PR #1, opened 2016-08-29
Add PDX Hierarchy Tree

inodb:
Add nodes to mock data, add React components to draw SVG Tree based on the
data. Eventually the nodes should be constructed from clinical data using
getTreeNodesFromClinicalData function.

### DIFF
--- a/src/features/patient_view/clinical_information/ClinicalInformationContainer.js
+++ b/src/features/patient_view/clinical_information/ClinicalInformationContainer.js
@@ -3,11 +3,12 @@ import { ClinicalInformationTable } from './ClinicalInformationTable';
 import { Link } from 'react-router';
 import { Tabs, Tab, ButtonGroup, Button } from 'react-bootstrap';
 import { default as $ } from 'jquery';
-import Spinner from 'react-spinkit';
+import PDXTree from './PDXTree'
+//import Spinner from 'react-spinkit';
 
-import styles from './test.scss';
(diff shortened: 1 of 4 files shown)


src/features/patient_view/clinical_information/ClinicalInformationContainer.js
```diff
@@ -130,7 +191,11 @@ export default class ClinicalInformationContainer extends React.Component {
                         data={ storeState.get('clinical_information').get('samples') }
                         title1="Attribute" title2="1202"
                     />
-
+                    <PDXTree
+                      width={800}
+                      height={300}
+                      nodes={ storeState.get('clinical_information').get('nodes').toJS() }
```
inodb: Maybe I shouldn't convert to JS object here?

src/features/patient_view/clinical_information/PDXTree.js
```diff
@@ -0,0 +1,116 @@
+import React from 'react'
+import SampleLabel from './SampleLabel'
+
+export default class PDXTree extends React.Component {
+  constructor(props) {
+    super(props)
+    this.renderNode = this.renderNode.bind(this)
```
alisman: you probably don't need to bind to this.  arrow functions (which you are using below) don't change this "this" value.

inodb: good point

PR #5, opened 2016-09-01
Add Mock data from new cBioPortal API

inodb:
Use mockdata from http://docs.cbioportal.apiary.io/ as data source for
ClinicalInformationTable.

### DIFF
--- a/src/features/patient_view/clinical_information/ClinicalInformationTable.js
+++ b/src/features/patient_view/clinical_information/ClinicalInformationTable.js
@@ -1,5 +1,6 @@
 import React, { PropTypes as T } from 'react';
 import { Table } from 'react-bootstrap';
+import Immutable from 'immutable';
 
 export class ClinicalInformationTable extends React.Component {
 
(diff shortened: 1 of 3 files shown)


src/features/patient_view/clinical_information/ClinicalInformationTable.js
```diff
@@ -8,32 +9,87 @@ export class ClinicalInformationTable extends React.Component {
...
+        } else {
+            return "No data to display";
+        }
+    }
+}
+
+export function convertPatientDataToTable(patient) {
+    let rv;
+
+    rv = patient.get('clinicalData').map(x => Immutable.List([x.get('name'), x.get('value')]));
```
alisman: We should probably change the format of the entrees in this list from array to key'd map. E.g. { name: "", value:"" }

inodb: As in the return value should be a List of maps? i.e.:
```
[{name:"",value:""},{name:"",value:""}]
```

alisman: Yes.  This is true also of the convertSampleDataToTable function below.  The code was very hard to read for two reasons, I think.  1. The format of the output was not human friendly (keyed objects) so it required that the reader understand the schema based on external knowledge.  2. maybe too much chaining and lookups.  I think in this case it would be better to forEach and build the desired output in a global variable for clarity.  Lets discuss.

PR #39, opened 2016-10-05
rename/reorg folders for clarity

alisman:
re name folders


adufilie:
no complaints about new folders

PR #40, opened 2016-10-05
added ts-loader to webpack config

adufilie:


webpack.config.js
```diff
@@ -110,8 +117,17 @@ var config = {
...
+        preLoaders: [
+            // All output '.js' files will have any sourcemaps re-processed by 'source-map-loader'.
+            {
+                test: /\.js$/,
+                loader: "source-map-loader"
+            }
+        ]
     },
     'postcss': [require('autoprefixer')],
+    'devtool': 'source-map',
```
alisman: in latest mast there's an optimization around sourcemap generation.
when we merge, lets make sure we retain it.

adufilie: I followed the instructions [here](http://www.typescriptlang.org/docs/handbook/react-&-webpack.html).  Is this line not required?

adufilie: What is the optimization?

adufilie:
removed the uses of es7 object rest spread because it's not supported by TypeScript yet and it causes errors to appear in IntelliJ

.eslintrc.json
```diff
@@ -8,7 +8,6 @@
 
     "parserOptions": {
         "ecmaFeatures": {
-            "experimentalObjectRestSpread": true,
```
inodb: Would it be possible to keep the spread syntax? I've used it a few times. Ok with tossing it if necessary

dist/index.html
```diff
@@ -5,10 +5,10 @@
...
     <script>window.defaultRoute = "/patient";</script>
     <script src="https://ajax.googleapis.com/ajax/libs/jquery/2.2.4/jquery.min.js"></script>
-<link href="/reactapp/css/styles.css" rel="stylesheet"></head>
+</head>
 <body class="cbioOnePageApp">
     <span id="reactHeader"></span>
     <div id="clinical_div"></div>
     <div id="reactRoot"></div>
-<script type="text/javascript" src="/reactapp/js/main.app.js"></script></body>
+<script type="text/javascript" src="http://localhost:3000/reactapp/js/main.app.js"></script></body>
```
inodb: why is this necessary? Windows thing? I think we should try not to add ports in the source file, since you should be able to override it with env variables on build. Don't feel particularly strongly about it tho, since this is just a dev file.

webpack.config.js
alisman: The devtool property is actually being overwritten down below on env specific basis.  As long as it's in dev, i'm fine with any sourcemap option that is relatively performant (on build).  i had chosen inline-sourcemap.  But not with any great reason.

dist/index.html
adufilie: I didn't do this on purpose.. Not sure what happened

.eslintrc.json
adufilie: Typescript does not allow it. It created ts errors in intellij

dist/index.html
adufilie: This is in the dist folder btw

inodb: 👌! Actually Travis should've caught this if it was only specific to your build env. Maybe something in build process that changed?

.eslintrc.json
inodb: Like nothing in TS works anymore, or it just gives some warning on that line in intellij?

adufilie: TS is unable to parse the file

PR #43, opened 2016-10-06
Add dist folder to repo 

inodb:
While waiting for npm/npm#3055 to be fixed add dist/ folder to repo. This will allow us to pull frontend compiled code in the parent cBioPortal project from github.

- Add test to Travis making sure the dist folder is the same as that produced by the
source code.
- Capitalize imports in  `routes.jsx`. OSX doesn't care about lower/uppercase, but Travis unix workers do.
- Add precommit hook running tests and adding the source code. Can be skipped with `--no-verify` or `-n`

Other **important** change requiring `.env` update see #44:

- Change __API_ROOT__ to include `api/`. New rc portal is using `api-legacy` for the old api, so this allows one to specify which one to use.


dist/package.json
```diff
@@ -0,0 +1,12 @@
+{
+	"name": "cbioportal-frontend",
+	"version": "0.0.3",
+	"description": "",
+	"author": "",
+	"license": "",
+	"pre-commit": [
+		"test",
```
inodb: You can always run these things yourself with `npm run test`, `npm run build` and then run `git commit -n` to ignore the precommit hook.

PR #52, opened 2016-10-14
Add mockData fetcher

inodb:
Add a mockData fetcher that pulls the mockData from a given API. This way the
data can stay up to date. Currently Travis-CI won't fail if the data is out of
sync. It will just give a warning.

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,6 +12,7 @@
     "lint": "eslint src/** --ext .js --ext .jsx --cache",
     "storedist": "npm run build && git add -f dist",
     "comparedist": "bash ./scripts/test_compare_dist_with_committed_dist.sh",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -12,6 +12,7 @@
     "lint": "eslint src/** --ext .js --ext .jsx --cache",
     "storedist": "npm run build && git add -f dist",
     "comparedist": "bash ./scripts/test_compare_dist_with_committed_dist.sh",
+    "syncmock": "node src/test/fetchMockData.js --diff",
```
pieterlukasse: how does this work? It compares the mock data with what?

inodb: Thanks for reviewing! It runs `git diff --quiet` on the file and looks at the exit code. This works because the mockData has been checked in. I followed the same strategy for the screenshot test.

PR #54, opened 2016-10-14
Add lint pre-commit hook and use babel-eslint

inodb:
Add lint to pre-commit hook. Ignore or fix all current lint issues. Use
babel-eslint because some files had parse errors.


alisman:
Just a reminder, we decided last week that since we are going to adopt TypeScript, eslint will not be compatible.  We will have to adopt TSLint instead.

inodb:
@alisman yeah that's a good point about switching to tslint. We can choose to:

1. switch all the files over to tsx, install tslint and call it a day
2. have a period of having all jsx/js/ts and only make new files in typescript. In that case I would argue for having both eslint & tslint in place for a while

I do think we should make the changes soon, since we're going to have a lot more people working on the project. It's a tricky merge now, but it'll be a bigger pain further down the line.

We can also decide not to enforce eslint at all, since we have codeclimate at least checking new changes. But I think it would be nicer to start out clean.

inodb:
We've added tslint. We're skipping eslint atm

PR #55, opened 2016-10-14
Fixed TS build issue in webpack.config.js and added VBox/HBox compone…

adufilie:
# What? Why?
- Fixed TS build issue in webpack.config.js (check es6 config here http://www.jbrantly.com/es6-modules-with-typescript-and-webpack/) 
- Added VBox/HBox components with updated sample code in HomePage.


src/shared/api/cbioportal-types.d.ts
```diff
@@ -0,0 +1,88 @@
+declare namespace cbio
```
inodb: I like this file. It's all for the old API I guess, but it's a good example imo

src/shared/components/flexbox/styles.js
```diff
@@ -0,0 +1,4 @@
+// this is a temporary solution to make TypeScript happy
+// there should be a way to resolve this the right way by modifying webpack.config.js
```
inodb: What's the problem exactly?

src/shared/lib/classNames.ts
```diff
@@ -0,0 +1,41 @@
+// Type definitions for classnames
```
inodb: Interesting, could be useful to mention this in some sort of style guide or the wiki

alisman:
I'm gonna boot this up so I can understand how they are being used but are FlexBox, VBox, HBox alternatives to Boostrap grid?

src/pages/home/HomePage.tsx
```diff
@@ -8,15 +9,44 @@ interface HomePageState
...
 export default class HomePage extends React.Component<HomePageProps, HomePageState>
 {
+	treeData:TreeStruct;
+
+	constructor(props:HomePageProps)
+	{
+		super(props);
+
+		// mock data for test
+		this.treeData = {label: "All", children: [
```
inodb: Once the mockData PR is merged #52, I think we should try to follow that pattern for getting mockData. But fine to code inline during dev

src/shared/components/flexbox/FlexBox.tsx
```diff
@@ -0,0 +1,55 @@
+/*
```
inodb: What is this for? To support use of flexbox?

src/shared/api/cbioportal-types.d.ts
adufilie: I wrote it to help myself understand the old api - we probably won't use it but I agree it's a good example.

src/shared/components/flexbox/FlexBox.tsx
adufilie: This is a file that contains the VBox and HBox components.  If you can think of a better name for the file, let me know. These VBox and HBox behave similar to their counterparts in Apache Flex.

adufilie: I've found VBox and HBox to be very useful for building UIs. If you can point me to a place using an alternative I'll try it out.

alisman: Lets look to see if this can be accomplished with Bootstrap grid.  we want to maintain a bias toward bootstrap for layout.

PR #61, opened 2016-10-21
Use new clinical-data endpoint from cbioportal-rc

inodb:
Use new clinical-data and samples endpoint from cbioportal-rc

Use clinical-data and samples endpoint from http://cbioportal-rc.herokuapp.com to get
clinical data instead of using old API wrapper. The JS client is automatically
generated by swagger-client.

Further

- rename clinical_div to clinical_div_prototype. We still have to fix some
  issues related to the tooltip before we can replace that in main cbioportal
  project.
- Add API_ROOT to .env file on Travis
- Use cbioportal-rc for mockData fetching. Should also use API_ROOT at some
  point
- Revert to using bootstrap table, since styling looks a better until we find a suitable table


adufilie:
[test](https://trello.com/c/OAP1Uh3q/2-test)

PR #64, opened 2016-10-26
Add pom.xml so jitpack can be used to download

inodb:
Remove dist folder from repo and all tests related to it. By adding a pom file
the dependencies can be downloaded through jitpack instead in the parent cBioPortal project.

See:

https://github.com/cBioPortal/cbioportal/pull/1844


adufilie:
Very nice!

PR #89, opened 2016-11-30
expose component renderers for legacy project

alisman:
# What? Why?
As we build components in the new project, we need a good way for code in the legacy project to mount the components in various places on the page.  We tried doing this from within the new project, but I realized this isn't sufficient.  For example, if we need to mount a component inside a jqueryUI tab in the legacy site.  We don't want to do that on page load, but when the tab is shown.  That's a key optimization and it would be a step backward to do otherwise.  So the legacy project needs to be able to control the mounting of the new components in some cases.  Be exposing render functions (which take the target node as arg) on the window object we can grant control to the legacy code.

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/shared/components/MSKPageDecorator/MSKPageDecorator.tsx
```diff
@@ -0,0 +1,13 @@
...
+    return class extends component<any, any> {
+        render() {
+            if ((window as any).devContext) {
+                return super.render();
+            } else {
+                return <span style={{ display:'none' }} />;
+            }
+        }
+    };
+};
```
adufilie: Why is this file prefixed with "MSK?"

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -1,16 +1,18 @@
...
 import PatientHeaderUnconnected from './patientHeader/PatientHeader';
 import {IPatientHeaderProps} from './patientHeader/PatientHeader';
 import {RootState} from "../../redux/rootReducer";
+import MSKPageDecorator from 'shared/components/MSKPageDecorator/MSKPageDecorator';
 
 interface IPatientViewPageProps {
     store?: RootState;
 }
 
+@MSKPageDecorator
```
alisman: This might be a little too fancy, but what this decorator does is disable the page components render function when we detect that we are in the legacy site context.  In this context, it's then at the discretion of the legacy code to invoke the renderers which we expose on the window object.   
What I like about this solution is that we still have page specific bundles which include all the components for a given page.   In dev context, we can render these as per nomal in the page component.  But in legacy project, that may not be appropriate because we want to render the components on various places in the page and at various times.  This solution allows this freedom.

src/shared/components/MSKPageDecorator/MSKPageDecorator.tsx
alisman: Good point.  No great reason.

src/shared/components/MSKPageDecorator/MSKPageDecorator.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+
+export default function (component: typeof React.Component): any {
```
adufilie: It doesn't matter much. The current decorator type restrictions are a little inconvenient. It wants the return type to be exactly the same as the parameter type.

src/pages/patientView/clinicalInformation/ClinicalInformationContainer.tsx
```diff
@@ -43,7 +44,7 @@ export default class ClinicalInformationContainer extends React.Component<IClini
                 return <div>{ this.buildTabs() }</div>;
 
             case 'error':
-                return <div>There was an error.</div>;
+                return <div>There was an loading error. Please try refreshing your browser.</div>;
```
inodb: *a loading

PR #106, opened 2016-12-09
proposal for get and observe

alisman:
# What? Why?
Components need to be able to fetch data and then observe change to that same data.   Rather than put the data in redux store, we allow components to subscribe to the QuerySession object. 

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/patientView/example.tsx
+++ b/src/pages/patientView/example.tsx
@@ -1,0 +1,64 @@
+import * as React from 'react';
+
+//// QuerySession.ts
+interface IExampleDataModel {
+    id:number;
+    type:string;
+}
+
+
+interface IExampleDataResponse {
+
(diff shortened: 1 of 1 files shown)


adamabeshouse:
Whereas this way, there has to be observer registration in any component that wants to use data, we could instead have only one object which observes, and keeps the data in the store fresh, and then use React and mapStateToProps to implement the observer pattern in components.

Why do you prefer the proposed approach?

adufilie:
I pushed some changes, adding missing parts

adufilie:
I've already solved this messy code issue in the past:
https://github.com/adufilie/WeaveJS/blob/namespaces/srcts/weavejs/util/ReactUtils.tsx#L339

PR #107, opened 2016-12-09
Query component

adufilie:
# What? Why?
Replacement query component which is used on the homepage and can also be embedded in other pages for modifying the query.

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/appShell/App/App.tsx
```diff
@@ -36,23 +40,21 @@ class App extends React.Component {
...
+        return <DevTools />;
       }
     }
   }
 
   render () {
      return (
        <Provider store={this.props.store}>
-         <div style={{ height: '100%' }}>
+         <div style={{ height: 700 }}>
```
alisman: why are we putting a fixed height on this?

src/appShell/App/Container.tsx
```diff
@@ -0,0 +1,44 @@
...
+            c => React.cloneElement(c as React.ReactElement<any>, childProps));
+    }
+    render() {
+    	let children = this.renderChildren();
+
+    	// don't create extra div elements if there is only one child
+    	if (children.length == 1)
+    		return children[0];
+
+		return <div style={{width:"100%", height: "100%", flex: 1}}>{children}</div>;
```
alisman: What is the reason for the height, width here?

src/pages/datasetView/DatasetPage.tsx
```diff
@@ -22,10 +24,10 @@ export default class DataSetPageUnconnected extends React.Component<IDatasetPage
         if (this.props.datasets) {
             const rows: Array<any> = [];
             this.props.datasets.forEach((item) => {
-                const tempObj: any = {
+				rows.push({
```
alisman: jaojoa will replace with unsafe with the custom column component

src/pages/patientView/clinicalInformation/SampleLabel.tsx
```diff
@@ -15,15 +22,13 @@ export default class SampleLabelSVG extends React.Component {
...
-    y: React.PropTypes.number.isRequired,
-};
+interface ISampleLabelHTMLProps {
+	label: string;
+	color: string;
+}
 
-export class SampleLabelHTML extends React.Component {
-    constructor(props) {
+export class SampleLabelHTML extends React.Component<ISampleLabelHTMLProps, void> {
```
alisman: curious why is the state interface set to void?

src/shared/api/memoize.ts
```diff
@@ -0,0 +1,155 @@
...
+		} as Transform;
+	}
+
+	get:Transform;
+}
+
+/**
+ * Creates a memoized version of a function.
+ */
+export default function memoize<F extends (...args:any[])=>any>(fn:F, getAdditionalMemoizeArgs:()=>any[]):F;
```
alisman: Could we use reselect instead of this?  Reselect is documented, tested and widely used.   https://github.com/reactjs/reselect.  I mean, if there's a good reason to write our own, then by all means.

src/appShell/App/App.tsx
adufilie: This was the easiest way to keep the contents on the screen - had a lot of trouble trying to make bootstrap do that. This change can be reverted.

src/appShell/App/Container.tsx
adufilie: Because I wanted it to fill the parent. Is there some problem?

src/pages/patientView/clinicalInformation/SampleLabel.tsx
adufilie: void, {}, null and undefined are pretty much the same

src/pages/patientView/patientHeader/PatientHeader.tsx
```diff
@@ -40,15 +34,15 @@ export default class PatientHeader extends React.Component<IPatientHeaderProps,
         );
     }
 
-    private getPopoverPatient(patient: typeof _ClinicalInformationData.patient) {
+    private getPopoverPatient(patient: ClinicalInformationData['patient']) {
         return patient && (
```
adufilie: TypeScript settings include strictNullChecks=true and the type of ClinicalInformationData['patient'] includes undefined. If we didn't have this check, it would crash when patient is undefined.

src/shared/api/memoize.ts
```diff
@@ -0,0 +1,155 @@
...
+		return cache;
+	}
+
+	// used to avoiding the varying-key-length limitation of the traverse() function below
+	private map_numArgs_cache = new Map<number, Cache<any, any>>();
+
+	// dual-purpose setter/getter
+	// note: does not work if subsequent calls vary the length of the keys array for the same cache param
+	private traverse(cache:Cache<any, any>, keys:any[], value?:T):T|undefined {
+		if (keys.length == 0)
```
adufilie: they are equivalent as long as the length value is a number.

src/shared/api/memoize.ts
adufilie: Reselect only saves the last result (does not cache all results), does not use weak references to parameter values, and has awkward syntax.

src/pages/patientView/clinicalInformation/getClinicalInformationData.ts
```diff
@@ -21,13 +21,13 @@ function transformClinicalInformationToStoreShape(patientId: string, studyId: st
...
     };
 
-    const samples: Array<ClinicalDataBySampleId> = chain(clinicalDataSample)
-        .groupBy('id')
-        .map((v, k) => ({
-            clinicalData: v,
-            id: k + '',
-        }))
-        .value();
+    const samples = _.map(
```
alisman: I actually find the initial chained code to be more readable.  It's nice to avoid nesting and its nice to have the type right next to the variable we are populating.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -0,0 +1,267 @@
+import * as _ from "lodash";
+import * as React from "react";
+import * as ReactDOM from "react-dom";
+import Dictionary = _.Dictionary;
+import {TypeOfCancer, CancerStudy} from "../../api/CBioPortalAPI";
+import {ITreeDescriptor, default as DescriptorTreeNode} from "../tree/DescriptorTreeNode";
+import * as ReactBootstrap from 'react-bootstrap';
+import {BootstrapTable} from "react-bootstrap-table";
```
adufilie: I had already started using it when it was discussed and it is easy to swap out later. I spent minimal time working with the library.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -0,0 +1,267 @@
...
+			if (this.props.onStateChange)
+				this.props.onStateChange({
+					selectedCancerType: this.selectedCancerType,
+					selectedStudies,
+				});
+		};
+	}
+
+    renderStudyName = (name:string, study:CancerStudy) => {
+    	let checked = !!_.find(this.selectedStudies, id => id == study.cancerStudyIdentifier);
```
alisman: as above, the == enables type coercion.  I've always understood that === is preferable unless we actually want coercion.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -0,0 +1,267 @@
...
+    		links.push({icon: 'book', url: `http://www.ncbi.nlm.nih.gov/pubmed/${study.pmid}`});
+    	if (study.cancerStudyIdentifier)
+    		links.push({icon: 'cube', url: `/study?id=${study.cancerStudyIdentifier}#summary`});
+
+		// put a space between links
+    	let children = [];
+		for (let link of links)
+		{
+			if (children.length)
+				children.push(<span>&nbsp;</span>)
```
alisman: lets use css for padding of this nature

src/pages/patientView/clinicalInformation/getClinicalInformationData.ts
adufilie: Me too but I was changing underscore code to lodash code and did not know the correct way to use chaining.

src/shared/components/query/CancerStudySelector.tsx
adufilie: In this case it makes no difference.

src/shared/components/query/styles.module.scss
```diff
@@ -0,0 +1,14 @@
+ul.cancerTypeTree {
+  color: black;/*test*/
+}
+
+ul.cancerTypeTree ul {
```
alisman: you could nest these children in the above definition to avoir repetition

src/shared/components/tree/DescriptorTreeNode.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+export default class DescriptorTreeNode extends React.Component<IDescriptorTreeNodeProps<any>, IDescriptorTreeNodeState>
+{
+	constructor(props:IDescriptorTreeNodeProps<any>)
+	{
+		super(props);
+		this.state = {};
+	}
+
+	onExpand = () =>
```
alisman: is this a way of getting around binding issue?

alisman: actually, it looks like this isn't used, right?  onExpand seems to come from props

src/shared/components/tree/DescriptorTreeNode.tsx
```diff
@@ -0,0 +1,104 @@
...
+        let content = this.props.treeDescriptor.getContent(this.props.tree);
+        let children = this.props.treeDescriptor.getChildren(this.props.tree);
+
+        let showArrow = !!children && children.length > 0;
+        let arrowStyle = showArrow ? undefined : {display: 'none'};
+        let arrow = (
+        	<FontAwesome
+        		name={expand ? 'caret-down' : 'caret-right'}
+        		style={arrowStyle}
+        		onMouseDown={this.onExpand}
```
alisman: hmmm. it is used here.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -0,0 +1,267 @@
...
+			dict_cancerTypeId_childCancerTypes,
+			dict_cancerTypeId_childStudies,
+			dict_cancerTypeId_flatStudiesList,
+			dict_studyId_study,
+		};
+	}
+
+	@memoizeWith(function() { return [this.props.cancerTypes, this.props.studies] })
+    getTreeDescriptor():ITreeDescriptor<TypeOfCancer|CancerStudy> & {rootNode: TypeOfCancer, setExpanded: (node:Node, value:boolean) => void} {
+		let {
```
alisman: some inconsistent indentation in this file.  that is one linting task i think is important.  otherwise things get very unreadable fast.  it's important that people be able to use the IDE reformat and that it aligns with linting

typings/missing.d.ts
```diff
@@ -5,3 +5,4 @@ declare module 'react-file-download';
 declare module 'react-zeroclipboard';
 declare module 'reactableMSK';
 declare module 'redux-seamless-immutable';
```
adufilie: noImplicitAny is very useful and should remain on, which means we need to declare these modules.

src/appShell/App/App.tsx
alisman: Yeah, I think this component itself should accept an optional fixed height prop, which would be put on its outermost el.

src/shared/components/query/CancerStudySelector.tsx
adufilie: indentation already addressed in latest commit

src/pages/patientView/patientHeader/PatientHeader.tsx
alisman: You mean runtime or ts? Is there a legitimate reason for patient to be undefined?  If not, it should crash or throw an error.  If so, then it should be handled in the control flow. Sorry if I'm misunderstanding.

src/shared/components/query/CancerStudySelector.tsx
adufilie: I am fine having a script that normalizes the indentation but it is anti-productive to worry about it while writing code.

src/shared/components/query/CancerStudySelector.tsx
alisman: ok.  admittedly the table component situation sucks.  we're going to have to use FB's fixed data table (in fact our enhancement of it) for tables where performance is a factor because of huge # of rows/columns.   it would be nice to keep this to a minimum.

src/shared/components/tree/DescriptorTreeNode.tsx
adufilie: This binding is required because the function is passed to a subcomponent.

src/pages/patientView/patientHeader/PatientHeader.tsx
adufilie: As I said above, TypeScript settings include strictNullChecks=true and the type of ClinicalInformationData['patient'] includes undefined. This means TypeScript displays an error if you do not check for undefined.

src/globalStyles/global.scss
```diff
@@ -15,4 +15,12 @@ th.reactable-header-sort-asc:after {
   content: '\25BC';
   padding-left: 10px;
   padding-bottom: 5px;
-}
\ No newline at end of file
+}
+
+/* makes react-bootstrap-table work in flexbox layout */
+.react-bs-table-container,.react-bs-table,.react-bs-container-body {
```
alisman: Putting aside whether we use react-bs-table or not, this would be better applied to the component instance itself.

src/appShell/App/Container.tsx
```diff
@@ -0,0 +1,44 @@
...
+        const {children} = this.props;
+        return React.Children.map(children,
+            c => React.cloneElement(c as React.ReactElement<any>, childProps));
+    }
+    render() {
+        let children = this.renderChildren();
+
+        // don't create extra div elements if there is only one child
+        if (children.length == 1)
+            return children[0];
```
alisman: We just need to set this back to what it was.  I don't think there is a reason that nested parent divs should cause a problem for the layout you are trying to achieve.   I'm happy to help address the underlying layout issue.

adufilie: Why do we need to change this back? The new code is better.  Why do you want two divs surrounding the child?

adufilie: Like I said, we can throw an error when there is more than one child if you prefer.
And here's a screenshot of what it shows with two wrapping divs: http://image.prntscr.com/image/7ef4209fefca47f88ce4b8ab5187c694.png

Here's a screenshot of what it shows with no wrapping div: http://image.prntscr.com/image/a966d92465154f93996fd90f6d39d8e9.png

src/globalStyles/global.scss
adufilie: i will move this to a better location

src/pages/home/HomePage.tsx
```diff
@@ -38,7 +46,10 @@ export default class HomePage extends React.Component<IHomePageProps, {}>
     public render()
     {
         return (
-       		<QueryContainer/>
+        	<FlexCol style={{height: '100%'}}>
+				<a style={{alignSelf: 'center'}} onClick={() => this.setState({ mode: this.state.mode == 2 ? 1 : 2 })}>Switch to {this.state.mode == 2 ? 'old' : 'new'} view</a>
```
alisman: is there a reason you're using double equals instead of the triple equals?    Also, did you manage to solve the setState ts issue that others have been encountering?

alisman: I see, in this case there's no problem because { mode } is entire state!

adufilie: == is just what I'm used to typing and it's what I type when it makes no difference.

alisman: It's been an established practice to use === by default.   In the spirit of typescript, it helps avoid errors.   You should use three unless you have a good reason.

adufilie: I don't have a good reason to use three. I know when three are necessary and I know in every case I use 2 it is ok to use 2. I never make errors with 3 versus 2 equals and typescript makes it even less necessary to use 3.

alisman: Never?! That's amazing!   Not you, but other developers (like me) sometimes do make mistakes.  So for us, every js book in existence recommends === unless we specifically want type coercion.  That's the best practice.  I suppose you are welcome to ignore it?

adufilie: There are very few situations where it matters, and using === everywhere without thinking about it or testing it actually has the potential to create bugs.  I am aware of every situation where it matters and I know that every place I have used == is a place that it does not matter.

adufilie:
Cancer study selector/study list still needs to be refactored to fix a performance issue. Also, still TODO:

- [ ] Submit button validation not done (does not make sure user has selected required fields)
- [ ] Search keyword OR/quotes don’t work yet
- [ ] New cancer study selector does not have the popup info (with clickable links) 
- [ ] Fix performance issue in cancer study selector
- “Send to GenomeSpace” button
  - [ ] is hidden by submit button styling
  - [ ] doesn’t work
  - [ ] does not use the image as seen on cbioportal.org

src/shared/api/genomespace/gsuploadwindow.ts
```diff
@@ -0,0 +1,89 @@
...
+    // expect an object with url, filename (optional), successCallback(optional), errorCallback (optional)
+
+    let gsUploadUrl = jsuiRoot + "/upload/loadUrlToGenomespace.html?uploadUrl=";
+    let dest = encodeURIComponent(config.url);
+    let filenameParam = "";
+    if (config['filename'] != null) {
+        filenameParam = "&fileName=" + config.filename;
+    }
+
+    let newWin:Window & any = window.open(gsUploadUrl + dest + filenameParam, "GenomeSpace Upload", "height=340px,width=550px");
```
adamabeshouse: wouldn't `& any` annihilate any type specificity?

src/shared/components/StudyList/StudyList.tsx
```diff
@@ -0,0 +1,197 @@
...
+import * as styles_any from './styles.module.scss';
+import classNames from "../../lib/classNames";
+import FontAwesome from "react-fontawesome";
+import LabeledCheckbox from "../labeledCheckbox/LabeledCheckbox";
+import {observer} from "mobx-react";
+import {getStudySummaryUrl, getPubMedUrl} from "../../api/urls";
+import {QueryStoreComponent} from "../query/QueryStore";
+
+const styles = {
+	...styles_any as {
```
adamabeshouse: can you explain this syntax?

src/shared/components/StudyList/StudyListLogic.ts
```diff
@@ -0,0 +1,372 @@
...
+			let clauseMatch = true;
+			for (let phrase of clause.data)
+			{
+				clauseMatch = clauseMatch && matchPhrase(phrase, node);
+			}
+			match = match || clauseMatch;
+		}
+	}
+	return {result: match, forced: forced};
+};
```
adamabeshouse: take out commented code?

src/shared/components/flexbox/FlexBox.tsx
```diff
@@ -0,0 +1,53 @@
...
+import classNames from "shared/lib/classNames";
+
+export interface IFlexBoxProps <T> extends React.HTMLProps<T>
+{
+	flex?:number;
+	padded?:boolean;
+	overflow?:boolean;
+}
+
+export class IFlexBoxProps<T>
```
adamabeshouse: whats going on here with the class and interface with the same name?

src/shared/components/query/GisticGeneSelector.tsx
```diff
@@ -0,0 +1,214 @@
...
+{
+	constructor(props: GisticGeneSelectorProps)
+	{
+		super(props);
+		this.map_geneSymbol_selected.replace(props.initialSelection.map(geneSymbol => [geneSymbol, true]));
+	}
+
+	map_geneSymbol_selected = observable.map<boolean>();
+
+	private columns:IColumnDefMap = {
```
adamabeshouse: we'll update this to interface with the new table

src/shared/api/genomespace/gsuploadwindow.ts
adufilie: Yes, you're right.  This is third party code and all I wanted to do was make it compile.

src/shared/components/StudyList/StudyList.tsx
adufilie: Because I'm adding properties below, I'm using the splat operator to copy the properties from styles_any, and I'm assigning strong typings.

src/shared/components/StudyList/StudyListLogic.ts
adufilie: This file is currently being refactored

src/shared/components/flexbox/FlexBox.tsx
adufilie: Interfaces and types are on a different plane of existence, so the names can overlap. I'm using the class to define a static function that can be accessed like IFlexBoxProps.renderBox(). It doesn't have to be defined this way. If you think it's important, I can change it.

adufilie: Interfaces and types are on a different plane of existence compared to javascript variables, so the names can overlap. I'm using the class to define a static function that can be accessed like IFlexBoxProps.renderBox(). It doesn't have to be defined this way. If you think it's important, I can change it.

adamabeshouse: I guess I don't see the point, it seems sort of confusing. But if you say it makes something cleaner then I believe it.

adufilie: It's just developer preference. TypeScript type variable versus JavaScript variables can be confusing in general. The shape of this code is the result of several refactorings. I can simplify it.

pieterlukasse:
Hi @adufilie, @jjgao , I am trying to test the heroku deployment but it only displays an empty screen. Any ideas what might be wrong?

adufilie:
Try this URL: https://cbioportal-frontend-pr-107.herokuapp.com/#/home

pieterlukasse:
@adufilie I can see the UI now, thanks.

pieterlukasse:
I found an issue in the OQL part, but I am not sure if I can expect this to work here. Is this heroku deployment an integrated deployment? If not, can we also see this query UI in the context of the rest of the portal? Or is this only coming later? For example, I found this issue:

image

Here the OQL part seems to fail and it tells me to click the "Submit" button, but this button is disabled.

pieterlukasse:
Another issue I found: the "DESELECT ALL" button only deselects within the current cancer type. Would be good to have a deselect all that does this across all types.

image

adufilie:
@pieterlukasse OQL / submit button enabling issue fixed.  Deselect all button intentionally deselects only those that are shown.  Alternatively you can click on the link that says how many are selected, which shows all selected without filtering and allows you to deseslect all from there.  You can discuss this with @ecerami.

pieterlukasse:
Hi @adufilie thanks for fixing it. 

While testing, I noticed another issue. Apparently if I type too fast, the OQL validation gets confused about what is valid or not. Here I managed to get a valid OQL expression to be considered invalid:

image

while if I remove all and paste same contents again, it is valid:

image

adufilie:
@pieterlukasse This was a recent regression involving upper- vs. lower-case in gene validation.  It's fixed now.

Issue #93, opened 2016-12-05
React JS table for the core-mutation columns for mock json data in the cbioportal-frontend repo


inodb:

onursumer:
@alisman @inodb was this the react table you were talking about: https://github.com/tannerlinsley/react-table

alisman:
Hi Onur, 
No, we settled on http://glittershark.github.io/reactable/
Unfortunately, it was missing a key feature, so we forked it.  You can use it as follows:

import {Table as DataTable} from "reactableMSK";

onursumer:
@alisman @inodb  here is my generic data table proposal:

- Initialize the table with a certain set of columns (example: initialize with core mutation data)
- Render the initial data.
- Fetch the data for additional columns on demand -- Additional data can be retrieved either from the server side, or via a cross-domain API call.
- Once the data is ready for a specific column, render the actual column content.

Let's say we want to render cosmic data on demand as an additional column, and assume that we have defined `CosmicCell` as: `columnFormatters={ { 'cosmic': CosmicCell } }`.

If we want to fetch the cosmic data in the `CosmicCell` component, we need the core mutation data accessible to that component in order to query the cosmic API endpoint.

Alternatively, we can keep `CosmicCell` simple. For example, if there is no data for cosmic column it renders a placeholder image, and when the data becomes available it replaces it with the actual content. In that case we need to deal with data fetching on a higher level.

Which one do you think is a better way to handle the additional data sources?

In any case, we need to provide a mechanism to add a custom column by defining a data fetcher component/function, and a renderer component/function. The data fetcher should be able to use the existing initial data for API queries.

PR #108, opened 2016-12-12
Mutation Table and Enhanced React Table components

onursumer:
Sample implementation of two react components `EnhancedReactTable` and `MutationTable`, built on `reactableMSK` component.

Please do not merge the PR, work is not complete yet. This is just a proof of concept.

@cBioPortal/frontend: Let me know if this design/implementation does not meet our current refactoring goals. Related issue: #93


src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    public static getText(proteinChange:string):string
+    {
+        return proteinChange;
+    }
+
+    public static getValue(data:IColumnFormatterData):string
+    {
+        let value:any;
```
alisman: probably can make this type string since we're returning it and method promises to return string

src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
```diff
@@ -0,0 +1,132 @@
+import * as React from 'react';
+import {Table as DataTable} from "reactableMSK";
```
alisman: If this feature will ultimately need to scale to handle large data sets, we should use FBs FixedDataTable, unless we're ok with pagination.

src/shared/components/enhancedReactTable/IColumnFormatterProps.ts
```diff
@@ -0,0 +1,29 @@
+/**
+ * @author Selcuk Onur Sumer
+ */
+export interface IColumnFormatterProps {
```
alisman: Typescript has namespaces, which would allow us to group these interfaces and tie them to the MSKTable, or EnhancedMSKTable as the case may be!

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,124 @@
...
+        return(
+            <EnhancedReactTable
+                reactTableProps={reactTableProps}
+                columns={columns}
+                rawData={rawData}
+            />
+        );
+    }
+
+    private mergeProps(props:IEnhancedReactTableProps):IEnhancedReactTableProps
```
alisman: I believe React has built in defaultProps mechanism:

https://facebook.github.io/react/docs/react-component.html#defaultprops

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
onursumer: @alisman Good catch. I think this method should be renamed as `getText` and we should make sure that it always returns string. In its current form it doesn't look safe to me.

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+            rawData: [],
+            columns: MutationTable.defaultColumns,
+            reactTableProps: {
+                className: "table",
+                id: "table"
+            }
+        };
+
+        // merge provided props with the default props
+        return lodash.merge(defaultProps, props);
```
onursumer: @adufilie Seems like object spread overrides existing nested objects (i.e: shallow merge). What I want is to merge
```
{
   columns: {
      a: {...},
      b: {...}
   }
}
```
with
```
{
   columns: {
      c: {...}
   }
}
```
and get
```
{
   columns: {
      a: {...},
      b: {...},
      c: {...}
   }
}
```
But with object spread, I only get
```
{
   columns: {
      c: {...}
   }
}
```

Am I missing something?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -36,6 +37,13 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
             );
         }
 
+        let mutationDiv: Element | null = document.getElementById('mutations_div_prototype');
```
alisman: this isn't necessary.  i can walk you through how to get the component on the page/expose it to legacy project.

src/shared/components/enhancedReactTable/IEnhancedReactTableProps.tsx
```diff
@@ -0,0 +1,24 @@
+import {IColumnFormatter, IColumnSortFunction, IColumnFilterFunction} from "./IColumnFormatterProps";
```
alisman: this file doesn't have any jsx in it so it should be a .ts file, just so compiler doesn't have waste time looking for jsx

src/shared/components/mutationTable/column/GeneColumnFormatter.tsx
```diff
@@ -0,0 +1,81 @@
...
+    {
+        return GeneColumnFormatter.getTextValue(data);
+    }
+
+    /**
+     * Default text value for a gene is its hugo gene symbol.
+     *
+     * @param data  column formatter data
+     * @returns {string}    hugo gene symbol
+     */
```
alisman: @adufilie would it be worth making a type hugoSymbol so this would be self documenting? (where hugoSymbol was just a string)

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -0,0 +1,174 @@
...
+ */
+export default class MutationAssessorColumnFormatter extends React.Component<IColumnFormatterProps, {}> implements IColumnFormatter
+{
+    /**
+     * Mapping between the functional impact score (data) values and
+     * view values.
+     */
+    public static get MA_SCORE_MAP():{[key:string]:any} {
+        return {
+            h: {label: "High", style: "oma-high", priority: 4},
```
alisman: typescript has enums.  would be good to have the style constants in an enum so they can be resused (e.g. down below)

src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
```diff
@@ -0,0 +1,132 @@
...
+    }
+
+    private convertToTableData(columns:IColumnDefMap, rawData:Array<any>):Array<any>
+    {
+        let rows:Array<any> = [];
+
+        _.each(rawData, function(element:any) {
+            let row:any = {};
+
+            _.each(columns, function(columnDef:any) {
```
alisman: I think maybe a small change to MSKReactable would make this much easier.   The column formatter component should be passed these three properties automatically: tableData, rowData, and columnData.   That way you don't have to do all this work to put references to the parent data structures in each cell data object.  I'm happy to make this change if you agree that it makes sense.

src/shared/components/mutationTable/MutationTable.tsx
adufilie: Just realized the defaultColumns is a getter - which means the in-place modification would not matter in this case.

src/shared/components/mutationTable/column/GeneColumnFormatter.tsx
adufilie: The Swagger API does that. It wouldn't hurt.

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -0,0 +1,157 @@
...
+        // let matched = proteinChange.match(/.*[A-Z]([0-9]+)[^0-9]+/);
+        let alleleAndPosition:RegExp = /[A-Za-z][0-9]+./g;
+        let position:RegExp = /[0-9]+/g;
+
+        // first priority is to match values like V600E , V600, E747G, E747, X37_, X37, etc.
+        let matched:RegExpMatchArray|null = proteinChange.match(alleleAndPosition);
+        let buffer:Array<number> = [];
+
+        // if no match, then search for numerical (position) match only
+        if (!matched || matched.length === 0)
```
adufilie: correct

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
adufilie: Here's a start:
```
type MA_IMPACT = 'h' | 'm' | 'l' | 'n';
type MA_STYLE = 'oma-high' | 'oma-medium' | 'oma-low' | 'oma-neutral';
type IMutationAssessorFormat = {label: string, style: MA_STYLE, priority: number};
```

and this function signature can then be:
```
public static get MA_SCORE_MAP():{[key:OMA_IMPACT]: IMutationAssessorFormat}
```

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -0,0 +1,174 @@
...
+            return "oma-link";
+        }
+        else {
+            return "";
+        }
+    }
+
+    public static getMapEntry(data:IColumnFormatterData):any // IMutationAssessorFormat?
+    {
+        let entry:any = null;
```
adufilie: There are very few situations where I would recommend using `any` type.  It's better to leave the type unspecified and let TypeScript figure it out through code flow analysis.

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
onursumer: Thanks @adufilie, this looks a lot better than `[key:string]:any`

onursumer: @adufilie when I set the key as a typed parameter (in this case `MA_IMPACT`), I get this TS error:
`An index signature parameter type must be 'string' or 'number'`

src/pages/patientView/mutation/mock/mutationData.json
```diff
@@ -0,0 +1,150 @@
+[
```
inodb: Let's explicitly put the endpoint for this data in here: https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/test/mockDataURLs.json. Set the flag `implemented: false`. Once the endpoint is implemented we can generate new mock data and set it to true, so we can make sure the mock data stays in sync.

src/pages/patientView/PatientViewPage.tsx
inodb: @alisman Ah my bad, I told @onursumer to do it like this. It's the same as we did before afaik. What is the new approach?

onursumer:
silently merged into integration branch, closing.

PR #109, opened 2016-12-12
Patient View - Move clinicalAttributesUtil

inodb:
Move clinicalAttributesUtil from parent project to frontend.

Still requires
- [ ] Tests
- [ ] Maybe rewrite in TS?
- [ ] Maybe just call when needed for now, instead of changing store (since it's only used at one location)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -158,7 +158,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
     public render() {
 
         let sampleManager: SampleManager | null = null;
-        let sampleHeader: JSX.Element[] | null = null;
+        let sampleHeader: JSX.Element[] | null | undefined = null;
```
alisman: what was reason for adding undefined?

PR #110, opened 2016-12-13
sandbox the new implementation of bootstrap so that it applies only w…

alisman:
# What? Why?
We want to have a fresh copy of bootstrap in the context of new components we are introducing.  We want to try to insulate the new markup from customizations to bootstrap in the legacy project.  To solve this, we build bootstrap with a wrapper class .cbioportal-frontend so that the new bootstrap styles will only appy when that class is applied to the root of a component

# Checks
  the `-s` option)


package.json
```diff
@@ -139,8 +140,5 @@
     "tslint-react": "^1.0.0",
     "webpack-dev-server": "^1.15.0",
     "yargs": "^4.8.0"
-  },
-  "pre-commit": [
```
alisman: just driving me crazy

alisman: we can restore it when we settle on linting rules

my-index.ejs
```diff
@@ -9,8 +9,8 @@
         // Set default API in case none is specified in .env
         __API_ROOT__ = 'cbioportal-rc.herokuapp.com/api';
     </script>
-    <script src="reactapp/js/common.bundle.js"></script>
-    <link href="reactapp/css/bootstrap.min.css" rel="stylesheet"></link>
+    <script src="reactapp/common.bundle.js"></script>
+    <link href="reactapp/prefixed-bootstrap.min.css" rel="stylesheet"></link>
```
inodb: Is it essential to change these paths? Might be better to keep them the same, so in the parent cBioPortal project we don't have to change anything.

package.json
inodb: I prefer to keep it in, since you can just ignore it with `-n` if you are certain you don't want the lint checking, seems like a pain update files afterwards, but your call.

webpack.config.js
```diff
@@ -47,9 +47,9 @@ var config = {
     ],
     'output': {
         path: './dist/',
-        filename: 'reactapp/js/[name].app.js',
-        chunkFilename: 'reactapp/js/[name].chunk.js',
-        cssFilename: 'reactapp/css/app.css',
+        filename: 'reactapp/[name].app.js',
+        chunkFilename: 'reactapp/[name].chunk.js',
+        cssFilename: 'reactapp/app.css',
```
inodb: As mentioned above, maybe better to keep current directory structure to prevent having to update parent project, but again your call.

PR #114, opened 2016-12-16
VAF plot wrapped into react component

adamabeshouse:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -57,9 +60,10 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
     }
 
     public render() {
-        return (
-            <div>
-                <ClinicalInformationContainer />
+        return (<div>
+            <ClinicalInformationContainer />
+                <ThumbnailExpandVAFPlot data= { vafPlotMockData } colors ={ vafPlotMockColors } labels= { vafPlotMockLabels }
```
alisman: minor,  but kinda strange spacing around those attibutes and = sings

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
```diff
@@ -0,0 +1,62 @@
...
+            show_controls: true,
+            nolegend: false,
+            init_show_histogram: true,
+            init_show_curve: true,
+        };
+
+        return (<span>
+            <Tooltip placement={this.props.overlayPlacement} trigger={['hover', 'focus']} overlay={<VAFPlot {...expandedProps}/>}
+                    arrowContent={<div className="rc-tooltip-arrow-inner"></div>}
+                    destroyTooltipOnHide={false}>
```
alisman: i think destroy:true is generally a good idea unless there's some big cost in reproducing the tooltip

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
```diff
@@ -0,0 +1,62 @@
...
+                <span><VAFPlot key="thumbnail" ref="thumbnail" {...thumbnailProps}/></span>
```
alisman: what's reason for wrapping span here?

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+import Tooltip from 'rc-tooltip';
+import 'rc-tooltip/assets/bootstrap_white.css';
+import {VAFPlot, IVAFPlotProps} from './VAFPlot';
+
+export type IThumbnailExpandVAFPlotProps = {
```
alisman: should all of these be optional? seems like data at least is necessary

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+            } else {
+                $(this.div).find('.viz_curve').hide();
+            }
+        }
+
+        return (nextState.show_histogram !== this.state.show_histogram) || (nextState.show_curve !== this.state.show_curve);
+    }
+
+    toggleShowHistogram(evt) {
+        // setState not necessarily synchronous, so to send the new value to callback we need to do this
```
alisman: not sure what this comment means

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+        this.setState({show_curve: new_show_curve});
+    }
+
+    render() {
+        let controls = [<label key="histogram-toggle"><input ref={(input) => { this.histogram_checkbox = input; }}
+                                      type="checkbox" checked={ this.state.show_histogram } onChange= { this.toggleShowHistogram }/>histogram</label>,
+                        <label key="curve-toggle"><input ref={(input) => { this.curve_checkbox = input; }}
+                                      type="checkbox" checked={ this.state.show_curve } onChange = { this.toggleShowCurve } />density estimation</label>];
+        return (
+            <div style={{display: "inline"}} ref={(div) => { this.div = div; }}>
```
alisman: layout conditions seems kinda funky.   maybe just avoid rendering div when there are no controls.  also, why inline but then <br> tag?

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+                { ( this.props.show_controls ? controls : [] ) }
+                {( this.props.show_controls ? <br/> : [])}
+            </div>
+        );
+    }
+}
+
+VAFPlot.defaultProps = {
```
alisman: does this work with typescript?  I think i tried it and was getting ts error.  
I think i found that with ts, default props can be handled with static member of component class:

 public static defaultProps: PageProps = {
        foo: "default"
    };

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+        this.toggleShowCurve = this.toggleShowCurve.bind(this);
+    }
+
+    componentDidMount() {
+        AlleleFreqPlotMulti(this, this.props.data, this.state.options, this.props.order, this.props.colors, this.props.labels);
+    }
+
+    shouldComponentUpdate(nextProps, nextState) {
+        // Hack around React: handle updating here, and return false
+        if (this.histogram_checkbox) {
```
alisman: if we determine that rendering AlleleFreqPlotMulti is not prohibitively expensive, we could err on the side of just fully re-rendering it (i.e. in render method).  That way we can avoid these kind of shenanigans and be doing real React.   Basically my point is that we should actually be biased toward real React flow in cases where we're certain there isn't a performance issue and shouldComponentUpdate conditions are simple.  We can discuss.

adamabeshouse: good idea

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: good call

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
adamabeshouse: done

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
adamabeshouse: oops, legacy

src/pages/patientView/vafPlot/ThumbnailExpandVAFPlot.tsx
adamabeshouse: good point, data should be necessary

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -0,0 +1,131 @@
...
+                top: this.props.margin_top,
+                left: this.props.margin_left,
+                right: this.props.margin_right,
+                bottom: this.props.margin_bottom
+            },
+            width: this.props.width,
+            height: this.props.height
+        };
+        // margin bottom must be computed before height because height computation makes use of it
+        options.margin.bottom = (isNaN(options.margin.bottom) ? (30 + (label_dist_to_axis / 2)) : options.margin.bottom);
```
adamabeshouse: these are used here actually, it's one of the ways we set up the thumbnail and the expanded version to be different.

But I've replaced isNaN with an undefined check.

src/pages/patientView/vafPlot/VAFPlot.tsx
adamabeshouse: oof, more legacy/dev code. sorry

src/pages/patientView/vafPlot/VAFPlot.tsx
adamabeshouse: The imperative rendering method `AlleleFreqPlotMulti` will populate this `div`

src/pages/patientView/vafPlot/VAFPlot.tsx
adamabeshouse: good idea, I'll do this

adamabeshouse: but if I type the variable (`: PageProps`) then I have to include values for mandatory members, which is kind of the opposite of defaultProps haha.. so I'm leaving it untyped

src/pages/patientView/vafPlot/VAFPlot.tsx
alisman: Sorry, I meant is it likely to be used anywhere else?

src/pages/patientView/vafPlot/VAFPlot.tsx
adufilie: Use Partial<PageProps>

src/pages/patientView/vafPlot/VAFPlot.tsx
adamabeshouse: I'm saying that it's not likely to be used elsewhere on the site, but it's used twice, once for the thumbnail, once for the expanded view, and these options need to be configurable to configure each of those cases.

PR #127, opened 2017-01-20
Styling clinical attributes at top of patient view page

inodb:
Add old clinical attributes util function to style clinical attributes at the top of page. This is being done using CSS selectors by adding the clinical attribute info to the elements. Reasoning at the time was, that replacing a CSS file is easier for other institutions to replace and less error prone. 

Some data normalization is performed in these functions as well that should preferably happen in the transformation step before import. Not much effort has been made therefore to improve these functions on the frontend.

screen shot 2017-01-20 at 12 04 05 pm

### DIFF
--- a/src/pages/pageHeader/styles.scss
+++ b/src/pages/pageHeader/styles.scss
@@ -1,7 +1,7 @@
 .cbioOnePageApp {
 
 
-  padding: 90px 20px 20px 20px;
+  padding: 99px 20px 20px 20px;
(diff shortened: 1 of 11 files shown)


inodb:
[Refactor top part of patient view (styling of clinical attributes)](https://trello.com/c/uqCBPj6Y/32-refactor-top-part-of-patient-view-styling-of-clinical-attributes)

PR #129, opened 2017-01-23
Added Cancer Hotspots and 3D Hotspots support for the annotation column

onursumer:
# What? Why?
- Added Cancer Hotspots and 3D Hotspots icons for the annotation column.
- Using an auto-generated Cancer Hotspots API client to retrieve the data on-the-fly from external Cancer Hotspots and 3D Hotspots web services.

# Checks
  the `-s` option)


src/shared/api/CancerHotspotsAPI.ts
```diff
@@ -0,0 +1,1044 @@
+import * as request from "superagent";
+
+type CallbackHandler = (err: any, res ? : request.Response) => void;
+export type HotspotMutation = any;
+export type SingleResidueHotspotMutation = any;
+export type IntegerRange = any;
+export type TumorTypeComposition = any;
+export type Cluster = any;
+export type ClusteredHotspotMutation = any;
```
onursumer: Seems like auto generator misses the actual field names for these types.
@adufilie : Do you have any idea why the swagger code generator ended up declaring these types as `any`?

onursumer:
[Integrate 3D Hotspots into Mutation Table](https://trello.com/c/ZU36oZxN/52-integrate-3d-hotspots-into-mutation-table)

onursumer:
[Integrate Mutation Hotspots into the Mutation Table [Uses External API]](https://trello.com/c/4bUUfoQ9/19-integrate-mutation-hotspots-into-the-mutation-table-uses-external-api)

src/shared/api/CancerHotspotsAPI.ts
alisman: I checked the swagger doc and it looks like there are some errors in it.  could be why the swagger generator fails even though the types do seem like the are specified.

alisman: Hmmm.  The online client generator does seem to work for the current swagger doc, so it might be our generator.   Will have to experiment more

src/pages/patientView/mutation/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,279 @@
...
+            if (is3dHotspot) {
+                publication.push("and ");
+            }
+        }
+
+        if (is3dHotspot) {
+            publication.push("a 3D clustered hotspot ");
+        }
+
+        publication.push("in a population-scale cohort of tumor samples of " +
```
alisman: use es6 string literals for multi-line strings

src/pages/patientView/mutation/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,279 @@
...
+            "various cancer types using methodology based in part on ");
+
+        if (isHotspot)
```
alisman: this concatenation could be handled using es6 template literals.  could be a little cleaner

onursumer: @alisman updated the tooltip generating code, no more string concat or arrays. Looks a bit cleaner now.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -43,6 +45,7 @@ interface IPatientViewState {
     mutationData: any;
     mrnaExprRankData?: MrnaRankData;
     activeTabKey: number;
+    hotspotsData: any;
```
adamabeshouse: how do you feel about the convention I started w MrnaRankData - define the type?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -93,6 +102,47 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+    fetchHotspotsData(mutations:Mutation[]) {
+        const generateMap = function(hotspots:HotspotMutation[]) {
+            // key => geneSymbol_proteinPosition
+            // protienPosition => start[_end]
+            const map: {[key:string]: boolean} = {};
+
+            // create a map for a faster lookup
+            _.each(hotspots, function(hotspot:HotspotMutation) {
+                const positions = hotspot.residue.match(/[0-9]+/g) || []; // start (and optionally end) positions
+                const key = [hotspot.hugoSymbol.toUpperCase()].concat(positions).join("_");
```
adamabeshouse: just confirming this will always be in the order you want

src/pages/patientView/PatientViewPage.tsx
onursumer: Good point!

src/pages/patientView/PatientViewPage.tsx
onursumer: I was assuming the match function should return the array in the correct order, but haven't really checked the documentation.

src/pages/patientView/PatientViewPage.tsx
onursumer: @alisman @adamabeshouse : Btw, we need this additional `HotspotData` mapping just because we don't have a dynamic mutation model. If it was possible to just attach the hotspot information to the mutation itself, say like `mutation.isHotspot = true;` then we wouldn't need to create such an additional data mapping.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: That and the amino acid codes too

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: @onursumer in my mind the jury is still out on whether that's preferable to indexing into a columnProps property. I kind of like columnProps because then we don't have to change the entire data (and thus, I think, force regenerate all the rows) when new data comes in.

alisman: the vdom will always be re-generated whenever there's a change to data either way.

onursumer: @adamabeshouse @alisman just realized that I already have a type defined `IHotspotData`, I just forgot to use that type in `PatientViewPage`. Looks like this PR is already merged, I will create another PR with some other minor improvements as well.

PR #135, opened 2017-01-25
Add sample label coloring to SampleInline

inodb:
Add coloring to header and mutation table sample labels based on DERIVED_NORMALIZED_CASE_TYPE clinical attribute.


inodb:
[Sample label coloring in mutation table](https://trello.com/c/lemE6RwS/69-sample-label-coloring-in-mutation-table)

PR #144, opened 2017-01-31
Show/Hide bug fix

onursumer:
# What? Why?
Fixes and improves the show/hide feature behavior.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
+++ b/src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
@@ -6,22 +6,12 @@ import * as _ from 'lodash';
 import TableHeaderControls from "shared/components/tableHeaderControls/TableHeaderControls";
 import {
     IEnhancedReactTableProps, IColumnDefMap, IEnhancedReactTableColumnDef, IColumnVisibilityState,
-    IEnhancedReactTableState, IColumnVisibilityDef
+    IEnhancedReactTableState, IColumnVisibilityDef, IColumnSort, IColumnFilter
 } from "IEnhancedReactTableProps";
 import {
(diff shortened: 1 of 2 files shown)


onursumer:
[(3) fix show/hide](https://trello.com/c/6TQmq9rC/86-3-fix-show-hide)

onursumer:
@alisman @adamabeshouse The fix is actually one line to make the show/hide work again: https://github.com/cBioPortal/cbioportal-frontend/pull/144/files#diff-dd74d85f67f0228723795bf4b6f21e1dR497

But I also made some improvements. Basically, now the `generateRows` method is called only when the column definitions are updated (since we are passing additional data via `columnProps` within the column definition). Change in `columnVisibility` now only affects `generateHeaders`, not `generateRows` anymore. Updating the `Thead` is sufficient to show/hide columns, since reactable is smart enough to ignore `Td` components with an unknown `column` value.

src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
```diff
@@ -106,33 +93,42 @@ export default class EnhancedReactTable<T> extends React.Component<IEnhancedReac
...
 
     constructor(props:IEnhancedReactTableProps<T>)
     {
         super(props);
 
+        this.colNameToId = this.mapColNameToId(props.columns);
+
         this.state = {
             columnVisibility: EnhancedReactTable.resolveVisibility(props.columns, props.rawData),
+            columnSort: this.resolveSortable(this.props.columns || {}),
```
adamabeshouse: this name doesn't make sense to me/is sort of vague. It's not quite a verb, not quite a noun. I think it should be sortableColumns or something. same with columnFilter

onursumer: okay, let's change to `sortableColumns` and `filterableColumns`.

PR #154, opened 2017-02-01
Added COSMIC column with tooltip

onursumer:
# What? Why?
Added COSMIC column with tool tip

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -22,8 +22,6 @@ import SelectCallback = ReactBootstrap.SelectCallback;
 import {
     default as CancerHotspotsAPI, HotspotMutation
 } from "../../shared/api/CancerHotspotsAPI";
-import CancerHotspotsAPI from "../../shared/api/CancerHotspotsAPI";
-import {HotspotMutation} from "../../shared/api/CancerHotspotsAPI";
 import {
     MutSig, MrnaPercentile, default as CBioPortalAPIInternal,
     VariantCountIdentifier, VariantCount, CosmicMutation
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -19,14 +19,18 @@ import { If, Then, Else } from 'react-if';
 import queryString from "query-string";
 import SampleManager from './sampleManager';
 import SelectCallback = ReactBootstrap.SelectCallback;
+import {
+    default as CancerHotspotsAPI, HotspotMutation
+} from "../../shared/api/CancerHotspotsAPI";
 import CancerHotspotsAPI from "../../shared/api/CancerHotspotsAPI";
 import {HotspotMutation} from "../../shared/api/CancerHotspotsAPI";
 import {
     MutSig, MrnaPercentile, default as CBioPortalAPIInternal,
-    VariantCountIdentifier, VariantCount
+    VariantCountIdentifier, VariantCount, CosmicMutation
 } from "../../shared/api/CBioPortalAPIInternal";
 import PatientHeader from './patientHeader/PatientHeader';
 import {TablePaginationControls} from "../../shared/components/tablePaginationControls/TablePaginationControls";
+import {ICosmicData} from "../../shared/components/mutationTable/column/CosmicColumnFormatter";
 import {IVariantCountData} from "./mutation/column/CohortColumnFormatter";
 
 import {IHotspotData, IMyCancerGenomeData, IMyCancerGenome} from "./mutation/column/AnnotationColumnFormatter";
@@ -53,6 +57,7 @@ interface IPatientViewState {
     mutationData: any;
     myCancerGenomeData?: IMyCancerGenomeData;
     hotspotsData?: IHotspotData;
+    cosmicData?: ICosmicData;
     mrnaExprRankData?: MrnaRankData;
     mutSigData?: MutSigData;
     variantCountData?: IVariantCountData;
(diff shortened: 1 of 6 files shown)


onursumer:
[(3) Add COSMIC counts to mutation table [API Ready]](https://trello.com/c/hi7y6nGE/16-3-add-cosmic-counts-to-mutation-table-api-ready)

onursumer:
[Add COSMIC Roll-over Table to the Mutation Table [API Ready]](https://trello.com/c/azghnWu6/11-add-cosmic-roll-over-table-to-the-mutation-table-api-ready)

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -167,6 +172,37 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         });
     }
 
+    fetchCosmicData(mutations:Mutation[]):Promise<ICosmicData> {
+        const generateMap = function(cosmicMutations:CosmicMutation[]) {
```
alisman: probalby would be better to expernalize this for testing purposes

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -167,6 +172,37 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+            // key => geneSymbol_proteinPosition
+            // protienPosition => start[_end]
+            const map: ICosmicData = {};
+
+            // create a map for a faster lookup
+            _.each(cosmicMutations, function(cosmic:CosmicMutation) {
+                if (!map[cosmic.keyword]) {
```
alisman: i like to use the in operator for this kind of check  if ( str in map )

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -167,6 +172,37 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+                    map[cosmic.keyword] = [];
+                }
+
+                map[cosmic.keyword].push(cosmic);
+            });
+
+            return map;
+        };
+
+        const queryKeywords:string[] = _.uniq(_.map(mutations, (mutation:Mutation) => {return mutation.keyword;}));
```
alisman: you should use the arrow function shortcut syntax:   (mutation:Mutation) => mutation.keyword

if no brackets it knows to return the value

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -272,9 +308,16 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                 });
 
                 this.fetchMutationData(sampleIds).then((_result) => {
-                    this.fetchHotspotsData(_result).then((hotspotsData:IHotspotData) => {
+                    const hotspotDataPromise = this.fetchHotspotsData(_result).then((hotspotsData:IHotspotData) => {
```
alisman: curcle brackets are not necessary here either in arrow callback

src/shared/components/cosmic/CosmicMutationTable.tsx
```diff
@@ -0,0 +1,95 @@
...
+            },
+            reactTableProps: {
+                className: "table table-striped table-border-top",
+                hideFilterInput:true,
+                defaultSort: {column: columns.occurrence.name, direction: 'desc'}
+            }
+        };
+
+        // merge provided props with the default props
+        return lodash.merge(defaultProps, props);
```
alisman: i think we should have convention of lodash = _   it's used that most everywhere else in project

src/pages/patientView/PatientViewPage.tsx
onursumer: In addition to that, we should also consider externalizing data fetcher functions to other classes

src/pages/patientView/PatientViewPage.tsx
onursumer: good to know, it looks more readable that way.

src/pages/patientView/PatientViewPage.tsx
onursumer: old habits die hard :)

src/shared/components/cosmic/CosmicMutationTable.tsx
onursumer: good point

src/pages/patientView/PatientViewPage.tsx
alisman: Yes, in mobx we do that.  Will show you that today.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -47,6 +51,7 @@ interface IPatientViewState {
     cnaSegmentData: any;
     mutationData: any;
     hotspotsData?: IHotspotData;
+    cosmicData?: ICosmicData;
```
adamabeshouse: how about moving this (and IHotspotData) definition into this file and importing it from here in the column formatters?

onursumer: That would create a dependency on `PatientViewPage` in the column formatters which I would like to avoid as much as possible :)

adamabeshouse: ah, that's a good point. Yes I vote for move them somewhere else

onursumer: @alisman @adamabeshouse @inodb technically all the type/interface definitions we have in general, except props and state, can be considered as our model and moved out of the view/component classes into separate model classes/files. Currently we don't really have a separation of model from view, maybe except the auto-generated API classes.

alisman: We should try to distinguish between models (api data) and view-models, which are transformations necessary for view layer and which therefor don't necessarily need to be decoupled from views.   We should wait to really worry about this when mobx is in the picture.  Which should be next week.

PR #155, opened 2017-02-01
Cohort variant frequency column

adamabeshouse:
When keyword data is present:
image

When no keyword data:
image

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -21,9 +21,14 @@ import SampleManager from './sampleManager';
 import SelectCallback = ReactBootstrap.SelectCallback;
 import CancerHotspotsAPI from "../../shared/api/CancerHotspotsAPI";
 import {HotspotMutation} from "../../shared/api/CancerHotspotsAPI";
-import {MutSig, MrnaPercentile, default as CBioPortalAPIInternal} from "../../shared/api/CBioPortalAPIInternal";
+import {
+    MutSig, MrnaPercentile, default as CBioPortalAPIInternal,
+    VariantCountIdentifier, VariantCount
+} from "../../shared/api/CBioPortalAPIInternal";
 import PatientHeader from './patientHeader/PatientHeader';
 import {TablePaginationControls} from "../../shared/components/tablePaginationControls/TablePaginationControls";
+import {IVariantCountData} from "./mutation/column/CohortColumnFormatter";
+
 import {IHotspotData, IMyCancerGenomeData, IMyCancerGenome} from "./mutation/column/AnnotationColumnFormatter";
 import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
 
@@ -50,6 +55,7 @@ interface IPatientViewState {
     hotspotsData?: IHotspotData;
     mrnaExprRankData?: MrnaRankData;
     mutSigData?: MutSigData;
+    variantCountData?: IVariantCountData;
     activeTabKey: number;
 }
 
@@ -91,6 +97,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
             cnaSegmentData: undefined,
             hotspotsData: undefined,
             mrnaExprRankData: undefined,
+            variantCountData: undefined,
             activeTabKey:1
         };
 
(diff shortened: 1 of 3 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -41,6 +44,13 @@ export interface IPatientViewPageProps {
 
 export type MrnaRankData = { [sampleId:string]: { [entrezGeneId:string]: {percentile:number, zScore:number}}};
 export type MutSigData = { [entrezGeneId:string]:{ qValue:number } }
+export type VariantCountData = {
```
onursumer: our current lint rules favor interface over type literals, but I am not sure that if we want to stick to that convention.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -247,6 +259,61 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+                        for (let keyword of keywords) {
+                            variantCountIdentifiers.push({
+                                entrezGeneId: entrezGeneIdInt,
+                                keyword: keyword
+                            });
+                        }
+                    }
+                }
+            }
+            const fetchPromise = this.tsInternalClient.getVariantCountsUsingPOST({geneticProfileId: this.mutationGeneticProfileId, variantCountIdentifiers});
```
onursumer: When the line gets too long, I like to break it into multiple lines.

src/pages/patientView/mutation/column/CohortColumnFormatter.tsx
```diff
@@ -5,35 +5,93 @@ import {IColumnFormatterData} from "../../../../shared/components/enhancedReactT
...
+            return null;
+        }
+        const entrezGeneId = data.rowData[0].entrezGeneId;
+        const hugoGeneSymbol = data.rowData[0].gene.hugoGeneSymbol;
+        const keyword = data.rowData[0].keyword;
+
+        const numberOfSamples = variantCountData.numberOfSamples;
+        const geneData = variantCountData.geneData[entrezGeneId];
+
+        let numberOfSamplesWithMutationInGene, numberOfSamplesWithKeyword;
```
onursumer: With the current lint settings, multiple variable declarations in the same statement are not allowed. But, we can discuss this one too.

src/pages/patientView/PatientViewPage.tsx
alisman: yeah, might as well use interface

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -247,6 +259,61 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         return this.tsInternalClient.getSignificantlyMutatedGenesUsingGET({studyId: this.studyId});
     }
 
+    fetchVariantCountData(mutations:Mutation[]):Promise<VariantCountData> {
+        return new Promise((resolve, reject) => {
+            const variantCountIdentifiers:VariantCountIdentifier[] = [];
+            const toQuery:{ [entrezGeneId:string]:{ [keyword:string]:boolean}} = {};
+            for (let mutation of mutations) {
```
alisman: lets default to using const instead of let.  only let when we know we are going to change the param

src/pages/patientView/mutation/column/CohortColumnFormatter.tsx
```diff
@@ -5,35 +5,93 @@ import {IColumnFormatterData} from "../../../../shared/components/enhancedReactT
...
+        const variantCountData = CohortColumnFormatter.getVariantCountData(data, (columnProps.variantCountData as VariantCountData));
+        const freqViz = CohortColumnFormatter.makeCohortFrequencyViz(data, (columnProps.variantCountData as VariantCountData))
         return (
-            <Td key={data.name} column={data.name}>
-                <If condition={qValue !== null}>
-                    {CohortColumnFormatter.makeMutSigIcon(qValue)};
-                </If>
+            <Td key={data.name} column={data.name} value={variantCountData && variantCountData.numberOfSamplesWithMutationInGene}>
+                <div>
+                    <If condition={freqViz !== null}>
```
alisman: i'm curious why this passes the ts compiler because if a variable is nullable then it should require a if statement to check for its null-ness.  this is why @adufilie thinks we should stay away from <if>, as pleasant as it is

adufilie: null is a valid value to have in JSX - we're not trying to access a property of freqViz here.

adufilie: actually, this `<If>` is unnecessary.

adufilie: Actually, this `<If>` is unnecessary.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: I only used it here because I thought that it does change by being reset during the iteration

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: done, I also moved the definition into CohortColumnFormatter.tsx like @onursumer and I talked about

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: done

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: done

src/pages/patientView/mutation/column/CohortColumnFormatter.tsx
adamabeshouse: done

PR #158, opened 2017-02-02
Fixed the copy/download button functionality for Mutation Table

onursumer:
# What? Why?
- Download data generators are implemented for all core data and allele frequency columns.
- Currently both visible and hidden columns are included in the copy/download data.
- Even if the table is filtered, always includes the entire unfiltered table in the copy/download data.

Columns with missing copy/download data (i.e. `downloader` functions need to be implemented):
- [ ] Tumors
- [ ] Annotation
- [ ] Allele Freq
- [ ] Allele Freq (N)
- [ ] Copy #
- [ ] mRNA Expr.
- [ ] Cohort
- [ ] COSMIC

# Checks
  the `-s` option)


onursumer:
[(5) Format mutation data for copy/download](https://trello.com/c/mcuS4r9p/74-5-format-mutation-data-for-copy-download)

src/shared/components/tableHeaderControls/TableHeaderControls.tsx
```diff
@@ -70,6 +77,8 @@ export default class TableExportButtons extends React.Component<ITableExportButt
         showPagination:false,
         searchClassName: '',
         copyDownloadClassName: '',
+        downloadFilename: 'patient-clinical-attributes.csv',
```
onursumer: @alisman `patient-clinical-attributes.csv` was hard-coded as the download file name value. I made it customizable and set the default value. But I still don't think it is a good idea to have `patient-clinical-attributes.csv` as a default value in this component. A generic name like `table.csv` might be better.

alisman: agreed.

PR #166, opened 2017-02-07
DefaultTooltip component

adamabeshouse:
Wrapper component for Tooltip with hardcoded settings for mouseEnterDelay, mouseExitDelay, and arrowContent

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/DefaultTooltip.tsx
+++ b/src/shared/components/DefaultTooltip.tsx
@@ -1,0 +1,18 @@
+import * as React from 'react';
+import Tooltip from 'rc-tooltip';
+import 'rc-tooltip/assets/bootstrap_white.css';
+
+export default class DefaultTooltip extends React.Component<Tooltip.Props, {}> {
+    render() {
+        const defaultProps = {
+            mouseEnterDelay:0.5,
+            mouseLeaveDelay:0.05,
+            arrowContent:(<div className="rc-tooltip-arrow-inner"/>)
+        };
+        return (
+            <Tooltip {...Object.assign({}, defaultProps, this.props)}>
+                {this.props.children}
+            </Tooltip>
+        );
+    }
+}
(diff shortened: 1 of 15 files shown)


src/shared/components/annotation/CancerHotspots.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import * as _ from 'lodash';
-import Tooltip from 'rc-tooltip';
+import DefaultTooltip from 'shared/components/DefaultTooltip';
```
alisman: small thing but I've stopped using the shared alias b/c the IDE doesn't seem to alble to resolve it.  let me know if it does for you.  i just do relative path to shared.  this gives me code completion on the exports of these modules

src/shared/components/DefaultTooltip.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+export default class DefaultTooltip extends React.Component<Tooltip.Props, {}> {
+    render() {
+        const defaultProps = {
+            mouseEnterDelay:0.5,
+            mouseLeaveDelay:0.05,
+            arrowContent:(<div className="rc-tooltip-arrow-inner"/>)
+        };
+        return (
+            <Tooltip {...Object.assign({}, this.props, defaultProps)}>
```
alisman: weird, i thought i commented on this but we should consider reversing this so that a developer can override the defaults if necessary.

adamabeshouse: yeah good call

src/shared/components/annotation/CancerHotspots.tsx
adamabeshouse: oh, it worked for me. I assumed it was going from the project root directory, I didn't realize it was an alias. Should I change it?

alisman: It works, as in, it compiles, but the IDE won't give hints to the path, which I find to be very annoying.

adamabeshouse: oh, yeah. I did it because I was finding it annoying to guess/figure out how many ".." to apply before I was in the right directory haha. Do you want me to change it?

alisman: no don't worry about it.

PR #171, opened 2017-02-13
Lazy load mrna column

adamabeshouse:
Please note and comment on my strategy for dealing with a function which has MobXPromise dependencies - I make it so that the function itself is the result of a MobXPromise which awaits the dependencies. See requestMrnaExprRankData in PatientViewPageStore.tsx

The advantage of this is it moves the async back into MobX territory, whereas the other thoughts I was having would have to mingle with ES6 Promises and callbacks within the function that had the async dependencies.

Issues:

- ~~Slightly wonky behavior if you try to sort by the column, because all the data is not yet loaded~~. We don't sort by this column.

- Potentially redundant rendering that should be optimized with shouldComponentUpdate

- With a lot of rows, it's really slow because every time you update the data (i.e. on every new page) you have to regenerate all the rows. We need a better system for this where you can update the rows by cell, and only the rows that need to be updated.

# Checks
  the `-s` option)


adamabeshouse:
[(8) Mutation Table --> retrieve mRNA Expr data lazily](https://trello.com/c/7SjkKJxc/81-8-mutation-table-retrieve-mrna-expr-data-lazily)

adufilie:
Take a look at expr() which dynamically creates a computed value.  This gives you more control over caching.

https://mobxjs.github.io/mobx/refguide/expr.html

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -85,11 +84,11 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
             mutationData: undefined,
             cnaSegmentData: undefined,
             hotspotsData: undefined,
-            mrnaExprRankData: undefined,
             activeTabKey:1
         };
 
         this.handleSelect = this.handleSelect.bind(this);
+        this.lazyFetchVisibleRowData = this.lazyFetchVisibleRowData.bind(this);
```
adufilie: You can avoid bind() by using the arrow syntax where you define the function.

src/shared/components/enhancedReactTable/EnhancedReactTable.tsx
```diff
@@ -145,19 +145,32 @@ export default class EnhancedReactTable<T> extends React.Component<IEnhancedReac
...
         this.shouldSetState = false;
 
-        const setFilteredDataLength = (n:number) => {
+        const setFilteredDataLength = ((n:number) => {
             if (n !== this.filteredDataLength) {
                 this.filteredDataLength = n;
                 this.shouldSetState = true;
             }
-        };
+        }).bind(this);
```
adufilie: .bind() has no effect on arrow functions.

src/shared/components/mutationTable/IMutationTableProps.ts
```diff
@@ -9,6 +9,7 @@ export interface IMutationTableProps extends IEnhancedReactTableProps<MutationTa
 {
     title?:string;
     samples?: ClinicalDataBySampleId;
+    onVisibleRowsChange?:(data:MutationTableRowData[]) => any;
```
adufilie: If you don't use the result of the function it should be specified as void.

src/shared/components/enhancedReactTable/IEnhancedReactTableProps.ts
```diff
@@ -15,6 +15,7 @@ export interface IEnhancedReactTableProps<T> {
     initItemsPerPage?: number; // initial number of items per page
     initPage?:number; // initial page
     itemsName?:string; // name of items, e.g. "mutations"
+    onVisibleRowsChange?:(data:T[]) => any; // executed whenever the visible rows change (sorting or paginating), argument is the row data of visible rows, in order
```
adufilie: If you don't use the result of the function it should be specified as void.

src/pages/patientView/mutation/column/MrnaExprColumnFormatter.tsx
```diff
@@ -27,17 +32,18 @@ export default class MrnaExprColumnFormatter {
 
     private static getTooltipContents(data:IColumnFormatterData<MutationTableRowData>, mrnaExprData: any) {
```
adufilie: Shouldn't the parameter type be MrnaRankData?

src/pages/patientView/mutation/MutationInformationContainer.tsx
```diff
@@ -25,6 +25,7 @@ export interface IMutationInformationContainerProps {
     sampleTumorType:{ [s:string]: string};
     sampleCancerType:{ [s:string]: string};
     sampleManager:SampleManager;
+    onVisibleRowsChange?:(data:Mutation[][]) => any;
```
adufilie: Function return type should be void.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,31 +161,42 @@ export class PatientViewPageStore
...
         invoke: () => {
-            const sampleToEntrezGeneIds = _.reduce(this.mutationData.result, (map:{ [s:string]:Set<number> }, next:Mutation) => {
-                const sampleId = next.sampleId;
-                map[sampleId] = map[sampleId] || new Set();
-                map[sampleId].add(next.entrezGeneId);
-                return map;
-            }, {});
-
-            return new Promise((resolve, reject) => {
+            return Promise.resolve(function(sampleToEntrezGeneIds:{ [sampleId:string]:Set<number> }) {
```
adufilie: Promise.resolve(func) is weird.  We should find another way.

src/shared/components/mutationTable/IMutationTableProps.ts
adamabeshouse: It doesn't have to be void though, if you wanted to use the same function for another purpose or something this lets you do that.

adufilie: Any function will match () => void, so you should make the return value void if you are not using it.

src/pages/patientView/mutation/column/MrnaExprColumnFormatter.tsx
adamabeshouse: good catch

src/shared/components/mutationTable/IMutationTableProps.ts
adamabeshouse: ah I see

adufilie: Correciton, `() => void` will match any zero-parameter function. I meant a function param with a void return type will accept functions with non-void return types.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,31 +161,42 @@ export class PatientViewPageStore
...
+                // Reference to new immutable map
+                // See which we need to fetch, and set "pending" for those data
+                const toQuery:{ [sampleId:string]:number[]} = {};
+                for (const sampleId in sampleToEntrezGeneIds) {
+                    if (sampleToEntrezGeneIds.hasOwnProperty(sampleId)) {
```
alisman: this check should really be handled by iterator.  so, for example, use Object.keys or lodash _.forEach.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,31 +161,42 @@ export class PatientViewPageStore
...
+                        this.mrnaExprRankData[sampleId] = this.mrnaExprRankData[sampleId] || {};
+                        for (const entrezGeneId of sampleToEntrezGeneIds[sampleId]) {
+                            if (!this.mrnaExprRankData[sampleId].hasOwnProperty(entrezGeneId)) {
```
alisman: use the in operator (if str in obj)  or we could maybe use Maps, which have the 'has' method

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,31 +161,42 @@ export class PatientViewPageStore
...
                         } else {
                             const fetchMrnaPercentilesPromise = internalClient.fetchMrnaPercentileUsingPOST({
                                 geneticProfileId:this.mrnaRankGeneticProfileId.result,
-                                sampleId:sampleId,
-                                entrezGeneIds: Array.from(entrezGeneIds)
+                                sampleId,
+                                entrezGeneIds
                             });
                             fetchMrnaPercentilesPromise.then((d) => {
                                 mrnaPercentiles.push.apply(mrnaPercentiles, d);
```
alisman: you don't need to use apply here because mrnaPercentiles is already the context (this) by virtue of method being called on instance

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,31 +161,42 @@ export class PatientViewPageStore
...
+                }
+                // Fetch that data
                 const mrnaPercentiles: MrnaPercentile[] = [];
-                const fetchAllMrnaPercentilesPromise = Promise.all(Object.keys(sampleToEntrezGeneIds).map((sampleId:string) =>
+                const fetchAllMrnaPercentilesPromise = Promise.all(Object.keys(toQuery).map((sampleId:string) =>
                     new Promise((sampleResolve, sampleReject) => {
-                        const entrezGeneIds = sampleToEntrezGeneIds[sampleId];
-                        if (typeof entrezGeneIds === "undefined" || entrezGeneIds.size === 0 || this.mrnaRankGeneticProfileId.result === null) {
+                        const entrezGeneIds = toQuery[sampleId];
+                        if (this.mrnaRankGeneticProfileId.result === null) {
```
alisman: what's condition for this being null here?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adufilie: apply() is being used to act like a spread operator.  Instead we can do `mrnaPercentiles.push(...d);`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adufilie: `hasOwnProperty` is better than `in` because `in` returns true if something is on the prototype.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -350,14 +361,17 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
-                                    mrnaExprRankData={patientViewPageStore.mrnaExprRankData.isComplete ? patientViewPageStore.mrnaExprRankData.result : undefined }
+                                    mrnaExprRankData={ patientViewPageStore.mrnaExprRankData }
                                     mutSigData={this.state.mutSigData}
                                     sampleOrder={sampleManager.sampleOrder}
                                     sampleLabels={sampleManager.sampleLabels}
                                     sampleColors={sampleManager.sampleColors}
                                     sampleTumorType={mockData.tumorType}
                                     sampleCancerType={mockData.cancerType}
                                     sampleManager={ sampleManager }
+                                    onVisibleRowsChange={ patientViewPageStore.requestMrnaExprRankData.isComplete
```
alisman: lets put this out in a component method.  we're going to want to test it.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: I was having some sort of issue with setting `mrnaPercentiles = mrnaPercentiles.concat(d)`, which is why I opted to use `.push` to do it in-place. Good call on the spread operator.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: which part?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -350,14 +361,17 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+                                    ? (data:Mutation[][]) => this.lazyFetchVisibleRowData(data, patientViewPageStore.requestMrnaExprRankData.result)
```
alisman: I think we can check/access the store inside of the lazyFetch handler so as to keep this cleaner.   To achieve functional purity, we can have another layer of indirection there.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: if there is no such genetic profile for the study

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: this doesn't strike me as dirty but I'm interested in what you have in mind

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: done

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: done

PR #174, opened 2017-02-15
Lazy load mRNA Expr rank column

adamabeshouse:
# Checks
  the `-s` option)

### DIFF
--- a/src/shared/api/LazyCache.ts
+++ b/src/shared/api/LazyCache.ts
@@ -1,0 +1,34 @@
+import {observable, action} from '../../../node_modules/mobx/lib/mobx';
+
+export default class LazyCache {
+    @observable.ref protected _cache:any;
+    protected dependencies:any[];
+
+    constructor(..._dependencies:any[]) {
+        this._cache = {};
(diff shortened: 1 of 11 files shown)


src/shared/api/LazyCache.ts
```diff
@@ -1,27 +1,34 @@
...
+        this.dependencies = _dependencies;
     }
 
     public get cache() {
         return this._cache;
     }
 
     public populate(args:any) {
-        this._populateCache(this._cache, args, this.dependencies).then((didChange:boolean) => {
+        this.populateCache(args, ...this.dependencies).then((didChange:boolean)=>{
```
alisman: lets think about whether it would be better to just use immutable to handle this.  the data arguably should be immutable for other reasons, so why not just use it to handle this business.

adamabeshouse:
[(8) Mutation Table --> retrieve mRNA Expr data lazily](https://trello.com/c/7SjkKJxc/81-8-mutation-table-retrieve-mrna-expr-data-lazily)

PR #191, opened 2017-02-28
Msktable

adamabeshouse:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -34,6 +34,7 @@ import ClinicalInformationSamples from "./clinicalInformation/ClinicalInformatio
 import {ICosmicData} from "../../shared/components/mutationTable/column/CosmicColumnFormatter";
 import {keywordToCosmic, geneToMyCancerGenome, geneAndProteinPosToHotspots} from 'shared/lib/AnnotationUtils';
 import {IVariantCountData} from "./mutation/column/CohortColumnFormatter";
+import {observable} from "mobx";
 import {observer} from "mobx-react";
 import {IHotspotData, IMyCancerGenomeData, IMyCancerGenome} from "./mutation/column/AnnotationColumnFormatter";
 import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
@@ -41,6 +42,11 @@ import CopyNumberAlterationsTable from "./copyNumberAlterations/CopyNumberAltera
 import CopyNumberTableWrapper from "./copyNumberAlterations/CopyNumberTableWrapper";
 import {reaction} from "mobx";
 import AppConfig from 'appConfig';
+import SimpleTable from "../../shared/components/simpleTable/SimpleTable";
+import MSKTable from "../../shared/components/msktable/MSKTable";
+
+class NameTable extends MSKTable<{first:string}> {
+}
 
 const patientViewPageStore = new PatientViewPageStore();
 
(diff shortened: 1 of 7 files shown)


src/shared/components/msktable/MSKTable.tsx
```diff
@@ -0,0 +1,110 @@
...
+    @observable public sortColumn:string;
+    @observable public sortAscending:boolean;
+    @observable public columns:Column<T>[];
+    @observable.ref public data:T[];
+    @computed get sortedFilteredData():T[] {
+        let filtered:T[];
+        if (this.filterString) {
+            filtered = this.data.filter((datum:T)=>{
+                let match = false;
+                for (let i=0; i<this.columns.length; i++) {
```
alisman: this is the inner loop.  gonna happen a lot so, worth optimizing.  you could cache this.columns.length in local variable.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -39,6 +39,11 @@ import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
 import CopyNumberAlterationsTable from "./copyNumberAlterations/CopyNumberAlterationsTable";
 import CopyNumberTableWrapper from "./copyNumberAlterations/CopyNumberTableWrapper";
 import {reaction} from "mobx";
+import SimpleTable from "../../shared/components/simpleTable/SimpleTable";
+import MSKTable from "../../shared/components/msktable/MSKTable";
+
+class TABLE extends MSKTable<{yo:string, text:string}> {
```
alisman: curious why you extend it.  define T?

src/shared/components/simpleTable/SimpleTable.tsx
```diff
@@ -0,0 +1,102 @@
...
+};
+
+class SimpleTableStore {
+    private onSort?:(name:string, sortAscending:boolean)=>void;
+
+    @observable public sortColumn:string;
+    @observable public sortAscending:boolean;
+    @observable.ref public rows:Row[];
+    @observable public columns:Column[];
+    @computed get thElements():JSX.Element[] {
```
alisman: just fyi, i think we're going to need to support custom column headers.  andy has a use case for this and asked me.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: yeah, because you can't use generic classes in JSX

src/shared/components/simpleTable/SimpleTable.tsx
adamabeshouse: Should be easy to change once the desired parameters are understood

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: better idea: for..of loop

src/shared/components/simpleTable/SimpleTable.tsx
```diff
@@ -0,0 +1,102 @@
...
+export type Row = { [columnName:string]:JSX.Element };
+
+
+type SimpleTableProps = {
+    columns:Column[];
+    rows:Row[];
+    onSort?:(name:string, sortAscending:boolean)=>void;
+};
+
+class SimpleTableStore {
```
adamabeshouse: That makes sense. Then we can make it onHeaderClick which is more general and have the parent handle all the logic there. So this component is really just a visualizing shell.

src/pages/patientView/PatientViewPage.tsx
alisman: how is T then used?  (not challenging this just curious.

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -0,0 +1,126 @@
...
+
+class MSKTableStore<T> {
+    @observable public filterString:string;
+    @observable public page:number;
+    @observable public itemsPerPage:number;
+    @observable public sortColumn:string;
+    @observable public sortAscending:boolean;
+    @observable public columns:Column<T>[];
+    @observable.ref public data:T[];
+    @computed get sortedFilteredData():T[] {
```
onursumer: @adamabeshouse @alisman Just realized if we agree on setting `T` to `Array<Mutation>`, then there would be no information about the additional columns in `datum`. Wouldn't sorting and filtering by additional columns be a problem in that case?

adamabeshouse: no, because you can reference that data via closure in the column definition of `sort` and `filter`

adamabeshouse: after a chat in person, onur agreed this seems at the moment to be a good solution, flexible enough to handle things

PR #192, opened 2017-02-28
Make patient navigator editable to jump to certain patient's page

ersinciftci:


ersinciftci:
[(8) Next / Previous samples button -> making `11` in the `#11 of 61 patients` editable](https://trello.com/c/3mQBdqjQ/129-8-next-previous-samples-button-making-11-in-the-11-of-61-patients-editable)

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -82,9 +132,25 @@ export class TablePaginationControls extends React.Component<ITablePaginationCon
...
+                                {this.props.textBetweenButtons}
+                            </Button>
+                        </Then>
+                        <Else>
+                            <FormControl
+                                autoFocus={true}
+                                className={styles["page-number-input"]}
+                                value={this.pageNumberValue}
+                                type="number"
+                                onKeyPress={this.handleCurrentPageSubmit as React.KeyboardEventHandler<any>}
```
alisman: hard to say without using it but it think it should require submission either by return key or blur? i.e. not just on keypress

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -82,9 +132,25 @@ export class TablePaginationControls extends React.Component<ITablePaginationCon
                     <Button key="prevPageBtn" disabled={!!this.props.previousPageDisabled} onClick={this.props.onPreviousPageClick}>
                         {this.props.previousButtonContent}
                     </Button>
-                    <Button key="textBetweenButtons" className={styles["auto-cursor"]} disabled={true}>
-                        {this.props.textBetweenButtons}
-                    </Button>
+                    <If condition={!this.textBoxActive}>
+                        <Then>
+                            <Button key="textBetweenButtons" className={styles["auto-cursor"]} disabled={!this.props.pageNumberEditable}
```
alisman: hard to say without using it but the text box could probably just always be active.  styling would be easier that way.  in other words, no activation button

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -45,24 +49,70 @@ export class TablePaginationControls extends React.Component<ITablePaginationCon
...
+    handleChangeCurrentPage(evt:React.FormEvent<HTMLInputElement>) {
+
+        this._pageNumberValue = parseInt(evt.currentTarget.value);
+    }
+
+    handleMiddleButtonClick() {
+        this.textBoxActive = true;
+    }
+
+    handleOnBlur() {
```
alisman: I think onBlur should give you the same evt argument as onChange, so you could potentially reuse handleChangeItemsPerPage here.

src/shared/components/paginationControls/PaginationControls.tsx
alisman: ah, i see.  it does check for enter key

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -45,24 +49,70 @@ export class TablePaginationControls extends React.Component<ITablePaginationCon
...
+    handleCurrentPageSubmit(evt:React.KeyboardEvent<HTMLInputElement>) {
+
+        if(evt.charCode == 13){
+            this.jumpToPage();
+        }
+    }
+
+    handleChangeCurrentPage(evt:React.FormEvent<HTMLInputElement>) {
+
+        this._pageNumberValue = parseInt(evt.currentTarget.value);
```
alisman: i don't think we should need to model this on the class.  We can always read it off the evt for the submission events.  Alternatively, we can set the input as a ref using react's ref mechanism and then just read it whenever we want.

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -45,24 +49,70 @@ export class TablePaginationControls extends React.Component<ITablePaginationCon
...
-        (this.props.onChangeItemsPerPage || (()=>0))(parseInt((evt.target as HTMLSelectElement).value,10));
+
+        if (this.props.onChangeItemsPerPage) {
+            this.props.onChangeItemsPerPage(parseInt((evt.target as HTMLSelectElement).value,10));
+        }
+    }
+
+    handleCurrentPageSubmit(evt:React.KeyboardEvent<HTMLInputElement>) {
+
+        if(evt.charCode == 13){
```
alisman: I had a big argument with Andy about == vs ===.   It's a pretty well established convention to use === by default.   == allows for type coercion.  In this case, it doesn't matter, but I want for people to adopt that standard.   We will enforce it with the linter.

adufilie: Actually in TypeScript it does not matter in all cases: https://gist.github.com/adufilie/0fef6f51652c352ac80856b2cf2078fa#file-output-ts

src/shared/components/paginationControls/PaginationControls.tsx
adufilie: This is required or the input will not work. https://facebook.github.io/react/docs/forms.html#controlled-components

src/shared/components/paginationControls/PaginationControls.tsx
ersinciftci: but the disabled button says `1 of 61  patients`, so it doesn't make sense to make the whole text editable. when user clicks the button(not disabled anymore), it turns to a text input with only `1` in it instead of `1 of 61 patients`, and in the event of enter key press or onBlur, it navigates to the new index and text input turns back to a button. you will see what I mean when you use it.

src/shared/components/paginationControls/PaginationControls.tsx
ersinciftci: Not sure what you mean. I use onBlur to jump to a page when the text input loses focus, just like Enter key press, but I use onChange just to update the value in text input. You might understand better when you use it, or I misunderstood what you meant.

src/shared/components/paginationControls/PaginationControls.tsx
alisman: I tried it on heroku deploy.  This is an Ethan question but I would vote for just the page number to become editable so you can still see the upper page bound.

src/shared/components/paginationControls/PaginationControls.tsx
alisman: Andy's right.     I've been using defaultValue attr to avoid controlled components but I see it's necessary here because it needs to adjust to arrow navigation.

PR #204, opened 2017-03-06
OncoKB icon tooltip support

onursumer:
# What? Why?

Added tool tip (OncoKB Card) to the annotation column for the OncoKB icon.

Note that currently `OncoKbUtils` class contains a lot of legacy code. I just made it work with TypeScript with minimal changes. We need to refactor both `OncoKBCard` and `OncoKBUtils` in a more modular and testable way.

# Checks
  the `-s` option)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -482,6 +482,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                     cosmicData={this.state.cosmicData}
                                     mrnaExprRankData={ patientViewPageStore.mrnaExprRankCache.cache }
                                     oncoKbData={patientViewPageStore.oncoKbData.isComplete ? patientViewPageStore.oncoKbData.result : undefined}
+                                    pmidData={patientViewPageStore.pmidData.isComplete ? patientViewPageStore.pmidData.result : undefined}
```
adufilie: These ternary operators are unnecessary

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
+            const evidenceLookupPromise = oncokbClient.evidencesLookupPostUsingPOST(
+                {body: generateEvidenceQuery(queryVariants)});
 
+            return new Promise((resolve, reject) => {
                 // TODO return type is not correct for the auto-generated API!
                 // generated return type is Array<IndicatorQueryResp>,
                 // but the actual return type is {meta: {} data: Array<IndicatorQueryResp>}
                 // that's why here we need to force data type to be any and get actual data by data.data
-                oncokbSearchPromise.then((data:any) => {
+                Promise.all([oncokbSearchPromise, evidenceLookupPromise]).then((values) => {
```
adufilie: Use destructuring here:
`let [onkokbSearch, evidenceLookup] = await Promise.all([oncokbSearchPromise, evidenceLookupPromise]);`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
                     sampleIdToTumorType[mutation.sampleId]);
             }), "id");
 
-            return new Promise((resolve, reject) => {
```
adufilie: This code would be much cleaner with async/await

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
                 // TODO return type is not correct for the auto-generated API!
                 // generated return type is Array<IndicatorQueryResp>,
                 // but the actual return type is {meta: {} data: Array<IndicatorQueryResp>}
                 // that's why here we need to force data type to be any and get actual data by data.data
-                oncokbSearchPromise.then((data:any) => {
+                Promise.all([oncokbSearchPromise, evidenceLookupPromise]).then((values) => {
                     const oncoKbData:IOncoKbData = {
                         sampleToTumorMap: sampleIdToTumorType,
-                        indicatorMap: generateIdToIndicatorMap(data.data)
+                        indicatorMap: generateIdToIndicatorMap((values[0] as any).data),
```
adufilie: Instead of letting the `any` propagate throughout the whole function, either fix the generated API code to have the correct type or cast the promise as the correct type. You can do `as any as CorrectType`.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
                     resolve(oncoKbData);
                 });
 
                 oncokbSearchPromise.catch(() => reject());
+                evidenceLookupPromise.catch(() => reject());
+            });
+        }
+    }, {sampleToTumorMap: {}, indicatorMap: {}, evidenceMap: {}});
+
+    readonly pmidData = remoteData<any>({
```
adufilie: It's better to let typescript determine the type instead of using `any`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
-            return new Promise((resolve, reject) => {
-                const oncokbSearchPromise = oncokbClient.searchPostUsingPOST(
-                    {body: generateEvidenceQuery(queryVariants)});
+            const oncokbSearchPromise = oncokbClient.searchPostUsingPOST(
+                {body: generateEvidenceQuery(queryVariants)});
+
+            const evidenceLookupPromise = oncokbClient.evidencesLookupPostUsingPOST(
+                {body: generateEvidenceQuery(queryVariants)});
 
+            return new Promise((resolve, reject) => {
```
adufilie: This code would be cleaner with async/await

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
+    }, {sampleToTumorMap: {}, indicatorMap: {}, evidenceMap: {}});
+
+    readonly pmidData = remoteData<any>({
+        await: () => [
+            this.oncoKbData
+        ],
+        invoke: () => {
+            let refs:any[] = [];
+
+            _.each(this.oncoKbData.result.evidenceMap, (evidence:any) => {
```
adufilie: Make the result have the correct type instead of using `any` everywhere.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -341,27 +341,104 @@ export class PatientViewPageStore
...
+                        {
+                            refs = refs.concat(item.articles.map((article:any) => {
+                                return Number(article.pmid);
+                            }));
+                        }
+                    });
+                }
+            });
+
+            return new Promise((resolve, reject) => {
```
adufilie: This would be cleaner with async/await.

src/pages/patientView/mutation/column/AnnotationColumnFormatter.tsx
```diff
@@ -37,6 +38,8 @@ export interface IAnnotation {
     is3dHotspot: boolean;
     myCancerGenomeLinks: string[];
     oncoKbIndicator?: IndicatorQueryResp;
+    oncoKbEvidence?: any;
```
adufilie: Try to avoid `any`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adufilie: I see, this would be an implicit any.  So we should define the correct type instead.

src/pages/patientView/PatientViewPage.tsx
onursumer: Good to know!

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: Long term plan is to fix this problem in the actual OncoKB API definition. Return types are not defined correctly there.

onursumer:
@adufilie I agree that we need proper types. I don't like seeing `any`s all over the place either, but the idea for this sprint was to make the legacy OncoKB code work with the current framework.

@alisman do you think we should try addressing typing issues in this PR, or is it okay to deal with them in the future?

onursumer:
@adufilie @alisman: updated the PR with Andy's suggestions. Also added a partial model for `Evidence` (just copied from the model from legacy code), and got rid of some of the `any`s. Now the code looks slightly better.

onursumer:
[(13) Mutation Table --> oncokb mouseover card](https://trello.com/c/adbegZpf/36-13-mutation-table-oncokb-mouseover-card)

onursumer:
@adamabeshouse yes, we should definitely defer fetching of the tooltip data until mouseover event. But it should be okay to do it in another PR. We should also move related data fetching functions from `PatientViewPage` to `PatientViewStore`.

PR #207, opened 2017-03-08
finish porting all columns to new table, except annotation (which onur is handling)

adamabeshouse:
# Checks
  the `-s` option)


adamabeshouse:
[(5) Port mutation table columns to new table component](https://trello.com/c/f917yHAl/141-5-port-mutation-table-columns-to-new-table-component)

onursumer:
Annotation column is handled in PR #204

src/pages/patientView/mutation/MutationInformationContainer.tsx
```diff
@@ -203,7 +203,7 @@ export default class MutationInformationContainer extends React.Component<IMutat
         };
 
         // some columns are dependent on number of samples
-        if (this.props.sampleManager.samples.length > 1) {
```
alisman: we have to make sure this condition is respected below.  @onursumer recommened that instead of conditionally including the column, we instead use excluded property.  Is this any different with the new table implementation?

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -341,6 +345,51 @@ export default class PatientViewMutationTable extends React.Component<PatientVie
...
+            render: (d:Mutation[])=>AlleleFreqColumnFormatter.renderFunction(d, this.props.sampleManager),
+            sort:(d1:Mutation[], d2:Mutation[], ascending:boolean)=>{
+                return numberListSort(AlleleFreqColumnFormatter.getSortValue(d1, this.props.sampleManager),
+                                AlleleFreqColumnFormatter.getSortValue(d2, this.props.sampleManager),
+                                ascending);
+            },
+            tooltip:(<span>Variant allele frequency in the tumor sample</span>)
+        };
+
+        this._columns[MutationTableColumn.TUMORS] = {
```
alisman: per above, this column should only be included when there are more than one sample in sampleManager.  see condition above.

adamabeshouse: the visible columns are specified here: https://github.com/cBioPortal/cbioportal-frontend/pull/207/files#diff-b02249405d34f258fb2e74c6a0e83c73R505

adamabeshouse: when onur is done with the visible column UI we'll implement this constraint and any others through that channel

PR #208, opened 2017-03-08
New table ui

adamabeshouse:
Depends on https://github.com/cBioPortal/cbioportal-frontend/pull/207

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/msktable/MSKTable.tsx
+++ b/src/shared/components/msktable/MSKTable.tsx
@@ -3,12 +3,13 @@ import * as React from 'react';
 import {observable, computed, action} from "mobx";
 import {observer} from "mobx-react";
 import './styles.scss';
-import {PaginationControls} from "../paginationControls/PaginationControls";
+import {SHOW_ALL_PAGE_SIZE as PAGINATION_SHOW_ALL, PaginationControls} from "../paginationControls/PaginationControls";
 import DefaultTooltip from "../DefaultTooltip";
+import {ButtonToolbar} from "react-bootstrap";
 
 export type Column<T> = {
     name: string;
-    filter?:(data:T, filterString:string)=>boolean;
+    filter?:(data:T, filterString:string, filterStringUpper?:string, filterStringLower?:string)=>boolean;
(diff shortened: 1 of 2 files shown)


adamabeshouse:
[filter and pagination UI in new mutation table](https://trello.com/c/4765QbvL/151-filter-and-pagination-ui-in-new-mutation-table)

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -7,17 +7,21 @@ import {compareNumberLists} from '../../../../shared/lib/SortUtils';
...
 export default class AlleleFreqColumnFormatter {
     static barWidth = 6;
     static barSpacing = 3;
     static maxBarHeight = 12;
     static indexToBarLeft = (n:number) => n*(AlleleFreqColumnFormatter.barWidth + AlleleFreqColumnFormatter.barSpacing);
 
-    public static renderFunction(data:IColumnFormatterData<MutationTableRowData>, columnProps:any) {
-        const sampleOrder = columnProps.sampleManager.getSampleIdsInOrder();
+    public static renderFunction(mutations:Mutation[], sampleManager:SampleManager|null) {
+        if (!sampleManager) {
```
alisman: what's case where sampleManager is null?

adamabeshouse: that's the type of `sampleManager` in `PatientViewPage.render`. From there it looks like it's null until the mobxpromise `patientViewPageStore.patientViewData` is complete.

alisman: And TS complains if you don't allow for null?  b/c i think by definition it should not be null when table renders because there's a check for sampleManager around the table.

adamabeshouse: it was definitely complaining, and as of now I don't have the new table wrapped that way, maybe I should?

PR #212, opened 2017-03-10
Show tooltip when hovering over sample name/description, not just label

### LINKED ISSUE #187: Show tooltip when hovering over sample name/description (not just label)
Currently the sample mouseover tooltip at the top of the page only works when mousing over the circle with the label. We would like the mouseover to work when hovering over the text as well, similar to how it did in the old page.

screen shot 2017-02-23 at 4 15 53 pm

aradhyamakkar97:
# What? Why?
Fix #187  .

Please review @inodb

### DIFF
--- a/src/pages/patientView/sampleManager.tsx
+++ b/src/pages/patientView/sampleManager.tsx
@@ -9,6 +9,11 @@ import DefaultTooltip from 'shared/components/DefaultTooltip';
 import {cleanAndDerive} from './clinicalInformation/lib/clinicalAttributesUtil.js';
 import styles from './patientHeader/style/clinicalAttributes.scss';
 
+import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
+import { PatientViewPageStore } from './clinicalInformation/PatientViewPageStore';
+
+const patientViewPageStore = new PatientViewPageStore();
+
 // we need this to account for issue with rc-tooltip when dealing with large tooltip overlay content
 export function placeArrow(tooltipEl: any) {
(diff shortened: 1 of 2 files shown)


src/pages/patientView/sampleManager.tsx
```diff
@@ -81,6 +86,10 @@ class SampleManager {
         this.samples.map((sample)=>this.getComponentForSample(sample.id));
     }
 
+    private handleSampleClick(id: string){
```
alisman: looks like indentation is off

src/pages/patientView/sampleManager.tsx
```diff
@@ -9,6 +9,11 @@ import DefaultTooltip from 'shared/components/DefaultTooltip';
 import {cleanAndDerive} from './clinicalInformation/lib/clinicalAttributesUtil.js';
 import styles from './patientHeader/style/clinicalAttributes.scss';
 
+import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
+import { PatientViewPageStore } from './clinicalInformation/PatientViewPageStore';
+
+const patientViewPageStore = new PatientViewPageStore();
```
alisman: we don't want to instantiate a new store inside this component.  The store instance should be managed by the page.  Presentational components should define handlers (callbacks) that allow access to the store.  See handleSampleClick above.   Since we're dealing with this legacy code and have to hack, what I would do is use jQuery to bind to clicks on the DOM and then call the handler.   you can email me at lisman.aaron@gmail.com for more advice!  thanks for your work.

PR #221, opened 2017-03-15
Fix debouncing and optimize sorting

adamabeshouse:
# Checks
  the `-s` option)


adamabeshouse:
[(5) Fix debouncing performance issue](https://trello.com/c/etpCW8BF/162-5-fix-debouncing-performance-issue)

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -9,16 +9,37 @@ import {CopyDownloadControls} from "../copyDownloadControls/CopyDownloadControls
...
+    sort?:undefined,
+    sortValue:(data:T)=>(number|null)[];
+    sortType:"number[]";
+} | {
+    sort?:undefined,
+    sortValue:(data:T)=>string|null;
+    sortType:"string";
+} | {
+    sort?:undefined,
+    sortValue:(data:T)=>(string|null)[];
```
alisman: getSortValue?

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -9,16 +9,37 @@ import {CopyDownloadControls} from "../copyDownloadControls/CopyDownloadControls
...
+    sortType:"string[]";
+} | {
+    sort?:(a:T, b:T, ascending:boolean)=>number
```
alisman: what's the case where one would use sort instead of sortValue?

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -102,10 +123,46 @@ class MSKTableStore<T> {
...
+    @computed get sortColumnObject():Column<T>|undefined {
+        return this.columns.find((col:Column<T>)=>this.isVisible(col) && (col.name === this.sortColumn));
+    }
+
+    @computed get sortedData():T[] {
+        let ret = this.data;
+        const column = this.sortColumnObject;
+        if (column) {
+            if (column.sortValue) {
+                switch (column.sortType) {
```
alisman: We could probably just detect the type, no?

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -36,9 +36,12 @@ export default class MutationAssessorColumnFormatter
         };
     }
 
-    public static getSortValue(d:Mutation[]):(number|undefined)[]
+    public static getSortValue(d:Mutation[]):(number|null)[]
     {
-        const score = MutationAssessorColumnFormatter.getData(d).score;
+        let score:number|undefined|null = MutationAssessorColumnFormatter.getData(d).score;
+        if (score === undefined) {
```
alisman: we should try to comment stuff that expresses some obscure business logic.  Like, why can store be undefined. E.g. "sometimes data is missing"

src/shared/lib/debounce.ts
```diff
@@ -0,0 +1,38 @@
+export default function debounce<AccumulatorType, ArgsType>(fn:(acc:AccumulatorType)=>void,
```
alisman: I think the name debounce is too general for this thing.  It accumulates and then does something.
Also, using a closure, you do create an instance scope, but this is not a very clear pattern.  Instead, maybe we should make this a class, instantiate it in the cache and then give the class an add method.  
I'm also not totally sold that we need to libraritize this.  I think the cache having a timeout and then checking for its existence would be enough.  Could go either way on this one.  We should just be careful not to add too many "things" to our app.

src/pages/patientView/clinicalInformation/CohortVariantCountCache.ts
```diff
@@ -65,7 +66,25 @@ export default class CohortVariantCountCache {
...
+            (toFetch:EntrezToKeywordSet)=>{
+                const entrezToKeywordList:EntrezToKeywordList = {};
+                for (const entrez of Object.keys(toFetch)) {
+                    const entrezGeneId = parseInt(entrez, 10);
+                    entrezToKeywordList[entrezGeneId] = Array.from(toFetch[entrezGeneId]);
+                }
+                this.populate(entrezToKeywordList);
+            },
+            (acc:EntrezToKeywordSet, entrezGeneId:number, keyword:string|null)=>{
+                acc[entrezGeneId] = acc[entrezGeneId] || new Set<string>();
```
adamabeshouse: oh, oops I thought it was core

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: sure

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: if you have complicated rules about what comes before what which aren't easily mapped to a numerical/string value? maybe unnecessary

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: nope 😛

adamabeshouse: the reason for this was I was multiplexing to which default sort method to use. but this problem may be obsolete bc _.sortBy handles every type of argument

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
adamabeshouse: sure

src/shared/lib/SortUtils.ts
```diff
@@ -1,5 +1,7 @@
 import * as _ from 'lodash';
```
adamabeshouse: yep

src/shared/lib/debounce.ts
adamabeshouse: I libraritized this to DRY it up, which I do think is necessary.

debounceAccumulate maybe?

Is this not a common pattern? debounce is a function creator, which includes its necessary variables in the closure of the returned function. I've seen and used this pattern before.

src/shared/lib/SortUtils.ts
adamabeshouse: done

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: done

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: done, only sortBy now

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: sortBy

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
adamabeshouse: done

src/shared/lib/debounce.ts
adamabeshouse: renamed to accumulatingDebounce

PR #227, opened 2017-03-17
New tabs

alisman:
This pull requests switches from legacy url qstring params to hash routing. Urls will now look like this:
http://localhost:3000/#/patient?studyId=lgg_ucsf_2014&tab=summaryTab&caseId=P04

Also introduces new MSKTabs, replacing Bootstrap tabs, which don't have the necessary "mount on show" but don't 'unmount on hide' behavior.

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -115,22 +112,34 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         //TODO: this should be done by a module so that it can be reused on other pages
         const qs = queryString.parse((window as any).location.search);
 
+        const reaction1 = reaction(
```
alisman: this may be temporary.  the url route is now modelled in it's own store.  page store could be made to observe this directly, but for now I decided just to wire them up this way.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -115,22 +112,34 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+            () => props.routing.query,
+            query => {
 
-        patientViewPageStore.studyId = qs['cancer_study_id'] + '';
+                if ('studyId' in query) {
+                    patientViewPageStore.studyId = query.studyId;
+                } else {
+                    alert("You must have a study Id");
```
onursumer: It may not be a good idea to handle errors with `alert`s

alisman: Yes, for sure.  This is just a placeholder.  I want to have a global error handling/messaging solution.

src/pages/patientView/patientHeader/PatientHeader.tsx
```diff
@@ -24,15 +24,15 @@ export default class PatientHeader extends React.Component<IPatientHeaderProps,
...
             <Popover key={patient.id} id={'popover-sample-' + patient.id}>
                 <ClinicalInformationPatientTable showTitleBar={false} data={patient.clinicalData} />
             </Popover>
         );
     }
 
-    private getOverlayTriggerPatient(patient: ClinicalInformationData['patient']) {
+    private getOverlayTriggerPatient(patient: any) {
         return patient && (
             <OverlayTrigger
```
adamabeshouse: should this be a DefaultTooltip?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -500,24 +509,24 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
-                        </Tab>
+                        </MSKTab>
                     )}
 
 
                     {  (patientViewPageStore.pathologyReport.isComplete && patientViewPageStore.pathologyReport.result.length > 0 ) &&
-                    (<Tab eventKey={8} id="summarTab" title="Pathology Report">
-                        <PathologyReport  pdfs={patientViewPageStore.pathologyReport.result} />
-                    </Tab>)
+                    (<MSKTab key={3} id="summarTab" linkText="Pathology Report">
```
adamabeshouse: while we're here, let's fix this typo summarTab?

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: is it conventional to trigger routing from the page store?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -45,13 +42,14 @@ import Timeline from "./timeline/Timeline";
...
 import {getCbioPortalApiUrl, getHotspotsApiUrl, getHotspots3DApiUrl} from "../../shared/api/urls";
+import { MSKTabs, MSKTab } from "../../shared/components/MSKTabs/MSKTabs";
 
 const patientViewPageStore = new PatientViewPageStore();
 
 (window as any).patientViewPageStore = patientViewPageStore;
 
 export interface IPatientViewPageProps {
-    store?: RootState;
+    routing: any;
```
adamabeshouse: can we type this any more specifically?

src/appBootstrapper.jsx
```diff
@@ -1,40 +1,41 @@
...
-import { configureStore } from './redux/configureStore';
+import { Provider } from 'mobx-react';
+import { hashHistory, Router } from 'react-router';
+import { RouterStore, syncHistoryWithStore  } from 'mobx-react-router';
+import ExtendedRoutingStore from './shared/lib/ExtendedRouterStore';
+import { computed, extendObservable } from 'mobx';
 import makeRoutes from './routes';
 import lodash from 'lodash';
 
+let RT = require('route-parser');
```
adamabeshouse: I was going to ask why you're using require instead of importing, but it doesn't look like you're using this?

src/shared/components/MSKTabs/MSKTabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+    render(){
+        if (this.props.children) {
+
+            let children: any[] = (this.props.children as any[]);
+
+            let hasActive: boolean = false;
+            let defaultTab: string;
+
+            const arr: any = _.reduce(React.Children.toArray(children), (memo: any, child: any) => {
```
adamabeshouse: I'm confused by what's going on with cloning, could you please explain what's going on here? Why can't you just set the prop on the existing child? Isn't cloning costly?

src/shared/components/MSKTabs/MSKTabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+export class MSKTab extends React.Component<IMSKTabProps,{}> {
+
+    constructor(props: IMSKTabProps){
+        super(props);
+    }
+
+    render(){
+        const hidden = ((this.props.hidden===true) ? 'hidden' : '');
+        return <div className={classnames({ 'msk-tab':true, 'hidden':(this.props.hidden === true)  })}>{this.props.children}</div>
```
adamabeshouse: on line 22 and 23, the predicate `this.props.hidden === true` is equivalent to the predicate `this.props.hidden`

src/shared/components/MSKTabs/MSKTabs.tsx
```diff
@@ -0,0 +1,106 @@
...
+                arr[0] = React.cloneElement(children[0], {
+                    hidden: false,
+                });
+                defaultTab = children[0].props.id;
+            }
+
+            return <div id={(this.props.id) ? this.props.id : ''}
+                        className={ classnames(this.props.className) }>
+                <ul className="nav nav-tabs">{
+                    React.Children.map(children, (tab: any) => {
```
adamabeshouse: doing some more explicit typing, e.g. children should be of type MSKTab, would be nice if it's not too much effort. But if it is then it would definitely be `Nice-to-have`

src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -0,0 +1,51 @@
...
+        return queryString;
+    }
+
+    pathname: string;
+
+    push: (newHash: string)=>void;
+
+    location: any;
+
+    @action updateRoute(newParams: any, newPath: string) {
```
adamabeshouse: I believe that URL manipulation is wrought with edge cases and annoying subtleties. It seems preferable to me to do use an existing library to do the URL manipulation work in this file.

src/pages/patientView/PatientViewPage.tsx
alisman: The page store in an instance of a library called mobx-react-router.  This handles sychornoization between store and url.  So yes, we only interact with the url via the store.

src/pages/patientView/patientHeader/PatientHeader.tsx
alisman: Yes.  Will make card to migrate it.

src/shared/components/MSKTabs/MSKTabs.tsx
alisman: Yeah, struggled with this.  I'm sure it can be done but couldn't get it to work.

src/shared/components/MSKTabs/MSKTabs.tsx
alisman: A component's props are immutable.  So this is the only way (in my research) for a parent to modify a child's props.  The clone is shallow so the DOM and children are merged not cloned.  I've verified that the tabs children are NOT re-instantiated.  Perhaps worth more research, but I think this is actually equivaent to what happens when declaratively in render. (not sure about that)

src/shared/lib/ExtendedRouterStore.ts
alisman: We can discuss.  Note that the actual reading and writing to URL is taken care of by library.  We construct that path, which is really just a matter of serializing a dictionary.  But I'm happy to consider a library here to avoid unknown complications.

src/shared/components/MSKTabs/MSKTabs.tsx
alisman: True.  22 is legacy.  Will delete.

src/appBootstrapper.jsx
alisman: Actually, this is a library which does route parsing and construction.  So it's possible it could be choice for your comment below.  I had some reason for moving away from it.  Not sure.

src/shared/components/MSKTabs/MSKTabs.tsx
adamabeshouse: got it, thanks

PR #228, opened 2017-03-17
sort samples by id

inodb:
Sort samples by id. Add style changes to top to be more in line with old page.


inodb:
superseded by https://github.com/cBioPortal/cbioportal-frontend/pull/230

PR #232, opened 2017-03-20
put pagination controls below MSKtable when nec

alisman:
Put pagination control at bottom of table, but only show it when there is pagination. 

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/shared/components/msktable/MSKTable.tsx
+++ b/src/shared/components/msktable/MSKTable.tsx
@@ -10,6 +10,7 @@ import {serializeData} from "shared/lib/Serializer";
 import DefaultTooltip from "../DefaultTooltip";
 import {ButtonToolbar} from "react-bootstrap";
 import * as _ from "lodash";
+import { If } from 'react-if';
 
 export type Column<T> = {
     name: string;
@@ -75,6 +76,10 @@ class MSKTableStore<T> {
     public set itemsPerPage(i:number) {
         this._itemsPerPage = i;
         this.page = this.page; // trigger clamping in page setter
+    }
+
+    @computed get showingAllRows(): Boolean {
+        return this.sortedFilteredData.length <= this.itemsPerPage || this.itemsPerPage === -1
     }
 
     @computed public get page() {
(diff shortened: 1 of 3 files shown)


src/pages/patientView/styles.scss
```diff
@@ -1,43 +1,46 @@
...
+    margin-bottom: 20px;
+
+    td {
+      vertical-align: top;
+      padding-left: 10px;
+      padding-bottom: 4px;
+    }
 
-  .patientSamples {
+    .patientIcon {
```
alisman: this stuff probably belongs in a css module associated with the patient header.  Also the patient header should be a component.

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -77,6 +78,10 @@ class MSKTableStore<T> {
         this.page = this.page; // trigger clamping in page setter
     }
 
+    public get showingAllRows(): Boolean {
```
adamabeshouse: I don't understand why this wouldn't be a computed. Let's talk tomorrow.

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -397,6 +403,23 @@ export default class MSKTable<T> extends React.Component<MSKTableProps<T>, {}> {
                 headers={this.store.headers}
                 rows={this.store.rows}
             />
+
+            <If condition={this.store.showingAllRows === false}>
+            <ButtonToolbar style={{marginLeft:0}} className='text-center'>
+                <PaginationControls
```
onursumer: Most of the options for this component is the same as the one on the top. Maybe we should we create a function which returns a `PaginationControls` component. We can pass `style` and `className` as parameters.

Also, would it be better to try to handle styling with the `className` property instead of introducing a new `style` property?

alisman: Point #1: done.

For #2, I used to be religious about avoiding inline styles, but now I think that small pixel pushing stuff should be inline.  Otherwise you end up bloating style sheets, having to come up with all kinds of names and worry about uniqueness.  It's more mentally efficient to put it inline.  Though we wouldn't want this for an element that gets repeated many times, e.g. table cell.

PR #236, opened 2017-03-21
Annotation column for the CNA table

onursumer:
Adds the annotation column to the CNA table

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/CopyNumberUtils.ts
+++ b/src/shared/lib/CopyNumberUtils.ts
@@ -1,0 +1,9 @@
+export enum Alterations {
+    Deletion = -2,
+    Amplification = 2
+}
+
+export function getAlterationString(alteration: number): string
+{
+    return Alterations[alteration] || '';
+}
--- a/src/shared/lib/OncoKbUtils.ts
+++ b/src/shared/lib/OncoKbUtils.ts
@@ -1,5 +1,5 @@
-import {Query, EvidenceQueries, EvidenceQueryRes, Evidence} from "shared/api/generated/OncoKbAPI";
 import * as _ from 'lodash';
+import {Query, EvidenceQueries, EvidenceQueryRes, Evidence, IndicatorQueryResp} from "shared/api/generated/OncoKbAPI";
 
 /**
  * @author Selcuk Onur Sumer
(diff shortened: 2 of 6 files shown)


onursumer:
[(5) Add annotation column to copy number table](https://trello.com/c/EU3a4qKE/180-add-annotation-column-to-copy-number-table)

src/shared/lib/CopyNumberUtils.ts
```diff
@@ -0,0 +1,22 @@
+export const ALTERATIONS = {
```
alisman: /src/pages/patientView/copyNumberAlterations/column/CnaColumnFormatter.tsx

Checkout out my solution in this file.  Is there a reason why it's deletion here and deep deletion there?  

Also, see use of ENUM.  Could possibly serve in place of switch statement.  Maybe we cansolidate code?

src/shared/lib/OncoKbUtils.ts
```diff
@@ -80,6 +80,28 @@ const LEVELS = {
     all: ['4', 'R3', '3B', '3A', 'R2', '2B', '2A', 'R1', '1', '0']
 };
 
+export const ALTERATIONS = {
+    mutation: "MUTATION",
+};
+
+/**
+ * Generates <query id, indicator> map.
```
alisman: are these comments autogenerated?  seems like typescript gives us this.  like, instead of putting return type in comment, why not specify with ts?

onursumer: Actually, this is a manual comment. Right, it seems a bit redundant, since we already have a `@returns` with the proper `TypeScript` definition.

src/shared/lib/CopyNumberUtils.ts
onursumer: For the OncoKB API, there are only 2 valid values for CNA: `DELETION` and `AMPLIFICATION`. And here is the corresponding legacy code:

```JavaScript
switch(mapData.getValue(_id, "alter")) {
    case 2:
        alter = 'amplification';
        break;
    case -2:
        alter = 'deletion';
        break;
    default:
        alter = null;
}
```

src/shared/lib/OncoKbUtils.ts
adamabeshouse: I think what @alisman means is 
`export function generateIdToIndicatorMap(data:IndicatorQueryResp[]):{ [id:string]: IndicatorQueryResp }
{
}`

onursumer: Oh, yea, that makes sense :)

src/shared/lib/CopyNumberUtils.ts
onursumer: @alisman now using enum for this, looks much simpler!

PR #240, opened 2017-03-22
Use URI library to parse hash and type MSKTabs strongly

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -35,16 +32,13 @@ export default class ExtendedRouterStore extends (RouterStore as FunctionConstru
             }
         });
```
alisman: I attempted to use the URI library to handle construction of the hash, but couldn't find a simple way to do it.   Also, I realized that $.param handles the uri encoding of parameter values, as demonstrated in test.  So, at this point, the only thing we're handling ourselves is the building of the parameter map, which seems very generic to me.

src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -2,19 +2,16 @@ import { RouterStore } from 'mobx-react-router';
...
     @computed get query(): {[key: string] : string} {
-        // var route = new RT(stores.routing.location.pathname);
-        var queryString: { [key: string] : string } = {};
-        decodeURIComponent(this.location.search).replace(
-            new RegExp("([^?=&]+)(=([^&]*))?", "g"),
-            function($0, $1, $2, $3) { queryString[$1] = $3; return $0}
-        )
-        return queryString;
+
+        return (URI.parseQuery(this.location.search) as {[key: string] : string});
```
alisman: A note, we already have at least 2 other URL building/parsing libraries installed.  So, at some point we should consolidate.  I will make a card for that.

src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -35,16 +32,13 @@ export default class ExtendedRouterStore extends (RouterStore as FunctionConstru
...
         let params = $.param(newQuery);
 
-        // if we have a new path, adopt it
-        // otherwise use current path
-
-        newPath = newPath || this.location.pathname;
-
-        newPath = (newPath[0] === '/') ? newPath : '/'+ newPath;
+        // put a leading slash if there isn't one
+        path = (path[0] === '/') ? path : '/' + path;
```
adamabeshouse: this is the kind of stuff that leads to issues in edge cases but this is probably simple enough to be fine

src/shared/lib/ExtendedRouterStore.ts
adamabeshouse: actually I was just playing around and it looks like the query is already parsed, I guess as part of RouterStore. But this.location.query is a map of parameter names to url decoded parameter values.

PR #241, opened 2017-03-22
Show image for patients that are in Darwin

inodb:
screen shot 2017-03-22 at 5 17 12 pm

This `darwinAccessUrl` parameter is currently injected in the `patient_view.jsp` from `cbioportal/cbioportal`. Therefore we can expect it to be there. Eventually we would like this to be provided through an API endpoint.

darwin


inodb:
[(5) Link to Darwin in patient view (Needed by MSKCC)](https://trello.com/c/4OH73UxL/95-5-link-to-darwin-in-patient-view-needed-by-mskcc)

src/pages/patientView/patientHeader/PatientHeader.tsx
```diff
@@ -18,11 +17,26 @@ export default class PatientHeader extends React.Component<IPatientHeaderProps,
...
+        // use JSP injected Darwin URL window.darwinAccessUrl
+        // TODO: use internal API service instead, once this exists
+        let darwinAccessUrl: any = ((window as any).darwinAccessUrl);
+
+        if (darwinAccessUrl !== undefined && darwinAccessUrl !== null && darwinAccessUrl !== '') {
+            // add link to darwin
+            let darwinImgSrc = require("./images/darwin_logo.png");
+            return (<a target={'_blank'} href={darwinAccessUrl}><img style={{paddingLeft:'5px'}} src={darwinImgSrc} /></a>);
+        } else {
+            return '';
```
alisman: I think return null would be better than empty string

src/pages/patientView/patientHeader/PatientHeader.tsx
```diff
@@ -18,11 +17,26 @@ export default class PatientHeader extends React.Component<IPatientHeaderProps,
...
 
+    private getDarwinAccessUrl() {
+        // use JSP injected Darwin URL window.darwinAccessUrl
+        // TODO: use internal API service instead, once this exists
+        let darwinAccessUrl: any = ((window as any).darwinAccessUrl);
+
+        if (darwinAccessUrl !== undefined && darwinAccessUrl !== null && darwinAccessUrl !== '') {
+            // add link to darwin
+            let darwinImgSrc = require("./images/darwin_logo.png");
+            return (<a target={'_blank'} href={darwinAccessUrl}><img style={{paddingLeft:'5px'}} src={darwinImgSrc} /></a>);
```
alisman: you can just use "_blank" if value is not dynamic.

src/pages/patientView/patientHeader/PatientHeader.tsx
```diff
@@ -18,11 +17,26 @@ export default class PatientHeader extends React.Component<IPatientHeaderProps,
...
+                {this.getDarwinAccessUrl()}
             </div>
         );
 
     }
 
+    private getDarwinAccessUrl() {
+        // use JSP injected Darwin URL window.darwinAccessUrl
+        // TODO: use internal API service instead, once this exists
+        let darwinAccessUrl: any = ((window as any).darwinAccessUrl);
```
alisman: I would do :string | null | undefined

typescript would probably be fine with just string.

PR #245, opened 2017-03-24
add sort by sample collection date

inodb:
Sort sample labels based on collection date


inodb:
[(5) Order samples in timeline based on time](https://trello.com/c/i1vjKkKI/167-5-order-samples-in-timeline-based-on-time)

src/pages/patientView/sampleManager.tsx
```diff
@@ -18,23 +17,71 @@ export function placeArrow(tooltipEl: any) {
...
-    let sortedSampleIDs: string[] = [];
-    sortedSampleIDs = sortedSampleIDs.concat(samples.map((sample) => sample.id)).sort(naturalSort).reverse();
-
-    return _.sortBy(samples, function(sample) {
-        return sortedSampleIDs.indexOf(sample.id);
+                            clinicalDataLegacyCleanAndDerived: { [s:string]:any },
+                            events?: any) {
+    // natural sort
+    let naturalSortedSampleIDs: string[] = [];
+    naturalSortedSampleIDs = naturalSortedSampleIDs.concat(samples.map((sample) => sample.id)).sort(naturalSort);
```
alisman: not sure you need concat.  map produces new array, so just assign result of map().sort() to naturalSortedSampleIDs?

src/pages/patientView/sampleManager.tsx
```diff
@@ -18,23 +17,71 @@ export function placeArrow(tooltipEl: any) {
...
+
+    // based on sample collection data (timeline event)
+    let collectionDayMap: {[s:string]:number} = {};
+    if (events) {
+        let specimenEvents = events.filter((e: ClinicalEvent) => (e.eventType === 'SPECIMEN'));
+
+        for (let i: number = 0; i < specimenEvents.length; i++) {
```
alisman: You can do specimentEvents.forEach

src/pages/patientView/sampleManager.tsx
```diff
@@ -18,23 +17,71 @@ export function placeArrow(tooltipEl: any) {
...
+            let sampleAttr = _.find(specimenEvents[i].attributes, (attr: ClinicalEventData) => {
+                // TODO: This is legacy support for old timeline data that does not use SAMPLE_ID, but one of the specrefnum
+                return (attr.key === "SAMPLE_ID" || attr.key === "SpecimenReferenceNumber" || attr.key === "SPECIMEN_REFERENCE_NUMBER") &&
+                    (naturalSortedSampleIDs.indexOf(attr.value) !== -1);
+            });
+            if (sampleAttr && sampleAttr.value !== null && sampleAttr.value !== undefined) {
```
alisman: i think the first condition actually covers both the subsequent null and undefined checks.

src/pages/patientView/sampleManager.tsx
```diff
@@ -18,23 +17,71 @@ export function placeArrow(tooltipEl: any) {
...
-    return _.sortBy(samples, function(sample) {
-        return sortedSampleIDs.indexOf(sample.id);
+                            clinicalDataLegacyCleanAndDerived: { [s:string]:any },
+                            events?: any) {
+    // natural sort
+    let naturalSortedSampleIDs: string[] = [];
+    naturalSortedSampleIDs = naturalSortedSampleIDs.concat(samples.map((sample) => sample.id)).sort(naturalSort);
+
+    // based on sample collection data (timeline event)
+    let collectionDayMap: {[s:string]:number} = {};
```
alisman: look into using reduce to accomplish building an object like this.  it's a little more semantic.

src/pages/patientView/sampleManager.tsx
```diff
@@ -18,23 +17,71 @@ export function placeArrow(tooltipEl: any) {
...
+    let sampleTypeOrdering: string[] = ['Primary'];
+    let sampleOrder: sampleOrderT[] = [];
+    
+    for (let i: number = 0; i < samples.length; i++) {
+        let id = samples[i].id;
+        // 1. based on sample collection data (timeline event)
+        let eventOrdering = collectionDayMap[id];
+
+        // 2. if cases have derived normalized case types, put primary first
+        let sampleTypeIndex = sampleTypeOrdering.indexOf(clinicalDataLegacyCleanAndDerived[id].DERIVED_NORMALIZED_CASE_TYPE) * -1;
```
alisman: why multiple by -1 here?

inodb: ah because `indexOf` returns -1 if it's not found. In this case the's only one item in the sampleTypeOrdering list, so it works. Prolly more clear to handle the -1 case separately.

src/pages/patientView/sampleManager.tsx
inodb: TypeScript gave me some error here, if I want to cast to `string[]`. Not sure why, I'll try another time

inodb:
Thanks for reviewing @alisman !

src/pages/patientView/sampleManager.tsx
inodb: Ah yes so the problem here is that it complains about not being able to sort on an immutable data structure. When I make a call to Mutable it also gives an error, which is why I for the contrived concatenation approach.

src/pages/patientView/sampleManager.tsx
inodb: Done, see below

Issue #205, opened 2017-03-07
Show error when patient/sample does not exist


inodb:
In case of going to a URL where there's no patient/sample, show a message that no patient or sample with that id was found. Only in log is a good start.

latifaabdullayeva:
Hi @inodb, 

After reading the source code, I came up with a few questions:
 - Do we need to show error message based on response from API or just local validation?
 - Are there any specific requirements to ids of patients / samples?

Thank you.

PR #251, opened 2017-03-28
Show error when patient/sample does not exist #205

Rohanhacker:
fix show error when patient/sample does not exist #205 

screenshot from 2017-03-28 12-16-41


alisman:
Hi, thanks for this work.  We've actually come up with a generalized parameter validation solution that handles this.  You can see this in current integration branch.  Sorry I cannot accept this PR.

PR #253, opened 2017-03-28
Portal integration

alisman:
Adjust styles for integration with portal project.

Map legacy qstring params for studyId, caseId, sampleId and cohort to new hash parameters

# Checks
  the `-s` option)

### DIFF
--- a/src/appBootstrapper.jsx
+++ b/src/appBootstrapper.jsx
@@ -8,13 +8,12 @@ import { computed, extendObservable } from 'mobx';
 import makeRoutes from './routes';
 import lodash from 'lodash';
 import $ from 'jquery';
+import queryString from 'query-string'
 
 // make sure lodash doesn't overwrite (or set) global underscore
 lodash.noConflict();
 
 const routingStore = new ExtendedRoutingStore();
-
-window.routingStore = routingStore;
 
 const stores = {
     // Key can be whatever you want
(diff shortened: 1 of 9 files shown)


src/appBootstrapper.jsx
```diff
@@ -8,21 +8,46 @@ import { computed, extendObservable } from 'mobx';
...
     // Key can be whatever you want
     routing: routingStore,
     // ...other stores
 };
 
-const history = syncHistoryWithStore(hashHistory, routingStore);
+var history = syncHistoryWithStore(hashHistory, routingStore);
+
+
+var qs = queryString.parse((window).location.search);
```
alisman: this code doesn't probably doesn't belong here because it's page specific, but at the same time we want to handle this mapping prior to initial routing.  it's a big complicated.  didn't want to hold it up since it works for purpose of this phase of project.

src/routes.jsx
```diff
@@ -34,8 +34,10 @@ function lazyLoadComponent(loader) {
 
 export const makeRoutes = () => (
         <Route path="/" component={Container}>
-            <Route path="home" getComponent={lazyLoadComponent(HomePage)} />
+            <Route path="/home" getComponent={lazyLoadComponent(HomePage)} />
             <Route path="/patient" getComponent={lazyLoadComponent(PatientViewPage)} />
+            <Redirect from="*" to="/patient" />
+            <IndexRedirect to="/patient" />
```
alisman: Star does not match empty string.  For that reason we have the IndexRedirect

src/appBootstrapper.jsx
```diff
@@ -8,21 +8,46 @@ import { computed, extendObservable } from 'mobx';
 import makeRoutes from './routes';
 import lodash from 'lodash';
 import $ from 'jquery';
+import queryString from 'query-string'
 
 // make sure lodash doesn't overwrite (or set) global underscore
 lodash.noConflict();
 
-const routingStore = new ExtendedRoutingStore();
+var routingStore = new ExtendedRoutingStore();
```
adamabeshouse: why const=>var, here and all over this file?

PR #254, opened 2017-03-28
Lazy loading OncoKB card content from cache

onursumer:
# What? Why?
- Deferred loading of the OncoKB card data (`Evidence` and `PMID` data) until an actual mouse-over event on a particular OncoKB icon.
- Added a generic `SimpleCache` implementation for Evidence and PMID data caching. We may want to replace these cache implementations with the ones that extend a generic version of `SampleGeneCache`, especially if we need the "accumulating debounce" feature.

# Checks
  the `-s` option)


onursumer:
[(5) OncoKB tooltip data should not be loaded before an actual mouse over event on the icon.](https://trello.com/c/WRLCC9km/146-5-oncokb-tooltip-data-should-not-be-loaded-before-an-actual-mouse-over-event-on-the-icon)

PR #263, opened 2017-04-04
move cosmic data to store.  don't fetch cosmic or onkobk when no muta…

alisman:
We were getting errors on patients where mutations service returned empty.  this work prevents cosmic data and onko kb services from being called when there are no mutations present.

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.spec.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.spec.ts
@@ -2,18 +2,20 @@
  * Created by aaronlisman on 3/2/17.
  */
 
-import { handlePathologyReportCheckResponse } from './PatientViewPageStore';
+import { handlePathologyReportCheckResponse, PatientViewPageStore } from './PatientViewPageStore';
 // import React from 'react';
 import { assert } from 'chai';
 // import { shallow, mount } from 'enzyme';
-// import sinon from 'sinon';
+import sinon from 'sinon';
 // //import AppConfig from 'appConfig';
 // import request from 'superagent';
 
 describe('ClinicalInformationSamplesTable', () => {
 
-    before(()=>{
+    let store: PatientViewPageStore;
 
+    before(()=>{
+        store = new PatientViewPageStore();
     });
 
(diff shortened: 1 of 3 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -420,45 +444,54 @@ export class PatientViewPageStore
...
                 });
             } else {
                 return [];
             }
-    }}, []);
+        }
+    }, []);
+
+
+    async oncoKbDataInvoke(){
```
alisman: we have to move these out to make them more testable

src/pages/patientView/clinicalInformation/PatientViewPageStore.spec.ts
```diff
@@ -33,4 +35,36 @@ describe('ClinicalInformationSamplesTable', () => {
         assert.deepEqual(result,[]);
     });
 
+    it('won\'t fetch cosmic data if there are no mutations', ()=>{
```
alisman: these test are not comprehensive but they catch the key thing, that we handle case where no mutations are present.

src/pages/patientView/clinicalInformation/PatientViewPageStore.spec.ts
```diff
@@ -33,4 +35,36 @@ describe('ClinicalInformationSamplesTable', () => {
...
+
+    it('won\'t fetch onkokb data if there are no mutations', ()=>{
+
+        const fetchStub = sinon.stub();
+
+        let mockInstance = {
+            mutationData: { result:[] }
+        };
+
+        store.oncoKbDataInvoke.apply(mockInstance).then((data: any)=>{
```
alisman: i should be checking to make sure the api client is not called, but this is still pretty good.  to make this easier we need to put api clients on instance of store OR set up a solution called rewire so that we can stub imports of modules.

PR #266, opened 2017-04-04
Use Darwin servlet

inodb:
Use `checkDarwinAcces.do` servlet from parent cBioPortal to determine Darwin URL for patient. Only show in case this is enabled. Previously this variable was passed through JSP, but now that we have the routing on frontend after `#` symbol, we had to use a servlet instead.

Requires https://github.com/cBioPortal/cbioportal/pull/2289 to be merged

### DIFF
--- a/my-index.ejs
+++ b/my-index.ejs
@@ -6,11 +6,9 @@
     <script>
         window.devContext = true;
         window.defaultRoute = "/patient";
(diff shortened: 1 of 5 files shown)


inodb:
[Darwin as a servlet](https://trello.com/c/htB15vRC/211-darwin-as-a-servlet)

PR #267, opened 2017-04-04
Improved OncoKB sort behavior (annotation column)

onursumer:
# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/OncoKbUtils.ts
+++ b/src/shared/lib/OncoKbUtils.ts
@@ -11,6 +11,7 @@ const ONCOGENIC_CLASS_NAMES:{[oncogenic:string]: string} = {
     'Likely Neutral': 'likely-neutral',
     'Unknown': 'unknown-oncogenic',
     'Inconclusive': 'unknown-oncogenic',
+    'Predicted Oncogenic': 'oncogenic',
     'Likely Oncogenic': 'oncogenic',
     'Oncogenic': 'oncogenic',
 };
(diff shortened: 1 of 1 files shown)


onursumer:
[Annotation Column Sorting does not work properly](https://trello.com/c/vRQMEI2j/206-annotation-column-sorting-does-not-work-properly)

src/shared/lib/OncoKbUtils.ts
```diff
@@ -19,8 +20,9 @@ const ONCOGENIC_CLASS_NAMES:{[oncogenic:string]: string} = {
 // (used for sorting purposes)
 const ONCOGENIC_SCORE:{[oncogenic:string]: number} = {
```
alisman: Should we use Typescript enum for this?

onursumer: That makes sense, we should use enum here for other numerical mappings as well.

onursumer: @alisman Actually, this looks a bit trickier than I initially thought. `oncogenic` is a variable string value coming from data. Trying to get the numerical value for a variable string (`OncogenicScore[oncogenic]`) does not work.

onursumer: To be more specific, I am getting this TS error:
`error TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'`

alisman: I see.  Look at documentation and it seems TS enums don't work with dynamic accessors.  So you are doing the right thing.

PR #271, opened 2017-04-05
Make COSMIC and Cohort sort descending first

adamabeshouse:
# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/msktable/MSKTable.tsx
+++ b/src/shared/components/msktable/MSKTable.tsx
@@ -12,6 +12,7 @@ import {ButtonToolbar} from "react-bootstrap";
 import * as _ from "lodash";
 import { If } from 'react-if';
 
+type SortDirection = 'asc' | 'desc';
 export type Column<T> = {
     name: string;
     filter?:(data:T, filterString:string, filterStringUpper?:string, filterStringLower?:string)=>boolean;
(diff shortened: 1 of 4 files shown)


src/shared/components/msktable/MSKTable.tsx
```diff
@@ -20,13 +21,14 @@ export type Column<T> = {
     render:(data:T)=>JSX.Element;
     download?:(data:T)=>string;
     tooltip?:JSX.Element;
+    firstSortDirection?:SortDirection;
```
onursumer: Why don't we name this one also `initialSortDirection`?

adamabeshouse: feels a little confusing, initial feels like it implies it'll be like that on start-up

alisman: What about defaultSortDirection?  I agree firstSortDirection is kind of strange.

onursumer: default sounds better than initial!

adamabeshouse: default it is!

adamabeshouse:
[sort cosmic and cohort columns descending at first](https://trello.com/c/a29zM76b/214-sort-cosmic-and-cohort-columns-descending-at-first)

PR #273, opened 2017-04-06
HotspotSet data structure for rapid checking for mutation overlap

adamabeshouse:
# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/HotspotSet.ts
+++ b/src/shared/lib/HotspotSet.ts
@@ -1,0 +1,95 @@
+/* HotspotSet
+ *
+ * This class is designed for quickly querying whether a mutation is a hotspot.
+ * It's supported to query by a single residue position, or a range of residue positions,
+ *  and it does a binary search to quickly check whether the input range or position
+ *  overlaps with any hotspot regions specified in the constructor.
+ *
+ * The class does not support telling you exactly which input intervals are overlapped-with,
+ *  that information is lost. But it will tell you true or false: whether there was some input interval
+ *  that is overlapped-with.
+ */
+export default class HotspotSet {
+    private regions:[number,number][];
+    constructor(intervals:[number,number][]) {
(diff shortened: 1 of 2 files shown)


src/shared/lib/HotspotSet.ts
```diff
@@ -0,0 +1,95 @@
...
+ *  that is overlapped-with.
+ */
+export default class HotspotSet {
+    private regions:[number,number][];
+    constructor(intervals:[number,number][]) {
+        // intervals: [L:number, U:number][] where L <= U
+        this.regions = createHotspotRegions(intervals);
+    }
+
+    public check = function(x:number, y?:number):boolean {
```
adamabeshouse: oops, accidentally left it as `= function`

src/shared/lib/HotspotSet.ts
```diff
@@ -0,0 +1,95 @@
...
+ *  and it does a binary search to quickly check whether the input range or position
+ *  overlaps with any hotspot regions specified in the constructor.
+ *
+ * The class does not support telling you exactly which input intervals are overlapped-with,
+ *  that information is lost. But it will tell you true or false: whether there was some input interval
+ *  that is overlapped-with.
+ */
+export default class HotspotSet {
+    private regions:[number,number][];
+    constructor(intervals:[number,number][]) {
```
onursumer: how about defining a type for interval? Something like this:
```JavaScript
type Interval {
    start: number,
    end: number
}
```

And then define `regions` and `intervals` like this:

```JavaScript
    regions: Interval[]
    intervals: Interval[]
```

onursumer:
@adamabeshouse have you tested `new HotspotSet(...)` in the browser?

I am getting a red screen of death with this error:
`TypeError: hotspotSet.check is not a function`

onursumer:
[Match hotspots properly](https://trello.com/c/OjU5RC58/213-match-hotspots-properly)

PR #275, opened 2017-04-06
Lazy mobx cache

adamabeshouse:
Implements generic LazyMobXCache class, and uses it to reimplement a few of the mutation table caches.

# Checks
  the `-s` option)


src/shared/lib/LazyMobXCache.ts
```diff
@@ -0,0 +1,171 @@
+import Immutable from "seamless-immutable"; // need to use immutables so mobX can observe the cache shallowly
+import accumulatingDebounce from "./accumulatingDebounce";
+import {observable, action} from "mobx";
+import {AccumulatingDebouncedFunction} from "./accumulatingDebounce";
+
+export type CacheData<T> = {
```
alisman: we could look into ts namespaces to group these types.

src/shared/lib/LazyMobXCache.ts
```diff
@@ -0,0 +1,171 @@
...
+        for (const datum of data) {
+            const datumKey = this.dataToKey(datum);
+            toMerge[datumKey] = {
+                status: "complete",
+                data: datum
+            };
+            queryKeyHasData[datumKey] = true;
+        }
+
+        for (const queryKey of Object.keys(queryKeyHasData)) {
```
alisman: could use a little commenting here to make this easier to understand

src/shared/lib/LazyMobXCache.ts
adamabeshouse: good idea, how do you think it should look?

src/shared/lib/LazyMobXCache.ts
adamabeshouse: done

adamabeshouse:
[generalize caches in a better way, and replace all caches with one cache to rule them all (SampleGeneCache, SimpleCache, CohortVariantCountCache)](https://trello.com/c/sgquyH0w/190-generalize-caches-in-a-better-way-and-replace-all-caches-with-one-cache-to-rule-them-all-samplegenecache-simplecache-cohortvaria)

src/shared/lib/LazyMobXCache.ts
alisman: i'm not totally sure!  I think they are for disambiguating very common names, like Cache for example.  Not a big deal.  I will just merge this and we can revisit namespaces later.

PR #277, opened 2017-04-07
In patient view page, show loading sign instead of tabs when target tab (eg from URL) is pending

adamabeshouse:
# Checks
  the `-s` option)


alisman:
I think this gets away from the declarative nature of MSKTabs.  We don't want the page to have know to much about tabs.  What about the following: An MSK tab can have an existencePending property which accepts a promise.  Then MSKTab parent can handle the business of what tabs (if any) to show.  This has benefit of generalizing this across use of MSKTab and hiding yucky logic.  Let me know if you think this would work.

adamabeshouse:
https://github.com/cBioPortal/cbioportal-frontend/pull/284

PR #278, opened 2017-04-07
improved hotspot mutation matching

onursumer:
Do not merge before merging #273

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/AnnotationUtils.ts
+++ b/src/shared/lib/AnnotationUtils.ts
@@ -1,7 +1,12 @@
+import * as _ from 'lodash';
 import {CosmicMutation} from "shared/api/generated/CBioPortalAPIInternal";
 import {ICosmicData} from "shared/components/mutationTable/column/CosmicColumnFormatter";
-import {IMyCancerGenome, IMyCancerGenomeData} from "pages/patientView/mutation/column/AnnotationColumnFormatter";
+import {
+    IMyCancerGenome, IMyCancerGenomeData, IHotspotIndex, IHotspotLookup
+} from "pages/patientView/mutation/column/AnnotationColumnFormatter";
 import {HotspotMutation} from "../api/generated/CancerHotspotsAPI";
+import {Mutation} from "../api/generated/CBioPortalAPI";
+import HotspotSet from "./HotspotSet";
 
 /**
  * Utility functions related to annotation data.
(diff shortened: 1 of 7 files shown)


src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,129 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
     return map;
 }
 
-export function geneAndProteinPosToHotspots(hotspots:HotspotMutation[]):{[key:string]: boolean}
+export function geneAndProteinPosToHotspots(hotspots:HotspotMutation[]):{[key:string]: HotspotMutation}
```
onursumer: not used anymore, should remove

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,129 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+}
+
+export function initHotspotSets(hotspotIndex: IHotspotIndex)
+{
+    _.keys(hotspotIndex).forEach((geneSymbol:string) => {
+        _.keys(hotspotIndex[geneSymbol]).forEach((hotspotType:string) => {
+            const intervals = hotspotIndex[geneSymbol][hotspotType].hotspotMutations.map(
+                (hotspot:HotspotMutation):[number, number] => {
+                    // start (and optionally end) positions
+                    const positions = hotspot.residue.match(/[0-9]+/g) || [];
```
onursumer: instead of parsing the residue, we should use `hotspot.aminoAcidPosition.start` and `hotspot.aminoAcidPosition.end` values

onursumer:
[Match hotspots properly](https://trello.com/c/OjU5RC58/213-match-hotspots-properly)

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+                }
+            );
+
+            hotspotIndex[geneSymbol][hotspotType].hotspotSet = new HotspotSet(intervals);
+        });
+    });
+}
+
+export function isHotspot(mutation:Mutation, index:IHotspotIndex):boolean
+{
```
adamabeshouse: since this function would be run many times if there are a lot of mutations, there should be a few optimizations:

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+    let isHotspot = false;
+    let hotspotType:string|null = null;
+
+    if (mutation.mutationType.toLowerCase().indexOf("missense") > -1) {
```
adamabeshouse: save `mutation.mutationType.toLowerCase()` in a variable instead of recomputing it a few times here

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+        hotspotType = "single residue";
+    }
+    else if (mutation.mutationType.toLowerCase().indexOf("inframe") > -1 ||
+        mutation.mutationType.toLowerCase().indexOf("in_frame") > -1 ||
+        mutation.mutationType.toLowerCase().indexOf("in frame") > -1)
+    {
+        hotspotType = "in-frame indel";
+    }
+
+    if (hotspotType &&
```
adamabeshouse: save `index[mutation.gene.hugoGeneSymbol]` in a variable instead of accessing twice

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+        index[mutation.gene.hugoGeneSymbol])
+    {
+        const indexedData = index[mutation.gene.hugoGeneSymbol][hotspotType];
+
+        if (indexedData)
+        {
+            isHotspot = indexedData.hotspotSet.check(mutation.proteinPosStart, mutation.proteinPosEnd);
```
adamabeshouse: just confirming this will work with missense, i.e. it will have both proteinPosStart and proteinPosEnd?

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+    }
+
+    return isHotspot;
+}
+
+export function is3dHotspot(mutation:Mutation, index:IHotspotIndex):boolean
+{
+    let isHotspot = false;
+
+    if (index[mutation.gene.hugoGeneSymbol]) {
```
adamabeshouse: same as above, I think you should save `index[mutation.gene.hugoGeneSymbol]` in a variable

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
-    return map;
+    initHotspotSets(index);
+
+    return index;
+}
+
+export function initHotspotSets(hotspotIndex: IHotspotIndex)
+{
+    _.keys(hotspotIndex).forEach((geneSymbol:string) => {
+        _.keys(hotspotIndex[geneSymbol]).forEach((hotspotType:string) => {
```
adamabeshouse: save `hotspotIndex[geneSymbol]` in a variable instead of multiple access

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,113 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
-    const map: {[key:string]: boolean} = {};
+    const index:IHotspotIndex = {};
 
-    // create a map for a faster lookup
-    hotspots.forEach(function(hotspot:HotspotMutation) {
-        const positions = hotspot.residue.match(/[0-9]+/g) || []; // start (and optionally end) positions
-        const key = [hotspot.hugoSymbol.toUpperCase()].concat(positions).join("_");
-        map[key] = true;
+    hotspots.forEach((hotspot:HotspotMutation) => {
+        let geneIndex = index[hotspot.hugoSymbol.toUpperCase()];
```
adamabeshouse: definitely worth saving `hotspot.hugoSymbol.toUpperCase()` in a variable, and also `index[..]`

src/shared/lib/AnnotationUtils.ts
onursumer: I think it is better to add more checks here, we may have `-1` or `0` instead of `undefined` as a `proteinPosEnd` value

src/shared/lib/AnnotationUtils.ts
```diff
@@ -45,18 +50,131 @@ export function geneToMyCancerGenome(myCancerGenomes:IMyCancerGenome[]):IMyCance
...
+    });
+}
+
+export function isHotspot(mutation:Mutation, index:IHotspotIndex):boolean
+{
+    let isHotspot = false;
+    let hotspotType:string|null = null;
+    const mutationType = mutation.mutationType && mutation.mutationType.toLowerCase();
+
+    if (mutationType.indexOf("missense") > -1) {
```
adamabeshouse: you need to check, `mutationType` could be undefined by this point

onursumer: Good catch! Actually the problem here is the `Mutation` model definition in the generated API client. The model assumes that those fields always exist, but in practice I don't think that's the case.

adamabeshouse: good point, definitely

PR #286, opened 2017-04-11
Don't show VAF plot in Genomic Overview if there are no mutations

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -304,7 +304,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                         sampleLabels={sampleManager.sampleLabels}
                                         sampleColors={sampleManager.sampleColors}
                                         sampleManager={sampleManager}
-                                        getWidth={()=>$(window).width()-140}
+                                        getContainerWidth={()=>$(window).width()}
                                     />
                                 )
                             }
--- a/src/pages/patientView/genomicOverview/GenomicOverview.tsx
+++ b/src/pages/patientView/genomicOverview/GenomicOverview.tsx
@@ -2,6 +2,7 @@ import * as React from 'react';
 import * as ReactDOM from 'react-dom';
 import * as _ from 'lodash';
 import $ from 'jquery';
+import {If, Then, Else} from 'react-if';
 import Tracks from './Tracks';
 import {ThumbnailExpandVAFPlot} from '../vafPlot/ThumbnailExpandVAFPlot';
 import {Mutation} from "../../../shared/api/generated/CBioPortalAPI";
@@ -15,7 +16,7 @@ interface IGenomicOverviewProps {
     sampleLabels: {[s:string]:string};
     sampleColors: {[s:string]:string};
     sampleManager: SampleManager;
-    getWidth: ()=>number;
+    getContainerWidth: ()=>number;
 }
 
 export type MutationFrequencies = number[];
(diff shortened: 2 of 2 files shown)


src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -45,6 +46,12 @@ export default class GenomicOverview extends React.Component<IGenomicOverviewPro
 
     }
 
+    private getWidth() {
+
+        return this.props.getWidth() - ((this.props.mutations.length > 0) ? 140 : 40);
```
alisman: in theory the adjustment really belongs outside component.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -304,7 +304,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                         sampleLabels={sampleManager.sampleLabels}
                                         sampleColors={sampleManager.sampleColors}
                                         sampleManager={sampleManager}
-                                        getWidth={()=>$(window).width()-140}
+                                        getWidth={()=>$(window).width() - ((patientViewPageStore.mutationData.result.length > 0) ? 140 : 40)}
```
adamabeshouse: can we move this logic into the GenomicOverview class as a static method or something?

src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -55,14 +57,16 @@ export default class GenomicOverview extends React.Component<IGenomicOverviewPro
...
                 <Tracks mutations={this.props.mutations} key={Math.random()} sampleManager={this.props.sampleManager} width={this.props.getWidth()} cnaSegments={this.props.cnaSegments} />
-                <ThumbnailExpandVAFPlot
-                    data={this.state.frequencies}
-                    order={this.props.sampleManager.sampleIndex}
-                    colors={this.props.sampleColors}
-                    labels={labels}
-                    overlayPlacement="right"
-                    cssClass="vafPlot"
-                />
+                <If condition={this.props.mutations.length > 0}>
```
adamabeshouse: why do you prefer to use `<If>` here instead of the ternary operator?

src/pages/patientView/PatientViewPage.tsx
alisman: I started with that but I actually think this shenanigans belongs outside, at the page (consumer) level.  The GenomicOverview should not have to know where it is being rendered.

src/pages/patientView/genomicOverview/GenomicOverview.tsx
alisman: you mean like ( condition && (<whatever/ >) ?
I mean, given the choice I prefer declarative.  We can't use it in all cases because ts doesn't recognize it as null check.  But where we can use it, I like it.  I suppose one could argue for consistency?

adamabeshouse: I'm not sure how I feel. On one hand I like this more. On the other hand I actually think in general it makes it slightly less readable, because you need to match up the `Else` and `If`. But in this case it's just the simple <If>. So I don't really care either way, was just curious.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: I just really don't like these exposed magic numbers, and the logic being duplicated out here and also inside it in the `<If>`.

Maybe the prop can be `getContainerWidth={()=>$(window).width()}` and inside it can use that to do this logic?

alisman: Yeah, I like that.  getContainerWidth.   and you're right the adjustment belongs inside because it's intrinsic information.  will do.

PR #287, opened 2017-04-11
Split timeline on TREATMENT_TYPE and AGENT

inodb:
Before only split occured on SUBTYPE and AGENT. In some cases there is no
SUBTYPE. Therefore better to split on TREATMENT_TYPE. Also floor
AGE/OS_MONTHS/DFS_MONTHS.

Fix https://github.com/cBioPortal/cbioportal/issues/1784

### DIFF
--- a/src/pages/patientView/timeline/legacy.js
+++ b/src/pages/patientView/timeline/legacy.js
@@ -101,7 +101,7 @@ export function buildTimeline(params, caseIds, patientInfo, clinicalDataMap, cas
         var prefix = prefixes[i];
         if (patientInfo[prefix + "OS_STATUS"] === "DECEASED" &&
             prefix + "OS_MONTHS" in patientInfo) {
-            var days = parseInt(parseInt(patientInfo[prefix + "OS_MONTHS"]) * 30.4);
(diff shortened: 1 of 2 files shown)


src/pages/patientView/timeline/legacy.js
```diff
@@ -166,6 +166,7 @@ export function buildTimeline(params, caseIds, patientInfo, clinicalDataMap, cas
     window.pvTimeline =
         window.pvTimeline
             .splitByClinicalAttributes("Treatment", ["SUBTYPE", "AGENT"])
+            .splitByClinicalAttributes("Treatment", ["TREATMENT_TYPE", "AGENT"])
```
onursumer: So, now we want to split both by `subtype` and `treatment_type`, right?

Just making sure that this is the desired behavior :)

inodb: Thanks for checking! It's splitting the treatment track by either subtype & agent or treatment_type & agent. Only if all events in that track have values for the given attributes. That's the way `splitByClinicalAttributes` works. I think it's best to try and split the treatment timelines as much as we can to make it more easy to interpret.

PR #291, opened 2017-04-14
More compact patient view page

onursumer:
This PR addresses the issue https://github.com/cBioPortal/cbioportal/issues/2300

# Checks
  the `-s` option)

### DIFF
--- a/src/globalStyles/global.scss
+++ b/src/globalStyles/global.scss
@@ -88,8 +88,8 @@ th.reactable-header-sort-asc:after {
     }
 
     hr {
-      margin-top:10px;
-      margin-bottom:15px;
+      margin-top:5px;
+      margin-bottom:5px;
     }
 
     > .msk-tab {
@@ -122,9 +122,10 @@ th.reactable-header-sort-asc:after {
 }
 
 .table tbody tr td {
-  padding:6px 10px !important;
+  padding:3px 10px !important;
   border-top:none;
 }
 
-
-
+.table {
+  margin-bottom: 5px;
+}
--- a/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
+++ b/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
@@ -106,11 +106,6 @@ export default class CopyNumberTableWrapper extends React.Component<{ store:Pati
 
         return (
             <div>
-                <FeatureTitle
-                    title="Copy Number Alterations"
-                    isLoading={this.props.store.discreteCNAData.isPending}
-                />
-
             {
                 (this.props.store.geneticProfileIdDiscrete.isComplete && this.props.store.geneticProfileIdDiscrete.result === undefined) && (
                     <div className="alert alert-info" role="alert">Copy Number Alterations are not available.</div>
@@ -127,7 +122,9 @@ export default class CopyNumberTableWrapper extends React.Component<{ store:Pati
                         data={this.props.store.discreteCNAData.result}
                         initialSortColumn="Annotation"
                         initialSortDirection="desc"
-                        initialItemsPerPage={25}
+                        initialItemsPerPage={10}
+                        itemsLabel="Copy Number Alteration"
+                        itemsLabelPlural="Copy Number Alterations"
                     />
                 )
             }
--- a/src/shared/components/msktable/MSKTable.tsx
+++ b/src/shared/components/msktable/MSKTable.tsx
@@ -33,6 +33,8 @@ type MSKTableProps<T> = {
     initialSortColumn?: string;
     initialSortDirection?:SortDirection;
     initialItemsPerPage?:number;
+    itemsLabel?:string;
+    itemsLabelPlural?:string;
     showFilter?:boolean;
     showCopyDownload?:boolean;
     copyDownloadProps?:ICopyDownloadControlsProps;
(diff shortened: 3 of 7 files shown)


src/globalStyles/global.scss
```diff
@@ -122,9 +122,10 @@ th.reactable-header-sort-asc:after {
...
-  padding:6px 10px !important;
+  padding:3px 10px !important;
   border-top:none;
 }
 
-
-
+.table {
+  margin-bottom: 5px;
+}
```
onursumer: If this global style is not desired, we can also define a style specific to patient view only

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -37,6 +37,7 @@ type MSKTableProps<T> = {
     showCopyDownload?:boolean;
     copyDownloadProps?:ICopyDownloadControlsProps;
     showPagination?:boolean;
+    paginationTextSuffix?:string;
```
onursumer: this may not be the best name for this property

src/shared/components/msktable/styles.scss
```diff
@@ -4,4 +4,16 @@ th.sort-des:after {
 
 th.sort-asc:after {
   content: ' \25B2';
-}
\ No newline at end of file
+}
+
+.topPagination {
+  .textBetweenButtons {
+    opacity: 1.00 !important;
```
onursumer: Since the text between the buttons is actually a "disabled button", it doesn't have full opacity. This restores the text to its full opacity

src/shared/components/msktable/styles.scss
```diff
@@ -4,4 +4,16 @@ th.sort-des:after {
...
+  }
+}
+
+.tableMainToolbar {
+  .form-group {
+    margin-bottom: 5px;
```
onursumer: this is to reduce the margin between the toolbar and the table

onursumer:
[Even more compact patient view](https://trello.com/c/Ni3UMp2Y/216-even-more-compact-patient-view)

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -436,27 +437,37 @@ export default class MSKTable<T> extends React.Component<MSKTableProps<T>, {}> {
         />
     }
 
-    render() {
+    buildTextBetweenTheButtons(): string {
```
adamabeshouse: I think we should make `@computed get textBetweenButtons` and put it in `MSKTableStore`

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -125,9 +120,10 @@ export default class CopyNumberTableWrapper extends React.Component<{ store:Pati
                     <CNATableComponent
                         columns={orderedColumns}
                         data={this.props.store.discreteCNAData.result}
+                        paginationTextSuffix="Copy Number Altered Gene(s)"
```
onursumer: let's change this to Copy Number Alteration(s)

src/shared/components/msktable/MSKTable.tsx
onursumer: @adamabeshouse Looks like this function depends on `paginationTextSuffix` which is a property of `MSKTable`. It might be better to make it `@computed` within `MSKTable` instead of `MSKTableStore`.

adamabeshouse: Look at `MSKTableStore.setProps`

When we had talked about this, Aaron preferred to keep all this logic in the store.

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: Maybe `itemsLabel`? (and `itemsLabelPlural` :) )

src/shared/components/msktable/MSKTable.tsx
onursumer: oh, I didn't notice that. okay then store it is!

src/globalStyles/global.scss
inodb: Let's maybe do patient view/table specific. Just to keep the bootstrap styles intact

src/shared/components/msktable/MSKTable.tsx
onursumer: done!

src/shared/components/msktable/MSKTable.tsx
onursumer: @adamabeshouse also moved `@computed get colVisProp(): IColumnVisibilityDef[]` to the store

src/globalStyles/global.scss
onursumer: We already have some global css for table components: `.table tbody tr td`

I did the modification in the same file for consistency. Actually this style also applies to the tooltip tables as well. For example, cosmic tooltip had unnecessary white-space below the table.

src/shared/components/msktable/MSKTable.tsx
```diff
@@ -278,6 +282,60 @@ class MSKTableStore<T> {
...
+                id: column.name,
+                name: column.name,
+                visible: this.columnVisibility[column.name]
+            });
+        });
+
+        return colVisProp;
+    }
+
+    @computed get textBetweenTheButtons(): string
```
adamabeshouse: `textBetweenPaginationButtons` I think. Or `paginationStatusText` would be even better I think

PR #299, opened 2017-04-18
Fix genomic overview rendering

alisman:
# What? Why?
https://github.com/cBioPortal/cbioportal/issues/2370

There was a problem with genomic overview chromosome numbers when the GO rendered in a hidden tab.  This has to do with some deep SVG box measuring bug.  Seemed to me the best way to fix and avoid issues like this going forward was to hide the tabs by position instead of display:none.  Can anyone think of a drawback?

image

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/globalStyles/global.scss
+++ b/src/globalStyles/global.scss
@@ -129,3 +129,9 @@ th.reactable-header-sort-asc:after {
 .table {
   margin-bottom: 5px;
 }
+
+.hiddenByPosition {
+  position: absolute;
+  top: -9999px;
+  left: -9999px;
+}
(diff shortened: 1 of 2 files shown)


src/globalStyles/global.scss
```diff
@@ -129,3 +129,9 @@ th.reactable-header-sort-asc:after {
 .table {
   margin-bottom: 5px;
 }
+
+.hiddenByPosition {
+  position: absolute;
```
adamabeshouse: there HAS to be a better way to do this

alisman: heh.   i think the fact that it happens only when tab renders in display:none indicates it a really nasty issue.   we could bust our asses to see if we can fix it (it might not be fixable) but then it will surely happen again in some other visualization down the road.  it's pretty common to hide things in this fashion.  i think this is actually relatively elegant.

PR #316, opened 2017-04-21
added loader icon for OncoKB

onursumer:
oncokb_loader

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2383.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/annotation/OncoKB.tsx
+++ b/src/shared/components/annotation/OncoKB.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {Modal} from 'react-bootstrap';
 import {observer} from "mobx-react";
+import {Circle} from "better-react-spinkit";
 import DefaultTooltip from 'shared/components/DefaultTooltip';
 import annotationStyles from "./styles/annotation.module.scss";
 import oncogenicIconStyles from "./styles/oncogenicIcon.module.scss";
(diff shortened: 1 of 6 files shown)


src/shared/components/annotation/OncoKB.tsx
```diff
@@ -100,7 +101,12 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
 
             if (this.showFeedback)
             {
-                oncoKbContent = this.feedbackModal(this.props.indicator);
+                oncoKbContent = (
```
onursumer: This is a minor fix related to the feedback modal. (OncoKB icon was disappearing upon click on feedback button)

typings/react-spinkit.d.ts
```diff
@@ -5,6 +5,8 @@ declare module 'react-spinkit'
         spinnerName?: string;
         style?: React.CSSProperties
         noFadeIn?: boolean;
+        className?: string;
+        overrideSpinnerClassName?: string;
```
onursumer: these properties are documented, but was missing and the typings

onursumer:
@alisman @adamabeshouse seems like OncoKB people very much disliked the `double-bounce` or `pulse` loader. I will change this back to the classical `circle` one.

onursumer:
Updated the PR to use `better-react-spinkit` library instead of `react-spinkit`. `better-react-spinkit` is better in terms of customizability and scalibility.

@alisman I will create another PR soon which replaces all instances of `react-spinkit` with `better-react-spinkit`

PR #320, opened 2017-04-25
Added a base class for Mutation table

onursumer:
- This PR adds a new base class `MutationTable`.
- `PatientViewMutationTable` now extends `MutationTable`
- Also moved some cache classes under shared/cache since they will be used by results page as well
- We will need to add definitions for results page specific columns (such as # of Muts, and Cancer Type) into the class `ResultsViewMutationTable` (partially implemented)

# Checks
  the `-s` option)


src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,340 @@
...
+    }
+
+    @computed protected get columns():Column<Mutation[]>[] {
+        return this.orderedColumns.reduce((columns:Column<Mutation[]>[], next:MutationTableColumnType) => {
+            columns.push(this._columns[next]);
+            return columns;
+        }, []);
+    }
+
+    public render()
```
onursumer: Classes extending `MutationTable` will most likely override `render` method. So, maybe we should not implement any default render method here?

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,340 @@
...
+    hotspots?: IHotspotData;
+    cosmicData?:ICosmicData;
+    oncoKbData?:IOncoKbData;
+    mrnaExprRankGeneticProfileId?:string;
+    discreteCNAGeneticProfileId?:string;
+    columns:MutationTableColumnType[];
+    data:Mutation[][];
+}
+
+export enum MutationTableColumnType {
```
onursumer: I was not able to find an easy way to extend an `enum`. So, we still have patient view specific columns within this enum. We will also need to add results page specific columns here as well...

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,340 @@
...
+    constructor(props:P)
+    {
+        super(props);
+        this._columns = {};
+        this.generateColumns();
+    }
+
+    protected getSamples():string[] {
+        if (this.props.sampleIds) {
+            return this.props.sampleIds as string[];
```
onursumer: Even if there is a check for undefined, typescript still complains here. I think this is related to the statement `P extends IMutationTableProps`. This is also the case for `this.props.discreteCNACache` below.

src/shared/components/mutationTable/MutationTable.tsx
adamabeshouse: that's okay, I think that still makes sense. The base mutation table should implement all the columns, and each use of it should just specify which ones and the order

src/shared/components/mutationTable/MutationTable.tsx
adamabeshouse: what's the purpose of the generic type P?

src/shared/components/mutationTable/MutationTable.tsx
adamabeshouse: ideally they would not override render, they would just change whatever properties are needed to be specified. i.e. the columns and initialItemsPerPage, and anything else.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -22,7 +22,8 @@ import {getSpans} from './clinicalInformation/lib/clinicalAttributesUtil.js';
 import CopyNumberTableWrapper from "./copyNumberAlterations/CopyNumberTableWrapper";
 import {reaction, computed} from "mobx";
 import Timeline from "./timeline/Timeline";
-import {default as PatientViewMutationTable, MutationTableColumnType} from "./mutation/PatientViewMutationTable";
+import {default as PatientViewMutationTable} from "./mutation/PatientViewMutationTable";
+import {MutationTableColumnType} from "shared/components/mutationTable/MutationTable"
 import PathologyReport from "./pathologyReport/PathologyReport";
 import { MSKTabs, MSKTab } from "../../shared/components/MSKTabs/MSKTabs";
 import validateParameters from '../../shared/lib/validateParameters';
```
adamabeshouse: I was thinking that we'd move `mutationTableColumns` on line 90, move all that stuff into PatientViewMutationTable

src/shared/components/mutationTable/MutationTable.tsx
onursumer: There are patient view speicific properties such as sample manager, mrna cache, etc. I wanted to define them in `PatientViewMutationTable` instead of the base class.

adamabeshouse: I think you don't need that because if PVTableProps has all the properties of TableProps, then PVTableProps can be passed in as TableProps even if it has other things too. Not 100% sure on that though.

src/shared/components/mutationTable/MutationTable.tsx
alisman: That's seems fair to me.

src/shared/components/mutationTable/MutationTable.tsx
alisman: Does PVTableProps actually extend TableProps?  if so, I think it will pass.

src/shared/components/mutationTable/MutationTable.tsx
alisman: I think I agree with adam.  What is the case for different render implementations?

onursumer: Makes sense, I will add more properties to avoid overriding the render method.

src/shared/components/mutationTable/MutationTable.tsx
onursumer: This is how I define PV Table properties:

```JavaScript
interface IPatientViewMutationTableProps extends IMutationTableProps {
    sampleManager:SampleManager | null;
    mrnaExprRankCache?:MrnaExprRankCache;
    variantCountCache?:VariantCountCache;
}
```

@adamabeshouse @alisman If I don't do `P extends IMutationTableProps`, typescript complains when I try to access patient view specific properties (such as `this.props.mrnaExprRankGeneticProfileId`) in `PatientViewMutationTable`. Because in that case `this.props` is an `IMutationTableProps` but not an `IPatientViewMutationTableProps`.

adamabeshouse: another option is to implement this as wrapper classes, where PatientViewMutationTable and ResultsViewMutationTable each have a render() method that outputs a <MutationTable> element

onursumer: At this point, I kind of liked the idea of inheritance with one render method over composition with multiple render methods. But that doesn't necessarily mean that inheritance is better than composition.

onursumer:
[Create ResultsPageMutationTable component](https://trello.com/c/gdxrMAKC/237-create-resultspagemutationtable-component)

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -1,167 +1,80 @@
...
             name:"mRNA Expr.",
             render:(d:Mutation[])=>(this.props.mrnaExprRankCache
-                                        ? MrnaExprColumnFormatter.renderFunction(d, this.props.mrnaExprRankCache)
-                                        : (<span></span>)),
+                ? MrnaExprColumnFormatter.renderFunction(d, this.props.mrnaExprRankCache)
+                : (<span></span>)),
             order: 182
         };
+
         this._columns[MutationTableColumnType.COHORT] = {
```
adamabeshouse: should we put these column defs into MutationTable?

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -0,0 +1,109 @@
...
+            MutationTableColumnType.COSMIC,
+            MutationTableColumnType.TUMOR_ALLELE_FREQ,
+            MutationTableColumnType.NORMAL_ALLELE_FREQ
+        ]
+    };
+
+    protected generateColumns() {
+        super.generateColumns();
+
+        this._columns[MutationTableColumnType.SAMPLE_ID] = {
```
adamabeshouse: should we put these columns into MutationTable?

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -0,0 +1,433 @@
...
+
+    protected generateColumns() {
+        this._columns = {};
+
+        this._columns[MutationTableColumnType.SAMPLE_ID] = {
+            name: "Sample ID",
+            render: SampleColumnFormatter.renderFunction,
+            sortBy: SampleColumnFormatter.getTextValue,
+            visible: true,
+            order: 10
```
onursumer: `order` should not be defined here, move this (and other order values) to either `PatientViewMutationTable` or `ResultsViewMutationTable`

PR #322, opened 2017-04-26
LazyMobXTable data store, and row highlighting

adamabeshouse:
Allow passing in a data store to LazyMobXTable, and implement row highlighting in LazyMobXTable, which is used according to `store.highlight`

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/simpleTable/styles.scss
+++ b/src/shared/components/simpleTable/styles.scss
@@ -5,6 +5,14 @@ th {
   background-color: rgb(255,255,255); // solid background
 }
 
+tbody>tr:nth-of-type(odd).highlight {
+  background-color: #E0E000 !important;
+}
+
 tbody>tr:nth-of-type(even) { // solid background - hard coded hack, react-bootstrap .table-striped makes ODD # rows grey
   background-color: rgb(255,255,255);
+}
+
+tbody>tr:nth-of-type(even).highlight {
+  background-color: #E9E900;
 }
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
@@ -2,7 +2,7 @@ import React from 'react';
 import {assert} from 'chai';
 import {shallow, mount, ReactWrapper} from 'enzyme';
 import sinon from 'sinon';
-import {lazyMobXTableSort, default as LazyMobXTable, Column} from "./LazyMobXTable";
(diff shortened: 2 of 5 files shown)


src/shared/components/msktable/MSKTable.tsx
```diff
@@ -356,20 +387,42 @@ class MSKTableStore<T> {
...
+
     @action setProps(props:MSKTableProps<T>) {
         this.columns = props.columns;
-        this.data = props.data;
         this._itemsLabel = props.itemsLabel;
         this._itemsLabelPlural = props.itemsLabelPlural;
         this._columnVisibility = this.resolveColumnVisibility(props.columns);
+
+        this.dataStore = props.dataStore || new MSKTableDataStore<T>(props.data || []);
+        this.dataStore.setSortAscending(this.sortAscending);
```
adamabeshouse: this may cause data binding direction problems later if we have more than one component influencing sorting. thoughts?

src/shared/components/simpleTable/styles.scss
```diff
@@ -5,6 +5,14 @@ th {
...
+tbody>tr:nth-of-type(odd).highlight {
+  background-color: #E0E000 !important;
+}
+
 tbody>tr:nth-of-type(even) { // solid background - hard coded hack, react-bootstrap .table-striped makes ODD # rows grey
   background-color: rgb(255,255,255);
+}
+
+tbody>tr:nth-of-type(even).highlight {
+  background-color: #E9E900;
```
onursumer: Should we also add `!important` here?

src/shared/components/msktable/MSKTable.tsx
adamabeshouse: fixed in upcoming commit

src/shared/components/simpleTable/styles.scss
adamabeshouse: it seems that it's not necessary because:
(1) bootstrap .table-striped doesn't define any style for the even rows, which is why we were having that problem with no background for those rows
(2) because it has `.highlight` it has higher specificity than the rule above

but I guess there's no harm in adding it? I'm not experienced with CSS so whatever you think.

onursumer: I think we are good as long as no other rule is overriding this one. I just wanted to make sure that we don't lose the highlight feature later at some point.

alisman: i think best to avoid !important if possible because then you can't override it!!

adamabeshouse: @alisman that's what I was thinking too

PR #323, opened 2017-04-27
Fix #2388 clinical data wrapping

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/2388. Puts clinical data and link in same span with flex-wrap set to wrap. So the clinical data in the patient view header wraps as expected.

screen shot 2017-04-26 at 8 35 07 pm


inodb:
[clinical data wrapping in header](https://trello.com/c/p77tO2YL/239-clinical-data-wrapping-in-header)

inodb:
There's a minor thing that the new line might start with `,`. That's because we use a CSS rule with `:before` for putting the commas. Unfortunately last-child and first-child don't work in flex display

PR #326, opened 2017-04-30
Fix cBioPortal/cbioportal#2306 mutations uncalled indicator

inodb:
Fix cBioPortal/cbioportal#2306. In mutation table on the patient view show a light sample label if the mutation wasn't called, but the alt read count > 0. This works if mutations are imported with a `*_uncalled` profile. It's useful for cases where the mutation is seen in the sequencing data, but we can't call the mutation confidently because it's there in such low abundance. This is common in ctDNA.

screen shot 2017-05-02 at 8 18 42 pm


inodb:
[cfDNA uncalled indicator](https://trello.com/c/YkmIgRXv/241-cfdna-uncalled-indicator)

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -524,6 +528,32 @@ export class PatientViewPageStore {
...
+                        geneticProfileId,
+                        mutationFilter: {
+                            sampleIds: this.samples.result.map((sample: Sample) => sample.sampleId)
+                        } as MutationFilter,
+                        projection: "DETAILED"
+                    });
+                for (const mutation of uncalledMutations) {
+                        // mutationStatus is hardly used, we can use it to keep
+                        // track of uncalled mutations
+                        mutation.mutationStatus = 'UNCALLED';
```
onursumer: I am not sure if it is a good idea to update/overwrite an API generated data field for custom data purposes...

jjgao:
From @schultzn:

reduce the number of grey numbered icons to only those samples in which there is some read support for the mutation

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: I agree, this smells bad. My opinion would be build a map that you can query to check if a mutation in a sample is uncalled, and pass that into the table and the tumor formatter.
`index.isMutationCalled(m:Mutation):boolean`

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -102,8 +103,14 @@ export default class PatientViewMutationTable extends MutationTable<IPatientView
...
         };
+        // only hide tumor column if there are no mutations
+        const hasUncalledMutations =  this.props.data.some((row:Mutation[]) => {
+            return row.some((m:Mutation) => {
+                return Boolean(m.geneticProfileId && m.geneticProfileId.endsWith(GENETIC_PROFILE_UNCALLED_POSTFIX));
+            });
+        });
         this._columns[MutationTableColumnType.TUMORS].shouldExclude = ()=>{
-            return this.getSamples().length < 2;
+            return this.getSamples().length < 2 && !hasUncalledMutations;
```
adamabeshouse: I think this should be ```@computed private get hasUncalledMutations```

src/shared/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const GENETIC_PROFILE_MUTATIONS = '_mutations';
```
adamabeshouse: seems like this should be `GENETIC_PROFILE_MUTATIONS_SUFFIX`

src/shared/constants.ts
```diff
@@ -0,0 +1,2 @@
...
+export const GENETIC_PROFILE_UNCALLED_POSTFIX = `${GENETIC_PROFILE_MUTATIONS}_uncalled`;
```
adamabeshouse: I'd call this `GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX`

inodb:
@adamabeshouse Thanks so much for the review! Let me know if there are any other comments

src/shared/constants.ts
```diff
@@ -0,0 +1,2 @@
+export const GENETIC_PROFILE_MUTATIONS_SUFFIX = '_mutations';
+export const GENETIC_PROFILE_UNCALLED_SUFFIX = `${GENETIC_PROFILE_MUTATIONS_SUFFIX}_uncalled`;
```
adamabeshouse: plz change to more unambiguous `GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX`

PR #328, opened 2017-05-01
Add Classname Prop to LazyMobXTable

Tamba32:
# What? Why?

Delete unused function in SampleTable.  Add optional className prop to LazyMobXTable and SimpleTable. Add CSS Module for SampleTable.

### DIFF
--- a/src/shared/components/simpleTable/SimpleTable.tsx
+++ b/src/shared/components/simpleTable/SimpleTable.tsx
@@ -2,32 +2,34 @@ import * as React from 'react';
 import './styles.scss';
 import { If, Then,  Else } from 'react-if';
 
-type SimpleTableProps = {
+interface ISimpleTableProps {
     headers:JSX.Element[];
     rows:JSX.Element[];
     noRowsText?:string;
+    className?:string;
 };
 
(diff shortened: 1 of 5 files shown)


src/shared/components/simpleTable/SimpleTable.tsx
```diff
@@ -2,32 +2,34 @@ import * as React from 'react';
...
-                <td style={{textAlign:'center'}} colSpan={this.props.headers.length}>
-                    {this.props.noRowsText || "There are no results."}
+                <td style={{textAlign:'center'}} colSpan={headers.length}>
+                    {noRowsText || "There are no results."}
                 </td>
             </tr>];
 
         return (
-            <table className="table table-striped table-border-top">
+            <table className={`table table-striped table-border-top${className ? ' ' + className : ''}`}>
```
alisman: https://www.npmjs.com/package/classnames  
this library makes it must easier to deal with conditional classes in jsx attribute

alisman: we have it installed already though we should be using it more religiously.

PR #329, opened 2017-05-01
Refactor DataSetPage

Tamba32:
# What? Why?
Refactor DataSetPage to use ReactJS and WebAPI. Add DataSetPage route.

### DIFF
--- a/src/pages/datasetView/DatasetList.tsx
+++ b/src/pages/datasetView/DatasetList.tsx
@@ -1,29 +1,61 @@
 import * as React from "react";
 import * as _ from 'lodash';
-import { Table, Tr, Td }  from 'reactableMSK';
-import { CancerStudy }  from 'shared/api/generated/CBioPortalAPI';
+import { Table, Tr, Td } from 'reactableMSK';
+import { CancerStudy } from 'shared/api/generated/CBioPortalAPI';
 import {ThreeBounce} from 'better-react-spinkit';
 import exposeComponentRenderer from 'shared/lib/exposeComponentRenderer';
 import TableHeaderControls from "shared/components/tableHeaderControls/TableHeaderControls";
+import LazyMobXTable from "shared/components/lazyMobXTable/LazyMobXTable";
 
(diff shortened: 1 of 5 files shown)


src/pages/datasetView/DatasetList.tsx
```diff
@@ -10,20 +10,24 @@ import TableHeaderControls from "shared/components/tableHeaderControls/TableHead
...
 
-    render(){
+    render() {
         return (
             <span>
-                <a href={`http://www.cbioportal.org/study?id=${ this.props.study.studyId }#summary`}
-                    target='_blank'>{ this.props.study.name }</a>&nbsp;
-                <a href={`https://github.com/cBioPortal/datahub/blob/master/public/${ this.props.study.studyId }.tar.gz`} download><i className='fa fa-download'></i></a>
+                <a
+                    href={`http://www.cbioportal.org/study?id=${ this.props.study.studyId }#summary`}
```
alisman: this should be relative link so we preserve host  e.g. href='study?id=....'

note, we cannot use leading slash because sometimes root is / and sometimes root is /beta/

alisman: i guess it will be easier to understand this when we integrate (which we'll try to do today)

src/pages/datasetView/DatasetList.tsx
```diff
@@ -1,121 +1,141 @@
...
 
-class ReferenceCell extends React.Component<{ study:CancerStudy },{}> {
+class ReferenceCell extends React.Component<IReferenceCellProps ,{}> {
 
-    render(){
+    render() {
         return (
-            <a target='_blank' href={`https://www.ncbi.nlm.nih.gov/pubmed/${ this.props.study.pmid }`}>{ this.props.study.citation }</a>
+            <a target='_blank' href={`https://www.ncbi.nlm.nih.gov/pubmed/${this.props.pmid}`}> {this.props.citation} </a>
         );
```
alisman: should there be a space outside the bracket here?

PR #333, opened 2017-05-02
When we encounter a 401 on ajax call, do redirect to login page

alisman:
When we get a 401 on any request from api, preserve the url hash (accomodates long url) in local storage, redirect to login page with key for url hash, then restore that hash when we get back to page

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/appBootstrapper.jsx
+++ b/src/appBootstrapper.jsx
@@ -1,14 +1,16 @@
 import React from 'react';
 import ReactDOM from 'react-dom';
-import { Provider } from 'mobx-react';
-import { hashHistory, Router } from 'react-router';
-import { RouterStore, syncHistoryWithStore  } from 'mobx-react-router';
+import {Provider} from 'mobx-react';
+import {hashHistory, Router} from 'react-router';
+import {RouterStore, syncHistoryWithStore} from 'mobx-react-router';
 import ExtendedRoutingStore from './shared/lib/ExtendedRouterStore';
-import { computed, extendObservable } from 'mobx';
+import {computed, extendObservable} from 'mobx';
 import makeRoutes from './routes';
 import lodash from 'lodash';
 import $ from 'jquery';
 import URL from 'url';
+import * as superagent from 'superagent';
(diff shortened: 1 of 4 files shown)


src/appBootstrapper.jsx
```diff
@@ -21,6 +23,27 @@ const stores = {
...
+superagent.Request.prototype.end = function (callback) {
+    return end.call(this, (error, response) => {
+        if (redirecting) {
+            return;
+        }
+        if (response.statusCode === 401) {
+            var storageKey = `redirect${Math.floor(Math.random() * 1000000000000)}`
+            localStorage.setItem(storageKey, window.location.hash);
+            const loginUrl = `${getHost()}/?spring-security-redirect=${window.location.pathname}${window.location.search}${encodeURIComponent('#/restore?key=' + storageKey)}`;
+            redirecting = true;
```
alisman: set this flag so that we don't redirect twice due to concurrent calls which both 401.

src/appBootstrapper.jsx
```diff
@@ -21,6 +23,27 @@ const stores = {
...
+
+let redirecting = false;
+
+superagent.Request.prototype.end = function (callback) {
+    return end.call(this, (error, response) => {
+        if (redirecting) {
+            return;
+        }
+        if (response.statusCode === 401) {
+            var storageKey = `redirect${Math.floor(Math.random() * 1000000000000)}`
```
onursumer: why do we need a random storage key here?

alisman: We want to make sure that after redirecting to login page and logging in, we restore the hashfragment that was created when the redirect was initiated, and not some other hashfragment that could possible have gotten stuck in local storage (e.g. failed login attempt).  Then unique key just ensures that will never happen.

alisman: Notice that we pass the key in redirect url.

PR #334, opened 2017-05-02
Add className package and tests to SimpleTable.

Tamba32:
# What? Why?
Add className package for SimpleTable.  Add tests for className prop of SimpleTable.

### DIFF
--- a/src/shared/components/simpleTable/SimpleTable.spec.tsx
+++ b/src/shared/components/simpleTable/SimpleTable.spec.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { assert } from 'chai';
-import { shallow, mount } from 'enzyme';
+import { shallow } from 'enzyme';
 import sinon from 'sinon';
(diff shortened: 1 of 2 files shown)


src/shared/components/simpleTable/SimpleTable.spec.tsx
```diff
@@ -1,35 +1,47 @@
...
+    const headers:any = [<th>one</th>];
+    const rows:any = [<tr><td>data</td></tr>, <tr><td>data</td></tr>];
+    const classNameTrue:string = 'test-class';
+    const classNameFalse:undefined = undefined;
 
-        let headers:any = [<th>one</th>];
-        let rows:any = [<tr><td>data</td></tr>, <tr><td>data</td></tr>];
+    before(()=> {
 
         table = shallow(<SimpleTable headers={headers} rows={rows} />);
```
alisman: you don't need to make any change, but i think we have typings for enzyme, so one would then get autocomplete on the enzyme methods (e.g. find).

PR #330, opened 2017-05-01
Results View Mutation Page & Store Cleanup

onursumer:
- added a new `ResultsViewPage` with a separate `ResultsViewPageStore`: Currently we have a single table with all mutations
- added a new route for results page -- example:  http://localhost:3000/#/query?geneList=BRCA1%20BRCA2&sampleListId=ov_tcga_pub_3way_complete&studyId=ov_tcga_pub
- extracted some of the store's `invoke` functions into `StoreUtils` for reuse purposes
- made results view page specific table columns fully functional

# Checks
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Airbnb React/JSX Style guide](https://github.com/airbnb/javascript/tree/master/react).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)

PR #337, opened 2017-05-03
Handle non-existent genetic profiles for mutations

inodb:
If a mutation profile did not exist, you would get a 404 on patient view. This
instead first checks what genetic profiles are available for the patient.

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -158,13 +158,30 @@ export class PatientViewPageStore {
         return this._sampleId ? 'sample' : 'patient';
     }
 
-    @computed get mutationGeneticProfileId() {
-        return `${this.studyId}${GENETIC_PROFILE_MUTATIONS_SUFFIX}`;
-    }
+    readonly mutationGeneticProfileId = remoteData({
+        await: () => [
+            this.geneticProfilesInStudy
+        ],
+        invoke: async() => {
+            const profile = this.geneticProfilesInStudy.result.find((p: GeneticProfile) => {
+                return p.geneticProfileId === `${this.studyId}${GENETIC_PROFILE_MUTATIONS_SUFFIX}`;
+            });
+            return profile ? profile.geneticProfileId : undefined;
+        }
+    });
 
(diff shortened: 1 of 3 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -492,8 +509,8 @@ export class PatientViewPageStore {
             this.geneticProfilesInStudy
         ],
         invoke: async() => {
-            const profile = this.geneticProfilesInStudy.result.find((profile: GeneticProfile) => {
-                return profile.datatype === 'DISCRETE';
+            const profile = this.geneticProfilesInStudy.result.find((p: GeneticProfile) => {
+                return p.datatype === 'DISCRETE';
```
onursumer: Why do we need to rename `profile` to `p`? :)

inodb: @onursumer typescript gave some error about shadow variables

onursumer: Also, looks like I already moved this function into `StoreUtils` in PR #330. (https://github.com/cBioPortal/cbioportal-frontend/pull/330/files#diff-2a7c36393265d7e597e9d30d3f0314a0R138)

I can handle the conflicts later, it shouldn't be a big deal.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,13 +158,30 @@ export class PatientViewPageStore {
...
 
-    @computed get mutationGeneticProfileId() {
-        return `${this.studyId}${GENETIC_PROFILE_MUTATIONS_SUFFIX}`;
-    }
+    readonly mutationGeneticProfileId = remoteData({
+        await: () => [
+            this.geneticProfilesInStudy
+        ],
+        invoke: async() => {
+            const profile = this.geneticProfilesInStudy.result.find((p: GeneticProfile) => {
```
onursumer: We should probably move this invoke function into `StoreUtils` to reuse within `ResultsPageViewStore` (after merging #330).

PR #341, opened 2017-05-08
expose jquery to page. make github call https always, to fix ie error…

alisman:
- a Use oncokb from portal.properties oncokb.api.url prop, transforming it properly
- b Expose new jquery globally for use by external code
- c always use https to call github search api

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/shared/api/urls.spec.ts
```diff
@@ -0,0 +1,25 @@
...
+    after(()=>{
+        delete (global as any).window.oncoKBApiUrl;
+    });
+
+    it('transforms oncokb url configuration url to proxied url: removes protocol and trailing slash', ()=>{
+        (global as any).window.oncoKBApiUrl = 'http://www.test.com/hello/';
+        // note that this is WRONG (http: should be followed by double shlash)
+        // but this is due to testing env and url builder library
+        // this works correctly in browser env
+        assert.equal(getOncoKbApiUrl(), 'http:proxy/www.test.com/hello');
```
alisman: see comment above this assertion for why it looks wrong.

alisman:
@zhx828 we need to make sure that cbioportal configs actually have the new api url.  also, what about defaults?  we don'e want to default to legacy.  since we don't control configs for all instances, maybe we handle this java?

src/shared/api/urls.ts
```diff
@@ -41,7 +41,17 @@ export function getHotspots3DApiUrl() {
...
-    return cbioUrl('proxy/oncokb.org/api/v1');
+    let url = (window as any).oncoKBApiUrl;
+
+    if (typeof url === 'string') {
+        // we need to support legacy configuration values
+        url = url.replace(/^http[s]?:\/\//,''); // get rid of protocol
+        url = url.replace(/\/$/,""); // get rid of trailing slashes
+        return cbioUrl(`proxy/${url}`)
+    } else {
+        return undefined;
```
alisman: we need to handle this case in views but in fact this will never the undefined for the moment b/c a default is provided by backend

src/pages/patientView/clinicalInformation/PatientViewPageStore.spec.ts
```diff
@@ -22,34 +22,37 @@ describe('PatientViewPageStore', () => {
...
+    //
+    //     result = handlePathologyReportCheckResponse({
+    //         total_count:0,
+    //     });
+    //     assert.deepEqual(result,[]);
+    // });
+    //
+    //
+    //
+    // it('won\'t fetch onkokb data if there are no mutations', ()=>{
```
onursumer: I think this one has already been moved into `StoreUtils.spec`

PR #343, opened 2017-05-10
Add data sets bar graph.

Tamba32:
# What? Why?
Add a bar graph to show study cases by primary site.

### DIFF
--- a/src/shared/components/barGraph/BarGraph.tsx
+++ b/src/shared/components/barGraph/BarGraph.tsx
@@ -1,0 +1,203 @@
+import * as React from "react";
+import * as _ from 'lodash';
+import {ThreeBounce} from 'better-react-spinkit';
+import { CancerStudy } from 'shared/api/generated/CBioPortalAPI';
+import * as Highcharts from 'highcharts';
+
+import HCE from 'highcharts/modules/exporting';
+HCE(Highcharts);
+
+
+export interface IBarGraphProps {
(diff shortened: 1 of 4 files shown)


src/shared/components/barGraph/BarGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+    [name:string]: ICancerTypeStudy;
+};
+
+export default class BarGraph extends React.Component<IBarGraphProps, {colors: string[]}> {
+
+    constructor() {
+        super();
+    }
+
+    getShortName(shortName:string) {
```
alisman: could you use a map instead of a switch statement.  If shortName is a key in the map, then return value, otherwise, just return shortName

src/shared/components/barGraph/BarGraph.tsx
```diff
@@ -0,0 +1,203 @@
...
+        let g = (num & 0x0000FF) + amt;
+
+        if (g > 255) g = 255;
+        else if (g < 0) g = 0;
+
+        return "#" + (g | (b << 8) | (r << 16)).toString(16);
+    }
+
+    componentDidMount() {
+        let cancerTypeStudiesObj = _.reduce(this.props.data, (counterObj:ICounterObj, study) => {
```
alisman: this reduction should be extracted to a pure function (or method) for testing purposes

PR #344, opened 2017-05-10
Add rotating testimonials component.

Tamba32:
# What? Why?
Make React Component rotating testimonials for sidebar.

### DIFF
--- a/src/shared/components/testimonials/Testimonials.tsx
+++ b/src/shared/components/testimonials/Testimonials.tsx
@@ -1,0 +1,109 @@
+import * as React from 'react';
+import CSSTransitionGroup from 'react-transition-group/CSSTransitionGroup';
+
+import {observable, action} from "mobx";
+import {observer} from "mobx-react";
+
+import styles from './testimonials.module.scss';
+
+interface ITestimonial {
+    cite: string;
+    quote: string;
+}
+
+class TestimonialStore {
+    @observable public testimonialIndex: number;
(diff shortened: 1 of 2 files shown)


src/shared/components/testimonials/Testimonials.tsx
```diff
@@ -0,0 +1,113 @@
...
+    quote: string;
+}
+
+class TestimonialStore {
+    @observable public testimonialIndex: number;
+    @observable public length: number;
+    // @observable public activeTestimonial: ITestimonial;
+    @observable public testimonials: ITestimonial[];
+
+    @action setIndex() {
```
alisman: incrementIndex?  nextQuote?

src/shared/components/testimonials/Testimonials.tsx
```diff
@@ -0,0 +1,113 @@
...
+import styles from './testimonials.module.scss';
+
+interface ITestimonial {
+    cite: string;
+    quote: string;
+}
+
+class TestimonialStore {
+    @observable public testimonialIndex: number;
+    @observable public length: number;
```
alisman: I try to would avoid keeping the length (instead of just reading it from array every time).  This is duplicating state.  Reading the length is not significant from performance standpoint.   The exception to this would be when you are optimizing a very large loop that executes sychronously.   There you can get a little benefit from caching the length in a local variable, since you know it won't change during the execution of loop.

PR #346, opened 2017-05-11
Cancer Type column changes

adamabeshouse:
(1) Use CANCER_TYPE_DETAILED clinical attribute
(2) Use cancer type of study if no clinical data for a sample

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1,5 +1,5 @@
 import {
-    Mutation, DiscreteCopyNumberFilter, DiscreteCopyNumberData, MutationFilter, Gene
+    Mutation, DiscreteCopyNumberFilter, DiscreteCopyNumberData, MutationFilter, Gene, CancerStudy
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
 import {computed, observable, action} from "mobx";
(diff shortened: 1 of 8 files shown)


adamabeshouse:
[(3) Results Page: Mutations Tab: Table: Cancer Type Column: use study cancer type if no data for the sample (use CANCER_TYPE_DETAILED clinical attribute)](https://trello.com/c/85lsy6qP/257-3-results-page-mutations-tab-table-cancer-type-column-use-study-cancer-type-if-no-data-for-the-sample-use-cancer-type-detailed-c)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -102,6 +102,17 @@ export class ResultsViewPageStore {
         invoke: async () => fetchSamples(this.sampleIds, this.studyId)
     }, []);
 
+    readonly studies = remoteData({
+        invoke: async()=>([await client.getStudyUsingGET({studyId: this.studyId})])
+    }, []);
+
+    @computed get studyToCancerType() {
+        return this.studies.result.reduce((map:{[studyId:string]:string}, next:CancerStudy)=>{
```
onursumer: Although this is a small function, it might still be better to move this into `StoreUtils` for testing purposes. Well, currently most of the `StoreUtils` functions are untested, but maybe one day we add more tests :)

onursumer:
After merging this one and #336, we will end up with a quite long cancer type column definition in `MutationTable`. We may need another PR which moves most of that code into `CancerTypeColumnFormatter`.

adamabeshouse:
@onursumer I think it will be easiest as far as merging to do that in another PR

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: done

PR #349, opened 2017-05-15
Use entrezGeneId instead of hugoSymbol when querying OncoKB service

zhx828:
# Checks
  the `-s` option)

# Notify reviewers
@onursumer

### DIFF
--- a/src/shared/lib/OncoKbUtils.ts
+++ b/src/shared/lib/OncoKbUtils.ts
@@ -107,7 +107,7 @@ export function generateEvidenceQuery(queryVariants:Query[]): EvidenceQueries
     };
 }
 
-export function generateQueryVariant(hugoSymbol:string,
+export function generateQueryVariant(entrezGeneId:number,
                                      tumorType:string,
                                      alteration?:string,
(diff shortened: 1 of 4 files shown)


src/shared/lib/OncoKbUtils.ts
```diff
@@ -116,11 +116,11 @@ export function generateQueryVariant(hugoSymbol:string,
                                      alterationType?:string): Query
 {
     return {
-        id: generateQueryVariantId(hugoSymbol, tumorType, alteration, mutationType),
-        hugoSymbol,
+        id: generateQueryVariantId(entrezGeneId, tumorType, alteration, mutationType),
+        hugoSymbol: '',
```
inodb: What's this empty `hugoSymbol` string for?

zhx828: In this particular case, we are sending request body with POST method which is using `Query` model. In typescript, we need to define a value for each property within that model(this is what I understand. Please correct me if I'm wrong @onursumer ). Similar to what we have currently that we always set entrezGeneId to 0.

PR #351, opened 2017-05-16
Improve cache by supporting metadata

adamabeshouse:
Use case: info that you want to use for indexing fetched data is available at fetch time, but not actually in the data itself (e.g. you fetch by study, but then the data doesn't echo 'studyId'). So this allows the fetcher to associate such metadata with the data which can then be used for indexing.

This PR then uses this capability to reimplement PmidCache

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/cache/PmidCache.ts
+++ b/src/shared/cache/PmidCache.ts
@@ -1,53 +1,42 @@
-import {default as SimpleCache, ICache} from "shared/lib/SimpleCache";
+import LazyMobXCache from "../lib/LazyMobXCache";
+import {AugmentedData} from "../lib/LazyMobXCache";
+import request from "superagent";
 
-export default class PmidCache extends SimpleCache<any, number[]>
+async function fetch(pmids: number[])
+{
+        const pmidData:any = await new Promise((resolve, reject) => {
+            // TODO better to separate this call to a configurable client
+            request.post('https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json')
+                .type("form")
+                .send({id: pmids.join(',')})
+                .end((err, res)=>{
+                    if (!err && res.ok) {
+                        resolve(JSON.parse(res.text));
+                    } else {
+                        reject(err);
+                    }
+                });
+        });
+
+        const ret:AugmentedData<any, string>[] = [];
+        for (const pmidStr of Object.keys(pmidData)) {
+            ret.push({
+                data: pmidData[pmidStr],
+                meta: pmidStr
+            });
+        }
+
+        return ret;
+}
+
+export default class PmidCache extends LazyMobXCache<number, any, string>
 {
     constructor()
     {
-        super();
-    }
-
-    protected async fetch(pmids: number[])
-    {
-        const cache: ICache<any> = {};
-
-        try {
-            const pmidData:any = await new Promise((resolve, reject) => {
-                // TODO better to separate this call to a configurable client
-                $.post(
-                    'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json',
-                    {id: pmids.join(',')}
-                ).then((data) => {
-                    resolve(data);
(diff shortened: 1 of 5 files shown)


adamabeshouse:
[update LazyMobXCache to allow indexing by fetch-time-known data](https://trello.com/c/5ewbKD8v/274-update-lazymobxcache-to-allow-indexing-by-fetch-time-known-data)

src/pages/patientView/clinicalInformation/CopyNumberCountCache.ts
```diff
@@ -21,7 +21,7 @@ function fetch(queries:CopyNumberCountIdentifier[], geneticProfileIdDiscrete:str
     }
 }
 
-export default class CopyNumberCountCache extends LazyMobXCache<CopyNumberCount, CopyNumberCountIdentifier> {
+export default class CopyNumberCountCache extends LazyMobXCache<CopyNumberCount, CopyNumberCountIdentifier, any> {
```
alisman: it seems like in all the uses of the cache, the type of Meta is any, so why does it need to be specified? (apologies if i'm missing the exception!)

src/shared/cache/PmidCache.ts
```diff
@@ -1,53 +1,39 @@
...
-                    status: "error"
-                };
-            });
 
-            this.putData(cache);
-        }
+        return ret;
+}
+
+export default class PmidCache extends LazyMobXCache<number, any, string>
```
alisman: ah!  the exception!

src/shared/cache/PmidCache.ts
```diff
@@ -1,53 +1,39 @@
...
-                // TODO better to separate this call to a configurable client
-                $.post(
-                    'https://eutils.ncbi.nlm.nih.gov/entrez/eutils/esummary.fcgi?db=pubmed&retmode=json',
-                    {id: pmids.join(',')}
-                ).then((data) => {
-                    resolve(data);
-                }).fail((err) => {
-                    reject(err);
-                });
+        const pmidData:any = await new Promise((resolve, reject) => {
```
alisman: lets always use superagent library b/c we get benefit of error handling and Sentry tracking out of box

adamabeshouse: sounds good

src/shared/cache/PmidCache.ts
adamabeshouse: yes, that's the whole point of this pull request :) in typescript 2.3 we can have default generic types but not in typescript 2. Maybe it's time to upgrade to 2.3 for this PR? because that really is quite inconvenient

alisman: yes, lets upgrade

alisman:
so we're not gonna merge this right?  (update ts instead?)

adamabeshouse:
no theres actual code in this PR too

src/shared/lib/LazyMobXCache.ts
```diff
@@ -135,17 +147,30 @@ export default class LazyMobXCache<Data, Query> {
...
         const keyHasData:{[key:string]:boolean} = {};
 
-        for (const datum of data) {
-            const datumKey = this.dataToKey(datum);
-            toMerge[datumKey] = {
-                status: "complete",
-                data: datum
-            };
-            keyHasData[datumKey] = true;
+        for (const dataElt of data) {
```
alisman: this could use some explanatory comment.  if i understand correctly, the presence of meta means that dataElt.data is a collection whose members need to be merged (and given ref to meta) wheareas without meta, only the parent would be merged?

adamabeshouse: right, the example case is if you want to augment data with the study that it came from, if that study id is not echoed in the API data. its cheaper to wrap a list of data that all have the same metadata (study id), rather than wrap each individual datum. so all the data in that AugmentedData gets associated with that metadata via it being added to the CacheData and it being passed into dataToKey to find the key

adamabeshouse: do you know what comment would make that clearer?

PR #353, opened 2017-05-17
Add browser error modal

Tamba32:
# What? Why?

Add browser/browser version detection so error message can be provided if version of IE used is earlier than IE11.


src/appShell/App/Container.tsx
```diff
@@ -10,31 +11,86 @@ interface IContainerProps {
...
+        this.handleCloseClick = this.handleCloseClick.bind(this);
+    }
 
     static contextTypes = {
         router: React.PropTypes.object,
     };
 
     context: {router: any};
 
+    findIEVersion(userAgent:string, versionIndex: number, localStorage:boolean = false) {
```
alisman: i think this method should have a more general name.  like, handleUnsupportedBrowsers.    also, since you are using localStorage within this method, you might as well internalize the check for browser setting, instead of passing it in.  in other words, this method is NOT pure in the functional-programming sense, so you might as well enjoy it, so to speak.  let me know if that needs more explanation.

src/shared/components/broswerErrorModal/BrowserErrorModal.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
```
alisman: I would give this file a different name.  UnsupportedBrowserModal.

src/appShell/App/Container.tsx
alisman: Also, it would be nice to package this method with the modal and not pollute the Container with this logic.   So, just always include the UnsupportedBrowserModal and it either renders something or not.  It manages its own state.

src/appShell/App/Container.spec.tsx
```diff
@@ -0,0 +1,87 @@
...
+
+
+    it('shows modal when IE10 is being used and site is accessed for first time', () => {
+        wrapper.instance().findIEVersion(ie10, ie10.indexOf("MSIE"));
+        expect(wrapper.contains(<BrowserErrorModal show={true} ieVersion='10' onHide={onHide}/> )).to.equal(true);
+    });
+
+
+    it.only('does not show modal when site is accessed after the first time, regardless of browser and/or version', () => {
+        wrapper.instance().findIEVersion(ie10, ie10.indexOf("MSIE"), true);
```
alisman: i would use a library to handle browser detection. https://github.com/DamonOehlman/detect-browser.

src/appShell/App/Container.tsx
alisman: also, we should use a library to do browser/version detection.https://github.com/DamonOehlman/detect-browser

alisman:
@inodb can you take take chage of getting this into rc.  testing it. 
@Tamba32 needs to be rebased on latest rc

PR #359, opened 2017-05-23
Add extra visual indicators for uncalled mutations

inodb:
- On mouseover of label in mutation table indicate the mutation wasn't called

screen shot 2017-05-23 at 3 28 11 pm

- In allele freq mouseover indicate which mutations were not called
screen shot 2017-05-24 at 6 36 23 pm

### DIFF
--- a/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
+++ b/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
@@ -4,6 +4,7 @@ import DefaultTooltip from "shared/components/DefaultTooltip";
 import 'rc-tooltip/assets/bootstrap_white.css';
 import {Mutation} from "../../../../shared/api/generated/CBioPortalAPI";
 import SampleManager from "../../sampleManager";
+import {isUncalled} from '../../../../shared/lib/mutationUtils';
 
 export default class AlleleFreqColumnFormatter {
     static barWidth = 6;
@@ -11,14 +12,27 @@ export default class AlleleFreqColumnFormatter {
     static maxBarHeight = 12;
     static indexToBarLeft = (n:number) => n*(AlleleFreqColumnFormatter.barWidth + AlleleFreqColumnFormatter.barSpacing);
 
-    public static renderFunction(mutations:Mutation[], sampleManager:SampleManager|null) {
-        if (!sampleManager) {
-            return (<span></span>);
-        }
+    public static getComponentForSampleArgs<T extends {tumorAltCount:number,geneticProfileId:string}>(mutation:T) {
+        const altReads = mutation.tumorAltCount;
 
-        const sampleOrder = sampleManager.getSampleIdsInOrder();
-        const barX = sampleOrder.reduce((map:{[s:string]:number}, sampleId:string, i:number) => {map[sampleId] = AlleleFreqColumnFormatter.indexToBarLeft(i); return map;}, {});
-        const sampleElements = mutations.map(function(mutation:Mutation) {
+        let opacity: number = 1;
+        let extraTooltipText: string = '';
+        if (isUncalled(mutation.geneticProfileId)) {
+            if (altReads > 0) {
+                opacity = 0.1;
+                extraTooltipText = "Mutation has supporting reads, but wasn't called";
+            } else {
+                opacity = 0;
+                extraTooltipText = "Mutation has 0 supporting reads and wasn't called";
+            }
+        }
+        return {
+           opacity,
+           extraTooltipText
+        };
+    }
+
+    public static convertMutationToSampleElement<T extends {sampleId:string, tumorRefCount:number, tumorAltCount:number, geneticProfileId:string}>(mutation:T, color:string, barX:number, sampleComponent:any) {
             const altReads = mutation.tumorAltCount;
             const refReads = mutation.tumorRefCount;
             if ((altReads < 0) || (refReads < 0)) {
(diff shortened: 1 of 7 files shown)


inodb:
[Add extra visual indicators for uncalled mutations](https://trello.com/c/Qmwk0NEj/287-add-extra-visual-indicators-for-uncalled-mutations)

src/pages/patientView/sampleManager.tsx
```diff
@@ -180,10 +180,23 @@ class SampleManager {
...
             <div style={{ maxHeight:400, overflow:'auto' }}>
-                <h5>{ sample.id }</h5>
+                <h5 style={{ marginBottom: 1 }}>
+                    <svg height="12" width="12" style={{ marginRight: 5}}>
+                        <SampleInline
+                                     sample={sample}
+                                     sampleNumber={sampleNumber}
+                                     sampleColor={sampleColor}
+                                     fillOpacity={fillOpacity}
+                                 >
```
onursumer: Indentation looks off here. Also, lint doesn't like empty bodies so it might be better to do 
```
<SampleInline
    sample={sample}
    ...
/>
```

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -31,12 +32,30 @@ export default class AlleleFreqColumnFormatter {
 
             const bar = (<rect x={barX[mutation.sampleId]} y={barY} width={AlleleFreqColumnFormatter.barWidth} height={barHeight} fill={sampleManager.getColorForSample(mutation.sampleId)}/>);
 
-            const circleRadius = 6;
```
onursumer: was this one an unused variable?

inodb: yup

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -31,12 +32,30 @@ export default class AlleleFreqColumnFormatter {
...
             const sampleId = mutation.sampleId;
-            const component = sampleManager.getComponentForSample(sampleId);
+            const isUncalled = mutation.geneticProfileId.endsWith(GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX);
```
alisman: I don't think endsWith is supported in IE or safari < 9.    i would build a regexp to do this.  
new RegExp(`${GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX}$`);

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -31,12 +32,30 @@ export default class AlleleFreqColumnFormatter {
...
+            let opacity: number = 1;
+            let extraTooltipText: string = '';
+            if (isUncalled) {
+                if (altReads > 0) {
+                    opacity = 0.1;
+                    extraTooltipText = "Mutation has supporting reads, but wasn't called";
```
alisman: we should write some tests for this code.  let me know if you want some help with that.

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -31,12 +32,30 @@ export default class AlleleFreqColumnFormatter {
...
+                    extraTooltipText = "Mutation has 0 supporting reads and wasn't called";
+                }
+            }
+            const component = sampleManager.getComponentForSample(sampleId, opacity, extraTooltipText);
+
+            let variantReadText:string;
+            if (isUncalled && altReads > 0) {
+                variantReadText = `(${altReads} variant reads out of ${altReads+refReads} total, but uncalled)`;
+            } else {
+                variantReadText = `(${altReads} variant reads out of ${altReads+refReads} total)`;
```
adamabeshouse: dry this up
```
`(${altReads} variant reads out of ${altReads+refReads} total${(isUncalled && altReads) ? ", but uncalled" : ""})`
```

adamabeshouse: it's sort of hard to read that at the end, what about putting it in front instead?
like
`(uncalled) (.. variant reads out of .. total)`

inodb: Good point, both changed as suggested. Slightly changed the workings, so it always shows `(uncalled)` if the mutation wasn't called. Instead of just indicating that for the ones with supporting reads. Given that the sample labels now also change colors it's pretty obvious what the difference is between the two cases imo: not called but with supporting reads and not called with supporting reads.

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
inodb: Let's discuss what kind of tests we want

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
adamabeshouse: why do you include a line for the sample that its uncalled without any supporting reads?

src/shared/lib/mutationUtils.ts
```diff
@@ -0,0 +1,5 @@
+import {GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX} from "../constants";
+
+export function isUncalled<T extends {geneticProfileId?:string}>(mutation:T) {
```
alisman: I think the appropriate type here is actually mutation:any, because all the T here does is say the arg may or may not have a geneticProfileId, which could be anything.   Alternatively, we could just have this method accept a string.

src/shared/lib/mutationUtils.ts
```diff
@@ -0,0 +1,5 @@
...
+    return Boolean(mutation.geneticProfileId && mutation.geneticProfileId.match(new RegExp(GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX + "$")));
```
alisman: this should be new RegExp(...).test(mutation.geneticProfileId).  the test method of RegExp returns boolean and handles undefined arg.

src/shared/lib/mutationUtils.ts
adamabeshouse: agreed w @alisman, I'd say either make it nonoptional or make it accept the genetic profile id directly

inodb: Good points, changed to make it accept genetic profile id directly

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
inodb: Added a few tests

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
inodb: As mentioned on the chat, it's to show that position was profiled. Otherwise you don't know whether or not that position was actually checked for read counts.

PR #364, opened 2017-05-25
Added store and utility functions to fetch and process PDB alignment data

onursumer:
Need to add more tests, especially for `PdbUtils`.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapperStore.ts
+++ b/src/pages/resultsView/mutation/MutationMapperStore.ts
@@ -7,10 +7,12 @@ import {remoteData} from "shared/api/remoteData";
 import {labelMobxPromises, MobxPromise, cached} from "mobxpromise";
 import {IOncoKbData} from "shared/model/OncoKB";
 import {IHotspotData} from "shared/model/CancerHotspots";
+import {IPdbChain} from "shared/model/Pdb";
+import {calcPdbIdNumericalValue} from "shared/lib/PdbUtils";
 import {
     indexHotspotData, fetchHotspotsData, fetchCosmicData, fetchOncoKbData,
     fetchMutationData, generateSampleIdToTumorTypeMap, generateDataQueryFilter,
-    ONCOKB_DEFAULT
+    ONCOKB_DEFAULT, fetchPdbAlignmentData, mergePdbAlignments, fetchSwissProtAccession, fetchUniprotId
 } from "shared/lib/StoreUtils";
 import {IMobXApplicationDataStore, SimpleMobXApplicationDataStore} from "../../../shared/lib/IMobXApplicationDataStore";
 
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -120,4 +177,21 @@ export class MutationMapperStore {
...
+
+        // define sort metric and direction at initialization time
+        dataStore.sortAscending = false;
+        dataStore.sortMetric = (pdbChain: IPdbChain) => [
+            pdbChain.identity,         // first, sort by identity
+            pdbChain.alignment.length, // then by alignment length
+            pdbChain.identityPerc,     // then by identity percentage
+            // current sort metric cannot handle mixed values so generating numerical values for strings
+            ...calcPdbIdNumericalValue(pdbChain.pdbId, true), // then by pdb id (A-Z): always returns an array of size 4
+            -1 * pdbChain.chain.charCodeAt(0)                 // then by chain id (A-Z): chain id is always one char
```
onursumer: I am not 100% satisfied with this solution...

@adamabeshouse, do we have a better way to solve the mixed type of sort metrics?

src/shared/components/structureViewer/StructureViewerPanel.tsx
```diff
@@ -520,8 +546,98 @@ export default class StructureViewerPanel extends React.Component<IStructureView
...
+            return undefined;
+        }
+    }
+
+    @computed get pdbChain()
+    {
+        let data = this.props.pdbChainDataStore.sortedFilteredSelectedData;
+
+        if (data.length === 0) {
+            data = this.props.pdbChainDataStore.sortedData;
```
onursumer: Here, I am assuming that we will only sort the chain data once. But if we make the PDB chain table sortable, then it makes things complicated. We don't want the order of chains in the PDB panel to change. Maybe we just don't pass this store to the table component?

On the other hand, maybe it is better to assume that there will always be selected data, and it will always contain one chain and not even worry about the sorted data?

src/shared/model/Pdb.ts
```diff
@@ -7,3 +7,18 @@ export interface IPdbPositionRange {
     start: IPdbPosition;
     end: IPdbPosition;
 }
+
+export interface IPdbChain {
```
onursumer: Actually, maybe rename this to `IPdbChainAlignment` to be more precise.

onursumer:
[(13) Results Page: Mutations Tab: 3D Alignment](https://trello.com/c/hoDzTWmH/227-13-results-page-mutations-tab-3d-alignment)

src/shared/components/structureViewer/StructureViewerPanel.tsx
```diff
@@ -520,8 +546,98 @@ export default class StructureViewerPanel extends React.Component<IStructureView
...
+                        position: 1815
+                    },
+                    end: {
+                        position: 1815
+                    }
+                },
+                color: "#00FF00",
+                highlighted: true
+            }
+        ];
```
onursumer: This will eventually be replaced with the actual mutation to residue mapping.

PR #367, opened 2017-05-29
Implement CIViC in the patient view

oplantalech:
# What? Why?
Implements CIViC in the new patient view (implementation in the old patient view: cBioPortal/cbioportal#2067).


Sjoerd-van-Hagen:
@jjgao @ecerami how would be the best person to review this PR? It is the porting of Civic to the new patient view!

jjgao:
@oplantalech I've changed the base branch to `rc`. Could you please rebase?

@onursumer @zhx828 Would you please review this pull request?

jjgao:
@alisman I assume the branch structure is the same so we should use rc?

oplantalech:
@jjgao Rebased.

src/shared/components/annotation/Civic.tsx
```diff
@@ -0,0 +1,117 @@
...
+        return (
+            <Circle size={18} scaleEnd={0.5} scaleStart={0.2} color="#aaa" className="pull-left"/>
+        );
+    }
+
+    private cardContent(civicInstance: ICivicInstance): JSX.Element
+    {
+        return (
+            <CivicCard
+                    title={`CIViC Variants`}
```
onursumer: Indentation is off here, should be one less tab for each line.

src/shared/components/annotation/styles/civicCard.scss
```diff
@@ -0,0 +1,172 @@
+.civic-card {
```
onursumer: This almost looks like `oncoKbCard.scss` (which is mostly legacy code that needs to be cleaned up). I don't think we really need to duplicate everything for Civic. For example there are styles for `qtip` in this file that we don't really use anymore. It would be nicer to copy only the related styles, not the entire `oncoKbCard.scss`.

src/shared/api/CivicAPI.ts
```diff
@@ -0,0 +1,265 @@
...
+            variantMap[variant.name] = variant.id;
+        });
+    }
+    return variantMap;
+};
+
+/**
+ * CIViC
+ */
+export function CivicAPI() {
```
onursumer: Why not a class?

src/shared/api/CivicAPI.ts
```diff
@@ -0,0 +1,265 @@
+import $ from 'jquery';
```
onursumer: Actually the content of this file looks like CivicAPI + CivicUtils to me. I mean it both contains the data retrieval functions (API) and data processing functions (Utils). Might be better to separate them.

src/shared/api/CivicAPI.ts
```diff
@@ -0,0 +1,265 @@
...
+            return mainPromise.then(function() {
+                return civicData;
+            });
+        },
+        
+        /**
+         * Asynchronously retrieve a map with Civic information from the mutationSpecs given for all genes in civicData.
+         * If no mutationSpecs are given, then return the Civic information of all the CNA variants of the genes in civicData.
+         */
+        getCivicVariants: function(civicData: ICivicData, mutationSpecs?: Array<{gene:{hugoGeneSymbol: string}, proteinChange: string}>): JQueryPromise<ICivicVariant> {
```
onursumer: ` Array<{gene:{hugoGeneSymbol: string}, proteinChange: string}>` is a bit confusing. We are actually passing Mutation instances to this function. How about something like this:
```JavaScript
function getCivicVariants<MutationSpec extends {gene:{hugoGeneSymbol: string}, proteinChange: string}>(..., mutationSpecs?: MutationSpecs[], ...)
```

That would allow passing a `Mutation` instance to this function in a clearer way.

src/shared/components/annotation/CivicCard.tsx
```diff
@@ -0,0 +1,111 @@
...
+    public variantItem(url?:string, name?:string, entryTypes?:string, description?:string)
+    {
+        let result;
+        
+        if (url || name || entryTypes || description) {
+            result = (
+            <div className="civic-card-variant">
+                <div className="civic-card-variant-header">
+                    <span className="civic-card-variant-name"><a href={url} target="_blank">{name}</a></span>
+                    <span className="civic-card-variant-entry-types">{" Entries: "+entryTypes}</span>
```
onursumer: Here, string template literal is more preferable to string concatenation: `Entries: ${entryTypes}`

src/shared/lib/StoreUtils.ts
```diff
@@ -314,6 +316,56 @@ export async function queryOncoKbData(queryVariants: Query[],
...
+        return civicData;
+    } else {
+        return {};
+    }
+}
+
+export async function fetchCivicVariants(civicData: ICivicData, mutationData?:MobxPromise<Mutation[]>,
+                                         uncalledMutationData?:MobxPromise<Mutation[]>) {
+
+    //For some reason, Typescript indicates that getCivicVariants can be undefined: it can't
```
onursumer: It might be because `CivicAPI` is a function but not a class. Just a guess though.

src/shared/model/Civic.ts
```diff
@@ -0,0 +1,27 @@
...
+    description: string;
+    url: string;
+    evidence: {[evidenceType: string]: number};
+};
+
+export interface ICivicData {[name: string]: ICivicGeneData};
+
+export interface ICivicVariant {[id: string]: {[name: string]: ICivicGeneVariant}};
+
+export interface ICivicInstance {
```
onursumer: A class/interface name ending with `instance` feels a bit counter-intuitive.

Say an object is an instance of `ICivicInstance`, should we call that object a `ICivicInstance` instance in that case? :)

src/shared/api/CivicAPI.ts
onursumer: Also, do we really need `jQuery` here? Maybe we can use `superagent` instead? Or even better, we can auto-generate the client API if there is a swagger definition for that.

zhx828:
@oplantalech how do we disable CIViC?

oplantalech:
@zhx828 To disable CIViC, you can set the parameter `enableCivic` to `false` in `MutationTable.tsx` and `CopyNumberTableWrapper.tsx`. There are also similar parameters for OncoKB, MyCancerGenome and Hotspots.

src/shared/api/CivicAPI.ts
oplantalech: There is no swagger definition for the Civic API. Is it that important to use `superagent` instead of `jQuery`?

src/shared/lib/StoreUtils.ts
oplantalech: I transformed `CivicAPI` into a class but the problem is still there.

src/shared/api/CivicAPI.ts
oplantalech: I just created a new type called `MutationSpecs`, I think that was what you wanted here.

src/shared/model/Civic.ts
oplantalech: I revised all the interface names and changed some of them to be more intuitive, also for some variable names that were related to those interfaces.

oplantalech:
@onursumer @zhx828 I have improved the PR according to your suggestions. I also created a few tests.

zhx828:
@oplantalech the feature should be disabled if the `show.civic` set to false in the property file. In line with other pages.

src/shared/api/CivicAPI.ts
onursumer: Ideally what I was thinking is something more generic:
```function getCivicVariants<T extends MutationSpec>(civicGenes: ICivicGene, mutationSpecs?: Array<T>)```

That would allow passing anything that extends `MutationSpec`, and that works with `Mutation` instances. Note that `Mutation` does not explicitly extend `MutationSpec`, but TypeScript is smart enough to consider that `Mutation` implicitly extends `MutationSpec`, because it contains all the fields defined in `MutationSpec`.

An example implementation: https://github.com/cBioPortal/cbioportal-frontend/blob/5fa910700d25a8c5ee307ac91d612f7fd50b0bde/src/pages/patientView/clinicalInformation/CopyNumberCountCache.ts#L5

File: src/shared/api/CivicAPI.ts (hunk 4 above)
Comment (onursumer) [inline src/shared/api/CivicAPI.ts]: @alisman might be a better person to answer this question

**Comment by oplantalech** [comment]
@zhx828 I've rebased again to `rc` and I've finished and tested the CIViC enabling/disabling option that @onursumer started in #380. So now you should be able to disable CIViC. As far as I know, the only thing that needs to be fixed is to change the API requests from `jQuery` to `superagent`, I'm in contact with @alisman to change that. If you guys see something else in this PR that needs to change, please comment it.

File: src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx [hunk 9]
```diff
@@ -6,30 +6,37 @@ import OncokbPubMedCache from "shared/cache/PubMedCache";
...
+    civicVariants?: ICivicVariant;
 }
 
 export interface IAnnotation {
     isHotspot: boolean;
     is3dHotspot: boolean;
     myCancerGenomeLinks: string[];
     oncoKbIndicator?: IndicatorQueryResp;
+    civicEntry?: ICivicEntry | null;
+    isCivicDisabled: boolean;
```
onursumer: `isCivicDisabled` flag should not be a part of the `IAnnotationModel`. We already have `enableCivic` option in the `IAnnotationColumnProps`. You can either use `columnProps.enableCivic` directly, or set `civicEntry` to `undefined` when `enableCivic` is false.

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
```diff
@@ -6,30 +6,37 @@ import OncokbPubMedCache from "shared/cache/PubMedCache";
...
+    civicGenes?: ICivicGene;
+    civicVariants?: ICivicVariant;
 }
 
 export interface IAnnotation {
     isHotspot: boolean;
     is3dHotspot: boolean;
     myCancerGenomeLinks: string[];
     oncoKbIndicator?: IndicatorQueryResp;
+    civicEntry?: ICivicEntry | null;
```
onursumer: I think you can just define this as `civicEntry?: ICivicEntry`. `null` seems a bit redundant to me.

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
```diff
@@ -185,6 +224,12 @@ export default class AnnotationColumnFormatter
                         is3dHotspot={annotation.is3dHotspot}
                     />
                 </If>
+                <If condition={columnProps.enableCivic || false}>
+                    <Civic
+                        civicEntry={annotation.civicEntry}
+                        isCivicDisabled={annotation.isCivicDisabled}
```
onursumer: To be consistent with the rest of the implementation, I think we should not render the Civic component at all if it is disabled. As far as I understand, current implementation shows a "disabled icon" in case civic is disabled. We prefer to show nothing in that case. It makes the implementation simpler.

src/shared/components/annotation/Civic.tsx
```diff
@@ -0,0 +1,117 @@
...
+import {Circle} from "better-react-spinkit";
+import DefaultTooltip from 'shared/components/DefaultTooltip';
+import annotationStyles from "./styles/annotation.module.scss";
+import {ICivicVariant, ICivicEntry} from "shared/model/Civic.ts";
+import {observable} from "mobx";
+import CivicCard from "./CivicCard";
+
+export interface ICivicProps { 
+    civicEntry: ICivicEntry | null | undefined;
+    isCivicDisabled: boolean;
```
onursumer: Let's remove this property (and the related code segment) from the component, and assume that if it is rendered, it is enabled.

onursumer:
@oplantalech Seems like `CivicMockUtils` is missing in the PR, so the tests are failing on the travis build. Can you add that file to the PR?

I think `src/test` directory is unintentionally excluded by `git.ignore`, so you may need to do `git add -f` to include that file. Ideally, we should fix the `git.ignore` file as well.

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
oplantalech: This is a bit tricky. This flag do not refer to enabling Civic but it defines if the icon should be the color one or the black and white one. The icon in black and white was defined as the 'disabled icon'. I have to think of another name.

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
oplantalech: Again, this name is confusing. The disabled icon only works on CNA when you have information about the gene but not about CNA variants, so when you hover over the gene you see the description of the gene.

oplantalech:
@onursumer I added `CivicMockUtils`, I renamed `isCivicDisabled` to avoid confusions, I added more tests and I transformed the `jQuery` calls into `superagent`. I think this is all!

onursumer:
Thanks @oplantalech, this looks good!

Just one more thing: can you rebase and squash commits into a single commit, once you are done with the tests?

oplantalech:
@onursumer Can you approve the PR if you do not have any other suggestions?

onursumer:
@oplantalech Can we disable the loader indicator (spinner) for Civic? Right now, the only annotation that has a loader indicator is OncoKB. Others just show up when the loading is finished.

When more than one annotation is loading at the same time (please see the screenshot), it doesn't look so good. @zhx828, what do you think?

selection_067

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -32,13 +41,50 @@ export default class AnnotationColumnFormatter
...
+        let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
+        let civicEntry = null;
+        //Only return data for genes with variants or it has a description provided by the Civic API
+        if (geneVariants ||
+               (geneEntry && geneEntry.description != "")) {
+            civicEntry = {name: geneEntry.name,
+                             description: geneEntry.description,
+                             url: geneEntry.url,
+                             variants: geneVariants,
+                            }
```
onursumer: A minor thing, but can we fix the indentation here? Something like this would look better:

```
civicEntry = {name: geneEntry.name,
    description: geneEntry.description,
    url: geneEntry.url,
    variants: geneVariants,
};
```

Also, can we add a semicolon at the end of the final line? BTW, it is always useful to run "npm run lint" and review the warnings for the new files/lines introduced. For example, there is a warning for missing semicolons. (You don't need to fix every lint error though)

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -32,13 +41,50 @@ export default class AnnotationColumnFormatter
...
+
+    public static hasCivicVariants (copyNumberData:DiscreteCopyNumberData[], civicGenes:ICivicGene, civicVariants:ICivicVariant): boolean
+    {
+        let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
+        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
+    
+        if (geneEntry && !geneVariants) {
+            return false;
+            }
```
onursumer: Can we fix the indentation here as well?

src/shared/api/CivicAPI.ts
```diff
@@ -0,0 +1,106 @@
...
+                   return {
+                       id: variantId,
+                       name: name,
+                       geneId: geneId,
+                       description: result.description,
+                       url: 'https://civic.genome.wustl.edu/#/events/genes/' + geneId +
+                            '/summary/variants/' + variantId + '/summary#variant',
+                       evidence: countEvidenceTypes(result.evidence_items)
+                   };
+              })
```
onursumer: missing semicolon...

src/shared/components/annotation/Civic.tsx
```diff
@@ -0,0 +1,117 @@
...
+            {
+                civicContent = (
+                    <span className={`${annotationStyles["annotation-item"]}`}>
+                            <img
+                                width={civicImgWidth}
+                                height={civicImgHeight}
+                                src={civicImgSrc}
+                                alt='Civic Variant Entry'
+                            />
+                        </span>
```
onursumer: indentation should be one tab less here and for the entire `img` component as well...

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
```diff
@@ -62,13 +74,35 @@ export default class AnnotationColumnFormatter
...
             };
         }
 
         return value;
     }
 
+    /**
+     * Returns an ICivicEntry if the civicGenes and civicVariants have information about the gene and the mutation (variant) specified. Otherwise it returns null.
+     */
+    public static getCivicEntry(mutation:Mutation, civicGenes:ICivicGene, civicVariants:ICivicVariant): ICivicEntry | null
```
onursumer: This method is the duplicate of the `getCivicEntry` defined in `copyNumberAlterations/column/AnnotationColumnFormatter` (except one line). Is it possible to separate the duplicated part as a shared utility function? Probably `CivicUtils` file is the best place to put it in.

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -32,13 +41,50 @@ export default class AnnotationColumnFormatter
...
         }
 
         return value;
     }
 
+   /**
+     * Returns an ICivicEntry if the civicGenes and civicVariants have information about the gene and the mutation (variant) specified. Otherwise it returns 
+     * an empty object.
+     */
+    public static getCivicEntry(copyNumberData:DiscreteCopyNumberData[], civicGenes:ICivicGene, civicVariants:ICivicVariant): ICivicEntry | null
```
onursumer: See my other comment in the `mutationTable/column/AnnotationColumnFormatter`related to duplication.

zhx828:
@onursumer @oplantalech Let's keep multiple loading bars for now. We are using the same logic in other pages. But we should definitely find a better solution if lateR we integrate more resources(like MatchMiner, MM or JAX)

oplantalech:
@onursumer Does it look better now? I don't know if you want to extract even more the `getCivicEntry` function...

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -50,23 +51,18 @@ export default class AnnotationColumnFormatter
...
         let civicEntry = null;
         //Only return data for genes with variants or it has a description provided by the Civic API
-        if (geneVariants ||
-               (geneEntry && geneEntry.description != "")) {
-            civicEntry = {name: geneEntry.name,
-                             description: geneEntry.description,
-                             url: geneEntry.url,
-                             variants: geneVariants,
-                            }
+        if (geneVariants || (geneEntry && geneEntry.description != "")) {
```
onursumer: Why do we need an extra `geneEntry` check here, but not within the other `getCivicEntry` function (the one in `mutationTable/column/AnnotationColumnFormatter`)?

If we can use the same condition for both functions (i.e: if we can add the `geneEnrty` check for the other function as well), then we can move even more functionality into the `buildCivicEntry` function. Well, basically everything except the first line where we get the gene symbol.

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -32,13 +42,45 @@ export default class AnnotationColumnFormatter
...
+    * an empty object.
+    */
+    public static getCivicEntry(copyNumberData:DiscreteCopyNumberData[], civicGenes:ICivicGene, civicVariants:ICivicVariant): ICivicEntry | null
+    {
+        let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
+        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
+        let civicEntry = null;
+        //Only return data for genes with variants or it has a description provided by the Civic API
+        if (geneVariants || (geneEntry && geneEntry.description != "")) {
```
alisman: try to always use double equal !== 
just a good standard practice to avoid type coercion.  our linting will catch this soon

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -32,13 +42,45 @@ export default class AnnotationColumnFormatter
...
+    public static hasCivicVariants (copyNumberData:DiscreteCopyNumberData[], civicGenes:ICivicGene, civicVariants:ICivicVariant): boolean
+    {
+        let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
+        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
+    
+        if (geneEntry && !geneVariants) {
+            return false;
+        }
+
```
alisman: I think this method could be more clearly written with use of hasOwnProperty method of object.   e.g.
if (!civicGenes.hasOwnProperty(geneSymbol) ...

onursumer:
[Mutation Table --> add CIVIC](https://trello.com/c/w8fqi2y5/53-mutation-table-add-civic)

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
oplantalech: Well, in mutations we only display the CIViC entries of the genes that have the specific mutation (in geneVariants). However, in CNA we want to display all the genes with an entry with description (with or without CNA variants), and but also the genes with no description but with CNA variants. The genes without CNA variants but with description will have the icon in grey instead of color.

src/shared/components/annotation/styles/civicCard.scss
oplantalech: I removed the copied part from `oncoKbCard.scss`. Now this file only contain styles that are used by CIViC.

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
oplantalech: I need to differentiate here between three "states" for `civicEntry`, and this is how I did it:
- `undefined`: the request has not been finished, so we still don't know if the gene will have a `civicEntry` or not.
- `ICivicEntry`: the request has been finished and we have Civic data for the gene to be displayed.
- `null`: the request has been finished, and there is no Civic data for the gene to be displayed.

oplantalech:
@jjgao Indeed, the calls were done if `show.civic=false`. This should be fixed now.

PR #370, opened 2017-06-01
Update OncoKB annotion sorting method

zhx828:
# What? Why?
variantExist and hotspot info do not play a role in sorting anymore
Fix https://github.com/cBioPortal/cbioportal/issues/2504.

# Checks
  the `-s` option)

# Notify reviewers
@cBioPortal/frontend

### DIFF
--- a/src/shared/components/annotation/OncoKB.spec.tsx
+++ b/src/shared/components/annotation/OncoKB.spec.tsx
@@ -160,6 +160,33 @@ describe('OncoKB', () => {
             sortedArray.indexOf(queryA), sortedArray.indexOf(queryB),
             'The score for Oncogenic variant(A) should always higher than other categories(B) even B has treatments.');
 
+        // GeneExist tests
+        queryA = initQueryIndicator({
+            geneExist: true
+        });
+        queryB = initQueryIndicator({
+            geneExist: false
+        });
+        array = [queryB, queryA];
+        sortedArray = lazyMobXTableSort<IndicatorQueryResp>(array, OncoKB.sortValue, true);
+        assert.isAbove(
+            sortedArray.indexOf(queryA), sortedArray.indexOf(queryB),
+            'geneExist test 1');
+
+        // GeneExist tests
+        queryA = initQueryIndicator({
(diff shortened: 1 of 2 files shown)


src/shared/components/annotation/OncoKB.spec.tsx
```diff
@@ -181,9 +183,40 @@ describe('OncoKB', () => {
...
+        queryB = initQueryIndicator({
+            oncogenic: 'Oncogenic',
+            hotspot: true,
+            highestSensitiveLevel: 'LEVEL_2A'
+        });
+        array = [queryB, queryA];
+        sortedArray = lazyMobXTableSort<IndicatorQueryResp>(array, OncoKB.sortValue, true);
+        assert.isBelow(
+            sortedArray.indexOf(queryA), sortedArray.indexOf(queryB),
+            'isHotspot test 1');
```
onursumer: Having a descriptive assertion message might be better here: something like, `A(LEVEL_2A) should be higher than B(...)`

src/shared/components/annotation/OncoKB.spec.tsx
```diff
@@ -181,9 +183,40 @@ describe('OncoKB', () => {
...
+        queryB = initQueryIndicator({
+            oncogenic: 'Oncogenic',
+            hotspot: false,
+            highestSensitiveLevel: 'LEVEL_2A'
+        });
+        array = [queryB, queryA];
+        sortedArray = lazyMobXTableSort<IndicatorQueryResp>(array, OncoKB.sortValue, true);
+        assert.isBelow(
+            sortedArray.indexOf(queryA), sortedArray.indexOf(queryB),
+            'isHotspot test 2');
```
onursumer: Change assertion message to `A(LEVEL_2A) should be higher than B(...) even if B is a hotspot`?

zhx828:
@onursumer I updated the descriptions and also included the issue https://github.com/cBioPortal/cbioportal/issues/2480 in this pull request as well since they are all about OncoKb sorting.

PR #372, opened 2017-06-01
Add styling for case when maxTreeDepth is 0.

Tamba32:
# What? Why?
Add styling for case when maxTreeDepth is 0.

### DIFF
--- a/src/shared/components/query/CancerStudySelector.tsx
+++ b/src/shared/components/query/CancerStudySelector.tsx
@@ -10,6 +10,7 @@ import StudyList from "./studyList/StudyList";
 import {observer, Observer} from "mobx-react";
 import {expr} from 'mobx';
 import memoize from "memoize-weak-decorator";
+import {If, Then} from 'react-if';
 import {QueryStoreComponent} from "./QueryStore";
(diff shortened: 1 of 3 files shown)


src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -240,9 +241,13 @@ export default class CancerStudySelector extends QueryStoreComponent<ICancerStud
 				</SectionHeader>
 
 				<FlexRow className={styles.cancerStudySelectorBody}>
-					<div className={styles.cancerTypeListContainer}>
-						<this.CancerTypeList/>
-					</div>
+					<If condition={this.store.maxTreeDepth}>
+						<Then>
```
alisman: We should try not rely on falsiness like this.  (this.store.maxTreeDepth > 0) is much more explicity/easier to understand for future devs.

src/shared/components/query/styles.module.scss
```diff
@@ -198,6 +198,9 @@ div.SelectedStudiesWindow {
   div.cancerStudyListContainer {
     overflow: auto;
     flex: 1;
+    &:only-child {
```
alisman: this is clever

PR #376, opened 2017-06-02
mapped mutations onto 3D protein structure

onursumer:
# Checks
  the `-s` option)


src/shared/lib/PdbUtils.ts
```diff
@@ -153,69 +131,80 @@ function groupAlignmentsByPdbIdAndChain(alignments: PdbUniprotAlignment[])
...
+ */
+export function mergeAlignments(alignments: PdbUniprotAlignment[]): IPdbChain|undefined
 {
     let mergedAlignment = "";
-    let end: number;
+    let start: number;
 
     if (alignments.length > 0) {
         // set the end just before the beginning of the first alignment
         // the iteration below will handle the rest
```
onursumer: oops, this comment is not valid anymore.

src/shared/lib/PdbUtils.ts
```diff
@@ -153,69 +131,80 @@ function groupAlignmentsByPdbIdAndChain(alignments: PdbUniprotAlignment[])
...
+        start = Math.min(start, alignment.uniprotFrom);
     });
 
     return {
         pdbId: alignments[0].pdbId,
         chain: alignments[0].chain,
-        uniprotStart: alignments[0].uniprotFrom,
-        uniprotEnd: alignments[0].uniprotFrom + mergedAlignment.length,
+        uniprotStart: start,
+        uniprotEnd: start + mergedAlignment.length - 1,
```
onursumer: In the legacy implementation we don't have the `-1`, but this would only affect the pdb chains by one pixel...

src/shared/lib/MutationUtils.ts
```diff
@@ -0,0 +1,93 @@
+import {
```
onursumer: Contents of this file is mostly moved from `LollipopMutationPlot`

src/shared/lib/PdbUtils.ts
adamabeshouse: oh ok, I understand now

onursumer:
[(8) Results Page: Mutation Table: 3D structure - diagram interaction (residue mapping)](https://trello.com/c/FYDjgdox/265-(8)-results-page%3A-mutation-table%3A-3d-structure---diagram-interaction-(residue-mapping))

src/shared/lib/MutationUtils.ts
```diff
@@ -0,0 +1,93 @@
...
+    if (priorityA < priorityB) {
+        return -1;
+    } else if (priorityA > priorityB) {
+        return 1;
+    } else {
+        return typeA.localeCompare(typeB);
+    }
+}
+
+export function getColorBasedOnProteinImpactType<T extends IProteinImpactTypeColors>(mutations: Mutation[],
```
adamabeshouse: why does this need to be generic typed?

src/shared/lib/PdbUtils.ts
adamabeshouse: maybe it would be good to document this, that `uniprotEnd` is inclusive.  Avoid some off-by-one errors :)

src/shared/lib/MutationUtils.ts
onursumer: This allows us to pass the whole `props` object as is, instead of cherry picking those 4 color props and instantiating another object. Any `props` object that has those color props can be passed to this function. But we can change it to a regular parameter if you think it is better.

src/shared/lib/PdbUtils.ts
onursumer: Good idea :)

src/shared/lib/MutationUtils.ts
onursumer: Actually, since I already explicitly added `extends IProteinImpactTypeColors` to the related component props, I think we don't need a generic type at this point.

adamabeshouse: that's already how types work

onursumer: Just removed the generic parameter, simpler is better for now

onursumer:
Needs to be rebased. Probably on rc.

onursumer:
[Results Page: Mutations Tab: 3D structure: Display a warning message in case no mutations can be mapped onto selected structure.](https://trello.com/c/Gsl8xohh/305-results-page-mutations-tab-3d-structure-display-a-warning-message-in-case-no-mutations-can-be-mapped-onto-selected-structure)

onursumer:
Closing this one due to too many rebase/merge conflicts. The new PR is #447.

PR #379, opened 2017-06-05
Fix PmidCache bug - not wrapping AugmentedData correctly

adamabeshouse:
Typescript didnt catch it because of `any` typing the server response.

As a way of adding what testing should have caught this, I've added a type for the pmid result, which catches this.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/cache/PmidCache.ts
+++ b/src/shared/cache/PmidCache.ts
@@ -2,26 +2,83 @@ import LazyMobXCache from "../lib/LazyMobXCache";
 import {AugmentedData} from "../lib/LazyMobXCache";
 import request from "superagent";
 
-async function fetch(pmids: number[])
+export type PmidData = {
+        [pmid:string]:PmidDatum;
+};
+
+export type PmidDatum = {
+    uid: string,
+    pubdate: string,
+    epubdate: string,
+    source: string,
+    authors: { name: string, authtype: string, clusterid: string}[],
+    lastauthor: string,
+    title: string,
+    sorttitle: string,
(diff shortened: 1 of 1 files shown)


adamabeshouse:
image

adamabeshouse:
[Bug: OncoKB tooltip only shows abstracts not PMIDs](https://trello.com/c/IHR08mxz/299-bug-oncokb-tooltip-only-shows-abstracts-not-pmids)

src/shared/cache/PmidCache.ts
```diff
@@ -2,34 +2,91 @@ import LazyMobXCache from "../lib/LazyMobXCache";
 import {AugmentedData} from "../lib/LazyMobXCache";
 import request from "superagent";
 
-async function fetch(pmids: number[])
+export type PmidData = {
+        [pmid:string]:PmidDatum;
+};
+
+export type PmidDatum = {
```
onursumer: Wow, thanks for doing this!

Can we move this one under model? I can update other references to use this model within the OncoKB related components later.

onursumer: Actually it might be better to name the model `PubMedRecord` and the cache `PubMedCache`. Let's merge this PR as is, we can make those changes later.

adamabeshouse:
https://github.com/cBioPortal/cbioportal-frontend/pull/382

src/shared/cache/PmidCache.ts
adamabeshouse: both done in other PR

PR #380, opened 2017-06-05
Using property values to enable/disable annotations

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2518

# Checks
  the `-s` option)


onursumer:
Note that this PR does not fully cover Civic annotation enabling/disabling, since the related PR #367 is still pending.

onursumer:
We should also do the same thing for `ResultsViewPage`, but that component is not yet in the `hotfix` branch.

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -44,6 +45,10 @@ type ICopyNumberTableWrapperProps = {
 @observer
 export default class CopyNumberTableWrapper extends React.Component<ICopyNumberTableWrapperProps, {}> {
 
+    public static defaultProps = {
+        enableOncoKb: true
+    };
+
```
inodb: @onursumer Don't fully follow this part, why not use `AppConfig` here?

PR #385, opened 2017-06-08
Ctdna tab

pambot:
This is a pull request to keep track of changes to the new ctDNA tab in the Patient View, as [proposed here](https://drive.google.com/open?id=1Xl-jJM8RkZ9SqfBr92sWqICG5f_er92nJAB_JpRlswY) for GSoC 2017.

Reviewers:
@inodb


src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
```diff
@@ -4,9 +4,7 @@ import {Mutation} from "../../../shared/api/generated/CBioPortalAPI";
 import {ClinicalDataBySampleId} from "../../../shared/api/api-types-extended";
 import SampleManager from "../sampleManager";
 
-const Highcharts = require('highcharts');
-const addHeatmap = require('highcharts/modules/heatmap');
-addHeatmap(Highcharts);
+const Chart = require('chart.heatmap.js/dst/Chart.HeatMap.S');
```
alisman: curious, what library is this?

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
```diff
@@ -25,64 +23,7 @@ export default class HeatmapVAFEvents extends React.Component<IHeatmapVAFEventsP
     this.chart = undefined;
```
alisman: you can do this assignment right where you declare the property.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
```diff
@@ -96,17 +37,62 @@ export default class HeatmapVAFEvents extends React.Component<IHeatmapVAFEventsP
...
+    let plotMaker = {
+      data: plotData,
+      xTicks: sampleOrder,
+      yTicks: heatmapGeneList,
+    };
+    return plotMaker;
+  }
+
+  componentDidMount() {
+    let plotMaker = this.extractPlotData();
```
alisman: If you do this in component did mount, you should make sure that this component cannot re-render in the future, as that could wipe out the canvas.   you should put the call to plot in componentDidUpdate.   that way it could re-execute if necessary.  then implement shouldComponentUpdate to determine if re-render/plot is actually necessary.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
```diff
@@ -96,17 +37,62 @@ export default class HeatmapVAFEvents extends React.Component<IHeatmapVAFEventsP
...
+          data: [4, 0, 7, 4, 6, 3, 2, 4, 2, 10, 8, 2]
+        },
+        {
+          label: 'Friday',
+          data: [1, 0, 0, 7, 0, 4, 1, 3, 4, 5, 1, 10]
+        }
+      ]
+    };
+
+    var canvas:any = document.getElementById("heatmap");
```
alisman: we should never use id's in this manner because then there can only be one instance of this on a page.  it is not self-contained as it relies on document global.   instead use React refs to get a reference to <canvas> element. https://facebook.github.io/react/docs/refs-and-the-dom.html

alisman:
you should make this pr to rc branch, not integration.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
pambot: @alisman It's a [heatmap module](http://tmroyal.github.io/Chart.HeatMap/) of Chart.js, which JJ showed me the other day.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
pambot: @alisman Cool, I can fix it. I only put it that way because Typescript won't let me assign the property in the constructor without declaring it above, which I wasn't used to, so I went with the first thing the worked.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
pambot: @alisman I think I fixed it later last night. It needed to be done, but I wanted to save working changes first.

src/pages/patientView/ctDNA/HeatmapVAFEvents.tsx
pambot: @alisman Okay cool

inodb:
We'll close this for now, because we're retiring `integration`

PR #386, opened 2017-06-08
Removed DataSets table row limit

Tamba32:
# What? Why?
Fix issue where DataSets table was limiting number of studies to 50. DataSets table now shows all studies.

Resolved https://github.com/cBioPortal/cbioportal/issues/2550


alisman:
We should probably change default behavior of table, but no worth it to do this now.

PR #388, opened 2017-06-08
Edit Subscribe Field

Tamba32:
# What? Why?
Remove subscribe input. Add anchor to subscribe button.

Resolves https://github.com/cBioPortal/cbioportal/issues/2551

### DIFF
--- a/src/pages/datasetView/DatasetList.tsx
+++ b/src/pages/datasetView/DatasetList.tsx
@@ -154,6 +154,7 @@ export default class DataSetsPageTable extends React.Component <IDataSetsTablePr
                         showColumnVisibility={true}
                         showFilter={true}
                         showCopyDownload={false}
+                        initialItemsPerPage={this.props.datasets.length}
                     />
                 </div>
             );
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -24,7 +24,7 @@ export default class RightBar extends React.Component<{}, {}> {
 
     static getExampleSection() {
         const title:string = 'Example Queries';
-        const defaultExamples:JSX.Element =
+        const defaultExamples:JSX.Element = (
             <div className="rightBarSection exampleQueries">
                 <h3>{title}</h3>
                 <ul>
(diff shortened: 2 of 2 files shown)


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -28,15 +28,10 @@ export default class RightBar extends React.Component<{}, {}> {
...
-                        <form className="form-inline">
-                            <div className="form-group form-group-sm">
-                                <input type="email" className="form-control" id="exampleInputEmail2" placeholder="jane.doe@example.com" />
-                            </div>
-
-                            <button type="submit" className="btn btn-default btn-sm">Subscribe</button>
-                        </form>
+                    <p>Sign up for low-volume email news alerts:</p>
+                    <a target="_blank" href="http://groups.google.com/group/cbioportal-news/boxsubscribe">
+                        <button type="button" className="btn btn-default btn-sm" style={{width: "100%"}}>Subscribe</button>
```
alisman: we shouldn't wrap a button in an a tag.  not semantically correct.  the bootstrap classes (btn btn-default btn-sm) can actually be assigned to <a> tag (style tag too) and then we can lose the <button> altogether

src/shared/components/rightbar/RightBar.tsx
```diff
@@ -50,14 +50,13 @@ export default class RightBar extends React.Component<{}, {}> {
                         <a href="case.do#/patient?studyId=ucec_tcga_pub&caseId=TCGA-BK-A0CC">Patient view of an endometrial cancer case</a>
                     </li>
                 </ul>
-            </div>;
+            </div>);
         console.log(AppConfig.skinRightNavExamplesHTML);
```
alisman: lest get rid of console statement

src/pages/datasetView/DatasetList.tsx
```diff
@@ -154,6 +154,7 @@ export default class DataSetsPageTable extends React.Component <IDataSetsTablePr
                         showColumnVisibility={true}
                         showFilter={true}
                         showCopyDownload={false}
+                        initialItemsPerPage={this.props.datasets.length}
```
alisman: seems like this shouldn't be part of this PR.  could an issue with git usage

PR #390, opened 2017-06-09
Populate query form with current selection when going from study view

adamabeshouse:
# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -19,6 +19,8 @@ import URL from 'url';
 import {buildCBioPortalUrl, BuildUrlParams} from "../../api/urls";
 import {SyntaxError} from "../../lib/oql/oql-parser";
 import StudyListLogic from "./StudyListLogic";
+import {QuerySession} from "../../lib/QuerySession";
+import {stringListToSet} from "../../lib/StringUtils";
 
 // interface for communicating
 type CancerStudyQueryUrlParams = {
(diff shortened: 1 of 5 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -124,6 +128,13 @@ export class QueryStore
 		this._selectedStudyIds = this.forDownloadTab ? ids.slice(-1) : ids;
 	}
 
+	@computed get selectedStudyIdsSet():{[studyId:string]:boolean}
+	{
+		return this.selectedStudyIds.reduce((ret:{[studyId:string]:boolean}, next:string)=>{
```
gideonite: Might be easier to read if you give this anonymous function a simple name like `setStudyIdTrue` or something like that.

src/shared/components/query/QueryStore.ts
```diff
@@ -87,6 +88,9 @@ export class QueryStore
 		labelMobxPromises(this);
 		if (urlWithInitialParams)
 			this.setParamsFromUrl(urlWithInitialParams);
+
```
alisman: are we sure it makes sense to override params from URL with params from query session?  i'm not suggesting otherwise i just want to make sure we thought about it

adamabeshouse:
[Query form not populating when coming from study summary](https://trello.com/c/0af1C5Zv/307-query-form-not-populating-when-coming-from-study-summary)

PR #391, opened 2017-06-09
Variant allele frequencies always shown with 2 decimal digits

adamabeshouse:
# Checks
  the `-s` option)


onursumer:
@adamabeshouse Is this any different than `toFixed()`?: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed

src/shared/lib/StringUtils.ts
```diff
@@ -15,4 +15,40 @@ export function longestCommonStartingSubstring(str1:string, str2:string) {
     }
 
     return str1.substring(0, i);
+}
+
+export function formatDecimal(x:number, numDigitsAfterDecimal:number) {
```
alisman: I think javascript's toFixed method does this. https://www.w3schools.com/jsref/jsref_tofixed.asp

adamabeshouse: oh wow haha

PR #399, opened 2017-06-13
Improved mutation id generating functions

onursumer:
# What? Why?
- Hotfix for https://github.com/cBioPortal/cbioportal/issues/2561.
- Need to apply the same fix and add tests for the release branch, since `PatientViewStore` is pretty outdated in hotfix branch.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -692,7 +692,7 @@ export class PatientViewPageStore {
         let idToMutations: {[key: string]: Array<Mutation>} = {};
         let mutationId: string;
         let MutationId: (m: Mutation) => string = (m: Mutation) => {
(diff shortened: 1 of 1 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -705,7 +705,7 @@ export class PatientViewPageStore {
     }
 
     static getMutationId(m: any): string {
```
adamabeshouse: (1) can we make the input parameter type any more specific?
(2) why not using this instead of `MutationId` on line 694?

onursumer: (1) already fixed in integration/release branches (was introduced by #326)
(2) already fixed in integration/release branches (was introduced by #326)

I don't want to fix those again in hotfix branch :)

PR #406, opened 2017-06-15
when samples are clicked with mod keys, sample should open in new tab…

alisman:
NOTE: probably we will merge this to 1.7.1, not 1.7.0.  No reason to destabilize.

image

When user is in a cohort, they want to be able to open specific samples in a new tab for future reference in isolation. We've changed that behavior, but we can restore it on conditional basis -> when user holds down modifier keys.

### DIFF
--- a/src/pages/patientView/PatientViewPage.spec.tsx
+++ b/src/pages/patientView/PatientViewPage.spec.tsx
@@ -1,9 +1,66 @@
-//import {  } from './PatientViewPage';
+import  PatientViewPage from './PatientViewPage';
 import React from 'react';
-import { assert } from 'chai';
-import { shallow, mount } from 'enzyme';
+import {assert} from 'chai';
+import {shallow, mount} from 'enzyme';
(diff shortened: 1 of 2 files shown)


src/pages/patientView/PatientViewPage.spec.tsx
```diff
@@ -1,9 +1,66 @@
...
+const componentUnderTest: PatientViewPage = (PatientViewPage as any).wrappedComponent;
+
+
 describe('PatientViewPage', () => {
 
+    describe('handleSampleClick', () => {
+
+        const handleSampleClick = (componentUnderTest as any).prototype.handleSampleClick;
+
+        let updateRouteStub: sinon.SinonStub, preventDefaultStub: sinon.SinonStub, mock: any, ev: Partial<React.MouseEvent<HTMLAnchorElement>>;
```
onursumer: might look nicer to have one variable per line, not a big deal though.

PR #409, opened 2017-06-16
Fix gene set dialog style issues (when on results page)

alisman:
Modals need cbioportal-frontend class and some help with style clash on buttons 

https://github.com/cBioPortal/cbioportal/issues/2598

### DIFF
--- a/src/shared/components/paginationControls/paginationControls.module.scss
+++ b/src/shared/components/paginationControls/paginationControls.module.scss
@@ -15,3 +15,7 @@
     padding: 0 5px 0 5px;
     display: inline-block;
 }
+
+:global(.legacy) .paginationControls button i {
+    line-height:1.5;
+}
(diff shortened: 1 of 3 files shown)


src/shared/components/paginationControls/paginationControls.module.scss
```diff
@@ -14,4 +14,8 @@
     height: 18px;
     padding: 0 5px 0 5px;
     display: inline-block;
+}
+
+:global(.legacy) .paginationControls button i {
```
alisman: using legacy here so that it only applies when on pages with html[class=legacy].  When we are finished with refactor, we can then delete legacy

onursumer: Is it possible to add a `TODO` comment inside the code itself as a secondary reminder to ourselves? :)

PR #418, opened 2017-06-21
Fix cBioPortal/cbioportal#2613 filter uncalled

inodb:
Only show uncalled mutations if the mutation is called in any of the patient's
other samples.

Fix cbioportal/cbioportal#2613

### DIFF
--- a/src/shared/lib/StoreUtils.spec.ts
+++ b/src/shared/lib/StoreUtils.spec.ts
@@ -16,6 +16,7 @@ describe('StoreUtils', () => {
     let mutationDataWithNoKeyword: MobxPromise<Mutation[]>;
     let mutationDataWithKeywords: MobxPromise<Mutation[]>;
     let mutationDataWithFusionsOnly: MobxPromise<Mutation[]>;
+    let uncalledMutationDataWithNewEvent: MobxPromise<Mutation[]>;
     let mutationDataWithMutationsOnly: MobxPromise<Mutation[]>;
     let mutationDataWithBothMutationsAndFusions: MobxPromise<Mutation[]>;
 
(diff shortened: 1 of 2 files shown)


src/shared/lib/StoreUtils.spec.ts
```diff
@@ -108,6 +109,27 @@ describe('StoreUtils', () => {
             }),
         ];
 
+        mutationDataWithNewEvent =  {
```
onursumer: Should we name this variable as `uncalledMutationDataWithNewEvent` or something like that. This is uncalled mutation data, right?

src/shared/lib/StoreUtils.ts
```diff
@@ -479,21 +479,24 @@ export function mergeMutationsIncludingUncalled(mutationData:MobxPromise<Mutatio
...
+    updateIdToMutationsMap(idToMutations, uncalledMutationData, generateMutationId, true);
 
     return Object.keys(idToMutations).map((id:string) => idToMutations[id]);
 }
 
 function updateIdToMutationsMap(idToMutations: {[key: string]: Mutation[]},
                                 mutationData:MobxPromise<Mutation[]>,
-                                generateMutationId:MutationIdGenerator = generateMutationIdByGeneAndProteinChangeAndEvent)
+                                generateMutationId:MutationIdGenerator = generateMutationIdByGeneAndProteinChangeAndEvent,
+                                onlyUpdateIfExists: boolean)
```
onursumer: I can't immediately see `what exists` by just looking at name of the variable? Maybe make it more clear?

src/shared/lib/StoreUtils.ts
```diff
@@ -479,21 +479,24 @@ export function mergeMutationsIncludingUncalled(mutationData:MobxPromise<Mutatio
...
 {
     if (mutationData.result) {
         for (const mutation of mutationData.result) {
             const mutationId = generateMutationId(mutation);
-            idToMutations[mutationId] = idToMutations[mutationId] || [];
-            idToMutations[mutationId].push(mutation);
+            if (!onlyUpdateIfExists || (onlyUpdateIfExists && (mutationId in idToMutations))) {
```
onursumer: You don't need the check if `onlyUpdateIfExist` on the right side of OR, since you are already checking if not `onlyUpdateIfExists` on the left side of OR :) When `onlyUpdateIfExists` is false, left hand side of the condition will always be true ;) So let's simplify this to:
`!onlyUpdateIfExists || (mutationId in idToMutations)`

inodb: Good point ;)

src/shared/lib/StoreUtils.ts
inodb: @onursumer how about `onlyUpdateExistingIds`?

PR #419, opened 2017-06-22
improved query selector style to reduce cross browser inconsistencies

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2625.
Also fixes inconsistency between gene set text area and case set text area.

Chrome:
queryselector_chrome

IE 11:
queryselector_ie

Firefox:
queryselector_firefox

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/query/styles.module.scss
+++ b/src/shared/components/query/styles.module.scss
@@ -61,7 +61,11 @@
   }
 
   :global(.sectionLabel) {
-    width:246px;
+    width:236px;
+  }
+
+  :global(.sectionContent) {
+    margin-left: 0 !important;
   }
 
   label {
@@ -192,7 +196,8 @@ div.SelectedStudiesWindow {
   }
 
   div.cancerTypeListContainer {
-    overflow: auto;
+    overflow-y: auto;
+    overflow-x: hidden;
   }
 
   div.cancerStudyListContainer {
@@ -219,7 +224,7 @@ div.SelectedStudiesWindow {
       color: #043852;
       border-bottom: 1px solid rgb(222, 233, 241);
       border-top: 1px solid rgb(243, 248, 251);
-      padding: 8px 10px;
+      padding: 8px 16px 8px 10px;
 
       span {
         &.cancerTypeListItemLabel {
@@ -344,7 +349,8 @@ div.SelectedStudiesWindow {
     width: 555px;
     resize: vertical;
     padding: 5px;
-    border-color:$borderColor;
+    border: 1px solid $borderColor;
+    border-radius: 2px;
   }
   .radioRow {
     div {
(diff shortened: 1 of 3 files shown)


onursumer:
Actually, maybe we should also increase the size of the cancer study selector panel to get rid of vertical scroll in IE...

src/shared/components/query/styles.module.scss
```diff
@@ -344,7 +349,8 @@ div.SelectedStudiesWindow {
     width: 555px;
     resize: vertical;
     padding: 5px;
-    border-color:$borderColor;
+    border: 1px solid $borderColor;
+    border-radius: 2px;
```
onursumer: Without these 2 lines, case set selector text area looks different in IE and Firefox.

src/shared/components/query/styles.module.scss
```diff
@@ -61,7 +61,12 @@
   }
 
   :global(.sectionLabel) {
-    width:246px;
+    width:236px;
+  }
+
+  :global(.centeredSectionContent) {
+    margin-left: auto !important;
+    margin-right: auto !important;
```
onursumer: another solution would be to justify the contents to the right instead of centering

src/shared/components/query/styles.module.scss
```diff
@@ -192,7 +197,8 @@ div.SelectedStudiesWindow {
   }
 
   div.cancerTypeListContainer {
-    overflow: auto;
+    overflow-y: auto;
+    overflow-x: hidden;
```
onursumer: This is to hide horizontal scrollbar, which is visible in IE (see screenshot above).

src/shared/components/query/styles.module.scss
```diff
@@ -219,7 +225,7 @@ div.SelectedStudiesWindow {
       color: #043852;
       border-bottom: 1px solid rgb(222, 233, 241);
       border-top: 1px solid rgb(243, 248, 251);
-      padding: 8px 10px;
+      padding: 8px 16px 8px 10px;
```
onursumer: This slightly increases right padding for the cancer type list item (the number next to cancer type). Those numbers are partially covered by the scrollbar in IE, Firefox, and Safari. This makes them fully visible.

src/shared/components/query/styles.module.scss
```diff
@@ -61,7 +61,11 @@
   }
 
   :global(.sectionLabel) {
-    width:246px;
+    width:236px;
+  }
+
+  :global(.sectionContent) {
+    margin-left: 0 !important;
   }
```
inodb: If we could prevent using globals here would be better imo

src/shared/components/query/styles.module.scss
inodb: Better to add this comment to the code

inodb:
This PR should prolly be to hotfix now?

alisman:
I'm gonna spend a little time look at this before we merge.

alisman:
I tried setting container of cancer types (left col) to overflow:scroll.  This normalizes width in all browsers because they all put the scroll outside the dev instead of inside.   The width of the react-select controls seems related to IE9 mishandling min-width of contentWidth class.  Please investigate that.

PR #420, opened 2017-06-23
Incorporate maxTreeDepth of 0 into Download Studies component 

Tamba32:
# What? Why?
Made it so download tab of home page preserves maxTreeDepth when it is set to 0.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -234,7 +234,7 @@ export class QueryStore
 	@observable private _maxTreeDepth:number = (window as any).maxTreeDepth;
 	@computed get maxTreeDepth()
 	{
-		return this.forDownloadTab ? 1 : this._maxTreeDepth;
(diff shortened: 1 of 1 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -234,7 +234,7 @@ export class QueryStore
 	@observable private _maxTreeDepth:number = (window as any).maxTreeDepth;
 	@computed get maxTreeDepth()
 	{
-		return this.forDownloadTab ? 1 : this._maxTreeDepth;
+		return (this.forDownloadTab && this._maxTreeDepth > 0) ? 1 : this._maxTreeDepth;
```
alisman: hi, not sure if i understand that logic here.  so, basically, the download tab can only ever have a depth of 0 or 1?  what's the reason for that?

alisman: also, this would be a good spot for a comment.  whenever we implement business logic that requires further knowledge to understand, a comment is called for.

PR #422, opened 2017-06-24
Add end-to-end tests webdriverio+browserstack

inodb:
Add end to end tests using webdriverio and browserstack. This runs cbioportal
frontend code running on http://localhost:3000 inside the cBioPortal instance
specified in __API_ROOT__  of my-index.ejs using the localStorage trick.

### DIFF
--- a/end-to-end-tests/specs/home.spec.js
+++ b/end-to-end-tests/specs/home.spec.js
@@ -1,0 +1,93 @@
+var assert = require('assert');
+
+const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+
+describe('homepage', function() {
+
+    it('it should show dev mode when testing', function() {
+        browser.url(CBIOPORTAL_URL);
+
+        browser.localStorage('POST', {key: 'localdev', value: 'true'});
+        browser.refresh();
+
+        var devMode = $('.alert-warning');
+
+        devMode.waitForExist(30000);
+        assert(browser.getText('.alert-warning').indexOf('dev mode') > 0);
+    });
+
+    it('it should have 29 studies in list', function () {
+        browser.url(CBIOPORTAL_URL);
+
+        var studies = $('[data-test="cancerTypeListContainer"] > ul > ul');
+        
+        studies.waitForExist(10000); // same as `browser.waitForExist('.notification', 10000)`
+        
+        //assert.equal(browser.elements('[@data-test=cancerTypeListContainer] > ul > ul').value.length, 29);
+        
+    });
+
+
+    it('should filter study list according to filter text input', function () {
+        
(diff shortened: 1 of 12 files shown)


end-to-end-tests/specs/home.spec.js
```diff
@@ -0,0 +1,93 @@
...
+
+    it('it should show dev mode when testing', function() {
+        browser.url(CBIOPORTAL_URL);
+
+        browser.localStorage('POST', {key: 'localdev', value: 'true'});
+        browser.refresh();
+
+        var devMode = $('.alert-warning');
+
+        devMode.waitForExist(30000);
```
alisman: how come so big?

end-to-end-tests/specs/home.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        assert(browser.getText('.alert-warning').indexOf('dev mode') > 0);
+    });
+
+    it('it should have 29 studies in list', function () {
+        browser.url(CBIOPORTAL_URL);
```
alisman: one thing i'm not sure about: if browser is already at this url, does it reload?  we should figure that out because it cost a lot of time to reload for every test.  i guess we can just watch the tests run.

end-to-end-tests/specs/home.spec.js
```diff
@@ -0,0 +1,93 @@
...
+        
+        var caseSetSelectorClass = '[data-test="CaseSetSelector"]';
+        
+        var checkBox = $('[data-test="StudySelect"]');
+
+        checkBox.waitForExist(10000);
+        
+        assert.equal(browser.isExisting(caseSetSelectorClass), false);
+        
+        browser.click('[data-test="StudySelect"]');
```
alisman: i think this i my code but we should figure out if we can just call click on checkbox itself.   also, we should write a little wrapper to abstract away data-test.  i.e. makeSelector('StudySelect")

end-to-end-tests/specs/home.spec.js
alisman: we should make defaultTimeout so we can easily adjust.  there might be one already.  30,0000 seems too high to me.

inodb: @alisman It's 30,000 so 30s. Travis can be quite slow (especially the startup), this number is based on experimentation. I think it uses default timeout if you don't put it a number, which is configured in the webdriver conf.

end-to-end-tests/specs/home.spec.js
inodb: @alisman good point

end-to-end-tests/specs/home.spec.js
inodb: @alisman yeah some helper function would be useful

PR #423, opened 2017-06-26
Overall Survival Months clinical attribute in patient view precision

adamabeshouse:
Now has no decimal digit precision

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
+++ b/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
@@ -21,6 +21,19 @@ interface IPatientRow {
 
 export default class ClinicalInformationPatientTable extends React.Component<IClinicalInformationPatientTableProps, {}> {
 
+    private getDisplayValue(data:{attribute:string, value:string}):string {
+        let ret:string;
+        switch (data.attribute) {
+            case "Overall Survival (Months)":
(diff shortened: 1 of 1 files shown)


src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -21,6 +21,14 @@ interface IPatientRow {
 
 export default class ClinicalInformationPatientTable extends React.Component<IClinicalInformationPatientTableProps, {}> {
 
+    private renderValue(attribute:string, value:string):string {
+        if (attribute === "Overall Survival (Months)") {
+            return parseInt(value, 10).toFixed(0) + '';
```
alisman: what does addition of empty string at the end do?  toString?

onursumer: Seems like `toFixed` already returns a string value...

adamabeshouse: @onursumer yes of course...how else could it work :man_facepalming:

onursumer: So, we don't need to append an empty string to the end, do we?

PR #424, opened 2017-06-27
added OncoKB error icon

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2655.

oncokb_error_icon

# Changes
- Showing an error icon in place of OncoKB icon in case cancer type cannot be resolved
- Falling back to `cancer_type` in case `cancer_type_detailed` cannot be found

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/annotation/OncoKB.tsx
+++ b/src/shared/components/annotation/OncoKB.tsx
@@ -19,7 +19,7 @@ import OncokbPubMedCache from "shared/cache/PubMedCache";
 import {default as TableCellStatusIndicator, TableCellStatus} from "shared/components/TableCellStatus";
 
 export interface IOncoKbProps {
-    indicator?: IndicatorQueryResp;
+    indicator?: IndicatorQueryResp | null;
     evidenceCache?: OncoKbEvidenceCache;
     evidenceQuery?: Query;
     pubMedCache?: OncokbPubMedCache;
@@ -46,7 +46,7 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
         };
     }
 
-    public static sortValue(indicator:IndicatorQueryResp|undefined):number[]
+    public static sortValue(indicator?: IndicatorQueryResp|undefined|null): number[]
     {
         if (!indicator) {
             return [];
@@ -78,7 +78,15 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
             <span className={`${annotationStyles["annotation-item"]}`} />
         );
 
-        if (this.props.indicator)
+        if (this.props.indicator === null) {
+            // null means there is an error...
+            oncoKbContent = this.errorIcon();
+        }
+        else if (this.props.indicator === undefined) {
+            // undefined means still loading...
+            oncoKbContent = this.loaderIcon();
+        }
+        else
         {
             oncoKbContent = (
                 <span className={`${annotationStyles["annotation-item"]}`}>
@@ -116,15 +124,6 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
                 );
             }
         }
-        else
-        {
-            // Here we assume that every OncoKB component is eventually updated with a valid
-            // IndicatorQueryResp property.
-            //
-            // Ideally we should distinguish "Loading", "NA" and "Error" states. This requires implementing
-            // an OncoKbIndicator cache...
-            oncoKbContent = this.loaderIcon();
-        }
 
         return oncoKbContent;
     }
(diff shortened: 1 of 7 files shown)


src/shared/components/annotation/OncoKB.tsx
```diff
@@ -78,7 +78,15 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
             <span className={`${annotationStyles["annotation-item"]}`} />
         );
 
-        if (this.props.indicator)
+        if (this.props.indicator === null) {
+            // null means there is an error...
+            oncoKbContent = this.errorIcon();
```
onursumer: Probably, this is not the best way to distinguish between `error` and `loading` states. Ideally there should be a `status` indicator for each remote data. But currently, we have the `status` indicator only for `cache` data.

onursumer: Actually, we have status in remote data (`MobXPromise`), but we are only passing the `result` field (the actual data part) to the components.

Maybe we should redesign our components so that they can be aware of the `status` of the data? I guess that is out of the scope of this hotfix PR...

src/shared/lib/StoreUtils.spec.ts
```diff
@@ -288,10 +288,20 @@ describe('StoreUtils', () => {
         });
     });
 
-    it("won't fetch onkokb data if there are no mutations", (done) => {
-        fetchOncoKbData({}, emptyMutationData).then((data: any) => {
-            assert.deepEqual(data, {sampleToTumorMap: {}, indicatorMap: {}});
-            done();
+    describe('fetchOncoKbData', () => {
+        it("won't fetch onkokb data if there are no mutations", (done) => {
+            fetchOncoKbData({}, emptyMutationData).then((data: any) => {
```
alisman: interesting that you have to use done.  i guess js promises are always async even then there is no async call inside.

src/shared/lib/StoreUtils.ts
```diff
@@ -287,15 +295,20 @@ export async function fetchCnaOncoKbData(sampleIdToTumorType:{[sampleId: string]
                                          discreteCNAData:MobxPromise<DiscreteCopyNumberData[]>,
                                          client: OncoKbAPI = oncokbClient)
 {
-    if (discreteCNAData.result && discreteCNAData.result.length > 0) {
+    if (!discreteCNAData.result || discreteCNAData.result.length === 0) {
```
alisman: not important for this release, but we really have to start commenting business logic.  like, for me, relationship between discreteCNAData and oncokb is opaque.

src/shared/lib/StoreUtils.spec.ts
adamabeshouse: if you wanted to avoid using done, you may be able to use the fake clock library `lolex`. It makes the test run a bit faster. Not super necessary though. It's already in package.json, I'm pretty sure it's used in the `accumulatingDebounce` tests

src/shared/components/annotation/OncoKB.tsx
adamabeshouse: I think it would be better to add a prop `error:boolean` - it's a bit confusing this way.

src/shared/components/annotation/OncoKB.tsx
```diff
@@ -78,7 +78,15 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
...
+        }
+        else if (this.props.indicator === undefined) {
+            // undefined means still loading...
```
adamabeshouse: it would be good, I think, to have a prop `loading`. This is a visualization, so `loading` and `error` as props, visualizations specs, makes sense to me, rather than overloading a prop whose name (indicator) doesn't indicate relation to `loading` or `error` viz.

src/shared/lib/StoreUtils.ts
onursumer: This one fetches OncoKB data for the CNA table. We are getting `gene` and `sample id` information from the CNA data. Rest of the business logic is identical to the method where we get OncoKB data for mutation table.

src/shared/components/annotation/OncoKB.tsx
```diff
@@ -136,6 +135,23 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
...
     }
 
+    public errorIcon()
+    {
+        return (
+            <DefaultTooltip
+                overlay={<span>Error fetching OncoKB data</span>}
+                placement="right"
+                trigger={['hover', 'focus']}
+                arrowContent={<div className="rc-tooltip-arrow-inner"/>}
```
adamabeshouse: `arrowContent` is one of the `<DefaultTooltip>` props, no need to add it again

onursumer: Good to know. Seems like it was added at a later stage in development. Initially we didn't have that prop. In this case, I think we still have some leftover default `arrowContent`s in other components, too...

src/shared/components/annotation/OncoKB.tsx
onursumer: makes sense.

src/shared/lib/StoreUtils.spec.ts
onursumer: Actually, I copied this test some time ago while introducing `StoreUtils`. So, that `done` has always been there :)

PR #427, opened 2017-06-27
Improved OncoKB icon status handling

onursumer:
# Why?
- Addresses remaining issues of PR #424 
- Better handling of missing cancer type value, and OncoKB API error
- Now also fixes https://github.com/cBioPortal/cbioportal/issues/2709

# Changes
- Added a `status` prop for `OncoKB` component
- Removed some redundant `arrowContent`s
- Sending a request to OncoKB API even if cancer type cannot be resolved (OncoKB annotation does not always require the cancer type)
- Not showing the `Something went wrong` popup when there is an OncoKB API error, instead showing the error icon in the Annotation column.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -25,6 +25,7 @@ import {default as PatientViewMutationTable} from "./mutation/PatientViewMutatio
 import PathologyReport from "./pathologyReport/PathologyReport";
 import { MSKTabs, MSKTab } from "../../shared/components/MSKTabs/MSKTabs";
 import { validateParametersPatientView } from '../../shared/lib/validateParameters';
+import {ONCOKB_ERROR} from "shared/lib/StoreUtils";
 import LoadingIndicator from "shared/components/loadingIndicator/LoadingIndicator";
 import ValidationAlert from "shared/components/ValidationAlert";
 import AjaxErrorModal from "shared/components/AjaxErrorModal";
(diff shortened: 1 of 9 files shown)


src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
```diff
@@ -29,14 +29,46 @@ export interface IAnnotation {
...
+            oncoKbStatus: "complete",
+            myCancerGenomeLinks: [],
+            isHotspot: false,
+            is3dHotspot: false
+        };
+    }
+
+    /**
+     * Derives status from data content.
+     * Default (empty) data means loading. Null data means error.
```
onursumer: This is still not ideal way of deriving the status...

onursumer:
@alisman @inodb I guess `master` is the wrong target for this PR. Should it be on `hotfix` or `release-1.7.1`?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -335,7 +336,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                         myCancerGenomeData={patientViewPageStore.myCancerGenomeData}
                                         hotspots={patientViewPageStore.indexedHotspotData}
                                         cosmicData={patientViewPageStore.cosmicData.result}
-                                        oncoKbData={patientViewPageStore.oncoKbData.result}
+                                        oncoKbData={patientViewPageStore.oncoKbData.error ? ONCOKB_ERROR : patientViewPageStore.oncoKbData.result}
```
onursumer: I think, what we really should do here is to pass the entire `patientViewPageStore.oncoKbData`  to the child component, instead of just the `result`. Or at least we should pass both the `result` and the `status`. That way, we don't need to derive status from the `result`.

PR #437, opened 2017-06-29
Implement Chartjs.

Tamba32:
# What? Why?
Add bar chart to right bar.


src/shared/components/barGraph/BarGraph.tsx
```diff
@@ -129,74 +137,84 @@ export default class BarGraph extends React.Component<IBarGraphProps, {colors: s
...
-                    labels: {style: {fontSize: '10px'}}},
-
-            yAxis: {
-                allowDecimals: false,
-                min: 0,
-                // max: 6000,
-                title: {text: ''}
-            },
+        const datasets = _.flattenDeep(cancerTypeStudiesArray.map((cancerStudySet, i) => (
+            cancerStudySet.studies.sort((a: ICounterObj, b:ICounterObj) => +b.allSampleCount - +a.allSampleCount).map((cancerStudy: CancerStudy, j: number) => {
```
alisman: had to look that syntax + syntax up.  i don't think this should be necessary.  what else could these counts be beside numbers?   also, where is ICounterObj declared?

src/shared/components/barGraph/BarGraph.tsx
```diff
@@ -1,18 +1,24 @@
...
-import HCE from 'highcharts/modules/exporting';
-HCE(Highcharts);
 
+import Chart from 'chart.js';
 
 export interface IBarGraphProps {
     data:CancerStudy[];
 };
 
+interface ITooltipItem {
```
alisman: chart.js typings has a type for ChartTooltipItem

src/shared/components/barGraph/BarGraph.tsx
```diff
@@ -129,74 +137,84 @@ export default class BarGraph extends React.Component<IBarGraphProps, {colors: s
...
+            type: 'horizontalBar',
+            data,
+            options
         });
+
     }
 
     render() {
-        return <div ref={el => this.chartTarget = el} />;
+        return <canvas id="chart" height="500px"/>;
```
alisman: use ref, not id

src/shared/components/barGraph/BarGraph.tsx
Tamba32: Without the "+", I receive a TS error:
```
TS2362: The left-hand side of an arithmetic operation must be of type 'any', 'number', or an enum type.
```
I do not know why I receive this error because on line 91, it allows `allSampleCount` to be added to another number.

ICounterObj is defined on line 28.

alisman: ICounterObj doesn't seem to have a property allSampleCount.  i'm confused

Tamba32:
[Finish the bar plot on the home page](https://trello.com/c/Ip5q3tYY/312-finish-the-bar-plot-on-the-home-page)

PR #439, opened 2017-06-30
Case set selector should filter on plain text of options.

alisman:
https://github.com/cBioPortal/cbioportal/issues/2679

Case Set selector in study view should filter on plain text of the set name (was filtering ID)

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/shared/components/query/CaseSetSelector.tsx
+++ b/src/shared/components/query/CaseSetSelector.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import * as styles_any from './styles.module.scss';
+import * as _ from 'lodash';
 import ReactSelect from 'react-select';
 import {observer} from "mobx-react";
 import {computed} from 'mobx';
@@ -8,6 +9,7 @@ import {QueryStore, QueryStoreComponent, CUSTOM_CASE_LIST_ID} from "./QueryStore
 import {getStudyViewUrl} from "../../api/urls";
 import DefaultTooltip from "../DefaultTooltip";
 import SectionHeader from "../sectionHeader/SectionHeader";
+import {ReactSelectOption} from "react-select";
 
 const styles = styles_any as {
 	CaseSetSelector: string,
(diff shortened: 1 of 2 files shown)


src/shared/components/query/CaseSetSelector.tsx
```diff
@@ -68,7 +76,7 @@ export default class CaseSetSelector extends QueryStoreComponent<{}, {}>
 				<ReactSelect
 					value={this.store.selectedSampleListId}
 					options={this.caseSetOptions}
-					clearable={this.store.selectedSampleListId != this.store.defaultSelectedSampleListId}
+					filterOption={(opt: ReactSelectOptionWithName, filter)=>_.includes(opt.textLabel.toLowerCase(), filter)}
```
alisman: Up to discussion whether we should write a unit test for something like this.   It seems to me that a unit test should: 
1. validate internal logic
2. protect against changes to external code on which this depends.  
For #1 there really isn't any logic here.  For #2 the only external dep here (that is our code) is existnence of textLabel, which is potected by TS.  discuss

adamabeshouse: I think it's hard to write unit tests with external libraries. Almost by definition it has to be an integration test because you cant reach inside it.

adamabeshouse: can you add a typing to `filter`?

adamabeshouse: I actually do think, strictly speaking, a unit test would be useful here - pull the lambda from filterOption into a function that can be unit tested. That's good React practice anyway, since with a lambda as a prop it will rerender every time even if the props didn't really change.

alisman:
@adamabeshouse fyi, i responded to your request here.  needed to merge to get this out though.

PR #442, opened 2017-07-03
Add UnsupportedBrowserModal.

Tamba32:
# What? Why?

Add browser/browser version detection so error message can be provided if the browser being used is not IE11, Chrome, or Firefox.


src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+    // let wrapper: any;
+    //
+    // before(()=> {
+    //
+    //      wrapper = shallow(<UnsupportedBrowserModal/>);
+    //
+    // });
+
+    it('does not show modal when Chrome is being used', () => {
```
inodb: @Tamba32 The tests are actually run in PhantomJS, so this is currently not what is being tested. It is rather `shows modal when PhantomJS is being used`. But yeah would be good to add some tests where you specifically set the name of the browser, to verify that it works.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+
+describe('UnsupportedBrowserModal', () => {
+
+    let wrapper: any;
+
+    before(()=> {
+         wrapper = shallow(<UnsupportedBrowserModal/>);
+    });
+
+    it.only('does not show modal when Chrome is being used', () => {
```
inodb: @Tamba32 Maybe update this text? It is using PhantomJS, not chrome.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+  public render() {
+    return (
+      <Modal show={this.state.show} onHide={() => this.setState({show: false})}>
+        <Modal.Header closeButton>
+          <Modal.Title>Sorry, we do not support your browser version!</Modal.Title>
+        </Modal.Header>
+        <Modal.Body>
+          <ul>
+            <li>It looks like your using an old version of Internet Explorer.</li>
+            <li>Please consider updating to IE 11 or using a different browser.</li>
```
inodb: This text is also shown when using another unsupported browser that is not IE. Maybe better to change body accordingly? That is, if the browser is `ie`, show this text. If it is something else, just mention the browser is unsupported and one should use Chrome, Firefox or latest IE.

inodb: @Tamba32 also when i use IE 11 from browserstack it is actually saying I should update to version 11. But that is what I am using?

Tamba32:
[(3) Add UnsupportedBrowserModal.](https://trello.com/c/TY3vphC3/294-3-add-unsupportedbrowsermodal)

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.spec.tsx
Tamba32: Resolved

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+  componentDidMount() {
+      this.handleUnsupportedBrowsers(this.state.name, this.state.version);
+  }
+
+  handleUnsupportedBrowsers(name:string, version: string) {
+      name = name.toLowerCase();
+      const localStorage = window.localStorage.browserError || false;
+      if (localStorage === true) {
+          this.setState({show: false});
+      } else if (String(name) === 'unsupported') {
```
alisman: not sure if i understand this.  home come name would be unsupported?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+          window.localStorage.browserError = true;
+          this.setState({show: true});
+      } else if (!(name === "chrome" || name === "firefox" || name === "edge") || (String(name) === "ie" && Number(version.slice(0,2)) < 11)) {
+          window.localStorage.browserError = true;
+          this.setState({show: true});
+      } else {
+          this.setState({show: false});
```
alisman: how come we should NOT show the error but set error in localstorage to true?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -0,0 +1,68 @@
...
+      }
+  }
+
+  componentDidMount() {
+      this.handleUnsupportedBrowsers(this.state.name, this.state.version);
+  }
+
+  handleUnsupportedBrowsers(name:string, version: string) {
+      name = name.toLowerCase();
+      const localStorage = window.localStorage.browserError || false;
```
alisman: not sure why you're involving local storage?  is this an optimization to prevent doing this check multiple times?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
Tamba32: Ino told me to add it in the case that `detect-browser` could not detect the browser. Up in the constructor, if `browser` is undefined, I set the browser state to be `unsupported`.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
Tamba32: Because I thought you said we only wanted to do this check on the first occurrence, meaning that the modal should not sure, regardless of the browser or its version, after the site first loads.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
Tamba32: I thought you said it would be best to only do the check once and to not display the modal after the first load.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
alisman: ok, sounds good.  good place for a comment, but no change necessary now.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
alisman: gotcha.  i would advise window.sessionStorage instead because it would get cleared on browser closing.  wherease localStorage is essentially forever.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
alisman: ok, i get it.  this makes sense.  see sessionStorage comment above.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.spec.tsx
inodb: @Tamba32 This still has `only` parameters, need to remove those (you can see in the code coverage report that everything turns red)

PR #443, opened 2017-07-05
Put data in the columns "Tumor" and "Allele Frequency" when downloading data

oplantalech:
Patients with multiple samples have the column `Tumor` which displays the sample number(s) where the alteration occurs. However, this column, and the column `Allele Frequency`, are empty in the exported tables. This PR fixes that by creating extra rows for those mutations. Also, the "Allele Counts" column have been rearranged. This changes do not alter the visualized tables in the Patient View.

For this table ([link to the Patient View](http://www.cbioportal.org/beta/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04#/patient?caseId=P04&studyId=lgg_ucsf_2014)):
captura de pantalla 2017-07-21 a les 9 43 13

This is (part of) the download table:
captura de pantalla 2017-07-21 a les 9 53 00

### DIFF
--- a/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
+++ b/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
@@ -60,6 +60,7 @@ export default class CopyNumberTableWrapper extends React.Component<ICopyNumberT
                 name: "Tumors",
                 render:(d:DiscreteCopyNumberData[])=>TumorColumnFormatter.renderFunction(d, this.props.sampleManager),
                 sortBy:(d:DiscreteCopyNumberData[])=>TumorColumnFormatter.getSortValue(d, this.props.sampleManager),
+                download: (d:DiscreteCopyNumberData[])=>TumorColumnFormatter.getSample(d),
                 order: 20
             });
         }
--- a/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.spec.tsx
+++ b/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.spec.tsx
@@ -1,4 +1,7 @@
 import AlleleFreqColumnFormatter from './AlleleFreqColumnFormatter';
+import {GENETIC_PROFILE_UNCALLED_MUTATIONS_SUFFIX, GENETIC_PROFILE_MUTATIONS_SUFFIX} from '../../../../shared/constants';
+import {initMutation} from "test/MutationMockUtils";
+import {Mutation} from "shared/api/generated/CBioPortalAPI";
 import {MOLECULAR_PROFILE_UNCALLED_MUTATIONS_SUFFIX, MOLECULAR_PROFILE_MUTATIONS_SUFFIX} from '../../../../shared/constants';
 import React from 'react';
 import { assert } from 'chai';
@@ -51,5 +54,12 @@ describe('AlleleFreqColumnFormatter', () => {
         };
         const res = AlleleFreqColumnFormatter.convertMutationToSampleElement(uncalledMutationWithSupport, 'red', 5, {});
         assert(res && mount(res.text).text().indexOf('(uncalled)') !== -1);
+    });
+    it('calculates the Allele frequency for a mutation', ()=> {
+        const mutation:Mutation[] = [initMutation({
+            tumorAltCount:1,
+            tumorRefCount:10
+        })];
+        assert.equal(AlleleFreqColumnFormatter.getFrequency(mutation), "0.09090909090909091");
     });
 });
--- a/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
+++ b/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
@@ -156,4 +156,21 @@ export default class AlleleFreqColumnFormatter {
 
         return false;
     }
+
+
+    public static getFrequency(data:Mutation[]): string|string[] {
+        let result = [];
+        if (data) {
+            for (let mutation of data) {
+                let frequency = mutation.tumorAltCount/(mutation.tumorAltCount+mutation.tumorRefCount);
+                if (frequency) {
+                    result.push(String(frequency));
+                }
+            }
+        }
+        if (result.length == 1) {
+            return result[0];
+        }
+        return result;
+    }
 }
--- a/src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
+++ b/src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
@@ -3,7 +3,6 @@ import 'rc-tooltip/assets/bootstrap_white.css';
 import SampleManager from "../../sampleManager";
 import {isUncalled} from 'shared/lib/MutationUtils';
 
-
 export default class TumorColumnFormatter {
 
     public static renderFunction<T extends {sampleId:string}>(data:T[], sampleManager:SampleManager|null) {
@@ -66,5 +65,18 @@ export default class TumorColumnFormatter {
             }
             return map;
         }, {});
+    }
+
+    public static getSample(data:Array<{sampleId:string}>): string|string[] {
+        let result: string[] =[];
+        if (data) {
+            data.forEach((datum:{sampleId:string}) => {
+                result.push(datum.sampleId);
+            })
+        }
+        if (result.length == 1) {
+            return result[0];
+        }
+        return result;
     }
 }
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
@@ -151,20 +151,73 @@ describe('LazyMobXTable', ()=>{
         invisibleString:"FILTER4"
     };
     let data = [datum0, datum1, datum2, datum3, datum4];
+    let multiData = [[datum0, datum1], [datum2, datum3], datum4];
     let sortedList:any[];
+    function downloadName(data:{name:string, num:number, str:string, numList:number[], strList:string[],
+        invisibleString:string}|{name:string, num:number, str:string, numList:number[], 
+        strList:string[], invisibleString:string}[]): string|string[] {
+            let result =[];
+            if (data instanceof Array) {
+                for (let datum of data) {
+                    result.push(datum.name);
+                }
+            } else {
+                return data.name;
+            }
+            return result;
+        }
+    function downloadNumber(data:{name:string, num:number, str:string, numList:number[], strList:string[],
+        invisibleString:string}|{name:string, num:number, str:string, numList:number[], 
+        strList:string[], invisibleString:string}[]): string|string[] {
+            let result =[];
+            if (data instanceof Array) {
+                for (let datum of data) {
+                    result.push(datum.num+'');
+                }
+            } else {
+                return data.num+'';
+            }
+            return result;
+        }
+    function downloadString(data:{name:string, num:number, str:string, numList:number[], strList:string[],
+        invisibleString:string}|{name:string, num:number, str:string, numList:number[], 
+        strList:string[], invisibleString:string}[]): string|string[] {
+            let result =[];
+            if (data instanceof Array) {
+                for (let datum of data) {
+                    result.push(datum.str+'');
+                }
+            } else {
+                return data.str+'';
+            }
+            return result;
+        }
+    function downloadInvisible(data:{name:string, num:number, str:string, numList:number[], strList:string[],
+        invisibleString:string}|{name:string, num:number, str:string, numList:number[], 
+        strList:string[], invisibleString:string}[]): string|string[] {
+            let result =[];
+            if (data instanceof Array) {
+                for (let datum of data) {
+                    result.push(datum.name+"HELLO123456");
+                }
+            } else {
+                return data.name+"HELLO123456";
+            }
+            return result;
+        }
     let columns:any[] = [{
         name: "Name",
         filter: (d:any,s:string)=>(d.name.indexOf(s) > -1),
         sortBy: (d:any)=>d.name,
         render:(d:any)=>(<span>{d.name}</span>),
-        download:(d:any)=>d.name,
+        download:(d:any)=>downloadName(d),
         tooltip:(<span>Name of the data.</span>),
         align: "right"
     },{
         name: "Number",
         sortBy: (d:any)=>d.num,
         render:(d:any)=>(<span>{d.num}</span>),
-        download:(d:any)=>d.num+'',
+        download:(d:any)=>downloadNumber(d),
         tooltip:(<span>Number of the data.</span>),
         defaultSortDirection: "desc",
         align: "left"
@@ -173,7 +226,7 @@ describe('LazyMobXTable', ()=>{
         filter: (d:any,s:string)=>(d.str && d.str.indexOf(s) > -1),
         sortBy: (d:any)=>d.str,
         render:(d:any)=>(<span>{d.str}</span>),
-        download:(d:any)=>d.str+'',
+        download:(d:any)=>downloadString(d),
         tooltip:(<span>String of the data</span>),
         defaultSortDirection: "desc",
         align: "center"
@@ -186,7 +239,7 @@ describe('LazyMobXTable', ()=>{
         visible: false,
         sortBy: (d:any)=>d.name,
         render:(d:any)=>(<span>{d.name}</span>),
-        download:(d:any)=>d.name+"HELLO123456",
+        download:(d:any)=>downloadInvisible(d),
     },{
         name: "Initially invisible column with no download",
         render:()=>(<span></span>),
@@ -999,7 +1052,6 @@ describe('LazyMobXTable', ()=>{
                 "3,-1,zijxcpo,,3HELLO123456,,\r\n"+
                 "4,90,zkzxc,,4HELLO123456,,\r\n");
         });
-
         it("gives data back in sorted order according to initially selected sort column and direction", ()=>{
             let table = mount(<Table columns={columns} data={data} initialSortColumn="Number" initialSortDirection="asc"/>);
 
@@ -1010,6 +1062,17 @@ describe('LazyMobXTable', ()=>{
                 "1,6,kdfjpo,,1HELLO123456,,\r\n"+
                 "4,90,zkzxc,,4HELLO123456,,\r\n" +
                 "2,null,null,,2HELLO123456,,\r\n");
+        });
+        
+        it("gives data for data with multiple elements", ()=>{
+            let table = mount(<Table columns={columns} data={multiData}/>)
+            assert.deepEqual((table.instance() as LazyMobXTable<any>).getDownloadData(),
+                "Name,Number,String,Number List,Initially invisible column,Initially invisible column with no download,String without filter function\r\n"+
+                "0,0,asdfj,,0HELLO123456,,\r\n"+
+                "1,6,kdfjpo,,1HELLO123456,,\r\n"+
+                "2,null,null,,2HELLO123456,,\r\n"+
+                "3,-1,zijxcpo,,3HELLO123456,,\r\n"+
+                "4,90,zkzxc,,4HELLO123456,,\r\n");
         });
     });
     describe('pagination', ()=>{
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
@@ -27,7 +27,7 @@ export type Column<T> = {
     visible?:boolean;
     sortBy?:((data:T)=>(number|null)) | ((data:T)=>(string|null)) | ((data:T)=>(number|null)[]) | ((data:T)=>(string|null)[]);
     render:(data:T)=>JSX.Element;
-    download?:(data:T)=>string;
+    download?:(data:T)=>string|string[];
     tooltip?:JSX.Element;
     defaultSortDirection?:SortDirection;
     togglable?:boolean;
@@ -135,6 +135,50 @@ export function lazyMobXTableSort<T>(data:T[], metric:SortMetric<T>, ascending:b
     return dataAndValue.map(x=>x.data);
 }
 
+function getListOfEmptyStrings<T>(maxColLength:number): string[] {
+     return Array(maxColLength).join(".").split(".");
+}
+
+function getAsList<T>(data: Array<string|string[]>, maxColLength: number): string[][] {
+    let returnList: string[][] = [];
+    data.forEach((datum: string|string[]) => {
+        if (datum instanceof Array) {
+            if (maxColLength != datum.length) {
+                throw new Error('Not all the arrays returned from the download functions are from the same length.');
+            }
+            returnList.push(datum);
+        } else {
+            let arr: string[] = [];
+            for (var i=0; i<maxColLength; i++) {
+                arr = arr.concat(datum);
+            };
+            returnList.push(arr);
+        }
+    })
+    return returnList;
+}
+
+
+function getDownloadObject<T>(columns: Column<T>[], rowData: T) {
+    let downloadObject: {data: Array<string[]|string>; maxColLength: number} = {data: [], maxColLength: 1};
+    columns.forEach((column:Column<T>) => {
+        if (column.download) {
+            let downloadData = column.download(rowData);
+            if (downloadData instanceof Array) {
+                downloadObject.data.push(downloadData);
+                if (downloadData.length > downloadObject.maxColLength) {
+                    downloadObject.maxColLength = downloadData.length;
+                }
+            } else {
+                downloadObject.data.push(downloadData);
+            }
+        } else {
+            downloadObject.data.push("");
+        }
+    });
+    return downloadObject;
+}
+
 class LazyMobXTableStore<T> {
     @observable public filterString:string;
     @observable private _page:number;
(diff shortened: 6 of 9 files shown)


alisman:
@oplantalech this is a fix for https://github.com/cBioPortal/cbioportal/issues/2672, correct?

oplantalech:
@alisman Yes! Thank you for referencing it!

inodb:
@oplantalech Thanks for submitting! The solution you submitted is different from what @pieterlukasse proposed. I think it would be nicer to have one row per sample if possible, instead of having extra columns for each sample. That would also get around the issue of having this weird `;` separated format for the variant/ref reads.

pieterlukasse:
@inodb before @oplantalech I reviewed the code with him and based on this we decided to go for the solution that is now being presented in this PR. I think either format works and has its advantages / disadvantages. For the "Allele Freq" column, which contains 1 value per sample, one possible future solution would be to also put this in multiple columns (e.g. " Allele Freq (sample 1), Alle Freq (sample 2), etc" ). 

We will also be presenting this to some end users to get their feedback. So the solution could still change based on that. Maybe you can also check with some users on your side?

inodb:
@pieterlukasse Gotcha. I think either users know how to do a groupby using R/python or they don't. For the latter one the proposed format in this PR is prolly better. But for any type of programmatic use I'd prefer the former.

inodb:
@pieterlukasse @oplantalech Another thing is that getting all the samples in the column dropdown seems a little much. Especially if you have over 10 samples. It would be nice to be able to see overlap between samples and sort. I think however we can just have this on the `Tumor` column. I think clicking on that should sort by (1) the number of samples with that mutation and (2) the id of the sample. If we do decide to add the extra column for the samples it's probably better if it is only in the csv

pieterlukasse:
I just saw this PR has not been updated to the latest version. In the latest version the solution is different than initially described. This can be found here: https://github.com/thehyve/cbioportal-frontend/tree/export_patient_tables.  @oplantalech can you archive this old code, and put the new code here? It is a bit confusing to discuss this PR and a different branch at the same time. We should all be looking at the same code.

src/shared/components/mutationTable/column/SampleDownloadColumnFormatter.spec.tsx
```diff
@@ -0,0 +1,79 @@
+import SampleDownloadColumnFormatter from "./SampleDownloadColumnFormatter";
+import {initMutation} from "test/MutationMockUtils";
+import React from 'react';
+import { assert } from 'chai';
+import {shallow, mount, ReactWrapper} from 'enzyme';
+import sinon from 'sinon';
+
+/**
+ * @author Selcuk Onur Sumer
```
onursumer: I don't think I am the author of this completely new file :)

src/shared/components/mutationTable/column/SampleDownloadColumnFormatter.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from 'react';
+import DefaultTooltip from 'shared/components/DefaultTooltip';
+import {Mutation} from "shared/api/generated/CBioPortalAPI";
+import styles from "./sample.module.scss";
+
+/**
+ * @author Selcuk Onur Sumer
```
onursumer: Do we really need to duplicate this file too?

oplantalech:
@onursumer @inodb @alisman @pieterlukasse Sorry for the duplication of the branch. Now the PR is up to date.

src/shared/components/mutationTable/column/SampleDownloadColumnFormatter.spec.tsx
pieterlukasse: @oplantalech if there is a new file, please change `@author`

src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
```diff
@@ -67,4 +66,12 @@ export default class TumorColumnFormatter {
             return map;
         }, {});
     }
+
+    public static getSample(data:{sampleId:string}): string {
+        let result ='';
+        if (data) {
+            result = data.sampleId;
+        }
+        return result;
```
inodb: This is more readable imo as `return data? data.sampleId : '';`

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -17,18 +17,23 @@ import { If } from 'react-if';
 import {SortMetric} from "../../lib/ISortMetric";
 import {IMobXApplicationDataStore, SimpleMobXApplicationDataStore} from "../../lib/IMobXApplicationDataStore";
 
+import {shallow, mount, ReactWrapper} from 'enzyme';
```
alisman: why is this imported here?

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -17,18 +17,23 @@ import { If } from 'react-if';
...
+export type TChild = {};
+
 export type Column<T> = {
     name: string;
     filter?:(data:T, filterString:string, filterStringUpper?:string, filterStringLower?:string)=>boolean;
     visible?:boolean;
     sortBy?:((data:T)=>(number|null)) | ((data:T)=>(string|null)) | ((data:T)=>(number|null)[]) | ((data:T)=>(string|null)[]);
     render:(data:T)=>JSX.Element;
-    download?:(data:T)=>string;
+    download?:(data:T, sample?:TChild)=>string;
```
alisman: it seems we are still referencing sample here.   LazyMobxTable should not be aware of sample.

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.spec.tsx
```diff
@@ -52,4 +54,11 @@ describe('AlleleFreqColumnFormatter', () => {
         const res = AlleleFreqColumnFormatter.convertMutationToSampleElement(uncalledMutationWithSupport, 'red', 5, {});
         assert(res && mount(res.text).text().indexOf('(uncalled)') !== -1);
     });
+    it('calculates the Allele frequency for a mutation', ()=> {
+        const mutation:Mutation = initMutation({
+            tumorAltCount:1,
+            tumorRefCount:10
+        });
+        assert.deepEqual(AlleleFreqColumnFormatter.getFrequency(mutation), "0.09090909090909091");
```
onursumer: Why do we need `deepEqual` for a string comparison?

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -149,40 +149,47 @@ describe('LazyMobXTable', ()=>{
         filter: (d:any,s:string)=>(d.name.indexOf(s) > -1),
         sortBy: (d:any)=>d.name,
         render:(d:any)=>(<span>{d.name}</span>),
-        download:(d:any)=>d.name,
+        download:(d:any, s:any)=>s.name,
+        toDownload: true,
```
onursumer: I don't think we need `toDownload`, instead we can check if `download` is defined or not

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -17,18 +17,23 @@ import { If } from 'react-if';
...
     tooltip?:JSX.Element;
     defaultSortDirection?:SortDirection;
     togglable?:boolean;
+    toDownload?:boolean;
```
onursumer: I don't think we need toDownload, instead we can check if download is defined or not

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -208,29 +213,33 @@ class LazyMobXTableStore<T> {
...
-            this.columns.forEach((column:Column<T>) => {
-                if (column.download) {
-                    rowDownloadData.push(column.download(rowData));
+            if (rowData instanceof Array) {
+                for (let sample of rowData) {
+                    const rowDownloadData:string[] = [];
+                    this.columns.forEach((column:Column<T>) => {
+                        if (column.toDownload) {
+                            if (column.download) {
+                                rowDownloadData.push(column.download(rowData, sample));
```
pieterlukasse: don' t use "sample" , but a generic variable name

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -975,17 +982,17 @@ describe('LazyMobXTable', ()=>{
...
+        //    let table = mount(<Table columns={columns} data={[data]} initialSortColumn="Number" initialSortDirection="asc"/>);
 
-            assert.deepEqual((table.instance() as LazyMobXTable<any>).getDownloadData(),
-                "Name,Number,String,Number List,Initially invisible column,Initially invisible column with no download,String without filter function\r\n"+
-                "3,-1,zijxcpo,,3HELLO123456,,\r\n"+
-                "0,0,asdfj,,0HELLO123456,,\r\n"+
-                "1,6,kdfjpo,,1HELLO123456,,\r\n"+
-                "4,90,zkzxc,,4HELLO123456,,\r\n" +
-                "2,null,null,,2HELLO123456,,\r\n");
-        });
```
pieterlukasse: from @inodb : please add specific test for multiple samples

pieterlukasse:
@oplantalech as @inodb mentioned: please also test other places where the table is used to make sure the new code does not break anything there.

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -149,40 +149,47 @@ describe('LazyMobXTable', ()=>{
...
+        download:(d:any, s:any)=>s.name,
+        toDownload: true,
         tooltip:(<span>Name of the data.</span>)
     },{
         name: "Number",
         sortBy: (d:any)=>d.num,
         render:(d:any)=>(<span>{d.num}</span>),
-        download:(d:any)=>d.num+'',
+        download:(d:any, s:any)=>s.num+'',
+        toDownload: true,
```
pieterlukasse: decision: please remove this "toDownload" feature

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -959,13 +966,13 @@ describe('LazyMobXTable', ()=>{
                 "Name,Number,String,Number List,Initially invisible column,Initially invisible column with no download,String without filter function\r\n");
         });
         it("gives one row of data when theres one row. data given for every column, including hidden, and without download def'n. if no data, gives empty string for that cell.", ()=>{
-            let table = mount(<Table columns={columns} data={[data[0]]}/>);
+            let table = mount(<Table columns={columns} data={[[data[0]]]}/>);
```
pieterlukasse: decision: revert all changes to test

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -208,29 +213,33 @@ class LazyMobXTableStore<T> {
...
+                    });
+                    tableDownloadData.push(rowDownloadData);
                 }
-                else {
-                    rowDownloadData.push("");
-                }
-            });
-
-            tableDownloadData.push(rowDownloadData);
+            }
```
pieterlukasse: decision: use extra "else"  here for when `rowData` is not an Array (which is indeed not always the case).

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -208,29 +213,33 @@ class LazyMobXTableStore<T> {
     @computed public get downloadData()
```
inodb: @onursumer @alisman What are your thought on changing this function this way? Should we override this function in the `MutationTable` and `CNATableComponent`? Or should we pass some function for downloading? I think eventually we want to be able to support multiple download formats from mutation table, e.g. MAF/VCF/TXT. But for now maybe better to support this proposed MAF-like format with one row per sample.

alisman: Hi guys, I think we should pass a function for downloading.  Formatting the data for download seems very specific to the nature of the particular data being displayed and can't be generalized here.   @adamabeshouse would be a good person to discuss this with.   Is it important that the downloaded data reflect ordering and filtering state or can we just deliver it as a dump given that someone will most likely be using another tool to manipulate the data?

inodb: I think use case depends on user. I would reckon that the folks who want an excel style output format, might want to get the filtered state, so they don't have to do any post processing. Bioinformaticians are probably happy with a raw data dump.

oplantalech:
Checking if the changes in this PR break the functionality of the other cBioPortal `LazyMobXTables`:

- Datasets: I could not check it, but columns have the download option set to 'false': 
```
columns={
    [
        {
            name:'Name',
            type: 'name',
            render:(data:IDataTableRow)=> <CancerStudyCell studyId={data.studyId} name={data.name}/>,
            filter:(data:any, filterString:string, filterStringUpper:string) => {
                return data.name.toUpperCase().indexOf(filterStringUpper) > -1;
             }

        },
        {name:'', sortBy:false, togglable:false, download: false, type:'download', render:(data:IDataTableRow)=> {
            const download = this.state.downloadable.indexOf(data.studyId) > -1;
            return (
                <a className="dataset-table-download-link" style={download ? {display:'block'} : {display:'none'}}
                   href={'https://media.githubusercontent.com/media/cBioPortal/datahub/master/public/' + data.studyId + '.tar.gz'} download>
                    <i className='fa fa-download'/>
                </a>
            );
        }},
```

- Clinical information tables: when you click the download button, they are empty in the PR and also in: http://www.cbioportal.org/beta/case.do#/patient?caseId=P04&studyId=lgg_ucsf_2014&tab=clinicalDataTab

- Cosmic tables in Patient View and Result's Mutation tab: no changes (as expected, since there is no download option for those tables).

- Result's Mutation tab: in beta, the column "Sample ID" in the download table is full, while in the PR the "Sample ID" column is empty. However, if I switch to rc, the "Sample ID" columns are still empty. So I believe the code used to fill the "Sample ID" column is outside this repository.

Sjoerd-van-Hagen:
@jjgao Hi JJ, do you think this is ready to be merged soon?

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -60,6 +60,8 @@ export default class CopyNumberTableWrapper extends React.Component<ICopyNumberT
                 name: "Tumors",
                 render:(d:DiscreteCopyNumberData[])=>TumorColumnFormatter.renderFunction(d, this.props.sampleManager),
                 sortBy:(d:DiscreteCopyNumberData[])=>TumorColumnFormatter.getSortValue(d, this.props.sampleManager),
+                download: (d:DiscreteCopyNumberData[], s:DiscreteCopyNumberData)=>TumorColumnFormatter.getSample(s),
+                toDownload: true,
```
adamabeshouse: what is the purpose of `toDownload`? can it not be inferred from the presence of a value for `download`?

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -142,4 +142,15 @@ export default class AlleleFreqColumnFormatter {
             return (altReads / (altReads + refReads));
         });
     }
+
+    public static getFrequency(data:Mutation): string {
```
adamabeshouse: more specific name would be good, `getTumorVariantAlleleFrequency` maybe?

src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
```diff
@@ -67,4 +66,12 @@ export default class TumorColumnFormatter {
             return map;
         }, {});
     }
+
+    public static getSample(data:{sampleId:string}): string {
```
adamabeshouse: `getSampleId`, because there are `Sample` objects so this is a bit confusing

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
adamabeshouse: yeah the table component should be completely agnostic about the form of the data, download should take the rowData and the table shouldn't have any opinion on whether its an array or not.

src/shared/components/mutationTable/column/AlleleCountColumnFormatter.tsx
```diff
@@ -48,4 +48,13 @@ export default class AlleleCountColumnFormatter
             </div>
         );
     }
+    
+    public static getReads(sample:Mutation, dataField:string): string
```
adamabeshouse: This isn't well-named, this function could really be summarized as `(sample ? sample[dataField] : '')`. It doesn't have anything to do with Reads

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -25,7 +25,7 @@ export type Column<T> = {
     visible?:boolean;
     sortBy?:((data:T)=>(number|null)) | ((data:T)=>(string|null)) | ((data:T)=>(number|null)[]) | ((data:T)=>(string|null)[]);
     render:(data:T)=>JSX.Element;
-    download?:(data:T)=>string;
+    download?:(data:T, element?:{})=>string;
```
adamabeshouse: I don't see the need for this change

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
oplantalech: @adamabeshouse I think it is not possible to move all the changes I made to the download functions: with this extra loop, we are creating new 'extra' rows (one row per every sample instead of one row per every mutation/alteration). Currently, the `push` function only adds one string to the row, according to the order of the columns. Therefore, if I want to maintain this 'extra' rows but remove the extra loop I introduced, I could make the `download` functions to return arrays instead of strings. However, these arrays should be parsed differently, and this should be done in LazyMobXTable, and I believe that's not what it is desired.

adamabeshouse: I see now, sorry I didn't think about it carefully enough, you're right. I actually think having `download` return an array would be an elegant way to handle it.

pieterlukasse: I think the current solution, which @adamabeshouse summarized well, could be the default behavior. Perhaps other (future) download formats could be implemented by overriding this default function? So by the time we need more specific download formats for the patient view, we could have a new type, e.g. PatientViewTable, which extends LazyMobXTable with its own extra download formats. Is such a scenario feasible with the current code structure?

adamabeshouse: @pieterlukasse actually I don't think what I said is the same - the current solution is, if the `rowData` is an array, loop through it and call `download` on each of its elements, then combine them. What I'm suggesting is, call `download` on `rowData`, always, and deal with whatever it returns.

I think a good way to handle the multiple desired download formats could be: put a prop on `LazyMobXTable` called `download` which takes the result of calling each of the column `download` functions on all the data, and decides how to combine them.

So each row gives an array of either `string` or `string[]`, i.e. `(string|string[])[]`, thus the input to `LazyMobXTable.props.download` would be of type `(string|string[])[][]`

pieterlukasse:
Approval based on code (I did no manual tests).

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -220,20 +249,27 @@ class LazyMobXTableStore<T> {
...
-                    rowDownloadData.push("");
+                    rowDownloadData.push(getListOfEmptyStrings(rowData));
                 }
             });
-
-            tableDownloadData.push(rowDownloadData);
+            
+          //rowDownloadData is list of lists iterate over each element (transpose)
+            let processedRowsDownloadData = rowDownloadData[0].map(function(col:string, i:number) { 
+              return rowDownloadData.map(function(row) {
```
adamabeshouse: row should be named col

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -220,20 +249,27 @@ class LazyMobXTableStore<T> {
...
                 else {
-                    rowDownloadData.push("");
+                    rowDownloadData.push(getListOfEmptyStrings(rowData));
                 }
             });
-
-            tableDownloadData.push(rowDownloadData);
+            
+          //rowDownloadData is list of lists iterate over each element (transpose)
+            let processedRowsDownloadData = rowDownloadData[0].map(function(col:string, i:number) {
```
adamabeshouse: col should be named row

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -220,20 +249,27 @@ class LazyMobXTableStore<T> {
 
         // add rows (including hidden columns)
         this.dataStore.sortedData.forEach((rowData:T) => {
-            const rowDownloadData:string[] = [];
+            const rowDownloadData:any[] = []; //string[]
```
adamabeshouse: rowDownloadData at index i is the output for column i

adamabeshouse: the type is string[][]

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -134,6 +134,35 @@ export function lazyMobXTableSort<T>(data:T[], metric:SortMetric<T>, ascending:b
...
+    } else {
+        return [""];
+    }
+}
+
+function getAsList<T>(rowData: T, data: string|string[]) {
+    let length = 1;
+    if (data instanceof Array) {
+        if (!(rowData instanceof Array) || rowData.length != data.length) {
+            throw new Error('Unexpected error!');
```
adamabeshouse: This is not good because it makes assumptions about the form of rowData. For example, in some applications we don't need the output of download to be the same length as the input rowData.

pieterlukasse: @adamabeshouse we can make it a bit more flexible by allowing the output of download to be either 0, 1 or the same length as rowData. But having output of download be e.g. 2 while rowData is 3 would still be an error, right?

adamabeshouse: I dont see why unique per row means it needs to be the length of rowData. You could end up outputting more than one row from a single rowData, or fewer rows than the length of rowData. That's the functional reason why I see this approach as wrong. From a code point of view, it violates one of the main rules of this library which is to be completely agnostic as to the form of rowData - once we start going down that road, it becomes less maintainable and general.

I don't even think the length needs to be the same across the row. You can just pad missing entries with the empty string, and build a number of rows equal to the max length column output.

pieterlukasse: @adamabeshouse "unique per row" indeed does not necessarily need to be the length of rowData. But the assumption of "same order" will remain. This means that in the example where output of download is length 2 while rowData is 3 we would map the first and second rows to the first and second output of download data while mapping the third row of rowData to some empty string. 

Although I agree this is more flexible, I tend to prefer a more defensive implementation for such parts. I.e. with such a flexible implementation, my question then would be: which part of the code is responsible for checking the integrity of the data? I mean, how do you avoid the scenario where the 2 items of the output of download should actually be bound to rows 2 and 3 instead of 1 and 2? An implementation that is too flexible will just silently map things incorrectly.

Maybe it would be a good idea if you could share your feedback in the form of `code` for this part? Perhaps an alternative PR that achieves the same so we can more easily compare?

adamabeshouse: If the user needs it to go to 2 and 3 then they can pad the beginning with an empty string. As I see it the issue you're bringing up is that if it's used improperly there won't be an easy way to see that, which is valid. In that case, we can, as you say, require that column.download(rowData) is the same length for all columns, without peeking into rowData itself at all.

src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
```diff
@@ -67,4 +66,17 @@ export default class TumorColumnFormatter {
             return map;
         }, {});
     }
+
+    public static getSample(data:Array<{sampleId:string}>): string|string[] {
+        let result =[];
+        if (data) {
+            for (let datum of data) {
```
alisman: i'm pretty sure for ... of is not supported in IE11.  there is a babel tranform for it but i don't think we have it installed and probably not worth it.  i would just use Array.forEach to be safe.  we definately already have a polyfill for that.

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -220,20 +252,28 @@ class LazyMobXTableStore<T> {
 
         // add rows (including hidden columns)
         this.dataStore.sortedData.forEach((rowData:T) => {
-            const rowDownloadData:string[] = [];
+            const rowDownloadData:any[] = []; //string[]
+            let refLength = getRefLength(this.columns, rowData);
```
adamabeshouse: okay it's almost there - the one remaining comment I have on this is the refLength approach is inefficient - you're generating the download data twice for each row. You could equivalently just use the length of the first column download - since we're not padding, we're requiring the same length, there's no need to find the max length to compare against.

adamabeshouse: okay I just realized the other reason for finding the max - duplicating non-array download outputs. I think there should be a way to do this without generating all the download data twice.

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
adamabeshouse: please do rename this, it's confusing

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
adamabeshouse: please rename this, its confusing

oplantalech:
I've squashed all commits into one and I've rebased, but Travis is failing. Apparently some `*-docs.json` are out of sync. @adamabeshouse @alisman @onursumer @inodb do you know how to fix that?

PR #448, opened 2017-07-05
Expose QueryStore::addGenesAndSubmit

adamabeshouse:
New frontend code to fix https://github.com/cBioPortal/cbioportal/issues/2673

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -23,11 +23,16 @@ import {QuerySession} from "../../lib/QuerySession";
 import {stringListToIndexSet, stringListToSet} from "../../lib/StringUtils";
 import chunkMapReduce from "shared/lib/chunkMapReduce";
 import formSubmit from "shared/lib/formSubmit";
-import {urlParams} from "./QueryStoreUtils";
+import {
+	GeneticProfileQueryParams, NonGeneticProfileQueryParams, queryUrl,
+	nonGeneticProfileParams, currentQueryParams, geneticProfileParams, queryParams
+} from "./QueryStoreUtils";
+import onMobxPromise from "shared/lib/onMobxPromise";
 
 // interface for communicating
 export type CancerStudyQueryUrlParams = {
 	cancer_study_id: string,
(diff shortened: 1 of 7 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -728,47 +744,52 @@ export class QueryStore
...
-			// Remove params with no value, because they may cause problems.
-			// For example, the server will always transpose if transpose_matrix is present, no matter the value.
-			for (let key in params)
-				if (!(params as any)[key])
-					delete (params as any)[key];
+		// Remove params with no value, because they may cause problems.
+		// For example, the server will always transpose if transpose_matrix is present, no matter the value.
+		for (let key in params)
+			if (!(params as any)[key])
+				delete (params as any)[key];
```
onursumer: Since we are already fixing the indentation here, can we also add the brackets for `for` and `if` blocks?

src/shared/components/query/QueryStore.ts
```diff
@@ -728,47 +744,52 @@ export class QueryStore
...
+		}
+
+		return {pathname: 'index.do', query: params};
+	}
 
-			return {pathname: 'index.do', query: params};
+	readonly asyncUrlParams = remoteData({
+		await: () => [this.asyncCustomCaseSet, this.geneticProfiles],
+		invoke: async () => {
+			return this.getUrlParams(this);
```
onursumer: `this` looks confusing :) Shouldn't this be `this.getUrlParams(this.initialParams)` or something like that?

adamabeshouse: nope, initialParams is only in addGenesToQuery, if you look at the type of initialParams, its a Pick type of QueryStore :)

adamabeshouse: but you're right that's redundant, I'll make the argument optional and default to `this`

inodb:
@adamabeshouse seems like mobxpromisehandler is missing? https://travis-ci.org/cBioPortal/cbioportal-frontend/jobs/250571354#L303

PR #449, opened 2017-07-06
Percentage display value fix for cohort tooltip

onursumer:
cohort_tooltip

# What? Why?
Fixes cBioPortal/cbioportal/issues/2691

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/cohort/FrequencyBar.tsx
+++ b/src/shared/components/cohort/FrequencyBar.tsx
@@ -1,10 +1,12 @@
 import * as React from 'react';
 import DefaultTooltip from 'shared/components/DefaultTooltip';
+import {getPercentage} from "shared/lib/FormatUtils";
 
 export interface IFrequencyBarProps
 {
     totalCount: number;
     counts: number[];
+    mainCountIndex?: number;
     freqColors?: string[];
     barColor?: string;
     barWidth?: number;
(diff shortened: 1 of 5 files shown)


src/shared/components/mutationTable/column/CohortColumnFormatter.tsx
```diff
@@ -104,11 +104,16 @@ export default class CohortColumnFormatter {
         );
     }
 
-    private static getBoldPercentage(proportion:number) {
+    public static getBoldPercentage(proportion: number)
```
onursumer: we may also consider moving this function into a general purpose util class...

adamabeshouse: not a bad idea at all

onursumer: okay, now we have new functions in `FormatUtils`!

src/shared/lib/FormatUtils.ts
```diff
@@ -13,3 +15,33 @@ export function toPrecision(value:number, precision:number, threshold:number)
...
 }
+
+/**
+ * Works the same as the default toFixed() function when the result string is not expected to be a zero value
+ * such as 0.0, 0.00, 0.000, etc.
+ *
+ * If the default toFixed() function returns a zero value, then converts it to an inequality such as
+ * <0.1, <0.01, <0.001, etc.
+ */
+export function toFixedWithThreshold(value: number, digits: number): string
```
onursumer: `toFixedWithThreshold` may not be the best name for this function

adamabeshouse: maybe just call it `toFixed` and have parameter `minBucket` ? or something like that

adamabeshouse: `toFixedWithThreshold` is fine with me too, but I think you should definitely make 0.1 a parameter

onursumer: `0.1` is dynamic actually. If the `digits` is 1 then it returns `<0.1`, so it is still "fixed" to one digit. If it is 2, returns `<0.01`, etc. This way, we still respect the `digits` parameter.

src/shared/components/cohort/FrequencyBar.tsx
```diff
@@ -63,7 +74,7 @@ export default class FrequencyBar extends React.Component<IFrequencyBarProps, {}
                     textAnchor="start"
                     fontSize="10"
                 >
-                    {`${(100 * mainProportion).toFixed(1)}%`}
+                    {getPercentage(mainProportion)}
```
onursumer: For consistency, using `getPercentage` function for the frequency bar as well...

src/shared/lib/FormatUtils.ts
adamabeshouse: oh I see. I think sometimes you may want to specify a different threshold than `10^-(#digits)` though

PR #451, opened 2017-07-06
Only allow dataset table download if data is in datahub

Tamba32:
# What? Why?
For each study, download icon no longer shows if the study is not in the datahub.

Resolves https://github.com/cBioPortal/cbioportal/issues/2703

### DIFF
--- a/src/pages/datasetView/DatasetList.tsx
+++ b/src/pages/datasetView/DatasetList.tsx
@@ -1,6 +1,5 @@
 import * as React from "react";
 import * as _ from 'lodash';
-import $ from 'jquery';
 import {CancerStudy} from 'shared/api/generated/CBioPortalAPI';
 import {ThreeBounce} from 'better-react-spinkit';
 import request from 'superagent';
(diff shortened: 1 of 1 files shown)


Tamba32:
On second thought, I should change some of the code. Currently, I just have some of this code from the legacy project: https://github.com/cBioPortal/cbioportal/pull/1767/files. But I think I should try to remove the jquery and replace it will vanilla js. So I will update this PR.

inodb:
@Tamba32 what about the jQuery here? Should we still update this?

src/pages/datasetView/DatasetList.tsx
```diff
@@ -1,7 +1,9 @@
 import * as React from "react";
 import * as _ from 'lodash';
+import $ from 'jquery';
```
inodb: @Tamba32 can u remove the unused jquery import here?

Tamba32: Done

PR #466, opened 2017-07-12
Improved OncoKB status handling for both patient view and results view

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2725.

- Now passing both status and data for OncoKB for better error handling at component level

# Checks
  the `-s` option)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -336,7 +335,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                         myCancerGenomeData={patientViewPageStore.myCancerGenomeData}
                                         hotspots={patientViewPageStore.indexedHotspotData}
                                         cosmicData={patientViewPageStore.cosmicData.result}
-                                        oncoKbData={patientViewPageStore.oncoKbData.error ? ONCOKB_ERROR : patientViewPageStore.oncoKbData.result}
+                                        oncoKbData={patientViewPageStore.oncoKbData}
```
onursumer: Now we are passing `oncoKbData` (which also includes the status) instead of just `oncoKbData.result`

src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -154,7 +154,10 @@ export class MutationMapperStore {
         await: () => [
             this.mutationData
         ],
-        invoke: async () => fetchOncoKbData(this.sampleIdToTumorType, this.mutationData)
+        invoke: async () => fetchOncoKbData(this.sampleIdToTumorType, this.mutationData),
+        onError: (err: Error) => {
+            // fail silently, leave the error handling responsibility to the data consumer
```
onursumer: In case of an OncoKB fetch error, we are showing the error icon within the Annotation column now, instead of a pop up "Sorry something went wrong" error message

src/shared/components/annotation/OncoKB.tsx
```diff
@@ -19,7 +19,7 @@ import OncokbPubMedCache from "shared/cache/PubMedCache";
 import {default as TableCellStatusIndicator, TableCellStatus} from "shared/components/TableCellStatus";
 
 export interface IOncoKbProps {
-    status: "complete" | "error" | "loading";
+    status: "pending" | "error" | "complete";
```
onursumer: renamed `loading` to `pending` to be consistent with `MobXPromise`

src/shared/components/mutationTable/column/AnnotationColumnFormatter.tsx
```diff
@@ -57,31 +57,10 @@ export default class AnnotationColumnFormatter
         };
     }
 
-    /**
-     * Derives status from data content.
-     * Default (empty) data means loading. Null data means error.
-     */
-    public static getOncoKbStatus(oncoKbData?: IOncoKbData): "complete"|"loading"|"error"
```
onursumer: We don't need to derive status from data anymore, since we are passing the complete `MobXPromise` object to the components

src/shared/model/OncoKB.ts
```diff
@@ -28,3 +28,8 @@ export interface IOncoKbData {
     indicatorMap: {[id:string]: IndicatorQueryResp} | null;
     sampleToTumorMap: {[sampleId:string]: string} | null;
 }
+
+export interface IOncoKbDataWrapper {
```
onursumer: Actually this is a simplified version of `MobXPromise` type. The only reason that I added this type is that I don't want lower level components to have dependency on the actual `MobXPromise` type.

This may not be the best name for this type though. Any other suggestions are welcome.

PR #469, opened 2017-07-13
Fixed initial sorting for the results view mutation table

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2736.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/IMobXApplicationDataStore.spec.ts
+++ b/src/shared/lib/IMobXApplicationDataStore.spec.ts
@@ -26,7 +26,7 @@ describe('SimpleMobXApplicationDataStore', ()=>{
         assert.deepEqual(store.tableData, store.sortedFilteredData, "with no selected data, sortedFilteredData is the tableData");
 
         assert.equal(store.filterString, "", "initial filter string is empty");
-        assert.isTrue(store.sortAscending, "initially sort ascending");
+        assert.isUndefined(store.sortAscending, "initially sort ascending if undefined");
         assert.equal(store.allData.filter((d:number)=>store.isHighlighted(d)).length, 0, "initially nothing is highlighted");
     });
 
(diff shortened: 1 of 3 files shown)


src/shared/lib/IMobXApplicationDataStore.ts
```diff
@@ -28,14 +28,18 @@ export class SimpleMobXApplicationDataStore<T> implements IMobXApplicationDataSt
...
+    @observable public sortAscending:boolean|undefined;
 
     @computed get allData() {
         return this.data;
     }
     @computed get sortedData() {
-        return lazyMobXTableSort(this.allData, this.sortMetric, this.sortAscending);
+        // if not defined, use default values for sortMetric and sortAscending
+        const sortMetric = this.sortMetric || (() => 0);
+        const sortAscending = this.sortAscending !== undefined ? this.sortAscending : true;
```
onursumer: Actually, another option is to return `allData` when undefined. But then, we need to update several test cases for this class.

src/shared/lib/IMobXApplicationDataStore.spec.ts
```diff
@@ -26,7 +26,7 @@ describe('SimpleMobXApplicationDataStore', ()=>{
         assert.deepEqual(store.tableData, store.sortedFilteredData, "with no selected data, sortedFilteredData is the tableData");
 
         assert.equal(store.filterString, "", "initial filter string is empty");
-        assert.isTrue(store.sortAscending, "initially sort ascending");
+        assert.isUndefined(store.sortAscending, "initially sort ascending is undefined");
```
inodb: `is` should be `if`?

onursumer: good catch :)

PR #470, opened 2017-07-14
Enabled filtering for sample id and cancer type columns

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2741.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/mutationTable/MutationTable.tsx
+++ b/src/shared/components/mutationTable/MutationTable.tsx
@@ -161,6 +161,8 @@ export default class MutationTable<P extends IMutationTableProps> extends React.
             name: "Sample ID",
             render: (d:Mutation[]) => SampleColumnFormatter.renderFunction(d, this.props.studyId),
             sortBy: SampleColumnFormatter.getTextValue,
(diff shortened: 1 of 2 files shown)


src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -405,6 +407,17 @@ export default class MutationTable<P extends IMutationTableProps> extends React.
...
             sortBy:(d:Mutation[])=>CancerTypeColumnFormatter.sortBy(d, this.props.studyId, this.props.cancerTypeCache),
+            filter:(d:Mutation[], filterString:string, filterStringUpper:string) => {
+                if (this.props.cancerTypeCache) {
+                    return CancerTypeColumnFormatter.filter(d,
+                        filterStringUpper,
+                        this.props.studyId,
+                        this.props.cancerTypeCache,
+                        this.props.studyToCancerType);
+                } else {
+                    return false;
```
adamabeshouse: at some point maybe we should implement a way to say that no filtering should be done on this column if the cache is missing. maybe return undefined or something

PR #471, opened 2017-07-16
Fix infinite pulling of source code

inodb:
Possibly related to this: https://github.com/facebook/react/issues/5591

When using `localDev = true` trick on the mutation mapper integration https://github.com/cBioPortal/cbioportal/pull/2713 branch. It infinitely pulls the source code. Seems to be related to the issue of setting state in `render`? Not 100% sure why everything works when uncommenting this line. I guess when the props of the component change they initiate a new call to render, causing the infinite loop. Things seem to work with this change, I'm not entirely sure what this component is supposed to do other than displaying the text of the protein domain or why there are two text elements created, one with reference and one without. Hope @onursumer or @adamabeshouse can clarify? Thanks!


inodb:
This is no longer on issue (only on unit tests, but seems to be a different problem)

PR #472, opened 2017-07-17
Remove characters that give issues on jitpack

inodb:
Remove characters in mutation mapper that result in weird characters when
compiling on jitpack.

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapper.tsx
+++ b/src/pages/resultsView/mutation/MutationMapper.tsx
@@ -99,7 +99,7 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                             <div style={{marginLeft:"45px", marginTop:"5px", marginBottom:"10px"}}>
                                 <ButtonGroup className="pull-right">
                                     <Button className="btn-default" disabled={this.props.store.pdbChainDataStore.allData.length === 0} onClick={this.toggle3dPanel}>
-                                        3D Structure »
+                                        3D Structure <i className="fa fa-angle-double-right" aria-hidden="true"></i>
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -98,9 +98,7 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                             />
                             <div style={{marginLeft:"45px", marginTop:"5px", marginBottom:"10px"}}>
                                 <ButtonGroup className="pull-right">
-                                    <Button className="btn-default" disabled={this.props.store.pdbChainDataStore.allData.length === 0} onClick={this.toggle3dPanel}>
-                                        3D Structure »
-                                    </Button>
+                                    <Button className="btn-default" disabled={this.props.store.pdbChainDataStore.allData.length === 0} onClick={this.toggle3dPanel} dangerouslySetInnerHTML={{ __html:['3D Structure &','raquo;'].join('') }} />
```
onursumer: actually, how about replacing this character with something like this: http://fontawesome.io/icon/angle-double-right/

inodb: @onursumer done. We actually used the same thing here: https://github.com/alisman/cbioportal-frontend/commit/5f2280fc4f721775f2715358dfaf9b1d78124b56. But yeah I agree that using awesome font is a lot easier to read

onursumer: Thanks! I must have missed the other one, but we can also update that one later.

PR #484, opened 2017-07-20
Fix issue where tooltip shows up for studies in wrong group.

Tamba32:
# What? Why?
Resolves https://github.com/cBioPortal/cbioportal/issues/2780


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -51,7 +51,6 @@ export default class RightBar extends React.Component<{}, {}> {
                     </li>
                 </ul>
             </div>;
-        console.log(AppConfig.skinRightNavExamplesHTML);
```
inodb: good catch

PR #485, opened 2017-07-20
Cancer type column improvements

onursumer:
# What? Why?
- Fix https://github.com/cBioPortal/cbioportal/issues/2807
- Fix https://github.com/cBioPortal/cbioportal/issues/2809
- Fix https://github.com/cBioPortal/cbioportal/issues/2811

- Cancer Type column is not lazily loaded anymore (https://github.com/cBioPortal/cbioportal/issues/2811)
- Cancer Type column is now using the `sampleIdToTumorType` map, which is defined in the store and also used for OncoKB. (https://github.com/cBioPortal/cbioportal/issues/2809)
- Cancer Type column is now conditionally visible (https://github.com/cBioPortal/cbioportal/issues/2807)

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapper.tsx
+++ b/src/pages/resultsView/mutation/MutationMapper.tsx
@@ -118,6 +118,7 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                                     </span>
                                 </div>)
                             }
+                            <LoadingIndicator isLoading={this.props.store.clinicalDataForSamples.isPending} />
                             {!this.props.store.clinicalDataForSamples.isPending && (
                                 <ResultsViewMutationTable
                                     studyId={this.props.studyId}
--- a/src/pages/resultsView/mutation/MutationMapper.tsx
+++ b/src/pages/resultsView/mutation/MutationMapper.tsx
@@ -29,7 +29,6 @@ export interface IMutationMapperProps {
     store: MutationMapperStore;
     config: IMutationMapperConfig;
     studyId?: string;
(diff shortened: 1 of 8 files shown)


src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -119,26 +118,27 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
-                                cosmicData={this.props.store.cosmicData.result}
-                                oncoKbData={this.props.store.oncoKbData}
-                                civicGenes={this.props.store.civicGenes.result}
-                                civicVariants={this.props.store.civicVariants.result}
-                                enableOncoKb={this.props.config.showOncoKB}
-                                enableHotspot={this.props.config.showHotspot}
-                                enableMyCancerGenome={this.props.config.showMyCancerGenome}
-                                enableCivic={this.props.config.showCivic}
-                            />
+                            {!this.props.store.clinicalDataForSamples.isPending && (
```
onursumer: This may not be an ideal solution, since now we are waiting for the clinical data to complete before starting rendering the table. If we don't wait for the clinical data, then conditional initial visibility setting won't work.

src/shared/components/mutationTable/column/CancerTypeColumnFormatterLazyLoaded.tsx
```diff
@@ -0,0 +1,87 @@
+import * as React from "react";
```
onursumer: Actually, this is the old `CancerTypeColumnFormatter`, just renamed it...

PR #487, opened 2017-07-21
fixed the broken sampleIds data in ResultsViewPageStore

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2818.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -76,7 +76,8 @@ export class ResultsViewPageStore {
     readonly sampleIds = remoteData(async () => {
         // first priority: user provided custom sample list
         if (this.sampleList) {
-            return this.sampleList;
+            // cannot return an observable array directly, need to create a copy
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -76,7 +76,7 @@ export class ResultsViewPageStore {
     readonly sampleIds = remoteData(async () => {
         // first priority: user provided custom sample list
         if (this.sampleList) {
-            return this.sampleList;
+            return this.sampleList.map(sampleId => sampleId);
```
onursumer: Since `this.sampleList` is a mobx observable array, it causes problems if we return it as is. So we need to create a plain copy.

inodb:
@onursumer nice fix! Also good to know that returning observable arrays causes problems

inodb:
Tested it, works!

inodb:
We might want to write an end to end test for thias

PR #490, opened 2017-07-21
Hide Allele Freq column when all NA

ersinciftci:


src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -49,6 +52,8 @@ export default class ResultsViewMutationTable extends MutationTable<IResultsView
         super.generateColumns();
 
         // override default visibility for some columns
+        this._columns[MutationTableColumnType.TUMOR_ALLELE_FREQ].visible = TumorAlleleFreqColumnFormatter.isVisible(
+            (this.props.dataStore as IMobXApplicationDataStore<Mutation[]>).allData as Mutation[][]);
```
onursumer: I don't think you need to do casting in this class. Can you double check? And if that's the case remove the casting and the corresponding import statement?

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -3,6 +3,9 @@ import {observer} from "mobx-react";
 import {
     IMutationTableProps, MutationTableColumnType, default as MutationTable
 } from "shared/components/mutationTable/MutationTable";
+import TumorAlleleFreqColumnFormatter from "../../../shared/components/mutationTable/column/TumorAlleleFreqColumnFormatter";
+import {Mutation} from "../../../shared/api/generated/CBioPortalAPI";
```
onursumer: I think `shared/api/...` is better than `../../../shared/api/...`

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
onursumer: Also, `this.props.dataStore` might be undefined, we need to check that case

PR #493, opened 2017-07-22
Hide Allele Freq column if all values are NA

ersinciftci:


inodb:
@ersinciftci do you have an example? Also would be useful to have a test of this

ersinciftci:
@inodb example URL http://www.cbioportal.org/beta/index.do?session_id=596f5ec4498e5df2e292b9ea&show_samples=false&#mutation_details&
I'll add tests later as I need to learn how to write them at first.

PR #494, opened 2017-07-24
change height of slide viewer

TaliaWise:
changing height of viewer


inodb:
We still have to test this

PR #499, opened 2017-07-26
Fix cbioportal/cbioportal#2830 show more button

inodb:
screen shot 2017-07-25 at 10 05 16 pm

http://cbioportal-frontend-pr-499.herokuapp.com/#/patient?caseId=P02&navCaseIds=P01%2CP02%2CP04%2CP05%2CP06%2CP07%2CP08%2CP09%2CP10%2CP11%2CP12%2CP13%2CP15%2CP16%2CP17%2CP18%2CP21%2CP24%2CP25%2CP26%2CP27%2CP28%2CP29&studyId=lgg_ucsf_2014

### DIFF
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
@@ -551,7 +551,8 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
             }
         };
         this.getDownloadData = this.getDownloadData.bind(this);
-        this.getToolbar = this.getToolbar.bind(this);
+        this.getTopToolbar = this.getTopToolbar.bind(this);
+        this.getBottomToolbar = this.getBottomToolbar.bind(this);
         this.getTable = this.getTable.bind(this);
         this.getPaginationControls = this.getPaginationControls.bind(this);
         this.filterInputReaction = reaction(
@@ -581,13 +582,14 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
         if (this.props.showPagination) {
             return (
                 <PaginationControls
-                    className="pull-left topPagination"
+                    className="text-center topPagination"
                     itemsPerPage={this.store.itemsPerPage}
                     currentPage={this.store.page}
                     onChangeItemsPerPage={this.handlers.changeItemsPerPage}
                     onPreviousPageClick={this.handlers.decPage}
                     onNextPageClick={this.handlers.incPage}
-                    textBetweenButtons={this.store.paginationStatusText}
+                    textBeforeButtons={this.store.paginationStatusText}
+                    showMoreButton={true}
                     previousPageDisabled={this.store.page === 0}
                     nextPageDisabled={this.store.page === this.store.maxPage}
                     {...this.props.paginationProps}
@@ -598,7 +600,7 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
         }
     }
 
-    private getToolbar() {
+    private getTopToolbar() {
         return (
             <ButtonToolbar style={{marginLeft:0}} className="tableMainToolbar">
                 { this.props.showFilter ? (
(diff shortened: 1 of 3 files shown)


inodb:
Seems to be an existing off by one error with current pagination (you can click next too many times):

screen shot 2017-07-25 at 10 09 43 pm

inodb:
Would also be nice to deal with situation when number of items in table < paginationOpts. Similarly when the show more button reaches its max items (instead of `all`)

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -149,7 +154,29 @@ export class PaginationControls extends React.Component<IPaginationControlsProps
...
+            <div className={classNames(styles.paginationControls, this.props.className)} style={this.props.style}>
+                <span style={{fontSize:12, marginRight:10}}>{this.props.textBeforeButtons}</span>
+                <ButtonGroup bsSize="sm" style={{float:'none'}}>
+                    <If condition={!!this.props.showFirstPage}>
+                        <Button key="firstPageBtn" disabled={!!this.props.firstPageDisabled} onClick={this.props.onFirstPageClick}>
+                            {this.props.firstButtonContent}
+                        </Button>
+                    </If>
+                    <Button className="prevPageBtn" key="prevPageBtn" disabled={!!this.props.previousPageDisabled} onClick={this.props.onPreviousPageClick}>
+                        {this.props.previousButtonContent}
```
inodb: Only this line is new in `render()`. Guess something funky with indentation

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -149,7 +154,29 @@ export class PaginationControls extends React.Component<IPaginationControlsProps
...
+
+    render() {
+        const pageSizeOptionElts = (this.props.itemsPerPageOptions || []).map((opt:number) => (<option key={opt} value={opt+""}>{opt}</option>));
+        if (this.props.showAllOption) {
+            pageSizeOptionElts.push(<option key="all" value={SHOW_ALL_PAGE_SIZE+""}>all</option>);
+        }
 
+        return (
+            <div className={classNames(styles.paginationControls, this.props.className)} style={this.props.style}>
+                <span style={{fontSize:12, marginRight:10}}>{this.props.textBeforeButtons}</span>
```
inodb: And this line is new

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -44,6 +46,7 @@ export class PaginationControls extends React.Component<IPaginationControlsProps
         itemsPerPage: SHOW_ALL_PAGE_SIZE,
         itemsPerPageOptions: [10, 25, 50, 100],
         showAllOption: true,
+        textBeforeButtons:"",
         textBetweenButtons: "text btwn buttons",
```
onursumer: Actually, now we should name this as something like this: `textForShowMoreButton` or `textForMiddleButton`. Now, that button is functional too.

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -602,16 +604,13 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
...
-                    { this.getPaginationControls }
-                </Observer>
                 {this.props.showColumnVisibility ? (
-                    <ColumnVisibilityControls
-                        className="pull-right"
-                        columnVisibility={this.store.colVisProp}
-                        onColumnToggled={this.handlers.visibilityToggle}
-                        {...this.props.columnVisibilityProps}
-                    />) : ""}
+                <ColumnVisibilityControls
```
onursumer: Do we really need this indentation change?

src/shared/components/paginationControls/PaginationControls.tsx
inodb: the show more button text can't be changed atm. So only if `showMoreButton==false`, you have `textBetweenButtons`, so the name still makes sense. We could update it to `textForMiddleButton` and use that to set the `show more` text as well, but it's not really necessary atm imo, since the text is always the same

PR #510, opened 2017-07-27
Displaying empty string for all unknown Mutation Assessor values

onursumer:
mutation_assessor

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2805.

- Now displaying empty string for any other value than "High", "Medium", "Low", and "Neutral"
- Slightly changed the sorting behavior to exclude empty values from sorting

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
@@ -28,22 +28,24 @@ export default class MutationAssessorColumnFormatter
             m: {label: "Medium", className: "oma-medium", priority: 3},
             l: {label: "Low", className: "oma-low", priority: 2},
             n: {label: "Neutral", className: "oma-neutral", priority: 1},
-            na: {label: "NA", className: "oma-na", priority: 0},
+            na: {label: "", className: "oma-na", priority: 0},
         };
     }
 
     public static getSortValue(d:Mutation[]):(number|null)[]
     {
-        // If data is missing, it returns undefined. For the way the table works, we map this to null.
         let score:number|undefined = MutationAssessorColumnFormatter.getData(d).score;
         let returnScore:number|null;
-        if (score === undefined) {
+
+        // If data is missing, it returns undefined. For the way the table works, we map this to null.
+        if (score === undefined || score === 0) {
             returnScore = null;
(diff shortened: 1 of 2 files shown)


src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -260,7 +262,9 @@ export default class MutationAssessorColumnFormatter
         );
 
         // add tooltip for valid values
-        if (text.length > 0 && text !== NA)
+        if (text.length > 0 &&
```
alisman: redundant.  change NA const to actual string value

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -260,7 +260,9 @@ export default class MutationAssessorColumnFormatter
...
+            text.toLowerCase() !== "na" &&
+            text.toLowerCase().indexOf("not available") === -1)
```
inodb: Can we have one utility function for checking NA values in a string `isNA()` or something?

inodb: There's a whole bunch of them here: https://github.com/cbioportal/cbioportal-frontend/blob/9cd67c047369eeeaf1ec7d7fee13871e8165cc70/src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js#L16-L33. Might be some duplication in importer/backend as well. Main problem is that there are a ton of these values in the db, but those studies have never been reimported, so the improved importer never fixed them. Prolly good to have similar checks in frontend

src/shared/components/mutationTable/column/MutationAssessorColumnFormatter.tsx
```diff
@@ -275,4 +273,12 @@ export default class MutationAssessorColumnFormatter
 
         return content;
     }
+
+    public static isValidValue(value: string) {
+        return (
+            value.length > 0 &&
+            value.toLowerCase() !== "na" &&
+            value.toLowerCase().indexOf("not available") === -1
```
alisman: use /not available|na/i.test(value) instead.  (no change necessary now)

PR #511, opened 2017-07-28
Falling back to study cancer type when no cancer type clinical data exists

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2863.

- Now using the study cancer type for samples that don't have cancer type clinical data

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -14,7 +14,8 @@ import PdbHeaderCache from "shared/cache/PdbHeaderCache";
 import {
     findGeneticProfileIdDiscrete, fetchMyCancerGenomeData,
     fetchDiscreteCNAData, findMutationGeneticProfileId, mergeDiscreteCNAData,
-    fetchSamples, fetchClinicalDataInStudy, generateDataQueryFilter, makeStudyToCancerTypeMap
+    fetchSamples, fetchClinicalDataInStudy, generateDataQueryFilter, makeStudyToCancerTypeMap,
+    fetchSamplesWithoutCancerTypeClinicalData, fetchStudiesForSamplesWithoutCancerTypeClinicalData
 } from "shared/lib/StoreUtils";
 import {MutationMapperStore} from "./mutation/MutationMapperStore";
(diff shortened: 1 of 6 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -106,6 +109,21 @@ export class ResultsViewPageStore {
         invoke: async () => fetchSamples(this.sampleIds, this.studyId)
     }, []);
 
+    readonly samplesWithoutCancerTypeClinicalData = remoteData({
+        await: () => [
+            this.sampleIds,
+            this.clinicalDataForSamples
+        ],
+        invoke: async () => fetchSamplesWithoutCancerTypeClinicalData(this.sampleIds, this.studyId, this.clinicalDataForSamples)
```
onursumer: here we still depend on the store's study id to fetch the samples...

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -122,8 +122,14 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                                     </span>
                                 </div>)
                             }
-                            <LoadingIndicator isLoading={this.props.store.clinicalDataForSamples.isPending} />
-                            {!this.props.store.clinicalDataForSamples.isPending && (
+                            <LoadingIndicator
+                                isLoading={
+                                    this.props.store.clinicalDataForSamples.isPending ||
```
alisman: we should probably hide this fallback thing.  Like this wrap this in another remoteData.  It's just too much to grock.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: is that a problem?

onursumer: not for now, but we will need to redesign it when we want to support multiple studies

adamabeshouse: true. I say just leave it and we'll make those changes all at once

PR #512, opened 2017-07-28
Fix Copy Number column default visibility

ersinciftci:


src/shared/components/mutationTable/column/DiscreteCNAColumnFormatter.tsx
```diff
@@ -127,15 +127,10 @@ export default class DiscreteCNAColumnFormatter {
         }
     }
 
-    public static isVisible(cache:DiscreteCNACache, allMutations?: Mutation[][]): boolean {
+    public static isVisible(cache:DiscreteCNACache): boolean {
```
alisman: a test would be good here

PR #514, opened 2017-07-31
Add mutation rate summary to mutation mapper

inodb:
Determine somatic mutation rate. That is divide number of cases with a somatic
mutation by total number of cases (patients) queried. Calculate germline mutation rate if
there are germline mutations. For the private impact study there is a specific clinical
attribute that indicates whether the patient was screened for germline
mutations `12_245_PARTC_CONSENTED`. Use number of cases that consented as
denominator if that is available (subset of total).

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1,5 +1,5 @@
 import {
-    DiscreteCopyNumberFilter, DiscreteCopyNumberData
+    DiscreteCopyNumberFilter, DiscreteCopyNumberData, ClinicalData
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
 import {computed, observable, action} from "mobx";
@@ -18,6 +18,8 @@ import {
 } from "shared/lib/StoreUtils";
 import {MutationMapperStore} from "./mutation/MutationMapperStore";
 import AppConfig from "appConfig";
+import * as _ from 'lodash';
+import { IMPACT_GERMLINE_TESTING_CONSENT } from "shared/constants";
 
 export class ResultsViewPageStore {
 
@@ -65,7 +67,9 @@ export class ResultsViewPageStore {
                 this.mutationGeneticProfileId,
                 this.sampleIds,
                 this.clinicalDataForSamples,
-                this.sampleListId);
+                this.sampleListId,
+                this.patientIds,
+                this.mskImpactGermlineConsentedPatientIds);
 
             this.mutationMapperStores[hugoGeneSymbol] = store;
 
@@ -95,8 +99,28 @@ export class ResultsViewPageStore {
         ],
         invoke: () => {
             const clinicalDataSingleStudyFilter = {attributeIds: ["CANCER_TYPE", "CANCER_TYPE_DETAILED"], ids: this.sampleIds.result};
-            return fetchClinicalDataInStudy(this.studyId, clinicalDataSingleStudyFilter)
+            return fetchClinicalDataInStudy(this.studyId, clinicalDataSingleStudyFilter, 'SAMPLE')
         }
+    }, []);
+
+    readonly mskImpactGermlineConsentedPatientIds = remoteData({
+        await: () => [this.patientIds],
+        invoke: async () => {
+            const clinicalDataSingleStudyFilter = {
+                attributeIds: [IMPACT_GERMLINE_TESTING_CONSENT],
+                ids: this.patientIds.result
+            };
+            const clinicalDataResponse = await fetchClinicalDataInStudy(
+                this.studyId, clinicalDataSingleStudyFilter, 'PATIENT'
+            );
+            if (clinicalDataResponse) {
+                return _.uniq(clinicalDataResponse.map(
+                    (cd:ClinicalData) => cd.entityId)
+                );
+            } else {
+                return [];
+            }
+        },
     }, []);
 
     readonly samples = remoteData({
@@ -104,6 +128,13 @@ export class ResultsViewPageStore {
             this.sampleIds
         ],
         invoke: async () => fetchSamples(this.sampleIds, this.studyId)
+    }, []);
+
+    readonly patientIds = remoteData({
+        await: () => [this.samples],
+        invoke: async () => {
+            return _.chain(this.samples.result).map('patientId').uniq().value();
+        },
     }, []);
 
     readonly studies = remoteData({
--- a/src/pages/resultsView/mutation/MutationMapper.tsx
+++ b/src/pages/resultsView/mutation/MutationMapper.tsx
@@ -17,6 +17,7 @@ import LollipopMutationPlot from "../../../shared/components/lollipopMutationPlo
 import ProteinImpactTypePanel from "../../../shared/components/mutationTypePanel/ProteinImpactTypePanel";
 import ProteinChainPanel from "../../../shared/components/proteinChainPanel/ProteinChainPanel";
 import {computed, action, observable} from "mobx";
+import MutationRateSummary from "pages/resultsView/mutation/MutationRateSummary";
 
 export interface IMutationMapperConfig {
     showCivic?: boolean;
@@ -61,6 +62,26 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
         };
     }
 
+    @computed get mutationRateSummary():JSX.Element|null {
+        if (this.props.store.gene.result &&
+            this.props.store.gene.result.length > 0 &&
+            this.props.store.patientIds.result &&
+            this.props.store.patientIds.result.length > 0 &&
+            this.props.store.mutationData.isComplete &&
+            this.props.store.mutationData.result.length > 0 &&
+            this.props.store.mskImpactGermlineConsentedPatientIds.result &&
+            this.props.store.mskImpactGermlineConsentedPatientIds.isComplete) {
+            return <MutationRateSummary
+                        hugoGeneSymbol={this.props.store.gene.result.hugoGeneSymbol}
+                        mutations={this.props.store.mutationData.result}
+                        patientIds={this.props.store.patientIds.result}
+                        mskImpactGermlineConsentedPatientIds={this.props.store.mskImpactGermlineConsentedPatientIds.result}
+                    />
+        } else {
+            return null;
+        }
+    }
+
     public render() {
         return (
             <div>
@@ -82,6 +103,7 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                 {
                     (this.props.store.mutationData.isComplete && this.props.store.gene.result) && (
                         <div>
+                            {this.mutationRateSummary}
                             <LollipopMutationPlot
                                 store={this.props.store}
                                 onXAxisOffset={this.handlers.onXAxisOffset}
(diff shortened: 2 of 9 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -95,17 +98,43 @@ export class ResultsViewPageStore {
...
+            return fetchClinicalDataInStudy(this.studyId, clinicalDataSingleStudyFilter, 'SAMPLE')
         }
     }, []);
 
+    readonly clinicalDataForPatients = remoteData({
+        await: () => [
+            this.patientIds
+        ],
+        invoke: () => {
+            const clinicalDataSingleStudyFilter = {attributeIds: ["12_245_PARTC_CONSENTED"], ids: this.patientIds.result};
```
alisman: lets make this ID a constant and add a comment explaining what it means.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -95,17 +98,43 @@ export class ResultsViewPageStore {
...
+            return fetchClinicalDataInStudy(this.studyId, clinicalDataSingleStudyFilter, 'PATIENT')
+        }
+    }, []);
+
+    readonly mskImpactGermlineConsentedPatientIds = remoteData({
```
alisman: it seems like clinicalDataForPatients is already specific to "consented" because of the filter that is passed.  So I would rename the above and just do this transformation in the response (unless you're trying to reuse the call)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -95,17 +98,43 @@ export class ResultsViewPageStore {
...
+        await: () => [this.clinicalDataForPatients],
+        invoke: async() => {
+            return _.chain(this.clinicalDataForPatients.result).map('entityId').uniq().value();
```
alisman: what does map do when passed a string like this?

inodb: same as `(x) => (x.entityId)`

src/pages/resultsView/ResultsViewPageStore.ts
inodb: Yes, I figured we would probably use other clinical attributes as well on the results page later on, so seems like it would make sense to have two functions

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
         };
     }
 
+    @computed get mutationRateSummary() {
+        if (this.props.store.gene.result &&
```
alisman: i believe that result properties of these will always evaluate as true because the default value is empty array.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+            return '';
+        }
+    }
+
+   private germlineMutationRate(mutations: Mutation[],
+       patientIds: string[],
+       mskImpactGermlineConsentedPatientIds: string[]) {
+
+       const nrCasesGermlineMutation:number =
+           _.chain(this.props.store.mutationData.result)
```
alisman: make this "functional" by passing in the mutationData instead of referencing it on "this".  That way this function would either be static method or even just a function outside of the class.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+           .filter((m:Mutation) => (m.mutationStatus === MUTATION_STATUS_GERMLINE))
+           .map('patientId')
+           .filter((patientId:string) => {
+               // only include patients that constented to germline testing in
+               // mskimpact
+               if (mskImpactGermlineConsentedPatientIds.length > 0) {
+                   return (
+                       mskImpactGermlineConsentedPatientIds.indexOf(patientId)
```
alisman: you can use array.includes method now.  babel-polyfill supplies it when browesr doesn't.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+                       > -1
+                   )
+               } else {
+                   return true;
```
alisman: seems odd that if list of consenting patients is empty, we return true?

src/pages/resultsView/ResultsViewPageStore.ts
alisman: I think we should still rename the clicalDataForPatients to reflect the fact that it's filtered, no? i.e. clinicalDataForConstenting patients?

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+                         [{germlineMutationRateElement}
+                      Somatic Mutation Rate:
+                         <span style={{paddingLeft:2}}>
+                         {this.somaticMutationRate(this.props.store.mutationData.result,
+                                                   this.props.store.patientIds.result).toFixed(1)}%]
+                         </span>
+                </h4>
+            );
+        } else {
+            return '';
```
alisman: can you return null instead of empty string?

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -82,6 +167,7 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                 {
                     (this.props.store.mutationData.isComplete && this.props.store.gene.result) && (
                         <div>
+                            {this.mutationRateSummary}
```
onursumer: Should we make this a component, and move all the related code into that class?

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
         };
     }
 
+    @computed get mutationRateSummary() {
```
alisman: lets put type annotation here.  just helps to know we are getting JSX back

src/shared/lib/StoreUtils.ts
```diff
@@ -118,13 +118,13 @@ export async function fetchClinicalData(clinicalDataMultiStudyFilter:ClinicalDat
 }
 
 export async function fetchClinicalDataInStudy(studyId:string,
-                                               clinicalDataSingleStudyFilter:ClinicalDataSingleStudyFilter,
+                                               clinicalDataSingleStudyFilter:ClinicalDataSingleStudyFilter, clinicalDataType: 'SAMPLE' | 'PATIENT',
```
onursumer: maybe add the new parameter in a new line (to be consistent with the rest of the parameters)

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+        if (this.props.store.gene.result &&
+            this.props.store.patientIds.result &&
+            this.props.store.mutationData.result &&
+            this.props.store.mskImpactGermlineConsentedPatientIds.result) {
+
+            const germlineMutationRate = this.germlineMutationRate(this.props.store.mutationData.result,
+                                                                   this.props.store.patientIds.result,
+                                                                   this.props.store.mskImpactGermlineConsentedPatientIds.result)
+            const germlineMutationRateElement:JSX.Element|null =
+                (this.props.store.mskImpactGermlineConsentedPatientIds.result.length
```
alisman: there's a weird relationship here.  how can germlineMutationRate be above zero if there are no germline consented patients.  i guess same question as as above.  probably just my ignorance.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+                       mskImpactGermlineConsentedPatientIds.indexOf(patientId)
+                       > -1
+                   )
+               } else {
+                   return true;
+               }
+           })
+           .uniq()
+           .value()
+           .length
```
alisman: missing semicolon

src/pages/resultsView/mutation/MutationMapper.tsx
onursumer: Maybe we can also move that function into `MutationUtils`.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -61,6 +64,88 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
...
+       const nrGermlineCases:number = mskImpactGermlineConsentedPatientIds.length > 0?
+           mskImpactGermlineConsentedPatientIds.length : patientIds.length;
+       return nrCasesGermlineMutation * 100.0 / nrGermlineCases;
+   }
+
+    private somaticMutationRate(mutations: Mutation[], patientIds: string[]) {
```
onursumer: I think, we can also make this one a static utility function

src/pages/resultsView/ResultsViewPageStore.ts
inodb: @alisman I imagine you would want to add clinical attributes to this function the more we end up needing? To prevent doing multiple calls in the future. The `clinicalDataForSamples` is similar in that it only gets `CANCER_TYPE` and `CANCER_TYPE_DETAILED`. Not sure how to best indicate

src/pages/resultsView/mutation/MutationMapper.tsx
inodb: @alisman Only impact study uses this consented patients clinical attribute. Everything else that has germline mutations should show germline mutation rate under the assumption that all patients in the study had germline screening.

inodb: Of course to do this right, you should really also take into account which genes/positions were actually screened for germline mutation. An improved implementation would take that into account. Now we're showing germline mutation rate for any gene in the impact study.

src/pages/resultsView/mutation/MutationMapper.tsx
inodb: This filter is actually redundant, since there would not be germline mutations if no consent was given. The idea was that we would consider all patients as having had germline mutation testing if there is no information about their consent (i.e. this study is not impact). Better to remove the filter maybe, it just obfuscates the error if the underlying data is incorrect.

src/pages/resultsView/mutation/MutationMapper.tsx
alisman: lets definately comment this assumption.  this is good example of business logic that is opaque to all but insiders

src/pages/resultsView/mutation/MutationMapper.tsx
onursumer: or how about a separate component for this?

src/pages/resultsView/ResultsViewPageStore.ts
inodb: I'll just remove `clinicalDataForPatients` for now and handle transformation in the response as originally proposed

src/pages/resultsView/mutation/MutationMapper.tsx
inodb: typescript complains about these properties being possibly undefined if not wrapped like this. Only `mutationData` is fine

src/pages/resultsView/mutation/MutationMapper.tsx
inodb: @alisman This actually did not work for me

PR #516, opened 2017-07-31
Change title of bar chart.

Tamba32:
# What? Why?
Change title of bar chart to "Cases by Top 20 Primary Sites".


inodb:
Bit nitpicky but feels more intuitive to me to call this `Cases by Primary Site (Top 20)`. And perhaps  `(Top 20 of x)`

PR #517, opened 2017-08-01
Fix cbioportal/cbioportal#2804

inodb:
Makes changes so pagination is again at the top for cosmic. This is reverting back to how it was before the `showMore` button. Also add `cbioportal-frontend` class to cosmic table. This fixes the styling issues in the parent project. 
screen shot 2017-08-02 at 7 50 50 pm

### DIFF
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
@@ -46,6 +46,8 @@ type LazyMobXTableProps<T> = {
     showCopyDownload?:boolean;
     copyDownloadProps?:ICopyDownloadControlsProps;
     showPagination?:boolean;
+	// used only when showPagination === true (show pagination at bottom otherwise)
+	showPaginationAtTop?:boolean; 
     paginationProps?:IPaginationControlsProps;
     showColumnVisibility?:boolean;
     columnVisibilityProps?:IColumnVisibilityControlsProps;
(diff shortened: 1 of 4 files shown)


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -580,19 +584,30 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
...
+				textBeforeButtons:this.store.paginationStatusText
+			};
+			// override with given paginationProps if they exist
+			if (this.props.paginationProps) {
+				// put status text between button if no show more button
+				if (this.props.paginationProps.showMoreButton === false) {
+					delete paginationProps['textBeforeButtons'];
+					paginationProps['textBetweenButtons'] = this.store.paginationStatusText;
+				}
+				paginationProps = _.extend(paginationProps, this.props.paginationProps)
```
alisman: you can use Object.assign instead of extend, so as to avoid lodash dependency.

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -139,7 +140,8 @@ export class PaginationControls extends React.Component<IPaginationControlsProps
         } else {
             return (<span
                         key="textBetweenButtons"
-                        className={"btn btn-default textBetweenButtons disabled" + styles["default-cursor"]}
+                        className={"btn btn-default textBetweenButtons disabled " + styles["default-cursor"]}
```
alisman: we should be using classnames module for this kind of thing (classnames(class1, class2, class3)

alisman:
This PR involves changing the default behavior of LazyMobxTable.  Are we sure this won't affect other uses of the table?

inodb:
@alisman Yeah I actually changed default behavior before (oops). This is making some changes to make it work again.

I think it is only used here, and these I test:
- [ ] pdb chain table
- [ ] cosmic mutation table (this is the one being fixed)
- [ ] mutation table patient view
- [ ] mutation table query view
- [ ] copy number table patient view

PR #518, opened 2017-08-01
Truncating long protein change values

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2871

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/TruncatedText.tsx
+++ b/src/shared/components/TruncatedText.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {If, Else, Then} from 'react-if';
+import classNames from 'classnames';
 import {observer} from "mobx-react";
 import {computed} from "mobx";
 import DefaultTooltip from "./DefaultTooltip";
@@ -7,6 +7,7 @@ import DefaultTooltip from "./DefaultTooltip";
 export interface ITruncatedTextProps
 {
     text: string;
+    className?: string;
     maxLength?: number; // max allowed length of the text
     buffer?: number;    // buffer length before considering truncating the text
     suffix?: string;    // will be added to the end of the text if truncated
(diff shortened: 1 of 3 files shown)


onursumer:
@adamabeshouse I was thinking the `maxLength` as the maximum number of characters excluding the suffix. Actually there is also a `buffer` property for a similar purpose, but they are not the same thing.

adamabeshouse:
I think in general it would make more sense to do that inside the component, because then you can make most use of the space, but it doesn't matter that much.

src/shared/components/TruncatedText.tsx
```diff
@@ -1,12 +1,13 @@
 import * as React from 'react';
-import {If, Else, Then} from 'react-if';
+import classNames from 'classnames';
```
alisman: not a big deal but in future lets keep it lowercase in keeping with module name

src/shared/components/TruncatedText.tsx
```diff
@@ -45,7 +46,10 @@ export default class TruncatedText extends React.Component<ITruncatedTextProps,
     public render()
     {
         let content = (
-            <span style={{whiteSpace: "nowrap"}}>
+            <span
+                style={{whiteSpace: "nowrap"}}
+                className={classNames(this.props.className)}
```
alisman: is classNames even necessary here?

src/shared/components/TruncatedText.tsx
onursumer: oops, this was a `copy&paste` from another file, apparently we have it like this at many places already.

src/shared/components/TruncatedText.tsx
onursumer: hmm, probably not. I used it to handle the case when className is `undefined`.

onursumer:
@adamabeshouse it might be better to improve `TruncatedText` in another PR, since it has been already used by other formatters.

Not related to this PR, but I also think that `<TruncatedText>{text}</TruncatedText>` looks better than `<TruncatedText text={text} />`.

PR #521, opened 2017-08-01
Add check for cancer type in bar graph.

Tamba32:
# What? Why?
Resolves https://github.com/cBioPortal/cbioportal/issues/2898

### DIFF
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -24,7 +24,8 @@ export default class RightBar extends React.Component<{}, {}> {
     };
 
     private CancerTypeDescendantStudies(cancerList:CancerType[]) {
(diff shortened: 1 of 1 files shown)


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -24,7 +24,8 @@ export default class RightBar extends React.Component<{}, {}> {
     };
 
     private CancerTypeDescendantStudies(cancerList:CancerType[]) {
-        return cancerList.filter(cancer => cancer.name !== "Other" && cancer.name.toLowerCase().indexOf("pancancer") < 0).map((filteredCancer:CancerType) => (
+        return cancerList.filter(cancer => (cancer.cancerTypeId !== "other" &&  cancer.cancerTypeId !== "mixed"))
```
alisman: you don't have to do this here but more readable and flexible method for writing the predicate here would be 
['mixed','other'].includes(cancer.cancerTypeId).

PR #525, opened 2017-08-02
Remove minHeight in DefaultTooltip

adamabeshouse:
NOTE: most of the changes in this PR are just updating the path to DefaultTooltip - because I moved it into a folder that would also house its styles.

### DIFF
--- a/src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
+++ b/src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
@@ -7,7 +7,7 @@ import {getPercentage} from "shared/lib/FormatUtils";
 
 import CopyNumberCountCache from "../../clinicalInformation/CopyNumberCountCache";
 import {CacheData} from "../../../../shared/lib/LazyMobXCache";
-import DefaultTooltip from "../../../../shared/components/DefaultTooltip";
+import DefaultTooltip from "../../../../shared/components/defaultTooltip/DefaultTooltip";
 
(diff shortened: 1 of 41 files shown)


src/shared/components/defaultTooltip/DefaultTooltip.spec.tsx
```diff
@@ -0,0 +1,17 @@
...
+
+describe("DefaultTooltip", ()=>{
+    it("has min-height: initial", ()=>{
+        /*const tooltip = render(<DefaultTooltip visible={true} overlay={<span>HELLO</span>}><span>TARGET</span></DefaultTooltip>);
+        const inner = tooltip.find('div.rc-tooltip-inner');
+        assert.equal(inner.length, 1, "should be exactly one rc-tooltip-inner element");
+        assert.equal(inner.first().css("min-height"), "initial", "min-height: initial");*/
+        // The above test should work, but its failing, it seems, because I'm losing any reference to the actual
+        //  tooltip, because its rendered in the body root and not in the given element. If I print tooltip.html(),
+        //  I only get '<span>TARGET</span>'. I'm not sure what to do about this.
```
adamabeshouse: help @onursumer @alisman

alisman: I don't think this test is necessary.  Am I right this just test that in fact there is a certain css property?

adamabeshouse: yeah... basically I was just trying to add a test that covers the bug, for regression

PR #536, opened 2017-08-04
Show CNA and MUT tracks in patient view regardless of if there is data

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/2865

Question for @schultzn: I have it always showing CNA tracks as well, is that right?

### DIFF
--- a/src/pages/patientView/genomicOverview/Tracks.tsx
+++ b/src/pages/patientView/genomicOverview/Tracks.tsx
@@ -22,9 +22,15 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
     componentDidMount() {
 
         // --- construct params ---
-        let uniqCnasampleIds = _.uniq(_.map(this.props.cnaSegments, 'sampleId'));
-        let uniqMutSampleIds = _.uniq(_.map(this.props.mutations, 'sampleId'));
-        var config = tracksHelper.GenomicOverviewConfig(uniqCnasampleIds.length + uniqMutSampleIds.length, this.props.width);
+        // <old way> only show a track if theres data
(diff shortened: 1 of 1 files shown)


src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -22,9 +22,15 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
...
 
         // --- construct params ---
-        let uniqCnasampleIds = _.uniq(_.map(this.props.cnaSegments, 'sampleId'));
-        let uniqMutSampleIds = _.uniq(_.map(this.props.mutations, 'sampleId'));
-        var config = tracksHelper.GenomicOverviewConfig(uniqCnasampleIds.length + uniqMutSampleIds.length, this.props.width);
+        // <old way> only show a track if theres data
+        //let uniqCnasampleIds = _.uniq(_.map(this.props.cnaSegments, 'sampleId'));
+        //let uniqMutSampleIds = _.uniq(_.map(this.props.mutations, 'sampleId'));
+        //var config = tracksHelper.GenomicOverviewConfig(uniqCnasampleIds.length + uniqMutSampleIds.length, this.props.width);
+        //</old way>
```
onursumer: Should we still keep the old way as an option that can be switched on/off via component props? Might be cleaner than just commenting out.

schultzn:
I can't see all the details of this issue now - but the title seems strange. When there is no data (no seg file or not sequenced), the tracks should not show up. When the sample has a seg file though or the sample has been sequenced, they should show up, even when there are 0 mutations.

PR #537, opened 2017-08-04
Add right margin to right-aligned text in mutation table

adamabeshouse:
image

### DIFF
--- a/src/shared/components/mutationTable/column/styles.module.scss
+++ b/src/shared/components/mutationTable/column/styles.module.scss
@@ -1,0 +1,9 @@
+.cosmic-count {
+  float: right;
+  margin-right: 50%;
+}
+
+.mutation-count {
+  float: right;
+  margin-right: 50%;
+}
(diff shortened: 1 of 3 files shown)


src/shared/components/mutationTable/column/styles.module.scss
```diff
@@ -0,0 +1,9 @@
+.cosmic-count {
```
onursumer: since `mutation-count` and `cosmic-count` styles are the same, maybe unite them as `numeric-column-value` class or something like that?

adamabeshouse: I thought since theres not much duplication it would be better to decouple them in case anything changes. Not opposed to unifying them if you think thats better.

alisman:
I think maybe right aligning the header would be better.  I'll try to look for some more examples
image

onursumer:
@alisman @adamabeshouse right aligned header doesn't look bad!

alisman:
Here is another example.  I think it actually looks quite good.  So, we would have an integer type which would be right aligned
image

adamabeshouse:
@alisman do you want to just push up the code you wrote? I'll close this PR

alisman:
No that was just stuff I faked in browser with jQuery manipulation.  So, no code written : (

onursumer:
@alisman @adamabeshouse should we create another PR and add a `className` or `style` to headers that we want to right align? Recently I added a `headerRender` option for columns, we may need to use that option.

alisman:
Could we either have an alignment property on column.  Or even, isInteger ?

adamabeshouse:
I would vote for `Column.props.align` - one of the most natural things you can want to do with a header, without having to rewrite the render method, is to align it differently.

PR #539, opened 2017-08-07
Visualize ccf and cluster information in mutation table

inodb:
Add clonality in mutation table (depends on cbioportal/cbioportal#2933)


my-index.ejs
```diff
@@ -11,7 +11,8 @@
         // window.enableDarwin = true;
 
         // Set default API in case none is specified in .env
-        __API_ROOT__ = 'cbioportal-rc.herokuapp.com';
+       // __API_ROOT__ = 'cbioportal-rc.herokuapp.com';
+	__API_ROOT__ = 'localhost:8080/cbioportal';
```
alisman: we should restore this to rc

inodb:
i think Avery has a PR for this now, so fine to close this

PR #540, opened 2017-08-07
Fix input:range style in lollipop for IE

alisman:
https://github.com/cBioPortal/cbioportal/issues/2826

In lollipol diagrame the input:range style doesn't look right.

Remove default padding.  Add vertical alignment center. 
Before:
image
After:
image


adamabeshouse:
can you add a screenshot?

PR #544, opened 2017-08-08
3D viewer panel updates

onursumer:
- Added a border around 3Dmol viewer
- Minor style improvements
- Added a button to toggle the panel size

3d_panel_default

3d_panel_large

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/structureViewer/StructureViewer.tsx
+++ b/src/shared/components/structureViewer/StructureViewer.tsx
@@ -11,6 +11,7 @@ export interface IStructureViewerProps extends IStructureVisualizerProps {
     chainId: string;
     bounds: {width: number|string, height: number|string};
     residues?: IResidueSpec[];
+    containerRef?: (div: HTMLDivElement) => void;
 }
 
 @observer
@@ -18,7 +19,6 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
 {
     private _3dMolDiv: HTMLDivElement|undefined;
     private _pdbId: string;
-    private _bounds: {width: number|string, height: number|string};
     private wrapper: StructureVisualizer3D;
 
     public constructor() {
@@ -47,11 +47,10 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
             this.wrapper = new StructureVisualizer3D(this._3dMolDiv, this.props);
             this.wrapper.init(this.props.pdbId, this.props.chainId, this.props.residues);
             this._pdbId = this.props.pdbId;
-            this._bounds = this.props.bounds;
         }
     }
 
-    public componentDidUpdate() {
+    public componentDidUpdate(prevProps: IStructureViewerProps) {
         if (this.wrapper) {
             // if pdbId is updated we need to reload the structure
             if (this.props.pdbId !== this._pdbId) {
@@ -63,7 +62,7 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
                 this.wrapper.updateViewer(this.props.chainId, this.props.residues, this.props);
             }
 
-            if (!_.isEqual(this.props.bounds, this._bounds)) {
+            if (!_.isEqual(this.props.bounds, prevProps.bounds)) {
                 this.wrapper.resize();
             }
         }
@@ -71,5 +70,9 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
 
     private divHandler(div:HTMLDivElement) {
          this._3dMolDiv = div;
+
+         if (this.props.containerRef) {
+             this.props.containerRef(div);
+         }
     }
 }
--- a/src/shared/components/structureViewer/StructureViewer.tsx
+++ b/src/shared/components/structureViewer/StructureViewer.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import * as _ from 'lodash';
 import {observer} from "mobx-react";
 import {
     IStructureVisualizerProps, IResidueSpec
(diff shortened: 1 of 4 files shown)


src/shared/components/structureViewer/StructureViewer.tsx
```diff
@@ -29,7 +33,12 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
         return (
             <div
                 ref={this.divHandler}
-                style={{height: 300}}
+                style={{
+                    height: this.props.bounds.height,
+                    width: this.props.bounds.width,
+                    padding: 0
```
onursumer: `padding: 0` is required to remove padding introduced by `borderedChart` style.

src/shared/components/structureViewer/StructureViewerPanel.tsx
```diff
@@ -369,6 +373,11 @@ export default class StructureViewerPanel extends React.Component<IStructureView
                             style={{marginRight: 5, cursor: "pointer"}}
                         />
                         <i
+                            className={classnames("fa", {"fa-compress": this.isIncreasedSize, "fa-expand": !this.isIncreasedSize})}
```
onursumer: We may want to use different icons here instead of `fa-compress` and `fa-expand`

src/shared/components/structureViewer/StructureViewerPanel.tsx
```diff
@@ -508,6 +530,24 @@ export default class StructureViewerPanel extends React.Component<IStructureView
         }
     }
 
+    @computed get structureViewerBounds(): {width: number|string, height: number|string} {
+        let width: number|string;
+        let height: number|string;
+
+        // if 3Dmol container div is not initialized yet, just set to a default value: width=auto; height=350
+        // otherwise toggle the size
+        if (this.isIncreasedSize) {
```
onursumer: `@computed` does not work as desired if the condition is written this way:
```JavaScript
if (this._3dMolDiv) {
	width = this.isDoubleSize ? this._3dMolDiv.offsetWidth * 2 : this._3dMolDiv.offsetWidth / 2;
	height = ...
} else {
	width = "auto";
	height = 350;
}
```
This is probably because `_3dMolDiv` is not observable.

src/shared/components/structureViewer/structureViewer.module.scss
```diff
@@ -1,10 +1,10 @@
 .main-3d-panel {
-  width: 420px;
+  width: 480px;
```
onursumer: increased the initial default width a little bit

src/shared/components/structureViewer/StructureViewer.tsx
```diff
@@ -53,10 +63,18 @@ export default class StructureViewer extends React.Component<IStructureViewerPro
             else {
                 this.wrapper.updateViewer(this.props.chainId, this.props.residues, this.props);
             }
+
+            if (!_.isEqual(this.props.bounds, this._bounds)) {
```
adamabeshouse: if I understand correctly what's going on here, you can do this without the extra `_bounds` variable by using the argument to `componentDidUpdate(prevProps)`

src/shared/components/structureViewer/StructureViewerPanel.tsx
```diff
@@ -414,8 +422,11 @@ export default class StructureViewerPanel extends React.Component<IStructureView
                             pdbId={this.pdbId}
                             chainId={this.chainId}
                             residues={this.residues}
+                            bounds={this.structureViewerBounds}
+                            containerRef={(div: HTMLDivElement) => {
```
adamabeshouse: its better practice to not use lambdas for rendering because that will cause unnecessary updates, since the anonymous function is recreated and so always different on each pass through. Better to hold onto this, maybe in a `handlers` object.

src/shared/components/structureViewer/StructureViewerPanel.tsx
adamabeshouse: its definitely because `_3dMolDiv` is not observable :) if written like that, on the first passthrough it only encounters `this._3dMolDiv`, `width`, and `height`, thus it never registers as having to do a recomputation, even when `_3dMolDiv` becomes defined.

onursumer: But why does it ignore the observable `isDoubleSize`? Whenever it changes, `computed` should recompute, but it doesn't if you write it this way...

src/shared/components/structureViewer/StructureViewerPanel.tsx
onursumer: good point

src/shared/components/structureViewer/StructureViewer.tsx
onursumer: That makes sense! I don't think we need the `_bounds` anywhere else.

src/shared/components/structureViewer/StructureViewerPanel.tsx
adamabeshouse: mobx registers observables per-execution. So whatever it sees on one execution, is listening for and then when it executes again, it forgets what happened last time and just remembers what its encountered this time.

So the first time the function executes, its before the 3dMolDiv ref is called, so it doesnt see the branch that has `isIncreasedSize` and doesnt register it. In the way you've written it in the PR, it always sees `isIncreasedSize`

onursumer: hmm, good to know. thanks!

alisman:
Do we really need the compress button? I mean, was that requested?

alisman:
@onursumer what status of this?  i think there was a change you wanted to make.

onursumer:
@alisman This is good to go. It was about a minor minimize bug, which I fixed already.

PR #546, opened 2017-08-08
Show more changes

adamabeshouse:
(1) Remove drop menu for page size from LazyMobXTable pagination 
(2) Show more increases until the last element in props.itemsPerPageOptions, then continues to increase in those increments 
(3) Reset button next to Show more 
(4) Spacing between pagination buttons, like in Google Scholar

image

image


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -587,8 +587,10 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
 			let paginationProps:IPaginationControlsProps = {
 				className:"text-center topPagination",
 				itemsPerPage:this.store.itemsPerPage,
+                totalItems:this.store.displayData.length,
```
adamabeshouse: I've tried to fix this indentation, for some reason when I try to change it git doesnt register it as a change and wont let me commit..

PR #547, opened 2017-08-08
Genomic Overview changes

adamabeshouse:
(1) Show MUT track for a sample iff its sequenced, fixing https://github.com/cBioPortal/cbioportal/issues/2865

(2) Position tooltips for right-side percentage and mutation count, so that they dont go off screen
image

@schultzn

### DIFF
--- a/src/pages/patientView/genomicOverview/Tracks.tsx
+++ b/src/pages/patientView/genomicOverview/Tracks.tsx
@@ -10,6 +10,7 @@ interface TracksPropTypes {
     mutations:Array<Mutation>;
     cnaSegments:Array<CopyNumberSeg>;
     sampleManager:SampleManager;
+    sequencedSamples:{[sampleId:string]:boolean};
     width:number;
 }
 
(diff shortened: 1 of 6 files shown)


src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -78,29 +79,27 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
...
-
-                    let comp: any = this.props.sampleManager.getComponentForSample(sample.id);
-
-                    ReactDOM.render(
-                        comp,
-                        $newContainer[0]
-                    );
-                }
+        _.each(sequencedSamples, (sample: ClinicalDataBySampleId) => {
+            var _trackData = _.filter(this.props.mutations, function (_mutObj: any) {
```
adamabeshouse: all the code here is the same except for the top line (the `_.each`) - everythings marked different because by removing the redundant `if`, now that Im computing `sequencedSamples`, the indentation goes left by one.

PR #552, opened 2017-08-09
Bar chart does not show when no appropriate studies exist.

Tamba32:
# What? Why?
Resolves https://github.com/cBioPortal/cbioportal/issues/

image
@alisman ^ what right bar looks like without studies


alisman:
Thanks @Tamba32!  Looks good

PR #555, opened 2017-08-10
Add package that detects browsers on mobile.

Tamba32:


src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import { Modal } from 'react-bootstrap';
-import browser from 'detect-browser';
+const browser = require('bowser');
```
inodb: @Tamba32 Maybe remove `detect-browser` from the dependencies and add `bowser`

Tamba32: @inodb done

inodb: @Tamba32 still missing bowser dependency, right? Surprised tests work fine, is it part of one of the other packages?

alisman: Yeah, i noticed that too.  So weird.

Tamba32: @inodb @alisman Sorry about that. At first, I only read the first part of your comment Ino so I only removed `detect-browser`. I had forgotten to add `bowser`.  Now it should be fixed.

Also, the tests only tests the `handleUnsupportedBrowser` fn.  They do not use `detect-browser` or `bowser`.  If you remember, when I first made this component about 2 months ago, we could not get the tests to work properly with the package.

inodb: @Tamba32 ok, that makes sense. Thanks! I think we should also try to use `import` instead of `require` if possible. Just to keep things consistent, there shouldn't be any difference in performance in this case afaik

PR #556, opened 2017-08-10
use global store in both page specific components and local components

alisman:
We were instantiating the queryStore twice, once for global components and once for page components.  This PR allows components to access global stores.  We should never use globalStores except in the globalComponents file.  Ultimately we won't need these "exposed" components anymore and stores will be passed using provider/inject, as they are now to pages.

https://github.com/cBioPortal/cbioportal/issues/2928

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/appBootstrapper.jsx
+++ b/src/appBootstrapper.jsx
@@ -4,6 +4,7 @@ import { Provider } from 'mobx-react';
 import { hashHistory, createMemoryHistory, Router } from 'react-router';
 import { RouterStore, syncHistoryWithStore  } from 'mobx-react-router';
 import ExtendedRoutingStore from './shared/lib/ExtendedRouterStore';
+import {QueryStore} from "./shared/components/query/QueryStore";
 import {computed, extendObservable} from 'mobx';
 import makeRoutes from './routes';
 import * as _ from 'lodash';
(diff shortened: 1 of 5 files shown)


src/globalComponents.tsx
```diff
@@ -1,33 +1,23 @@
...
 exposeComponentRenderer('renderRightBar', ()=> {
-    return <RightBar store={queryStore}/>;
+    return <RightBar store={GlobalStores.queryStore} />;
 });
 
-exposeComponentRenderer('renderQuerySelectorInModal', ()=><QueryModal store={queryStore} />);
-
 exposeComponentRenderer('renderQuerySelector', (props:{[k:string]:string|boolean|number})=> {
-    return <QueryAndDownloadTabs {...props} store={queryStore} />;
+    (window as any).addGenesAndSubmitQuery = GlobalStores.queryStore.addGenesAndSubmit.bind(GlobalStores.queryStore);
```
alisman: @adamabeshouse I put this inside the renderer because the global store isn't available at module's rune time.  Do you see any problem here?

src/appBootstrapper.jsx
```diff
@@ -4,6 +4,7 @@ import { Provider } from 'mobx-react';
 import { hashHistory, createMemoryHistory, Router } from 'react-router';
```
alisman: Yes, I would really like to make this file typescript, but there are too many complications right now.

src/globalComponents.spec.ts
```diff
@@ -1,9 +1,3 @@
 import { assert } from 'chai';
 import * as React from 'react';
-import "./globalComponents";
 
-describe('globalComponents', ()=>{
```
alisman: @adamabeshouse Here is a test I don't think we need.  I mean, there needs to be a line drawn somewhere.  A good one I've seen is basically, test logic, don't test declaration.   I think exposing a function is declaration.

src/globalComponents.tsx
adamabeshouse: no not at all

src/globalComponents.spec.ts
adamabeshouse: thanks for pointing this out, I didn't notice you took this out. I disagree. Lets talk in person

PR #559, opened 2017-08-11
Using germline specific sample list id instead of 12_245_PARTC_CONSENTED

onursumer:
- tested with actual data
screen shot 2017-08-21 at 2 53 47 pm

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2918.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1,5 +1,5 @@
 import {
-    DiscreteCopyNumberFilter, DiscreteCopyNumberData, ClinicalData
+    DiscreteCopyNumberFilter, DiscreteCopyNumberData
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
 import {computed, observable, action} from "mobx";
@@ -14,13 +14,11 @@ import PdbHeaderCache from "shared/cache/PdbHeaderCache";
 import {
     findGeneticProfileIdDiscrete, fetchMyCancerGenomeData,
     fetchDiscreteCNAData, findMutationGeneticProfileId, mergeDiscreteCNAData,
-    fetchSamples, fetchClinicalDataInStudy, generateDataQueryFilter, makeStudyToCancerTypeMap,
+    fetchSamples, fetchClinicalDataInStudy, generateDataQueryFilter,
     fetchSamplesWithoutCancerTypeClinicalData, fetchStudiesForSamplesWithoutCancerTypeClinicalData
 } from "shared/lib/StoreUtils";
 import {MutationMapperStore} from "./mutation/MutationMapperStore";
 import AppConfig from "appConfig";
-import * as _ from 'lodash';
-import { IMPACT_GERMLINE_TESTING_CONSENT } from "shared/constants";
 
 export class ResultsViewPageStore {
 
@@ -71,8 +69,7 @@ export class ResultsViewPageStore {
                 this.studiesForSamplesWithoutCancerTypeClinicalData,
                 this.samplesWithoutCancerTypeClinicalData,
                 this.sampleListId,
-                this.patientIds,
-                this.mskImpactGermlineConsentedPatientIds);
+                this.germlineConsentedSampleIds);
 
             this.mutationMapperStores[hugoGeneSymbol] = store;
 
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -106,24 +103,19 @@ export class ResultsViewPageStore {
...
+            if (this.germlineSampleListId) {
+                return await client.getAllSampleIdsInSampleListUsingGET({
+                    sampleListId: this.germlineSampleListId
+                });
             } else {
                 return [];
             }
         },
+        onError: () => {
+            // fail silently
```
onursumer: Since for most of the studies we don't have `${this.studyId}_germline` sample list id, most of the times we end up with an API error, and ignore it. I don't think this is the best solution...

onursumer:
Apparently, case sensitive comparison of the string `Germline` was the problem, since the data is sometimes `GERMLINE`, sometimes `germline`, etc. Updated the PR with a case insensitive comparison. It should work as expected now.

inodb:
@onursumer so we just discussed this, but ideally I think we should keep the calculation as it was. Calculating mutation rate per patient instead of per sample (at least in the cases where we don't have the `{study_id}_germline` genetic profile). Also since this is causing some confusion, I think it would be good to add some extra info on mouseover, maybe with a question mark symbol, that explains what the calculation does exactly.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: could you catch the error within the invoke and fall back to 12_245_PARTC_CONSENTED if so?

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -84,10 +84,8 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
                     hugoGeneSymbol={this.props.store.gene.result.hugoGeneSymbol}
                     mutations={this.props.store.mutationData.result}
                     sampleIds={this.props.store.sampleIds.result}
-                    patientIds={this.props.store.patientIds.result}
-                    patientIdsStatus={this.props.store.patientIds.status}
-                    mskImpactGermlineConsentedPatientIds={this.props.store.mskImpactGermlineConsentedPatientIds.result}
-                    mskImpactGermlineConsentedPatientIdsStatus={this.props.store.mskImpactGermlineConsentedPatientIds.status}
+                    germlineConsentedSampleIds={this.props.store.germlineConsentedSampleIds.result}
```
adamabeshouse: maybe just pass in the mobxpromise, thus exposing both result and status?

onursumer: I have mix feelings about this. Does `MutationRateSummary` really need to be aware of `MobXPromise`? Or is it better to just pass 2 separate parameters?

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: I don't think we need to fall back to `12_245_PARTC_CONSENTED`. Patients with the `12_245_PARTC_CONSENTED` info should also be in the `${this.studyId}_germline` sample list.

Here actually, I wanted to point out that for most of the studies (for now studies other than MSK-IMPACT) don't have  `${this.studyId}_germline` (neither they have `12_245_PARTC_CONSENTED`, I think). But since this solution is generic, it queries all studies no matter what, and we got a `not found` error from API and then conclude that there is no such sample list (which should be the correct conclusion). What I am saying is that we depend on an API error to reach that conclusion.

Maybe we specifically check the error type and make sure it is a `not found` error, but not something else?

src/pages/resultsView/mutation/MutationMapper.tsx
adamabeshouse: you could type it as an interface satisfying { result: T, status: ...}, which is a type we could use elsewhere too

onursumer: `MobXPromise` it is then :)

alisman:
@inodb in @onursumer absence, can you drive this PR?

inodb:
Decided to stick to percentage of samples with given mutation, but added an info icon to indicate how the calculation works.

PR #562, opened 2017-08-14
Fix cbioportal/cbioportal#2778 external frontend

inodb:
Fix cBioPortal/cbioportal#2778. Allows hosting cbioportal frontend on given `window.frontendUrl` instead of inside main `cbioportal/cbioportal` instance. See also https://github.com/cBioPortal/cbioportal/pull/2974


alisman:
@inodb we need to make sure this goes out with corresponding backend version, yes?

PR #564, opened 2017-08-15
Added missing getter for dataTypePriorityCode

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2962.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -513,7 +513,8 @@ export class QueryStore
 
 	// DATA TYPE PRIORITY
 
-	private getDataTypePriorityCode(dataTypePriority:{mutation: boolean, cna: boolean}) {
+	private calculateDataTypePriorityCode(dataTypePriority:{mutation: boolean, cna: boolean}): '0'|'1'|'2'
+	{
 		let {mutation, cna} = dataTypePriority;
 		if (mutation && cna)
 			return '0';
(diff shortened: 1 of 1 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -542,6 +543,11 @@ export class QueryStore
 		}
 	}
 
+	get dataTypePriorityCode(): '0'|'1'|'2'
```
adamabeshouse: `@computed`?

onursumer: yea, makes sense!

src/shared/components/query/QueryStore.ts
```diff
@@ -513,7 +513,8 @@ export class QueryStore
 
 	// DATA TYPE PRIORITY
 
-	private getDataTypePriorityCode(dataTypePriority:{mutation: boolean, cna: boolean}) {
+	private getDataTypePriorityCode(dataTypePriority:{mutation: boolean, cna: boolean}): '0'|'1'|'2'
```
alisman: just curious but is there a reason we use strings for integers?

alisman: lets rename this to calculateDataType...

onursumer: good point, no idea, I just added the return type for clarity :)

onursumer: sounds good

PR #565, opened 2017-08-15
Create Mutual Exclusivity Tab component

ersinciftci:


src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+export default class MutualExclusivityTable extends React.Component<IMutualExclusivityTableProps, {}> {
+
+    @observable protected _columns:{[columnEnum:number]:MutualExclusivityTableColumn};
+    @observable data: MutualExclusivity[];
+
+    constructor(props:IMutualExclusivityTableProps)
+    {
+        super(props);
+        this._columns = {};
+        this.data = props.data;
```
alisman: you probably should not assign prop to local property here.  is there a reason to do so?

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.tsx
```diff
@@ -0,0 +1,143 @@
...
+            MutualExclusivityTableColumnType.GENE_B,
+            MutualExclusivityTableColumnType.P_VALUE,
+            MutualExclusivityTableColumnType.LOG_ODDS_RATIO,
+            MutualExclusivityTableColumnType.ASSOCIATION
+        ],
+        initialSortColumn: "p-Value"
+    };
+
+    public setData(data: MutualExclusivity[]) {
+        this.data = data;
```
alisman: this strikes me as a React antipattern.  Any data used by this component should be passed in a prop.   Since data is already assigned to a prop value above, what is purpose of this?

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.tsx
ersinciftci: Removed the local property

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.tsx
ersinciftci: Removed this method, I missed the point of @observer, so I was trying to set the data manually.

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
```diff
@@ -0,0 +1,183 @@
...
+        if (exclusiveLength == 0) {
+            exclusiveCount = <span><b>no</b> gene pair</span>;
+        } else if (exclusiveLength == 1) {
+            exclusiveCount = <span><b>1</b> gene pair</span>;
+        } else {
+            exclusiveCount = <span><b>{exclusiveLength}</b> gene pairs</span>;
+        }
+
+        if (exclusiveLength > 0) {
+            if (significantLength == 0) {
```
alisman: always use ===

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.spec.tsx
```diff
@@ -0,0 +1,59 @@
...
+        it("should return 3.000 for 3", () => {
+            assert.equal(formatLogOddsRatio(3), "3.000");
+        });
+
+        it("should return >3 for 4.32", () => {
+            assert.equal(formatLogOddsRatio(4.32), ">3");
+        });
+
+        it("should return >3 for NaN", () => {
+            assert.equal(formatLogOddsRatio(NaN), ">3");
```
jjgao: When would `NaN` happen? Completely mutually exclusive? How about completely co-occuring (<-3)?

jjgao:
@ersinciftci I know we decided not to do it on Monday but adding a Bonferroni adjusted p-value seems really easy (https://github.com/cBioPortal/cbioportal/issues/2981). Do you think you can quickly add that? Feel free to say no and we can do it later :)

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.spec.tsx
ersinciftci: `NaN` only happens when the sample list is empty, and it's not a valid case anyway. I did this because the old code was doing this, I'll just remove it because it's confusing.

ersinciftci:
@jjgao I think it's better if we do it later, as we will want to compare p-Values in the old site with the p-Values on the beta site once we deploy it. If we do it now we won't be able to do that, and people will think it's a bug. Let's stick to "no new feature" rule :)

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
```diff
@@ -0,0 +1,183 @@
...
+import {observer} from "mobx-react";
+import {Checkbox} from 'react-bootstrap';
+import styles from "./styles.module.scss";
+import {computed, observable} from "mobx";
+import {MutualExclusivity} from "../../../shared/model/MutualExclusivity";
+import Combinatorics from 'js-combinatorics';
+import fisherTest from 'fishertest';
+
+export interface IMutualExclusivityTabProps {
+    isSampleAlteredMap: {[key: string]: boolean[]};
```
alisman: I wonder if we can no longer consider sampleIds unique since we're handling multiple studies?  : (

ersinciftci: that string is hugo gene symbol, I'll add a comment.

alisman: You could even make a type HugoGeneSymbol which is just a proxy for string.

ersinciftci:
screen shot 2017-08-22 at 9 32 10 am

This is how it looks with mock data, by the way.

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
```diff
@@ -0,0 +1,185 @@
...
+    // a mapping from Hugo Gene Symbol to list of booleans,
+    // each element of the list representing the altered status of a sample
+    isSampleAlteredMap: {[key: string]: boolean[]};
+}
+
+export function calculateAssociation(logOddsRatio: number): string {
+    return logOddsRatio > 0 ? "Tendency towards co-occurrence" : "Tendency towards mutual exclusivity";
+}
+
+export function count(valuesA: boolean[], valuesB: boolean[]): [number, number, number, number] {
```
alisman: this is kind of an important function so maybe a more descriptive name would be helpful.  like, countCoocurrences or something?

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
```diff
@@ -0,0 +1,185 @@
...
+        const association = calculateAssociation(logOddsRatio);
+        data.push({geneA, geneB, pValue, logOddsRatio, association});
+    });
+    return data;
+}
+
+export function getFilteredData(data: MutualExclusivity[], mutualExclusivityFilter: boolean, coOccurenceFilter: boolean,
+                                significantPairsFilter: boolean): MutualExclusivity[] {
+
+    return data.filter(mutualExclusivity => {
```
alisman: not neccessary to change but would this be more clearly written as
if (mutualExclusivityFilter && logOddsRatio <=0) return true;
...

jjgao:
@ersinciftci I was suggesting to have an extra column "Adjusted P-value" in addition to "P-value", but I am fine with postponing it.

alisman:
I noticed that when filter text is set in the table and one changes filter checkboxes, the filter text is wiped out.  couldn't quite track down why, so I thought I would ask if this is expected/desired behavior.

ersinciftci:
@jj Oh, I thought you meant to change it, but yeah let's do it later. 
@alisman I noticed that too and I couldn't track down either, I was going to ask if it was fine like that but I forgot.

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
ersinciftci: Updated

src/pages/resultsView/mutualExclusivity/MutualExclusivityTab.tsx
ersinciftci: I thought about it but I find the current way less complicated/more clear.

PR #567, opened 2017-08-16
Now using g2s endpoint for residue alignment and mapping

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2994.

# Changes proposed in this pull request:
- Replaced `PdbAnnotation API` with `g2s API` (except for the PDB header data)
- Added new API generation npm scripts and auto-generated files (those may not be the final version, we may need to update them)

# Know Issues
multiple_residue_mapping
- [FIXED] Currently there are cases where multiple PDB position mappings exist for a single uniprot position (i.e: when you mouse over or select a single position on the lollipop diagram, multiple locations might be highlighted on the 3D structure viewer). We may want to fix this on the API side
- We cannot completely remove  `PdbAnnotation API` yet, because we fetch `PDB header` data from there.
- `ResidueMappingCache` is a very basic implementation where caching is done wrt the full query parameters. We should probably implement a `LazyMobXCache` version of it.

# Checks
  the `-s` option)


onursumer:
@sheridancbio Here is an example query where `1677` is mapped to multiple different locations in different alignments: ```http://g2s.genomenexus.org/api/alignments/uniprot/BRCA1_HUMAN/pdb/4u4a_B/residueMapping?positionList=1265&positionList=1216&positionList=1217&positionList=654&positionList=1756&positionList=1345&positionList=722&positionList=1538&positionList=1779&positionList=1835&positionList=1495&positionList=504&positionList=825&positionList=813&positionList=1676&positionList=1677&positionList=1710&positionList=1395&positionList=519&positionList=1726&positionList=1727&positionList=797&positionList=1815```

src/shared/api/urls.ts
```diff
@@ -65,6 +65,9 @@ export function getOncoKbApiUrl() {
 export function getPdbAnnotationApiUrl() {
     return 'https://cbioportal.mskcc.org/pdb-annotation';
 }
+export function getG2SApiUrl() {
+    return 'http://g2s.genomenexus.org';
```
onursumer: This one will NOT work with secure portal instances, we need to replace this one with the `https` version, or we need to use `proxy` for a temporary workaround.

alisman:
This is ready to go except that api needs to update and then clients regenerated.  
There is a caching problem but it's functional.  Will just load data in duplicate. @jjgao @sheridancbio  will decide when this should be merged/deployed.

jjgao:
Let's merge this once @sheridancbio implemented the https g2s server. We can log the caching and 
 PDB header issues and resolve them later.

PR #572, opened 2017-08-18
Pull Mutation Assessor from Genome Nexus

inodb:
- Use genome nexus (genomenexus.org) to get Mutation Assessor column on patient view page

TODO:
- [x] cache in bulk like mRNA column
- [x] put logo in annotation column (don't sort by that column, so pull lazily)


inodb:
Superseded by https://github.com/cBioPortal/cbioportal-frontend/pull/588

PR #574, opened 2017-08-18
Fix type of unsupported browser modal for new ts version

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
+++ b/src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
@@ -14,23 +14,11 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
     constructor() {
         super();
 
-        let state;
-        if (!browser)  {
-            state = {
-                name: "Unsupported",
-                show: true,
-                version: ""
-            };
-        } else {
-            state = {
-                name: browser.name,
-                show: false,
-                version: browser.version
-            };
-        }
         this.state = {
-            noShowModal: window.localStorage.noShowModal || false,
-            ...state
+            name: browser.name,
+            show: false,
+            version: browser.version.toString(),
+            noShowModal : window.localStorage.noShowModal || false
         };
 
         this.handleCheckboxChange = this.handleCheckboxChange.bind(this);


src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -14,24 +14,21 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
     constructor() {
         super();
 
-        let state;
         if (!browser)  {
```
alisman: i'm confused about this condition.  why would browser ever be undefined?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -14,24 +14,21 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
...
-            state = {
+            this.state = {
                 name: "Unsupported",
```
alisman: seems like at line 64 we are testing a lowercased "unsupported"    is that right?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
Tamba32: @alisman I think I just had this there as a fail safe - if for some reason the package did not work.  We can remove it though

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
Tamba32: @alisman yes. it works either way though because 2 lines before that, we set `name=name.toLowerCase()`. You can change it to lowercase in the state though for better clarity.

alisman: k will do.

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
alisman: yeah, i think unless we know there's really a condition where this would be true we should remove b/c it implies that's some actual function of the module.  like, if we're running in test, there's no browser or something.  or on a certain device.  when really that thing will never be undefined

PR #577, opened 2017-08-21
add default to protein impact type so that 'other' category is correc…

alisman:
Add default to protein impact type so that 'other' category is correctly populated

https://github.com/cBioPortal/cbioportal/issues/3023

image

correct:

image

### DIFF
--- a/src/shared/lib/getCanonicalMutationType.ts
+++ b/src/shared/lib/getCanonicalMutationType.ts
@@ -98,6 +98,7 @@ export function getProteinImpactTypeFromCanonical(mutationType:CanonicalMutation
         case "fusion":
         case "silent":
         case "other":
+        default:
             return "other";
     }
 }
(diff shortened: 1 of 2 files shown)


src/shared/lib/getCanonicalMutationType.ts
```diff
@@ -98,6 +98,7 @@ export function getProteinImpactTypeFromCanonical(mutationType:CanonicalMutation
         case "fusion":
         case "silent":
```
alisman: @adamabeshouse seems like this code should just have case for the main named types and that have all over cases handled by default.  
(Incidentally, there's a linting rule that requires a default for all switch statements. Would've helped here.)
I'll write a test for this after we decide whether to change code in way I suggested above.

adamabeshouse: that makes sense from a code simplicity view but I think it's better in this case to spell it out so we can see exactly what's getting assigned to other

alisman: So, I can add this last one, but I think i should leave other just in case there are others what we don't know about, agreed?

adamabeshouse: I think `default` is a good idea, I'm just explaining why we should still keep the other cases (fusion, silent..) even though logically it would be equivalent to remove them

PR #581, opened 2017-08-23
IE11 and Edge are now properly detected.

Tamba32:
# What? Why?
Unsupported browser modal now no longer shows when using IE11 and Edge.

### DIFF
--- a/src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
+++ b/src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
@@ -38,30 +38,25 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
     }
 
     handleUnsupportedBrowsers(name:string, version: string, localStorage: boolean) {
-        return;
-        // if (localStorage) {
-        //     this.setState({show: false});
-        //     return;
-        // }
-        // const sessionStorage = window.sessionStorage.browserError || false;
-        // if (sessionStorage === true) {
-        //     this.setState({show: false});
-        // } else {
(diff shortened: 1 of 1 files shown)


src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -38,30 +38,28 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
...
+            return;
+        }
+        const sessionStorage = window.sessionStorage.browserError || false;
+        if (sessionStorage === true) {
+            this.setState({show: false});
+        } else {
+            name = name.toLowerCase();
+            const isIE11 = (name === "internet explorer" || name === "msie") && Number(version.slice(0,2)) === 11;
+
+            if (String(name) === 'unsupported') {
```
inodb: where is the name set to 'unsupported'?

inodb: Seems like this case is unnecessary?

src/shared/components/unsupportedBrowserModal/UnsupportedBrowserModal.tsx
```diff
@@ -38,30 +38,28 @@ export default class UnsupportedBrowserModal extends React.Component<{}, IBrowse
     }
 
     handleUnsupportedBrowsers(name:string, version: string, localStorage: boolean) {
```
inodb: In the `render` function (not shown in this diff). There's is still a reference to name being `ie` which it will never be with browser package afaik

PR #583, opened 2017-08-24
GSoC2017: Integrating MolecularMatch clinical trials data into cBioPortal

emswbandara:
# What? Why?
Fix # .

This is the frontend component for the GSoC project done to integrate MolecularMatch clinical trials into cBioPortal. This will include another annotation in the annotation column on PatientViewPage for the mutations that has matching clinical trials in MolecularMatch. Similar to other annotations, on hovered over the molecularmatch annotation it will display the tooltip. In the tooltip (molecularmatch card) it displays the top 5 clinical trial details with highest matching scores along with the links to relevant trials in MolecularMatch website. 

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


inodb:
@zhx828 @emswbandara I think we should lazy load molecular match annotations. Similar to mRNA expression. That would also significantly decrease load. The only reason we fetch everything at the start for OncoKB and hotspot is because we sort on those. I think we should change this architecture a bit so we can have some annotations that can lazy load (only load visible ones).

inodb:
@emswbandara Good job though otherwise! Thanks for all your work! My comments are more on general architecture of mutations table, not towards this specific PR

PR #587, opened 2017-08-25
Mutations tab supporting multiple studies

adamabeshouse:
Corresponding to: https://github.com/cBioPortal/cbioportal/pull/3045

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -331,6 +331,17 @@ export class PatientViewPageStore {
         })
     }, []);
 
+    readonly geneticProfileIdToGeneticProfile = remoteData<{[geneticProfileId:string]:GeneticProfile}>({
+        await:()=>[this.geneticProfilesInStudy],
+        invoke:()=>{
+            return Promise.resolve(this.geneticProfilesInStudy.result.reduce((map:{[geneticProfileId:string]:GeneticProfile}, next:GeneticProfile)=>{
+                map[next.geneticProfileId] = next;
+                return map;
+            }, {}));
+        }
+    }, {});
+
+
     public readonly mrnaRankGeneticProfileId = remoteData({
         await: () => [
             this.geneticProfilesInStudy
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -388,6 +388,17 @@ export class PatientViewPageStore {
         }
     });
 
+    readonly studyToGeneticProfileDiscrete = remoteData({
+        await: ()=>[this.geneticProfileIdDiscrete],
+        invoke:async ()=>{
+            const ret:{[studyId:string]:GeneticProfile} = {};
+            if (this.geneticProfileIdDiscrete.result) {
+                ret[this.studyId] = await client.getGeneticProfileUsingGET({geneticProfileId:this.geneticProfileIdDiscrete.result});
+            }
+            return ret;
+        }
+    }, {});
+
     readonly darwinUrl = remoteData({
         await: () => [
             this.derivedPatientId
@@ -580,7 +591,7 @@ export class PatientViewPageStore {
     }
 
     @cached get discreteCNACache() {
-        return new DiscreteCNACache(this.geneticProfileIdDiscrete.result);
+        return new DiscreteCNACache(this.studyToGeneticProfileDiscrete.result);
     }
 
     @cached get oncoKbEvidenceCache() {
@@ -596,11 +607,11 @@ export class PatientViewPageStore {
     }
 
     @cached get cancerTypeCache() {
-        return new CancerTypeCache(this.studyId);
+        return new CancerTypeCache();
     }
 
     @cached get mutationCountCache() {
-        return new MutationCountCache(this.mutationGeneticProfileId.result);
+        return new MutationCountCache();
     }
 
     @action setActiveTabId(id: string) {
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -391,6 +391,7 @@ export class PatientViewPageStore {
     readonly studyToGeneticProfileDiscrete = remoteData({
         await: ()=>[this.geneticProfileIdDiscrete],
         invoke:async ()=>{
+            // we just need it in this form for input to DiscreteCNACache
             const ret:{[studyId:string]:GeneticProfile} = {};
             if (this.geneticProfileIdDiscrete.result) {
                 ret[this.studyId] = await client.getGeneticProfileUsingGET({geneticProfileId:this.geneticProfileIdDiscrete.result});
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -7,6 +7,8 @@ import AjaxErrorModal from "shared/components/AjaxErrorModal";
 import exposeComponentRenderer from 'shared/lib/exposeComponentRenderer';
 import {ResultsViewPageStore} from "./ResultsViewPageStore";
 import Mutations from "./mutation/Mutations";
+import {stringListToSet} from "../../shared/lib/StringUtils";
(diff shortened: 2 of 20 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -388,6 +388,17 @@ export class PatientViewPageStore {
         }
     });
 
+    readonly studyToGeneticProfileDiscrete = remoteData({
+        await: ()=>[this.geneticProfileIdDiscrete],
+        invoke:async ()=>{
+            const ret:{[studyId:string]:GeneticProfile} = {};
+            if (this.geneticProfileIdDiscrete.result) {
+                ret[this.studyId] = await client.getGeneticProfileUsingGET({geneticProfileId:this.geneticProfileIdDiscrete.result});
```
alisman: not sure if i understand.  wont this map only ever have one key?  the studyId?

adamabeshouse: thats right, but this is just to fit the interface of DiscreteCNACache

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -61,24 +73,27 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
-            resultsViewPageStore.studyId = props.studyId;
-            if (typeof props.samples === "string") {
-                resultsViewPageStore.sampleListId = props.samples;
-            } else {
-                resultsViewPageStore.sampleList = props.samples;
-            }
+        exposeComponentRenderer('renderMutationsTab',
+            (props:MutationsTabInitProps)=>{
+                resultsViewPageStore.hugoGeneSymbols = props.genes;
+                if (props.studyToSampleMap) {
```
alisman: seems like these modes are exclusive, so should setting one clear the other automatically?  Like, store should not be allowed to have  studyToSampleMap and studyToSampleIds?  We did this with patient view.  You can't have both patientId and sampleId set at once.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -61,24 +73,27 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+                    let _studyToSampleIds:{[studyId:string]:{[sampleId:string]:boolean}} = {};
+                    _.each(props.studyToSampleMap, (sampleIds:string[], studyId:string)=>{
+                        _studyToSampleIds[studyId] = stringListToSet(sampleIds);
```
alisman: also, it's confusing that we keep list of string sampleIds in study buckets and then have another flat sampleId collection.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -34,17 +40,61 @@ export class ResultsViewPageStore {
...
+    //@observable studyId: string = '';
+    //@observable sampleListId: string|null = null;
     @observable hugoGeneSymbols: string[]|null = null;
-    @observable sampleList: string[]|null = null;
+    //@observable sampleList: string[]|null = null;
+    @observable _studyToSampleIds:{[studyId:string]:{[sampleId:string]:boolean}}|null = null;
+    @observable studyToSampleListId:{[studyId:string]:string}|null = null;
+
+    readonly studyToSampleIds = remoteData<{[studyId:string]:{[sampleId:string]:boolean}}>(async()=>{
+        if (this._studyToSampleIds) {
```
alisman: yeah, i would love to have some computed mode property that would hide this.  like, mode:'customSampleList'|'definedSampleList'
(maybe there's something better than 'definedSampleList')
My brain has a hard time distinguishing sampleId from sampleListId

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -34,17 +40,61 @@ export class ResultsViewPageStore {
...
+                const studyId = profile.studyId;
+                if (!ret[studyId] && isMutationProfile(profile)) {
+                    ret[studyId] = profile;
+                }
+            }
+            return Promise.resolve(ret);
+        }
+    }, {});
+
+    @computed get studyIds():string[] {
```
alisman: we could derive this from samples.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -63,55 +113,49 @@ export class ResultsViewPageStore {
         else {
             const store = new MutationMapperStore(AppConfig,
```
alisman: @adamabeshouse @onursumer this is not related to this review but i've been wanting to discuss, the collection of MutationMapperStores is not "reactive" because they get instantiated by view layer calling this method.  This became a problem for loading the MutEx tab prior to the Mutations tab because these stores had not yet been instantiated.  I am ended up just instantiating them on that tab and I do see that they will be resused when switching tabs, but still, I think we should consider a different solution.  Lets discuss soon.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -120,17 +164,34 @@ export class ResultsViewPageStore {
...
+            this.samples,
             this.clinicalDataForSamples
         ],
-        invoke: async () => fetchSamplesWithoutCancerTypeClinicalData(this.sampleIds, this.studyId, this.clinicalDataForSamples)
+        invoke: () => {
+            const sampleHasData:{[sampleUid:string]:boolean} = {};
+            for (const data of this.clinicalDataForSamples.result) {
+                sampleHasData[toSampleUuid(data.clinicalAttribute.studyId, data.entityId)] = true;
+            }
+            return Promise.resolve(this.samples.result.filter(sample=>{
```
alisman: why Promise.resolve?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -141,41 +202,69 @@ export class ResultsViewPageStore {
     }, []);
 
     readonly studies = remoteData({
-        invoke: async()=>([await client.getStudyUsingGET({studyId: this.studyId})])
+        invoke: ()=>Promise.all(this.studyIds.map(studyId=>client.getStudyUsingGET({studyId})))
```
alisman: I guess we don't need async because we already return promise?

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: invoke has to return a promise. when its async thats done automatically, but I wanted to not use async here as a way of denoting that no async stuff is actually done

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: I think that may not be future proof - eg one study could have a sample list id, another a custom sample list

src/pages/resultsView/ResultsViewPageStore.ts
alisman: also, might be nice to return map keyed by studyId?

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: you're right -- it should be a computed map or a computed flat list or something - something reactive

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -34,17 +42,63 @@ export class ResultsViewPageStore {
...
-    @observable studyId: string = '';
-    @observable sampleListId: string|null = null;
     @observable hugoGeneSymbols: string[]|null = null;
-    @observable sampleList: string[]|null = null;
+    @observable samplesSpecification:SamplesSpecificationElement[] = [];
+
+    readonly studyToSampleIds = remoteData<{[studyId:string]:{[sampleId:string]:boolean}}>(async()=>{
+        const sampleListsToQuery:{studyId:string, sampleListId:string}[] = [];
+        const ret:{[studyId:string]:{[sampleId:string]:boolean}} = {};
+        for (const sampleSpec of this.samplesSpecification) {
```
alisman: i don't think we can for ... of yet.  lets just use Array.forEach

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: if that need comes up we can definitely derive it

PR #588, opened 2017-08-30
Genome nexus functional consequence Column (Genome Nexus)

inodb:
Add `Functional Consequence` column to Mutations Table with annotations from Mutation Assessor & Sift pulled from Genome Nexus (http://genomenexus.org). The annotations are lazily loaded, similar to Cohort column.

https://cbioportal-frontend-pr-588.herokuapp.com/#/patient?caseId=P04&studyId=lgg_ucsf_2014

TODO:
- [x] Make logos for mutation assessor H/M/L
- [ ] Fix mutation assessor link in tooltip
- [x] Fix Styling (looks a little wonky behind COSMIC column)
- [x] Wonder whether it makes more sense to show just letters and no icons? Might be more readable (discuss) Have similar color scheme for all (from neutral gray to high impact red) and perhaps the logo in the header instead.
- [ ] Sift how to deal with multiple transcripts (just picking first transcript now)

screen shot 2017-08-30 at 10 03 21 pm
screen shot 2017-08-30 at 10 04 58 pm


jjgao:
Super cool! Great job. @inodb 

The MA icons look good. The SIFT icon look funny. I am wondering if it is possible to have sub-columns? Then we can put the icon in the column header, and use only letters in the table.

In the mouse over, we should have the resource name, maybe with a link out too.

inodb:
Thanks @jj! Yeah I think sub columns would make sense at some point. Opened an issue for this. Not sure how hard implementation would be. Sorting and filtering by one or more annotations comes to mind. A quick fix for this might be to just add a new line in the column header and include the logos on that second line. Then in a later iteration we can go about improving the mutation table to sub columns.

inodb:
Replaced https://github.com/cBioPortal/cbioportal-frontend/pull/607

PR #590, opened 2017-08-30
Study column in mutation mapper for multiple studies

adamabeshouse:
It shows the (truncated) study name, linking to that study summary, tooltip the full study name and description.

The column is always available but only visible by default if there is more than one study in the query.

image

### DIFF
--- a/src/shared/components/mutationTable/column/StudyColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/StudyColumnFormatter.tsx
@@ -26,7 +26,9 @@ export default class StudyColumnFormatter
                 <a href={getStudySummaryUrl(study.studyId)} target="_blank">
                     <TruncatedText
                         text={study.name}
-                        tooltip={<div style={{maxWidth: 300}}>{`${study.name}: ${study.description}`}</div>}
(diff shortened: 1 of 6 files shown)


src/shared/components/mutationTable/column/StudyColumnFormatter.tsx
```diff
@@ -0,0 +1,56 @@
...
+            return study.name;
+        }
+    }
+
+    public static filter(d:Mutation[], filterStringUpper:string, geneticProfileIdToGeneticProfile?:{[geneticProfileId:string]:GeneticProfile}, studyIdToStudy?:{[studyId:string]:CancerStudy}) {
+        const study = StudyColumnFormatter.getStudy(d, geneticProfileIdToGeneticProfile, studyIdToStudy);
+        if (!study) {
+            return false;
+        } else {
+            return study.name.toUpperCase().indexOf(filterStringUpper) > -1;
```
zhx828: study.name is always guaranteed?

adamabeshouse: I think so? every study in production has a name

PR #594, opened 2017-09-05
Implement cumulative fisher exact test

ersinciftci:


src/shared/lib/FisherExactTestCalculator.ts
```diff
@@ -0,0 +1,35 @@
...
+
+    p += getPValue(a, b, c, d, f);
+    if ((a * d) >= (b * c)) {
+        min = (c < b) ? c : b;
+        for (i = 0; i < min; i++) {
+            p += getPValue(++a, --b, --c, ++d, f);
+        }
+    }
+
+    if ((a * d) < (b * c)) {
```
adamabeshouse: else?

src/shared/lib/FisherExactTestCalculator.ts
```diff
@@ -0,0 +1,35 @@
...
+    const p: number = (f[a + b] + f[c + d] + f[a + c] + f[b + d]) - (f[a] + f[b] + f[c] + f[d] + f[n]);
+    return Math.exp(p);
+}
+
+export function getCumulativePValue(a: number, b: number, c: number, d: number): number {
+
+    let min: number, i: number;
+    const n: number = a + b + c + d;
+    let p: number = 0;
+    let f: number[] = new Array(n + 1);
```
adamabeshouse: seems like f is log factorial, could you write that out, in the argument names too?

src/shared/lib/FisherExactTestCalculator.ts
```diff
@@ -0,0 +1,35 @@
+function getPValue(a: number, b: number, c: number, d: number, f: number[]): number {
```
adamabeshouse: this is the probability of a config, not a p-value :)

src/shared/lib/FisherExactTestCalculator.ts
```diff
@@ -0,0 +1,35 @@
...
+            p += getPValue(++a, --b, --c, ++d, f);
+        }
+    }
+
+    if ((a * d) < (b * c)) {
+        min = (a < d) ? a : d;
+        for (i = 0; i < min; i++) {
+            p += getPValue(--a, ++b, ++c, --d, f);
+        }
+    }
```
adamabeshouse: I haven't thought through the math - does this strategy add all of, and only, the probabilities that are <= prob(a,b,c,d)?

adamabeshouse:
Also, it would be easier to use if you could use more descriptive variable names for the input than a,b,c,d

ersinciftci:
It seems like a,b,c,d is standard in this type of calculations, for example https://github.com/sackio/fishertest/blob/master/lib/bigfisher.js

PR #597, opened 2017-09-06
Hide pagination buttons in PaginationControls if all fits in smallest page

adamabeshouse:
https://github.com/cBioPortal/cbioportal/issues/3075

### DIFF
--- a/src/shared/components/paginationControls/PaginationControls.tsx
+++ b/src/shared/components/paginationControls/PaginationControls.tsx
@@ -41,6 +41,7 @@ export interface IPaginationControlsProps {
     lastPageDisabled?:boolean;
     pageNumberEditable?:boolean;
     groupButtons?:boolean;
+    hidePaginationIfOnePage?:boolean;
 }
 
 @observer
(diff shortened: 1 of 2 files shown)


src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -41,6 +41,7 @@ export interface IPaginationControlsProps {
     lastPageDisabled?:boolean;
     pageNumberEditable?:boolean;
     groupButtons?:boolean;
+    hidePaginationIfOnePage?:boolean;
```
alisman: i don't think this needs to be an option

src/shared/components/paginationControls/PaginationControls.tsx
```diff
@@ -147,6 +149,13 @@ export class PaginationControls extends React.Component<IPaginationControlsProps
         }
     }
 
+    private get shouldHidePaginationControls() {
+        return (this.props.hidePaginationIfOnePage
+            && this.props.itemsPerPageOptions
+            && typeof this.props.totalItems!=="undefined"
```
alisman: you can just use this.props.totalItems !== undefined.   I believe typeof === "undefined" is necessary only to support IE8

adamabeshouse: I think you're right when it comes to undefined properties on an object. I think I was used to it because I'm pretty sure if its not an undefined property but a possibly undefined variable you may get a reference error, or something like that.

src/shared/components/paginationControls/PaginationControls.tsx
adamabeshouse: really? I could see an aesthetic use case for it

PR #600, opened 2017-09-07
Add bar graph.

Tamba32:
# What? Why?
Cancer Summary Plot


src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+        this.handleTotalSliderChange = this.handleTotalSliderChange.bind(this);
+        this.toggleShowControls = this.toggleShowControls.bind(this);
+        this.setPngAnchor = this.setPngAnchor.bind(this);
+        this.setPdfAnchor = this.setPdfAnchor.bind(this);
+        this.downloadPdf = this.downloadPdf.bind(this);
+        this.resetSliders = this.resetSliders.bind(this);
+    }
+
+    @computed get datasets() {
+        const {data} = this.props;
```
alisman: lets change the name data to something that better describes what it is.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+        const {yAxis, showGenomicAlt, multiSelectValue:msv} = this;
+        return _.without(
```
alisman: code this deeply nested is very hard to read.  instead of using _.without, I think you should use _.reduce to either discard items or include the transformed item.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+            _.map(data, (cancer:ICancerAlterationData, cancerTypeS:string) => {
```
alisman: i think you named it cancerTypeS to indicate types string, but i kept reading it as plural.   i would get rid of the S

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+                const totalCases = cancer.total;
+                const cancerAlterations:ICancerAlterationPlotData = _.omit(cancer, ['total']);
+                const altTotalCount = _.reduce(cancerAlterations, (total:number, value:number) => total + value, 0);
+                const altTotalPercent = this.toRoundedNumber(altTotalCount / totalCases * 100);
+                if (_.intersection(["all", "", cancerTypeS.toLowerCase()], msv.split(",").map(val => val.toLowerCase())).length && totalCases >= this.totalCasesValue) {
```
alisman: lets just call msv "multiselectvalue." 
take the msv.split.map statement and assign it to a local variable above that gives reader a sense of what it represents.   like, selected types or something.  just too hard to read.
i think you could also do toLowerCase operation on msv prior to split, to avoid map.
also, why is empty string "" a possibility?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+                            const percent = count / totalCases * 100;
+                            const total = yAxis === "abs-count" ? count : percent;
+                            return {
+                                label: altType,
+                                totalCases,
+                                altTotalPercent,
+                                total,
+                                count,
+                                percent,
+                                backgroundColor: showGenomicAlt  ? this.getColors(altType) : '#aaaaaa'
```
alisman: lets get the hex values into some configuration map

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+            mutated:"#36a2eb",
+            amplified:"#ff6384",
+            deleted:"#ffce56",
+            multiple:"#cc65fe",
+        };
+        return colors[color];
+    }
+
+    private toRoundedNumber(decimal: string | number) {
+        return Math.round(Number(decimal) * 100) / 100;
```
alisman: i think you can use lodash _.round instead of this helper method.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    @computed private get altCasesMax() {
+        return Math.max(...this.datasets.map(data => data!.sortCount));
```
alisman: seems like we shouldn't need the ! to assert that a member of dataset is not undefined.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+    }
+
+    @computed private get totalCasesMax() {
+        const value = this.multiSelectValue;
+        return Math.max(..._.map(
+            _.filter(this.props.data, (o, label) => (
+                !!_.intersection(["all", "", label.toLowerCase()], value.split(",").map(val => val.toLowerCase())).length
```
alisman: since this is used above, maybe make a computed getter selectedValues or something, which handles split and lowercasing of values.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+            )),
+            (cancer:ICancerAlterationData) => cancer.total));
+    }
+
+    @computed private get chartData():IBarGraphData {
+        const {datasets, xAxis, altCasesValue} = this;
+
+        const some2 = _.orderBy(datasets,
```
alisman: better name!

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,448 @@
...
+                                [xAxis === "y-axis" ? 'sortCount': 'label'],
+                                [xAxis === "y-axis" ? 'desc' : 'asc']);
+
+        const datasetsF = _.flatten(
+            some2.map((line, i) => (
+                line!.data.map(thing => (
```
alisman: again would question need for !

src/shared/components/cancerSummary/CancerSummary.tsx
Tamba32: I'm not sure I know what configuration map means: Isn't what I have on line 189?  I just did the same thing as https://github.com/cBioPortal/cbioportal-frontend/blob/842b4d69f5e9747661e475b696bf2dd4cbcdf299/src/shared/components/barGraph/BarGraph.tsx#L33

src/shared/components/cancerSummary/CancerSummary.tsx
Tamba32: The empty string is a possibility because when you deselect all values in the multiselect, it's value is returned back as `""`

src/shared/components/cancerSummary/CancerSummary.tsx
alisman: yes, 189 is right approach, but the #aaaa should also be a constant or in a map.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -85,7 +86,7 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
 
     public render() {
 
-        return null;
+        return <CancerSummaryContainer/>;
```
ersinciftci: I think you were supposed to not commit this? Isn't it for your local testing?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+    total: number;
+    count: number;
+    percent: number;
+    data: number[];
+    backgroundColor: string;
+}
+
+export interface IBarGraphData {
+    labels: string[];
+    datasets: IBarGraphDatasets[];
```
adamabeshouse: are these labels different than the `label` property of `IBarGraphDatasets`?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+
+export interface ICancerAlterationPlotData {
+    [alteration:string]:number;
+    mutated:number;
+    amplified:number;
+    deleted:number;
+    multiple:number;
+}
+
+export interface IBarGraphDatasets {
```
adamabeshouse: the name should be `IBarGraphDataset` (singular), right?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+}
+
+// interface IGeneAltertionData {
+//     [gene:string]: {
+//         [cancerType:string]:ICancerAlterationData
+//     };
+// }
+
+interface ICancerSummaryContentProps {
+    data: {
```
adamabeshouse: As per what we've been trending towards, I think it would be better for this prop to be an array of `ICancerAlterationData`, and have `ICancerAlterationData` have a `cancerType` member.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+        [cancerType:string]:ICancerAlterationData
+    };
+}
+
+export interface IBarChartDatasets {
```
adamabeshouse: Why do you have this defined here instead of more proximate to the bar graph code in `SummaryBarGraph.tsx`?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+    }
+}
+
+// interface ICancerSummaryContainerState {
+//     activeTab?: string;
+//     data: IGeneAltertionData;
+// }
+
+@observer
+export default class CancerSummaryContainer extends React.Component<{},{}> {
```
adamabeshouse: I think it would be better to move this into its own file.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+                </div>
+                {controls}
+                <SummaryBarGraph data={this.chartData} yAxis={this.yAxis} xAxis={this.xAxis}
+                                 setPdfAnchor={this.setPdfAnchor} setPngAnchor={this.setPngAnchor} legend={this.showGenomicAlt}/>
+            </div>
+        );
+    }
+}
+
+// interface ICancerSummaryContainerState {
```
adamabeshouse: remove comment

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -0,0 +1,215 @@
...
+interface IChartLegendItem extends ChartLegendItem {
+    datasetIndex: number;
+}
+
+@observer
+export default class SummaryBarGraph extends React.Component<ISummaryBarGraphProps, {}> {
+
+    private chartTarget:HTMLCanvasElement;
+
+    @observable private chart:any;
```
adamabeshouse: if you dont want every member of `chart` to become observable but you just want to be able to react to when `SummaryBarGraph::chart` is changed, you should use `@observable.ref`

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -0,0 +1,215 @@
...
+            return {label, count, percent};
+        }
+
+        // Set Text
+        if (tooltipModel.body) {
+
+            const bodyLines:{ label:string; count:number; percent:number }[] = tooltipModel.body.map(getBody);
+            const {totalCases, altTotalPercent} = data.datasets[tooltipModel.dataPoints[0].datasetIndex];
+
+            let innerHtml = (
```
adamabeshouse: better to write this as JSX, here and everywhere else instead of string innerHTML.

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+import {ThreeBounce} from 'better-react-spinkit';
+import {FormGroup, ControlLabel, FormControl} from 'react-bootstrap';
+import jsPDF from 'jspdf';
+
+import 'react-select/dist/react-select.css';
+import 'react-rangeslider/lib/index.css';
+
+import SummaryBarGraph from './SummaryBarGraph';
+
+export interface ICancerAlterationPlotData {
```
adamabeshouse: shouldnt it be`ICancerTypeAlterationPlotData`?

src/shared/components/cancerSummary/CancerSummary.tsx
```diff
@@ -0,0 +1,431 @@
...
+    data: number[];
+    backgroundColor: string;
+}
+
+export interface IBarGraphData {
+    labels: string[];
+    datasets: IBarGraphDatasets[];
+}
+
+export interface ICancerAlterationData extends ICancerAlterationPlotData {
```
adamabeshouse: like above, `CancerAlteration` => `CancerTypeAlteration`

src/shared/components/cancerSummary/CancerSummary.tsx
adamabeshouse: should it be singular, `IBarChartDataset`?

src/shared/components/cancerSummary/SummaryBarGraph.tsx
alisman: lets discuss how/if we should make this react.  it's possible to do.

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -0,0 +1,215 @@
...
+        tooltipEl.style.opacity = '1';
+        // tooltipEl.style.display = 'block';
+        tooltipEl.style.left = position.left + tooltipModel.caretX + 'px';
+        tooltipEl.style.top = position.top + tooltipModel.caretY + 5 + 'px';
+    }
+
+    private componentDidMount() {
+        this.chartConfig.data = this.props.data;
+        this.chartConfig.options = this.chartOptions;
+        this.loadChart();
```
alisman: because this code should only be called once, i would just put it up here rather than in method.

src/shared/components/cancerSummary/SummaryBarGraph.tsx
adamabeshouse: what do you mean?

adamabeshouse: I understand now, didn't look carefully enough. It's fine the way it is.

src/shared/components/cancerSummary/CancerSummary.tsx
Tamba32: No. The `labels` array is just the full list of all the `label`s

adamabeshouse: in my opinion that should be thrown out, if its accessible from the `IBarGraphDataset(s)` - better to have one source of truth

PR #601, opened 2017-09-11
Showing message when unrecognized string is interpreted as mutation code

adamabeshouse:
Addresses https://github.com/cBioPortal/cbioportal/issues/3031

image

In this example, it does not have any notes, because it recognizes the form AMINO_ACID_CODE + NUMBER + STRING, but we can easily change it to have a note in this case as well
image

In this example, it recognizes the form AMINO_ACID_CODE + NUMBER, and will be submitting V600 as a position filter. But with just the number, it does not recognize that form, which is something to consider.
image

### DIFF
--- a/src/shared/components/query/styles.module.scss
+++ b/src/shared/components/query/styles.module.scss
@@ -73,6 +73,11 @@
     color: $redError;
     font-weight: bold;
   }
+
+  .oqlMessage {
+    color: $brand-info;
+    font-weight: bold;
+  }
 }
 
 .CancerStudySelector,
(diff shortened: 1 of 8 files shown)


src/shared/components/query/styles.module.scss
```diff
@@ -73,6 +73,11 @@
     color: $redError;
     font-weight: bold;
   }
+
+  .oqlMessage {
+    color: $blueNotice;
```
alisman: i thought we were going to use the class text-info?

adamabeshouse: I didn't want the other styling associated with it so I just copied the color

alisman: ok, but in variables we should be able to reference the underlying bootstrap color name.  i think it's important to establish a practice wherein we standardize these sort of semantic color.

PR #602, opened 2017-09-11
Fix LazyMobXTable maxPage calculation

adamabeshouse:


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -552,6 +549,8 @@ class LazyMobXTableStore<T> {
 
         this._page = 0;
         this.itemsPerPage = lazyMobXTableProps.initialItemsPerPage || 50;
+
+        reaction(()=>this.displayData.length, ()=>{ this.page = this.page; /* trigger recomputation of page when data length changes */});
```
alisman: this line seems potentially like a mistake
  this.page = this.page?

src/shared/components/lazyMobXTable/utils.ts
```diff
@@ -0,0 +1,8 @@
+import {SHOW_ALL_PAGE_SIZE} from "../paginationControls/PaginationControls";
+export function maxPage(displayDataLength:number, itemsPerPage:number) {
```
alisman: i think "pages" should not be zero indexed.  i.e. we don't ever have zero pages.  we have one page, two pages.

inodb:
@adamabeshouse This might be better to `hotfix` no?

alisman:
I agree.  So, switch this pr to hotfix/

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
adamabeshouse: no its right - in `set page` it clamps the input value, and when the display data length changes it needs to be potentially clamped. So this just triggers that recomputation, as it says in the comment

src/shared/components/lazyMobXTable/utils.ts
adamabeshouse: in my experience indexing math always works out easier when things are zero-indexed

adamabeshouse:
see PR to hotfix https://github.com/cBioPortal/cbioportal-frontend/pull/611

PR #607, opened 2017-09-15
Genome nexus Functional Impact Column

inodb:
Connect to Genome Nexus to get mutation assessor, PhyloPhen-2 and SIFT
functional impact predictions. Show the functional impact predictions in a separate column. This allows us to get rid of the current Mutation Assessor column. After this is merged we can remove the related data from the database as well.
screen shot 2017-09-15 at 7 21 40 pm

- [x] Hide Functional Impact Column on default on both query page and patient view
- [x] Prettify header (put link)
- [x] Fix mutation assessor link in tooltip
- [x] Figure out why SIFT/Polyphen Score is often 0

**NOTE to reviewers:** We will hide this column on default so it's not essential this column looks superb. We will eventually prolly replace the entire column with some sort of mouseover summary on the protein change itself for other annotation sources.


adamabeshouse:
does this PR implement headers that span multiple columns? or are you just aligning the header icons and the cell positions manually?

inodb:
@adamabeshouse manual alignment atm, similar to annotation column.

PR #624, opened 2017-09-21
Handle CIViC API errors and replace "/" in queried gene symbols

oplantalech:
# What? Why?
Fix cBioPortal/cbioportal#3120. All failed CIViC requests now have a "error icon" like OncoKB annotation.

The error is caused because two gene symbol contain a `/`. Therefore, I've also replaced all `/` in queried gene symbols by an url-escaped slash (`%2F`). Currently this solution is not yet working (so you will see error icons) but I've reported it to CIViC so that they can fix it.


oplantalech:
The ticket I logged for CIViC is: griffithlab/civic-server#361

src/shared/lib/CivicUtils.ts
```diff
@@ -30,6 +30,8 @@ export function getCivicGenes(geneSymbols: Array<string>): Promise<ICivicGene> {
     let promises: Array<Promise<Array<ICivicGeneData>>> = [];
     let ids = '';
     geneSymbols.forEach(function(geneSymbol: string) {
+        //Encode "/" characters
+        geneSymbol = geneSymbol.replace(/[\/]/g,'%2F');
```
alisman: minor quibble, but i don't think you need the [] brackets here.  you would use them if you were specifying a class of acceptable chars.   but here there's ony one, so /\// will work.

src/shared/lib/CivicUtils.ts
```diff
@@ -30,6 +30,8 @@ export function getCivicGenes(geneSymbols: Array<string>): Promise<ICivicGene> {
...
         // Check if we already have it in the cache
         if (civicGenes.hasOwnProperty(geneSymbol)) {
             return;
```
alisman: another quibble: I noticed that we're building the ids string with concatenation.  would be better to build an array and then use the array.join method to build the coma delim'd string that is required by the client.

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -56,19 +58,32 @@ export default class AnnotationColumnFormatter
...
         if (geneVariants || (geneEntry && geneEntry.description !== "")) {
             civicEntry = buildCivicEntry(geneEntry, geneVariants);
         }
 
         return civicEntry;
     }
+    
+    public static getCivicStatus(civicGenesStatus:"pending" | "error" | "complete", civicVariantsStatus:"pending" | "error" | "complete"): "pending" | "error" | "complete"
+    {
+    if (civicGenesStatus == "error" || civicVariantsStatus == "error") {
```
alisman: please always use === instead of double equals.   just a good habit to get into.  (we need to enforce this with linting.)

src/shared/model/Civic.ts
```diff
@@ -25,3 +25,13 @@ export interface ICivicEntry {
     url: string;
     variants: {[name: string]: ICivicVariantData};
 };
+
+export interface ICivicGeneDataWrapper {
+    status: "pending" | "error" | "complete";
```
alisman: We really need to have a single interface for these status values.  Why don't you make one for civic and they we can easily replace it when we have a better for whole site.  Mostly we rely on status of MobxPromise, so this isn't an issue.

PR #636, opened 2017-09-26
Adjusted max width of chart.  Removed cancer types with 0 data.

Tamba32:


src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -192,9 +191,8 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
                     ticks: {
                         fontSize: 11,
                         callback: function(value:number) {
-                            return that.props.yAxis === "abs-count" ? value : (Math.round(value*10)/10) + '%';
+                            return _.round(value, 1) + (that.props.yAxis === "abs-count" ? '': '%');
```
alisman: I changed this from what it was before because it was rounding, for example, 1.2 down to 1.  There are instances where the whole chart is only 2% tall and we need decimal granularity.

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -250,7 +248,11 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
     }
 
     private get width() {
-        return 220 + this.props.data.labels.length * 57;
+        const maxWidth = 220 + this.props.data.labels.length * 57;
+        if (document.getElementById("pancancer_study_summary")) {
```
alisman: hmm. why is there an element of Id pancancer_study_summary?  lets think of a way to do this without cloodge.   like, lets give the chart a max width of whatever the fixed width of the content container of site is.  
we could measure this, but lets do this kind of "dirty" thing in the page component and pass it in through props to the this component.   in general, keep these presentation components "clean."  They shouldn't have to know about their environment.

src/shared/components/cancerSummary/SummaryBarGraph.tsx
Tamba32: Having `_.round(value, 1)` should make it so it rounds to 1 decimal place.  Without this function, labels sometimes appear as 0.40000000... or 0.6000000...

PR #638, opened 2017-09-26
Move sequenced and copyNumberSegmentPresent to DETAILED projection

ersinciftci:


src/shared/api/generated/CBioPortalAPI.ts
```diff
@@ -485,7 +485,7 @@ export type ClinicalAttribute = {
  */
 export default class CBioPortalAPI {
 
-    private domain: string = "";
+    private domain: string = "http://localhost:8080/api";
```
alisman: mistake?

ersinciftci: yeah, just realized that API was already up to date, so reverted api.json change too

PR #639, opened 2017-09-27
sequencedSamples call should fail silently

alisman:
sequenced samples call is currently erroring for some patients and showing a big ugly error dialog.  it should not.

https://github.com/cBioPortal/cbioportal/issues/3127

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -321,8 +321,13 @@ export class PatientViewPageStore {
         )
     }, {});
 
-    readonly sequencedSampleIdsInStudy = remoteData(async() => {
-        return stringListToSet(await client.getAllSampleIdsInSampleListUsingGET({sampleListId:`${this.studyId}_sequenced`}));
+    readonly sequencedSampleIdsInStudy = remoteData({
+        invoke: async () => {
+            return stringListToSet(await client.getAllSampleIdsInSampleListUsingGET({sampleListId:`${this.studyId}_sequenced`}));
+        },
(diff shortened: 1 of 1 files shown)


alisman:
@inodb right now if sequenced samples fails, the genomic overview will not render. is this appropriate?

inodb:
@alisman ah nvm i see example in the issue

inodb:
@alisman ok yeah so this looks good to me. Don't fully get where the copy number information is coming from if this sample wasn't sequenced. But guess they had some orthogonal method to determine that. For a separate PR we might want to not show the mutations table in case there are no sequenced samples, instead of showing that there are 0 mutations.

pieterlukasse:
just saw this. I think it should *not* silently fail. Instead, the call should not be made if mutation data is not available.

pieterlukasse:
@inodb CNA info comes from _cna or _gistic case list. The _sequenced case list seems to be linked to mutations profile in most of the existing logic.

alisman:
in hotfix right now, we have what I think is effectively the right behavior.  IE, no dialog will show when that call fails and user can continue to use page.   @jjgao @inodb if you can confirm that @pieterlukasse is right that there will NEVER be data returned from getAllSampleIdsInSampleListUsingGET endpoint when there are no mutations that I can add logic to avoid making call.

jjgao:
@pieterlukasse I agree with you, ideally, we should not even try to pull mutations data (including case list) if we know there is no mutations profile.

But I can live with this for now. We should use Ersin's change in https://github.com/cBioPortal/cbioportal/pull/3070 once rc is released.

PR #640, opened 2017-09-27
Change bar order.

Tamba32:
# What? Why?
Change bar order.

### DIFF
--- a/src/shared/components/cancerSummary/CancerSummaryContent.tsx
+++ b/src/shared/components/cancerSummary/CancerSummaryContent.tsx
@@ -27,7 +27,6 @@ export interface IBarGraphDataset {
     total: number;
     count: number;
     percent: number;
-    data: number[];
     backgroundColor: string;
 }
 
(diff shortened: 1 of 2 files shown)


src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -171,14 +177,14 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
...
-            gain: "rgb(255,182,193)",
-            homdel: "rgb(0,0,255)",
-            mrnaExpressionUp: "#FF989A",
-            mrnaExpressionDown: "#529AC8",
-            protExpressionUp: "#FF989A",
-            protExpressionDown: "#E0FFFF",
+            multiple:"#383838",
+            fusion:"#8B00C9",
+            gain:"rgb(255,182,193)",
+            homdel:"rgb(0,0,255)",
```
jjgao: should this be hetloss? deleted above looks like homdel...

Tamba32: No, `homdel` shouldn't be `hetloss`.  

The store returns `hetloss` as `deleted`.  Therefore, here, to get the color of `hetloss`, I use the `deleted` key.  Then, on line 109, if the alteration is labeled as `deleted`, I change its value to `hetloss`.

PR #644, opened 2017-09-29
Add strong typings to collections based on alteration subtypes.  Fix bar chart perf issue.

alisman:
Gene tags in cacner type summury were re-rendering unecessarily because of error in assignment of gene attribute.
Also, the alteration subtypes weren't lining up properly with color and label text collections.  Application of stronger types prevents this.

### DIFF
--- a/src/shared/components/cancerSummary/CancerSummaryContent.tsx
+++ b/src/shared/components/cancerSummary/CancerSummaryContent.tsx
@@ -13,12 +13,20 @@ import 'react-rangeslider/lib/index.css';
 
 import SummaryBarGraph from './SummaryBarGraph';
 
+
 export interface ICancerTypeAlterationPlotData {
-    mutated:number;
-    amplified:number;
-    deleted:number;
+    mutated: number;
+    amp:number;
+    homdel:number;
+    hetloss:number;
+    gain:number;
+    fusion:number;
+    mrnaExpressionUp:number;
+    mrnaExpressionDown:number;
+    protExpressionUp:number;
+    protExpressionDown:number;
     multiple:number;
-}
+};
 
 export interface IBarGraphDataset {
     label: string;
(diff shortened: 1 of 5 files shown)


src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -174,20 +182,21 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
...
+            fusion:"#8B00C9",
             mrnaExpressionUp:"#FF989A",
             mrnaExpressionDown:"#529AC8",
             protExpressionUp:"#FF989A",
             protExpressionDown:"#E0FFFF",
+            multiple:"#666"
         };
-        return this.showGenomicAlt ? (colors[color] || "#000000") : '#aaaaaa';
+        // TODO: fix ts index signature issue so we don't have to cast alterationToColor as any
+        return this.showGenomicAlt ? ((alterationToColor as any)[color] || "#000000") : '#aaaaaa';
```
alisman: if anyone knows how to do resolve this issue, i'm all ears.  basically, if you type a collection as Recorder (with keyof) or even as  { [k:key of Whatever]: string }
then ts throws an error if you do collection[someKey]

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -174,20 +182,21 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
     }
 
     private getColors(color:string) {
```
adamabeshouse: the naming here is very bad.. you're getting a _single_ color, and the parameter is the alteration type not the color...

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -174,20 +182,21 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
...
-        const colors:{[id:string]:string} = {
+        const alterationToColor: Record<keyof ICancerTypeAlterationPlotData, string> = {
```
adamabeshouse: there's no reason for this map to be rebuilt on every call

src/shared/components/cancerSummary/CancerSummaryContent.tsx
adamabeshouse: can you do `{[keyof ICancerTypeAlterationPlotData]:string}` ?

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -133,21 +136,21 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
     }
 
     private getLegendNames(id:string) {
```
adamabeshouse: `getLegendName(alterationType:keyof ICancerTypeAlterationPlotData)` (not sure if that typing is possible, but the fn and param name should be changed either way)

src/shared/components/cancerSummary/CancerSummaryContent.tsx
alisman: tried it.  same issue.  index signature must be string or number and for some reason it won't accept that object keys are just strings

src/shared/components/cancerSummary/CancerSummaryContent.tsx
alisman: @computed?

src/shared/components/cancerSummary/SummaryBarGraph.tsx
alisman: yes

src/shared/components/cancerSummary/CancerSummaryContent.tsx
adamabeshouse: thats dumb as heck

alisman: i'm sure there is some way to do it.

PR #646, opened 2017-09-29
Cancer Summary Bar Chart Changes

Tamba32:
Provide white background to chart.  Fix properties of PDF download so now whole chart properly fits on page.  Change ordering of legend and bar stacks.  Allowed slider inputs to now change slider value.

### DIFF
--- a/src/shared/components/cancerSummary/CancerSummaryContent.tsx
+++ b/src/shared/components/cancerSummary/CancerSummaryContent.tsx
@@ -14,6 +14,20 @@ import 'react-rangeslider/lib/index.css';
 
 import SummaryBarGraph from './SummaryBarGraph';
 
+const orderedLabels:Record<keyof ICancerTypeAlterationCounts, string> = {
+    multiple: "Multiple Alterations",
+    protExpressionDown: "Protein Downregulation",
+    protExpressionUp: "Protein Upregulation",
+    mrnaExpressionDown: "mRNA Downregulation",
+    mrnaExpressionUp: "mRNA Upregulation",
+    hetloss: "Shallow Deletion",
+    homdel: "Deep Deletion",
+    gain: "Gain",
+    amp: "Amplification",
+    fusion: "Fusion",
+    mutated: "Mutation"
+};
+
 export interface ICancerTypeAlterationCounts {
     mutated: number;
     amp:number;
@@ -73,8 +87,10 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
 
     private inputYAxisEl:any;
     private inputXAxisEl:any;
+    @observable private tempAltCasesInputValue = 0;
+    @observable private tempTotalCasesInputValue = 0;
     @observable private pngAnchor = '';
-    @observable private pdfAnchor = '';
+    @observable private pdf:{anchor:string;width:number;height:number} = {anchor:'',width: 0,height:0};
     @observable private showControls = false;
     @observable private showGenomicAlt = true;
     @observable private yAxis: 'alt-freq' | 'abs-count' = 'alt-freq';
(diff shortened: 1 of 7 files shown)


src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -219,16 +218,21 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
                                 }
                                 return obj;
                             }, {} as any)
-                            return _.reduce(alterationCounts, (arr, value:{count:number, backgroundColor: string}, label) => {
+                            const reducedAlterations = _.reduce(alterationCounts, (arr, value:{count:number, backgroundColor: string}, label) => {
```
adamabeshouse: more descriptive name for what this actually is would be better

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -219,16 +218,21 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
...
                                     arr.push({
                                         text: this.getLegendNames(label),
-                                        fillStyle: value.backgroundColor,
-                                        index: counter ++
+                                        fillStyle: value.backgroundColor
                                     });
                                 }
                                 return arr;
                             }, [] as any);
+                            const orderedAlts = ['Multiple Alterations', 'Amplification', 'Mutation', 'Fusion', 'Gain',
```
adamabeshouse: it would be best to declare this somewhere else so that it's not rebuilt on every function call

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -125,9 +125,9 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
                     });
                     return memo;
                 }, [] as IBarGraphDataset[]);
+                const orderedAlts = ['multiple', 'amp', 'mutated', 'fusion', 'gain', 'homdel', 'hetloss',
```
adamabeshouse: it would be best to declare this somewhere else so its not rebuilt on every call. Also, it would be great to somehow link this order with the order below to DRY it up

jjgao: +1. The order should be defined once.

jjgao: ... and amp is still not in the correct position.

jjgao:
@Tamba32 please change the order as requested by Niki: https://github.com/cBioPortal/cbioportal/issues/3214

Please note that: he requested to have the same order top-down.

Thanks!

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -219,16 +217,18 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
...
                                 if (value.count) {
                                     arr.push({
                                         text: this.getLegendNames(label),
-                                        fillStyle: value.backgroundColor,
-                                        index: counter ++
+                                        fillStyle: value.backgroundColor
                                     });
                                 }
                                 return arr;
                             }, [] as any);
```
alisman: id you type the accumulator passed to reduce, the ts will consider alterationLabels type and you shouldn't need to type of params passed to sort below

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -115,6 +119,13 @@ export default class SummaryBarGraph extends React.Component<ISummaryBarGraphPro
     }
 
     public componentDidMount() {
+        (Chart as any).plugins.register({
```
alisman: seems like this is applied to all charts then?  
if so, we should see if there's a different way to accomplish this.

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -142,6 +154,22 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
         }, [] as IBarChartSortedData[]);
     }
 
+    private get orderedLabels() {
```
alisman: this should be typed as Record<keyof ICancerTypeAlterationPlotData, string> 
this will make sure it has the right properties.  also, this doesn't need to be a class member (getter).  just assign this to a local variable of the module above the class

src/shared/components/cancerSummary/SummaryBarGraph.tsx
```diff
@@ -17,9 +17,12 @@ interface ISummaryBarGraphProps {
     xAxis: 'y-axis' | 'can-types';
     legend: boolean;
     setPngAnchor:any;
-    setPdfAnchor:any;
+    setPdf:any;
```
alisman: look up in ts documentation how to type these functions.

alisman: also, please give it a more descriptive name.  like, setPDFProperties?

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -124,9 +142,8 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
                     });
                     return memo;
                 }, [] as IBarGraphDataset[]);
+                const orderedAlts = _.keys(orderedLabels);
```
adamabeshouse: is this definitely stable? keys usually isnt order-guaranteed. It would probably be better to make orderedLabels an array of objects `{ alteration: keyof ICancerTypeAlterationCounts, label: string }`

PR #647, opened 2017-09-29
Multiple studyify cancer type summary and mutex

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -193,16 +193,16 @@ export class ResultsViewPageStore {
...
 
-    readonly molecularData = remoteData({
+    readonly molecularData = remoteData<GeneMolecularData[]>({
         await: () => [
             this.studyToDataQueryFilter,
             this.genes,
             this.selectedMolecularProfiles
         ],
-        invoke: async() => {
+        invoke: () => {
```
alisman: why take away async?

adamabeshouse: for code legibility, since nothing async is actually happening

alisman: we do call api here though.  215

adamabeshouse: but we're returning the promise, theres no awaiting going on inside

PR #648, opened 2017-09-29
Show all clinical info in patient view

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/3206


alisman:
I changed this to hofix so we can deploy today.

alisman:
@adamabeshouse I think we should revisit design of this component so that this is never possible.  And also the interface of the table component is confusing.  I know how we got here but lets discuss to see if we can improve it.

PR #650, opened 2017-10-02
Sort clinical info table in alph order

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/3206#issuecomment-333542383


src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -49,7 +49,8 @@ export default class ClinicalInformationPatientTable extends React.Component<ICl
                       {   name:'Attribute',
                           render:(data)=><span>{data.attribute}</span>,
                           filter: (data:IPatientRow, filterString:string, filterStringUpper:string) =>
-                            data.attribute.toString().toUpperCase().indexOf(filterStringUpper) > -1
+                            data.attribute.toString().toUpperCase().indexOf(filterStringUpper) > -1,
```
alisman: i believe that string.includes method is safe to use (b/c we polyfill it).  we can use it instead of indexOf

PR #651, opened 2017-10-02
Oncoprint

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -704,6 +714,37 @@ export class ResultsViewPageStore {
...
+        await: ()=>[
+            this.molecularProfilesInStudies
+        ],
+        invoke:()=>{
+            const MRNA_EXPRESSION = "MRNA_EXPRESSION";
+            const PROTEIN_LEVEL = "PROTEIN_LEVEL";
+            const selectedMolecularProfileIds = stringListToSet(this.selectedMolecularProfileIds);
+
+            return Promise.resolve(_.sortBy(_.filter(this.molecularProfilesInStudies.result!, profile=>{
+                return (profile.molecularAlterationType === MRNA_EXPRESSION ||
```
alisman: lets try to use map for these.

src/shared/components/editableSpan/EditableSpan.tsx
```diff
@@ -24,45 +36,109 @@ export default class EditableSpan extends React.Component<IEditableSpanProps, {}
...
     render() {
-        const {contentEditable, onKeyPress, onBlur,
+        const {contentEditable, onKeyPress, onKeyUp, onBlur,
             value, setValue, maxChars,
-            numericOnly, ...spanProps} = this.props;
+            numericOnly, isPercentage,
+            textFieldAppearance,
+            style, ...spanProps} = this.props;
+        let spanStyle = {};
+        if (this.props.textFieldAppearance) {
```
alisman: rather than complicating the component with this kind of conditional style stuff, why not just create and apply a css class?

package.json
```diff
@@ -130,6 +132,7 @@
     "mobx-utils": "^2.0.1",
     "mobxpromise": "^1.2.0",
     "node-sass": "^4.3.0",
+    "oncoprintjs": "file:///Users/abeshoua/cbioportal/oncoprintjs",
```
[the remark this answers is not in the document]
adamabeshouse: this is actually just for development so I dont have to keep publishing updates to npm

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: ?

src/shared/components/editableSpan/EditableSpan.tsx
adamabeshouse: the actual reason why is that I'm not as comfortable/familiar with how scss module magic works. But I can do that.

src/shared/lib/onMobxPromise.ts
```diff
@@ -1,15 +1,32 @@
 import {MobxPromise} from "mobxpromise";
 import {autorun, IReactionDisposer} from "mobx";
 
-export default function onMobxPromise<T>(promise:MobxPromise<T>,
-                                         onComplete:(result:T)=>void,
+export function onMobxPromise<T>(promise:MobxPromise<T>,
+                          onComplete:(result:T)=>void,
```
alisman: can onComplete fire multiple times as observed MobxPromise go from pending to complet to pending to complete?

alisman: onError?

src/shared/components/editableSpan/EditableSpan.tsx
alisman: You wouldn't actually have to use css modules.  I would recommend that EditableSpan accept a custom class name.  Then for a particular instance, you define that class in the feature's scss file.  Let me know if you want more detail.

src/pages/resultsView/ResultsViewPageStore.ts
alisman: I mean the constants like MRNA_EXPRESSION should come from one place in our app

src/shared/lib/onMobxPromise.ts
adamabeshouse: thats what the `times` parameter determines

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: oh yeah. do they?

adamabeshouse: good point

src/shared/lib/onMobxPromise.ts
adamabeshouse: I havent had a need for that yet so I didn't implement it

PR #662, opened 2017-10-02
Fix #3138 Move uniprot link to right

inodb:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3138

Disable tooltips on gray segments. Move uniprot link to the right

screen shot 2017-10-02 at 6 55 54 pm

Another idea would be to move uniprot link in a `?` so these details only show up for those who are interested about the details

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapper.tsx
+++ b/src/pages/resultsView/mutation/MutationMapper.tsx
@@ -67,8 +67,11 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
 
     @computed get geneSummary():JSX.Element {
         return (
-            <div>
+            <div style={{'paddingBottom':10}}>
                 <h4>{this.props.store.gene.hugoGeneSymbol}</h4>
+                {this.props.store.uniprotId.result && (
+                    <span>UniProt: <a href={`http://www.uniprot.org/uniprot/${this.props.store.uniprotId.result}`}>{this.props.store.uniprotId.result}</a></span>
+                )}
             </div>
         );
     }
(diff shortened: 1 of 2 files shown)


jjgao:
Another option is to append after the gene symbol separate by `|`. We can add ENSG and ENST (maybe on the 2nd line) later.

schultzn:
But shouldn’t we distinguish between the gene itself and its isoforms? I.e., TP53 is the gene, but P53_HUMAN, NM_…, ENST… all refer to specific isoforms? Just an argument for not placing them into the same line.

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -67,8 +67,13 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
 
     @computed get geneSummary():JSX.Element {
         return (
-            <div>
+            <div style={{'paddingBottom':10}}>
                 <h4>{this.props.store.gene.hugoGeneSymbol}</h4>
+                {this.props.store.uniprotId.result?
```
alisman: yeah, lets switch to (condition) && (<whatever />)

src/pages/resultsView/mutation/MutationMapper.tsx
```diff
@@ -67,8 +67,13 @@ export default class MutationMapper extends React.Component<IMutationMapperProps
 
     @computed get geneSummary():JSX.Element {
         return (
-            <div>
+            <div style={{'paddingBottom':10}}>
```
alisman: what's bottom padding for?

inodb: @alisman  have some space before `Somatic Mutation Frequency`

PR #663, opened 2017-10-03
Fix to remove hard coded URLs and add https

pieterlukasse:
# What? Why?
This removes hard coded references to www.cbioportal.org and adds https to the external service calls that don't go via the internal proxy (this is important for cbioportal installations hosted on a https address).

# Notify reviewers
@adamabeshouse @alisman

### DIFF
--- a/src/pages/datasetView/DatasetList.tsx
+++ b/src/pages/datasetView/DatasetList.tsx
@@ -49,7 +49,7 @@ class CancerStudyCell extends React.Component<ICancerStudyCellProps,{}> {
         return (
             <span>
                 <a
-                    href={`http://www.cbioportal.org/study?id=${this.props.studyId}#summary`}
+                    href={`study?id=${this.props.studyId}#summary`}
                     target='_blank'
                 >
                     {this.props.name}
--- a/src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
@@ -95,7 +95,7 @@ export default class CosmicColumnFormatter
     {
         const cosmic:CosmicMutation[]|null = CosmicColumnFormatter.getData(data, cosmicData);
 
-        let value:number = -1;
+        let value:number|undefined = -1;
         let display:string = "";
         let overlay:JSX.Element|null = null;
         let content:JSX.Element;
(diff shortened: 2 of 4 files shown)


src/pages/datasetView/DatasetList.tsx
```diff
@@ -49,7 +49,7 @@ class CancerStudyCell extends React.Component<ICancerStudyCellProps,{}> {
         return (
             <span>
                 <a
-                    href={`http://www.cbioportal.org/study?id=${this.props.studyId}#summary`}
```
pieterlukasse: @inodb @alisman: it seems this hard coded reference to www.cbioportal.org has been there for quite some time. At the same time, the datasets page seems to work correctly on a local installation. How is this possible? Is there a global search+replace going on during build time?

src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
```diff
@@ -95,7 +95,7 @@ export default class CosmicColumnFormatter
     {
         const cosmic:CosmicMutation[]|null = CosmicColumnFormatter.getData(data, cosmicData);
 
-        let value:number = -1;
+        let value:number|undefined = -1;
```
pieterlukasse: another question is this compilation error fix: isn't the whole project compiled at each commit? How could a compilation error creep in?

src/pages/datasetView/DatasetList.tsx
alisman: look like you had an old version of this file.  In the below, you can see this was fixed.  On the other hand, there are other references to our datahub.  Not sure if those should also be configurable?

https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/datasetView/DatasetList.tsx

src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
alisman: hmm.  you should not need to allow undefined here as far as i can see.  what is compiler error?

src/pages/datasetView/DatasetList.tsx
pieterlukasse: indeed old code. Sorry for alarming you. Will make a new PR.

src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
pieterlukasse: was old code....didn't encounter this in new PR #664

PR #671, opened 2017-10-04
Update API clients

inodb:


onursumer:
@inodb Seems like mutation model was updated at some point, and also the Civic mock data, interesting...

PR #680, opened 2017-10-06
Removed legend display for case where genomic alterations aren't showing

Tamba32:


alisman:
What is case for this?

PR #685, opened 2017-10-10
Fix pdf sizing.

Tamba32:


src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -343,8 +343,12 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
 
     private downloadPdf() {
         const {anchor, width, height} = this.pdf;
-        const pdf = new jsPDF({orientation: 'l', unit:'mm', format:[width + 20, height + 20]})
-            .addImage(anchor, 'JPEG', 0,0)
+        let orientation = 'p';
```
alisman: no need to change b/c it's a super minor quibble but ternary would actually make this clearer.  again, no need to do anything

PR #690, opened 2017-10-11
Query Page: configurable list of sets of genes

oplantalech:
# What? Why?
We have created a property in portal.properties (`querypage.setsofgenes.location`) where the user can specify a different file that describes the list of sets of genes displayed in the query page. This file must be a JSON string.

This PR goes together with the PR cbioportal/cbioportal#3249.

### DIFF
--- a/src/config/production.config.ts
+++ b/src/config/production.config.ts
@@ -22,7 +22,8 @@ const config:IAppConfig = {
     skinRightNavShowExamples: (window as any).skinRightNavShowExamples, 
     skinRightNavShowTestimonials: (window as any).skinRightNavShowTestimonials,
     skinRightNavExamplesHTML: (window as any).skinRightNavExamplesHTML,
-    skinRightNavWhatsNewBlurb: (window as any).skinRightNavWhatsNewBlurb
+    skinRightNavWhatsNewBlurb: (window as any).skinRightNavWhatsNewBlurb,
+    querySetsOfGenes: (window as any).querySetsOfGenes
 };
 
 export default config;
--- a/src/shared/components/query/GeneSetSelector.tsx
+++ b/src/shared/components/query/GeneSetSelector.tsx
@@ -13,6 +13,7 @@ import {QueryStoreComponent} from "./QueryStore";
 import MutSigGeneSelector from "./MutSigGeneSelector";
 import GisticGeneSelector from "./GisticGeneSelector";
 import SectionHeader from "../sectionHeader/SectionHeader";
+import AppConfig from "appConfig";
 
 const styles = styles_any as {
 	GeneSetSelector: string,
@@ -40,16 +41,21 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
 
 	@computed get geneListOptions()
 	{
-		return [
-			{
-				label: 'User-defined List',
-				value: ''
-			},
-			...gene_lists.map(item => ({
-				label: `${item.id} (${item.genes.length} genes)`,
-				value: item.genes.join(' ')
-			}))
-		];
+	    let geneList: {"id": string, "genes": string[]}[] = gene_lists;
+	    if (AppConfig.querySetsOfGenes) {
+	        geneList = AppConfig.querySetsOfGenes;
+	    }
+
+	    return [
+	        {
+	            label: 'User-defined List',
+	            value: ''
+	        },
+	        ...geneList.map(item => ({
+	            label: `${item.id} (${item.genes.length} genes)`,
+	            value: item.genes.join(' ')
+	        }))
+	    ];
 	}
 
 	@computed get textAreaRef()
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -23,7 +23,8 @@ export interface IAppConfig {
     skinRightNavShowExamples?: boolean, 
     skinRightNavShowTestimonials?: boolean,
     skinRightNavExamplesHTML?: string,
-    skinRightNavWhatsNewBlurb?: string 
+    skinRightNavWhatsNewBlurb?: string,
+    querySetsOfGenes?: {"id": string, "genes":string[]}[]
 }
 
 export type PriorityStudies = {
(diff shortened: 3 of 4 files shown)


src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -45,7 +45,7 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
 				label: 'User-defined List',
 				value: ''
 			},
-			...gene_lists.map(item => ({
+			...AppConfig.querySetsOfGenes == undefined ? [{label:"", value:""}] : AppConfig.querySetsOfGenes.map(item => ({
```
inodb: why not `AppConfig.querySetsOfGenes?`

src/config/production.config.ts
```diff
@@ -22,7 +22,8 @@ const config:IAppConfig = {
     skinRightNavShowExamples: (window as any).skinRightNavShowExamples, 
     skinRightNavShowTestimonials: (window as any).skinRightNavShowTestimonials,
     skinRightNavExamplesHTML: (window as any).skinRightNavExamplesHTML,
-    skinRightNavWhatsNewBlurb: (window as any).skinRightNavWhatsNewBlurb
+    skinRightNavWhatsNewBlurb: (window as any).skinRightNavWhatsNewBlurb,
+    querySetsOfGenes: (window as any).querySetsOfGenes
```
inodb: There's also a `development.config.ts` you should update. It's a little wonky but these are basically the same files. We should prolly change it to just read one config file.

src/shared/components/query/gene_lists.ts
```diff
@@ -1,72 +0,0 @@
-const gene_lists = [{
```
inodb: How do you handle the default on the frontend now? We would like everything to work also outside of `cbioportal` war, so that at one point we can really separate them. If gene lists are only used in the new frontend maybe better to only store the default here. If they are used in both places, might be good to store a default in both places. If we only store it in `cbioportal` repo, I don't see how we can make the frontend work standalone.

src/shared/components/query/GeneSetSelector.tsx
oplantalech: You can take a look here: https://reactjs.org/docs/jsx-in-depth.html#spread-attributes

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -40,16 +41,29 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
...
+                        value: item.genes.join(' ')
+                    }))
+                ];
+	    } else {
+	        return [
+	                {
+	                    label: 'User-defined List',
+	                    value: ''
+	                },
+	                ...AppConfig.querySetsOfGenes == undefined ? [{label:"", value:""}] : AppConfig.querySetsOfGenes.map(item => ({
```
alisman: curious why we want [{label:"", value:""}]  if querySetsofGenes is undefined.

oplantalech: This is an error, the first if should contain `null` and `undefined`.

src/config/production.config.ts
oplantalech: I've updated the `development.config.ts`, too.

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -40,16 +41,29 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
...
-			{
-				label: 'User-defined List',
-				value: ''
-			},
-			...gene_lists.map(item => ({
-				label: `${item.id} (${item.genes.length} genes)`,
-				value: item.genes.join(' ')
-			}))
-		];
+	    if (AppConfig.querySetsOfGenes == null || AppConfig.querySetsOfGenes == undefined) {
```
ersinciftci: I think you can just say `if (AppConfig. querySetsOfGenes == null)`, it covers `undefined` too as long as you don't use `===`

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -40,16 +41,29 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
...
+                        value: item.genes.join(' ')
+                    }))
+                ];
+	    } else {
+	        return [
+	                {
+	                    label: 'User-defined List',
+	                    value: ''
+	                },
+	                ...AppConfig.querySetsOfGenes.map(item => ({
```
ersinciftci: Seems like these 2 cases are copy/paste of each other besides the list name, can be improved by creating a local variable for the list in this function, something like:
```
let geneList = gene_lists;
if (AppConfig. querySetsOfGenes) {
    geneList = AppConfig. querySetsOfGenes;
}

return [
	{
		label: 'User-defined List',
		value: ''
	},
	...geneList.map(item => ({
		label: `${item.id} (${item.genes.length} genes)`,
		value: item.genes.join(' ')
	}))
];

File: package-lock.json [hunk 7]
```diff
@@ -35,7 +35,7 @@
     "@types/cheerio": {
       "version": "0.22.2",
       "resolved": "https://registry.npmjs.org/@types/cheerio/-/cheerio-0.22.2.tgz",
-      "integrity": "sha512-7YJQbllem83umRzZ+ry6RZFAbBWNoHvbiW6tE5VEAJIoX6F22hrKrBw3ibRvHm2zPnOhZGrZJusAbp1zZKcLzQ=="
+      "integrity": "sha1-U5Ylh0vIVghq1JHC/cmxDAWuMI4="
```
ersinciftci: I think you use a different version of npm, so the sha algorithm is different. So I think you should remove this file from the PR, and try to upgrade your npm.

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -40,16 +41,21 @@ export default class GeneSetSelector extends QueryStoreComponent<GeneSetSelector
...
-				label: 'User-defined List',
-				value: ''
-			},
-			...gene_lists.map(item => ({
-				label: `${item.id} (${item.genes.length} genes)`,
-				value: item.genes.join(' ')
-			}))
-		];
+	    let geneList: {"id": string, "genes": string[]}[] = gene_lists;
+	    if (AppConfig. querySetsOfGenes) {
```
ersinciftci: There is an extra white space after `.`, not sure if it effects the execution.

oplantalech: hmm, that shouldn't be here (although I tested it and it was working). I will fix it.

PR #691, opened 2017-10-12
Temporary patch to use gene length when no PFAM domain information available

onursumer:
# What? Why?
- Partial fix for https://github.com/cBioPortal/cbioportal/issues/3218.

unknown_entrez_gene

- In case of an error with the PFAM graphics API, using the `gene length / 3` approach to determine the protein length. This prevents infinite spinning of the lollipop diagram loader

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapperStore.ts
+++ b/src/pages/resultsView/mutation/MutationMapperStore.ts
@@ -94,6 +94,12 @@ export class MutationMapperStore {
 
     readonly swissProtId = remoteData({
         invoke: async() => {
+            // do not try fetching swissprot data for invalid entrez gene ids,
+            // just return the default value
+            if (this.gene.entrezGeneId < 1) {
+                return "";
+            }
+
             const accession:string|string[] = await fetchSwissProtAccession(this.gene.entrezGeneId);
 
             if (_.isArray(accession)) {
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -94,6 +94,10 @@ export class MutationMapperStore {
 
     readonly swissProtId = remoteData({
         invoke: async() => {
+            if (this.gene.entrezGeneId < 1) {
+                return "";
```
adamabeshouse: can you add a comment explaining the logic here

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -468,7 +468,7 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
                         dataStore={this.props.store.dataStore}
                         vizWidth={this.props.geneWidth}
                         vizHeight={130}
-                        xMax={this.props.store.pfamGeneData.result.length}
+                        xMax={this.props.store.pfamGeneData.result.length || this.props.store.gene.length / 3}
```
adamabeshouse: order of operations stuff makes me uneasy.. can you put parens around the division

PR #692, opened 2017-10-13
Survival Tab implementation with Chart.js

ersinciftci:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -451,6 +452,15 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly filteredAlterationsAsUniquePatientKeyArrays = remoteData({
+        await: () => [
+            this.filteredAlterations
+        ],
+        invoke: async() => {
+            return _.mapValues(this.filteredAlterations.result, (mutations: Mutation[]) => _.map(mutations, 'uniquePatientKey'));
+        }
```
alisman: minor but for map i like to use (whatever)=>whatever.property
to get typechecking on the prop name

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -570,32 +602,126 @@ export class ResultsViewPageStore {
...
+                    const statusClinicalData: ClinicalData[] = clinicalData.filter(c => c.clinicalAttributeId === statusAttributeId);
+                    const monthsClinicalData: ClinicalData[] = clinicalData.filter(c => c.clinicalAttributeId === monthsAttributeId);
+                    if (statusClinicalData[0] && monthsClinicalData[0] && statusClinicalData[0].value != 'NA' &&
+                        monthsClinicalData[0].value != 'NA') {
+                        const patient: Patient = this.patients.result.filter(p => p.uniquePatientKey === uniquePatientKey)[0];
+                        patientSurvivals.push({
+                            patientId: patient.patientId,
+                            studyId: patient.studyId,
+                            status: statusFilter(statusClinicalData[0].value),
+                            months: Number.parseFloat(monthsClinicalData[0].value)
```
alisman: I don't think Number.parseFloat is not supported by IE.  It's possible that it's polyfilled by Babel, but I'm not sure.  You can probably just use the global parseFloat.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -570,32 +602,126 @@ export class ResultsViewPageStore {
...
+    readonly survivalClinicalDataGroupByUniquePatientKey = remoteData<Dictionary<ClinicalData[]>>({
+        await: () => [
+            this.survivalClinicalData,
+        ],
+        invoke: async() => {
+            return _.groupBy(this.survivalClinicalData.result, 'uniquePatientKey');
+        }
+    });
+
+    private getPatientSurvivals(targetUniquePatientKeys: string[], statusAttributeId: string, monthsAttributeId: string,
```
alisman: just for discussion: this method relies on async data but has no mechanism to wait for it.  this sort of bothers me.  one solution would be to just pass in the data (e.g. survivalClinicalDataGroupByUniquePatientKey and patient) so that the moxPromise would then be responisble for awaiting that data and feeding in to this method (which would become more pure)

PR #694, opened 2017-10-16
Show Twitter Embed

inodb:
Fix https://github.com/cbioportal/cbioportal/issues/3115

Do we like something like this?

screen shot 2017-10-16 at 1 45 22 pm

We can add a prop to change the twitter handle if we want. This is just overriding default.

In the spec we mentioned 3 tweets, but yeah with a scrollbar it might be ok to do more as I've done in this example. How many tweets are visible depends on the size of the tweets themselves.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3971,6 +3971,11 @@
         "original": "1.0.0"
       }
     },
+    "exenv": {
+      "version": "1.2.2",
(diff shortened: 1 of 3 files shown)


ecerami:
@inodb looks good to me.  tx

schultzn:
Agreed, looks great. We should always have Ethan’s picture up there. :-)

PR #700, opened 2017-10-17
Temporary workaround for missing residue mappings

onursumer:
# What? Why?
Temporary fix for https://github.com/cBioPortal/cbioportal/issues/3259

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/StoreUtils.ts
+++ b/src/shared/lib/StoreUtils.ts
@@ -76,7 +76,13 @@ export async function fetchPdbAlignmentData(uniprotId: string,
                                             client:Genome2StructureAPI = g2sClient)
 {
     if (uniprotId) {
-        return await client.getAlignmentUsingGET({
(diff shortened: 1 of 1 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -76,7 +76,13 @@ export async function fetchPdbAlignmentData(uniprotId: string,
...
 {
     if (uniprotId) {
-        return await client.getAlignmentUsingGET({
+        // return await client.getAlignmentUsingGET({
+        //     idType: "uniprot",
+        //     id: uniprotId
+        // });
+
+        // TODO temporary workaround to exclude alignments with no residue mapping data
+        return await client.postResidueMappingUsingGET({
```
inodb: @onursumer maybe good to explain a bit more clearly why this works and the other doesn't. Anyway good for now

PR #704, opened 2017-10-18
Only display the CNA variant shown in the table in CIViC tooltip

### LINKED ISSUE #674: Issue with CIVIC and CNA data?
I just noticed an example of an unexpected CIVIC annotation. The alteration is AMP but the CIVIC information is about a DELETION:

image

I thought we were using AMP/DEL info to find a correct match, and if a correct match was not found but some other information about the gene *was* found, we would make the icon grey?

oplantalech:
# What? Why?
Fix #674.

### DIFF
--- a/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.spec.tsx
+++ b/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.spec.tsx
@@ -55,7 +55,7 @@ describe('AnnotationColumnFormatter', () => {
         
         assert.deepEqual(
             AnnotationColumnFormatter.hasCivicVariants(cna, civicGenes, civicVariants), false,
-            'Civic has variants');
+            'Civic has no variants');
     });
 
 
--- a/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
+++ b/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
@@ -11,7 +11,7 @@ import {generateQueryVariantId, generateQueryVariant} from "shared/lib/OncoKbUti
 import {IndicatorQueryResp, Query} from "shared/api/generated/OncoKbAPI";
 import {getAlterationString} from "shared/lib/CopyNumberUtils";
 import {ICivicVariant, ICivicGene, ICivicEntry, ICivicVariantData, ICivicGeneData, ICivicGeneDataWrapper, ICivicVariantDataWrapper} from "shared/model/Civic.ts";
-import {buildCivicEntry} from "shared/lib/CivicUtils";
+import {buildCivicEntry, getCivicCNAVariants} from "shared/lib/CivicUtils";
 
 /**
  * @author Selcuk Onur Sumer
@@ -73,10 +73,10 @@ export default class AnnotationColumnFormatter
     {
         let civicEntry = null;
         let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
-        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneVariants:{[name: string]: ICivicVariantData} = getCivicCNAVariants(copyNumberData, geneSymbol, civicVariants);
         let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
         //Only return data for genes with variants or it has a description provided by the Civic API
-        if (geneVariants || (geneEntry && geneEntry.description !== "")) {
+        if (!_.isEmpty(geneVariants) || geneEntry && geneEntry.description !== "") {
             civicEntry = buildCivicEntry(geneEntry, geneVariants);
         }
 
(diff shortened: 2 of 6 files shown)


src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -73,10 +73,10 @@ export default class AnnotationColumnFormatter
     public static hasCivicVariants (copyNumberData:DiscreteCopyNumberData[], civicGenes:ICivicGene, civicVariants:ICivicVariant): boolean
     {
         let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
-        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneVariants:{[name: string]: ICivicVariantData} = getCivicCNAVariants(copyNumberData, geneSymbol, civicVariants);
         let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
 
-        if (geneEntry && !geneVariants) {
+        if (geneEntry && Object.keys(geneVariants).length === 0) {
```
onursumer: I would prefer `_.isEmpty(geneVariants)` here

src/shared/components/annotation/CivicCard.tsx
```diff
@@ -25,15 +25,19 @@ export default class CivicCard extends React.Component<ICivicCardProps, {}> {
...
 
         if (variantMap) {
-            for (let name in variantMap) {
-                let variant = variantMap[name];
-                let entryTypes: string = '';
-                for (let evidenceType in variant.evidence) {
-                    entryTypes += evidenceType.toLowerCase() + ': ' + variant.evidence[evidenceType] + ', ';
-                }
-                entryTypes = entryTypes.slice(0, -2) + '.';
+            if (Object.keys(variantMap).length === 0) {
```
onursumer: `_.isEmpty(variantMap)`

src/shared/lib/CivicUtils.ts
```diff
@@ -134,3 +135,21 @@ export function buildCivicEntry(geneEntry: ICivicGeneData, geneVariants: {[name:
         variants: geneVariants
     };
 }
+
+export function getCivicCNAVariants(copyNumberData:DiscreteCopyNumberData[], geneSymbol: string, civicVariants:ICivicVariant): {[name: string]: ICivicVariantData} {
+    let geneVariants: {[name: string]: ICivicVariantData} = {};
+    if (copyNumberData[0].alteration == 2) {
+        for (let alteration in civicVariants[geneSymbol]) {
+            if (alteration == "AMPLIFICATION") {
```
onursumer: In general, `===` is safer than `==`

inodb: @oplantalech Yes I mentioned this on a previous PR as well I believe, I don't think there's really any reason to use `==` in JS, always use `===`

oplantalech:
@onursumer @inodb Can you approve and merge? Thank you.

oplantalech:
@inodb It should be fixed now, so can you merge the PR?

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.spec.tsx
```diff
@@ -25,6 +25,8 @@ describe('AnnotationColumnFormatter', () => {
         let cna = getCnaData();
 
         let expectedCivicEntry = getExpectedCnaCivicEntry();
+        
+        console.log(AnnotationColumnFormatter.getCivicEntry(cna, civicGenes, civicVariants));
```
onursumer: I think we should try to avoid using `console.log`

src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx
```diff
@@ -73,10 +73,10 @@ export default class AnnotationColumnFormatter
     {
         let civicEntry = null;
         let geneSymbol: string = copyNumberData[0].gene.hugoGeneSymbol;
-        let geneVariants: {[name: string]: ICivicVariantData} = civicVariants[geneSymbol];
+        let geneVariants:{[name: string]: ICivicVariantData} = getCivicCNAVariants(copyNumberData, geneSymbol, civicVariants);
         let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
         //Only return data for genes with variants or it has a description provided by the Civic API
-        if (geneVariants || (geneEntry && geneEntry.description !== "")) {
+        if ((Object.keys(geneVariants).length !== 0) || geneEntry && geneEntry.description !== "") {
```
onursumer: `_.isEmpty(geneVariants)`?

src/shared/components/annotation/CivicCard.spec.tsx
```diff
@@ -35,7 +35,7 @@ describe('CivicCard', () => {
 
     it('should not have variant', () => {
         const wrapper = shallow(<CivicCard {...props}/>);
-        expect(wrapper.find('.civic-card-variant')).to.have.length(0);
+        expect(wrapper.find('.civic-card-variant')).to.have.length(1);
```
onursumer: So, the unit test definition claims that `it should not have variant`, but then we expect to have 1 `civic-card-variant`. Seems a bit contradicting to me.

onursumer:
@oplantalech I just requested a few more small changes, can you look into them as well? Thanks.

src/shared/components/annotation/CivicCard.spec.tsx
oplantalech: Now, with the new implementation, `civic-card-variant` is created when there are no variants, only with the `civic-card-variant-description summary` item. However, other items of this part like `civic-card-variant-header`, `civic-card-variant-name` and `civic-card-variant-entry-types` are not created, so I am going to add this three parts as the test to check if there is a variant or not, I think it would make more sense.

oplantalech:
Thank you @onursumer for your comments. Now everything should be corrected.

PR #705, opened 2017-10-18
Cancer Summary 100% max y-axis

Tamba32:
Reimplement maximum for cancer type summary of 100% when y-axis is alteration frequency and max data point is above 90%.


jjgao:
Should this go to hotfix instead?

src/shared/components/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -476,7 +476,7 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
                 {controls}
                 <SummaryBarGraph data={this.chartData} yAxis={this.yAxis} xAxis={this.xAxis} gene={this.props.gene} width={this.props.width}
                                  setPdfProperties={this.setPdfProperties} setPngAnchor={this.setPngAnchor} legend={this.showGenomicAlt}
-                                 orderedLabels={orderedLabels}
+                                 orderedLabels={orderedLabels} altCasesMax={this.altCasesMax}
```
alisman: Seems like it would be better to derive this from the chartData passed in, but I guess we're already calculating this outside?

alisman: In any case, we should bother because we're probably going to use a new library and hopefully that library will not try to pad a percentage axis above 100%

PR #706, opened 2017-10-18
use proteinChange property of Mutation instead of aminoAcidChange

alisman:
The new api doesn't return aminoAcidChange property on Mutations.   We need to use the property proteinChange instead.

### DIFF
--- a/src/shared/lib/oql/accessors.ts
+++ b/src/shared/lib/oql/accessors.ts
@@ -99,7 +99,7 @@ export default class accessors {
         if (this.alterationType(d.molecularProfileId) === 'MUTATION_EXTENDED') {
             if (d.mutationType === "fusion") {
                 return null;
-            } else if (d.aminoAcidChange && d.aminoAcidChange.toLowerCase() === "promoter") {
+            } else if (d.proteinChange && d.proteinChange.toLowerCase() === "promoter") {
                 return "promoter";
             } else {
                 return getSimplifiedMutationType(d.mutationType);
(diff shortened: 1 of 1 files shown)


src/shared/lib/oql/accessors.ts
```diff
@@ -99,7 +99,7 @@ export default class accessors {
         if (this.alterationType(d.molecularProfileId) === 'MUTATION_EXTENDED') {
             if (d.mutationType === "fusion") {
                 return null;
-            } else if (d.aminoAcidChange && d.aminoAcidChange.toLowerCase() === "promoter") {
+            } else if (d.proteinChange && d.proteinChange.toLowerCase() === "promoter") {
```
alisman: @jj @onursumer this is not related to the bug but i decided to change this reference to aminoAcidChange as well.  Let me know if that's right or not.

src/shared/lib/oql/accessors.ts
```diff
@@ -125,7 +125,7 @@ export default class accessors {
 
     public mut_amino_acid_change(d: Mutation) {
         if (this.alterationType(d.molecularProfileId) === 'MUTATION_EXTENDED') {
```
alisman: we need to refactor oqlFilter.js to typescript and deal with the null case in some fashion.

src/shared/lib/oql/accessors.ts
onursumer: @alisman looks right to me

jjgao:
Not sure why travis fails. Please feel free to merge the release if code works. @inodb

PR #709, opened 2017-10-19
We should fallback to study cancer type when there is no CANCER_TYPE clinData

alisman:
image

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1,7 +1,7 @@
 import {
     DiscreteCopyNumberFilter, DiscreteCopyNumberData, ClinicalData, ClinicalDataMultiStudyFilter, Sample,
     SampleIdentifier, MolecularProfile, Mutation, GeneMolecularData, MolecularDataFilter, Gene,
-    ClinicalDataSingleStudyFilter
+    ClinicalDataSingleStudyFilter, CancerStudy
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
 import {computed, observable, action} from "mobx";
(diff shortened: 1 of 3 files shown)


alisman:
The second commit is the beginning of a tool I'm using to generate test boilerplate

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -49,12 +50,12 @@ export const AlterationTypeConstants = {
     FUSION: 'FUSION'
 }
 
-interface ExtendedAlteration extends Mutation, GeneMolecularData {
+export interface ExtendedAlteration extends Mutation, GeneMolecularData {
```
alisman: We have to figure out where we should put these kind of suplemmentary types.  They don't belong just wherever we first use them.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -175,6 +176,55 @@ export function countAlterationOccurences(samplesByCancerType: {[cancerType: str
...
+            }
+        }
+        if (sample.cancerType && !sample.cancerTypeDetailed) {
+            sample.cancerTypeDetailed = sample.cancerType;
+        }
+    });
+
+    return extendedSamples;
+
+}
```
adamabeshouse: maybe for clarity (I dont know if necessary) but we could have
```
export interface ExtendedSample extends Sample {
    clinCancerType: string;
    clinCancerTypeDetailed: string;
    studyCancerType: string;
    cancerType: string; // derived based on which of the above three are available
}
```

This may be useful?

alisman: I see.  Yeah.  I like it.

PR #717, opened 2017-10-25
Use user logged in email address when fill in OncoKB feedback form

zhx828:
Signed-off-by: Hongxin Zhang <jackson.zhang.828@gmail.com>

# What? Why?
Fix #3232.

# Checks
  the `-s` option)

# Notify reviewers
@cBioPortal/frontend

### DIFF
--- a/src/shared/components/annotation/OncoKB.tsx
+++ b/src/shared/components/annotation/OncoKB.tsx
@@ -17,6 +17,7 @@ import OncoKbEvidenceCache from "shared/cache/OncoKbEvidenceCache";
 import OncoKbTooltip from "./OncoKbTooltip";
 import OncokbPubMedCache from "shared/cache/PubMedCache";
 import {default as TableCellStatusIndicator, TableCellStatus} from "shared/components/TableCellStatus";
+import AppConfig from "appConfig";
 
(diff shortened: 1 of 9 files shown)


src/shared/components/annotation/OncoKB.tsx
```diff
@@ -17,6 +17,7 @@ import OncoKbEvidenceCache from "shared/cache/OncoKbEvidenceCache";
 import OncoKbTooltip from "./OncoKbTooltip";
 import OncokbPubMedCache from "shared/cache/PubMedCache";
 import {default as TableCellStatusIndicator, TableCellStatus} from "shared/components/TableCellStatus";
+import AppConfig from "appConfig";
```
onursumer: I think it is better to make `userEmailAddress` a part of `IOncoKbProps`. In general we should try to avoid "injecting" `AppConfig` into the shared components, since it will cause problems when we want to separate certain components out of this repository.

zhx828: Only if other parts will not use this property. I'm so sure about all features.

zhx828:
@onursumer @inodb updated the pull request from the suggestions. Could you review it again? Thanks!

PR #723, opened 2017-11-02
Fix frontendConfig passing

inodb:
There were still some places that were directly acessing window instead of
AppConfig.


alisman:
Not critical but are any of the props in myindex.ejs file (template for html file that we sometimes use in dev) involved (and going to break)?

PR #724, opened 2017-11-02
Add Genome Driven Diagnosis on Patient View

inodb:
Uses an internal service to get the cancer type prediction for each impact
case. This cancer type prediction is called Genome Driven Diagnosis and is a
predicted probability for 22 tumor types based on genomic features, using a
calibrated RandomForest algorithm. We would like to indicate mismatches between
these and pathology based assessments of cancer type. Furthermore it's useful
for samples of unkown primary. Currently the service will only be available
internally at MSKCC.

The coloring follows @anoopbr's thresholds, so they are the same in CVR.

The GDD predicted cancer type shows up in the patient view header when it
differs from the diagnosed cancer type.

In red if it's a high confident
mismatch (>75%),

screen shot 2018-02-16 at 5 17 16 pm

 in orange if it's a medium confident mismatch (>50%),

screen shot 2018-02-16 at 5 15 45 pm

otherwise the text shows up in blue.The top bar in the tooltip is colored similarly
(just gray instead of blue for low confident calls).

screen shot 2018-02-16 at 5 18 19 pm

When the cancer type
matches the diagnosed one, the high confident match bar is colored green.

screen shot 2018-02-16 at 7 43 33 pm

When the cancer type
matches the diagnosed one, the medium confident match bar is colored blue.

screen shot 2018-02-16 at 6 41 13 pm

When the diagnosed cancer type is not part of the 22 cancer types a warning
message is displayed in the tooltip.

screen shot 2018-02-16 at 5 18 53 pm

Similarly when the diagnosed cancer type
is undefined.
screen shot 2018-02-16 at 5 40 50 pm

See example:
https://cbioportal.mskcc.org/beta/case.do#/patient?caseId=P-0000004&navCaseIds=P-0000004%2CP-0000015%2CP-0000025%2CP-0000027%2CP-0000037&studyId=msk_impact_2017


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -187,6 +188,18 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                     sampleManager.clinicalDataLegacyCleanAndDerived[sample.id] &&
                     sampleManager.clinicalDataLegacyCleanAndDerived[sample.id].DERIVED_NORMALIZED_CASE_TYPE === 'Xenograft'
                 );
+				// check whether the sample has a Genomic Driven Diagnosis
+				// prediction
+				const hasGDDPred:boolean = (
```
onursumer: Seems like there is too much indentation here

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -210,6 +223,11 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                 )
                             }
                         </span>
+						{hasGDDPred && patientViewPageStore.sampleIdToTumorType &&
```
onursumer: here, too :)

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,85 @@
...
+        return (
+            <div>
+                <h6 style={{textAlign:"center"}}>
+                    Genome Driven Diagnosis (GDD)
+                    <DefaultTooltip
+                        mouseEnterDelay={0}
+                        placement="right"
+                        overlay={<span>Predicted probability for 22 tumor types
+                            based on genomic features, using a calibrated
+                            RandomForest algorithm.</span>}
```
onursumer: Minor aesthetic detail, but I would style it this way :)

```
overlay = {
    <span>
        Predicted probability for 22 tumor types based on genomic features, 
        using a calibrated RandomForest algorithm.
    </span>
}
```

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,85 @@
...
+                    >
+                        <span style={{paddingLeft: 2}}><FontAwesome name='question-circle'/></span>
+                    </DefaultTooltip>
+                </h6>
+                <SimpleTable 
+                    headers={[
+                        <th></th>,
```
onursumer: No text for the header, is this intentional?

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,85 @@
+import * as React from "react";
+import DefaultTooltip from "shared/components/defaultTooltip/DefaultTooltip";
+import {placeArrowBottomLeft} from "shared/components/defaultTooltip/DefaultTooltip";
+import SimpleTable from "shared/components/simpleTable/SimpleTable";
+import FontAwesome from "react-fontawesome";
+
+export type GDDOutput = {
+    Pred1: string;
```
onursumer: Why not `pred1` (all lowercase variable name)? Would it be better two have 2 arrays here instead of 6 variables, something like:

```
export type GDDOutput = {
    preds: string[];
    confs: string[]
}
```

Also, maybe use the full names `prediction` and `confidence` instead of truncated versions `pred` and `conf` for clarity?

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,85 @@
...
+}
+
+export default class GenomeDrivenDiagnosis extends React.Component<IGenomeDrivenDiagnosisProps, {}> {
+    public render() {
+        return (
+            <DefaultTooltip
+                placement='bottomLeft'
+                trigger={['hover', 'focus']}
+                overlay={this.tooltipContent()}
+                arrowContent={<div className="rc-tooltip-arrow-inner" />}
```
onursumer: I think we don't need the `arrowContent` anymore, it should already be defined in the `DefaultTooltip`.

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
inodb: @onursumer this is the data format as returned from GDD.  We could transform it for esthetic purposes, though not sure if that adds confusion

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
inodb: yes first column is an index

inodb:
Thanks for reviewing @onursumer !

src/pages/patientView/patientHeader/mock/gdd/P-0001055-T01-IM3
```diff
@@ -0,0 +1,38 @@
+{
```
inodb: example file of output of one patient. Eventually the data will just be stored in clinical attributes, so this is a temporary thing.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -200,6 +201,18 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
                 );
+                // check whether the sample has a Genomic Driven Diagnosis
+                // prediction
+                const hasGDDPred:boolean = (
+                    AppConfig.genomeDrivenDiagnosisUrl &&
+                    AppConfig.genomeDrivenDiagnosisUrl.length > 0 &&
+                    patientViewPageStore.genomeDrivenDiagnosis.isComplete &&
+                    patientViewPageStore.genomeDrivenDiagnosis.result &&
+                    patientViewPageStore.genomeDrivenDiagnosis.result[sample.id] &&
+                    _.every(["Pred1","Pred2","Pred3","Conf1","Conf2","Conf3"]
```
alisman: this is a bit of a strange use of _.every i think, thought it works.   i think you don't need the map.  just pass as second param to every the callback you are passing to map

alisman: also, this should probably go into a computed getter on this component.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -254,6 +254,32 @@ export class PatientViewPageStore {
...
+        ],
+        invoke: async() => {
+            let genomeDrivenDiagnosisUrl = getGenomeDrivenDiagnosisUrl();
+
+            if (genomeDrivenDiagnosisUrl) {
+                let resp: any = await Promise.all(this.samples.result.map(s => request.get(`${genomeDrivenDiagnosisUrl}/${s.sampleId}`)));
+
+                const parsedResp: any = _.keyBy(resp.map((r:any) => JSON.parse(r.text)), 'Tumor_Sample_Barcode');
+                // remove undefined keys
+                delete parsedResp["undefined"];
```
alisman: would it ever have undefined?

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,197 @@
...
+                onPopupAlign={placeArrowBottomLeft}
+            >
+                <span style={{paddingLeft:2}}>
+                    {this.predictionContent()}
+                </span>
+            </DefaultTooltip>
+        );
+    }
+
+    @computed private get isInAllowedCancerTypes() {
```
alisman: i would be careful with use of computed here.  The thing is, if the items reference inside are not observable, then computed will not reflect changes to them.   Unless you're using obviously observable properties, or there is a really strong case for optimization, I would avoid using computed because it's easy to mess up.

src/pages/patientView/PatientViewPage.tsx
inodb: yes good points. Im just gonna add this to an issue that the sampleHeader should be it's own component. We prolly change this a bit more once it goes through beta testing. Also the data might come from clinical attributes instead once we get it fed from cvr. So not super keen on restructuring this rn: https://github.com/cBioPortal/cbioportal-frontend/issues/988

src/pages/patientView/patientHeader/GenomeDrivenDiagnosis.tsx
```diff
@@ -0,0 +1,197 @@
...
+                  // Show warning when diagnosis not part of 22 types
+                  (!this.isInAllowedCancerTypes && (
+                    <p className="small text-danger" style={{textAlign:"center"}}>
+                        The diagnosed
+                        cancer type "{this.props.prediction.Diagnosed_Cancer_Type}" is not
+                        part of the 22 tumor types GDD can predict.
+                    </p>
+                  ))
+                }
+                <SimpleTable
```
alisman: I would not use SimpleTable unless you are going to use a feature of it.  Just use a plain html table and give it a class="table"    That we get consistent style but not the probably pointless dependency

jjgao:
@inodb I am closing this for now. Let's create a new one when/if we work on this again.

PR #727, opened 2017-11-06
Add addthis.

Tamba32:
Connected with cbioportal/cbioportal PR: https://github.com/cBioPortal/cbioportal/pull/3318

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -3,7 +3,7 @@ import * as ReactDOM from 'react-dom';
 import * as _ from 'lodash';
 import $ from 'jquery';
 import {observer, inject, Observer} from "mobx-react";
-import {reaction, computed} from "mobx";
+import {reaction, computed, observable} from "mobx";
 import validateParameters from 'shared/lib/validateParameters';
 import ValidationAlert from "shared/components/ValidationAlert";
 import AjaxErrorModal from "shared/components/AjaxErrorModal";
@@ -14,6 +14,8 @@ import Mutations from "./mutation/Mutations";
 import MutualExclusivityTab from "./mutualExclusivity/MutualExclusivityTab";
 import SurvivalTab from "./survival/SurvivalTab";
 import Chart from 'chart.js';
+import AppConfig from 'appConfig';
+import AddThisBookmark from 'shared/components/addThis/addThisBookmark';
 import {CancerStudy} from "../../shared/api/generated/CBioPortalAPI";
 import getOverlappingStudies from "../../shared/lib/getOverlappingStudies";
 import OverlappingStudiesWarning from "../../shared/components/overlappingStudiesWarning/OverlappingStudiesWarning";
@@ -101,6 +103,8 @@ type OncoprintTabInitProps = {
 @observer
 export default class ResultsViewPage extends React.Component<IResultsViewPageProps, {}> {
 
+    private showTwitter = AppConfig.showTwitter === true;
+
     constructor(props: IResultsViewPageProps) {
         super();
         this.exposeComponentRenderersToParentScript();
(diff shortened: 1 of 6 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -91,7 +94,56 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+    }
+
+    componentDidUpdate() {
+        this.addBookmarkButtons();
+    }
+
+    public addBookmarkButtons() {
+        let shareURL = window.location.href;
+
+        if (shareURL.indexOf("session_id") < -1) return;
```
alisman: you can now use string.includes or /session_id/.test(str) which are boolean

Tamba32: Fixed.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -91,7 +94,56 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+
+        const showSamples = shareURL.indexOf("&show");
```
alisman: you should really use a url parser for this.  we have one installed already called :"url"

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -91,7 +94,56 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+        if (showSamples > -1) {
+            shareURL = shareURL.slice(0, showSamples);
+        }
+
+        if (shareURL.indexOf('localhost') > -1) {
+            this.addAddThis(shareURL);
+        } else {
+            $.ajax('http://' + location.host + "/api/url-shortener?url=" + shareURL)
```
alisman: for consistency, use superagent for ajax requests.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -91,7 +94,56 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+                    this.addAddThis(res.shortURL);
+                })
+                .fail(function() {
+                    this.addAddThis(shareURL);
+                });
+        }
+    }
 
+    public addAddThis(shareURL:string) {
+        const showTwitter = Boolean(AppConfig.showTwitter) === true;
```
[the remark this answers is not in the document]
Tamba32: For some reason,  `AppConfig.showTwitter` is a string. When I don't cast it, the comparison evaluates to `false`.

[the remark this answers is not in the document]
alisman: ok.  can you bring that up on refactoring chat?  We should try to avoid this situation.   Lets ask @inodb about it

alisman:
It would be best to keep the ResultsViewPage component uncluttered with feature specific stuff.  We should move that add this stuff into a component that would make it reusable.  This component should accept the url and twitter boolean as props.   As for the ajax to retrieve shortened url, lets put this in the resultsViewPageStore.  This way, all we have in the ResulstViewPage file is a component <AddThis>.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -91,7 +95,58 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+        } : {};
+        addthis(
+            {pubid: 'ra-59959e35901cdde7'},
+            {url: shareURL, ...twitterParam},
+            () => {
+                const addthisDiv = document.createElement("div");
+                Object.assign(addthisDiv.style, {display: 'flex', justifyContent: 'flex-end'});
+                if (showTwitter) addthisDiv.classList.add("addthis_inline_share_toolbox");
+                else addthisDiv.classList.add("addthis_inline_share_toolbox_ubww");
+                document.getElementsByClassName("main_smry")[0].appendChild(addthisDiv);
```
alisman: We want to avoid this kind of DOM reference.  Instead, we should use a component which can manage its own dom and call addthis against it.

PR #735, opened 2017-11-09
Change cross cancer end to end test for iviz

inodb:
Before the test looked for the old cross cancer plot. Now we only check whether
the gene that was searched for is displayed on the tumor type page.


inodb:
Only unit tests are failing, but unrelated to this PR

PR #749, opened 2017-11-17
Disable showing virtual studies in home page

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3364


zhx828:
@kalletlak I'm sure this will solve the issue, but I don't understand why it will be converted to a string null...

kalletlak:
its because of https://github.com/cBioPortal/cbioportal-frontend/pull/749/files#diff-2edbc62f162a228681eff5291dc29be8R201

however in my next pr I am going to cleanup most of the code since we are not saving them to localStorage

zhx828:
This will be cleaned up in next release.

PR #750, opened 2017-11-17
Survival Tab with Victory

ersinciftci:


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -0,0 +1,345 @@
...
+          }];
+
+        return (
+            
+            <div className={styles.SurvivalChart}>
+                <VictoryChart containerComponent={<VictoryContainer responsive={false} containerRef={(ref:any) => this.svgContainer = ref}/>}
+                height={650} width={1150} padding={{top: 50, bottom: 50, left: 60, right: 300}}>
+                    <VictoryLabel x={50} y={15} text={this.props.title} style={{fontSize: 24}}/>
+                    <VictoryLine interpolation="stepAfter" data={getLineData(this.sortedAlteredPatientSurvivals, this.alteredEstimates)}
+                    style={{data: {stroke: "red", strokeWidth: 1}}}/>
```
onursumer: According to our current lint settings, this is the preferred indentation for multi-line JSX components (one property per line):

```JavaScript
<VictoryLine 
    interpolation="stepAfter" 
    data={getLineData(this.sortedAlteredPatientSurvivals, this.alteredEstimates)}
    style={{data: {stroke: "red", strokeWidth: 1}}}
/>
```

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -0,0 +1,345 @@
...
+              onMouseOut: () => {
+                return [
+                  {
+                    target: "data",
+                    mutation: () => ({ active: false })
+                  },
+                  {
+                    target: "labels",
+                    mutation: async () => {
+                        await sleep(100);
```
alisman: is this like a wrapper for setTimeout?  I just want to make sure it's not some kind of looping that is actually sychronous

ersinciftci: @alisman it's a function I created in src/shared/lib/TimeUtils.ts, a new file present in this PR.

src/pages/resultsView/survival/SurvivalChart.tsx
ersinciftci: I tried to do this but it gets too long. Maybe we should change the rule :)

src/pages/resultsView/survival/SurvivalChart.tsx
adamabeshouse: even though it makes the function execution "synchronous", await is asynchronous https://hackernoon.com/6-reasons-why-javascripts-async-await-blows-promises-away-tutorial-c7ec10518dd9

alisman: i know that await is async.  i think i got hung up on "sleep" because in my experience sleep means the processor locks.   i think maybe "wait" might be better.  but it's not a big deal.  i just wanted to make sure it wasn't some thing like for (i < 1trillion) because people have written such things!

src/pages/resultsView/survival/SurvivalTab.tsx
```diff
@@ -0,0 +1,100 @@
...
+}
+
+@observer
+export default class SurvivalTab extends React.Component<ISurvivalTabProps, {}> {
+
+    private overallSurvivalTitleText = 'Overall Survival Kaplan-Meier Estimate';
+    private diseaseFreeSurvivalTitleText = 'Disease Free Survival Kaplan-Meier Estimate';
+
+    componentWillMount() {
+        Chart.pluginService.register({
```
alisman: we can take this out no?

src/pages/resultsView/survival/styles.module.scss
```diff
@@ -0,0 +1,42 @@
+.SurvivalTable {
+  margin: 30px auto;
+  width: 65%;
+}
+
+table {
```
alisman: i don't think this is nested so this will apply to all <table>?

src/pages/resultsView/survival/SurvivalTab.tsx
```diff
@@ -0,0 +1,100 @@
...
+                           yAxisLabel="Disease Free"
+                           totalCasesHeader="#total cases"
+                           statusCasesHeader="#cases relapsed"
+                           medianMonthsHeader="median months disease free"
+                           yLabelTooltip="Disease free estimate"
+                           xLabelWithEventTooltip="Time of relapse"
+                           xLabelWithoutEventTooltip="Time of last observation"
+                           fileName="Disease_Free_Survival"/>);
+        } else {
+            content.push(<div className={styles.NotAvailable}>{this.diseaseFreeSurvivalTitleText} not available</div>);
```
alisman: seems like this would read non available when in pending state

src/pages/resultsView/survival/SurvivalTab.tsx
ersinciftci: Yes, will do later, we can merge for now.

src/pages/resultsView/survival/styles.module.scss
ersinciftci: Yes, will make it more specific later, we can merge for now.

src/pages/resultsView/survival/SurvivalTab.tsx
ersinciftci: Not sure what you mean, but we can discuss later, can merge for now.

ersinciftci: @alisman There is
```
if (this.props.store.overallAlteredPatientSurvivals.isPending ||
            this.props.store.overallUnalteredPatientSurvivals.isPending ||
            this.props.store.diseaseFreeAlteredPatientSurvivals.isPending ||
            this.props.store.diseaseFreeUnalteredPatientSurvivals.isPending) {
            return <Loader isLoading={true}/>;
        }
```
on the top.

PR #753, opened 2017-11-17
Update genome nexus and use new endpoints lollipop

inodb:
We now return the pfam ids and ranges as part of ensembl canonical transcript
response. Update lollipop store and plot accordingly.

### DIFF
--- a/src/shared/api/generated/CBioPortalAPI.ts
+++ b/src/shared/api/generated/CBioPortalAPI.ts
@@ -186,6 +186,8 @@ export type SampleList = {
 
         'sampleCount': number
 
+        'sampleIds': Array < string >
+
         'sampleListId': string
 
         'studyId': string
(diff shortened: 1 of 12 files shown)


src/shared/api/generated/CBioPortalAPI.ts
```diff
@@ -186,6 +186,8 @@ export type SampleList = {
```
inodb: Ther'es also a new endpoint of cBioPortal, but not related to this PR

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -208,26 +208,45 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
...
                 };
             });
         }
     }
 
     @computed private get domainColorMap(): {[pfamAccession:string]: string}
     {
-        if (!this.props.store.pfamDomainData.isPending && this.props.store.pfamDomainData.result) {
-            return generatePfamDomainColorMap(this.props.store.pfamDomainData.result);
+        if (!this.props.store.canonicalTranscript.isPending &&
```
inodb: now use canonicalTranscript to make the domain color map instead of using the pfam domain data

src/shared/lib/PfamUtils.spec.ts
```diff
@@ -1,71 +1,35 @@
 import * as _ from 'lodash';
 import { assert } from 'chai';
-import {PfamDomain} from "shared/api/generated/GenomeNexusAPI";
+import {PfamDomainRange} from "shared/api/generated/GenomeNexusAPI";
 import {generatePfamDomainColorMap} from "./PfamUtils";
 
-let domains: PfamDomain[];
+let domains: PfamDomainRange[];
```
inodb: we're making the colors based on the pfam domain ranges

PR #766, opened 2017-11-28
add unknown/unauthorized error message

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3446

### DIFF
--- a/src/shared/components/query/QueryContainer.tsx
+++ b/src/shared/components/query/QueryContainer.tsx
@@ -15,6 +15,7 @@ import {providesStoreContext} from "../../lib/ContextUtils";
 import AppConfig from "appConfig";
 import CaseSetSelector from "./CaseSetSelector";
 import OverlappingStudiesWarning from "../overlappingStudiesWarning/OverlappingStudiesWarning";
+import UnknownStudiesWarning from "../unknownStudies/UnknownStudiesWarning"
 
 const styles = styles_any as {
 	QueryContainer: string,
@@ -64,6 +65,9 @@ export default class QueryContainer extends React.Component<QueryContainerProps,
 
                 {
                     <OverlappingStudiesWarning studies={this.store.selectedStudies}/>
+                }
+                {
+                    <UnknownStudiesWarning ids={this.store.unknownStudyIds} />
                 }
 
 				<CancerStudySelector/>
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -276,6 +276,7 @@ export class QueryStore
 		ids = ids.filter(id=>!!selectableStudies[id]);
 		return this.forDownloadTab ? ids.slice(-1) : ids;
 	}
+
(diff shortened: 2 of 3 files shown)


src/shared/components/query/QueryContainer.tsx
```diff
@@ -64,6 +65,10 @@ export default class QueryContainer extends React.Component<QueryContainerProps,
 
                 {
                     <OverlappingStudiesWarning studies={this.store.selectedStudies}/>
+				}
+				
+				{
+                    <UnknownStudiesWarning ids={this.store.unknownStudyIds} />
```
inodb: @kalletlak seems to be a combo of tabs and spaces, let's stick to spaces

src/shared/components/query/QueryContainer.tsx
```diff
@@ -65,6 +66,9 @@ export default class QueryContainer extends React.Component<QueryContainerProps,
                 {
                     <OverlappingStudiesWarning studies={this.store.selectedStudies}/>
                 }
+                {
+                    <UnknownStudiesWarning ids={this.store.unknownStudyIds} />
```
alisman: make sure that if both of these errors are showing, layout looks ok. please put screenshot in PR

src/shared/components/query/QueryStore.ts
```diff
@@ -730,6 +731,14 @@ export class QueryStore
 		return this.selectedStudyIds.map(id => this.treeData.map_studyId_cancerStudy.get(id) as CancerStudy).filter(_.identity);
 	}
 
+	@computed get unknownStudyIds()
+	{
+		let ids:string[] = this._selectedStudyIds.keys();
+		const selectableStudies = this.selectableStudiesSet;
+		ids = ids.filter(id=>!selectableStudies[id]);
```
alisman: just a quibble but you should use "in" operator. more semantic

src/shared/components/query/QueryContainer.tsx
inodb: @alisman @kalletlak 
screen shot 2017-12-07 at 1 58 08 pm

src/shared/components/unknownStudies/UnknownStudiesWarning.tsx
```diff
@@ -0,0 +1,15 @@
+import * as React from 'react';
+
```
alisman: you should tag this with @observer decorator so that it will react to changes in observable properties (ids).  for example, the session might change, in which case the warning should go away.

inodb:
@kalletlak ill fix @alisman's comments

kalletlak:
Thanks @inodb

inodb:
FYI we can't show the full study name, because access to those study ids are forbidden

src/shared/components/query/QueryContainer.tsx
inodb: FYI we can't show the full study name, because access to those study ids are forbidden

PR #768, opened 2017-11-29
Do not request data for invalid germline sample lists

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3471.

- Fixed by using `getAllSampleListsInStudyUsingGET` and filtering sample ids by `germline`.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1,7 +1,7 @@
 import {
     DiscreteCopyNumberFilter, DiscreteCopyNumberData, ClinicalData, ClinicalDataMultiStudyFilter, Sample,
     SampleIdentifier, MolecularProfile, Mutation, GeneMolecularData, MolecularDataFilter, Gene,
-    ClinicalDataSingleStudyFilter, CancerStudy
+    ClinicalDataSingleStudyFilter, CancerStudy, SampleList
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
 import {computed, observable, action} from "mobx";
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -608,17 +608,23 @@ export class ResultsViewPageStore {
...
-                return sampleIds.map(sampleId => ({sampleId, studyId}));
-            }));
+            if (this.germlineSampleLists.result) {
+                const ids: string[][] = await Promise.all(this.germlineSampleLists.result.map(sampleList => {
+                    return client.getAllSampleIdsInSampleListUsingGET({
+                        sampleListId: sampleList.sampleListId
+                    });
+                }));
+
+                return _.flatten(ids.map((sampleIds: string[], index: number) => {
```
adamabeshouse: this doesnt seem multiple-study-safe - need to perform conversion from sampleId to uniqueSampleKey, or just return Sample objects

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -675,9 +681,35 @@ export class ResultsViewPageStore {
...
     }, {});
 
-    private getGermlineSampleListId(studyId:string):string {
-        return `${studyId}_germline`;
-    }
+    readonly sampleListIdsInStudies = remoteData({
+        await: () => [this.studies],
+        invoke: async () => {
+            const studies: string[] = this.studyIds;
+            return await Promise.all(studies.map(studyId => {
```
adamabeshouse: return _.flatten(await... ) would be more intuitive I think, to return a flat list

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: Does this mean that the current code is already not multiple-study-safe?

```JavaScript
return _.flatten(ids.map((sampleIds: string[], index: number) => {
    const studyId = studies[index];
    return sampleIds.map(sampleId => ({sampleId, studyId}));
}));
```

adamabeshouse: nevermind, I didn't read carefully enough. looks good

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: I did it this way to keep the current logic in `germlineConsentedSamples` intact. Seems like we are already doing the flattening there.

onursumer: Hmm, actually it looks okay to be flattened here...

jjgao:
I think we may want to have a better solution. Let's discuss.

PR #771, opened 2017-11-30
Download support for the mutation table annotation, functional impact, and mutation count columns

onursumer:
# What? Why?
- Fixed https://github.com/cBioPortal/cbioportal/issues/3427.
- Also implemented download function for `# of Mut`

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
+++ b/src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
@@ -1,15 +1,20 @@
 import * as React from 'react';
-import {Button, ButtonGroup} from 'react-bootstrap';
+import {Modal, Button, ButtonGroup} from 'react-bootstrap';
+import {ThreeBounce} from 'better-react-spinkit';
 import DefaultTooltip from 'shared/components/defaultTooltip/DefaultTooltip';
 import {If} from 'react-if';
 import fileDownload from 'react-file-download';
+import {observable} from "mobx";
+import {observer} from "mobx-react";
 const Clipboard = require('clipboard');
+
+import copyDownloadStyles from "./copyDownloadControls.module.scss";
 
 export interface ICopyDownloadControlsProps {
     className?: string;
     showCopy?: boolean;
     showDownload?: boolean;
-    downloadData?: ()=>string;
+    downloadData?: () => Promise<string>;
     downloadFilename?: string;
 }
 
@@ -17,8 +22,11 @@ export interface ICopyDownloadControlsProps {
  * @author Selcuk Onur Sumer
  * @author Aaron Lisman
  */
+@observer
 export class CopyDownloadControls extends React.Component<ICopyDownloadControlsProps, {}>
 {
+    @observable downloadingData:boolean = false;
+
     private _copyButton: HTMLElement;
 
     public static defaultProps:ICopyDownloadControlsProps = {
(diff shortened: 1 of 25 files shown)


jjgao:
Does this handle Functional Consequence and "#Muts in sample"?

onursumer:
@jjgao As of the latest commit it handles both `Functional Impact` and `# of Mut` columns. It is also functional on the patient view page.

src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
```diff
@@ -31,77 +44,154 @@ export class CopyDownloadControls extends React.Component<ICopyDownloadControlsP
...
+                            placement="top"
+                            arrowContent={arrowContent}
+                        >
+                            <Button className="btn-sm" onClick={this.handleDownload}>
+                                <i className='fa fa-cloud-download'/>
+                            </Button>
+                        </DefaultTooltip>
+                    </If>
+                </ButtonGroup>
+                <Modal
```
onursumer: This is the modal component to show during data download

src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
```diff
@@ -31,77 +44,154 @@ export class CopyDownloadControls extends React.Component<ICopyDownloadControlsP
...
+                    onHide={() => undefined}
+                    bsSize="sm"
+                    className={`${copyDownloadStyles["centered-modal-dialog"]}`}
+                >
+                    <Modal.Body>
+                        <ThreeBounce style={{ display:'inline-block', marginRight:10 }} />
+                        <span>Downloading Table Data...</span>
+                    </Modal.Body>
+                </Modal>
+                <Modal
```
onursumer: We need to show another modal to be able to copy data to clipboard. It is not possible to trigger an async copy to clipboard function programmatically due to browser security restrictions.

src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
```diff
@@ -31,77 +44,154 @@ export class CopyDownloadControls extends React.Component<ICopyDownloadControlsP
...
+                    className={`${copyDownloadStyles["centered-modal-dialog"]}`}
+                >
+                    <Modal.Header>
+                        Download Complete!
+                    </Modal.Header>
+                    <Modal.Body>
+                        Please click on Copy to copy downloaded data to clipboard.
+                    </Modal.Body>
+                    <Modal.Footer>
+                        <span
```
onursumer: The button needs to be wrapped with another HTML element, in order for the clipboard to work in a Modal: https://stackoverflow.com/questions/38398070/bootstrap-modal-does-not-work-with-clipboard-js-on-firefox

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -578,9 +582,23 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
...
 
-    public getDownloadData(): string
+    public getDownloadData(): Promise<string>
     {
-        return serializeData(this.store.downloadData);
+        return new Promise<string>((resolve, reject) => {
+            if (this.store.downloadStore) {
+                this.store.downloadStore.loadAllData().then(isLoaded => {
+                    if (isLoaded) {
+                        resolve(serializeData(this.store.downloadData));
```
adamabeshouse: I think the use of this is confusing, where the downloadstore only reports on the data status and requests it but does not return it. I think the downloadstore should return the data as well

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -578,9 +582,23 @@ export default class LazyMobXTable<T> extends React.Component<LazyMobXTableProps
         showCountHeader: false
     };
 
-    public getDownloadData(): string
+    public getDownloadData(): Promise<string>
     {
-        return serializeData(this.store.downloadData);
+        return new Promise<string>((resolve, reject) => {
+            if (this.store.downloadStore) {
```
alisman: We're going to get rid of nested "download" store.  Change some names to make this more clear.  That way it will make more sense why data doesn't come from store as promise return (b/c it's not in store!).  We have to leverage existing download functions to get it.

adamabeshouse: can you explain more what's going on in this PR?

PR #775, opened 2017-12-04
Fetch mutation data with single call to cross study endpoint

alisman:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -535,8 +525,53 @@ export class ResultsViewPageStore {
         return fetchMyCancerGenomeData();
     }
 
+    @computed mutationData: {
+
+    }
+
+    readonly mutationDataNew = remoteData<{ [geneId:string]:Mutation[] }>({
```
adamabeshouse: bad name

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -535,8 +525,53 @@ export class ResultsViewPageStore {
         return fetchMyCancerGenomeData();
     }
 
+    @computed mutationData: {
```
adamabeshouse: whats this?

src/pages/resultsView/ResultsViewPageStore.ts
alisman: true.

PR #778, opened 2017-12-05
Updates to query interface for multi-study query

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3513

The backend change: https://github.com/cBioPortal/cbioportal/pull/3799

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -63,8 +63,7 @@ function initStore(queryStore: QueryStore) {
     // ultimate we will phase this out and this information will be stored in router etc.
     //const qSession:any = (window as any).QuerySession;
     var samplesSpecification:any = [];
-    if (serverVars.caseSetProperties.case_set_id === "all") {
-        // "all" means all cases in the queried stud(y/ies) - not an actual case set that could be queried
+    if(_.includes(['all', '0', '1', '2'],serverVars.caseSetProperties.case_set_id)){
         var studyToSampleMap = serverVars.studySampleObj;
         var studies = Object.keys(studyToSampleMap);
         for (var i=0; i<studies.length; i++) {
--- a/src/shared/components/query/DataTypePrioritySelector.tsx
+++ b/src/shared/components/query/DataTypePrioritySelector.tsx
@@ -25,15 +25,15 @@ export default class DataTypePrioritySelector extends QueryStoreComponent<{}, {}
 			return null;
 
 		let flexRowContents:JSX.Element[] = [];
-		flexRowContents.push(<LoadingIndicator key="loading" isLoading={this.profileAvailability.isPending}/>);
-		if (this.profileAvailability.isError) {
+		flexRowContents.push(<LoadingIndicator key="loading" isLoading={this.store.profileAvailability.isPending}/>);
+		if (this.store.profileAvailability.isError) {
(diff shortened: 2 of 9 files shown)


src/shared/components/query/DataTypePrioritySelector.tsx
```diff
@@ -22,18 +22,32 @@ export default class DataTypePrioritySelector extends QueryStoreComponent<{}, {}
 
 		return (
 			<FlexRow padded className={styles.DataTypePrioritySelector}>
-				<SectionHeader className="sectionLabel">Select Data Type Priority:</SectionHeader>
+				<SectionHeader className="sectionLabel">Select Genomic Profiles:</SectionHeader>
```
inodb: @kalletlak let's do `Molecular Profiles` here

kalletlak: sure!

inodb:
@kalletlak could u resolve these conflicts?

kalletlak:
Some of the changes here are already done in https://github.com/cBioPortal/cbioportal-frontend/pull/816

kalletlak:
@inodb fixed conflicts, its ready to review. Thanks!

src/shared/components/query/QueryStore.ts
```diff
@@ -969,7 +969,15 @@ export class QueryStore
 	@computed get defaultSelectedSampleListId()
 	{
 		if (this.isVirtualCohortQuery) {
-			return ALL_CASES_LIST_ID;
+			let toReturn = ALL_CASES_LIST_ID;
```
alisman: this method really needs some tests.  this would a good opportunity to add them, at least for this logic here.

alisman: you can put the body of htis method in a helper function and then call it here.

src/shared/components/query/QueryStore.ts
```diff
@@ -1162,6 +1170,9 @@ export class QueryStore
 				return "Expression specified in the list of genes, but not selected in the Molecular Profile Checkboxes.";
 
 		}
+		if(this.isVirtualCohortQuery && !(this.dataTypePriority.mutation || this.dataTypePriority.cna)){
```
alisman: this also should have a test.  these are the new rules : )

kalletlak: done!

alisman:
I think this PR should be to rc, no?  Otherwise looks good.

kalletlak:
Since these are minor changes, I think its ok to go with master. But whatever you and @inodb says, I'll rebase and point it that branch.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -62,8 +62,8 @@ function initStore(queryStore: QueryStore) {
     // ultimate we will phase this out and this information will be stored in router etc.
     //const qSession:any = (window as any).QuerySession;
     var samplesSpecification:any = [];
-    if (serverVars.caseSetProperties.case_set_id === "all") {
-        // "all" means all cases in the queried stud(y/ies) - not an actual case set that could be queried
+    if (["-1", "0", "1", "2", "all"].indexOf(serverVars.caseSetProperties.case_set_id) > -1) {
```
alisman: you can now use [].includes instead of indexOf.   You don't need to make this change though.

PR #781, opened 2017-12-06
Switch to sorting by data in oncoprint if clinical or heatmap track...

adamabeshouse:
...sort direction is selected

This only happens if 'Sort a->Z' or 'Sort Z->a' is selected, if 'Dont sort on this track' is selected then nothing else happens.

### DIFF
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -23,7 +23,7 @@ import _ from "lodash";
 import onMobxPromise from "shared/lib/onMobxPromise";
 import AppConfig from "appConfig";
 import LoadingIndicator from "shared/components/loadingIndicator/LoadingIndicator";
-import OncoprintJS from "oncoprintjs";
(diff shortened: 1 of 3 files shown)


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -598,6 +599,12 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         this.selectedClinicalAttributeIds.delete(this.clinicalTrackKeyToAttributeId(clinicalTrackKey));
     }
 
+    private onTrackSortDirectionChange(trackId:TrackId, dir:number) {
+        if (dir === 1 || dir === -1) {
```
alisman: is this right?  where is direction state stored if it's not passed to sortByData function and sortByData is called in either case?   maybe a comment would be good.

adamabeshouse: comment added. This direction state is stored within the oncoprintjs object, this callback is called whenever the user selects a sort direction for a track. See https://github.com/cBioPortal/cbioportal/issues/3374 for more info

PR #782, opened 2017-12-07
Fix bug in how we compute sample clinical data for patient attributes

adamabeshouse:


src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -389,9 +410,10 @@ export function makeClinicalTrackData(
         = _.groupBy(data,
         isSampleList(cases) ?
             datum=>datum.uniqueSampleKey :
-            datum=>datum.uniquePatientKey);// TS error will be gone when ersin updates API
+            datum=>datum.uniquePatientKey);
 
     // Create oncoprint data
+    const getDataForCase = makeGetDataForCase(attribute, isSampleList(cases) ? "sample" : "patient", data);
```
alisman: i'd vote for using parens around both the condition and the entire ternary here to make this more readable.

adamabeshouse: around the condition seems excessive... around the ternary I agree. but if you feel strongly I dont care that much

adamabeshouse: ah whatever it does look a little nicer that way

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -379,6 +379,27 @@ function fillClinicalTrackDatum(
     return trackDatum;
 }
 
+function makeGetDataForCase(
```
alisman: this is sort of crying out for a test.  i have a tool that should make it quite easy.  i'll merge and we can revisit.

adamabeshouse: true, show me the tool when you get a chance

PR #792, opened 2017-12-08
Use symbolic links for getting pathology reports

inodb:
Fix cbioportal/cbioportal#3367

Instead of using the Github search API, which is rate limited, directly access
the symbolic links of type PATIENT_ID.number (patients can have multiple
reports). The symbolic links contain the name of the actual PDF file, which can
then be used to display the report.

3 reports:
http://www.cbioportal.org/case.do#/patient?caseId=TCGA-TQ-A7RK&navCaseIds=TCGA-TQ-A7RK&studyId=lgg_tcga&tab=pathologyReportTab

1 report:
http://www.cbioportal.org/case.do#/patient?caseId=TCGA-BK-A0CC&studyId=ucec_tcga_pub&tab=pathologyReportTab

no report:
http://www.cbioportal.org/case.do#/patient?studyId=lgg_ucsf_2014&caseId=P04

### DIFF
--- a/src/pages/patientView/pathologyReport/PathologyReport.tsx
+++ b/src/pages/patientView/pathologyReport/PathologyReport.tsx
@@ -3,6 +3,7 @@ import {PathologyReportPDF} from "../clinicalInformation/PatientViewPageStore";
 import { If, Then, Else } from 'react-if';
 import * as _ from 'lodash';
 import IFrameLoader from "../../../shared/components/iframeLoader/IFrameLoader";
+import {observer} from "mobx-react";
 
 export type IPathologyReportProps = {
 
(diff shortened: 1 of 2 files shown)


src/pages/patientView/pathologyReport/PathologyReport.tsx
```diff
@@ -21,15 +23,16 @@ export default class PathologyReport extends React.Component<IPathologyReportPro
 
         super();
 
-        this.state = { pdfUrl: this.buildPDFUrl(props.pdfs[0].name) }
+        debugger;
```
ersinciftci: debugger line

inodb:
good catch @ersinciftci , removed it

PR #793, opened 2017-12-08
Make getPutativeDriverInfo a mobxpromise

adamabeshouse:
…so we can await it before showing oncoprint


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1429,39 +1430,59 @@ export class ResultsViewPageStore {
...
-        return {
-            oncoKb,
-            hotspots,
-            cbioportalCount,
-            cosmicCount
-        };
-    }
+    readonly getPutativeDriverInfo = remoteData({
+        await:()=>{
+            const toAwait = [];
```
adamabeshouse: dynamic `await`... this is necessary because otherwise we're querying for, and waiting for, data before we need it

PR #800, opened 2017-12-12
Bring over minimap icon from legacy oncoprint

adamabeshouse:
image

### DIFF
--- a/src/shared/components/oncoprint/controls/OncoprintControls.tsx
+++ b/src/shared/components/oncoprint/controls/OncoprintControls.tsx
@@ -838,8 +838,9 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
                 <Button
                     active={this.showMinimap}
                     onClick={this.toggleShowMinimap}
+                    className="oncoprint__controls__minimap_button"
                 >
-                    <i className="fa fa-expand"></i>&thinsp;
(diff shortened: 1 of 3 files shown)


src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -838,8 +838,9 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
                 <Button
                     active={this.showMinimap}
                     onClick={this.toggleShowMinimap}
+                    className="oncoprint__controls__minimap_button"
                 >
-                    <i className="fa fa-expand"></i>&thinsp;
+                    <img src="images/fitalteredcases.svg" alt="icon" width="20" height="20"/>&thinsp;
```
alisman: lets copy over this image and then use request() so that webpack will handle it

alisman: webpack will also probably automatically inline the svg into css file.

adamabeshouse: when you get back let's talk about how to do this

alisman: src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
line 49

PR #803, opened 2017-12-12
Survival tab text modifications

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/3530


alisman:
@ersinciftci i think we should bring styles of table in line with other tables on site.  also add a little more formatting near the bottom.  i can do this.

PR #806, opened 2017-12-13
Remove discreteCNAData from ResultsViewPageStore

adamabeshouse:


src/shared/lib/StoreUtils.ts
```diff
@@ -465,6 +465,30 @@ export async function fetchCnaOncoKbData(uniqueSampleKeyToTumorType:{[uniqueSamp
...
+                                          molecularProfileIdToMolecularProfile:{[molecularProfileId:string]:MolecularProfile},
+                                         client: OncoKbAPI = oncokbClient)
+{
+    if (!geneMolecularData.result || geneMolecularData.result.length === 0) {
+        return ONCOKB_DEFAULT;
+    }
+    else
+    {
+        const alterationsToQuery = _.filter(geneMolecularData.result, d=>{
+            return molecularProfileIdToMolecularProfile[d.molecularProfileId].molecularAlterationType === "COPY_NUMBER_ALTERATION" &&
```
alisman: There's a constants collection for this AlterationTypeConstants

src/shared/lib/StoreUtils.ts
```diff
@@ -465,6 +465,30 @@ export async function fetchCnaOncoKbData(uniqueSampleKeyToTumorType:{[uniqueSamp
...
+                                         geneMolecularData:MobxPromise<GeneMolecularData[]>,
+                                          molecularProfileIdToMolecularProfile:{[molecularProfileId:string]:MolecularProfile},
+                                         client: OncoKbAPI = oncokbClient)
+{
+    if (!geneMolecularData.result || geneMolecularData.result.length === 0) {
+        return ONCOKB_DEFAULT;
+    }
+    else
+    {
+        const alterationsToQuery = _.filter(geneMolecularData.result, d=>{
```
alisman: i think d is MolecularData. we shoudl just name is molecularData or some such.  so much easier to read

src/shared/lib/StoreUtils.ts
```diff
@@ -465,6 +465,30 @@ export async function fetchCnaOncoKbData(uniqueSampleKeyToTumorType:{[uniqueSamp
...
+            return molecularProfileIdToMolecularProfile[d.molecularProfileId].molecularAlterationType === "COPY_NUMBER_ALTERATION" &&
+                !!annotatedGenes[d.entrezGeneId];
+        });
+        const queryVariants = _.uniqBy(_.map(alterationsToQuery, (datum: GeneMolecularData) => {
+            return generateQueryVariant(datum.entrezGeneId,
+                cancerTypeForOncoKb(datum.uniqueSampleKey, uniqueSampleKeyToTumorType),
+                getAlterationString(parseInt(datum.value, 10)));
+        }), "id");
```
alisman: as nice as it is to use the string shortcut, we should use (molduclarData)=>molecularData.id  
to get the typescript safety.   We should just establish that habit.

adamabeshouse: I just copied this over but you're right

src/shared/lib/StoreUtils.ts
adamabeshouse: sure

src/shared/lib/StoreUtils.ts
adamabeshouse: right

PR #814, opened 2017-12-14
Update code for new samples endpoint params

adamabeshouse:
(1) Update calls to use sampleFilter

(2) Query by sample list id for samples

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -3,7 +3,7 @@ import {
     SampleIdentifier, MolecularProfile, Mutation, GeneMolecularData, MolecularDataFilter, Gene,
     ClinicalDataSingleStudyFilter, CancerStudy, PatientIdentifier, Patient, GenePanelData, GenePanelDataFilter,
     SampleList, MutationCountByPosition, MutationMultipleStudyFilter, SampleMolecularIdentifier,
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1156,19 +1156,37 @@ export class ResultsViewPageStore {
...
+                    sampleIdentifiers = sampleIdentifiers.concat(dataQueryFilter.sampleIds.map(sampleId => ({
+                        sampleId,
+                        studyId
+                    })));
+                } else if (dataQueryFilter.sampleListId) {
+                    sampleListIds.push(dataQueryFilter.sampleListId);
+                }
             });
+            let promises:Promise<Sample[]>[] = [];
+            if (sampleIdentifiers.length) {
```
alisman: i see, so there is possiblity we have BOTH sample lists and sampleIdentifiers?

adamabeshouse: I think the way it currently is, there is no such possibility - but it's easy to imagine this would happen down the line if our cohort creation UI becomes more sophisticated.

alisman: but is it possible now we would somehow have overlapping?  i.e. both list and samples contained in that list?

adamabeshouse: no thats not possible

PR #816, opened 2017-12-14
populate data priority into QueryStore

adamabeshouse:


src/shared/components/query/QueryStore.ts
```diff
@@ -1241,6 +1241,10 @@ export class QueryStore
 		// Populate OQL
 		if ((window as any).serverVars) {
 			this.geneQuery = normalizeQuery((window as any).serverVars.theQuery);
+			const dataPriority = (window as any).serverVars.dataPriority;
```
alisman: can we have enum that helps explain what these numbers are?

alisman: lets follow up with it

PR #819, opened 2017-12-15
Check for nonempty oql query w virtual study query

adamabeshouse:


src/shared/components/query/QueryStore.ts
```diff
@@ -1190,7 +1190,11 @@ export class QueryStore
 	                return "Please enter one or more gene symbols.";
 	            }
             }
-        }
+        } else {
+			if (!this.oql.query.length) {
+				return "Please enter one or more gene symbols.";
```
adamabeshouse: this section is pretty dirty.. we should move these messages into constants and clean/DRY up this logic

alisman: agreed

src/shared/components/query/QueryStore.ts
```diff
@@ -1190,7 +1190,11 @@ export class QueryStore
 	                return "Please enter one or more gene symbols.";
 	            }
             }
-        }
+        } else {
```
alisman: i wish we could understand what changed to cause this logic to fail.

adamabeshouse: yes, its that the check for empty oql is gated by the check on line 1149 for nonempty list of selected profile ids, and I guess for multiple study queries, you dont select profile ids explicitly so that list is empty

PR #829, opened 2017-12-18
use gene panel multiple study endpoint

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -483,43 +483,41 @@ export class ResultsViewPageStore {
...
-                        genePanelDataFilter:{
-                            entrezGeneIds: this.genes.result!.map(x=>x.entrezGeneId),
-                            sampleIds: dataQueryFilter.sampleIds,
-                            sampleListId: dataQueryFilter.sampleListId
-                        } as GenePanelDataFilter
+            const studyToMutationMolecularProfile = this.studyToMutationMolecularProfile.result!;
+            const sampleMolecularIdentifiers:SampleMolecularIdentifier[] = [];
+            this.samples.result!.forEach(sample=>{
+                const profile = studyToMutationMolecularProfile[sample.studyId];
+                if (profile) {
```
alisman: provide for case where there's no profile

adamabeshouse: if theres no profile, we simply dont add that sample to the query

alisman: but if sampleMolecularIdentifiers doesn't accumulate anything (as would happen with CNA only query?) then we'll call service with empty array?

adamabeshouse: oh oops i forgot to push that update

PR #830, opened 2017-12-19
Implement volcano plot popup for GSVA data

oplantalech:
# What? Why?
We have incorporated a functionality to select gene sets using a volcano plot, for studies with GSVA data. This functionality is in a popup that opens when clicking the button "Select Gene Sets From Volcano Plot" above the gene sets query text box (this text box appears only when the "GSVA scores" genomic profile is selected.

The popup allows to select a percentile to calculate the GSVA scores, and it is divided in two main parts: the volcano plot and the table. You can select points in the volcano plot using the mouse (the points will change from blue to red), and automatically* the table will be filtered with the correspondent gene sets. Also, if you select any gene set from the table, the color of its point will change in the plot.

You can find a demo of the PR here: https://cbioportal-test.thehyve.net:8081/cbioportal/

***Note:** there is a performance issue in VictoryPlot when selecting points. I've raised an issue in the correspondent repository, you are welcome to contribute: FormidableLabs/victory#873.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -147,9 +147,9 @@
       "integrity": "sha1-dsvqSlVrol6xxtf6XnGsSOcvgc8="
     },
     "@types/rc-slider": {
-      "version": "8.2.0",
-      "resolved": "https://registry.npmjs.org/@types/rc-slider/-/rc-slider-8.2.0.tgz",
-      "integrity": "sha512-E01Y6Rdp6+jpnpETXe4IpApUlhU/a8ctXTJIJvvRT9z8VvN1r+7ZfwyJh9cHYlgl2JEO6xKATeuIDymY/rOzvw==",
+      "version": "8.2.1",
+      "resolved": "https://registry.npmjs.org/@types/rc-slider/-/rc-slider-8.2.1.tgz",
+      "integrity": "sha512-CNbk/Rlbhl4tNhjLLy5VCtGfMvj37+ukTVAB0MpQtd19Rs8e/GUgsx61jt1YHiXGC4ljtkUPkhaozQtJjPk5Gw==",
       "requires": {
         "@types/react": "15.6.2"
       }
--- a/src/shared/components/query/GenesetsSelector.tsx
+++ b/src/shared/components/query/GenesetsSelector.tsx
@@ -3,7 +3,7 @@ import * as styles_any from './styles/styles.module.scss';
 import {Modal} from 'react-bootstrap';
 import ReactSelect from 'react-select';
 import {observer} from "mobx-react";
-import {computed} from 'mobx';
+import {computed, ObservableMap} from 'mobx';
 import {FlexRow, FlexCol} from "../flexbox/FlexBox";
 import gene_lists from './gene_lists';
 import GenesetsValidator from "./GenesetsValidator";
@@ -11,12 +11,14 @@ import classNames from 'classnames';
 import {getOncoQueryDocUrl} from "../../api/urls";
 import {QueryStoreComponent} from "./QueryStore";
 import GenesetsHierarchySelector from "./GenesetsHierarchySelector";
+import GenesetsVolcanoSelector from "./GenesetsVolcanoSelector";
 import SectionHeader from "../sectionHeader/SectionHeader";
 import AppConfig from "appConfig";
 
 const styles = styles_any as {
     GeneSetSelector: string,
     GenesetsSelectorWindow: string,
+    GenesetsVolcanoSelectorWindow: string,
     buttonRow: string,
     geneSet: string,
     empty: string,
@@ -82,7 +84,10 @@ export default class GenesetsSelector extends QueryStoreComponent<GenesetsSelect
 
                     <FlexRow padded className={styles.buttonRow}>
                     <button className="btn btn-default btn-sm" onClick={() => this.store.showGenesetsHierarchyPopup = true}>
-                        Select Gene Sets
+                        Select Gene Sets from Hierarchy
+                    </button>
+                    <button className="btn btn-default btn-sm" onClick={() => this.store.showGenesetsVolcanoPopup = true}>
+                        Select Gene Sets from Volcano Plot
                     </button>
                     </FlexRow>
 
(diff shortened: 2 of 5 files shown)


jjgao:
Maybe volcano plot should be implemented as a new component in the frontend repo?

package-lock.json
```diff
@@ -88,6 +88,19 @@
         "@types/react": "15.6.2"
       }
     },
+    "@types/geojson": {
+      "version": "7946.0.0",
+      "resolved": "https://registry.npmjs.org/@types/geojson/-/geojson-7946.0.0.tgz",
+      "integrity": "sha512-lbWmXFxIpEzpH7OprsCRvxj7kie+248Y2ItjeVsF+0+IqvwG+R+0xgZmxq1ofYNTszvuihDahas7O5dscfxTsw=="
+    },
+    "@types/highcharts": {
```
alisman: what's reason for highcharts?

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,313 @@
...
+    onSelect: (map_genesets_selected:ObservableMap<boolean>) => void;
+}
+
+@observer
+export default class GenesetsVolcanoSelector extends React.Component<GenesetsVolcanoSelectorProps, {plotData:{x: number, y: number, fill: string}[]}>
+{
+    @observable tableData: Geneset[] = [];
+    @observable allTableData: Geneset[] = [];
+    maxY = 1;
+    plotData = autorun(() => {
```
alisman: elsewhere in the we've avoided using React state, instead just using observable properties and computed getters.  Is there a reason not to just make this @computed getter?

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,313 @@
...
+    componentDidUpdate()
+    {
+        if (this.percentileHasChanged) {
+            this.isLoading = true;
+            this.fetchData();
+            this.percentileHasChanged = false;
+        }
+    }
+    
+    private columns:IColumnDefMap = {
```
alisman: this is probably our fault for not having better documentation but we don't want to use ReactTable or EnhancedReactTable anymore.   I don't want to make you do this work though.   Let me take a look at the functionality in the browser and then get back to you.  sorry for this confusion.

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,313 @@
...
+                            )}
+                    </Observer>
+                    </div>
+                    )}
+                    </Td>
+                ),
+            },
+    };
+    
+    async fetchData() {
```
alisman: i think this should live in queryStore.  State/data should not be managed in component.

alisman: And we should use a mobxPromise to handle loading state.

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,313 @@
...
+    {
+        this.selectedPercentile = val || {label: '75%', value: '75'};
+        this.percentileHasChanged = true;
+    }
+    
+    updateSelectionFromPlot(points: any, bounds: any) {
+        const selectedPoints = [];
+        const newTableData: Geneset[] = [];
+        const newPlotData = this.state.plotData;
+        this.map_genesets_selected = new ObservableMap<boolean>();
```
alisman: we want to keep this state in the store, not in component

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,313 @@
...
+        if (points && points[0].data) {
```
alisman: use map or reduce to build/tranform objects and array.  it makes code clearer

src/shared/components/query/GenesetsVolcanoSelector.tsx
alisman: You probably don't need an ObservableMap, which is more costly.  For example, it's going to issue synchronous update events every time you call set down below.  Instead, build the table as a plain object and then either assign it to a observable.ref (look that up) or THEN instantiate an Observable Map with it if you really need to functionality of an ObservableMap

src/shared/components/query/QueryStore.ts
```diff
@@ -1110,6 +1114,64 @@ export class QueryStore
...
+            return tableData;
+        }
+    });
+    
+    readonly maxYVolcanoPlot = remoteData<number>({
+        await: () => [this.volcanoPlotTableData],
+        invoke: () => {
+            let maxY = 1;
+            const tableData = this.volcanoPlotTableData.result!;
+            tableData.map(({representativePvalue}) => {
```
alisman: seems like maybe this "map" is not necessary?  representativePvalue doesn't seem to be used anywhere.

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -0,0 +1,220 @@
...
+        this.percentileChange = this.percentileChange.bind(this);
+        this.updateSelectionFromPlot = this.updateSelectionFromPlot.bind(this);
+    }
+    
+    percentileChange(val: {label: string, value: string} | null)
+    {
+        this.store.volcanoPlotSelectedPercentile = val || {label: '75%', value: '75'};
+    }
+    
+    updateSelectionFromPlot(points: any, bounds: any) {
```
alisman: this function should be @action since it updates observable state (map_genesets_selected_volcano) and does so multiple times.  @action will allow batch updating. this is critical to performance because it prevents duplicate render cycles.
also, this should have a unit test that makes sure map_genesets_selected_volcano is update properly.

src/shared/components/query/QueryStore.ts
```diff
@@ -1104,6 +1108,62 @@ export class QueryStore
     {
         return this.genesetIdsQuery.query;
     }
+    
+    readonly volcanoPlotTableData = remoteData<Geneset[]>({
+        invoke: async()=>{
+            const tableData: Geneset[] = [];
+            const hierarchyData = await getHierarchyData(
```
alisman: this fetch should go in a util file.  also it should be decoupled from the logic which processes the result so that you can easily write a test of the processing logic.  by processing logic i mean the for loop that populates the table.

src/shared/components/query/QueryStore.ts
```diff
@@ -1104,6 +1108,62 @@ export class QueryStore
...
     }
+    
+    readonly volcanoPlotTableData = remoteData<Geneset[]>({
+        invoke: async()=>{
+            const tableData: Geneset[] = [];
+            const hierarchyData = await getHierarchyData(
+                    this.getFilteredProfiles("GENESET_SCORE")[0].molecularProfileId, 
+                    Number(this.volcanoPlotSelectedPercentile.value),
+                    0, 1, this.defaultSelectedSampleListId);
+            for (const node of hierarchyData) {
```
alisman: when you are writing logic which populates a variable (in this case tableData), you should try to use either map or reduce.  in this case, reduce would work because you are filtering.  the reason this is better is because the code is more semantic.  const result  = source.reduce(...).    a reader can see immediately that the aim of this code is to set the result variable to the return value of reduce.

src/shared/components/query/GenesetsSelector.tsx
```diff
@@ -114,12 +119,36 @@ export default class GenesetsSelector extends QueryStoreComponent<GenesetsSelect
...
+                    >
+                    <Modal.Header closeButton>
+                    <Modal.Title>Select Gene Sets From Volcano Plot</Modal.Title>
+                    </Modal.Header>
+                    <Modal.Body>
+                    <GenesetsVolcanoSelector
+                        initialSelection={this.store.genesetIds}
+                        data={this.store.volcanoPlotTableData.result}
+                        plotData={this.store.volcanoPlotGraphData}
+                        maxY={this.store.maxYVolcanoPlot ? -(Math.log(this.store.maxYVolcanoPlot)/Math.log(10)) : undefined}
```
alisman: i think this will show an empty chart.  you should check to see if the promise on which moxYVolcanoPlot depends isComplete

oplantalech: No because it is the return data of get maxYVolcano plot: if the table data is not there, this function returns 'undefined', else returns the number.

PR #831, opened 2017-12-19
Show user virtual studies on home page

kalletlak:
This is related to #https://github.com/cBioPortal/cbioportal/issues/3450

TODO:
- [x] When virtual study is deleted, make sure its disabled


src/shared/api/sessionServiceAPI.ts
```diff
@@ -0,0 +1,59 @@
+import * as request from 'superagent';
+import {getSessionServiceApiUrl} from "./urls";
+import {IVirtualStudy} from "shared/model/VirtualStudy";
+
+
+type VirtualStudyData = {
```
zhx828: No users?

src/shared/components/query/QueryStore.ts
```diff
@@ -192,61 +174,16 @@ export class QueryStore
 		return Object.keys(ret);
 	}
 
-	readonly temporaryVirtualCohortId = remoteData({
```
zhx828: Could you give me some info how the temporaryVirtualCohort works in the new implementation?

src/shared/api/sessionServiceAPI.ts
```diff
@@ -0,0 +1,59 @@
...
+                    });
+                });              
+    }
+
+    deleteVirtualStudy(id:string){
+        return request
+                .delete(`${getSessionServiceApiUrl()}/${id}`)
+                             
+    }
+    deleteVirtualStudy1(id:string) {
```
zhx828: What's this one for?

src/shared/api/sessionServiceAPI.ts
kalletlak: not all fields are required in the front-end. Unused and unwanted fields are filtered in the session-service proxy api

src/shared/api/sessionServiceAPI.ts
kalletlak: hmm... good catch!

src/shared/components/query/QueryStore.ts
kalletlak: `temporaryVirtualCohort` was legacy and bit hard to understand (btw that method is not invoked at all). So just moved the api call and object mapping to `src/shared/api/sessionServiceAPI.ts`

src/shared/api/sessionServiceAPI.ts
```diff
@@ -0,0 +1,59 @@
...
+                .delete(`${getSessionServiceApiUrl()}/${id}`)
+                             
+    }
+
+    addVirtualStudy(id:string){
+        return request
+                .put(`${getSessionServiceApiUrl()}/${id}`)
+                             
+    }
+}
```
inodb: @kalletlak Instead of using a custom client. Generate the typescript client from the swagger JSON of session-service: http://cbioportal-session-service.herokuapp.com/swagger-ui.html. See: https://github.com/cBioPortal/cbioportal-frontend/blob/master/package.json#L16-L18

kalletlak: @inodb This client doesn't directly refer to session-service API. It is exposed through a wrapper which is in cbio web `api-legacy`. Will discuss this on slack.

src/shared/components/query/QueryAndDownloadTab.spec.tsx
```diff
@@ -1,18 +1,107 @@
...
+
+        it("View VirtualStudies", ()=>{
+            assert.deepEqual(
+               Object.keys(store.selectableStudiesSet).length,
+               component.find('[data-test="StudySelect"]').getNodes().length);
+            assert.equal(component.find('[data-test="StudySelect"]').find('.fa-trash').getNodes().length, virtualStudies.length);
+        });
+
+         it("Delete VirtualStudy", ()=>{
+          let button2 = component.find('.fa-trash').at(0)
```
inodb: @kalletlak we usually add `data-test="virtualStudyItem"` attributes but this works as well. Could just lead to some strange errors if we end up using `.fa-trash` at multiple locations

kalletlak: Sure! will update this

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -43,7 +43,7 @@ export default class QuerySummary extends React.Component<{ queryStore:QueryStor
 
     private get singleStudyUI() {
         return <div>
-            <h4><StudyLink study={this.props.store.studies.result[0]}/></h4>
+            <h4><StudyLink study={this.props.queryStore.selectedStudies[0]}/></h4>
```
alisman: I'm not sure this makes sense.  selectedStudies in queryStore will change prior to user submitting the new query.  We don't want to this change just when user checks checkboxes.

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -63,14 +64,20 @@ export default class CancerStudyTreeData
 		const virtualCohortStudies = [];
 		for (let virtualCohort of virtualCohorts) {
 			let study = {
-				allSampleCount: virtualCohort.samples.length,
+				//allSampleCount: [].concat.apply([], virtualCohort.data.studies.map(study => study.samples)).length,
+				allSampleCount:virtualCohort.data.studies
```
alisman: i think you can use _.sumBy to make this cleaner

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -63,14 +64,20 @@ export default class CancerStudyTreeData
...
+														return prev.concat(curr);
+													})
+													.length,
 				studyId: virtualCohort.id,
-				name: virtualCohort.name,
-				description: virtualCohort.description,
+				name: virtualCohort.data.name,
+				description: virtualCohort.data.description,
 				cancerTypeId: virtualCohortsName
 			} as CancerStudy;
```
alisman: if at all possible, avoid using "as" because actually then lose type checking on this object.  try to actually construct a study that satisfies typing.   then you can do const study: CancerStudy = {completes study}

src/shared/components/query/QueryStore.ts
```diff
@@ -143,46 +145,35 @@ export class QueryStore
 	}
 
 	@observable studiesHaveChangedSinceInitialization:boolean = false;
-	@observable savedVirtualCohorts:VirtualCohort[] = [];
+	@observable userVirtualStudies:VirtualStudy[] = [];
+	@observable deletedVirtualStudies:string[] = [];
 
 	@action public deleteVirtualCohort(id:string) {
-		VirtualCohorts.delete(id);
+		sessionServiceClient.deleteVirtualStudy(id)
```
alisman: i think it's antipattern to tag an async method with action.   instead action should just suround the code which actually mutation state.

alisman: see mobx action documentation for how to do that.

src/shared/components/query/QueryStore.ts
```diff
@@ -143,46 +145,35 @@ export class QueryStore
...
-	@computed get virtualCohorts():VirtualCohort[] {
-		const ret:VirtualCohort[] = [];
-		// if (this.temporaryVirtualCohort.result) {
-		// 	ret.push(this.temporaryVirtualCohort.result);
-		// }
-		// for (let i=0; i<this.savedVirtualCohorts.length; i++) {
-		// 	ret.push(this.savedVirtualCohorts[i]);
-		// }
-		return ret;
+	@computed get virtualCohorts():VirtualStudy[] {
```
alisman: maybe we get rid of this getter since we're just accessing another property?  we certainly don't need computed decorator here.

src/shared/components/query/QueryStore.ts
```diff
@@ -143,46 +145,35 @@ export class QueryStore
...
+		return this.userVirtualStudies;
 	}
 
-	@computed get virtualCohortsSet():{[id:string]:VirtualCohort} {
-		return this.virtualCohorts.reduce((acc:{[id:string]:VirtualCohort}, next:VirtualCohort)=>{
+	@computed get virtualCohortsSet():{[id:string]:VirtualStudy} {
+		return this.virtualCohorts.reduce((acc:{[id:string]:VirtualStudy}, next:VirtualStudy)=>{
```
alisman: i think we could use _.keyBy to accomplish this in more readable fashion

src/shared/components/query/QueryStore.ts
```diff
@@ -195,7 +186,7 @@ export class QueryStore
 		for (const studyId of this.selectedStudyIds) {
 			const vc = virtualCohortsSet[studyId];
 			if (vc) {
-				for (const constStudyId of vc.constituentStudyIds) {
+				for (const constStudyId of vc.data.studies.map(study => study.id)) {
```
alisman: i think you could just use data.studies.each((study)=>ret[study.id]=true)

src/shared/components/query/QueryStore.ts
```diff
@@ -297,8 +230,28 @@ export class QueryStore
 	{
 		let ids:string[] = this._selectedStudyIds.keys();
 		const selectableStudies = this.selectableStudiesSet;
-		ids = ids.filter(id=>!!selectableStudies[id]);
-		return this.forDownloadTab ? ids.slice(-1) : ids;
+		let selectableIds = ids.filter(id=>!!selectableStudies[id]);
```
alisman: this logic could use some commenting.  someone coming to it without background would struggle to understand what is going on with "selectable" studies and also the normalization.  (I do realize a lot of this is legacy code from another author but no time like the present to make it easier to understand)

src/shared/components/query/QueryStore.ts
```diff
@@ -297,8 +230,28 @@ export class QueryStore
...
 		const selectableStudies = this.selectableStudiesSet;
-		ids = ids.filter(id=>!!selectableStudies[id]);
-		return this.forDownloadTab ? ids.slice(-1) : ids;
+		let selectableIds = ids.filter(id=>!!selectableStudies[id]);
+
+		if(ids.length !== selectableIds.length) {
+			let entities = this.caseIds.trim().split(/\s+/g);
+			let normStudies:string [] = entities
+			.map(entity=>{
+				let splitEntity = entity.split(':');
```
alisman: since it seems we don't use the id portion, you could just make a regexp that matches the string prior to :.   This will then work for all strings, even those without : compound nature.

src/shared/components/query/QueryStore.ts
```diff
@@ -297,8 +230,28 @@ export class QueryStore
...
 		let ids:string[] = this._selectedStudyIds.keys();
 		const selectableStudies = this.selectableStudiesSet;
-		ids = ids.filter(id=>!!selectableStudies[id]);
-		return this.forDownloadTab ? ids.slice(-1) : ids;
+		let selectableIds = ids.filter(id=>!!selectableStudies[id]);
+
+		if(ids.length !== selectableIds.length) {
+			let entities = this.caseIds.trim().split(/\s+/g);
+			let normStudies:string [] = entities
+			.map(entity=>{
```
alisman: this should be a reduce function which can then map and filter at the same time and won't produce the undefined items.

src/shared/components/query/QueryStore.ts
```diff
@@ -297,8 +230,28 @@ export class QueryStore
...
+				let splitEntity = entity.split(':');
+				if (splitEntity.length === 2) {
+					const study = splitEntity[0];
+					const id = splitEntity[1];
+					return study;
+				}
+			})
+			.filter(element => element !== undefined) as string[];
+
+			normStudies = _.uniq(normStudies);
```
alisman: you can also get of this uniq by just checking to see if the string is already in the reduce memo above.

src/shared/components/query/QueryStore.ts
```diff
@@ -297,8 +230,28 @@ export class QueryStore
...
+
+			selectableIds = normStudies.filter(id=>!!selectableStudies[id]);
```
alisman: same thing here.  this can be accomplished inside reduce and without extra iteration

src/shared/components/query/QueryStore.ts
```diff
@@ -143,46 +145,35 @@ export class QueryStore
...
 	@observable studiesHaveChangedSinceInitialization:boolean = false;
-	@observable savedVirtualCohorts:VirtualCohort[] = [];
+	@observable userVirtualStudies:VirtualStudy[] = [];
+	@observable deletedVirtualStudies:string[] = [];
 
 	@action public deleteVirtualCohort(id:string) {
-		VirtualCohorts.delete(id);
+		sessionServiceClient.deleteVirtualStudy(id)
+		.then(() => {
+			this.deletedVirtualStudies.push(id)
```
alisman: we probaby don't need to bother waiting for request to respond remove it from deletedVirtualStudies.  If there's a service error, then we show an error message, otherwise, assume success?

src/shared/components/query/QueryStore.ts
```diff
@@ -143,46 +145,35 @@ export class QueryStore
...
-				constituentStudyIds.push(selectedCasesObj.studyID);
-			}
-			return {
-				id: x.virtualCohortID,
-				name: x.studyName,
-				description: x.description,
-				samples,
-				constituentStudyIds
-			};
+		sessionServiceClient.getUserVirtualStudies().then((response) => {
```
alisman: we could probably use remoteData for this, to keep consistency

alisman:
@kalletlak apologies I took so long to really look at this.  There are number of things to discuss. Let me know if you want to chat at some point.

kalletlak:
@alisman thanks for your time in reviewing this pr. Let me do a first round of changes based on your comments. Will ping you on slack if I have any questions.
Thanks!

src/pages/resultsView/querySummary/QuerySummary.tsx
kalletlak: Yeah, True. Currenlt `ResultsViewPageStore` studies (`store.studies.result`) would just have predefined studies. we should even render virtual studies in this case.

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -216,94 +228,117 @@ export default class StudyList extends QueryStoreComponent<IStudyListProps, {}>
...
+                    {study.studyId && (
+                        <DefaultTooltip
+                            mouseEnterDelay={0}
+                            placement="top"
+                            overlay={
+                                <div className={styles.tooltip}
+                                >View study summary</div>
+                            }
+                            children={
+                                <span onClick={()=>openStudySummaryFormSubmit(study.studyId)}
```
alisman: maybe just put this inside tag instead of assigning to children property

src/shared/components/query/QueryStore.ts
kalletlak: True, I should have added some comments here. Now I don't remember why I added this block. For now I'll revert it back.

src/shared/components/query/studyList/StudyList.tsx
kalletlak: true, I just changes the parent logic and ported everything what was present. I'll try to update this

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1248,7 +1252,44 @@ export class ResultsViewPageStore {
...
+    
+    readonly virtualStudies = remoteData(sessionServiceClient.getUserVirtualStudies(), []);
+    
+    readonly virtualStudyIdToStudy = remoteData({
+        await: ()=>[this.virtualStudies],
+        invoke: async ()=>{
+            return _.keyBy(
+                this.virtualStudies.result.map(virtualStudy=>{
+                    let study = {
+                        allSampleCount:_.sumBy(virtualStudy.data.studies, function(study) { return study.samples.length }),
```
alisman: use arrow function

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -43,7 +43,7 @@ export default class QuerySummary extends React.Component<{ queryStore:QueryStor
 
     private get singleStudyUI() {
         return <div>
-            <h4><StudyLink study={this.props.store.studies.result[0]}/></h4>
+            <h4><StudyLink study={this.props.store.userQueryStudies.result[0]}/></h4>
```
alisman: we need to be 100% sure we get all references to studies.  TS can really help us with that.   1. rename store.studies to something like nonVirtualStudies (@adamabeshouse comment?)  2. refactor studyIdToSummary to use that.   3. TS is going to complain for every remaining instance of studies.   You could then fix all the ts errors with the new userQueryStudies and be sure you're getting everything.    I would actually change userQueryStudies to just queriedStudies

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -63,14 +64,14 @@ export default class CancerStudyTreeData
 		const virtualCohortStudies = [];
 		for (let virtualCohort of virtualCohorts) {
 			let study = {
-				allSampleCount: virtualCohort.samples.length,
+				allSampleCount:_.sumBy(virtualCohort.data.studies, function(study) { return study.samples.length }),
```
alisman: use arrow function

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -84,9 +85,8 @@ export default class CancerStudyTreeData
 				cancerTypeId: name
 			});
 		}
-		//cancerTypes = [virtualCohortsCategory].concat(this.priorityCategories).concat(this.rootCancerType, cancerTypes);
-		//studies = virtualCohortStudies.concat(studies);
-		cancerTypes = this.priorityCategories.concat(this.rootCancerType, cancerTypes);
+		cancerTypes = [virtualCohortsCategory].concat(this.priorityCategories).concat(this.rootCancerType, cancerTypes);
```
alisman: not too important but look at new es6 spread operator to do this kind of concatenation

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -84,9 +85,8 @@ export default class CancerStudyTreeData
 				cancerTypeId: name
 			});
 		}
-		//cancerTypes = [virtualCohortsCategory].concat(this.priorityCategories).concat(this.rootCancerType, cancerTypes);
-		//studies = virtualCohortStudies.concat(studies);
-		cancerTypes = this.priorityCategories.concat(this.rootCancerType, cancerTypes);
+		cancerTypes = [virtualCohortsCategory].concat(this.priorityCategories).concat(this.rootCancerType, cancerTypes);
+		studies = CancerStudyTreeData.sortNodes(virtualCohortStudies).concat(studies);
```
alisman: Is it right at we sort the nodes first and THEN concat?  Do we want to sort after concatenation?

src/shared/components/query/QueryStore.ts
```diff
@@ -143,126 +143,43 @@ export class QueryStore
...
-			return {
-				id: x.virtualCohortID,
-				name: x.studyName,
-				description: x.description,
-				samples,
-				constituentStudyIds
-			};
-		});
-	}
+	@action
```
alisman: you should not put @action decorator on an async method.   The actual mutation of state will end up executing outside of the action, thus nullifying purpose of action.   See documentation for further explaination.  https://mobx.js.org/best/actions.html.   i would reccomend using action() around the actual mutation  (inside callback)

src/pages/resultsView/querySummary/QuerySummary.tsx
kalletlak: `userQueryStudies` is only used to render summary section on the query(mainly name and description). and `studies` contains physical studies in queried session. 
Anyways will rename `studies` to `physicalStudies` and `userQueryStudies` to `queriedStudies`

src/shared/components/query/CancerStudyTreeData.ts
kalletlak: I think it doesn't matter. Will clean it up.

src/pages/resultsView/querySummary/QuerySummary.tsx
adamabeshouse: nonVirtualStudies makes sense

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1248,10 +1252,62 @@ export class ResultsViewPageStore {
...
+                    return study;
+                }), x =>x.studyId);
+        }
+    },{});
+
+    //this is only required to show study name and description on the results page
+    readonly queriedStudies = remoteData({
+		await: ()=>[this.studyIdToStudy, this.virtualStudyIdToStudy],
+		invoke: async ()=>{
+            const queriedStudies = [];
```
alisman: i would take this and put it into a helper file.  this function should be pure and accept as arguments all the instance properties it uses (i.e. this references)

src/shared/components/query/QueryStore.ts
```diff
@@ -296,13 +225,21 @@ export class QueryStore
...
+			if(selectableStudies[next]){
+				return obj.concat(selectableStudies[next])
+			}
+			return obj;
+
+		},[]);
 		return this.forDownloadTab ? ids.slice(-1) : ids;
 	}
 
 	set selectedStudyIds(val:string[]) {
```
alisman: this should be an @action since it sets an observable value

src/shared/components/query/QueryStore.ts
```diff
@@ -473,20 +417,46 @@ export class QueryStore
...
-		}
-		return ret;
-	}
+	readonly virtualStudies = remoteData(sessionServiceClient.getUserVirtualStudies(), []);
+
+	readonly selectableIdToPhysicalIdsMap = remoteData<{[studyId:string]:string[]}>({
+		await: ()=>[this.cancerStudies, this.virtualStudies],
+	    invoke: async ()=>{
+			let physicalStudiesIdsMap:{[studyId:string]:string[]} = this.cancerStudies.result.reduce((obj:{[studyId:string]:string[]}, item) =>{
+				obj[item.studyId] = [item.studyId]
```
alisman: how come the value here is an array of studies?  is this because a virtual study can be composed of multiple physical studies?  but if that's so, i would expect that virtual studies below would allow for multiple elements in the value, but i don't see that.  it just seems like an array of one element.

src/shared/components/query/QueryStore.ts
```diff
@@ -473,20 +417,46 @@ export class QueryStore
...
+
+			const unknownIds:string[] = this._defaultSelectedIds.keys().filter(id => !knownSelectableIds[id]);
+
+			await Promise.all(unknownIds.map(id =>{
+				return new Promise((resolve, reject) => {
+					sessionServiceClient.getVirtualStudy(id).then((response)=>{
+						knownSelectableIds[id] = response.data.studies.map(study=>study.id)
+						resolve();
+					}).catch(() => {
+						resolve();
```
alisman: I think we should handle this better.  If we get an error code if that id doesn't exists in session then we should message the user because the query is basically invalid.  also, we should wait until we have all responses before assigning to knownSelectableIds.    i believe that Promise.all().then() would give you aggregated collection of responses.  Make sense?

alisman: I will let you know what the right way is to do an error message shortly.

src/shared/components/query/QueryStore.ts
```diff
@@ -807,24 +777,52 @@ export class QueryStore
...
 	@computed get selectedStudies_totalSampleCount()
 	{
-		return this.selectedStudies.reduce((sum:number, study:CancerStudy) => sum + study.allSampleCount, 0);
+		const result:{[id:string]:number} = {};
+		const virtualStudySamples:{[id:string]:string[]} = {};
+
+		this.selectedStudies.forEach(study => {
+			//merge samples for a study across all virtual studies 
+			if(this.isVirtualCohort(study.studyId)){
+				this.virtualStudiesSet[study.studyId]
```
alisman: lets make this explicit with if statement.  too hard to read otherwise

src/shared/components/query/QueryStore.ts
```diff
@@ -807,24 +777,52 @@ export class QueryStore
...
+						})
+			}else{
+				//add selected physical studies samples count to result
+				result[study.studyId] =study.allSampleCount;
+			}
+		})
+		//if the physical study in virtual study is not present result object, then 
+		//add the study and samples count into result object
+		Object.keys(virtualStudySamples).forEach(id => {
+			if(!result[id])
```
alisman: use brackets

src/shared/components/query/QueryStore.ts
```diff
@@ -807,24 +777,52 @@ export class QueryStore
...
 
 	public isVirtualCohort(studyId:string):boolean {
 		// if the study id doesn't correspond to one in this.cancerStudies, then its a virtual cohort
 		return !this.cancerStudyIdsSet.result[studyId];
 	}
 
-	public isTemporaryVirtualCohort(studyId:string):boolean {
-		return this.temporaryVirtualCohortId.isComplete && this.temporaryVirtualCohortId.result === studyId;
+	public isDeletedVirtualStudy(studyId:string):boolean {
+		if(this.isVirtualCohort(studyId) && this.deletedVirtualStudies.indexOf(studyId) > -1){
```
alisman: how come we need to know that something is a deleted virtual study?  seems like this should be handled simple by removal of study from collection of studies.

src/shared/components/query/QueryStore.ts
```diff
@@ -296,13 +225,21 @@ export class QueryStore
...
+			}
+			return obj;
+
+		},[]);
 		return this.forDownloadTab ? ids.slice(-1) : ids;
 	}
 
 	set selectedStudyIds(val:string[]) {
-		this._selectedStudyIds = observable.map(stringListToSet(val));
+		//filter out deleted virtual study
```
alisman: yeah, i don't think this is right.  we shouldn't need to filter out deleted studies at this point.  is there a reason we need to keep track of deleted studies?  (as in, we need to display them in some deleted list).

src/shared/components/query/StudyListLogic.ts
```diff
@@ -223,10 +223,28 @@ export class FilteredCancerTreeView
 		{
 			let study = node as CancerStudy;
 			let checked = !!this.store.selectedStudyIds.find(id => id == study.studyId);
-			return {checked};
+			let disabled = this.store.isDeletedVirtualStudy(study.studyId);
```
alisman: i see, we are continuing to show deleted studies, but disabled.  was that a requirement?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -793,7 +797,7 @@ export class ResultsViewPageStore {
     });
 
     public get studyMap():{ [studyId:string]:CancerStudy } {
```
alisman: this should be physicalStudyMap

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1248,10 +1252,62 @@ export class ResultsViewPageStore {
...
+                                        this.virtualStudyIdToStudy.result) as {[id:string]:CancerStudy}; 
+
+            for(const cohortId of this.cohortIdsList){
+                if(cohorts[cohortId]){
+                    queriedStudies.push(cohorts[cohortId])
+                }
+            }
+
+            if(this.cohortIdsList.length === 1 && queriedStudies.length === 0){
+                return sessionServiceClient.getVirtualStudy(this.cohortIdsList[0]).then(function(virtualStudy){
```
alisman: i think this should be part of virtualStudies promise.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -717,10 +721,10 @@ export class ResultsViewPageStore {
         await: () => [
             this.samples,
             this.clinicalDataForSamples,
-            this.studies
+            this.physicalStudies
         ],
         invoke: () => {
-            return Promise.resolve(extendSamplesWithCancerType(this.samples.result, this.clinicalDataForSamples.result,this.studies.result));
+            return Promise.resolve(extendSamplesWithCancerType(this.samples.result, this.clinicalDataForSamples.result,this.physicalStudies.result));
```
alisman: return physicalStudies to studies

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -39,19 +40,18 @@ export default class CancerStudyTreeData
 	map_cancerTypeId_cancerType = new Map<string, CancerType>();
 	map_studyId_cancerStudy = new Map<string, CancerStudy>();
 
-	constructor({cancerTypes = [], studies = [], priorityStudies = {}, virtualCohorts=[]}: {cancerTypes: CancerType[], studies: CancerStudy[], priorityStudies?:PriorityStudies, virtualCohorts?:VirtualCohort[]})
+	constructor({cancerTypes = [], studies = [], priorityStudies = {}, virtualCohorts=[]}: {cancerTypes: CancerType[], studies: CancerStudy[], priorityStudies?:PriorityStudies, virtualCohorts?:VirtualStudy[]})
```
alisman: lets get rid entirely of virtualChorts language.  call it virtualStudies

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -39,19 +40,18 @@ export default class CancerStudyTreeData
...
 		let meta:NodeMetadata;
 
 		// sort by name
 		cancerTypes = CancerStudyTreeData.sortNodes(cancerTypes);
-		studies = CancerStudyTreeData.sortNodes(studies);
 
 		// add virtual cohort category, and studies
 		// disabled, for now
-		/*const virtualCohortsName = "My Virtual Studies";
+		const virtualCohortsName = "My Virtual Studies";
```
alisman: VIRTUAL_COHORTS_NAME

src/shared/components/query/QueryStore.ts
```diff
@@ -143,126 +143,55 @@ export class QueryStore
...
-		const ret:VirtualCohort[] = [];
-		// if (this.temporaryVirtualCohort.result) {
-		// 	ret.push(this.temporaryVirtualCohort.result);
-		// }
-		// for (let i=0; i<this.savedVirtualCohorts.length; i++) {
-		// 	ret.push(this.savedVirtualCohorts[i]);
-		// }
-		return ret;
-	}
+	addVirtualStudy(id:string) {
```
alisman: restoreVirtualStudy

src/shared/components/query/QueryStore.ts
```diff
@@ -143,126 +143,55 @@ export class QueryStore
...
+		)
+	};
 
-	@computed get virtualCohortsSet():{[id:string]:VirtualCohort} {
-		return this.virtualCohorts.reduce((acc:{[id:string]:VirtualCohort}, next:VirtualCohort)=>{
-			acc[next.id] = next;
-			return acc;
-		}, {});
+	@computed get virtualStudiesSet():{[id:string]:VirtualStudy} {
+		return _.keyBy(this.virtualStudies.result, study => study.id);
```
alisman: virtualStudiesMap

src/shared/components/query/QueryStore.ts
```diff
@@ -143,126 +143,55 @@ export class QueryStore
...
 	}
 
 	@computed get studyIdsInSelection():string[] {
 		// Gives selected study ids and study ids that are in selected virtual cohorts
-		const virtualCohortsSet = this.virtualCohortsSet;
 		const ret:{[id:string]:boolean} = {};
 		for (const studyId of this.selectedStudyIds) {
-			const vc = virtualCohortsSet[studyId];
+			const vc = this.virtualStudiesSet[studyId];
```
alisman: no abbr

src/shared/components/query/QueryStore.ts
```diff
@@ -143,126 +143,55 @@ export class QueryStore
...
-	@computed get virtualCohortsSet():{[id:string]:VirtualCohort} {
-		return this.virtualCohorts.reduce((acc:{[id:string]:VirtualCohort}, next:VirtualCohort)=>{
-			acc[next.id] = next;
-			return acc;
-		}, {});
+	@computed get virtualStudiesSet():{[id:string]:VirtualStudy} {
+		return _.keyBy(this.virtualStudies.result, study => study.id);
 	}
 
 	@computed get studyIdsInSelection():string[] {
```
alisman: make this physicalStudyIdsInSelection

alisman: put a comment

src/shared/components/query/QueryStore.ts
alisman: action(()=>)

src/shared/components/query/QueryStore.ts
```diff
@@ -321,8 +258,15 @@ export class QueryStore
 			}
 		}
 	}
-	private isStudyIdSelected(studyId:string):boolean {
-		return !!this._selectedStudyIds.get(studyId);
+
+	@observable private _defaultSelectedIds:ObservableMap<boolean> = observable.map<boolean>();
```
alisman: should not be observable.  initiallySelectedIds

src/shared/components/query/QueryStore.ts
```diff
@@ -321,8 +258,15 @@ export class QueryStore
 			}
 		}
 	}
-	private isStudyIdSelected(studyId:string):boolean {
-		return !!this._selectedStudyIds.get(studyId);
+
+	@observable private _defaultSelectedIds:ObservableMap<boolean> = observable.map<boolean>();
+	
+	@action private setDefaultSelectedIds(studyId:string, selected:boolean) {
```
alisman: get rid of method

src/shared/components/query/QueryStore.ts
```diff
@@ -807,24 +777,52 @@ export class QueryStore
 
 	@computed get unknownStudyIds()
```
alisman: comment explaining unknown

src/shared/components/query/QueryStore.ts
```diff
@@ -807,24 +777,52 @@ export class QueryStore
...
+				result[study.studyId] =study.allSampleCount;
+			}
+		})
+		//if the physical study in virtual study is not present result object, then 
+		//add the study and samples count into result object
+		Object.keys(virtualStudySamples).forEach(id => {
+			if(!result[id])
+				result[id] = _.uniq(virtualStudySamples[id]).length
+		});
+		return Object.keys(result).reduce((sum, id) => sum + result[id], 0);
```
alisman: sumBy

kalletlak: `_.sumBy` works on List. In our case its a object

src/shared/components/query/StudyListLogic.ts
kalletlak: https://github.com/cBioPortal/cbioportal/issues/3777

PR #837, opened 2017-12-20
Only show data priority options depending on profile availability for…

adamabeshouse:
… multiple study query form

### DIFF
--- a/src/shared/components/query/DataTypePrioritySelector.spec.tsx
+++ b/src/shared/components/query/DataTypePrioritySelector.spec.tsx
@@ -1,0 +1,109 @@
+import {assert} from 'chai';
+import React from 'react';
+import expect from 'expect';
+import expectJSX from 'expect-jsx';
+import {shallow} from "enzyme";
+import {profileAvailability, radioButtons} from "./DataTypePrioritySelector";
+import {AlterationTypeConstants} from "../../../pages/resultsView/ResultsViewPageStore";
+import {MolecularProfile} from "../../api/generated/CBioPortalAPI";
+import {stringListToSet} from "../../lib/StringUtils";
+import {QueryStore} from "./QueryStore";
+
+expect.extend(expectJSX);
+
+describe("DataTypePrioritySelector", ()=>{
+    describe("profileAvailability", ()=>{
+        it("returns correct profile availability result in case of zero profiles", ()=>{
+            assert.deepEqual(profileAvailability([]), {mutation:false, cna:false});
+        });
+        it("returns correct profile availability result in case of one profile", ()=>{
+            let profiles = [{
+                molecularAlterationType: AlterationTypeConstants.MUTATION_EXTENDED,
+                showProfileInAnalysisTab: true
+            }] as MolecularProfile[];
+            assert.deepEqual(profileAvailability(profiles), {mutation:true, cna:false});
+
+            profiles = [{
+                molecularAlterationType: AlterationTypeConstants.MUTATION_EXTENDED,
+                showProfileInAnalysisTab: false
+            }] as MolecularProfile[];
+            assert.deepEqual(profileAvailability(profiles), {mutation:false, cna:false});
+
+            profiles = [{
+                molecularAlterationType: AlterationTypeConstants.COPY_NUMBER_ALTERATION,
+                showProfileInAnalysisTab: true
+            }] as MolecularProfile[];
+            assert.deepEqual(profileAvailability(profiles), {mutation:false, cna:true});
+
+            profiles = [{
+                molecularAlterationType: AlterationTypeConstants.COPY_NUMBER_ALTERATION,
+                showProfileInAnalysisTab: false
+            }] as MolecularProfile[];
+            assert.deepEqual(profileAvailability(profiles), {mutation:false, cna:false});
+        });
+        it("returns correct profile availability result in case of two profiles", ()=>{
+            let profiles = [{
+                molecularAlterationType: AlterationTypeConstants.MUTATION_EXTENDED,
+                showProfileInAnalysisTab: true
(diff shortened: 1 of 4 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -477,6 +479,14 @@ export class QueryStore
 		}
 	});
 
+	readonly molecularProfilesInSelectedStudies = remoteData<MolecularProfile[]>({
+		invoke: async()=>{
+			const profiles:CacheData<MolecularProfile[], string>[] =
+				await this.molecularProfilesInStudyCache.getPromise(this.selectedStudyIds, true);
+			return _.flatten(profiles.map(d=>(d.data ? d.data : [])));
```
alisman: what's reason that data would not be present?

src/shared/components/query/DataTypePrioritySelector.tsx
```diff
@@ -20,33 +24,84 @@ export default class DataTypePrioritySelector extends QueryStoreComponent<{}, {}
...
+		buttons.push(<DataTypePriorityRadio key="C" label={`${hasBoth ? 'Only ' : ''}CNA`} state={{mutation:false, cna:true}}/>);
+	}
+	return buttons;
+}
+
+export function profileAvailability(molecularProfiles:MolecularProfile[]) {
+	let hasMutationProfile = false;
+	let hasCNAProfile = false;
+	for (const profile of molecularProfiles) {
+		if (!profile.showProfileInAnalysisTab)
```
alisman: just curious but what is reason a profile would have this set to false?

alisman: showProfileInAnalysisTab

src/shared/components/query/DataTypePrioritySelector.spec.tsx
```diff
@@ -0,0 +1,108 @@
...
+            },{
+                molecularAlterationType: AlterationTypeConstants.COPY_NUMBER_ALTERATION,
+                showProfileInAnalysisTab: false
+            }] as MolecularProfile[];
+            assert.deepEqual(profileAvailability(profiles), {mutation:true, cna:false});
+        });
+    });
+    describe("radioButtons",()=>{
+        it("shows the right buttons when theres mutation and cna profiles available",()=>{
+            const buttonLabels = stringListToSet(radioButtons({mutation:true, cna:true}).map(x=>x.props.label));
```
alisman: i would use enzyme here.  it's more semantic and doing so in general would allow easier extension of the tests if necessary.

src/shared/components/query/QueryStore.ts
adamabeshouse: if theres an error in the cache loading then data would be null

src/shared/components/query/DataTypePrioritySelector.spec.tsx
alisman: one could extract the label logic into a pure function so as to avoid the difficult of testing component itself.  but actually i think it's better to test component because it makes it easier to test other behaviors of the component, should there be any. let me know if that makes sense.

src/shared/components/query/DataTypePrioritySelector.tsx
adamabeshouse: im not sure but it is a parameter

src/shared/components/query/DataTypePrioritySelector.spec.tsx
adamabeshouse: thats more complicated than necessary imo

adamabeshouse: when writing this test I'm thinking, what is the behavior thats really at risk of regression? I dont think theres any reason to test for a regression in what specifically are the elements, what relative position, etc.

adamabeshouse:
@ersin I think it still makes sense this way..

ersinciftci:
@adamabeshouse Actually http://yeoman.io/contributing/testing-guidelines.html says your way is better, so I guess there is no convention yet.

ersinciftci:
@adamabeshouse Btw my point was not about making sense, it was about conventions.

adamabeshouse:
I dont see why we should always adhere to conventions just because they're a convention... in this case it makes just as much sense and doesnt create any problems.

ersinciftci:
Because consistency across the project is a good thing? And the best way to pick which way to go is to look at the conventions, so that it would be consistency with even other projects.

adamabeshouse:
but this isnt code consistency, its english language consistency. I just really dont see the point except for dogma

ersinciftci:
When I look at the test execution report, I personally like seeing the test names in the same format. But it's just me and we obviously don't need to change it just because I like it. And since there is apparently no convention on this, I'll change my tests and remove `should`s so that the format would be the same.
screen shot 2017-12-21 at 5 36 38 pm

adamabeshouse:
lets just each write it how we like it :)

jjgao:
@adamabeshouse @ersinciftci I think it's good to have some consistency within the team. Maybe this discussion can lead to our convention?

alisman:
My two cents on this issue, is, on the describe messages, I do think we should decide upon some standard because it relates to how we group things.  On the issue of the it() message, I think as long as it follows the grammatical construct "it does something" then we're good.  I don't think the repetition of "should" in the message makes much sense as it doesn't seem to carry any meaning.

ersinciftci:
:+1: I just thought `should` was the generally accepted convention at first, but then realized that it was actually not, and I agree that if it's not a convention, then why write the extra word every time. Hence removed `should`s from my tests in https://github.com/cBioPortal/cbioportal-frontend/pull/845/files

PR #838, opened 2017-12-21
Add tests of oncoprint data creation

adamabeshouse:
These are tests of the functionality that turns web API data into oncoprint data, as well as some code reorganization to facilitate testing.

Besides this, I think the rest of oncoprint-related code should be tested via e2e

### DIFF
--- a/src/shared/components/oncoprint/DataUtils.spec.ts
+++ b/src/shared/components/oncoprint/DataUtils.spec.ts
@@ -1,0 +1,785 @@
+import { assert } from 'chai';
+import {fillClinicalTrackDatum, fillGeneticTrackDatum, fillHeatmapTrackDatum, selectDisplayValue} from "./DataUtils";
+import {GeneticTrackDatum} from "shared/components/oncoprint/Oncoprint";
+import {AlterationTypeConstants, AnnotatedExtendedAlteration} from "../../../pages/resultsView/ResultsViewPageStore";
+import {ClinicalAttribute, GeneMolecularData, Sample} from "../../api/generated/CBioPortalAPI";
+import {SpecialAttribute} from "../../cache/ClinicalDataCache";
+import {OncoprintClinicalAttribute} from "./ResultsViewOncoprint";
+import {MutationSpectrum} from "../../api/generated/CBioPortalAPIInternal";
+
+describe("DataUtils", ()=>{
+   describe("selectDisplayValue", ()=>{
+       it("returns undefined if no values", ()=>{
+           assert.equal(selectDisplayValue({}, {}), undefined);
+       });
+       it("returns the lone value if one value", ()=>{
+           assert.equal(selectDisplayValue({"a":0}, {"a":0}), "a");
+       });
+       it("returns the lowest priority value if two values", ()=>{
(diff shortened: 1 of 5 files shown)


src/shared/components/oncoprint/DataUtils.spec.ts
```diff
@@ -0,0 +1,785 @@
...
+           assert.deepEqual(
+               fillGeneticTrackDatum({}, "gene", data),
+               {
+                   gene:"gene",
+                   data: data,
+                   disp_cna: undefined,
+                   disp_mrna: undefined,
+                   disp_prot: undefined,
+                   disp_mut: "missense_rec"
+               } as any, "test1");
```
ersinciftci: Does `test1` correspond to the message in here http://chaijs.com/api/assert/? If so, it seems it's optional, so you don't need to send something like `test1`.

adamabeshouse: I added it because when the tests were failing, the diff printing wasn't working properly so I couldn't tell which assertion within the `it` was failing.

ersinciftci: I thought yesterday we agreed on having only one case in each `it`, meaning whenever we call the tested method (`fillGeneticTrackDatum` in this case), we should have a separate `it` for it. I can see that it feels convenient to group some cases under one `it`, but with that logic one can create just one `it("should work correctly")` in every `describe`, so `it` would lose it's meaning. So as an example `it("should fill a datum for one putative missense mutation data correctly"` and `it("should fill a datum for one non-putative in_frame_del mutation data correctly"` looks better to me, and you wouldn't need the `test1` trick anymore.

adamabeshouse: that seems unnecessary to me

ersinciftci: What is a broad case? Isn't every case is a separate case?

adamabeshouse: if you read it you'll see, one case is "builds a datum correctly w a single mutation data", so it goes through a few ways for a single mutation data to happen.

ersinciftci: Yesterday you said you would split them up in the other PR, and it's the same situation here, so I thought we agreed on it. But not a big deal.

adamabeshouse: I think its a good point in some cases but here unnecessary imo

alisman: I think there are a couple decent reasons to group multiple assertions in same it.  One, I don't like an endless list of test results, so it's nice to be able to nest highly related assertions inside of one it (with use of assert message to distinguish failure, if dev isn't lazy ; ).  Two, if you need to run multiple assertions on the same returned item, you don't want to have to use closure to accomplish that.   An argument for one it one assert would be if the thing under test is stateful and dev should be forced to re-instantiate the thing for every test (using the beforeEach callback).   So, there are times when you definately should obey that rule.

ersinciftci: Good points 👍 Yeah as long as the assert message is meaningful, sounds good to me.

PR #843, opened 2017-12-21
enable tslint @ codeclimate

inodb:
Unfortunately codeclimate checking only works against one branch


jjgao:
@inodb should we try to merge this one?

PR #845, opened 2017-12-22
Add Survival Tab data button, fix style, add unit tests

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/3618 and https://github.com/cBioPortal/cbioportal/issues/3629


src/pages/resultsView/survival/SurvivalChart.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+            assert.deepEqual(getStats([], []), [0, 0, "NA"]);
+        });
+
+        it("returns correct stats for the example data", () => {
+            assert.deepEqual(getStats(examplePatientSurvivals, exampleEstimates), [6, 3, "182.19"]);
+        });
+    });
+
+    describe("#calculateLogRank()", () => {
+        it("returns NaN for empty list", () => {
```
alisman: this seems like a good example for discussion.  there is some complicated code inside this function that would require many more assertions to cover.   we need to figure out how to define "covered".

src/pages/resultsView/survival/SurvivalChart.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+                {x: 0.09, y: 80},
+                {x: 0, y: 80},
+                {x: 63.83, y: 53.3333333333333333},
+                {x: 0.13, y: 53.3333333333333333},
+                {x: 182.19, y: 0}
+            ]);
+        });
+    });
+
+    describe("#getScatterData()", () => {
```
alisman: this is a function which arguably has no logic inside it.  (in function itself we should use _.map instead of foreach/push).  i would be comfortable not testing this function

src/pages/resultsView/survival/SurvivalChart.spec.tsx
```diff
@@ -0,0 +1,184 @@
...
+];
+
+const examplePatientSurvivals = exampleAlteredPatientSurvivals.concat(exampleUnalteredPatientSurvivals);
+
+const exampleAlteredEstimates = [1,0.8];
+const exampleUnalteredEstimates = [0.8,0.5333333333333333,0.5333333333333333,0];
+const exampleEstimates = exampleAlteredEstimates.concat(exampleUnalteredEstimates);
+
+describe("SurvivalChart", () => {
+    describe("#getEstimates()", () => {
```
alisman: doesn't test the survivalStatus fork

ersinciftci: the example data has both status=true and status=false patients, so it enters both forks.

PR #853, opened 2018-01-03
Support prev/next button for multiple studies

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3639

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -81,17 +81,10 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                     }
 
                     let patientIdsInCohort = ('navCaseIds' in query ? (query.navCaseIds as string).split(",") : []);
-                    
-                    let multiStudyCohorts: string[] = patientIdsInCohort
-                    .filter(entity=> entity.split(':').length == 2);
 
-                    if(multiStudyCohorts.length > 0) {
-                        patientViewPageStore.patientIdsInCohort = multiStudyCohorts;
-                    } else {
-                        patientViewPageStore.patientIdsInCohort = patientIdsInCohort.map(entityId=>{
-                            return patientViewPageStore.studyId + ':' + entityId;
-                        });
-                    }
+                    patientViewPageStore.patientIdsInCohort = patientIdsInCohort.map(entityId=>{
+                        return entityId.includes(':') ? entityId : patientViewPageStore.studyId + ':' + entityId;
+                    });
                 } else {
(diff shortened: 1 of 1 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -79,8 +79,19 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                     {
                         patientViewPageStore.setSampleId(query.sampleId as string);
                     }
-                    patientViewPageStore.patientIdsInCohort = ('navCaseIds' in query ? (query.navCaseIds as string).split(",") : []);
 
+                    let patientIdsInCohort = ('navCaseIds' in query ? (query.navCaseIds as string).split(",") : []);
+                    
+                    let multiStudyCohorts: string[] = patientIdsInCohort
+                    .filter(entity=> entity.split(':').length == 2);
```
alisman: why not just do entity.contains(":").  no need to split since you're not saving the result

alisman:
Hi guys, 
Adding @ersinciftc and @jji here.  I'm ok with this being a temporary fix but it seems a slippery slope to me where if we don't start using uniqueId for things, our code will get polluted with this kind of thing.  @jj not that heatmap and pathology report just accept patientId, so seems like they could be susceptible to duplicate ID issue.   
To sum up, maybe we should just look at what it would take to start using a uniqueId (handled by backend)

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -79,8 +79,19 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+
+                    if(multiStudyCohorts.length > 0) {
+                        patientViewPageStore.patientIdsInCohort = multiStudyCohorts;
+                    } else {
+                        patientViewPageStore.patientIdsInCohort = patientIdsInCohort.map(entityId=>{
```
alisman: why not just map patientIdsInCohort and where we don't detect a compound, make it compound (add studyId)?  Seems simpler to me.

src/pages/patientView/PatientViewPage.tsx
alisman: actually, see comment below. maybe not necessary to determine if we have multiStudyCohort?

jjgao:
@alisman Agreed on support uniqueId in the long run.

For the moment, I am ok with this solution. In fact, `study_id:sample_id` is now pretty standard throughout the interface -- in query page, download tab, and study view page. 

For heatmap and path report, since they are TCGA samples, it's ok just use the sample ID to pull.

jjgao:
It'd be great if we can release this soon.

PR #854, opened 2018-01-03
Fixed broken patient view page table download

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3563.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
+++ b/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
@@ -131,17 +131,9 @@ export default class AlleleFreqColumnFormatter {
             map[next.sampleId] = next;
             return map;
         }, {} as {[s:string]:Mutation});
-        return sampleManager.getSampleIdsInOrder().map(sampleId=>sampleToMutation[sampleId]).map(mutation=>{
-            if (!mutation) {
-                return null;
-            }
-            const altReads = mutation.tumorAltCount;
-            const refReads = mutation.tumorRefCount;
-            if ((altReads < 0) || (refReads < 0)) {
-                return null;
-            }
-            return (altReads / (altReads + refReads));
-        });
+        return sampleManager.getSampleIdsInOrder().map(sampleId=>sampleToMutation[sampleId]).map(mutation=>
+            AlleleFreqColumnFormatter.calcFrequency(mutation)
+        );
     }
 
     public static isVisible(sampleManager:SampleManager|null, allMutations?: Mutation[][]): boolean {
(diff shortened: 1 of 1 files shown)


onursumer:
I think we definitely needed to add tests for this fairly complex logic of "breaking single rows into multiple rows when downloading"...

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -161,18 +153,37 @@ export default class AlleleFreqColumnFormatter {
 
 
     public static getFrequency(data:Mutation[]): string|string[] {
-        let result = [];
+        const result: string[] = [];
+
         if (data) {
-            for (let mutation of data) {
-                let frequency = mutation.tumorAltCount/(mutation.tumorAltCount+mutation.tumorRefCount);
-                if (frequency) {
```
onursumer: This was filtering out zero values like `0.0`, that's why we were ending up with an smaller array size and getting the error message `Not all the arrays returned from the download functions are from the same length.`

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -161,18 +153,37 @@ export default class AlleleFreqColumnFormatter {
 
 
     public static getFrequency(data:Mutation[]): string|string[] {
```
onursumer: Should we name this function `download` for consistency?

oplantalech: No objections, go ahead to change the name if it is more consistent with the code.

oplantalech:
@onursumer I've tested your code with the sample I put as example in my PR ([link here](http://www.cbioportal.org/beta/case.do?cancer_study_id=lgg_ucsf_2014&case_id=P04#/patient?caseId=P04&studyId=lgg_ucsf_2014)) and some allelic frequencies now appear empty:
captura de pantalla 2018-01-08 a les 16 08 03

onursumer:
@oplantalech an empty value indicates that there is no allele frequency data for that specific sample. For example, the first row with empty allele frequency value in your example (the one starting with the protein change `N49Tfs*23`) does not have any allele frequency value. See the screenshot below (taken from the example link you shared).

empty_allele_frequency

Current download logic generates a file with an allele frequency value of `1.05882352941176` for the same column though. Any idea where that value is coming from?

oplantalech:
@onursumer good you discovered this! It is a bug of course, I don't have any idea how this can happen, if I have time I'll look it up and post it here. Meanwhile, I approve the PR.

pieterlukasse:
@oplantalech did you find the bug?

PR #855, opened 2018-01-04
when study selector is filtered, unselect should clear ALL selected s…

alisman:
Formerly, unselect cleared only currently visible selections (in study list).  Instead we want to clear ALL selection so that the filtered selection can then be selected

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/shared/components/query/CancerStudySelector.tsx
+++ b/src/shared/components/query/CancerStudySelector.tsx
@@ -8,9 +8,9 @@ import classNames from 'classnames';
 import ReactSelect from 'react-select';
 import StudyList from "./studyList/StudyList";
 import {observer, Observer} from "mobx-react";
-import {expr} from 'mobx';
+import {action, expr} from 'mobx';
 import memoize from "memoize-weak-decorator";
-import {If, Then} from 'react-if';
+import {If, Then, Else} from 'react-if';
 import {QueryStoreComponent} from "./QueryStore";
 import SectionHeader from "../sectionHeader/SectionHeader";
 import {Modal, Button} from 'react-bootstrap';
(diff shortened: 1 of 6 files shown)


alisman:
Decided to test this using e2e tests.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -8,9 +8,9 @@ import classNames from 'classnames';
 import ReactSelect from 'react-select';
```
alisman: had to reformat this file heavily b/c andy was using very different formatting settings.  so diff includes a lot of code which is reformatted and not refactored

adamabeshouse: could you point out which lines are new code?

src/shared/components/query/StudyListLogic.ts
```diff
@@ -227,6 +227,12 @@ export class FilteredCancerTreeView
 		}
 	}
 
+	@action clearAllSelection(): void
+	{
+        this.store.selectedStudyIds = []
```
adamabeshouse: while this doesn't bother me too much, I do feel like this would be something that might be more natural as a method of the store.

src/shared/components/query/CancerStudySelector.tsx
adamabeshouse: nevermind

src/shared/components/query/StudyListLogic.ts
alisman: i'm just kind of following Andy's separation of concerns, which is pretty different from what we've done.

PR #856, opened 2018-01-06
Fix survival tab when there are NaN months values

ersinciftci:


src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+import {PatientSurvival} from "../../shared/model/PatientSurvival";
+
+export function getPatientSurvivals(survivalClinicalDataGroupByUniquePatientKey: any, patients: Patient[], 
+    targetUniquePatientKeys: string[], statusAttributeId: string, monthsAttributeId: string, 
+    statusFilter: (s: string) => boolean): PatientSurvival[] {
+
+    let patientSurvivals: PatientSurvival[] = [];
+    if (targetUniquePatientKeys) {
+        targetUniquePatientKeys.forEach((uniquePatientKey: string) => {
+            const clinicalData: ClinicalData[] = survivalClinicalDataGroupByUniquePatientKey![uniquePatientKey];
```
alisman: shouldn't need the !

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -0,0 +1,29 @@
+import {Patient, ClinicalData} from "shared/api/generated/CBioPortalAPI";
+import {PatientSurvival} from "../../shared/model/PatientSurvival";
+
+export function getPatientSurvivals(survivalClinicalDataGroupByUniquePatientKey: any, patients: Patient[], 
+    targetUniquePatientKeys: string[], statusAttributeId: string, monthsAttributeId: string, 
+    statusFilter: (s: string) => boolean): PatientSurvival[] {
+
+    let patientSurvivals: PatientSurvival[] = [];
+    if (targetUniquePatientKeys) {
+        targetUniquePatientKeys.forEach((uniquePatientKey: string) => {
```
alisman: minor quibble but i like to use reduce in a case like this because it make it more clear that the the goal of the enclosed code is to produce the return value

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -0,0 +1,29 @@
...
+            const clinicalData: ClinicalData[] = survivalClinicalDataGroupByUniquePatientKey![uniquePatientKey];
+            if (clinicalData) {
+                const statusClinicalData: ClinicalData = clinicalData.filter(c => c.clinicalAttributeId === statusAttributeId)[0];
```
alisman: you could use [].find instead of [].filter here.  It will return first found item or undefined. Will stop iteration when it finds item and also doesn't require index.   You should double check though that it works in IE11 via babel-polyfill.

src/pages/resultsView/SurvivalStoreHelper.ts
ersinciftci: Isn't `reduce` used in order to reduce an array to a single value? In here I create a new array by transforming the valid elements of the original array. So it's more like a combination of `filter` and `map`, I can convert to them if you think it would be more readable.

alisman: i think new array with potentially fewer elements qualifies semantically as a "reduction" of original.  you're right _filterMap would be better but I don't think it exists.  the key point to me is that it enapsulates the transformation and assigns the result to a value.  whereas, forEach, you have to actually read the forEach to understand that what you're doing is pushing values to a variable outside the foreach.

ersinciftci: I see, done.

PR #858, opened 2018-01-08
Change webservice to validate gene sets

oplantalech:
In cbioportal/cbioportal#3281 I implemented a new webservice, which is used here to improve the already implemented GSVA frontend: the code is simplified and the validation is faster.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -600,16 +600,8 @@ export class QueryStore
 	        async (genesetIds:string[]):Promise<{found: Geneset[], invalid: string[]}> =>
 	        {
 	            const getGenesetResults = async () => {
-	                const found:Geneset[] = [];
-	                const invalid: string[] = [];
-	                for (const genesetId in genesetIds) {
-	                    try {
-	                        const retrievedGeneset = await internalClient.getGenesetUsingGET({genesetId: genesetIds[genesetId]});
-	                        found.push(retrievedGeneset);
-	                    } catch (error) {
-	                        invalid.push(genesetIds[genesetId]);
-	                    }
-	                }
+	                const found = await internalClient.fetchGenesetsUsingPOST({genesetIds: genesetIds});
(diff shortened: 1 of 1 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -572,16 +572,8 @@ export class QueryStore
...
-	                for (const genesetId in genesetIds) {
-	                    try {
-	                        const retrievedGeneset = await internalClient.getGenesetUsingGET({genesetId: genesetIds[genesetId]});
-	                        found.push(retrievedGeneset);
-	                    } catch (error) {
-	                        invalid.push(genesetIds[genesetId]);
-	                    }
-	                }
+	                const found = await internalClient.fetchGenesetsUsingPOST({genesetIds: genesetIds});
+	                const invalid = _.difference(genesetIds, found.map(geneset => geneset.genesetId + ''));
```
alisman: + '' is to make it a string?   if so, you could just call toString()

src/shared/components/query/QueryStore.ts
```diff
@@ -572,16 +572,8 @@ export class QueryStore
...
-	                for (const genesetId in genesetIds) {
-	                    try {
-	                        const retrievedGeneset = await internalClient.getGenesetUsingGET({genesetId: genesetIds[genesetId]});
-	                        found.push(retrievedGeneset);
-	                    } catch (error) {
-	                        invalid.push(genesetIds[genesetId]);
-	                    }
-	                }
+	                const found = await internalClient.fetchGenesetsUsingPOST({genesetIds: genesetIds});
+	                const invalid = _.difference(genesetIds, found.map(geneset => geneset.genesetId));
```
oplantalech: @alisman Turns out `geneset.genesetId` is already a string, so I removed the `+ ''` part.

src/shared/components/query/QueryStore.ts
oplantalech: @alisman Turns out `geneset.genesetId` is already a string, so I removed the `+ ''` part.

PR #859, opened 2018-01-08
Update cBioPortal and OncoKB API

inodb:
Both cBioPortal and OncoKB API were out of sync
  
Include a hack for dealing with svType field that should actually be
optional.

### DIFF
--- a/my-index.ejs
+++ b/my-index.ejs
@@ -11,7 +11,7 @@
             // uncomment to enable Darwin internal MSKCC service patient view link
             // uses checkDarwinAccess.do service from cbioportal/cbioportal
             // enableDarwin : true,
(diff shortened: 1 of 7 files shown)


my-index.ejs
```diff
@@ -11,7 +11,7 @@
             // uncomment to enable Darwin internal MSKCC service patient view link
             // uses checkDarwinAccess.do service from cbioportal/cbioportal
             // enableDarwin : true,
-            apiRoot: 'www.cbioportal.org/beta',
```
alisman: do we want to change this?

inodb: @alisman yeah i think so. There's a new endpoint on `/beta` that's not in production

PR #862, opened 2018-01-08
dont include clinical track description in tooltip if same as name

adamabeshouse:
Also update oncoprintjs version, which fixes tooltip line break issue shown in the screenshot in https://github.com/cBioPortal/cbioportal/issues/3596

### DIFF
--- a/src/shared/components/oncoprint/DeltaUtils.ts
+++ b/src/shared/components/oncoprint/DeltaUtils.ts
@@ -448,7 +448,7 @@ function transitionClinicalTrack(
             data: nextSpec.data,
             data_id_key: "uid",
             label: nextSpec.label,
-            description: nextSpec.description,
+            description: ((nextSpec.label || "").trim() === (nextSpec.description || "").trim()) ? undefined : nextSpec.description,
             removable: true,
             removeCallback: ()=>{
                 delete getTrackSpecKeyToTrackId()[nextSpec.key];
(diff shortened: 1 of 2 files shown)


src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -448,7 +448,7 @@ function transitionClinicalTrack(
             data: nextSpec.data,
             data_id_key: "uid",
             label: nextSpec.label,
-            description: nextSpec.description,
+            description: (nextSpec.label.trim() === nextSpec.description.trim()) ? "" : nextSpec.description,
```
alisman: are we 100% sure these props are both non-null?

adamabeshouse: good point

adamabeshouse: yes we are :)

alisman: the reason i ask is that we know client generator isn't accurate in reflecting the true "nullability" of fields and "description" strikes me as something that we might know is sometimes missing.  would be worth checking

alisman: btw, i'm assuming description comes from api data though i can't see it.

adamabeshouse: good catch!

adamabeshouse: okay I updated it

PR #870, opened 2018-01-09
Custom driver annotations in refactored oncoprint

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1365,6 +1367,36 @@ export class ResultsViewPageStore {
...
+    readonly customDriverAnnotationReport = remoteData<{ hasBinary:boolean, tiers:string[] }>({
+        await:()=>[
+            this.mutations
+        ],
+        invoke:()=>{
+            let hasBinary = false;
+            let tiersMap:{[tier:string]:boolean} = {};
+            for (const mutation of this.mutations.result!) {
+                hasBinary = hasBinary || !!mutation.driverFilter;
+                mutation.driverTiersFilter && (tiersMap[mutation.driverTiersFilter] = true);
```
alisman: i don't think we should use this syntax for normal control logic.   we use it in jsx for a specic typescript reason.  lets use if () {}

adamabeshouse: one

adamabeshouse:
@oplantalech please see updated code, I've addressed your comment

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1368,30 +1372,30 @@ export class ResultsViewPageStore {
...
-                    annotated.push(Object.assign({
-                        putativeDriver,
-                        isHotspot: putativeDriverInfo.hotspots,
-                        oncoKbOncogenic: putativeDriverInfo.oncoKb,
-                        simplifiedMutationType: getSimplifiedMutationType(mutation.mutationType)
-                    }, mutation) as AnnotatedMutation);
-                }
-                return annotated;
-            }, []));
+            return Promise.resolve(computePutativeDriverAnnotatedMutations(this.mutations.result!, this.getPutativeDriverInfo.result!, this.mutationAnnotationSettings.ignoreUnknown));
```
oplantalech: @adamabeshouse Typescript complains about `this.mutationAnnotationSettings.ignoreUnknown`: 

```
Argument of type 'boolean | undefined' is not assignable to parameter of type 'boolean'.
  Type 'undefined' is not assignable to type 'boolean'.
```

adamabeshouse: good catch, simple fix

src/pages/resultsView/ResultsViewPageStoreUtils.spec.ts
```diff
@@ -0,0 +1,425 @@
...
+            bothMutation = {
+                driverFilter:"ADFADF",
+                driverTiersFilter:"SDPOIFJP"
+            } as Mutation;
+
+            neitherMutation = {
+            } as Mutation;
+        });
+
+        it("returns the right report for empty list", ()=>{
```
alisman: for discussion, but  i would argue only two tests are necessary for this function.  1. mutation list with element sans driverFilter and 2. mutation list where they all have driver filter.   it seems to me that would:
1. cover all logical branches. 
2. cover the format of the return value

I don't think we need to test empty array, because that's just testing that the for loop will properly handle a zero length array.  I.e. testing the javascript engine.  
@ersinciftci please comment

adamabeshouse: I agree with @ersinciftci , thats how I was taught at school too - treat it as black box and cover "representative cases" to try to cover all functionality, including edge cases like empty and singleton.

PR #872, opened 2018-01-09
fill sample ids, bug in tests though

adamabeshouse:


src/shared/lib/onMobxPromise.spec.ts
```diff
@@ -72,4 +72,32 @@ describe('onMobxPromise', ()=>{
             });
         });
     });
+    it.only('executes immediately if the promise is already resolved, does not execute again', (done)=>{
```
alisman: only.      also, just curious, toes this have to do with pr?

src/shared/components/query/QueryStore.ts
```diff
@@ -1278,8 +1278,18 @@ export class QueryStore
 			const caseSetId =  (window as any).serverVars.caseSetProperties.case_set_id;
 			if (caseSetId !== undefined) {
 				this.selectedSampleListId = caseSetId;
+				this.initiallySelected.sampleListId = true;
```
alisman: how come we need to remember initially selected distinct from selectedSampleListId?  could use a comment

PR #874, opened 2018-01-10
Fix sample labels disappearing on timeline

inodb:
Add d3 as dependency to legacy timeline code. It was throwing an
undefined error after upgrading to webpack 4 in
https://github.com/cBioPortal/cbioportal-frontend/commit/b5621cf8a4dff08d3d9aef1f3aa558857fdb294b.
Surprising that this code worked originally, since it requires d3.


inodb:
Need to merge this before we can merge master back to release-1.11.0: https://github.com/cBioPortal/cbioportal/pull/3687

PR #876, opened 2018-01-11
Fix linux removing fsevents (osx) package-lock

inodb:
Upgrade to npm 5.6.0 seems to solve the issue:

https://github.com/npm/npm/issues/17722

This will allow osx users to update package lock file without having to get it form circle-ci

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,8 +38,8 @@
     "validate": "npm ls"
   },
   "engines": {
-    "npm": "5.5.1",
-    "node": "8.9.1"
+    "npm": "5.6.0",
+    "node": "8.9.4"
   },
   "author": "",
   "license": "",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -38,8 +38,8 @@
     "validate": "npm ls"
   },
   "engines": {
-    "npm": "5.5.1",
-    "node": "8.9.1"
+    "npm": "5.6.0",
+    "node": "8.9.4"
```
inodb: this is for heroku

pom.xml
```diff
@@ -24,8 +24,8 @@
                         </goals>
                         <configuration>
                             <!-- See https://nodejs.org/en/download/ for latest node and npm (lts) versions -->
-                            <nodeVersion>v8.9.1</nodeVersion>
-                            <npmVersion>5.5.1</npmVersion>
+                            <nodeVersion>v8.9.4</nodeVersion>
+                            <npmVersion>5.6.0</npmVersion>
```
inodb: this is for jitpack

PR #877, opened 2018-01-11
Optimize oncoprint

adamabeshouse:
Fix some bugs affecting oncoprint refresh optimization.

TODO: needs testing

### DIFF
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -836,7 +836,7 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                     divId={this.props.divId}
                     width={1050}
                     suppressRendering={this.clinicalTracks.isPending || this.geneticTracks.isPending || this.heatmapTracks.isPending}
-                    hiddenIds={!this.showUnalteredColumns ? this.unalteredKeys.result : []}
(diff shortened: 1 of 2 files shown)


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -836,7 +838,7 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                     divId={this.props.divId}
                     width={1050}
                     suppressRendering={this.clinicalTracks.isPending || this.geneticTracks.isPending || this.heatmapTracks.isPending}
-                    hiddenIds={!this.showUnalteredColumns ? this.unalteredKeys.result : []}
+                    hiddenIds={!this.showUnalteredColumns ? this.unalteredKeys.result : (EMPTY_ARRAY as string[])} // reuse empty array constant to be compatible w shallow equality check
```
alisman: this doesn't seem like generally a great practice.  if for some reason something is pushed into that array in the future, now you're updating multiple owners of that reference inadvertantly.  why not just add zero length array check in condition?

alisman: or actually, just make it null?

adamabeshouse: went with undefined

PR #878, opened 2018-01-11
Fix survival chart tooltip positioning and download button function

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/resultsView/survival/SurvivalChart.tsx
+++ b/src/pages/resultsView/survival/SurvivalChart.tsx
@@ -137,13 +137,11 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps, 
 
         return (
 
-            <div>
+            <div className="posRelative">
 
+                <div className="borderedChart">
 
-
-                <div className="borderedChart" style={{position:'relative'}}>
-
-                    <div className="btn-group" style={{position:'absolute', right:10 }} role="group">
+                    <div className="btn-group" style={{position:'absolute', zIndex:10, right:10 }} role="group">
                         <button className={`btn btn-default btn-xs`} onClick={this.downloadSvg}>
                             SVG <i className="fa fa-cloud-download" />
                         </button>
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -183,6 +181,7 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
                 </div>
                 {this.tooltipModel &&
                     <Popover arrowOffsetTop={48} positionLeft={this.tooltipModel.x + 15}
+                             { ...{container:this} }
```
alisman: basically this causes popover to be placed inside the top level div of this component instead of body.  Since mouseover event seems to give offset relative to parent div, this is necessary.  (Problem was caused by adding header and padding to chart.)
The reason I'm using this funky { ...{ }} syntax is that React-Bootstrap crappy typings don't include this container property.

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -137,13 +137,11 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
-            <div>
+            <div className="posRelative">
 
+                <div className="borderedChart">
 
-
-                <div className="borderedChart" style={{position:'relative'}}>
-
-                    <div className="btn-group" style={{position:'absolute', right:10 }} role="group">
+                    <div className="btn-group" style={{position:'absolute', zIndex:10, right:10 }} role="group">
```
alisman: it would have been nice if we had a test to detect this failure.  because it's css issue, I think the only possiblity would e2e test.  Even then it would be hard because the test would have had to use positioning to click rather than $(el).click in order to be subject to the zindex issue.

ersinciftci:
@alisman in the previous PR you set titles to empty string https://github.com/cBioPortal/cbioportal-frontend/pull/867/files#diff-f9e8d59ac1083e4f061c185936ada406R65, but actually they are also used as the file name when you download svg,png and data. So you can keep the titles and also use `this.overallSurvivalTitleText` and `this.diseaseFreeSurvivalTitleText` in `SurvivalTab.tsx` as the new chart titles, so would prevent duplication.

PR #880, opened 2018-01-11
Put genetic track at top of oncoprint legend

adamabeshouse:
TODO: E2E testing

@inodb @alisman I'm changing package.json, do I need to do anything else with package-lock?

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11982,9 +11982,9 @@
       }
     },
     "oncoprintjs": {
-      "version": "1.0.74",
-      "resolved": "https://registry.npmjs.org/oncoprintjs/-/oncoprintjs-1.0.74.tgz",
-      "integrity": "sha1-26SlGEivOCrASK6Fbqdx5n2l7fM=",
+      "version": "1.0.75",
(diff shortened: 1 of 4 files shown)


alisman:
By running npm install, a new package.json should be generated automatically.  You can then commit this package.json.    THere was an issue where on mac, a package called fs would be added, which would then break on linux.  I think @inodb fixed that.

src/shared/components/query/QueryStore.ts
```diff
@@ -1467,7 +1467,7 @@ export class QueryStore
 		if (historyUrl != newUrl)
 			window.history.pushState(null, window.document.title, historyUrl);*/
 
-		formSubmit(urlParams.pathname, urlParams.query);
+		formSubmit(urlParams.pathname, urlParams.query, undefined, "smart");
```
alisman: seems like these changes are from other pull request, no?

adamabeshouse: done

adamabeshouse:
@alisman having a lot of issues with the screenshot tests, but the effort is ongoing, and in a separate PR https://github.com/cBioPortal/cbioportal/pull/3712 - it safe to merge this, we'll eventually get those screenshots in

PR #881, opened 2018-01-11
Added a study config parameter for germline consented samples

onursumer:
- Fix https://github.com/cBioPortal/cbioportal/issues/3471
- Now we need to make sure that we provide the config parameter for the private instances.

# Checks
  the `-s` option)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1126,16 +1126,35 @@ export class ResultsViewPageStore {
...
-            const ids: string[][] = await Promise.all(studies.map(studyId => {
-                return client.getAllSampleIdsInSampleListUsingGET({
-                    sampleListId: this.getGermlineSampleListId(studyId)
-                });
-            }));
-            return _.flatten(ids.map((sampleIds: string[], index: number) => {
-                const studyId = studies[index];
-                return sampleIds.map(sampleId => ({sampleId, studyId}));
-            }));
+            const studiesWithGermlineConsentedSamples = AppConfig.studiesWithGermlineConsentedSamples;
```
alisman: we gotta write tests for this.  : (  
shouldn't be too hard if you mock client.  you can do this easily by making a method on store that return client.  then use sinon stub to return a function that returns promise with fake data

onursumer: Added a new function `fetchGermlineConsentedSamples` into `StoreUtils`, will add a test for that soon.

PR #901, opened 2018-01-17
Use OncoKB data for hotspot annotation

adamabeshouse:
Also implement function that will later be used to handle oncoKB timeout

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -60,6 +60,7 @@ import {countMutations, mutationCountByPositionKey} from "./mutationCountHelpers
 import {getPatientSurvivals} from "./SurvivalStoreHelper";
 import {QueryStore} from "shared/components/query/QueryStore";
 import {
+    annotateMolecularDatum, getOncoKbOncogenic,
     computeCustomDriverAnnotationReport, computePutativeDriverAnnotatedMutations,
     initializeCustomDriverAnnotationSettings
 } from "./ResultsViewPageStoreUtils";
@@ -280,8 +281,6 @@ export function extendSamplesWithCancerType(samples:Sample[], clinicalDataForSam
     return extendedSamples;
 
 }
-
-export const ONCOKB_ONCOGENIC_LOWERCASE = ["likely oncogenic", "predicted oncogenic", "oncogenic"];
 
 export class ResultsViewPageStore {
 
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1489,18 +1504,51 @@ export class ResultsViewPageStore {
...
+                return addTimeoutToPromise(
+                    this.hotspotDataPromise.result!.promise,
+                    HOTSPOTS_TIMEOUT_MS
+                );
+            } else {
+                return Promise.resolve(undefined);
+            }
+        },
+        onError:()=>{
+            alert("Hotspot data not available, or timeout encountered. Please try again in a few moments.");
```
onursumer: Should we alert inside a UI component instead?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -282,6 +283,7 @@ export function extendSamplesWithCancerType(samples:Sample[], clinicalDataForSam
 }
 
 export const ONCOKB_ONCOGENIC_LOWERCASE = ["likely oncogenic", "predicted oncogenic", "oncogenic"];
+export const HOTSPOTS_TIMEOUT_MS = 5000;
```
adamabeshouse: @jjgao @alisman @onursumer is this a good timeout? for oncoprint only, no timeout with the mutation table

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: I feel like its okay to trigger the alert here as opposed to in the UI, it really should be triggered by the error event. @alisman what do you think?

alisman: I don't think we should be using alert.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1491,17 +1491,12 @@ export class ResultsViewPageStore {
...
         ],
         invoke:()=>{
             return Promise.resolve((mutation:Mutation)=>{
-                const indexedHotspotData = this.indexedHotspotData;
-                if (!indexedHotspotData.isComplete) {
-                    return false;
-                } else {
-                    return isHotspot(mutation, indexedHotspotData.result!.single);
-                }
+                const oncokbAnnotation = this.getOncoKbMutationAnnotationForOncoprint.result!(mutation);
```
alisman: so the result is a function?

adamabeshouse: this is a function that should be tracked as a mobx promise, ie not available before its dependencies are. If you have a better way to handle it, I'm open to it

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1430,9 +1430,9 @@ export class ResultsViewPageStore {
                     this.getOncoKbMutationAnnotationForOncoprint.isComplete &&
                     this.getOncoKbMutationAnnotationForOncoprint.result(mutation);
 
-                let oncoKb:string = ((oncoKbDatum && oncoKbDatum.oncogenic) || "").toLowerCase();
-                if (ONCOKB_ONCOGENIC_LOWERCASE.indexOf(oncoKb) === -1) {
-                    oncoKb = "";
+                let oncoKb:string = "";
+                if (oncoKbDatum && ONCOKB_ONCOGENIC_LOWERCASE.indexOf((oncoKbDatum.oncogenic || "").toLowerCase()) > -1) {
```
alisman: nbd, but you can now use string.includes method instead of indexOf.

src/pages/resultsView/ResultsViewPageStore.ts
alisman: we've talked bout this before.  i would rather see dependencies "on the surface" and passed to pure functions because i think it's more flexible and explicit.  this caught my eye just because i hadn't seen a result be a function.  anyway, just an academic discussion.

adamabeshouse: that would be nicer actually, I agree. lets handle that later?

PR #902, opened 2018-01-18
Fix addGenesAndSubmit, test broken though

adamabeshouse:


inodb:
@adamabeshouse Thanks for the help! I added an end to end test for the enrichment tab. Works well. Not sure how to stub a computed get from mobx using Sinon. @alisman ?

src/shared/components/query/QueryStore.ts
```diff
@@ -275,6 +275,11 @@ export class QueryStore
 	@observable searchText:string = '';
 
 	@observable private _selectedStudyIds:ObservableMap<boolean> = observable.map<boolean>();
+
+	@computed get allSelectedStudyIds():string[] {
```
alisman: Is there a reason we couldn't use the selectedStudyIds below?  Do we understand what this selectable studies thing is about?   We might take this opportunity to clean this up, or add some comments explaining why we need both of these.

src/shared/components/query/QueryStoreUtils.ts
```diff
@@ -65,8 +66,8 @@ export function nonMolecularProfileParams(store:QueryStore):NonMolecularProfileQ
         Action: 'Submit',
     };
 
-    if (store.selectedStudyIds.length !== 1) {
-        ret = Object.assign(ret, { cancer_study_list: store.selectedStudyIds.join(",") });
+    if (selectedStudyIds.length !== 1) {
+        ret = Object.assign(ret, { cancer_study_list: selectedStudyIds.join(",") });
```
alisman: why are we using assign instead of just ret.cancer_study_list = ...   ?

end-to-end-tests/specs/home.spec.js
```diff
@@ -259,4 +259,36 @@ describe('single study query', function() {
...
+        it('should show mutations plot', function() {
+            browser.url(`${CBIOPORTAL_URL}/index.do?cancer_study_id=ov_tcga_pub&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=ov_tcga_pub_cna_seq&gene_list=BRCA1+BRCA2&geneset_list=+&tab_index=tab_visualize&Action=Submit&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ov_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=ov_tcga_pub_gistic`);
+
+            // click enrichments tab
+            $('#enrichments-result-tab').waitForExist(30000);
+            $('#enrichments-result-tab').click();
+
+            // wait for plot to show
+            $('#ov_tcga_pub_mutations_plot_div').waitForExist(60000);
+            assert(true);
```
alisman: i wonder if we even needs this assert?  would be good to know.

src/shared/components/query/QueryStore.ts
inodb: The `selectedStudyIds` uses `selectableStudiesSet` which isn't always available. It could be changed into a readonly and do an await, but would require more rewriting. Since a lot of the properties would then have to go from `@computed` to `readonly`. Not sure how much effort that would be

adamabeshouse: to be honest, I don't actually remember why I did things that way. I don't think its a good approach. A better system would, I think, be to just delete studies from the selected study ids when we load all the studies and can see which ones are not available. This filtering system doesn't really make sense, I think, and is causing all these conceptual and coding difficulties.

src/shared/components/query/QueryStoreUtils.ts
adamabeshouse: I think that may be legacy from when I was assigning more than one member here.

end-to-end-tests/specs/home.spec.js
inodb: @alisman yes we can remove it

src/shared/components/query/QueryStore.ts
inodb: @adamabeshouse @alisman I've created a new issue for this as discussed so we can at least include this fix for a critical bug: https://github.com/cBioPortal/cbioportal/issues/3771

PR #903, opened 2018-01-18
Expect serverVars.theQuery to be decoded

adamabeshouse:


alisman:
I think the only important part of this review is seeing if we can address the confusing nature of selectedStudyIds  in the store.   Maybe we can discuss live.

PR #913, opened 2018-01-23
Display germline mutation as white stripe in oncoprint view

rnugraha:
# What? Why?

- Display germline mutation as white stripe in the oncoprint track
- Add background to the oncoprint rule legends

# Dependency
https://github.com/cBioPortal/oncoprintjs/pull/55

# Checks
  the `-s` option)

# Any screenshots or GIFs?

## Before
schermafbeelding 2018-02-08 om 13 00 25

## After
after1

### In zoom-out
schermafbeelding 2018-02-08 om 11 26 37

# Notify reviewers
cBioPortal organization, notify the appropriate team (remove inappropriate):

@cBioPortal/frontend

and notify them here:


pieterlukasse:
perhaps remove all white-space changes in `src/shared/components/oncoprint/geneticrules.ts`...?

jjgao:
@rnugraha Can we still color based on mutation types (missense, truncating, inframe) for germline mutations?

inodb:
Hey thanks for submitting @rnugraha ! I don't think the white line can convey all the information. For instance: I can't see the difference between truncating germline mutation and a patient that has both a somatic truncating mutation and a germline mutatation. I think it would make more sense to make a different track for germline mutations. So in the example four tracks, two for somatic BRCA1/2 and two for germline BRCA 1/2 mutation (if there are any). That would also be useful in showing which patients had germline+somatic profiling. Thoughts  @adamabeshouse @schultzn ?

adamabeshouse:
just from a design perspective also, I find white is a little confusing because the other time we use white in the oncoprint is when whitespace means not sequenced, so it seems like it indicates the mutation is not as valid or important somehow (also because it makes the color of the mutation behind it quite difficult to spot)

jjgao:
cc'ing @schultzn on this discussion.

Sjoerd-van-Hagen:
@inodb Currently, we have one gene per track and using the OQL it can be split up into multiple. We now also hide mutations in the oncoprint if there is a mutation that 'trumps' it. I think it would be good to keep it consistent. Your suggestion is good, but would be for a different feature in my opinion: 'germline in OQL'.

pieterlukasse:
@jjgao the mutation type color will be there. It is missing in the screenshot because @rnugraha disabled it from the menu when making the screenshot. 

@inodb thanks for the ideas. We discussed with @schultzn in a call and I collected some of his feedback regarding the issue you raised. See last page of [RFC 40](https://docs.google.com/document/d/1DRmABEBAdtP8kxsp5yevKeCZSlIE4LmBu-2sPBmho44).

adamabeshouse:
@pieterlukasse how about a slightly different shape? I think that would better accomplish the goal of indicating that its still a mutation of the same type, but with an additional modifier. What about making it a diamond instead of a square? or a square with triangles on the top and bottom to make it some kind of elongated diamond.

Sjoerd-van-Hagen:
@adamabeshouse I think the problem with using a different shape is that it will not be visible when you zoom uit.

adamabeshouse:
@Sjoerd-van-Hagen that's a good point, that would be a reason to favor the elongated diamond instead of a similar-sized diamond - the elongation would still be visible zoomed out

Sjoerd-van-Hagen:
@adamabeshouse I am not sure what you mean or how it would remain visible when zooming out?

adamabeshouse:
@Sjoerd-van-Hagen heres a mockup:
germline-diamond

germline-diamond-zoomed

I actually agree its not that clear zoomed out. But I also don't think putting a shape in front of the square is good, because it seems to modify the graphic in a way that detracts from it, when the message should be that theres just an EXTRA piece of info about the mutation. 

Maybe something like this, a little cap on top of the mutation square?
In this image is also a possibility of having a cap on top and bottom of the square
image

Sjoerd-van-Hagen:
@adamabeshouse I would then go for a line at the top and the bottom of the square. It will still look good when zooming out and it is easier to implement.

schultzn:
We won't be able to show all details. I still think the white line looks best - that could be used to indicate the presence of a germline mutation in that gene in that patient. If there are multiple mutations in the same gene and sample, the tooltip can reveal that.

jjgao:
@rnugraha could you please make the white line thinner and add another screenshot here? The current white color makes the green faded away -- maybe that's part of the confusion.

Sjoerd-van-Hagen:
@jjgao Riza is out of the office this week. Perhaps @adamabeshouse  can do this, if it is not too much trouble?

jjgao:
@Sjoerd-van-Hagen we can wait for @rnugraha.

rnugraha:
Thanks all for the input and discussion. I would personally agree with the thinner white line. I shall  update soon the screenshots.

rnugraha:
PR's description and Screen shots are updated. Please beware that this PR has dependency on https://github.com/cBioPortal/oncoprintjs/pull/55.

pieterlukasse:
Looks great @rnugraha , thanks for the changes! :+1:

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -123,13 +133,19 @@ export function fillGeneticTrackDatum(
...
+
                 let oncoprintMutationType = getOncoprintMutationType(getSimplifiedMutationType(event.mutationType)!);
+
                 if (oncoprintMutationType === "fusion") {
                     dispFusion = true;
                 } else {
                     if (event.putativeDriver) {
                         oncoprintMutationType += "_rec";
                     }
+                    if (event.mutationStatus === 'Germline') {
```
adamabeshouse: why do we need this if you're already tracking germline status with `disp_germline`?

adamabeshouse:
@rnugraha I think there was a misunderstanding - changing disp_germline to isGermline doesn't make any difference, the issue is the redundancy between disp_germline and disp_mut. The solution I proposed was to replace every usage of disp_germline with a helper function, isGermline(d) which would check whether disp_mut contained the _germ suffix. In that case you'd have to add genetic rules for each of the disp_mut options.

rnugraha:
ok make sense

rnugraha:
@adamabeshouse Okay, so different approach. Now only using disp_germline instead of adding more suffix '_germ' to the existing mutation types. Sorting and priorities are managed by the germline comparation metrix.

adamabeshouse:
@rnugraha that sounds good, thanks for doing that.

The last thing I need to ask is that you implement/adjust end to end tests having to do with the sort order. In home.spec.js, there is a section on oncoprint sorting. I'm not sure whether those tests will be affected by this change. If not, could you just add a simple test of a query that includes germline mutations, and validate the sort order using the method you can find in those tests ( checking the value of `browser.execute(function() { frontendOnc.getIdOrder(); }).value`)

Sjoerd-van-Hagen:
@adamabeshouse could you re-review this PR and see if it is ready to be merged?

rnugraha:
@adamabeshouse Unit test & e2e have been provided. Could you pls have a look?

end-to-end-tests/specs/home.spec.js
```diff
@@ -508,6 +508,44 @@ describe('oncoprint', function() {
...
+                "patient id order correct"
+            );
+        });
+
```
adamabeshouse: looks good, please just add a check for sample order as well

adamabeshouse:
Looks good @rnugraha could you please squash to one commit?

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -128,13 +129,16 @@ export function fillGeneticTrackDatum(
...
+
                 let oncoprintMutationType = getOncoprintMutationType(getSimplifiedMutationType(event.mutationType)!);
+
                 if (oncoprintMutationType === "fusion") {
                     dispFusion = true;
                 } else {
                     if (event.putativeDriver) {
                         oncoprintMutationType += "_rec";
                     }
+                    dispGermline[oncoprintMutationType] = event.mutationStatus === 'Germline';
```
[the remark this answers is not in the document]
adamabeshouse: please add a unit test like "fills a datum w one germline one non- " or something that could catch this

rnugraha:
@adamabeshouse  https://github.com/cBioPortal/cbioportal-frontend/pull/913/files#diff-831da00dd128eebcf1b83908cb0fdd1cR84 should cover the test for germline flag. Let me know if you think it's not sufficient.

adamabeshouse:
@rnugraha thanks for making the fix. I think it's a good practice for us to make unit tests for any bugs we find, that's why I think it would be good to have a unit test where a datum is filled using two mutations - the first germline, the second not.

rnugraha:
@adamabeshouse okay, pls check again

adamabeshouse:
@rnugraha could you point me to the test, I don't see it? I see a test containing two separate tests w single mutations w germline. I think it would be good to have one test where a datum is filled with two mutations, one is germline and the *second* is not (thats the key), and the result should be that germline is true. This would catch the bug we're talking about

rnugraha:
@adamabeshouse https://github.com/cBioPortal/cbioportal-frontend/pull/913/files#diff-831da00dd128eebcf1b83908cb0fdd1cR223

adamabeshouse:
@rnugraha it would be great to have a test where fillGeneticTrackDatum is called with data.length = 2, where the first element of data is Germline and the second is not

rnugraha:
ah that what you mean, ok pls check now

adamabeshouse:
@rnugraha thanks! can you squash your commits and then Ill merge

rnugraha:
@adamabeshouse looks like all the failing tests have something to do with unable to reached certain amount of time to display something as well as the last failing test you mentioned. https://github.com/thehyve/cbioportal-frontend/blob/607915a7a919fa83f1bea77df57e382fdcb700f0/end-to-end-tests/specs/home.spec.js#L481

The e2e tests I created to test the germline seems to be fine.

rnugraha:
@adamabeshouse Test failing you mentioned is not caused by my changes. It's already happening in rc https://circleci.com/gh/cBioPortal/cbioportal-frontend/4458?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #915, opened 2018-01-23
Change attribute ids of special clinical attributes

adamabeshouse:


alisman:
@jjgao merging this would break some of the urls that might be out there with integer values for clinical attributes.  Do you think we should go through the trouble of maintaining compatibility with those (they have only been in place since yesterday)

jjgao:
@alisman not worries about existing ones.

jjgao:
@alisman for old urls, it's ok if a track is not displayed, but the oncoprint should still load.

alisman:
@adamabeshouse can you confirm that's the case?

adamabeshouse:
@alisman @jjgao fixed now, it wont break

alisman:
@adamabeshouse e2e test that is failing now looks it might actually be legit?

adamabeshouse:
I'm almost positive that would be because I updated a link in the tests to use FRACTION_GENOME_ALTERED instead of 1, and currently in production that wont load the correct clinical tracks, which is going to break that

adamabeshouse:
oh oops, no I actually forgot to change another link in the e2e tests. tests.. they work

PR #917, opened 2018-01-23
/beta bug fixes

adamabeshouse:
(1) Add typing and check to prevent undefined dereference exception
(2) Improve system of tracking changes to QueryStore to fix bug in case set selection


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1571,7 +1571,7 @@ export class ResultsViewPageStore {
         )
     }, ONCOKB_DEFAULT);
 
-    readonly getOncoKbMutationAnnotationForOncoprint = remoteData({
+    readonly getOncoKbMutationAnnotationForOncoprint = remoteData<(mutation:Mutation)=>(IndicatorQueryResp|undefined)>({
```
adamabeshouse: can typing qualify as test coverage? :p in conjunction with e2e

PR #927, opened 2018-01-26
Fail gracefully in oncoprint if webGL not available

adamabeshouse:
needs screenshot test


alisman:
@adamabeshouse how did you manage to test this manually?

PR #935, opened 2018-01-29
Reduce Survival chart size, make axis color black

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/3749

screen shot 2018-01-29 at 1 02 29 pm

### DIFF
--- a/src/pages/resultsView/survival/SurvivalChart.tsx
+++ b/src/pages/resultsView/survival/SurvivalChart.tsx
@@ -139,9 +139,9 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps, 
 
             <div className="posRelative">
 
-                <div className="borderedChart">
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -192,11 +193,10 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
                             : {this.tooltipModel.datum.x.toFixed(2)} months {this.tooltipModel.datum.status ? "" :
                                 "(censored)"}
                         </div>
-
                     </Popover>
                 }
 
-                <table className="table table-striped" style={{marginTop:20}}>
+                <table className="table table-striped" style={{marginTop:20, width: "79%"}}>
```
alisman: I would do with pixels because we may soon want to use a responsive layout in SPA

PR #936, opened 2018-01-29
Fix addGenesToQuery in case of custom case list

adamabeshouse:
fixes bug

incurring/perpetuating some technical debt https://github.com/cBioPortal/cbioportal/issues/3834

### DIFF
--- a/src/shared/components/query/QueryStoreUtils.ts
+++ b/src/shared/components/query/QueryStoreUtils.ts
@@ -50,15 +50,18 @@ export function queryParams(nonMolecularProfileParams:NonMolecularProfileQueryPa
     return {pathname: path, query:params};
 }
 
-export function nonMolecularProfileParams(store:QueryStore):NonMolecularProfileQueryParams {
+export function nonMolecularProfileParams(store:QueryStore, case_ids?:string):NonMolecularProfileQueryParams {
     const selectedStudyIds = store.allSelectedStudyIds;
+
+    case_ids = case_ids || store.asyncCustomCaseSet.result.map(caseRow => (caseRow.studyId + ':' + caseRow.sampleId)).join('\r\n');
+
     let ret:NonMolecularProfileQueryParams = {
         cancer_study_id: selectedStudyIds.length === 1 ? selectedStudyIds[0] : 'all',
         Z_SCORE_THRESHOLD: store.zScoreThreshold,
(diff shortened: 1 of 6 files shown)


src/shared/components/query/QueryStoreUtils.ts
```diff
@@ -52,13 +52,19 @@ export function queryParams(nonMolecularProfileParams:NonMolecularProfileQueryPa
 
 export function nonMolecularProfileParams(store:QueryStore):NonMolecularProfileQueryParams {
     const selectedStudyIds = store.allSelectedStudyIds;
+    let case_ids: string;
+    if (store.asyncCustomCaseSet.isComplete) {
+        case_ids = store.asyncCustomCaseSet.result.map(caseRow => (caseRow.studyId + ':' + caseRow.sampleId)).join('\r\n');
+    } else {
+        case_ids = store.caseIds;
+    }
```
inodb: @adamabeshouse how does this fix the issue exactly? This was failing before because `asyncCustomCaseSet` wasn't complete ye?

adamabeshouse: yes that's right, now instead we'll pass the caseIds as given from the page straight through

src/shared/components/query/QueryStoreUtils.ts
```diff
@@ -52,13 +52,19 @@ export function queryParams(nonMolecularProfileParams:NonMolecularProfileQueryPa
 
 export function nonMolecularProfileParams(store:QueryStore):NonMolecularProfileQueryParams {
     const selectedStudyIds = store.allSelectedStudyIds;
+    let case_ids: string;
```
alisman: this seems funky to me.  i would think that if there's a custom caseCase set call that isn't complete, then all of caseIds is pending and user should be waiting for that.  i.e. why show them caseIds preliminarily if custom caseIds is pending and might wipe caseIds out?

adamabeshouse: you're right, I should wait until the custom case set call completes. I'll make it work like that

inodb: We have an issue to clean some of this stuff up, a lot of things that are computed should prolly just be readonly and await the response: https://github.com/cBioPortal/cbioportal/issues/3771

PR #937, opened 2018-01-29
truncate numbers in tooltips to 2 decimal digits

adamabeshouse:


adamabeshouse:
this is a "nice" typescript "quirk" where `[0,1]` is interpreted as `number[]` and not as `[number, number]` so the type check fails

adamabeshouse:
@alisman I dont think the circleCI error is related to this PR

PR #939, opened 2018-01-29
Add methylation heatmap tracks to oncoprint

adamabeshouse:
needs screenshot tests


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1286,26 +1287,34 @@ export class ResultsViewPageStore {
...
+            const MRNA_EXPRESSION = AlterationTypeConstants.MRNA_EXPRESSION;
+            const PROTEIN_LEVEL = AlterationTypeConstants.PROTEIN_LEVEL;
+            const METHYLATION = AlterationTypeConstants.METHYLATION;
             const selectedMolecularProfileIds = stringListToSet(this.selectedMolecularProfileIds);
 
             return Promise.resolve(_.sortBy(_.filter(this.molecularProfilesInStudies.result!, profile=>{
-                return (profile.molecularAlterationType === MRNA_EXPRESSION ||
-                        profile.molecularAlterationType === PROTEIN_LEVEL) && profile.showProfileInAnalysisTab;
+                return ((profile.molecularAlterationType === MRNA_EXPRESSION ||
+                        profile.molecularAlterationType === PROTEIN_LEVEL) && profile.showProfileInAnalysisTab) ||
```
adamabeshouse: @angelicaochoa I have this in here for now because otherwise no methylation profiles show up, because none of them are shown in analysis tab. Should I keep it like this, or are you going to change that on the data side?

adamabeshouse:
@jjgao @schultzn @alisman what do we want to do with this PR? i dont think we should let it hang indefinitely or else it may become incompatible/need to be co-maintained

angelicaochoa:
@adamabeshouse Approved. Any feedback from @jjgao @schultzn?

schultzn:
Agree that we should merge it in to make it easier to maintain. We need it for the POETIC collaboration,

PR #946, opened 2018-01-30
Hide mutex tab from frontend if only one gene

adamabeshouse:


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -120,6 +121,13 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
         this.exposeComponentRenderersToParentScript(props);
 
         win.renderQuerySummary(document.getElementById('main_smry_info_div'));
+
+        // hide mutex tab
+        $(document).ready(()=>{
```
adamabeshouse: thoughts?

adamabeshouse:
@alisman I dont think the failing screenshot test is related...

PR #959, opened 2018-02-07
Override window.frontendConfig w localStorage

inodb:
This allows trying different frontend configurations configurations without
having to redeploy the war or change the frontend config file.

Try e.g.:
```
localStorage.frontendConfig = '{"showCivic":false}'
```

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -1,0 +1,13 @@
+// Override frontendConfig with localStorage frontendConfig if available
+let localStorageFrontendConfig:any = {};
+if (localStorage.frontendConfig) {
+    try {
+        localStorageFrontendConfig = JSON.parse(localStorage.frontendConfig);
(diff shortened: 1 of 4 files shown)


src/config/config.ts
```diff
@@ -0,0 +1,14 @@
+const config:any = (window as any).frontendConfig;
+// Override frontendConfig with localStorage frontendConfig if available
+if (localStorage.frontendConfig) {
+    try {
+        let localStorageFrontendConfig:any = JSON.parse(localStorage.frontendConfig);
+        for (var key in localStorageFrontendConfig) {
```
alisman: not important but you could use Object.assign to do this merge.

alisman: in fact, I would avoid using for in loop on objects because it will pick up properties from objects prototype.  unlikely, but has definately caused bug before.

PR #962, opened 2018-02-07
Counting multiple mutations for the same patient only once

onursumer:
# What? Why?
Related to https://github.com/cBioPortal/cbioportal/issues/3375

lollipop_mutation_count

- Counting multiple mutations for the same patient only once when generating the lollipops for the mutation diagram.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/lib/MutationUtils.ts
+++ b/src/shared/lib/MutationUtils.ts
@@ -100,6 +100,27 @@ export function groupMutationsByProteinStartPos(mutationData: Mutation[][]): {[p
     return map;
 }
 
+export function groupMutationsByGeneAndPatientAndProteinChange(mutations: Mutation[]): {[key: string]: Mutation[]}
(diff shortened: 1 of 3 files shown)


src/shared/lib/MutationUtils.ts
```diff
@@ -100,6 +100,27 @@ export function groupMutationsByProteinStartPos(mutationData: Mutation[][]): {[p
     return map;
 }
 
+export function groupMutationsByGeneAndPatientAndProteinChange(mutations: Mutation[]): {[key: string]: Mutation[]}
+{
+    // key = <gene>_<patient>_<proteinChange>
+    const map: {[key: string]: Mutation[]} = {};
+
+    for (const mutation of mutations)
```
alisman: I think you could use _.groupBy for more semantic code here

PR #965, opened 2018-02-08
run tests on circleci machine

alisman:
running webdriver manager locally seems more robust than running on browserstack

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,6 +83,10 @@ jobs:
           name: "Spin up frontend repo and run end to end tests"
           command: |
             npm run start & \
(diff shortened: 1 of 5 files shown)


.circleci/config.yml
```diff
@@ -94,7 +94,10 @@ jobs:
             sleep 20s && \
             cd end-to-end-tests && \
             npm install && \
-            npm run test-travis
+            npm install -g webdriver-manager && \
+            webdriver-manager update && \
+            webdriver-manager start && \
```
inodb: you'll want to start that one in the background, run with `&`. Then prolly sleep 1s or so after

PR #967, opened 2018-02-09
Add env files per branch

inodb:
Instead of only specifying one environment in my-index, we can now select the
master or rc environment by setting `export BRANCH_ENV=rc` or `export
BRANCH_ENV=master` and running `eval $(./scripts/env_vars.sh)`. That gets the
exports for each particular branch.` In webpack we then use the DefinePlugin to
get the correct environment variables. This means developers have to do an
extra step of selecting what env variables to use.

CircleCI finds out what branch the PR is pointing to automatically and gets the
appropriate env variables.


end-to-end-tests/wdio/webdriver-manager.conf.js
```diff
@@ -135,7 +135,15 @@ exports.config = {
     // Test reporter for stdout.
     // The only one supported by default is 'dot'
     // see also: http://webdriver.io/guide/testrunner/reporters.html
-    reporters: ['spec'],
+    reporters: ['spec', 'junit'],
```
inodb: this is to enable reports again on circleCI

inodb: the summary at the top of the run. kinda useful now the end to end test stuff is so bloated with selenium logs

package.json
```diff
@@ -206,7 +206,7 @@
     "swagger-client": "^2.1.32",
     "swagger-js-codegen": "git+https://github.com/cBioPortal/swagger-js-codegen.git#77031560146c6985c99cd40eda208b04052ab631",
     "ts-loader": "^2.0.0",
-    "typescript": "^2.5.0",
+    "typescript": "^2.6.2",
```
inodb: had to set 2.6 to be able to use ts-ignore for the global variables coming from the define plugin

src/appShell/App/App.spec.tsx
```diff
@@ -21,7 +21,7 @@ describe('<App />', () => {
     });
 
     it('passes a history prop', () => {
-        const props = wrapper.find('Router').props() as Router.RouterProps;
+        const props = wrapper.find('Router').props() as RouterProps;
```
inodb: TS 2.6 error

PR #970, opened 2018-02-13
Restore gene panel functionality in oncoprint

adamabeshouse:


src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -85,6 +89,49 @@ export function getOncoKbOncogenic(response:IndicatorQueryResp):string {
...
+export function computeGenePanelInformation(
+    genePanelData:GenePanelData[],
+    samples: Sample[],
+    patients: Patient[],
+    genes:Gene[]
+):GenePanelInformation {
+    const entrezToGene = _.keyBy(genes, gene=>gene.entrezGeneId);
+    const sampleInfo:GenePanelInformation["samples"] = {};
+    const patientInfo:GenePanelInformation["patients"] = {};
+    for (const sample of samples) {
```
alisman: not too important but i think use of reduce to build these collections would make this code clearer.  the reason is that you get simple assignment.   const sampleInfo = _.reduce(...).

adamabeshouse: good idea

adamabeshouse: done

alisman:
@adamabeshouse please rebase and merge

PR #973, opened 2018-02-15
Prevent submission of cross cancer queries if PROT in OQL

adamabeshouse:


src/shared/components/query/QueryStore.ts
```diff
@@ -1149,6 +1149,10 @@ export class QueryStore
 			return (result.alterations || []).some(alt => alt.alteration_type === 'exp');
 		});
 
+		const haveProtInQuery = this.oql.query.some(result => {
+			return (result.alterations || []).some(alt => alt.alteration_type === 'prot');
```
alisman: wow, some!  seems it is fully supported, too.

PR #974, opened 2018-02-16
Filter by and display sample counts in oncoprint clinical attribute selector

adamabeshouse:
image

image


pieterlukasse:
@adamabeshouse looks good to me. My only question would be related to the plans for adding this to the study view as well: this PR is still quite specific for results view and oncoprint. Are you planning to move parts like `clinicalAttributeIdToAvailableSampleCount` to a shared library so that it can easily be used by Study View? Or perhaps encapsulate this drop-down and related code as a generic component that can then be easily be added to both Oncoprint and Study View? 
Another question: will this PR disable the clinical attributes that have 0% coverage? Or can one still select these? Are the attributes sorted by % in the frontend (couldn't find any code doing that - not sure if you want to rely on the order of the ws for this)?

adamabeshouse:
@pieterlukasse 
- The clinical attributes are still sorted in alphabetical order, except for a few which are pushed to the top (which I'm now realizing this PR completely hides, which I need to fix).

- Attributes with 0% are filtered out.

- Study view is not currently in the new frontend project. And as of now this is not a reusable component - if that's a goal, I can refactor it to work that way.

jjgao:
@adamabeshouse great to see this. Is there any way to sort by availability?

adamabeshouse:
@jjgao of course, but there's no easy way to toggle between alphabetical and availability. Should we just sort by availability, with alphabetical between same availability?

jjgao:
@adamabeshouse ok to order just by availability if we cannot find an easy way to add both options.

cc'ing @schultzn

jjgao:
How could I test it? The `localStorage.heroku` trick didn't work for me... @adamabeshouse

adamabeshouse:
@jjgao maybe because its out of date, ill rebase and ping you

schultzn:
Looks good! If you want me to test it, can you send me a working link?

adamabeshouse:
@jjgao what should I show next to the names of attributes like mutation spectrum which cannot be queried for availability the same way because they're not real clinical attributes?

pieterlukasse:
@adamabeshouse thanks for the reply. Here are my comments:

- I also prefer to sort on availability. 
- Would be nice to leave in the 0% items, but perhaps disable them (so they can't be selected). I think it is good for users and data curators to see that there is a clinical attribute in the study that is still empty in all (selected) cases. Just hiding it would lead to more confusion and questions on why the variable is "missing". Remember also that a variable that is ~50% available for example, might be 0% available in a different case list. I think it will be confusing to sometimes have a clinical variable in the list and sometimes not.

adamabeshouse:
@schultzn good idea, I just pushed inferring % of cases for special attributes using sequenced/cna segment present data is implemented (though it seems the API endpoint is broken, so we'll wait for that to be fixed https://github.com/cBioPortal/cbioportal/issues/3951)

@pieterlukasse just pushed changes disabling instead of hiding 0% attributes. It's also now sorting on availability, displayname among equally available attributes

jjgao:
cc'ing @zhx828 and @kalletlak on this later. We should use this component later for attribute selection on the study view page as well.

jjgao:
@adamabeshouse could you updated once 1.13 is merged to rc?

jjgao:
@adamabeshouse @alisman is it possible to make the dropdown wider after I click?

adamabeshouse:
@jjgao try it now

jjgao:
@adamabeshouse somehow I couldn't make the code on beta site or production...

adamabeshouse:
@jjgao it seems like the heroku app was deleted, i just redeployed it hopefully it works now

jjgao:
looks great. Thanks, @adamabeshouse 

One tiny issue. Somehow the placeholder text for this query is missing.

http://www.cbioportal.org/beta/index.do?session_id=5af2fef1498eb8b3d5660b9c

image

adamabeshouse:
@jjgao I believe that is related to this issue: https://github.com/thehyve/react-select-checked/issues/17 @pieterlukasse @fedde-s

fedde-s:
Yes, it probably is. If so, I'm addressing it in thehyve/react-select-checked#19, and it will probably be solved in a bugfix release of the `react-select-checked` library soon.

fedde-s:
[Release 0.1.9](https://github.com/thehyve/react-select-checked/releases/tag/v0.1.9) of the `react-select-checked` component should now be available, including a fix for the missing placeholder text.

jjgao:
@adamabeshouse @alisman maybe we can prioritize this one for v1.15.0?

adamabeshouse:
Just rebased and updated react-select-checked version and pushed

jjgao:
Let's make this to 1.15.0 too.

adamabeshouse:
could I get a final review on this? @pieterlukasse

pieterlukasse:
@adamabeshouse I would like to test it. Can it be deployed to Heroku again?

adamabeshouse:
@pieterlukasse just deployed :+1:

fedde-s:
One check since I see a commit about gene panel data: does this change what data curators need to know about whether they should include case lists, gene panel data and special clinical attributes in their studies, relative to what's currently documented in [File-Formats.md](https://github.com/cBioPortal/cbioportal/blob/v1.14.0/docs/File-Formats.md) and assured by the validator?

pieterlukasse:
thanks @adamabeshouse . Unfortunately the deployment at heroku does not seem to be working...I get a blank screen.

adamabeshouse:
@pieterlukasse remember with frontend deploys you have to do `localStorage.heroku = 'cbioportal-frontend-pr-974'`

PR #980, opened 2018-02-20
Replaced CancerHotspots web API endpoints with Genome Nexus end point

onursumer:
# What? Why?
- Using genome nexus endpoint for cancer hotspots data. 

# Changes proposed in this pull request:
- Now mutations tab uses the data fetched from Genome Nexus.
- Oncoprint still uses the data fetched from OncoKB API (better to address this in a separate PR)

# Checks
  the `-s` option)


src/shared/lib/AnnotationUtils.spec.tsx
```diff
@@ -86,34 +166,23 @@ describe('AnnotationUtils', () => {
         hotspot3dIndex = indexHotspots(hotspots3d);
     });
 
-    it('properly creates hotspot index', () => {
```
onursumer: Moved to `CancerHotspotsUtils`

env/rc.sh
```diff
@@ -1,2 +1,2 @@
 export CBIOPORTAL_URL="http://www.cbioportal.org/beta"
-export GENOME_NEXUS_URL="http://genomenexus.herokuapp.com"
+export GENOME_NEXUS_URL="http://genomenexus.org/beta"
```
onursumer: Should this be the default URL for RC environment?

src/shared/model/CancerHotspots.ts
```diff
@@ -1,23 +1,5 @@
...
-
-export interface IHotspotData {
-    single: IHotspotIndex,
-    clustered: IHotspotIndex
-}
-
-export interface ICancerHotspotData {
-    single: HotspotMutation[],
-    clustered: HotspotMutation[]
+    [genomicLocation: string]: AggregatedHotspots;
```
onursumer: Now we are mostly relying on the data structure provided by Genome Nexus, thus the frontend cancer hotspots model is now a lot simpler.

src/shared/lib/StoreUtils.ts
```diff
@@ -687,53 +676,7 @@ export function findMrnaRankMolecularProfileId(molecularProfilesInStudy: MobxPro
     }
 }
 
-export async function fetchHotspotsData(mutationData:MobxPromise<Mutation[]>,
```
onursumer: This method is now simplified and moved into `CancerHotspotsUtils.ts`

src/shared/lib/StoreUtils.ts
```diff
@@ -687,53 +676,7 @@ export function findMrnaRankMolecularProfileId(molecularProfilesInStudy: MobxPro
...
-        })
-    ]);
 
-    return {
-        single: dataSingle,
-        clustered: data3d
-    };
-}
-
-export function indexHotspotData(hotspotData:MobxPromise<ICancerHotspotData>): IHotspotData|undefined
```
onursumer: simplified and moved into `CancerHotspotsUtils.ts`

env/rc.sh
inodb: yeah that makes sense

src/shared/lib/CancerHotspotsUtils.ts
```diff
@@ -0,0 +1,66 @@
...
+}
+
+export function genomicLocationString(genomicLocation: GenomicLocation)
+{
+    return `${genomicLocation.chromosome},${genomicLocation.start},${genomicLocation.end},${genomicLocation.referenceAllele},${genomicLocation.variantAllele}`;
+}
+
+export function isHotspot(mutation: Mutation, index: IHotspotIndex, filter?: (hotspot: Hotspot) => boolean): boolean
+{
+    let isHotspot = false;
```
adamabeshouse: shadowing is sort of confusing to read IMO - better to have different name for variable

onursumer:
@inodb good catch! Updated the PR to remove Cancer Hotspots urls as well.

src/shared/api/generated/GenomeNexusAPIInternal-docs.json
```diff
@@ -12,7 +12,7 @@
       "url": "https://github.com/cBioPortal/genome-nexus/blob/master/LICENSE"
     }
   },
-  "basePath": "/",
+  "basePath": "/beta",
```
alisman: @inodb is this right?  shouldn't this file be env agnostic?

onursumer: @alisman Probably not right, good point. I think the API generator tries to guess the `basePath` depending on the provided url, and that url depends on the genome nexus env parameter.

inodb: I don't think this parameter ends up in the generated client so it might not be a big deal. Just API sync error will prolly be raised on master when  generating client from genomenexus.org . Maybe we should strip out basepath info from the docs (add grep -v in the updateAPI script)

PR #898, opened 2018-01-17
Add gene set heatmap tracks to the Oncoprint

fedde-s:
If the user queries gene sets and a gene set profile in addition to
genes, add the corresponding tracks to the bottom of the Oncoprint with
an appropriate rule set.

## Screenshot ##
Screenshot of an Oncoprint showing some genetic tracks, some expression heatmap tracks, and below that, two gene set tracks in a green-to-dark-pink scale

PR #981, opened 2018-02-21
Expand correlated genes from gene set heatmaps

fedde-s:
Register callbacks to enable expansion options in the track menu of gene
set heatmap tracks in the Oncoprint. When these are clicked, retrieve
the list of correlated genes for the gene set and render these tracks as
the expansions of the gene set tracks. Tweak the colour of the gene set
tracks' labels in order to clearly distinguish them from gene tracks in
the same group.

This is a rewrite of the functionality originally reviewed in pull request
cBioPortal/cbioportal#2190, using the functionality I later ported to
the new stack via cBioPortal/oncoprintjs#34 and
cBioPortal/cbioportal-frontend#898.

## Screenshot ##
Screenshot showing a gene set heatmap with grey track labels, also containing five gene expression heatmaps in the regular colours under one of the tracks and showing the words ‘Show genes’ in the track menu for another

### DIFF
--- a/src/shared/cache/GenesetCorrelatedGeneCache.ts
+++ b/src/shared/cache/GenesetCorrelatedGeneCache.ts
@@ -1,0 +1,90 @@
+import {GenesetCorrelation} from "../api/generated/CBioPortalAPIInternal";
+import client from "shared/api/cbioportalInternalClientInstance";
+import _ from "lodash";
+import {IDataQueryFilter} from "../lib/StoreUtils";
+
+interface IQuery {
+    genesetId: string;
+    molecularProfileId: string;
+}
+
+type SampleFilterByProfile = {
+    [molecularProfileId: string]: IDataQueryFilter
+};
+
+async function fetch(
+    {genesetId, molecularProfileId}: IQuery,
+    sampleFilterByProfile: SampleFilterByProfile
+): Promise<GenesetCorrelation[]> {
+    const param = {
+        genesetId,
+        geneticProfileId: molecularProfileId,
+        ...sampleFilterByProfile[molecularProfileId]
+    };
+    return client.fetchCorrelatedGenesUsingPOST(param);
+}
+
+class GenesetCorrelatedGeneIteration {
+    private query: IQuery;
+    private sampleFilterByProfile: SampleFilterByProfile;
+    private nextGeneIndex = 0;
+    private data: undefined | GenesetCorrelation[] = undefined;
+
+    constructor(query: IQuery, sampleFilterByProfile: SampleFilterByProfile) {
+        this.query = query;
+        this.sampleFilterByProfile = sampleFilterByProfile;
+    }
+
+    async next(maxNumber: number) {
+        if (this.data === undefined) {
+            this.data = await fetch(this.query, this.sampleFilterByProfile);
+        }
+        // select the first 5 genes starting from the index,
+        // up to the end of the array
+        const nextGenes = this.data.slice(
+            this.nextGeneIndex,
+            this.nextGeneIndex + maxNumber
+        );
+        this.nextGeneIndex += nextGenes.length;
+        return nextGenes;
+    }
+
+    /**
+     * Resets the iteration so that next() will start from the beginning.
+     */
+    reset(): void {
+        this.nextGeneIndex = 0;
+    }
+}
+
+export default class GenesetCorrelatedGeneCache  {
+
+    private sampleFilterByProfile: SampleFilterByProfile;
+    private iterations: {
+        [iterationKey: string]: GenesetCorrelatedGeneIteration
+    } = {};
+
+    constructor(sampleFilterByProfile: SampleFilterByProfile) {
+        this.sampleFilterByProfile = sampleFilterByProfile;
+    }
+
+    initIteration(iterationKey: string, query: IQuery) {
+        if (!Object.prototype.hasOwnProperty.call(this.iterations, iterationKey)) {
+            this.iterations[iterationKey] = new GenesetCorrelatedGeneIteration(
+                query, this.sampleFilterByProfile
+            );
+        }
+    }
+
+    async next(
+        iterationKey: string,
+        maxNumber: number
+    ): Promise<GenesetCorrelation[]> {
+        return this.iterations[iterationKey].next(maxNumber);
+    }
+
+    reset(iterationKey: string): void {
+        this.iterations[iterationKey].reset();
+    }
+
+}
--- a/src/shared/cache/GenesetCorrelatedGeneCache.ts
+++ b/src/shared/cache/GenesetCorrelatedGeneCache.ts
@@ -24,36 +24,52 @@ async function fetch(
     return client.fetchCorrelatedGenesUsingPOST(param);
 }
 
+class AsyncStateChain<S> {
+    private stateChain: Promise<S>;
+    constructor(initState: S) {
+        this.stateChain = Promise.resolve(initState);
+    }
+    appendTransition<A>(
+        useState: (currentState: S) => Promise<{newState: S, output: A}>
+    ): Promise<A> {
+        // Synchronously replace stateChain by a promise of the new state,
+        // and return a promise of the output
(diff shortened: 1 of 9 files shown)


src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -301,6 +353,70 @@ export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sam
...
+            oncoprint.props.store.genesetCorrelatedGeneCache
+        ],
+        invoke: async () => {
+            const samples = oncoprint.props.store.samples.result!;
+            const patients = oncoprint.props.store.patients.result!;
+            const molecularProfileIdToMolecularProfile = oncoprint.props.store.molecularProfileIdToMolecularProfile.result!;
+            const dataCache = oncoprint.props.store.geneMolecularDataCache.result!;
+            const genesetGeneCache = oncoprint.props.store.genesetCorrelatedGeneCache.result!;
+
+            const trackGroup = oncoprint.genesetHeatmapTrackGroup;
```
fedde-s: There's a new issue here after a change in PR  #982; the expansions that end up being used in the gene set heatmap tracks no longer seem to correctly observe the track group observable.

fedde-s: I pushed two commits to fix the issue!

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -123,6 +130,8 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
     private heatmapGeneInputValueUpdater:IReactionDisposer;
 
     public selectedClinicalAttributeIds = observable.shallowMap<boolean>();
+    public genesetHeatmapTrackExpansionGenes =
```
adamabeshouse: can you make more explicit in the name what the map key represents? is it the track id?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -301,6 +364,70 @@ export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sam
...
+
+            const tracksByGenesetTrack: {[genesetTrackKey: string]: IGeneHeatmapTrackSpec[]} = {};
+            expansionsByGenesetTrack.entries().forEach(
+                ([gsTrack, genes]) => {
+                    tracksByGenesetTrack[gsTrack] = genes.map(
+                        ({entrezGeneId, hugoGeneSymbol, molecularProfileId, correlationValue}) => {
+                            const data = dataCache.get({entrezGeneId, molecularProfileId})!.data!;
+                            const profile = molecularProfileIdToMolecularProfile[molecularProfileId];
+                            return {
+                                key: `EXPANSIONTRACK_${gsTrack},${hugoGeneSymbol},GROUP${trackGroup}`,
```
adamabeshouse: the key shouldnt include the trackGroup, otherwise if the track group changes (e.g. "pushed up") then it won't be associated as the same track and will be removed and re-added by `DeltaUtils`

I'd say use the gene set id (is that the molecularProfileId?) instead

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: could you please explain in more detail what the problem was and what your fix is?

src/shared/components/oncoprint/OncoprintUtils.ts
fedde-s: That's not something I introduce in this PR, it's already in rc. Please log an issue there.

src/shared/components/oncoprint/OncoprintUtils.ts
fedde-s: Since there's no way to represent a track group as a whole moving in the APIs for either the `<Oncoprint>` component or the OncoprintJS library, moving the gene set heatmap to a different track group index involves removing the tracks in one heatmap and adding them back in another, as per [this commit in PR #898](https://github.com/cBioPortal/cbioportal-frontend/pull/898/commits/e19c30078d4389c0b3a7a1eac8d5c69007ffc71e). At the time I wrote the expansion feature there was no reason to ever remove or add a gene set track that already has expansions, so I never wrote the `DeltaUtils` logic to update its expansions in these cases; the fix is writing that logic.

src/shared/components/oncoprint/OncoprintUtils.ts
fedde-s: See my [other response](https://github.com/cBioPortal/cbioportal-frontend/pull/981#discussion_r171168585) to this review for an explanation on why this logic exists in rc.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
fedde-s: Basically yes, it's the track key (not specific to the track group). I've gotten used to clarifying these things in the type, but sadly `ObserveableMap` doesn't allow annotating the keys with more type information than ‘any string of characters will do’. Will change the name to reflect it.

src/shared/components/oncoprint/OncoprintUtils.ts
fedde-s: That commit might have been easier to review if I'd moved the function declaration to a different place in a separate commit though, sorry for that.

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: I see, so you have to force the oncoprint to register this as a new track in order to change its trackGroupIndex.

fedde-s:
I added a commit to make the relationship between the expansion promises and the gene set heatmap track promises more explicit — the former only serve as inputs to the latter.

fedde-s:
As for tests, any pointers to an existing test suite I might want to add to? I don't see many examples for the code that defines and deals with the MobXPromises of `OncoprintUtils.ts`. Would it suffice to test something about whether `DeltaUtils.ts` responds to gene set track specifications with and without expansion tracks by calling the expected OncoprintJS methods to add those tracks? That covers how `transitionGenesetHeatmapTracks()` is defined in terms of `transitionHeatmapTracks()`

adamabeshouse:
@fedde-s code looks great

src/shared/cache/GenesetCorrelatedGeneCache.ts
```diff
@@ -0,0 +1,90 @@
+import {GenesetCorrelation} from "../api/generated/CBioPortalAPIInternal";
```
adamabeshouse: this could use some tests

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -553,10 +563,36 @@ function transitionGenesetHeatmapTrack(
...
     oncoprint:OncoprintJS<any>,
     nextProps:IOncoprintProps,
-    trackIdForRuleSetSharing:{genesetHeatmap?:TrackId}
+    trackIdForRuleSetSharing:{genesetHeatmap?:TrackId, heatmap?:TrackId}
 ) {
+    function updateExpansionTracks(
+        nextParentSpec: IGenesetHeatmapTrackSpec | undefined,
+        prevParentSpec: IGenesetHeatmapTrackSpec | undefined
+    ) {
+        const expansionTrackList = prevParentSpec ? prevParentSpec.expansionTrackList || [] : [];
```
adamabeshouse: something like how many times transitionHeatmapTrack is called could be tested, not sure if its worth it. i'll leave it to your judgment

fedde-s:
@adamabeshouse, I included an update to the OncoprintJS release that includes the clustering fix, added tests for the cache class as you added, and immediately fixed a synchronisation issue between `.reset()` and `.next()` that I found because of the tests!

adamabeshouse:
@fedde-s glad that testing helped!! looks good, could you just rebase to rc and then i'll merge

PR #982, opened 2018-02-21
Enable clustering for gene set heatmaps

fedde-s:
List any queried gene set profile in the heatmap menu, displaying only
the clustering button as it's queried via the query page and the ‘Add
Genes’ and ‘Remove Heatmap’ buttons do not apply to it.

Collapse expansion tracks before clustering a heatmap, as these are
generally not of the same data type as the main tracks of the heatmap,
and including them when clustering would not make sense. 

This is a rewrite of functionality originally reviewed in pull request
cBioPortal/cbioportal#2186, using the functionality I later ported to
the new stack via cBioPortal/oncoprintjs#34 and
cBioPortal/cbioportal-frontend#898.

## Screenshot ##

Screenshot showing an Oncoprint with a clustered gene set heatmap and a heatmap menu containing only the cluster button for this profile

### DIFF
--- a/src/shared/components/oncoprint/DeltaUtils.ts
+++ b/src/shared/components/oncoprint/DeltaUtils.ts
@@ -32,11 +32,11 @@ export function transition(
         oncoprint.keepSorted(false);
     }
     trySuppressRendering(nextProps, prevProps, oncoprint);
-    transitionSortConfig(nextProps, prevProps, oncoprint);
     transitionWhitespaceBetweenColumns(nextProps, prevProps, oncoprint);
     transitionShowMinimap(nextProps, prevProps, oncoprint);
     transitionOnMinimapCloseCallback(nextProps, prevProps, oncoprint);
     transitionTracks(nextProps, prevProps, oncoprint, getTrackSpecKeyToTrackId);
+    transitionSortConfig(nextProps, prevProps, oncoprint);
     transitionTrackGroupSortPriority(nextProps, prevProps, oncoprint);
     transitionHiddenIds(nextProps, prevProps, oncoprint);
     transitionHorzZoomToFit(nextProps, prevProps, oncoprint);
@@ -162,7 +162,7 @@ function shouldNotKeepSorted_ClinicalTracksHelper(
     // Check if
     // (1) A track added/removed
     // (2) Track data changed
-    if (differentTracksOrChangedData(nextProps.geneticTracks || [], prevProps.geneticTracks || [])) {
+    if (differentTracksOrChangedData(nextProps.clinicalTracks || [], prevProps.clinicalTracks || [])) {
         return true;
     }
 }
@@ -174,7 +174,10 @@ function shouldNotKeepSorted_HeatmapTracksHelper(
     // Check if
     // (1) A track added/removed
     // (2) Track data changed
-    if (differentTracksOrChangedData(nextProps.geneticTracks || [], prevProps.geneticTracks || [])) {
+    if (
+        differentTracksOrChangedData(nextProps.heatmapTracks, prevProps.heatmapTracks || [] )
+        || differentTracksOrChangedData(nextProps.genesetHeatmapTracks, prevProps.genesetHeatmapTracks || [])
+    ) {
         return true;
     }
 }
--- a/src/shared/components/oncoprint/OncoprintUtils.ts
+++ b/src/shared/components/oncoprint/OncoprintUtils.ts
@@ -285,10 +285,13 @@ export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sam
                             trackGroup.genes.delete(gene);
                             if (!trackGroup.genes.size) {
                                 molecularProfileIdToHeatmapTracks.delete(molecularProfileId);
-                                if (oncoprint.sortMode.type === "heatmap" && oncoprint.sortMode.clusteredHeatmapProfile === molecularProfileId) {
-                                    oncoprint.sortByData();
-                                }
                             }
+                        }
+                        if (!molecularProfileIdToHeatmapTracks.has(molecularProfileId)
+                            && oncoprint.sortMode.type === "heatmap"
+                            && oncoprint.sortMode.clusteredHeatmapProfile === molecularProfileId
+                        ) {
+                            oncoprint.sortByData();
                         }
                     })
                 };
--- a/src/shared/components/oncoprint/OncoprintUtils.ts
+++ b/src/shared/components/oncoprint/OncoprintUtils.ts
@@ -310,8 +310,7 @@ export function makeGenesetHeatmapTracksMobxPromise(
             oncoprint.props.store.patients,
             oncoprint.props.store.genesetMolecularProfile,
             oncoprint.props.store.genesetMolecularDataCache,
-            oncoprint.props.store.genesetLinkMap,
-            oncoprint.genesetHeatmapTrackGroup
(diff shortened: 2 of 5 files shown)


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -431,7 +449,15 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                 this.heatmapGeneInputValue = s;
                 this.heatmapGeneInputValueUpdater(); // stop updating heatmap input if user has typed
             }),
-            onSelectHeatmapProfile:(id:string)=>{this.selectedHeatmapProfile = id;},
+            onSelectHeatmapProfile:(id:string)=>onMobxPromise(
+                this.props.store.molecularProfileIdToMolecularProfile,
+                profileMap => {
+                    this.selectedHeatmapProfile = id;
+                    this.heatmapIsDynamicallyQueried = (
```
adamabeshouse: cleaner to move this into an `autorun` that you set up in `constructor`, and remove it here and on line 164

fedde-s:
Nice catch with the error, I think I see the cause of it: something about the gene set tracks being re-rendered into a new track group, which is supposed to be the clustered track group after the re-render. I'll look into it later.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -242,7 +242,11 @@ export function transitionSortConfig(
         nextSortConfig.type === "order" &&
         (prevSortConfig.order !== nextSortConfig.order))
         ||
-        !_.isEqual(prevSortConfig, nextSortConfig)) {
+        !_.isEqual(prevSortConfig, nextSortConfig)
+    ) {
+        if (nextSortConfig.type === "cluster") {
+            oncoprint.removeAllExpansionTracksInGroup(nextSortConfig.track_group_index);
```
fedde-s: And indeed, this is wrong too: I should remove these expansion tracks in the React state, not just in the Oncoprint model.

fedde-s:
I fixed the issues you mentioned and three overlapping issues in how gene set heatmaps and addition/removal of other heatmaps play along, but haven't found the right place to remove expansions when starting to cluster a track group. Do you have an idea, @adama? Two actions can currently transition a track group to being clustered: clicking the cluster button in the heatmap menu, and moving the clustered gene set profile to a different track group by adding/removing other heatmaps above it.

fedde-s:
Either way, that missing functionality will not cause any issues until both this PR and #981 are merged into the same branch.

fedde-s:
The best place to program a solution would probably be in that pull request, not this one.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
fedde-s: Since I intended this observable to list a value that should at all times indicate whether the the profile ID in `this.selectedHeatmapProfile` identifies a non-geneset profile, I think a computed property that observes both `this.selectedHeatmapProfile` and `this.props.store.molecularProfileIdToMolecularProfile` is an even better fit than an autorun reaction. I pushed a commit!

fedde-s:
I think I addressed both your issues now, @adamabeshouse.

fedde-s:
Shall I squash the fixup commits again?

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -162,7 +162,7 @@ function shouldNotKeepSorted_ClinicalTracksHelper(
     // Check if
     // (1) A track added/removed
     // (2) Track data changed
-    if (differentTracksOrChangedData(nextProps.geneticTracks || [], prevProps.geneticTracks || [])) {
+    if (differentTracksOrChangedData(nextProps.clinicalTracks || [], prevProps.clinicalTracks || [])) {
```
adamabeshouse: oops!!

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -174,7 +174,10 @@ function shouldNotKeepSorted_HeatmapTracksHelper(
     // Check if
     // (1) A track added/removed
     // (2) Track data changed
-    if (differentTracksOrChangedData(nextProps.geneticTracks || [], prevProps.geneticTracks || [])) {
+    if (
+        differentTracksOrChangedData(nextProps.heatmapTracks, prevProps.heatmapTracks || [] )
```
adamabeshouse: wow I really f*cked this up! :D

fedde-s: At least it was only a little performance thingy up until I built heatmap tracks that jump around to different track groups 👼

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -435,11 +449,8 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
             onClickAddGenesToHeatmap:()=>{
                 this.addHeatmapTracks(this.selectedHeatmapProfile, this.heatmapGeneInputValue.toUpperCase().trim().split(/\s+/));
             },
-            onClickRemoveHeatmap:action(()=>{
+            onClickRemoveHeatmap:action(() => {
                 this.molecularProfileIdToHeatmapTracks.clear();
-                if (this.sortMode.type === "heatmap") {
```
adamabeshouse: whats going on here?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -285,11 +285,14 @@ export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sam
                             trackGroup.genes.delete(gene);
                             if (!trackGroup.genes.size) {
                                 molecularProfileIdToHeatmapTracks.delete(molecularProfileId);
-                                if (oncoprint.sortMode.type === "heatmap" && oncoprint.sortMode.clusteredHeatmapProfile === molecularProfileId) {
```
adamabeshouse: why move this down?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
fedde-s: For the reason I pointed out in the commit message for 499c12806ed72cd0415f30d8a3d2788112c6582d: there's no reason to uncluster the gene set heatmap when removing other heatmaps.

adamabeshouse: but this way we'll lose the correct behavior when theres no geneset..

adamabeshouse: oh maybe not because of the onRemove handler above, I see

src/shared/components/oncoprint/OncoprintUtils.ts
fedde-s: This removes duplicate logic to say ‘reset sorting when all tracks in the clustered profile are removed’.

When the tracks have been removed because the ‘Remove heatmap’ button has cleared the map via the code you commented on [above](https://github.com/cBioPortal/cbioportal-frontend/pull/982#discussion_r170300827), the callback does not need to update the map, but it does have to reset sorting.

fedde-s:
Squashed the commits that were amendments to the first one; the two other commits are related improvements to code and functionality that already existed in `rc`.

PR #983, opened 2018-02-21
Include reference links in Oncoprint gene set track tooltips

fedde-s:
When exploring the relevance of gene sets through the heatmap functionality in the Oncoprint, users sometimes want to look up what a gene set is actually about. The metadata known to cBioPortal include the URL of the main reference for the gene set. By making the Oncoprint setup retrieve this URL via the web API, a hyperlink can be presented to the user in the track tooltip.

In addition, mark word breaks to allow browsers to split the line after punctuation commonly found in long MSigDB gene set names, in case the name overflows the width of the tooltip.

This is a translation of the code originally reviewed in pull request thehyve/cbioportal#19,
using the functionality I later ported to the new stack via cBioPortal/oncoprintjs#34 and cBioPortal/cbioportal-frontend#898.

## Screenshot ##
Screenshot of a gene set track mouseover starting with the track label marked in blue and split over two lines

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1365,6 +1365,23 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly genesetLinkMap = remoteData<{[genesetId: string]: string}>({
+        invoke: async () => {
+            if (this.genesetIds && this.genesetIds.length) {
+                const genesets = await internalClient.fetchGenesetsUsingPOST(
+                    {genesetIds: this.genesetIds.slice()}
+                );
+                const linkMap: {[genesetId: string]: string} = {};
+                genesets.forEach(({genesetId, refLink}) => {
+                    linkMap[genesetId] = refLink;
+                });
+                return linkMap;
+            } else {
+                return {};
+            }
+        }
+    });
+
     readonly customDriverAnnotationReport = remoteData<{ hasBinary:boolean, tiers:string[] }>({
         await:()=>[
             this.mutations
--- a/src/shared/components/oncoprint/DeltaUtils.ts
+++ b/src/shared/components/oncoprint/DeltaUtils.ts
@@ -11,7 +11,10 @@ import {
     getGenesetHeatmapTrackRuleSetParams, getHeatmapTrackRuleSetParams
 } from "./OncoprintUtils";
 import {getClinicalTrackSortComparator, getGeneticTrackSortComparator, heatmapTrackSortComparator} from "./SortUtils";
-import {makeClinicalTrackTooltip, makeGeneticTrackTooltip, makeHeatmapTrackTooltip} from "./TooltipUtils";
+import {
+    makeClinicalTrackTooltip, makeGeneticTrackTooltip, makeHeatmapTrackTooltip,
+    linebreakGenesetId
+} from "./TooltipUtils";
 import {MolecularProfile} from "../../api/generated/CBioPortalAPI";
 
 export function transition(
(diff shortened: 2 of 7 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1365,6 +1365,23 @@ export class ResultsViewPageStore {
...
     });
 
+    readonly genesetLinkMap = remoteData<{[genesetId: string]: string}>({
+        invoke: async () => {
+            if (this.genesetIds && this.genesetIds.length) {
+                return (await internalClient.fetchGenesetsUsingPOST(
+                    {genesetIds: this.genesetIds.slice()}
+                )).reduce(
+                    (linkMap, {genesetId, refLink}) => (
+                        _.assign(linkMap, {[genesetId]: refLink})
```
adamabeshouse: I see you're doing this for a one-line-return, and it is nice and clever :) but not very readable, in my opinion.. can you just expand this all into a for-loop with more common syntax?

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -561,11 +564,13 @@ function transitionGenesetHeatmapTrack(
             has_column_spacing: false,
             track_padding: 0,
             label: nextSpec.label,
+            html_label: linebreakGenesetId(nextSpec.label),
             target_group: nextSpec.trackGroupIndex,
             sort_direction_changeable: true,
             sortCmpFn: heatmapTrackSortComparator,
             init_sort_direction: 0 as 0,
-            description: `${nextSpec.label} data from ${nextSpec.molecularProfileId}`,
+            description: `gene set scores from ${nextSpec.molecularProfileId}`,
```
adamabeshouse: capitalize "Gene"?

src/shared/components/oncoprint/TooltipUtils.ts
```diff
@@ -39,6 +39,20 @@ function makeGenePanelPopupLink(gene_panel_id:string) {
...
     return anchor;
 }
+
+export function linebreakGenesetId(
+    genesetId: string
+): string {
+    return (
+        // encode the string as the textual contents of an HTML element
+        $('<div>').text(genesetId).html()
+        // Include zero-width spaces to allow line breaks after punctuation in
```
adamabeshouse: nice idea, and helpful comment

src/pages/resultsView/ResultsViewPageStore.ts
fedde-s: I'm mostly doing this because I find the reduce more readable as an explicit transformation of an array to an object of a different shape, but I do agree that it would at the very least be better to name the intermediate array here.

adamabeshouse: what is slightly non-readable for me is _.assign(linkMap, ...) instead of linkMap[genesetId] = refLink

fedde-s: Right that's a Lodash fill-in for Object.assign(), a combination of that and returning the mutated linkMap to the next iteration. I can easily split those two things over two lines!

adamabeshouse: thanks :)

fedde-s: Even better solution that fixes both our objections (mine being pretty awkward to read in a language that doesn't properly understand immutability and yours talking about array indexes in logic that isn't about array indexes) — `Array.prototype.forEach()` to the rescue 😄

src/shared/components/oncoprint/TooltipUtils.ts
fedde-s: I considered building it in OncoprintJS itself, but it sort of depends on knowledge about what the headers will generally look like and where they can be sensibly split, so I had to keep it specific to gene set IDs.

fedde-s:
Addressed your comments, @adamabeshouse. Would you like me to squash the fixup commits?

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: good point and good solution

adamabeshouse:
thanks, yes please @fedde-s

fedde-s:
Done.

adamabeshouse:
Please rebase

fedde-s:
Done.

PR #987, opened 2018-02-23
Updating pull request template

pieterlukasse:
Adding a check note to remind developers of the importance of keeping the validation script up to date when implementing new assumptions about clinical attributes in the frontend.

Notify:
@sandertan

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -6,7 +6,7 @@ Changes proposed in this pull request:
 - b
 
 # Checks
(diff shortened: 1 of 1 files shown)


fedde-s:
Does this also apply to the semantics of other names and vocabularies listed verbatim in the data files, including genetic profile data types, case list categories, and (even if we agreed to stop doing this) hard-coded genetic profile IDs / case list IDs?

Or would that take more time to agree on? In that case I see no reason to postpone this improvement until it's done.

pieterlukasse:
Thanks @fedde-s , those are good points as well. I will add them to a next PR. Let us merge this one.

PR #997, opened 2018-03-01
Test genome nexus backwards compatability

inodb:


inodb:
ok seems to be working - also tested functional impact column. Still works

PR #999, opened 2018-03-01
Fix oncoprint sorting by custom case list

adamabeshouse:


kalletlak:
@adamabeshouse which format are we using for caseIds 1. `study1:sample1+study1:sample2` or 2. `study1:sample1\nstudy1:sample2`?
I think when sending request from study view it is sent in 1st format. @zhx828 correct me if I am wrong

adamabeshouse:
@kalletlak that seems to be true. I'm lost... is it just handled both ways everywhere?

jjgao:
@adamabeshouse we should support blank characters as the separator, e.g. space, tab, new line.

zhx828:
@adamabeshouse @kalletlak in the QueryBuilder, we use `study1:sample1+study1:sample2`

src/shared/components/query/QueryStore.ts
```diff
@@ -140,6 +140,8 @@ export class QueryStore
 				this.studiesHaveChangedSinceInitialization = true;
 			}
 		);
+
+		(window as any).queryStore = this;
```
alisman: queryStore is already exposed on window object under globalStores

adamabeshouse:
@jjgao it automatically gets URL encoded, so that's not a problem. I think we should stick with the + everywhere

PR #1002, opened 2018-03-03
Enable retries again for e2e tests

inodb:


env/master.sh
```diff
@@ -1,2 +1,2 @@
 export CBIOPORTAL_URL="http://www.cbioportal.org"
-export GENOME_NEXUS_URL="http://genomenexus.herokuapp.com"
+export GENOME_NEXUS_URL="https://genomenexus.herokuapp.com"
```
inodb: genomenexus.org is down :(

PR #1003, opened 2018-03-05
Fixes in query page related to GSVA

oplantalech:
Fix cbioportal/cbioportal#3899:
* Keep the gene sets in the query text box under "Modify Query" in results
* Disallow geneless queries

PR #1008, opened 2018-03-08
 Adapt some functions to deal with queries with only gene sets

oplantalech:
This PR makes some fixes to functions so that they don't crash when queries with only gene sets are submitted, including hiding gene-related tabs in the result page. 

Note: the first two commits are also present in PR #1003 (bugfixes). This last PR should be merged first, and then rebase the PR explained here.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -133,7 +133,7 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
 
         // hide mutex tab
         $(document).ready(()=>{
-            if (genes((window as any).serverVars.theQuery).length <= 1) {
+            if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length <= 1) {
                 $('a#mutex-result-tab').parent().hide();
             }
         });
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -136,7 +136,7 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
             if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length <= 1) {
                 $('a#mutex-result-tab').parent().hide();
             }
-            //hide gene-specific tabs when no genes are queried
+            //hide gene-specific tabs when we only query gene sets (and no genes are queried)
             if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length == 0) {
                 $('a#cancer-types-result-tab').parent().hide();
                 $('a#plots-result-tab').parent().hide();
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -136,6 +136,18 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
             if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length <= 1) {
                 $('a#mutex-result-tab').parent().hide();
             }
+            //hide gene-specific tabs when no genes are queried
+            if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length == 0) {
+                $('a#cancer-types-result-tab').parent().hide();
+                $('a#plots-result-tab').parent().hide();
+                $('a#mutation-result-tab').parent().hide();
+                $('a#coexp-result-tab').parent().hide();
+                $('a#enrichments-result-tab').parent().hide();
+                $('a#survival-result-tab').parent().hide();
+                $('a#network-result-tab').parent().hide();
+                $('a#igv-result-tab').parent().hide();
+                $('a#data-download-result-tab').parent().hide();
+            }
         });
     }
 
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -298,8 +298,9 @@ export class ResultsViewPageStore {
 
             // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
             const profilesWithoutMutationProfile = _.filter(this.selectedMolecularProfiles.result, (profile: MolecularProfile) => profile.molecularAlterationType !== 'MUTATION_EXTENDED');
+            const genes = this.genes.result;
 
-            if (profilesWithoutMutationProfile.length) {
+            if (profilesWithoutMutationProfile.length && genes != undefined && genes.length) {
 
                 const identifiers : SampleMolecularIdentifier[] = [];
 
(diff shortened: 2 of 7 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -131,11 +131,14 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
         win.renderQuerySummary(document.getElementById('main_smry_info_div'));
 
         // hide mutex tab
-        $(document).ready(()=>{
-            if (genes((window as any).serverVars.theQuery).length <= 1) {
-                $('a#mutex-result-tab').parent().hide();
-            }
-        });
+        if ((window as any).serverVars.theQuery.trim() != "") {
```
alisman: i think it would be safer to put this inside the ready callback.  otherwise, there's potential exec order issues.

alisman: actually, i'm not sure i understand this.  what is relationship here to genesets?

oplantalech: Since now we're supporting queries without genes, `genes((window as any).serverVars.theQuery)` fails as `window.serverVars.theQuery` is empty.

adamabeshouse: this logic is backwards - the mutex tab will never get hidden for a geneset query in this case

I think what you want is to put this on line 136: 
`if (!(window as any).serverVars.theQuery.trim().length || ...)`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -368,9 +369,12 @@ export class ResultsViewPageStore {
             this.defaultOQLQuery
         ],
         invoke:()=>{
-            return Promise.resolve(
-                filterCBioPortalWebServiceData(this.oqlQuery, this.unfilteredAlterations.result!, (new accessors(this.selectedMolecularProfiles.result!)), this.defaultOQLQuery.result!)
-            );
+            if (this.unfilteredAlterations.result!.length) {
```
adamabeshouse: this is necessary?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -368,9 +369,12 @@ export class ResultsViewPageStore {
...
+                return Promise.resolve(
+                        filterCBioPortalWebServiceData(this.oqlQuery, this.unfilteredAlterations.result!, (new accessors(this.selectedMolecularProfiles.result!)), this.defaultOQLQuery.result!)
+                );
+            }
```
adamabeshouse: add `} else {` for readability

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -432,21 +436,24 @@ export class ResultsViewPageStore {
...
+                                groupBy(oql.data, datum=>datum.uniqueSampleKey, this.samples.result!.map(sample=>sample.uniqueSampleKey)),
+                            patients:
+                                groupBy(oql.data, datum=>datum.uniquePatientKey, this.patients.result!.map(sample=>sample.uniquePatientKey))
+                        };
+                        return {
+                            cases,
+                            oql
+                        };
+                    }));
+            }
```
adamabeshouse: add `} else {` for readability

src/shared/components/query/QueryStore.ts
```diff
@@ -1273,6 +1273,7 @@ export class QueryStore
 		if ((_window as any).serverVars) {
 			// Populate OQL
 			this.geneQuery = normalizeQuery((_window as any).serverVars.theQuery);
+			this.genesetQuery = normalizeQuery((_window as any).serverVars.genesetIds);
```
adamabeshouse: Just confirming - `normalizeQuery` was written for OQL. it works for this too?

oplantalech: Yes!

adamabeshouse:
console errors:
image

adamabeshouse:
empty tab: 
image

adamabeshouse:
you should hide irrelevant tabs in this PR - its something to do from the frontend, and as it is this PR leads to broken situations.

you can do that in same place/similar way we're hiding mutex tab: https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/ResultsViewPage.tsx#L127

oplantalech:
@adamabeshouse I've moved the hiding of the tabs here in the frontend, and fixed the bug regarding alterations in the Oncoprint.

oplantalech:
@adamabeshouse @alisman I've added a new commit, which removes OQL references to the gene set queries, since this was confusing and illogical. Also, since they are problems to merge #1003 due to issues in rc, and #1003 is basically a temporary fix until the PR here is merged, I would suggest to finish this PR and try to merge it, and forget about #1003. What do you think?

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -132,9 +132,21 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
 
         // hide mutex tab
         $(document).ready(()=>{
-            if (genes((window as any).serverVars.theQuery).length <= 1) {
+            if (!(window as any).serverVars.theQuery.trim().length || genes((window as any).serverVars.theQuery).length <= 1) {
```
alisman: @adamabeshouse lets rename this function "genes" to something more descriptive.   it seems to parse query and return genes symbols, so something to that effect.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -132,9 +132,21 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
                 $('a#mutex-result-tab').parent().hide();
             }
+            //hide gene-specific tabs when no genes are queried
```
alisman: lets add to this comment to explain why there might be no genes.  namely that we have a geneset query.  we might want to actually have a function isGensetQuery and test against that here, which would remove the ambiguity.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -297,8 +297,9 @@ export class ResultsViewPageStore {
 
             // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
             const profilesWithoutMutationProfile = _.filter(this.selectedMolecularProfiles.result, (profile: MolecularProfile) => profile.molecularAlterationType !== 'MUTATION_EXTENDED');
+            const genes = this.genes.result;
 
-            if (profilesWithoutMutationProfile.length) {
+            if (profilesWithoutMutationProfile.length && genes != undefined && genes.length) {
```
alisman: rather than put this check here, would it be possible instead of just not rely on molecularData mobxpromise when we're in geneset mode?  just worth a look because i would rather not introduce into this logic the case where there are no genes, since that is not part of its use case.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -368,9 +369,13 @@ export class ResultsViewPageStore {
             this.defaultOQLQuery
         ],
         invoke:()=>{
-            return Promise.resolve(
-                filterCBioPortalWebServiceData(this.oqlQuery, this.unfilteredAlterations.result!, (new accessors(this.selectedMolecularProfiles.result!)), this.defaultOQLQuery.result!)
-            );
+            if (this.oqlQuery.trim() != "") {
```
alisman: again, i think a better way to solve this would be to not actually invoke filteredAlterations when there are no genes.

alisman: @adamabeshouse what do you think about this idea?

adamabeshouse: @alisman and I discussed and decided it doesnt make sense to think of two different modes

oplantalech:
@adamabeshouse I've discussed this PR with @alisman and as far as I know, the only thing remaining to do here was to improve a comment in the code. So, can you guys approve and merge this PR? Thanks.

PR #1010, opened 2018-03-08
Fix lollipop when canonical transcript is missing

inodb:
Get all transcripts with given hugo symbol and show longest.

Fix https://github.com/cBioPortal/cbioportal/issues/3922

### DIFF
--- a/src/pages/resultsView/mutation/MutationMapperStore.ts
+++ b/src/pages/resultsView/mutation/MutationMapperStore.ts
@@ -18,7 +18,7 @@ import {
     indexHotspotData, fetchHotspotsData, fetchCosmicData, fetchOncoKbData,
     fetchMutationData, generateUniqueSampleKeyToTumorTypeMap, generateDataQueryFilter,
     ONCOKB_DEFAULT, fetchPdbAlignmentData, fetchSwissProtAccession, fetchUniprotId, indexPdbAlignmentData,
-    fetchPfamDomainData, fetchCivicGenes, fetchCivicVariants, IDataQueryFilter, fetchCanonicalTranscript,
+    fetchPfamDomainData, fetchCivicGenes, fetchCivicVariants, IDataQueryFilter, fetchCanonicalTranscriptWithFallback,
     fetchEnsemblTranscriptsByEnsemblFilter
 } from "shared/lib/StoreUtils";
 import MutationMapperDataStore from "./MutationMapperDataStore";
@@ -152,7 +152,7 @@ export class MutationMapperStore {
             }
         },
         onError: (err: Error) => {
-            console.log("Getting all transcripts failed");
+            throw new Error("Failed to fetch all transcripts");
         }
     }, undefined);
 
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -142,10 +143,33 @@ export class MutationMapperStore {
...
+    }, undefined);
+
     readonly canonicalTranscript = remoteData<EnsemblTranscript | undefined>({
+        await: () => [
+            this.allTranscripts
+        ],
         invoke: async()=>{
             if (this.gene) {
-                return fetchCanonicalTranscript(this.gene.hugoGeneSymbol, this.isoformOverrideSource);
+                return fetchCanonicalTranscript(this.gene.hugoGeneSymbol, this.isoformOverrideSource).catch(async() => {
```
onursumer: This seems to me a bit complicated as a return statement, maybe we should move the logic within the catch block into another utility function?

src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -142,10 +143,33 @@ export class MutationMapperStore {
...
+                    console.log("No canonical transcript for: " + this.gene.hugoGeneSymbol);
+                    const transcript = _.maxBy(this.allTranscripts.result, (t:EnsemblTranscript) => t.proteinLength);
+                    if (transcript) {
+                        console.log("Using transcript with longest length instead: " + transcript);
+                        return transcript;
+                    }
+                })
```
onursumer: missing semicolon!

src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -142,10 +143,33 @@ export class MutationMapperStore {
...
+            this.allTranscripts
+        ],
         invoke: async()=>{
             if (this.gene) {
-                return fetchCanonicalTranscript(this.gene.hugoGeneSymbol, this.isoformOverrideSource);
+                return fetchCanonicalTranscript(this.gene.hugoGeneSymbol, this.isoformOverrideSource).catch(async() => {
+                    console.log("No canonical transcript for: " + this.gene.hugoGeneSymbol);
+                    const transcript = _.maxBy(this.allTranscripts.result, (t:EnsemblTranscript) => t.proteinLength);
+                    if (transcript) {
+                        console.log("Using transcript with longest length instead: " + transcript);
```
onursumer: I have always had mixed feelings about `console.log` :)

src/pages/resultsView/mutation/MutationMapperStore.ts
alisman: is it a failure that we expect or a real break down?

inodb: one that we expect. As a backup it uses the same service, so a real breakdown would still fail

src/pages/resultsView/mutation/MutationMapperStore.ts
inodb: same, if we have a better way of logging warnings would be good

inodb:
Changed this to master, since current bug is there

PR #1019, opened 2018-03-12
Use new render completion callback in oncoprint to hide while rendering

adamabeshouse:


adamabeshouse:
https://github.com/cBioPortal/cbioportal-frontend/pull/1035

PR #1020, opened 2018-03-12
Expose OQL.genes to window

adamabeshouse:


src/globalComponents.tsx
```diff
@@ -15,6 +16,9 @@ class GlobalStores {
 }
 
 (window as any).getStudySummaryUrl = getStudySummaryUrl;
+(window as any).oqlGenes = (oqlQuery:string)=>{
```
inodb: @adamabeshouse Prolly not in this PR. But would be nice to expose these in a namespaced manner. E.g. `window.frontendVars.oqlGenes` similar for other variables/functions. Makes it a little easier to see in backend project where tight integration happens

adamabeshouse: good idea

alisman:
should this go to rc though?

adamabeshouse:
@alisman I thought master since its kind of a bug... but rc fine too. whatever you think

fedde-s:
I ran into the issue when trying to implement a new feature, so for that I'll only need it in `rc` for now. I'll probably have time to check if it fixes the specific problem I was having in the next few days, but I don't mind if it's already been merged into `rc` by then, in which case I'll just address any follow-up issues iteratively.

src/shared/lib/oql/oqlfilter.js
```diff
@@ -1,4 +1,4 @@
-// Requires parser to be window object, window.oql_parser
+// Requires parser to be window object, oql_parser
```
fedde-s: I still find this comment a bit confusing. Does it actually require anything from the window object, or does our tooling (Webpack, maybe helped by Babel?) just make the import statement below work?

alisman: no this is not related to tooling.  it's basically just saying, this is an expected global

alisman: actually this comment is defunct, right? b/c we now import it below.  strike comment?

adamabeshouse: heh this is the result of a find and replace, this comment should be deleted. good catch

PR #1027, opened 2018-03-15
Remove TESTS.xml file

ersinciftci:


fedde-s:
I would like some feedback from whoever configured `npm run test` to generate it in the first place though, to see if they agree that we should add it to `.gitignore`.

fedde-s:
@inodb, is that you?

PR #1007, opened 2018-03-07
Refactor co-expressions tab

adamabeshouse:

PR #1040, opened 2018-03-16
Mobx cache

adamabeshouse:


src/shared/lib/MobxCache.ts
```diff
@@ -1,64 +1,39 @@
...
 
-    public destroy() {
-        this.clearReaction && this.clearReaction();
+    public getAll(queries:Query[]):MobxPromise<Result>[] {
+        const key = queries.map(q=>this.queryToKey(q));
+        const existing = this.allCache.get(key);
+        if (existing) {
+            return existing;
+        } else {
+            const newPromise = MobxPromise.all(
```
adufilie: I realize you already changed this in another commit, but I would like to note that here you are creating a new MobxPromise each time, which is not in the spirit of MobxPromise.  Instead, if you do indeed want a MobxPromise for all queries, it should be created once and updated dynamically like this:

```
    readonly all = new MobxPromise({
        await: () => this.queries.map(q=>this.get(q)),
        invoke: async () => this.queries.map(q=>this.get(q).result),
        default: []
    });
```

adufilie: Well, looks like I misread the code a bit. _I See What You Did There_ with the `key`.  The readonly MobxPromise would still be better I think.

adufilie: Man, I need sleep. I keep misreading the code. :slightly_smiling_face: I'll leave you to it.

adamabeshouse:
included in https://github.com/cBioPortal/cbioportal-frontend/pull/1007

PR #1045, opened 2018-03-19
Fix N/S in oncoprint

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/3990 and https://github.com/cBioPortal/cbioportal/issues/3988


src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -116,7 +116,10 @@ export function computeGenePanelInformation(
         const patientSequencingInfo = patientInfo[gpData.uniquePatientKey];
         const hugo = entrezToGene[gpData.entrezGeneId].hugoGeneSymbol;
 
-        if (gpData.genePanelId !== undefined) {
+        if (gpData.genePanelId && gpData.sequenced) {
```
alisman: is there an assertion against this?

adamabeshouse: the changes in this PR to ResultsViewPageStoreUtils.spec.ts made the relevant tests fail, and this change makes them pass again.

PR #1057, opened 2018-03-22
Update frontend to work with new gene panel API

adamabeshouse:


inodb:
@ersinciftci @adamabeshouse I just deployed to cbioportal.org/beta. Seems like gene panel data is much faster again now, (2.6s). Most time spent in TTFB (1.4s). Anyway this is not that big of a deal. Biggest performance gains we can get is prolly still on frontend itself. At least performance wise this looks good to go for me

adamabeshouse:
@inodb I think the tests are failing because master isn't deployed yet

inodb:
@adamabeshouse I believe so. Problem is that they aren't backwards compatible. So I can't deploy master to cbioportal.org - we could point the env/master.sh environment temporarily to cbioportal.org/beta

inodb:
@adamabeshouse ill add a commit to this pr to do that

PR #1059, opened 2018-03-23
Oncoprint optimizations and bug fixes

adamabeshouse:
Use new oncoprintjs feature in https://github.com/cBioPortal/oncoprintjs/pull/67 to do a bucket sort instead of comparison sort for genetic tracks (impact x 30 genes loads in 19s now instead of 45s)

Also fixes oncoprint tooltip positioning issue


inodb:
@adamabeshouse That's quite a speed-up - nice. Would be nice to know why this data specifically works well w bucket sort vs any of the others: http://bigocheatsheet.com/.

adamabeshouse:
@inodb bucket sort is average case n+k, vs comparison-based sorts which are average case nlogn, and the difference becomes quite large when n is mskimpact-sized

src/shared/components/oncoprint/SortUtils.ts
```diff
@@ -44,53 +49,50 @@ export function getGeneticTrackSortComparator(sortByMutationType?: boolean, sort
...
+        } else {
+            vector.push(1);
         }
 
         // Next, CNA
-        const cna_diff = sign(cna_order[d1.disp_cna+""] - cna_order[d2.disp_cna + ""]);
-        if (cna_diff !== 0) {
-            return cna_diff;
-        }
+        vector.push(cna_order[d.disp_cna+""]);
```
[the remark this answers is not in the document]
adamabeshouse: this is still pretty idiomatic and concise but if you think that would be more readable im fine w it

package-lock.json
```diff
@@ -12054,7 +12054,7 @@
     },
     "onetime": {
       "version": "1.1.0",
-      "resolved": "https://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
+      "resolved": "http://registry.npmjs.org/onetime/-/onetime-1.1.0.tgz",
```
inodb: @adamabeshouse so i think best to revert this change and rebase on https://github.com/cBioPortal/cbioportal-frontend/commit/0b6823c795e659a4fd5745e9b34d6b17ba1683a4

adamabeshouse: yep just waiting for the e2e tests to run so i can see if any failures

src/shared/components/oncoprint/Oncoprint.tsx
```diff
@@ -1,14 +1,8 @@
...
-    ClinicalAttribute, GeneMolecularData, GenePanelData, MolecularProfile,
-    Mutation
-} from "../../api/generated/CBioPortalAPI";
+import OncoprintJS, {TrackId} from "oncoprintjs";
+import {GenePanelData, MolecularProfile} from "../../api/generated/CBioPortalAPI";
 import {observer} from "mobx-react";
-import {computed, observable} from "mobx";
-import {doWithRenderingSuppressedAndSortingOff, getClinicalTrackRuleSetParams, getGeneticTrackRuleSetParams} from "./OncoprintUtils";
-import {getClinicalTrackSortComparator, getGeneticTrackSortComparator, heatmapTrackSortComparator} from "./SortUtils";
+import {computed} from "mobx";
```
adamabeshouse: this section is just cleaning up imports

adamabeshouse:
@inodb actually, thinking about it more, the reason this algorithm works well here is because we can specify the sort order of the oncoprint elements with arrays with small numbers. Instead of doing comparison based O(nlogn) sort on all of the n=10000 samples, it reduces it to a series of bucket sorts, where the samples are put into buckets depending on their array element values, and then we do a comparison sort but only on the values. So instead of O(nlogn), we have basically a round of bucket-izing for each array element, and then a comparison sort of the buckets. So if there are less than M buckets (distinct values) in each array element, and the arrays are of length K, then the total sort complexity is O(K*(MlogM + n)) = O(KMlogM + Kn). In this case, K is less than 10, M is less than 20, but n grows up to 10000 for impact, so it makes a big difference. But it wouldn't really make any difference if the values are roughly unique per sample, like in fraction genome altered.

inodb:
@adamabeshouse makes sense - Thanks for the details!

PR #1068, opened 2018-04-04
Removed PdbAnnotation API dependency

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3159.

- Now using Genome Nexus for `PdbHeader` data
- Removed all the references, and files related to `PdbAnnotationAPI`

# Checks
  the `-s` option)


src/shared/cache/PdbPositionMappingCache.ts
```diff
@@ -1,44 +0,0 @@
...
-    const uniqueAlignments = _.uniq(queries.map(q => q.alignmentId));
-    const uniquePositions = _.uniq(queries.map(q => q.uniprotPosition));
-
-    // TODO postPositionMap actually returns a map of <uniprotPosition: PdbUniprotResidueMapping[]>
-    // so we need to flatten to match the actual promise data
-    return _.flatten(_.values(await fetchPdbPositionMappingData(uniquePositions, uniqueAlignments)));
-
-}
-
-export default class PdbPositionMappingCache extends LazyMobXCache<PdbUniprotResidueMapping, AlignmentIdAndUniportPos> {
```
onursumer: We haven't been using this cache since merging https://github.com/cBioPortal/cbioportal-frontend/pull/567, so removing until we re-implement it for the Genome Nexus endpoint data.

src/shared/model/Pdb.ts
```diff
@@ -12,7 +12,7 @@ export interface IPdbPositionRange {
 
 export type PdbAlignmentIndex = {
     [pdbId: string]: {
-        [chainId: string]: PdbUniprotAlignment[]
+        [chainId: string]: Alignment[]
```
onursumer: This should have been done by https://github.com/cBioPortal/cbioportal-frontend/pull/567. Seems like it slipped away type-checking because in the `PdbAnnotationAPI` it was generated as `any`: 

```JavaScript
export type PdbUniprotAlignment = any;
```

src/pages/resultsView/mutation/MutationMapperStore.ts
```diff
@@ -244,9 +243,12 @@ export class MutationMapperStore {
         return new PdbChainDataStore(this.sortedMergedAlignmentData);
     }
 
+    // TODO currently not used, implement cache based on Genome Nexus endpoint data
+    // (not straightforward to implement, data structure is different from the previous one)
     @cached get pdbPositionMappingCache()
     {
-        return new PdbPositionMappingCache();
+        // return new PdbPositionMappingCache();
+        return null;
```
adamabeshouse: do we not want to just delete this?

onursumer: Ideally we should implement a new cache for the new structure, but for now it might be better to just remove it completely.

onursumer: removed it completely for now, and added a new issue: https://github.com/cBioPortal/cbioportal/issues/4071

src/shared/components/proteinChainPanel/column/OrganismColumnFormatter.tsx
```diff
@@ -12,12 +12,12 @@ export default class OrganismColumnFormatter {
 
         _.find(pdbHeader.compound, (mol:any)=>{
             if (_.indexOf(mol.chain, chainId.toLowerCase()) != -1 &&
-                pdbHeader.source[mol.mol_id] != null)
+                (pdbHeader.source as any)[mol.mol_id] != null)
```
alisman: curious why this isn't part of generated type?

end-to-end-tests/specs/results.spec.js
```diff
@@ -135,6 +135,29 @@ describe('Results Page', function() {
...
+            before(() => {
+                var url = `${CBIOPORTAL_URL}/index.do?tab_index=tab_visualize&cancer_study_list=ov_tcga_pub&cancer_study_id=ov_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ov_tcga_pub_mutations&Z_SCORE_THRESHOLD=2.0&case_set_id=ov_tcga_pub_3way_complete&case_ids=&gene_list=BRCA1+BRCA2&gene_set_choice=user-defined-list&Action=Submit`;
+                browser.url(url);
+                browser.click("[href='#mutation_details']");
+                browser.waitForEnabled('[data-test=view3DStructure]', 10000);
+            });
+
+            it('populates PDB info properly', () => {
+                browser.click('[data-test=view3DStructure]');
+                browser.waitUntil(() => (browser.getText('[data-test=pdbChainInfoText]') !== "LOADING"), 10000);
```
onursumer: @alisman @inodb not sure if this is the best way to wait until the actual text loads

inodb: @onursumer could use `waitForExist` but this works

onursumer: @inodb The problem is that `pdbChainInfoText` always exists, but its text content changes from `LOADING` to a real value. And the actual content is generated by another generic component called `TextExpander`. 

We could potentially render separate `div`s or `span`s for different states to be able to use `waitForExist` though.

PR #1078, opened 2018-04-09
Remove scroll:auto on oncoprint because it is causing various problem…

alisman:
I made a change to oncoprint css to hide scrollbars when they were not needed (no overflow).  This caused various problems in different browsers so I am reverting the cosmetic change.

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


adamabeshouse:
I cant find it now but, wont this bring back the issue you were fixing, with the vertical scrollbar not appearing or something like that?

end-to-end-tests/wdio/webdriver-manager.conf.js
```diff
@@ -35,18 +35,14 @@ exports.config = {
     // and 30 processes will get spawned. The property handles how many capabilities
     // from the same test should run tests.
     //
-    maxInstances: 1,
+    maxInstances: 5,
```
adamabeshouse: :+1:

alisman:
@adamabeshouse no, this is reversion to the way the scrolling initially was: forced-on.  I don't like it when the scrollbars show when no scrolling is actually possible, but that seems to be the most robust way of solving all these various subtle but horrible behavior inconsisteencies.

end-to-end-tests/wdio/webdriver-manager.conf.js
inodb: @alisman Wondering how this relates to number of cores on physical machine.

alisman:
@adamabeshouse I think we reached conclusion that commits don't always have to be squashed if they actually represent different tasks/updates and we want that preserved in commit history.  In this case they do, although I think the screenshot updates should probably be merged with the updates that necessitated them.  I will do that.

alisman:
squashing done!

PR #1080, opened 2018-04-10
Add tests for volcano plot popup and fix error

oplantalech:
I've added tests for #830 and I fixed an error I found when calculating the max Y value for the volcano plot.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -36,6 +36,7 @@ import {CacheData} from "../../lib/LazyMobXCache";
 import { getHierarchyData } from "shared/lib/StoreUtils";
 import sessionServiceClient from "shared/api//sessionServiceInstance";
 import {VirtualStudy} from "shared/model/VirtualStudy";
+import { getGenesetsFromHierarchy, getVolcanoPlotMinYValue, getVolcanoPlotData } from "shared/components/query/GenesetsSelectorStore";
 
 // interface for communicating
 export type CancerStudyQueryUrlParams = {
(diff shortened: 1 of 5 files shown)


src/shared/components/query/QueryStore.spec.ts
```diff
@@ -1,10 +1,12 @@
 import {assert, expect} from "chai";
-import {QueryStore, CUSTOM_CASE_LIST_ID} from "./QueryStore";
+import {QueryStore, CUSTOM_CASE_LIST_ID, buildTableData, generateMaxYValue, generateVolcanoPlotData} from "./QueryStore";
```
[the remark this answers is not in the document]
alisman: really these should probably be in a helper file, not in QueryStore, as they pertain only to a very specific feature.

src/shared/components/query/QueryStore.ts
```diff
@@ -1684,3 +1658,48 @@ export const QueryStoreComponent = ComponentGetsStoreContext(QueryStore);
 
 const selectedGeneSets = '';
 export default selectedGeneSets;
+
+export function buildTableData(hierarchyData: any[]): Geneset[] {
+	let tableData: Geneset[] = [];
```
alisman: i think this function makes an unnecessary copy of the data and iterates needlessly through all the items when what you're doing is just concatenating.  I think _.flatMap will actually accomplish this in one statement:
_.flatMap((node)=>return node.genesets||[]) might do the trick.

alisman: ah, i see you are renaming gensesetId -> name.   seems like a big cost for this, no?

src/shared/components/query/QueryStore.ts
```diff
@@ -1684,3 +1658,48 @@ export const QueryStoreComponent = ComponentGetsStoreContext(QueryStore);
...
+	            });
+	        }
+	    }
+	}
+	return tableData;
+};
+
+export function generateMaxYValue(tableData: Geneset[]|undefined): number|undefined
+{
+    if (tableData) {
```
alisman: why would tabledata be undefined?  getter to handle this outside of the function.  this function should assume data.

src/shared/components/query/QueryStore.spec.ts
```diff
@@ -122,4 +124,92 @@ describe("QueryStore", ()=>{
...
+                	genesetId: "AKT_UP.V1_UP",
+                	name: "AKT_UP.V1_UP",
+                	refLink: "http://www.example.com",
+                	representativePvalue: 0.0862,
+                	representativeScore: 0.3945
+                }]
+            );
+        });
+        it("calculates the correct max Y value for the volcano plot", ()=>{
+            assert.equal(maxYValue, -(Math.log(0.0736)/Math.log(10)));
```
alisman: you want to try to put the call to the code you are testing (generateMaxYValue) inside of the call to it.  this makes it a lot clearer.

src/shared/components/query/QueryStore.spec.ts
```diff
@@ -122,4 +124,92 @@ describe("QueryStore", ()=>{
...
+        });
+        it("builds the data for the volcano plot graph with the correct values", ()=>{
+            assert.deepEqual(graphData, 
+            	[{
+                	x: 0.1986,
+                	y: -(Math.log(0.0736)/Math.log(10)),
```
alisman: instead of hardcoding the number, make a reference back to seed data so it's not duplicated.

src/shared/components/query/QueryStore.spec.ts
```diff
@@ -122,4 +124,92 @@ describe("QueryStore", ()=>{
             // },1000)  
         });
     });
-});
\ No newline at end of file
+});
+describe("QueryStore", ()=>{
+    describe("Volcano Plot related functions", ()=>{
+        const genesets: Geneset[] = [{
```
alisman: see mocha documentation for use of before and beforeEach for data "set up"  this is a good habit to get into because for more involved tests which might involve state or instances of classes, it's very important to prevent tests from interfering with each other

src/shared/components/query/QueryStore.ts
```diff
@@ -1684,3 +1658,48 @@ export const QueryStoreComponent = ComponentGetsStoreContext(QueryStore);
...
+        let genesetWithMaxY = _.maxBy(tableData, function(o) { return -(Math.log(o.representativePvalue)/Math.log(10)); });
```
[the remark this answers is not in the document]
oplantalech: I think so, because we need to find the higher number after the computation with the `representativePValue`.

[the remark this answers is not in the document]
alisman: but the function is constant no?  So the max (or perhaps min b/c of minus sign?) representativePValue with always win, no?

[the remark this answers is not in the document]
oplantalech: Oh, true. Let me see if I can fix it and move back the calculation to the props.

PR #1086, opened 2018-04-11
Fail gracefully in oncoprint if OncoKB fails

adamabeshouse:
image

image

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -6,7 +6,7 @@ import {
     MolecularDataMultipleStudyFilter, SampleFilter, MolecularProfileFilter, GenePanelMultipleStudyFilter, PatientFilter
 } from "shared/api/generated/CBioPortalAPI";
 import client from "shared/api/cbioportalClientInstance";
-import {computed, observable, action} from "mobx";
+import {computed, observable, action, reaction, IObservable, IObservableValue, ObservableMap} from "mobx";
 import {remoteData, addErrorHandler} from "shared/api/remoteData";
 import {labelMobxPromises, cached, MobxPromise} from "mobxpromise";
 import OncoKbEvidenceCache from "shared/cache/OncoKbEvidenceCache";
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -214,6 +214,18 @@ export class ResultsViewPageStore {
         //     this.ajaxErrors.push(error);
         // });
         this.getURL();
+
```
alisman: this seems super icky to me.  this is so we don't make further calls to it?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -201,6 +201,31 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
             }
         );
 
+        this.putativeDriverSettingsReaction = reaction(
```
alisman: as above, i think we should really try to avoid making reactions unless it's really necessary.  could this not be accomplished with a computed getters?   i think the setting should be immutable.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: thats so the oncoprint setting goes to false instead of the default true. at this point it will become impossible to change from the UI. maybe it would be a better idea to model this situation more explicitly in mutationAnnotationSettings

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: this is to maintain the relationship between distinguishDrivers and any of the mutationAnnotationSettings being set to true. I think this is a clean way to do it because the mutationAnnotationSettings can be modified from here or from the store.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1478,15 +1529,20 @@ export class ResultsViewPageStore {
         invoke: ()=>Promise.resolve(indexHotspotsData(this.hotspotData))
     });
 
-    public readonly isHotspot = remoteData({
+    public readonly isHotspot = remoteData<Error|((mutation:Mutation)=>boolean)>({
```
onursumer: Since we are not getting hotspot info from OncoKB anymore, changes related to hotspots will conflict with https://github.com/cBioPortal/cbioportal-frontend/pull/1083/files#diff-a1e9801197401cd10e07242602cf120dL1481.

PR #1092, opened 2018-04-16
Customize surival plot by passing victory options

zhx828:
# Checks
  the `-s` option)


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -30,9 +31,44 @@ export interface ISurvivalChartProps {
     xLabelWithEventTooltip: string;
     xLabelWithoutEventTooltip: string;
     fileName: string;
-    showTable: boolean;
-    showLegend: boolean;
-    showDownloadButtons: boolean;
+    showTable?: boolean;
```
alisman: @ersinciftci did you make similar changes or is it my imagination?

ersinciftci: Hongxin converted them to optional, so they are `true` now by default. In my version they were required and didn't have a default value. I think this is better, it makes sense to show everything by default.

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -45,6 +81,60 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
+        },
+        legend: {
+            x: 600,
+            y: 40
+        }
+    };
+
+    @computed
+    get styleOpts() {
+        let configurableOpts: StyleOpts = _.merge(this.styleOptsDefaultProps, this.props.styleOpts);
```
alisman: It turns out that _.merge method treats the first param as the "target."  You wouldn't think it would work this way but, the defaults object ends up getting overwritten.  So, it's a good practice to use _.merge({}, obj1, obj2).  That way obj1 doesn't get mutated.  Weird, I know.

alisman: this probably doesn't matter in this case because you probably won't use defaults again in life of component, but in another circumstance you might.

zhx828: Yeah, the default should be immutable. I will change to the suggestion above.

zhx828:
Merging this pull request since it has exact same tests failed as in rc branch. Need this component for the study view prototype demos on Monday.

PR #1094, opened 2018-04-16
Fix genomic overview when dot in sample name

inodb:
The genomic overview would show MUT for the sample if it contains a dot in the
name, because jquery assumes one is looking for an element with a certain id +
a certain class. Use `$([id="name"])` syntax instead.

See:

https://cbioportal.mskcc.org/case.do#/patient?studyId=all_phase2_target_2017&caseId=TARGET-10-PAKKCA

Try fix with:

```
localStorage.heroku = "cbioportal-frontend-pr-1094"
```


src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -89,8 +89,8 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
                 rowIndex = rowIndex + 1;
 
                 if (this.props.sampleManager.samples.length > 1) {
-                    const id = `#mutTrack${sample.id}`;
-                    const $container = $(id);
+                    const id = `mutTrack${sample.id}`;
+                    const $container = $(`[id="${id}"]`);
                     const pos = {x: parseInt($container.attr('x')) - 10, y: parseInt($container.attr('y')) - 5};
                     const $newContainer = $(`<svg id="${id}" height="12" width="12" />`);
```
inodb: i kept the only other reference to id without `#` because that seems to have been a typo

PR #1096, opened 2018-04-16
CircleCI split end to end tests

inodb:
One job for each spec file so one can retry a single spec file. Move DLL building to install step


inodb:
Doesn't do much (still takes about 9m, splitting up spec files prolly better strategy)

PR #1104, opened 2018-04-19
Disable session test iff no session service

inodb:
This will make it easier to run e2e tests on backend repo

### DIFF
--- a/end-to-end-tests/specs/screenshot.spec.js
+++ b/end-to-end-tests/specs/screenshot.spec.js
@@ -2,9 +2,11 @@ var assert = require('assert');
 var expect = require('chai').expect;
 var waitForOncoprint = require('./specUtils').waitForOncoprint;
 var goToUrlAndSetLocalStorage = require('./specUtils').goToUrlAndSetLocalStorage;
+var sessionServiceIsEnabled = require('./specUtils').sessionServiceIsEnabled;
 var assertScreenShotMatch = require('../lib/testUtils').assertScreenShotMatch;
 
 const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+
 
 function runResultsTestSuite(){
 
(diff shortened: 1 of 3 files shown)


end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -86,6 +88,7 @@ describe('result page screenshot tests', function(){
     before(function(){
         var url = `${CBIOPORTAL_URL}/index.do?tab_index=tab_visualize&cancer_study_list=coadread_tcga_pub&cancer_study_id=coadread_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=coadread_tcga_pub_nonhypermut&case_ids=&gene_list=KRAS+NRAS+BRAF&gene_set_choice=user-defined-list&Action=Submit&show_samples=false&`;
         goToUrlAndSetLocalStorage(url);
+        sessionServiceIsEnabled();
```
adamabeshouse: wrong function name?

inodb: @adamabeshouse good catch - i kinda dislike using that as a function name here because the same name is being used as a property on the frontendConfig object

src/config/IAppConfig.ts
```diff
@@ -47,6 +47,7 @@ export interface IAppConfig {
     oncoprintOncoKbHotspotsDefault?:"disable"|"custom";
     // Select hide VUS by default
     oncoprintHideVUSDefault?:boolean;
+    sessionServiceIsEnabled:boolean;
```
inodb: this is passed in https://github.com/cBioPortal/cbioportal/pull/4111 - tests are only skipped if this is `false` so they will continue to run when this is undefined in older versions of the backend

end-to-end-tests/specs/screenshot.spec.js
adamabeshouse: fair enough

PR #1110, opened 2018-04-23
Fix "not sequenced" in oncoprint

adamabeshouse:
Update oncoprint to use new system of coverage information.

image

image

image

image

non-screenshot E2E tests passing on local using `cbioportal/master`


jjgao:
@adamabeshouse thanks. Looks good. Maybe also highlight profiles that were not profiled in the tooltip.

pieterlukasse:
I am interested in reviewing this. @adamabeshouse can you please add some screenshots of before vs after?

pieterlukasse:
@jjgao should we consider a sample with alteration data to be "profiled", irrespective of whether it has panel information or not?

end-to-end-tests/specs/home.spec.js
```diff
@@ -23,7 +23,7 @@ describe('homepage', function() {
         });
     }
 
-    it('it should have 27 (small test db), 29 (public test db) or 31 studies (production) in list', function () {
+    it.only('it should have 27 (small test db), 29 (public test db) or 31 studies (production) in list', function () {
```
inodb: let's revert this before merging

adamabeshouse: good catch

jjgao:
@pieterlukasse I was thinking of that too. But since we are pull samples from sample_profile table, we should cover all samples in a profile already.

jjgao:
@pieterlukasse `_sequenced` case list was used for mutation for now. I have asked Ben to fix that so all sequenced samples will be in `sample_profile` table.

AFAIR, data in `sample_profile` is complete for other types of data including cna.

pieterlukasse:
@jjgao @n1zea144 can you please point to the PR where the `_sequenced` change is made? This PR would have to ensure the `_sequenced` case list input file is used to populate the `sample_profile` table, right. This would mean nothing has to change in the validation scripts, correct?

PR #1119, opened 2018-04-25
Refac plots

adamabeshouse:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+        this.searchCaseTimeout = setTimeout(()=>this.executeSearchCase(this.searchCaseInput), searchInputTimeoutMs);
+    }
+
+    private setSearchMutationInput(e:any) {
+        this.searchMutationInput = e.target.value;
+        clearTimeout(this.searchMutationTimeout);
+        this.searchMutationTimeout = setTimeout(()=>this.executeSearchMutation(this.searchMutationInput), searchInputTimeoutMs);
+    }
+
+    private executeSearchCase(caseId:string) {
```
alisman: i think these should all be actions since they are setting an observable

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+    private onVerticalAxisGeneSelect(option:any) {
+        this.onGeneSelect(true, option);
+    }
+
+    private onHorizontalAxisGeneSelect(option:any) {
+        this.onGeneSelect(false, option);
+    }
+
+    @computed get geneOptions() {
+        if (this.props.store.genes.isComplete) {
```
alisman: i think it would be better to check load state in view rather than in computed.   that way status isn't hidden.  components which consume this data should handle loadind status

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+    @computed get cnaPromise() {
+        if (this.viewType === ViewType.CopyNumber && this.props.store.studyToMolecularProfileDiscrete) {
+            if (!this.sameGeneInBothAxes) {
+                throw ("Plots tab invariant error: viewType is CopyNumber must mean that both axes are molecularProfile and have the same gene");
+            }
+            const entrezGeneId = this.horzSelection.entrezGeneId!;
+            const queries = _.values(this.props.store.studyToMolecularProfileDiscrete.result)
+                            .map(p=>({molecularProfileId: p.molecularProfileId, entrezGeneId}));
+            const promises = this.props.store.numericGeneMolecularDataCache.getAll(queries);
+            return remoteData({
```
alisman: could this whole @computed by represented as a remoteData in the usual fashion?   when cnaPromise recomputes, I don't think that the remoteData of prior computations will every be garbage collected

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                this.props.store.molecularProfileIdToMolecularProfile.result,
+                this.props.store.entrezGeneIdToGene.result,
+                this.clinicalAttributeIdToClinicalAttribute
+            );
+        } else {
+            return "";
+        }
+    }
+
+    @computed get horzDescription() {
```
alisman: i'm sort of torn about this, but i don't think we should be checking status of promises inside computed getters.   it hides the status of the promises from consuming code.   we need to make sure that loading status is handled properly at the view level and this pattern will let devs off that hook.  that is, components do not render and invoke these "getters" until they know the data is loaded.   instead of this, i would advocate we use pure functions which accept data as arguments.  in order to get the optimization of @computed, we can use @memoize which caches the result of functions based on arguments.  i.e. will not recompute result if it's done so already for equal arguments.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                                        onChange={vertical ? this.onVerticalAxisProfileIdSelect : this.onHorizontalAxisProfileIdSelect}
+                                        options={this.profileNameOptionsByType[axisSelection.molecularProfileType+""] || []}
+                                        clearable={false}
+                                        searchable={false}
+                                    />
+                                </div>
+                                <InfoIcon
+                                    tooltip={<span>{vertical ? this.vertDescription : this.horzDescription}</span>}
+                                    tooltipPlacement="right"
+                                    style={{marginTop:"0.9em", marginLeft:5}}
```
alisman: is there a reason units em?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                <div style={{margin:5, padding:10, border: "1px solid #aaaaaa", borderRadius:4}}>
+                    <Observer>
+                        {this.getUtilitiesMenu}
+                    </Observer>
+                </div>
+            </div>
+        );
+    }
+
+    readonly plotType = remoteData({
```
alisman: we shouldn't have remoteData on a view component.  the reason is that view components can and will be reinstantiated.   remoteData the are properties of component class will then be duplicated, with old ones, I believe never getting cleaned up.     Stores on the other should only be instantiated once (we used to break this rule but in SPA, we have to obey it).

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: i guess theres no rational justification except it feels more semantic to me bc its sort of related to line height

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+
+        this.geneLock = true;
+        this.searchCaseInput = "";
+        this.searchMutationInput = "";
+        this.viewType = ViewType.MutationType;
+
+        this.controls = this.controls.bind(this);
+        this.plot = this.plot.bind(this);
+        this.onInputClick = this.onInputClick.bind(this);
+        this.getHorizontalAxisMenu = this.getHorizontalAxisMenu.bind(this);
```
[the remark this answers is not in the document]
adamabeshouse: done

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+    }
+
+    @computed get downloadFilename() {
+        return "plot"; // todo: more specific?
+    }
+
+    private initAxisMenuSelection():AxisMenuSelection {
+        const self = this;
+
+        return observable({
```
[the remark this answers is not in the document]
adamabeshouse: well, i guess it could be. maybe more like an imported utility function. because it needs reference to the PlotsTab itself

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: done

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+        switch (this.viewType) {
+            case ViewType.MutationType:
+                return (d:IScatterPlotSampleData)=>this.scatterPlotAppearance(d).symbol!;
+            case ViewType.CopyNumber:
+            case ViewType.MutationSummary:
+                return "circle";
+        }
+    }
+
+    private getAxisMenu(vertical:boolean) {
```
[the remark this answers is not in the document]
adamabeshouse: this is a helper function for getters which are used in `<Observer>`

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                                    <ReactSelect
+                                        name={`${vertical ? "v" : "h"}-gene-selector`}
+                                        value={axisSelection.entrezGeneId}
+                                        onChange={vertical ? this.onVerticalAxisGeneSelect : this.onHorizontalAxisGeneSelect}
+                                        options={this.geneOptions}
+                                        clearable={false}
+                                        searchable={false}
+                                    />
+                                </div>
+                                <div style={{marginLeft:7, display:"inline"}}>
```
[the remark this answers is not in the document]
adamabeshouse: I think I had that and something was going wrong with the layout

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                                    <DefaultTooltip
+                                        placement="right"
+                                        overlay={<span>{this.geneLock ? "Gene selection synchronized in both axes" : "Gene selections independent in each axis"}</span>}
+                                    >
+                                        <div>
```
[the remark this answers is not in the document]
adamabeshouse: yes it is because of some issue with `<DefaultTooltip>` placement

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -0,0 +1,1110 @@
...
+                                onChange={vertical ? this.onVerticalAxisProfileTypeSelect : this.onHorizontalAxisProfileTypeSelect}
+                                options={this.profileTypeOptions}
+                                clearable={false}
+                                searchable={false}
+                            />
+                        </div>
+                        <div>
+                            Profile Name
+                            <div style={{display:"flex", flexDirection:"row"}}>
+                                <div style={{width:"93%"}}>
```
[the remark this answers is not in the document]
adamabeshouse: otherwise flex makes it collapse. open to flex wisdom that would do this better

src/pages/resultsView/plots/PlotsTab.tsx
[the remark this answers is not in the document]
alisman: it seems like a slightly weird pattern to me.   i don't think it should reference the instance of PlotTab via closure (self).

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: done

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: in this particular case i think it actually makes sense because the return type is just a string for a tooltip (and in `horzLabel` and `vertLabel` its the string for the axis labels), and we wouldn't be showing a loading spinner anyway

alisman: k

src/pages/resultsView/plots/PlotsTab.tsx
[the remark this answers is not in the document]
adamabeshouse: why not?

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: i think we already decided that in general it's not pleasing to have to separate things into stores? my opinion is it's better to start explicitly disposing once we do SPA

PR #1121, opened 2018-04-26
Revert "using Genome Nexus hotspot data for oncoprint instead of OncoKB"

inodb:
This reverts commit 7c3ccfcd5845ffa33d45e21d2373c4fd7f28189c


inodb:
@onursumer fixed ts error, thanks

PR #1124, opened 2018-04-30
Change the study view to a new route; Add tabs; Add study description

zhx828:

### DIFF
--- a/my-index.ejs
+++ b/my-index.ejs
@@ -5,7 +5,7 @@
     <title><%= htmlWebpackPlugin.options.title %></title>
     <script>
         window.devContext = true;
-        window.defaultRoute = "/study";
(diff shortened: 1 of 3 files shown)


my-index.ejs
```diff
@@ -5,7 +5,7 @@
     <title><%= htmlWebpackPlugin.options.title %></title>
     <script>
         window.devContext = true;
-        window.defaultRoute = "/study";
+        window.defaultRoute = "/newstudy";
```
kalletlak: We could still use the same `/study` and change it at servlet (jsp) side

kalletlak: We need to add new servlet `newstudy` and `study` route in cbioportal-fronted for the above url to work

PR #1128, opened 2018-04-30
Fix errors in release-1.13.0

adamabeshouse:


inodb:
@adamabeshouse just merging this to get rid of compile errors

inodb:
let's look at e2e test today

PR #1133, opened 2018-05-02
Down sampling if survival curve has more than 1000 samples; Allow zoom in

zhx828:

## Before down sampling:
before-down-sampling

## After down sampling
after-down-sampling

## Zoom in
zoom in

## Zoom out
zoom out

## Drag
may-03-2018 15-56-44


zhx828:
@alisman @ersinciftci there is a performance issue with the method `onZoomDomainChange` I tried to add a setTimeout to prevent multiple updates within 500ms, but it didn't work. Any thoughts?

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -128,9 +135,67 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
+            }
+        };
+        return a;
+    }
+
+    // Only recalculate the scatter data based on the plot filter.
+    // The filter is only available when user zooms in the plot.
+    @computed
+    get scatterData() {
+        let filteredData = _.cloneDeep(this.allScatterData);
```
ersinciftci: Can you move this whole calculation to SurvivalUtil, and also add unit tests for new methods in SurvivalUtil?

alisman: could you do all three of these iterations (cloning, forEach1 and forEach2) with one single map?     would be a little more semantic too.

zhx828: I can group two forEach together. Not sure what's the best way to have lodash cloning method in.

zhx828: Sure.

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -128,9 +135,40 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
         let configurableOpts: StyleOpts = _.merge({}, this.styleOptsDefaultProps, this.props.styleOpts);
         configurableOpts.padding.right = this.props.showLegend ? 300 : configurableOpts.padding.right;
         configurableOpts.legend.x = configurableOpts.width - configurableOpts.padding.right;
+        this.downSamplingDenominatorX = configurableOpts.width - configurableOpts.padding.left - configurableOpts.padding.right;
```
alisman: computed getters should be pure functions and should not have sideffects (setting instance properties).   it looks like these denomintator props should be computed properties.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -152,4 +184,87 @@ export function convertScatterDataToDownloadData(patientData: any[]): any[] {
...
+}
+
+/**
+ * Down sampling based on the hypotenuse difference.
+ * The cross dot is guaranteed to be drawn when hidden dots are taking into account.
+ * @param {DownSampling} opts
+ * @returns {ScatterData[]}
+ */
+export function downSampling(data: ScatterData[], opts: DownSamplingOpts): ScatterData[] {
+    let x = data.map(item => item.x);
```
alisman: use _.maxBy to avoid doing these initial maps, which are unnecessary

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -152,4 +184,87 @@ export function convertScatterDataToDownloadData(patientData: any[]): any[] {
...
+
+    let yMax = _.max(y) || 0;
+    let yMin = _.min(y) || 0;
+
+    if (opts.xDenominator <= 0 || opts.yDenominator <= 0)
+        return data;
+
+    let timeThreshold = (xMax - xMin) / opts.xDenominator;
+    let survivalRateThreshold = (yMax - yMin) / opts.yDenominator;
+    let averageThreshold = Math.hypot(timeThreshold, survivalRateThreshold);
```
alisman: I just checked and looks like Math.hypot is not supported in IE11.  May see if jStat has it.  Or install a polyfill.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -152,4 +184,87 @@ export function convertScatterDataToDownloadData(patientData: any[]): any[] {
...
+    });
+}
+
+export function filteringScatterData(allScatterData: GroupedScatterData, filters: SurvivalPlotFilters | undefined, downSamplingOpts: DownSamplingOpts):GroupedScatterData {
+    let filteredData = _.cloneDeep(allScatterData);
+    _.forEach(filteredData, (value:SurvivalCurveData, key:string) => {
+        if (value.numOfCases > downSamplingOpts.threshold) {
+            if (filters) {
+                let filteredIndex: number[] = [];
+                _.forEach(value.scatter, (_val, _key) => {
```
alisman: i think you could do this with a call to reduce that would avoid double iterations (forEach and then filter). and would also be more semantic because you just get assignement  (e.g. result = arr.reduce() )

alisman: in places where we are probably dealing with large amounts of data, we should use the native api instead of lodash.  you can use arr.forEach, arr.filter and even arr.includes.    (Arr.includes might have typing issue)

zhx828: I was trying to reduce the unnecessary loops since this index will be used to get data out of both `.scatter` and `.scatterWithOpacity`. But seems I did it wrong, my intention is using indices to pick up the data instead of `.filter`. But this may not be necessary... after second thought, if most of the dots are qualified the filter, the performance would be worse compared to only two loops. I will change it back.

PR #1134, opened 2018-05-02
Results page session screenshots separate files

inodb:
Before they were using the same file, so it's impossible to inspect the
mismatch when the first one fails.


inodb:
@adamabeshouse there's a test here: `msk_impact_2017 with SOS1 with CNA profile - SOS1 should be sequenced`. The reference screenshot was never included, it shows up as not profiled. I think SOS1 should actually show up as not profiled in this case right. Since the gene panel for copy number is now the same as for simple somatic mutations

end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -169,6 +169,8 @@ describe('study view screenshot test', function(){
 
         // give charts time to render
         browser.setViewportSize({ height: 1600, width: 1000 })
+        // move mouse somewhere it doesn't show active shart icons
```
adamabeshouse: *chart xD

inodb: hahah

PR #1141, opened 2018-05-02
Use localdist in end to end tests

inodb:
- use localdist=true in URL to serve frontend artifacts during e2e tests. This serves the artifacts from dist/ folder when running e2e using http-server instead of `npm run start`. This makes the testing more like production environment (we never serve artifacts through `npm run start`). This functionality can be turned on by setting env parameter `export FRONTEND_TEST_USE_LOCAL_DIST=true`. Otherwise it will use localdev=true.
- move building to install job, since both unit and end_to_end_tests require
  build artifacts
- several screenshots updated. It seems that styling is slighlty different when
  serving dist artifacts, specifically the borders of tables
- update to new version of screenshot regression testing and disable anti aliasing mismatches with `ignoreComparison`


.circleci/config.yml
```diff
@@ -36,6 +36,15 @@ jobs:
...
             - node_modules
           key: v1-dependencies-{{ checksum "package-lock.json" }}
+      - run:
+          name: "Run build"
+          command: 'npm run build'
+      - save_cache:
+          paths:
+            - common-dist
+            - dist
+          key: asset-cache-{{ .Environment.CIRCLE_SHA1 }}
```
inodb: store `common-dist` and `dist` so other jobs in the workflow can just load it from the cache

.circleci/config.yml
```diff
@@ -79,13 +91,16 @@ jobs:
...
           # fallback to using the latest cache if no exact match is found
           - v1-dependencies-
+      - restore_cache:
+          keys:
+          - asset-cache-{{ .Environment.CIRCLE_SHA1 }}
       - run:
-          name: "Build DLL files"
-          command: "npm run buildDLL:dev"
+          name: "Install http server"
+          command: "npm i http-server"
```
inodb: the http server used to serve the frontend artifacts in `dist/` folder

.circleci/config.yml
```diff
@@ -99,11 +114,12 @@ jobs:
             curl $CBIOPORTAL_URL > /dev/null && \
             sleep 20s && \
             curl http://localhost:3000 > /dev/null && \
-            sleep 20s && \
+            sleep 1s && \
```
inodb: no need to sleep this long now since we don't have to wait for frontend artifacts to be build as is the case with `npm run start`

.circleci/config.yml
```diff
@@ -99,11 +114,12 @@ jobs:
...
             cd end-to-end-tests && \
             npm run test-webdriver-manager
           environment:
             JUNIT_REPORT_PATH: ./junit/
+            FRONTEND_TEST_USE_LOCAL_DIST: true
```
inodb: enable local dist testing in e2e test

end-to-end-tests/package-lock.json
```diff
@@ -1917,17 +2164,18 @@
       }
     },
     "wdio-visual-regression-service": {
-      "version": "0.8.0",
-      "resolved": "https://registry.npmjs.org/wdio-visual-regression-service/-/wdio-visual-regression-service-0.8.0.tgz",
-      "integrity": "sha1-3OhPm/B6B4O1DCAN94M0b89HVaM=",
+      "version": "0.9.0",
```
inodb: updated visual regression service for enabling `ignoreComparison` antialiasing

end-to-end-tests/specs/results.spec.js
```diff
@@ -24,7 +21,7 @@ describe('Results Page', function() {
         describe('single study query with four genes', ()=>{
             before(()=>{
                 var url = `${CBIOPORTAL_URL}/index.do?tab_index=tab_visualize&cancer_study_list=coadread_tcga_pub&cancer_study_id=coadread_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=coadread_tcga_pub_nonhypermut&case_ids=&gene_list=BRAF+KRAS+NRAS&gene_set_choice=user-defined-list&Action=Submit#pancancer_study_summary`;
-                browser.url(url);
+                goToUrlAndSetLocalStorage(url);
```
inodb: removed all references to `browser.url`, always use `goToUrlAndSetLocalStorage`. There's no extra cost to this now, because we no longer do a refresh

src/appBootstrapper.jsx
```diff
@@ -17,10 +17,12 @@ import AppConfig from "appConfig";
 import browser from 'bowser';
 import './shared/lib/tracking';
 
-if (localStorage.localdev === 'true') {
+if (localStorage.localdev === 'true' || localStorage.localdist === 'true') {
     __webpack_public_path__ = "//localhost:3000/"
+    localStorage.setItem("e2etest", "true");
```
inodb: set e2etest to true for devs (when testing google analytics, one can always disable this manually here)

PR #1099, opened 2018-04-18
Parse ‘merged track’ syntax in OQL queries

fedde-s:
As the first step towards the ‘merged track’ feature discussed in cBioPortal RFC42, make the OQL parser support the syntax to query for them: square brackets around some of the lines/genes in the query, optionally including a label between double quotes.

```
TP53 [MDM2 MDM4]
["CDK INHIBITORS" CDKN2A CDKN2B]
TP53BP1
```

All results view tabs still receive a flat list of genes as of this pull request, as they do not yet have a use for the grouping, and most never will.

Notable detail: to control the alteration filter of all genes in a merged track, any `DATATYPES` commands within a merged track expression only affect genes inside the expression (up to the closing bracket), such that this query:
```
DATATYPES: AMP
[TP53
DATATYPES: GAIN
SOX9
BRCA2]
MDM2
```
is effectively equivalent to this one:
```
[TP53: AMP
SOX9: GAIN
BRCA2: GAIN]
MDM2: AMP
```

The new syntax is not yet documented on the OQL page, as it has a negligible effect on the output for now.

PR #1143, opened 2018-05-03
Render merged track queries as single Oncoprint tracks

fedde-s:
Recognising the OQL syntax introduced #1099, render each merged track query as a single Oncoprint track that aggregates the alterations for the genes in the same way they're aggregated for patients with multiple samples. Cover several of the functions that gain additional behaviour with unit tests, correct the documentation for the function that filters data by query statement, and refactor the interface of the function that performs the cell-wise aggregation for clarity.

## Screenshot ##
Partial screenshot of an Oncoprint showing a track labelled ‘CDK INHIBITORS’ that displays alterations for both CDKN2A and CDKN2B (explained by the OQL in its tooltip), as well as a track labelled ‘MDM2 / MDM4’

### DIFF
--- a/src/shared/components/oncoprint/DataUtils.ts
+++ b/src/shared/components/oncoprint/DataUtils.ts
@@ -98,22 +98,11 @@ export function selectDisplayValue(counts:{[value:string]:number}, priority:{[va
     }
 };
 
-type FilledGeneticTrackDatum<T> = T & Pick<GeneticTrackDatum, (
-    'gene'
-    | 'data'
-    | 'disp_fusion'
-    | 'disp_cna'
-    | 'disp_mrna'
-    | 'disp_prot'
-    | 'disp_mut'
-    | 'disp_germ'
-)>;
-export function fillGeneticTrackDatum<T>(
-    oldDatum: T,
+export function fillGeneticTrackDatum(
+    newDatum:Partial<GeneticTrackDatum>,
     hugoGeneSymbol:string,
     data:AnnotatedExtendedAlteration[]
-): FilledGeneticTrackDatum<T> {
-    const newDatum: Partial<FilledGeneticTrackDatum<T>> = oldDatum;
+):GeneticTrackDatum {
     newDatum.gene = hugoGeneSymbol;
     newDatum.data = data;
 
@@ -171,11 +160,10 @@ export function fillGeneticTrackDatum<T>(
     newDatum.disp_cna = selectDisplayValue(dispCnaCounts, cnaRenderPriority);
     newDatum.disp_mrna = selectDisplayValue(dispMrnaCounts, mrnaRenderPriority);
     newDatum.disp_prot = selectDisplayValue(dispProtCounts, protRenderPriority);
-    const disp_mut = selectDisplayValue(dispMutCounts, mutRenderPriority);
-    newDatum.disp_mut = disp_mut;
-    newDatum.disp_germ = disp_mut ? dispGermline[disp_mut] : undefined;
+    newDatum.disp_mut = selectDisplayValue(dispMutCounts, mutRenderPriority);
+    newDatum.disp_germ = newDatum.disp_mut ? dispGermline[newDatum.disp_mut] : undefined;
 
-    return newDatum as FilledGeneticTrackDatum<T>;
+    return newDatum as GeneticTrackDatum; // return for convenience, even though changes made in place
 }
 
 export function makeGeneticTrackData(
--- a/src/shared/components/oncoprint/DataUtils.ts
+++ b/src/shared/components/oncoprint/DataUtils.ts
@@ -98,11 +98,22 @@ export function selectDisplayValue(counts:{[value:string]:number}, priority:{[va
     }
 };
 
-export function fillGeneticTrackDatum(
-    newDatum:Partial<GeneticTrackDatum>,
+type FilledGeneticTrackDatum<T> = T & Pick<GeneticTrackDatum, (
+    'gene'
+    | 'data'
+    | 'disp_fusion'
+    | 'disp_cna'
+    | 'disp_mrna'
+    | 'disp_prot'
+    | 'disp_mut'
+    | 'disp_germ'
+)>;
+export function fillGeneticTrackDatum<T>(
+    oldDatum: T,
     hugoGeneSymbol:string,
     data:AnnotatedExtendedAlteration[]
-):GeneticTrackDatum {
+): FilledGeneticTrackDatum<T> {
+    const newDatum: Partial<FilledGeneticTrackDatum<T>> = oldDatum;
     newDatum.gene = hugoGeneSymbol;
     newDatum.data = data;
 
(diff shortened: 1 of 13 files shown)


src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -98,11 +98,22 @@ export function selectDisplayValue(counts:{[value:string]:number}, priority:{[va
...
+    | 'disp_mut'
+    | 'disp_germ'
+)>;
+export function fillGeneticTrackDatum<T>(
+    oldDatum: T,
     hugoGeneSymbol:string,
     data:AnnotatedExtendedAlteration[]
-):GeneticTrackDatum {
+): FilledGeneticTrackDatum<T> {
+    const newDatum: Partial<FilledGeneticTrackDatum<T>> = oldDatum;
```
adamabeshouse: why are you doing this, why not just keep the parameter name and type it Partial<FilledGeneticTrackDatum<T>> instead of T

adamabeshouse:
still looking, but one note: it doesnt work with multiple study queries because legacy oql-parser doesnt support merged queries. Just try it and you'll see. Some backend changes may be required

src/shared/components/oncoprint/DataUtils.ts
adamabeshouse: in general I don't understand the point of what you've done here with FilledGeneticTrackDatum and the generic type T. can you explain more about that?

fedde-s: I made the type of the function's interface more explicit for clarity, but that might not be very effective if understanding it requires questions 😄  What I found (and documented) that this function does is take any object and return something that has whatever properties that object had, plus several specific `GeneticTrackDatum` properties. It doesn't depend on any of the input object's properties, and can't claim to return a complete `GeneticTrackDatum` if given a partial one (in which all of the properties may be missing). It *can* claim to add its list of properties to whatever object you pass to it though.

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -199,18 +212,25 @@ export function makeGeneticTrackData(
...
+                geneSymbolArray,
+                hugoGeneSymbol => sampleSequencingInfo.notProfiledByGene[hugoGeneSymbol] || []
+            );
             newDatum.not_profiled_in = newDatum.not_profiled_in.concat(sampleSequencingInfo.notProfiledAllGenes);
 
-            fillGeneticTrackDatum(
-                newDatum, hugoGeneSymbol,
+            _.assign(newDatum, fillGeneticTrackDatum(
+                {},
+                geneSymbolArray.join(' / '),
```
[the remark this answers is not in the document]
fedde-s: I tried to keep the impact of the change small, but in effect I changed the semantics of all downstream code by assuming that the ‘hugoGeneSymbol’ field holds a label to be used verbatim in views rather than an actual HGNC-approved gene symbol to integrate with web APIs. It would make sense to either change the name of the attribute to make that explicit, or, if the code downstream of this object isn't strictly limited to view logic, adjust all of that code to expect an array. Do you think the latter choice is the right one then? Note that actual alterations in the `data` property still list the details of the genes they were measured in.

fedde-s:
I believe that the legacy parser was recently phased out in favour of a global variable named `frontendVars`, which provides access to the updated one via the updated `oql-filter.js` functions. The developers of the multi-study query feature must have been unaware of that. Do you know who they are, so that we can ask them to fix it?

src/shared/components/oncoprint/DataUtils.ts
adamabeshouse: I think that's being too pedantic.. typescript has certain limitations of expressiveness and inference that I'm getting around by using `Partial` for the parameter, and then casting at the end.

src/shared/components/oncoprint/DataUtils.ts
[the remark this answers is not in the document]
adamabeshouse: I see your point. I think the `gene` property of `GeneticTrackDatum` is only used in the tooltip. So I think it would be a good idea to leave the code you've written but change the attribute name to `trackLabel`

adamabeshouse:
I *believe* that's @alisman ?

src/shared/components/oncoprint/DataUtils.ts
fedde-s: I was able to express it in a way that helped me understand the function without too much effort, but if you feel like it will confuse others I can leave this commit out; it's otherwise unrelated to the merged track feature, after all.

fedde-s:
I found the line that was using the old parser, and proposed a fix in cBioPortal/cbioportal#4187; I can now query for merged tracks in merged studies.

src/shared/components/oncoprint/DataUtils.ts
adamabeshouse: yes please take it out, i think it adds unnecessary complexity

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -102,8 +102,8 @@ export function fillGeneticTrackDatum(
     newDatum:Partial<GeneticTrackDatum>,
     hugoGeneSymbol:string,
     data:AnnotatedExtendedAlteration[]
-):GeneticTrackDatum {
-    newDatum.gene = hugoGeneSymbol;
+): Partial<GeneticTrackDatum> {
```
adamabeshouse: no this is what we were talking about, just please leave as is, and leave the cast on line 166

fedde-s: This return type correction is here to catch mismatched usage of the new `trackLabel` property in the various pieces of code and test that use the function. I already removed the complicated dependent type you objected to.

fedde-s: Listing an incorrect return type of `GeneticTrackDatum` complicates that, as pretty much all code that references the function would have to disable type checking.

adamabeshouse: I don't understand what you mean with the mismatched `trackLabel` usage?

Instead of changing the return type I'd rather change the argument type of newDatum to be more precise, because there's nothing imprecise about it - this function does return a complete GeneticTrackDatum

fedde-s: Makes sense—that would make the usage clearer as well, as `makeGeneticTrackData()` could just use the return value without fighting against the type checker. I was going off the use-cases demonstrated in the tests, which all fed it an empty object.

adamabeshouse: makes sense. thanks :+1:

adamabeshouse: if you want, could you also change the argument type of `newDatum`?

fedde-s: I'm not actually sure if that would be an improvement. On the one hand it would point out missing required inputs to the `fillGeneticTrackDatum()` function when using it, but on the other hand I don't know of a way to express it without making the function implementation fragile to any unrelated changes in the `GeneticTrackDatum` type. So for me, it's clear enough as-is now that I've adjusted the tests to document the intended use-case.

adamabeshouse: okay, sounds good, thanks

adamabeshouse:
@fedde-s could you just please rebase onto rc

fedde-s:
Rebased on rc, @adamabeshouse! I could try to squash the six back-and-forth commits of clarifying the signature of `fillGeneticTrackDatum` into one, but merging it like this is fine with me too as long as there's a merge commit for the PR.

adamabeshouse:
I think its fine as is 

but do you know if the failing e2e tests are related to this?

fedde-s:
I don't expect so; they're failing everywhere, and have been most of the time for the past month. I believe @inodb is tracking the progress making the test suites usable again, and the last thing I heard was to ignore them for now.

fedde-s:
Looks like someone introduced a conflict. I should rebase and see what it's about.

adamabeshouse:
@fedde-s thanks fedde, rebasing should also clear up the failing tests since we merged fixes to rc yesterday

fedde-s:
The `rc` branch contains a type error, and somehow lists all unit tests twice in the terminal output. It seems very unreliable even if all the tests are have a green tick mark (twice).

fedde-s:
Rebased on `rc`. The unit test output is a bit difficult to read while all lines are printed twice in different colours, but they do pass apart from the pre-existing error mentioned above (which is being addressed in other channels).

fedde-s:
Rebased again, as @alisman asked via Slack. Test output is still duplicated and still includes the upstream error message, but I believe this to be unrelated to the feature introduced here. @alisman said I could merge this now, so that I can present the next feature that depends on this.

PR #1147, opened 2018-05-07
Revert back pie-chart filtering and add chart header

kalletlak:


src/pages/studyView/chartHeader/styles.scss
```diff
@@ -0,0 +1,69 @@
...
+    }
+    >button:first-child {
+        margin-left: 10px !important;
+    }
+    .fa-arrows:hover {
+        cursor: move;
+    }
+}
+
+main {
```
alisman: remember that these are global, meaning e.g. "main" will apply outside of studyview.     just because it's inside a file specific to study view doesn't scope it.  you should scope this entire file by wrapping it in an outer class .study-view or some such.

alisman: also, main is not descriptive enough.

src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -1,152 +1,87 @@
...
 export class PieChart extends React.Component<IPieChartProps, {}> {
 
+    private colorSet:{[id:string]:string} = {};
 
-  constructor(props: IPieChartProps) {
-    super(props);
-  }
+    constructor(props: IPieChartProps) {
+        super(props);
+        this.colorSet = getPieSliceColors(props.data.counts);
```
alisman: if props.data.counts is observable, then I it would seem colorSet should be @computed, no?

src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -1,152 +1,87 @@
...
-              this.props.onUserSelection(props.datum.x);
+    private events = [{
+        target: "data",
+        eventHandlers: {
+        onClick: () => {
+            return [
+            {
+                target: "data",
+                mutation: (props:any) => {
+                    this.props.onUserSelection(props.datum.x);
```
alisman: somehow doesn't seem quite right to refer to "this" inside of a property assignment.  It might work, but I would instead make events @computed getter.    also, remember that in theory the onUserSelection property can change, so you would want this to be @computed for that reason also (so that it is reactive)

src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -1,152 +1,87 @@
...
+        }
+    }];
 
-  @computed get totalCount(){
-    return this.props.data.counts.reduce((acc:number,next)=>{
-      return acc+next.count
-    },0)
-  }
+    @computed get totalCount(){
+        return this.props.data.counts.reduce((acc:number,next)=>{
```
alisman: use lodash _.sum

src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -1,152 +1,87 @@
...
-          style={{
-            data: { fillOpacity: 0.9 },
-            labels: { fill: "white" }
-          }}
-        />
-    );
-  }
+    public render() {
+        return (
+        <div className="study-view-pie">
```
alisman: i think you could put the class on the VictoryContainer itself, which has a property className.

src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -1,152 +1,87 @@
...
+            <VictoryPie
```
alisman: we should be using our custom theme for all victory charts.  see cancer type summary for example.

src/shared/components/defaultTooltip/DefaultTooltip.tsx
```diff
@@ -1,10 +1,10 @@
 import * as React from 'react';
-import Tooltip from 'rc-tooltip';
+import Tooltip, {RCTooltip} from 'rc-tooltip';
```
alisman: one would think that Props would be on Tooltip?

src/pages/studyView/charts/pieChart/PieChart.tsx
kalletlak: `colorSet` should be calculated initially when the chart is loaded. We need to have the same slice colors even after applying filters.

src/shared/components/defaultTooltip/DefaultTooltip.tsx
kalletlak: had to `rc-tooltip` library to latest version, which has Props in RCTooltip

kalletlak:
@zhx828 instead of debounce I prefer block selection until the charts are re-loaded, may be some loading icon on each of those charts

ersinciftci:
@kalletlak I agree, the user shouldn't be able to make a new selection until every API call from the old selection is finished and every chart is updated accordingly. So we wouldn't need debounce in that case.

zhx828:
@kalletlak @ersinciftci If I, for instance, are interested in both stage I and stage II. I don't want to see the results with the only stage I applied. Why do I need to wait for the page is fully loaded after select stage I to select stage II?

ersinciftci:
@zhx828 but that means delaying the request of the users who are only interested in one slice, it would make the page even less responsive.

ersinciftci:
@zhx828 it might be a good idea to have debounce for big studies, but let's add it after we do some performance and user experience tests.

zhx828:
@ersinciftci sounds good to me

src/pages/studyView/StudyViewUtils.ts
```diff
@@ -0,0 +1,116 @@
...
+                return acc;
+            }, {});
+        });
+}
+
+export function getClinicalDataType(clinicalAttribute: ClinicalAttribute){
+    return clinicalAttribute.patientAttribute?ClinicalDataType.PATIENT:ClinicalDataType.SAMPLE;
+}
+
+export function getPieSliceColors(data: ClinicalDataCount[]):{[id:string]:string}{
```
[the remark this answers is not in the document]
kalletlak: This method is invoked initially when the chart is loaded (without any filters) and result is saved in the component. For the subsequent filtered data this saved result is used. This ensures slice color remains same

PR #1174, opened 2018-05-15
Clean up study view styles; hidden nonfunctional features

zhx828:


src/pages/studyView/charts/pieChart/styles.scss
```diff
@@ -1,5 +0,0 @@
-.study-view-pie{
```
kalletlak: Any specific reason for not including it?

PR #1175, opened 2018-05-15
Fixed broken CN segments file

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4220.

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/cnSegments/CNSegments.tsx
+++ b/src/pages/resultsView/cnSegments/CNSegments.tsx
@@ -3,8 +3,10 @@ import $ from 'jquery';
 import {observer} from "mobx-react";
 import {MSKTabs, MSKTab} from "../../../shared/components/MSKTabs/MSKTabs";
 import {Gene, Sample} from "../../../shared/api/generated/CBioPortalAPI";
+import {unescapeTabDelimited, unquote} from "shared/lib/StringUtils";
 import {ResultsViewPageStore} from "../ResultsViewPageStore";
 import {observable} from "mobx";
+import fileDownload from 'react-file-download';
 
 interface CNSegmentsIframeProps {
     sampleIds:string[];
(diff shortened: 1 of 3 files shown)


src/shared/lib/StringUtils.ts
```diff
@@ -29,4 +29,31 @@ export function stringListToIndexSet(alos:ReadonlyArray<string>):{[s:string]:num
...
+}
+
+/**
+ * Unquote a string wrapped with single or double quotes.
+ *
+ * @param {string} str  input string wrapped with single or double quotes
+ */
+export function unquote(str: string): string
+{
+    if ((str.startsWith('"') && str.endsWith('"')) ||
```
alisman: startsWith/endsWith not supported in IE11.  Possibly already polyfilled by babel but please check to make sure.

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -46,32 +48,32 @@ class CNASegmentIframe extends React.Component<CNSegmentsIframeProps,{}>{
...
-                    //triger download
-                    a.click();
-                    setTimeout(function(){
-                       $(a).remove();
-                    },200);
+                    const reader = new FileReader();
+
+                    reader.addEventListener('loadend', (e: any) => {
+                        // seems like the downloaded string is wrapped with quotes,
+                        // also it contains "\\n", "\\t", instead of "\n", and "\t"
```
alisman: where is the data coming from?  i'm just curious why it's escaped like this.   would be good to include that in comment.   if we don't control it, then indicate that.

src/shared/lib/StringUtils.ts
onursumer: good point

src/pages/resultsView/cnSegments/CNSegments.tsx
onursumer: Seems like it is coming from `api-legacy/segmentfile` endpoint, but I haven't really checked the actual code. It might be possible to fix the problem there.

onursumer: Here is the legacy API code: https://github.com/cBioPortal/cbioportal/blob/94273890828889b775ef0b7d698ab617c72edf0a/business/src/main/java/org/mskcc/cbio/portal/service/CNSegmentServiceImpl.java#L29

Not sure why we end up with a wrapped and escaped string in the end.

PR #1176, opened 2018-05-16
Fix missing Modify button

kalletlak:
Fixes https://github.com/cBioPortal/cbioportal/issues/4219

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -64,11 +64,10 @@ import {
     annotateMolecularDatum, getOncoKbOncogenic,
     computeCustomDriverAnnotationReport, computePutativeDriverAnnotatedMutations,
     initializeCustomDriverAnnotationSettings, computeGenePanelInformation,
-    getQueriedStudies, CoverageInformation
+    fetchQueriedStudies, CoverageInformation
 } from "./ResultsViewPageStoreUtils";
 import {getAlterationCountsForCancerTypesForAllGenes} from "../../shared/lib/alterationCountHelpers";
 import sessionServiceClient from "shared/api//sessionServiceInstance";
-import { VirtualStudy } from "shared/model/VirtualStudy";
 import MobxPromiseCache from "../../shared/lib/MobxPromiseCache";
 import OncoprintClinicalDataCache from "../../shared/cache/OncoprintClinicalDataCache";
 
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1228,33 +1227,14 @@ export class ResultsViewPageStore {
...
-
     //this is only required to show study name and description on the results page
     readonly queriedStudies = remoteData({
-		await: ()=>[this.studyIdToStudy, this.virtualStudyIdToStudy],
 		invoke: async ()=>{
-            return getQueriedStudies(this.studyIdToStudy.result,
-                                     this.virtualStudyIdToStudy.result,
-                                     this.cohortIdsList);
+            if(!_.isEmpty(this.cohortIdsList)){
+                return getQueriedStudies(toJS(this.cohortIdsList));
```
alisman: just curious why calling toJS is necessary?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1228,33 +1227,14 @@ export class ResultsViewPageStore {
...
-                        description: virtualStudy.data.description,
-                        cancerTypeId: "My Virtual Studies"
-                    } as CancerStudy;
-                    return study;
-                }), x =>x.studyId);
-        }
-    },{});
-
     //this is only required to show study name and description on the results page
     readonly queriedStudies = remoteData({
```
alisman: there's something weird about the name of this.  queriedStudies suggests relationship with the studies collection, but in fact is just returning contents of cohortIdsList, should there be anything inside of it.   Seems to the the name queriedStudies needs to be more descriptive.

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -208,36 +209,39 @@ export function annotateMolecularDatum(
     return Object.assign({oncoKbOncogenic: oncogenic}, molecularDatum);
 }
 
-export async function getQueriedStudies(
-    physicalStudies:{[id:string]:CancerStudy},
-    virtualStudies:{[id:string]:CancerStudy},
-    queriedIds:string[]
-):Promise<CancerStudy[]>{
+export async function getQueriedStudies(queriedIds:string[]):Promise<CancerStudy[]>{
```
alisman: lets rename this fetchQueriedStudies

PR #1185, opened 2018-05-18
Send mobx-promise instead of data to chart component

kalletlak:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -239,31 +267,62 @@ export class StudyViewPageStore {
...
-            attributeIds.length > this._previousQueriedPatientAttributesId.length ||
+        if(attributeIds.length > this._previousQueriedPatientAttributesId.length ||
             (JSON.stringify(this.filters) !== JSON.stringify(this._previousFilters))){
             return attributeIds;
         } else {
             return this._previousQueriedPatientAttributesId
         }
     }
 
+    @computed get clinicalAttributePromises(): { [id: string]: MobxPromise<ClinicalDataCount[]> } {
```
alisman: is there a reason this doesn't just use remoteData (factory for mobx promises)?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -45,6 +46,19 @@ export type SurvivalType = {
 export class StudyViewPageStore {
 
     constructor() {
+        reaction(
```
alisman: not sure I understand what this does.

src/pages/studyView/StudyViewPageStore.ts
kalletlak: I just require a list of `MobxPromise` for each of the attribute. May I'll change to `remoteData` if needed

src/pages/studyView/StudyViewPageStore.ts
kalletlak: I need to hide charts initially when the page is loaded depending on number of categories(slices) each clinical attribute has.

alisman: this needs a lot of comments.  if i understand correctly, you want to determine what charts to show initially based on the initial response of the clinicalData service, but NOT on subsequent?  if that's the case there may be a much simpler way to accomplish this.  in general we should almost never use reactions/autorun unless really necessary

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -45,6 +46,19 @@ export type SurvivalType = {
...
+            () => this._loadedInitially,
+            (loadedInitially, reaction) => {
+                if (loadedInitially) {
+                    _.forEach(this.clinicalData.result, (obj, key) => {
```
alisman: is this similar to defaultVisibleAttributes?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: Yes, that's true. I'll add comments and see if I can get rid of reaction block

alisman: a different strategy might be, make an observable property on store, say, initialData=undefined.    Then give it a setter than only ever updatest that value if it is undefined.    this.intitialData = this.initialData || newData;

alisman: would that work?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: Similar? you mean executed once initially?

alisman: i didn't look too closely but it seemed to doing the same thing.

src/pages/studyView/StudyViewPageStore.ts
kalletlak: I'll give it a try

src/pages/studyView/StudyViewPageStore.ts
kalletlak: `defaultVisibleAttributes` pick 20 attributes from all clinical attributes, but this block filters the attributes that needs to be show depending on the data what we got for those 20 attributes. this logic is very basic. there is much more complicated logic in current study-view to handle this which will be covered in the next releases

PR #1169, opened 2018-05-11
Refactored Expression tab

alisman:
# What? Why?
Fix # .

Changes proposed in this pull request:
- a
- b

# Checks
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Airbnb React/JSX Style guide](https://github.com/airbnb/javascript/tree/master/react).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)

# Any screenshots or GIFs?
If this is a new visual feature please add a before/after screenshot or gif
here with e.g. [GifGrabber](http://www.gifgrabber.com/).

# Notify reviewers
Read our [Pull request merging
policy](../CONTRIBUTING.md#pull-request-merging-policy). If you are part of the
cBioPortal organization, notify the approprate team (remove inappropriate):

@cBioPortal/frontend

If you are not part of the cBioPortal organization look at who worked on the
file before you. Please use `git blame <filename>` to determine that
and notify them here:

PR #1186, opened 2018-05-18
Refactor Enrichments tab

ersinciftci:
@alisman both this and https://github.com/cBioPortal/cbioportal-frontend/pull/1169 has your src/shared/lib/boxPlotUtils.ts, so the PR that gets merged last will need to remove it after rebasing.


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -283,6 +283,11 @@ export class ResultsViewPageStore {
 
     public mutationAnnotationSettings:MutationAnnotationSettings;
 
+    @observable selectedEnrichmentMutationProfile: MolecularProfile;
```
alisman: use @observable.ref here.  @observable makes the object deeply @observable, which is unnecessary since the data is/should be immutable.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -617,6 +622,18 @@ export class ResultsViewPageStore {
...
     });
 
+    readonly alteredSamples = remoteData<Sample[]>({
+        await: () => [
+            this.sampleKeyToSample,
+            this.alteredSampleKeys
+        ],
+        invoke: () => {
+            const alteredSamples: Sample[] = [];
+            this.alteredSampleKeys.result!.forEach(a => alteredSamples.push(this.sampleKeyToSample.result![a]));
```
alisman: nitpick, but you could just use map, no?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1792,6 +1821,150 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly mutationEnrichmentProfiles = remoteData<MolecularProfile[]>({
+        await: () => [
+            this.molecularProfilesInStudies
+        ],
+        invoke: async () => {
+            const profiles: MolecularProfile[] = _.filter(this.molecularProfilesInStudies.result,
```
alisman: another nitpick, but _.find return first matching item

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1792,6 +1821,150 @@ export class ResultsViewPageStore {
...
+    });
+
+    readonly mutationEnrichmentData = remoteData<AlterationEnrichment[]>({
+        await: () => [
+            this.alteredSamples,
+            this.unalteredSamples,
+            this.mutationEnrichmentProfiles
+        ],
+        invoke: async () => {
+            return this.selectedEnrichmentMutationProfile ? this.sortEnrichmentData(
```
alisman: maybe a comment here.  seems like this would return either a single profile or a an array of profiles?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1792,6 +1821,150 @@ export class ResultsViewPageStore {
...
+                        unalteredIds: this.unalteredSamples.result.map(s => s.sampleId)}})) : [];
+        }
+    });
+
+    readonly copyNumberEnrichmentProfiles = remoteData<MolecularProfile[]>({
+        await: () => [
+            this.molecularProfilesInStudies
+        ],
+        invoke: async () => {
+            const profiles: MolecularProfile[] = _.filter(this.molecularProfilesInStudies.result,
```
alisman: _.find

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1792,6 +1821,150 @@ export class ResultsViewPageStore {
...
+                (profile: MolecularProfile) => profile.molecularAlterationType === AlterationTypeConstants.COPY_NUMBER_ALTERATION
+                    && profile.datatype === "DISCRETE");
+            this.selectedEnrichmentCopyNumberProfile = profiles[0];
```
alisman: see above

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1792,6 +1821,150 @@ export class ResultsViewPageStore {
...
+                unalteredIds: unalteredSamples.map(s => s.sampleId)
+        }}));
+    }
+
+    readonly mRNAEnrichmentProfiles = remoteData<MolecularProfile[]>({
+        await: () => [
+            this.molecularProfilesInStudies
+        ],
+        invoke: async () => {
+            let profiles: MolecularProfile[] = _.filter(this.molecularProfilesInStudies.result,
```
alisman: not a big deal but couldn't the filter and _remote steps be done in the filter?

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+    @observable mutualExclusivityFilter: boolean = true;
+    @observable coOccurenceFilter: boolean = true;
+    @observable significanceFilter: boolean = false;
+    @observable.shallow checkedGenes: string[] = [];
+    @observable clickedGene: string;
+    @observable selectedGenes: string[]|null;
+
+    constructor(props: IAlterationEnrichmentContainerProps) {
+        super(props);
+        this.mutualExclusivityFilterChange = this.mutualExclusivityFilterChange.bind(this);
```
alisman: you can install react-autobind, which lets you do binding with a decorator on method.   adam and i started using this but it's not merged yet.

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -0,0 +1,142 @@
...
+
+        const clickedAlterationEnrichment: AlterationEnrichment = _.find(this.props.data, ['hugoGeneSymbol', this.clickedGene])!;
+
+        return [this.props.totalAlteredCount - clickedAlterationEnrichment.alteredCount, 
+            clickedAlterationEnrichment.alteredCount, 
+            clickedAlterationEnrichment.unalteredCount, 
+            this.props.totalUnalteredCount - clickedAlterationEnrichment.unalteredCount];
+    }
+
+    private mutualExclusivityFilterChange() {
```
alisman: maybe give this a more active name, like "toggleMutualExclusivityFilter" ?

src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -0,0 +1,170 @@
...
+        this.generateColumns();
+        this.props.onCheckGene(hugoGeneSymbol);
+    }
+
+    private geneNameClick(hugoGeneSymbol: string) {
+        this.props.onGeneNameClick(hugoGeneSymbol);
+    }
+
+    protected generateColumns() {
+        this._columns = {};
```
alisman: make this a pure function.  it should return the columns object and calling code can assign it to the instance prop

src/pages/resultsView/enrichments/DataSetDropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+        this.change = this.change.bind(this);
+    }
+
+    private change(e: any) {
+        this.props.onChange(JSON.parse(e.target.value));
+    }
+
+    public render() {
+
+        if (this.props.dataSets.length < 2 && (this.props.dataSets[0].molecularAlterationType === "MUTATION_EXTENDED"
```
alisman: maybe test for length  === 1 since dataSets[0] assumes at least 1

src/pages/resultsView/enrichments/DataSetDropdown.tsx
```diff
@@ -0,0 +1,51 @@
...
+            || this.props.dataSets[0].molecularAlterationType === "COPY_NUMBER_ALTERATION")) {
+            return null;
+        }
+
+        const options: any = [];
```
alisman: HTMLOptionElement[]

src/pages/resultsView/enrichments/EnrichmentsTab.tsx
```diff
@@ -0,0 +1,64 @@
...
+        if (this.props.store.mutationEnrichmentProfiles.isPending ||
+            this.props.store.copyNumberEnrichmentProfiles.isPending ||
+            this.props.store.mRNAEnrichmentProfiles.isPending ||
+            this.props.store.proteinEnrichmentProfiles.isPending) {
+            return <Loader isLoading={true} />;
+        }
+
+        return (
+            <MSKTabs activeTabId={this.currentTabId} onTabClick={this.handleTabChange}>
+                {this.props.store.mutationEnrichmentProfiles.result!.length > 0 && <MSKTab id="mutations" linkText="Mutations">
```
alisman: i kind of like parenthesis around clauses when we use this syntax

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -0,0 +1,271 @@
...
+    } else if (logOddsRatio > 10) {
+        return ">10";
+    }
+    return logOddsRatio.toFixed(2);
+}
+
+export function formatLogOddsRatioWithStyle(logOddsRatio: number): JSX.Element {
+
+    if (logOddsRatio >= 0) {
+        return <span style={{ color: "green" }}>{formatLogOddsRatio(logOddsRatio)}</span>;
```
alisman: better to use hex values

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -0,0 +1,271 @@
...
+    } else {
+        return <span style={{ color: "red" }}>{formatLogOddsRatio(logOddsRatio)}</span>;
+    }
+}
+
+export function formatValueWithStyle(value: number): JSX.Element {
+
```
alisman: i've started using a library called http://numeraljs.com/, which has some nice formatting/precision features.  not merged yet, but you could install it

alisman: but you don't have to

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -0,0 +1,271 @@
...
+        currentGene = hugoGeneSymbol;
+    });
+
+    return result;
+}
+
+export function shortenGenesLabel(genes: string[], limit: number): string {
+
+    if (genes.length > limit) {
+        return genes.slice(0, limit).join(" ") + "...";
```
alisman: there is an elipsis unicode char 
&#8230;

src/shared/lib/NumberUtils.ts
```diff
@@ -0,0 +1,8 @@
+export function toConditionalPrecision(number: number, precision: number, threshold: number): string {
```
alisman: yeah, see if you could substitute numeral.js

src/pages/resultsView/ResultsViewPageStore.ts
ersinciftci: I need both the first item (to set as default) and all filtered items (to return from the function) in this function.

alisman: gotcha. so per the note below, i think the default needs to be a computation of this, or we should set it in onResult.  probably a computation which only ever sets it once is best.  i don't know the exact timing/lifecycle of onResult

src/pages/resultsView/ResultsViewPageStore.ts
ersinciftci: _.remove is actually used to move the removed elements at the top of the list, so not a permanent _.remove operation. I'll add a comment.

PR #1188, opened 2018-05-22
MyCancerGenome links now open in a separate tab

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3722.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/annotation/MyCancerGenome.tsx
+++ b/src/shared/components/annotation/MyCancerGenome.tsx
@@ -14,6 +14,40 @@ export function placeArrow(tooltipEl: any) {
 }
 
 /**
+ * This is NOT a generic html anchor tag parser. Assuming all MyCancerGenome links have the same simple structure.
+ * DOMParser can be used for generic parsing purposes, but it is not compatible with the current test environment.
+ * (i.e: DOMParser unit tests always fail, but it works fine in the browser)
+ */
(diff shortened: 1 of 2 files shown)


src/shared/components/annotation/MyCancerGenome.tsx
```diff
@@ -13,6 +13,40 @@ export function placeArrow(tooltipEl: any) {
...
+ * (i.e: DOMParser unit tests always fail, but it works fine in the browser)
+ */
+export function parseMyCancerGenomeLink(link: string)
+{
+    // Generic DOMParser solution (not compatible with the current unit test environment):
+    // const parser = new DOMParser();
+    // const htmlDoc = parser.parseFromString(link, "text/html");
+    // const links = htmlDoc.getElementsByTagName('a');
+
+    // assuming that we only have one attribute which is href
```
alisman: you don't have to rewrite this stuff but in general, would be simpler to write a regexp with capture groups to do this, no?

onursumer: It would be simpler/shorter to capture with regex. But personally I find regex expressions difficult to read/implement. Ideally we should use some parser library here, but the libraries that I tried (`DOMParser`, `parse5`, etc.) didn't work in the test environment. And, even more ideally we should fix the data :)

PR #1199, opened 2018-05-25
Enable expansion for ‘merged gene’ tracks

fedde-s:
Building on the feature introduced by PR #1143, add an expansion option to the track menu of tracks queried by ‘merged track’ syntax. Tweak the markup of gene set tracks to match the colours and indentation of these tracks.

Depends on the API exposed by pull request cBioPortal/oncoprintjs#73.

## Screenshot ##
Screenshot of an Oncoprint showing a merged track with indented, grey expansions, and another with an ‘Expand’ option in its track options menu


fedde-s:
@adamabeshouse, could you review this one as well? It requires two methods to be exposed to TypeScript by the OncoprintJS API, as I presented in https://github.com/cBioPortal/oncoprintjs/pull/73.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -471,7 +472,11 @@ export class ResultsViewPageStore {
 
     readonly putativeDriverFilteredCaseAggregatedDataByUnflattenedOQLLine = remoteData<{
         cases: CaseAggregatedData<AnnotatedExtendedAlteration>,
-        oql: UnflattenedOQLLineFilterOutput<AnnotatedExtendedAlteration>
+        oql: UnflattenedOQLLineFilterOutput<AnnotatedExtendedAlteration>,
+        list?: {
+            cases: CaseAggregatedData<AnnotatedExtendedAlteration>,
+            oql: OQLLineFilterOutput<object>
+        }[]
```
adamabeshouse: could you pull this out into a type def and add a comment explaining it?

package.json
```diff
@@ -140,7 +140,7 @@
     "mobxpromise": "^1.3.1",
     "node-sass": "^4.3.0",
     "numeral": "^2.0.6",
-    "oncoprintjs": "^1.0.95",
+    "oncoprintjs": "^1.0.98",
```
fedde-s: Thanks for including this part, @alisman! I was out of reach of my laptop when the fix came in.

src/pages/resultsView/ResultsViewPageStore.ts
fedde-s: Right, this shape of object occurs all over the page store; it would be useful to give it a name.

fedde-s: Unless you're referring to the meaning of the `list:` property itself, which I now realise not to be exactly self-explanatory either if you haven't just written the code that interprets the OQL parser tree. I'd have time to clarify that on Monday.

fedde-s: I extracted the interfaces and added comments to them. In doing so I found that I'd accidentally left in an unused function before I made the PR, and I removed that function.

PR #1209, opened 2018-05-30
Enrichments style fixes

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4286, https://github.com/cBioPortal/cbioportal/issues/4301 and fixes tooltip positions and some other styles, also fixes Survival tooltip overflow.

### DIFF
--- a/src/pages/resultsView/enrichments/styles.module.scss
+++ b/src/pages/resultsView/enrichments/styles.module.scss
@@ -28,6 +28,12 @@
     font-size: 11px;
 }
 
+.TableContainer {
+    :global(.tableSearchInput) {
+        width: 196px !important;
+    }
+}
+
 .Checkbox {
     margin-top: -2px !important;
     margin-bottom: 0 !important;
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/enrichments/styles.module.scss
```diff
@@ -18,6 +18,7 @@
   
 .TableContainer {
     width: 800px;
+    margin-right: -10px;
```
alisman: what does this fix?

alisman: Not sure but I would think this would just increase the width of the table to 810px. If that's true, then maybe just increase width?

src/pages/resultsView/enrichments/styles.module.scss
```diff
@@ -28,6 +28,10 @@
     font-size: 11px;
 }
 
+.TableContainer input {
```
alisman: i think this might be to general because there are other inputs inside the container no?

PR #1216, opened 2018-06-01
Add hover color to clickable row lazymobxtable

adamabeshouse:


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -469,25 +469,37 @@ class LazyMobXTableStore<T> {
         }
     }
 
+    @computed get clickableClassName() {
```
alisman: seems like it would be better to leave specific color out of this and just have semantic states which could be targeted in css?

src/shared/components/simpleTable/styles.scss
```diff
@@ -4,22 +4,26 @@ th {
   background-color: rgb(255,255,255); // solid background
 }
 
-tbody>tr:nth-of-type(odd).highlight {
-  background-color: #E0E000 !important;
-}
-
-tbody>tr:nth-of-type(odd).highlight-bluegray {
+tbody>tr:nth-of-type(odd).highlighted {
```
alisman: this will be applied to every tbody globally, not just those of the simpleTable

vendor-bundles.webpack.config.js
```diff
@@ -57,7 +57,7 @@ config.plugins = [
 
 ];
 
-if (process.env.NODE_ENV !== 'development') {
+if (process.env.NODE_ENV !== 'development' && process.env.NODE_ENV !== "test") {
```
alisman: i don't think we should do this quiet yet.

PR #1222, opened 2018-06-04
Set a minimalistic default style to MSKTab

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4280


ersinciftci:
Apparently pill style gene tabs are also MSKTab, I thought they were different. So I'll create a new style for pill tabs in `global.scss`, and use `secondaryTabs` for the style in this PR.

alisman:
Agreed

PR #1227, opened 2018-06-05
Fix two sources of issues with unit tests

adamabeshouse:


src/global.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+    for (const stub of clientStubs) {
+        stub.restore();
+    }
+});
+
+function stubClient():any[] {
+    // returns stubs
+    const stubs:any[] = [];
+    const methodNames:(keyof CBioPortalAPI)[] = (Object.keys(client) as (keyof CBioPortalAPI)[]).filter(m=>(typeof (client as any)[m]) === "function");
+    for (const methodName of methodNames) {
```
alisman: there are a couple methods on client that aren't fetchers, so you might want to only stub /^get|fetch/

adamabeshouse: oh ok

adamabeshouse: example?

PR #1233, opened 2018-06-07
Show sample counts next to profile name in Enrichments

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4322

### DIFF
--- a/src/pages/resultsView/enrichments/MutationEnrichmentsTab.tsx
+++ b/src/pages/resultsView/enrichments/MutationEnrichmentsTab.tsx
@@ -4,7 +4,7 @@ import { ResultsViewPageStore } from "../ResultsViewPageStore";
 import { observable } from 'mobx';
 import AlterationEnrichmentContainer from 'pages/resultsView/enrichments/AlterationEnrichmentsContainer';
 import Loader from 'shared/components/loadingIndicator/LoadingIndicator';
-import DataSetDropdown from 'pages/resultsView/enrichments/DataSetDropdown';
+import EnrichmentsDataSetDropdown from 'pages/resultsView/enrichments/EnrichmentsDataSetDropdown';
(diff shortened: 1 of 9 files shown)


src/pages/resultsView/enrichments/MutationEnrichmentsTab.tsx
```diff
@@ -28,7 +28,8 @@ export default class MutationEnrichmentsTab extends React.Component<IMutationEnr
         return (
             <div>
                 <DataSetDropdown dataSets={this.props.store.mutationEnrichmentProfiles.result!} onChange={this.onProfileChange}
-                    selectedValue={JSON.stringify(this.props.store.selectedEnrichmentMutationProfile)} />
+                    selectedValue={JSON.stringify(this.props.store.selectedEnrichmentMutationProfile)}
```
alisman: how come we are stringifying an object?

ersinciftci: I need all the details of a profile (not just the id),when user changes the profile in the dropdown, so instead of setting the profileId as option's value, I set the stringified object so that I can do `JSON.parse(e.target.value)` in `onChange`.

alisman: I would say in general this is not a great idea.  Better to use an ID or index and look it up on selection.  You could do this easily inside the DataSetDropDown.  And onChange could then give the selected object as an argument.     Also, you might want to give DataSetDown a more specific name since it's not intended as a shared component.  I know it's inside folder, but because of ts completion, people could get confused.

ersinciftci: Fixed

PR #1252, opened 2018-06-13
Add screenshot tests for Enrichments

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4368


end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -58,6 +58,13 @@ function runResultsTestSuite(prefix){
         assertScreenShotMatch(res);
     });
```
alisman: lets also get one of the mRNA tab where you select something and bar chart shows up

ersinciftci: yeah I'll add that too, just trying to get this simple one to work consistently first

PR #1253, opened 2018-06-14
Add loader for all study view charts

zhx828:
Show loading gif when loading data

# Checks
  the `-s` option)

# Any screenshots or GIFs?
jun-14-2018 16-40-49

# Notify reviewers
@kalletlak @alisman

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -1,15 +1,15 @@
 import * as React from 'react';
-import { inject, observer, Observer } from "mobx-react";
+import {inject, observer} from "mobx-react";
 import styles from "./styles.module.scss";
 import {MutatedGenesTable} from "./table/MutatedGenesTable";
 import {CNAGenesTable} from "./table/CNAGenesTable";
 import {ChartContainer, ChartType} from 'pages/studyView/charts/ChartContainer';
 import SurvivalChart from "../resultsView/survival/SurvivalChart";
 import {MSKTab, MSKTabs} from "../../shared/components/MSKTabs/MSKTabs";
-import { StudyViewPageStore, ClinicalDataType, SurvivalType } from 'pages/studyView/StudyViewPageStore';
-import { reaction } from 'mobx';
-import { ClinicalAttribute } from 'shared/api/generated/CBioPortalAPI';
-import _ from "lodash";
+import {ClinicalDataType, StudyViewPageStore, SurvivalType} from 'pages/studyView/StudyViewPageStore';
+import {reaction} from 'mobx';
+import {ClinicalAttribute} from 'shared/api/generated/CBioPortalAPI';;
+import {StudyViewComponentLoader} from "./charts/StudyViewComponentLoader";
 
 export interface IStudyViewPageProps {
     routing: any;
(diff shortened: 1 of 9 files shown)


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -129,18 +139,18 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                                 </div>
                             }
                             <div className={styles.flexContainer}>
-                                {(this.store.mutatedGeneData.isComplete && <MutatedGenesTable
-                                    data={mutatedGeneData}
```
kalletlak: you could get rid off local variable mutatedGeneData

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -107,26 +107,32 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
-                    <If condition={this.props.chartType===ChartType.PIE_CHART}>
-                        <PieChart
-                            ref={this.handlers.ref}
-                            onUserSelection={this.handlers.onUserSelection}
-                            filters={this.props.filters}
-                            data={this.data} />
-                    </If>
+                    <div style={{position: 'relative', height: 'calc(100% - 20px)'}}>
+                        <If condition={!this.props.promise.isComplete}>
+                            <StudyViewComponentLoader></StudyViewComponentLoader>
```
kalletlak: may be `<StudyViewComponentLoader/>`

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -2,14 +2,18 @@ import * as React from "react";
...
 import {default as LazyMobXTable} from "shared/components/lazyMobXTable/LazyMobXTable";
-import {action} from "mobx";
+import {action, computed} from "mobx";
 import {observer} from "mobx-react";
 import styles from "./tables.module.scss";
 import {CopyNumberCountByGene, CopyNumberGeneFilterElement} from "shared/api/generated/CBioPortalAPIInternal";
 import LabeledCheckbox from "../../../shared/components/labeledCheckbox/LabeledCheckbox";
+import MobxPromise from "mobxpromise";
+import {If} from 'react-if';
+import {ThreeBounce} from 'better-react-spinkit';
```
kalletlak: never used

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -71,25 +73,33 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
-                           showLegend={false}
-                           styleOpts={{
-                               width: 450,
-                               height: 300
-                           }}
-                           fileName="Overall_Survival"/>
+            <div className={styles.survivalPlotTitle}>{data.title}</div>
+            <div className={styles.survivalPlotBody}>
+                <If condition={!this.store.survivalPlotData.isComplete}>
+                    <StudyViewComponentLoader></StudyViewComponentLoader>
```
kalletlak: Why not use `Loader` component from `src/shared/components/loadingIndicator/LoadingIndicator`?

zhx828: They are both wrappers of TreeBounce but this one only has styles specifically applied to study view. I don't see the benefits of making LoadingIndicator accept one more property for the div and duplicate the styles every place uses it.

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -20,6 +23,9 @@ class CNAGenesTableComponent extends LazyMobXTable<CopyNumberCountByGene> {
 
 @observer
 export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
+    @computed get data() {
```
alisman: we don't need to access this via computed property. basically, the result property should be dereferenced in the same place as where its status is checked, so that the relationship between the two is made obvious.  as it stands now, a developer would have know idea they need to check status before referening the data property.   Alternatively, the status can be dealt with by the parent of this component and the dereferenced data passed directly to the component.  That way the component doesn't need to have knowledge of status because the data will always be populated.

alisman: But I can see you are checking status in the render method, so just reference result directly there.

src/pages/studyView/table/tables.module.scss
```diff
@@ -13,4 +13,5 @@
 
 .body {
```
alisman: i think we should avoid a class named "body" as it's too easily confused with <body>.  lets make this more specific

src/pages/studyView/table/MutatedGenesTable.tsx
```diff
@@ -19,17 +23,23 @@ class MutatedGenesTableComponent extends LazyMobXTable<MutationCountByGene> {
 @observer
 export class MutatedGenesTable extends React.Component<IMutatedGenesTablePros, {}> {
 
+    @computed get data() {
```
alisman: again, no reason to put this in computed.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -43,15 +42,15 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
     private plot:AbstractChart;
 
     @observable mouseInPlot:boolean = false;
-    
+
     constructor(props: IChartContainerProps) {
         super(props);
         let fileName = props.clinicalAttribute.displayName.replace(/[ \t]/g,'_')
```
alisman: fileName should be computed because otherwise it won't always reflect prop.  it's possible that this property is static, but in general we want to always make sure that our archictecture is reactive and doesn't have cached state.

src/pages/studyView/charts/StudyViewComponentLoader.tsx
```diff
@@ -0,0 +1,29 @@
...
+export interface StudyViewComponentLoaderProps {
+    promise: MobxPromise<any>
+}
+
+export class StudyViewComponentLoader extends React.Component<StudyViewComponentLoaderProps> {
+    public render() {
+        return (
+            <div>
+                <If condition={this.props.promise.isPending}>
+                    <div style={{position: 'absolute', left: '50%', top: '40%'}}>
```
alisman: we should do this centering with https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -107,26 +109,29 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
                     />
-
-                    <If condition={this.props.chartType===ChartType.PIE_CHART}>
-                        <PieChart
-                            ref={this.handlers.ref}
-                            onUserSelection={this.handlers.onUserSelection}
-                            filters={this.props.filters}
-                            data={this.data} />
-                    </If>
+                    <div style={{position: 'relative', height: 'calc(100% - 20px)'}}>
```
alisman: Not sure about this height calculation.  Why not have a fixed height and then padding?

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -69,13 +71,13 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
             onMouseLeavePlot: action(()=>{ this.mouseInPlot = false;}),
             handleDownloadDataClick:()=>{
                 let firstLine = this.props.clinicalAttribute.displayName+'\tCount'
-                fileDownload(firstLine+'\n'+this.plot.downloadData(), fileName);
+                fileDownload(firstLine+'\n'+this.plot.downloadData(), this.fileName);
             },
             handleSVGClick:()=>{
```
alisman: i'm surprised that this works.  i would think the click handlers need to be bound to instance (which we can now with auto-bind decorator).

zhx828: The download feature is not updated in this pull request and not used in anyway currently. We will send out another pull request to fix the issue.

src/pages/studyView/styles.module.scss
```diff
@@ -1,10 +1,21 @@
-.flexContainer {
+.studyViewFlexContainer {
     display: inline-flex;
     flex-direction: row;
     flex-wrap: wrap;
     width: 100%;
 }
-.survivalPlot {
+.studyViewSurvivalPlot {
     float: left;
```
alisman: we can clean this up loater but we shouldn't need float anymore.  In a flex box layout, float has no impact.

PR #1254, opened 2018-06-15
Study view refactoring

kalletlak:
This includes
1. Table tooltip for pie chart
2. Delete chart
3. Improve pie chart interaction performance
4. Support multi study

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -3,13 +3,13 @@ import {inject, observer} from "mobx-react";
 import styles from "./styles.module.scss";
 import {MutatedGenesTable} from "./table/MutatedGenesTable";
 import {CNAGenesTable} from "./table/CNAGenesTable";
-import {ChartContainer, ChartType} from 'pages/studyView/charts/ChartContainer';
+import {ChartContainer} from 'pages/studyView/charts/ChartContainer';
 import SurvivalChart from "../resultsView/survival/SurvivalChart";
 import {MSKTab, MSKTabs} from "../../shared/components/MSKTabs/MSKTabs";
-import {ClinicalDataType, StudyViewPageStore, SurvivalType} from 'pages/studyView/StudyViewPageStore';
-import {reaction} from 'mobx';
-import {ClinicalAttribute} from 'shared/api/generated/CBioPortalAPI';;
 import {StudyViewComponentLoader} from "./charts/StudyViewComponentLoader";
+import { StudyViewPageStore, ClinicalDataType, SurvivalType, ChartMeta, ChartType } from 'pages/studyView/StudyViewPageStore';
+import { reaction } from 'mobx';
+const CheckedSelect = require("react-select-checked").CheckedSelect;
 
 export interface IStudyViewPageProps {
     routing: any;
@@ -31,15 +31,24 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps, 
         this.store = new StudyViewPageStore();
 
         this.handlers = {
-            onUserSelection: (attrId: string, clinicalDataType: ClinicalDataType, values: string[])=> {
-                this.store.updateClinicalDataEqualityFilters(attrId, clinicalDataType, values)
+            onUserSelection: (chartMeta : ChartMeta, values: string[])=> {
+                this.store.updateClinicalDataEqualityFilters(chartMeta, values)
             },
             updateGeneFilter:(entrezGeneId: number)=>{
                 this.store.updateGeneFilter(entrezGeneId);
             },
             updateCNAGeneFilter:(entrezGeneId: number, alteration: number)=> {
                 this.store.updateCNAGeneFilter(entrezGeneId, alteration);
-            }
+            },
+            onDeleteChart:(uniqueKey:string)=>{
+                this.store.changeChartVisibility(uniqueKey, false);
+            },
+            changeChartType: (uniqueKey: string, chartType: ChartType, updateDefaultType?:boolean)=> {
+                this.store.changeChartType(uniqueKey, chartType, updateDefaultType)
+            },
+            onChangeSelectedClinicalChart:(options: {label: string; value: string;}[])=>{
+                this.store.onChangeSelectedClinicalChart(options)
+            } 
         }
 
         //TODO: this should be done by a module so that it can be reused on other pages
@@ -56,16 +65,15 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps, 
         );
     }
 
-    renderAttributeChart = (clinicalAttribute: ClinicalAttribute) => {
-        const data = this.store.studyViewClinicalDataCountsCache.get({attribute:clinicalAttribute,filters:this.store.filters})
-        const filters = this.store.getClinicalDataEqualityFilters(clinicalAttribute)
+    renderAttributeChart = (chartMeta: ChartMeta) => {
         return (<ChartContainer
-            chartType={ChartType.PIE_CHART}
-            clinicalAttribute={clinicalAttribute}
+            chartMeta={chartMeta}
             onUserSelection={this.handlers.onUserSelection}
-            key={clinicalAttribute.clinicalAttributeId}
-            promise={data}
-            filters={filters}
+            key={chartMeta.uniqueKey}
+            filter={this.store.filters}
+            dataCache ={this.store.studyViewClinicalDataCountsCache}
+            onDeleteChart={this.handlers.onDeleteChart}
+            changeChartType={this.handlers.changeChartType}
         />)
     };
 
@@ -116,21 +124,22 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps, 
                          className="mainTabs">
 
                     <MSKTab key={0} id="summaryTab" linkText="Summary">
-                            {
-                                this.store.defaultVisibleAttributes.isComplete && 
-                                this.store.initalClinicalDataCounts.isComplete &&
-                                this.store.initialized && 
-                                (
-                                    <div className={styles.studyViewFlexContainer}>
-                                        {this.store.visibleAttributes.map(this.renderAttributeChart)}
-                                    </div>
-                                )
-                            }
-                            {
-                                <div className={styles.studyViewFlexContainer}>
-                                    {this.store.survivalPlotData.result.map(this.renderSurvivalPlot)}
-                                </div>
-                            }
+                        <div style={{"width":"250px"}}>
+                            <CheckedSelect
+                                placeholder="Add clinical tracks.."
+                                isLoading={this.store.initialized}
+                                noResultsText={this.store.initialized ? "Loading..." : "No matching clinical tracks found"}
+                                onChange={this.handlers.onChangeSelectedClinicalChart}
+                                options={this.store.clinicalChartOptions}
+                                value={this.store.selectedClinicalCharts}
+                            />
+                        </div>
+                            <div className={styles.studyViewFlexContainer}>
+                                {this.store.visibleAttributes.map(this.renderAttributeChart)}
+                            </div>
+                            <div className={styles.studyViewFlexContainer}>
+                                {this.store.survivalPlotData.result.map(this.renderSurvivalPlot)}
+                            </div>
                             <div className={styles.studyViewFlexContainer}>
                                 <MutatedGenesTable
                                     promise={this.store.mutatedGeneData}
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -1,15 +1,15 @@
 import * as React from 'react';
-import {inject, observer} from "mobx-react";
+import { inject, observer } from "mobx-react";
 import styles from "./styles.module.scss";
-import {MutatedGenesTable} from "./table/MutatedGenesTable";
-import {CNAGenesTable} from "./table/CNAGenesTable";
-import {ChartContainer} from 'pages/studyView/charts/ChartContainer';
+import { MutatedGenesTable } from "./table/MutatedGenesTable";
+import { CNAGenesTable } from "./table/CNAGenesTable";
+import { ChartContainer } from 'pages/studyView/charts/ChartContainer';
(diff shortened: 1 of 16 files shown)


src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -139,8 +178,8 @@ class CustomSlice extends React.Component<{}, {}> {
         const d:any = this.props;
         return (
         <g>
-            <Slice {...this.props}/>
-            <title>{`${d.datum.x}:${d.datum.y}`}</title>
+            <Slice {...this.props} active={true}/>
```
kalletlak: cleanup `active`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -146,13 +145,63 @@ export class StudyViewPageStore {
...
+        if(!visibility){
+            //TODO: Currently clears only clinicalDataEqualityFilters,Need to implement for others
+            this._clinicalDataEqualityFilterSet.delete(uniqueKey);
+        }
+        this._chartVisibility.set(uniqueKey, visibility);
+    }
+
+    @action changeChartType(uniqueKey: string, chartType:ChartType, updateDefaultType?:boolean){
+        let chartMetaObject = this._clinicalAttributesMetaSet.get(uniqueKey);
+        if(chartMetaObject){
```
alisman: is there a reason there would not be a meta object?  just seems strange you don't deal with that case if it in fact is met

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -146,13 +145,63 @@ export class StudyViewPageStore {
...
+    }
+    @action onChangeSelectedClinicalChart(options: {label: string; value: string;}[]) {
+        this._chartVisibility.clear();
+        for (const option of options) {
+            this.changeChartVisibility(option.value, true)
+        }
+    }
+
+    @computed get clinicalChartOptions() {
+        return _.map(this.clinicalAttributesMeta.result, clinicalAttributeMeta=>({
```
alisman: if clinicalAttributesMeta is asychronous (i.e. mobxpromise) then the wrapper should also be a moxpromise so that consumer knows they have to check for status=complete.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -146,13 +145,63 @@ export class StudyViewPageStore {
...
+            label: clinicalAttributeMeta.clinicalAttribute.displayName,
+            value: clinicalAttributeMeta.uniqueKey
+        }));
+    }
+
+    @computed get selectedClinicalCharts() {
+        return _.reduce(this._chartVisibility.keys(),(acc:{label: string; value: string;}[], next)=>{
+            if(this._chartVisibility.get(next)) {
```
alisman: i would put === true so it's more clear that you aren't just testing for presence of item in collection, but checking its value.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -146,13 +145,63 @@ export class StudyViewPageStore {
...
+                        value: chartMeta.uniqueKey
+                    })
+                }
+            }
+            return acc;
+        },[]);
+    }
+
+    @computed get emptyFilters() {
+        let filters: StudyViewFilter = {} as any;
```
alisman: just use object literal.  return { studyIds:[this.StudyId]  }

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -146,13 +145,63 @@ export class StudyViewPageStore {
...
+        filters.studyIds = [this.studyId]
```
alisman: will we sometimes have more than one studyId?  If so, should we engineer for that?  i.e. this.studyIds    instead of single studyId

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -283,49 +331,59 @@ export class StudyViewPageStore {
...
+                    .get({ attribute: next.clinicalAttribute, filters: this.emptyFilters })
                     .result!;
                 return acc;
             }, {} as any);
         },
         default: {},
-        onResult:() => {
-            this.initialized=true;
+        onResult:(result) => {
+            if(!this._initialized){
```
alisman: not sure i understand need for initialization.  we should discuss this on chat.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -90,50 +91,122 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+                this.props.changeChartType(this.props.chartMeta.uniqueKey, ChartType.TABLE, true);
+            }
+        }
+    })
+
+
+
+    @computed get chartWidth(){
+        let chartWidth = styles.chartWidthTwo
+        if(this.props.chartMeta.chartType === ChartType.PIE_CHART){
```
alisman: how many different chart types are there?

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -90,50 +91,122 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+            chartWidth = styles.chartWidthOne
+        }
+        return chartWidth;
     }
 
-    @computed get isChartVisible(){
-        //check if there is chart implementation is present
-        return this.props.chartType === ChartType.PIE_CHART && this.props.clinicalAttribute.datatype==='STRING';
+    @computed get chartHeight(){
+        let chartHeight = styles.chartHeightTwo
```
alisman: semicolons

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -90,50 +91,122 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+        if((this.props.chartMeta.chartType === ChartType.PIE_CHART) || (this.props.chartMeta.chartType === ChartType.TABLE && (this.initialData.result!).length<=3)){
+            chartHeight = styles.chartHeightOne
+        }
+        return chartHeight;
     }
 
-    @computed get data(){
-        return this.props.promise.result || [];
+    @computed get hideLabel(){
+        return this.props.chartMeta.chartType === ChartType.TABLE ? true : false;
```
alisman: can you just return the condition

src/pages/studyView/StudyViewPageStore.ts
kalletlak: `.get` is of type `chartMeta | undefined`. I used `if` condition to check it. I could you post operator`!`

src/pages/studyView/charts/ChartContainer.tsx
kalletlak: there are 5.

alisman: seems like then are going to have a lot of switch statements and they need to have all five options.  hard to maintain.  would it be better to have subclasses which implement these members?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -229,47 +262,62 @@ export class StudyViewPageStore {
...
+            });
+        },
         default: [],
-        onResult:(attributes:ClinicalAttribute[])=>{
-            attributes.forEach(attribute=>{
-                this._attributeChartVisibility.set(attribute.clinicalAttributeId,false);
-            })
+        onResult:(result)=>{
+            _.forEach(result,obj=>{
+                this._clinicalAttributesMetaSet.set(obj.uniqueKey, obj)
```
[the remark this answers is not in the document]
kalletlak: there will be some actions, example changing chart type from pie to table, which is stored in the value via `chartType`. May be we can discuss this over chat.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -91,49 +91,124 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+                    active={this.mouseInChart}
+                    resetChart={this.handlers.resetFilters}
+                    deleteChart={this.handlers.onDeleteChart}
+                    hideLabel={this.hideLabel}
+                    chartControls={this.chartControls}
+                    changeChartType={this.changeChartType}
+                />
+
+                <If condition={this.props.chartMeta.chartType === ChartType.PIE_CHART}>
+                    <StudyViewComponentLoader promise={this.data}>
```
zhx828: The if condition should be placed inside of StudyViewComponentLoader. This component will be used for all charts, then you can avoid duplication later.

src/pages/studyView/charts/StudyViewComponentLoader.tsx
```diff
@@ -10,7 +10,7 @@ export interface StudyViewComponentLoaderProps {
 export class StudyViewComponentLoader extends React.Component<StudyViewComponentLoaderProps> {
     public render() {
         return (
-            <div style={{margin: 'auto'}}>
```
zhx828: Why did you remove the margin?

src/pages/studyView/charts/ChartContainer.tsx
kalletlak: if I do that and try changing the chart type, pie to table or vice-versa, nothing shows up.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -1,38 +1,34 @@
 import * as React from "react";
```
zhx828: This ChartContainer should soon be used by Scatter, Survival and Tables.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -75,62 +76,57 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
-                                (
-                                    <div className={styles.studyViewFlexContainer}>
-                                        {this.store.visibleAttributes.map(this.renderAttributeChart)}
-                                    </div>
-                                )
-                            }
-                            {
-                                <div className={styles.studyViewFlexContainer}>
-                                    {this.store.survivalPlotData.result.map(this.renderSurvivalPlot)}
+        if (this.store.studies.isComplete) {
```
zhx828: Should we add conditions when isPending and isError?

kalletlak: I can add this in up coming pr, by showing a loading symbol while its still pending

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -269,77 +262,71 @@ export class StudyViewPageStore {
...
+                    .get({ attribute: next, filters: this.emptyFilter })
                     .result!;
                 return acc;
             }, {} as any);
         },
         default: {},
-        onResult:() => {
-            this.initialized=true;
+        onResult: (result) => {
+            _.forEach(result, (obj, uniqueKey) => {
```
zhx828: Should we restart the conversation about the same uniqueKey(clinical attribute) in multiple studies?

kalletlak: Sure, I think its better to deal with this logic in backend

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -269,77 +262,71 @@ export class StudyViewPageStore {
             let patientAttributeCount = 0;
             let filterAttributes: ClinicalAttribute[] = []
             queriedAttributes.forEach(attribute => {
-                if(attribute.patientAttribute){
-                    if(patientAttributeCount<10){
+                if (attribute.patientAttribute) {
+                    if (patientAttributeCount < 10) {
```
alisman: put comments explaining business logic (i.e. <10)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -269,77 +262,71 @@ export class StudyViewPageStore {
...
-        onResult:(attributes:ClinicalAttribute[])=>{
-            attributes.forEach(attribute=>{
-                this._attributeChartVisibility.set(attribute.clinicalAttributeId,true);
-            })
-        }
+        default: []
     });
 
-    readonly initalClinicalDataCounts = remoteData<{[id:string]:ClinicalDataCount[]}>({
+    readonly initalClinicalDataCounts = remoteData<{ [id: string]: ClinicalDataCount[] }>({
```
alisman: initial not "inital"

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -179,84 +175,81 @@ export class StudyViewPageStore {
...
-    @computed get visibleAttributes(){
-        const attributes = this.clinicalAttributeSet.result
-        return _.reduce(this._attributeChartVisibility.toJS(),(acc:ClinicalAttribute[],next,key)=>{
-            if(next && attributes[key]){
-                acc.push(attributes[key])
-            }
-            return acc;
-        },[])
+    @computed get visibleAttributes(): ChartMeta[] {
+        if (this.initalClinicalDataCounts.isComplete) {
```
alisman: loading state should not handled inside of @computed getter.    this hides the loading state from calling code, which only knows whether array is empty or not (could be empty for other reasons).   Conditional should be in calling code.

kalletlak:
@alisman Still there are lot of moving pieces. Will write tests once there is a stable code

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -75,62 +76,58 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
+                                <h3>Multiple Studies</h3>
+                            </If>
+                        </div>
+                    </div>
+                    <MSKTabs id="studyViewTabs" activeTabId={this.props.routing.location.query.tab}
+                        className="mainTabs">
+
+                        <MSKTab key={0} id="summaryTab" linkText="Summary">
+                            <div className={styles.studyViewFlexContainer}>
+                                {this.store.initialClinicalDataCounts.isComplete &&
```
alisman: so, initialClinicalDataCounts will ensure the presenc eof visisbleAttributes, yes?  because onResult populates it?

kalletlak: yes

PR #1263, opened 2018-06-22
Add SVG and PNG buttons to volcano plot in enrichments

ersinciftci:
Fixes https://github.com/cBioPortal/cbioportal/issues/4381
No Data download button as it's the same data with the table


src/pages/resultsView/enrichments/MiniScatterChart.tsx
```diff
@@ -72,8 +87,16 @@ export default class MiniScatterChart extends React.Component<IMiniScatterChartP
         return (
             <div className="posRelative">
                 <div className="borderedChart inlineBlock">
-                    <VictoryChart containerComponent={<VictorySelectionContainer responsive={false}
-                        onSelection={(points: any, bounds: any, props: any) => this.handleSelection(points, bounds, props)} 
+                    <div className="btn-group" style={{position:'absolute', zIndex:10, right: 26 }} role="group">
```
alisman: @adamabeshouse  has a component for this now I think.

ersinciftci: Updated

src/shared/components/DownloadControls.tsx
```diff
@@ -17,6 +17,7 @@ interface IDownloadControlsProps {
     additionalLeftButtons?:ButtonSpec[],
     additionalRightButtons?:ButtonSpec[]
     dontFade?:boolean;
+    dontDisable?:boolean;
```
adamabeshouse: why would we want to have the buttons not be disabled if the svg isnt available?

ersinciftci: I thought it had a dependency on having a `<svg ref={this.svgRef}` wrapper, but forgot that `<VictorySelectionContainer containerRef={this.svgRef}` also takes a function, so fixed.

PR #1274, opened 2018-07-02
OncoKB Fixes

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/2934
Fix https://github.com/cBioPortal/cbioportal/issues/3817
Fix https://github.com/cBioPortal/cbioportal/issues/4422


src/shared/components/annotation/OncoKbCard.tsx
```diff
@@ -399,9 +407,16 @@ export default class OncoKbCard extends React.Component<OncoKbCardProps, IOncoKb
                     )}
 
                     <div className="item footer">
-                        <a href={`http://oncokb.org/#/gene/${this.props.gene}`} target="_blank">
-                            <img src={require("./images/oncokb.png")} className="oncokb-logo" alt="OncoKB"/>
-                        </a>
+                        <If condition={this.oncokbLinkOut===undefined}>
+                            <Then>
+                                <img src={require("./images/oncokb.png")} className="oncokb-logo" alt="OncoKB"/>
```
onursumer: maybe separate this `img` as a variable, so that we don't need to duplicate the same content in the `Else` block? Something like:

```JavaScript
const oncokbLogo = <img src={require("./images/oncokb.png")} className="oncokb-logo" alt="OncoKB"/>;
```

PR #1277, opened 2018-07-05
SV: Add custom case selection

kalletlak:


src/pages/studyView/SummaryHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+    @computed
+    get selectedPatientsCount() {
+        return _.uniq(this.props.selectedSamples.map(sample => sample.uniquePatientKey)).length;
+    }
+
+    @bind
+    private openCases() {
+
+        const firstSample = this.props.selectedSamples[0];
```
zhx828: You can't assume `selectedSamples` has an element.

src/pages/studyView/SummaryHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+        const groupedSamples = _.groupBy(this.props.selectedSamples, sample => sample.studyId);
+        const IncludeStudyId: boolean = Object.keys(groupedSamples).length > 1;
+
+        let navCaseIds = _.map(groupedSamples, (samples) => {
+            return _.map(samples, sample => {
```
adamabeshouse: I dont see the point of computing groupedSamples?

src/pages/studyView/customCaseSelection/CustomCaseSelection.tsx
```diff
@@ -0,0 +1,186 @@
...
+export default class CustomCaseSelection extends React.Component<ICustomCaseSelectionProps, {}> {
+
+    @observable showCaseIds: boolean = false;
+    @observable caseIdsMode: 'sample' | 'patient' = 'sample';
+    @observable caseIds: string = ''
+    @observable invalidCases: string[] = [];
+    @observable validCases: Sample[] = []
+
+    @computed get sampleSet(): { [id: string]: Sample } {
+        return _.reduce(this.props.selectedSamples, (acc: { [id: string]: Sample }, next) => {
```
adamabeshouse: you can write this as `_.keyBy(this.props.selectedSamples, s=>`${s.studyId}:${s.sampleId}`)`

src/pages/studyView/styles.scss
```diff
@@ -0,0 +1,22 @@
+.studyview-summary-header{
```
zhx828: Same here, we don't do dash anymore. All camelcase.

src/pages/studyView/customCaseSelection/CustomCaseSelection.tsx
kalletlak: User actually enters `sampleId` or `studyId:sampleId` in text area. 
Is there a login in front-end to convert `studyId:sampleId` to `uniqueSampleKey`?

src/pages/studyView/SummaryHeader.tsx
kalletlak: Here its used to check whether `studyId` needs to be prefixed or not. and also to order samples by study

src/pages/studyView/SummaryHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+}
+
+@observer
+export default class SummaryHeader extends React.Component<ISummaryHeaderProps, {}> {
+
+    @observable private isCustomCaseBoxOpen = false;
+
+    @computed
+    get selectedPatientsCount() {
+        return _.uniq(this.props.selectedSamples.map(sample => sample.uniquePatientKey)).length;
```
zhx828: _.uniq is unnecessary.

src/pages/studyView/SummaryHeader.tsx
adamabeshouse: ah yes, makes sense

src/pages/studyView/customCaseSelection/CustomCaseSelection.tsx
adamabeshouse: oh I see, this seems fine and a little different than whats going on in Query

src/pages/studyView/SummaryHeader.tsx
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    @bind
+    private openCases() {
+
+        const firstSample = this.props.selectedSamples[0];
+        const groupedSamples = _.groupBy(this.props.selectedSamples, sample => sample.studyId);
+        const IncludeStudyId: boolean = Object.keys(groupedSamples).length > 1;
```
zhx828: Shouldn't we only uppercase first letter when it is a component/type/interface?

src/pages/studyView/customCaseSelection/CustomCaseSelection.tsx
zhx828: It should be the same though. Just curious how this different from the one in Query

kalletlak: Query page includes some logic if Virtual studies are in selection. eventually this component should be updated to support that functionality and replace it(what's in query page)

adamabeshouse: the difference I was thinking of is that here theres already the selected study, and the samples in the study are already known, so you dont need to fetch more. that may be wrong though and im misunderstanding whats going on here

kalletlak: yes, even that's true.

src/pages/studyView/customCaseSelection/styles.module.scss
```diff
@@ -0,0 +1,55 @@
+.customCaseSelection {
```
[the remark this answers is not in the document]
kalletlak: Its gonna be an independent component right, I think this should be ok.

PR #1283, opened 2018-07-10
SV: Add gene selection box

kalletlak:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -78,6 +82,35 @@ export class StudyViewPageStore {
...
     private _clinicalAttributesMetaSet: { [id: string]: ChartMeta } = {} as any;
 
+    @observable private _query: SingleGeneQuery[] = [];
+    @observable private _genes: Gene[] = [];
+    @observable geneQuery: string = '';
+
+    @bind
+    @action onCheckGene(hugoGeneSymbol: string) {
+        let geneSet= stringListToSet(this.geneQuery.trim().split(/\s+/g));
+        if (geneSet[hugoGeneSymbol]) {
```
adamabeshouse: `geneSet[hugoGeneSymbol] = !geneSet[hugoGeneSymbol]`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -78,6 +82,35 @@ export class StudyViewPageStore {
...
+            geneSet[hugoGeneSymbol] = false;
+        } else {
+            geneSet[hugoGeneSymbol] = true;
+        }
+
+        this.geneQuery = _.reduce(geneSet, (acc: string[], next, geneId) => {
```
adamabeshouse: `next` could be renamed, maybe `inSet`

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: I don't think you're using the right data structures for the task here. What you want is a set of selected genes, then you can do 
```
@computed get geneQuery() { 
    return Object.keys(this._selectedGenes).filter(g=>!!this._selectedGenes[g]).join(' ');
}
@action set geneQuery(q:string) {
    this._selectedGenes = stringListToSet(q);
}
```

src/pages/studyView/geneSymbolSelection/GeneSelectionBox.tsx
```diff
@@ -0,0 +1,343 @@
...
+        }
+    });
+
+    private validationStatus() {
+
+        if (this.oql.error)
+            return (
+                <span className={styles.errorMessage}>
+                    {`Cannot validate gene symbols because of invalid OQL. ${
+                        this._geneQueryErrorDisplayStatus === 'unfocused'
```
adamabeshouse: I think its better practice to use constants/enums instead of the string values, like 'unfocused' and on line 275 'studyView'. That way the compiler can catch any typos which could otherwise become pernicious bugs, and also when programming the IDE can autocomplete

src/pages/studyView/geneSymbolSelection/GeneSelectionBox.tsx
```diff
@@ -0,0 +1,343 @@
...
+        );
+    }
+    @computed get placeHolder() {
+        if(this.props.location === 'studyView' && !this._isFocused){
+            return 'query genes - click to expand';
+        }
+        return 'Enter HUGO Gene Symbols or Gene Aliases';
+    }
+
+    @computed get fousOutValue() {
```
adamabeshouse: typo, `focus`?

src/pages/studyView/StudyViewUtils.ts
```diff
@@ -63,3 +63,35 @@ export const COLORS = [
 export const NA_COLOR = '#CCCCCC'
 
 export const UNSELECTED_COLOR = '#808080'
+
+
+export function updateGeneQuery(geneQuery: string, selectedGene: string): string {
+
+    let isSelectedGenePresent = false;
+    let geneQueryLines = geneQuery.includes(':') ? geneQuery.split('\n') : _.isEmpty(geneQuery) ? [] : [geneQuery];
```
adamabeshouse: I don't think this is safe, you should run OQL parser to get lines

src/pages/studyView/StudyViewUtils.ts
```diff
@@ -63,3 +63,35 @@ export const COLORS = [
...
+        if (queryLine.includes(':')) {
+            const gene = queryLine.substring(0, queryLine.indexOf(':')).trim();
+            const alteration = queryLine.substring(queryLine.indexOf(':') + 1).trim();
+            if (gene.toUpperCase() === selectedGene.toUpperCase()) {
+                isSelectedGenePresent = true;
+            } else {
+                result.push(gene + (alteration ? ': ' + alteration : ''));
+            }
+        } else {
+            _.forEach(queryLine.trim().split(/\s+/g), gene => {
```
adamabeshouse: this method is doing a lot of unsafe string processing on OQL it seems, that should really be done w the OQL parser. you'll get a lot of this work done for free that way too, like pasting together the gene and the alterations and such. the OQL parser output gives you each line's query string

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -78,6 +87,22 @@ export class StudyViewPageStore {
...
+    @bind
+    @action onCheckGene(hugoGeneSymbol: string) {
+        this.geneQuery = updateGeneQuery(this.geneQuery, hugoGeneSymbol);
+    }
+
+    @computed get selectedGenes(): string[] {
+        return this._genes.map(gene => gene.hugoGeneSymbol);
+    }
+
+    @action updateSelectedGenes(query: SingleGeneQuery[], genes: Gene[], queryStr: string) {
```
adamabeshouse: why are you accepting `query` as argument here?

src/pages/studyView/geneSymbolSelection/GeneSelectionBox.tsx
```diff
@@ -0,0 +1,342 @@
...
+import { getFocusOutText } from 'pages/studyView/geneSymbolSelection/GeneSelectionBoxUtils';
+
+export interface IGeneSelectionBoxProps {
+    inputGeneQuery?: string;
+    geneQueryErrorDisplayStatus?: DispalyStatus;
+    location?: GeneBoxType;
+    callback: (query: SingleGeneQuery[], genes: Gene[], queryStr: string, status: boolean) => void;
+}
+
+export enum DispalyStatus {
```
adamabeshouse: typo `DisplayStatus`

src/pages/studyView/geneSymbolSelection/GeneSelectionBox.tsx
```diff
@@ -0,0 +1,342 @@
...
+        default: { found: [], suggestions: [] },
+        onResult: (genes) => {
+            let status = _.isUndefined(this.oql.error) && genes.suggestions.length === 0;
+            this.props.callback(this.oql.query, genes.found, this._geneQuery, status)
+        }
+    });
+
+    @computed get validationStatusMessage() {
+
+        if (this.oql.error)
```
adamabeshouse: I do think this method should be tested somehow.. if you can find a way to abstract out the text computation maybe, that could be a nice way to test it easily, just test the text output

src/pages/studyView/geneSymbolSelection/GeneSelectionBox.tsx
```diff
@@ -0,0 +1,342 @@
...
+        return (
+            <div className={styles.validBubble} title="You can now submit the list.">
+                <FontAwesome className={styles.icon} name='check-circle' />
+                <span>All gene symbols are valid.</span>
+            </div>
+        );
+    }
+
+    @action private replaceGene(oldSymbol: string, newSymbol: string) {
+        this._geneQuery = normalizeQuery(this._geneQuery.toUpperCase().replace(new RegExp(`\\b${oldSymbol.toUpperCase()}\\b`, 'g'), () => newSymbol.toUpperCase()));
```
adamabeshouse: this is probably fine but again it's not really safe to do manual string processing on OQL (for example, what if there happens to be a gene with the same name as one of the OQL keywords... unlikely but not impossible). best to use OQL parser then modify that structured data

src/pages/studyView/StudyViewUtils.ts
kalletlak: is there a way to get back the query string from OQL parser?. Currently OQL parser is just used to validate gene query. its result is not used anywhere

adamabeshouse: I just made a PR to expose this function: https://github.com/cBioPortal/cbioportal-frontend/pull/1288

zhx828:
Featurewise it looks good, could you make the following changes?
- Please remove the `Query` button and others which are not properly working at this moment.
- When select genes, there is a half-second latency, is this something we can change?
- Could you apply the patient view button style to `Select cases` button?
- Could you apply the following animation similarly as the production page?
jul-14-2018 13-51-25

kalletlak:
@zhx828 that latency is because of gene API call.
there is no need to use GeneSelectionBox in QueryStore, we just need to update QuerySelector

zhx828:
@kalletlak I'm thinking when you click through tables, these hugosymbols are all valid. There is no need to do the gene validation anymore. What do you think?

zhx828:
@kalletlak the tooltip when you mouse over the gene is still missing

zhx828:
@kalletlak could you change the color of the gene box border from #d3d3d3 to #ddd to keep the consistency with other buttons?

src/pages/studyView/StudyViewUtils.ts
```diff
@@ -58,8 +57,21 @@ export const COLORS = [
...
+export function updateGeneQuery(geneQueries: SingleGeneQuery[], selectedGene: string): string {
+
+    let updatedQueries = _.filter(geneQueries,query=> query.gene !== selectedGene)
+    if(updatedQueries.length === geneQueries.length){
+        updatedQueries.push({
+            gene: selectedGene,
+            alterations: false
+        })
+    }
+    return updatedQueries.map(query=>unparseOQLQueryLine(query)).join('\n');
```
adamabeshouse: looks good

src/pages/studyView/SummaryHeader.tsx
```diff
@@ -74,9 +101,18 @@ export default class SummaryHeader extends React.Component<ISummaryHeaderProps,
...
+                    />
+                    {/* <span style={{ margin: "0px 5px" }}><i className="fa fa-arrow-right fa-lg" aria-hidden="true"></i></span>
+                    <button disabled={this.isQueryButtonDisabled} className="btn" onClick={() => null}>
+                        Query
+                    </button> */}
                     <button
-                        className="btn btn-default btn-sm"
-                        onClick={event=>this.isCustomCaseBoxOpen = true}
+                        className="btn"
+                        onClick={()=>this.isCustomCaseBoxOpen = true}
```
adamabeshouse: for react to not rerender this unnecessarily, the value of onClick should be a defined function, because otherwise the anonymous function gets redefined on every render and always shows up as new props. its  probably not worth the effort here bc its just a single button but just wanted to point it out

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -51,16 +55,34 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
...
+                                            const overlay = () => <span>{`Click ${data.hugoGeneSymbol} to ${_.includes(this.props.selectedGenes, data.hugoGeneSymbol) ? 'remove' : 'add'} from your query`}</span>;
+                                            return (
+                                                <DefaultTooltip
+                                                    placement="left"
+                                                    overlay={overlay}
+                                                    destroyTooltipOnHide={true}
+                                                >
+                                                    <span
+                                                        className={classnames(styles.geneSymbol, _.includes(this.props.selectedGenes, data.hugoGeneSymbol) ? styles.selectd : undefined)}
+                                                        onClick={() => this.props.onGeneSelect(data.hugoGeneSymbol)}>
```
adamabeshouse: see comment above about anonymous functions as props. I think its probably not worth the effort here but just wanted to point it out

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -78,6 +83,27 @@ export class StudyViewPageStore {
 
     private _clinicalAttributesMetaSet: { [id: string]: ChartMeta } = {} as any;
 
+    @observable geneQueryStr: string;
+
+    @observable private geneQueries: SingleGeneQuery[] = [];
+
+    @observable private tableGenes = observable.map<boolean>();
```
adamabeshouse: tableGenes?

PR #1286, opened 2018-07-12
StudyViewScatterPlot, applied to Mutation Count vs CNA

adamabeshouse:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -482,4 +492,78 @@ export class StudyViewPageStore {
...
     });
+
+    readonly mutationCounts = remoteData<MutationCount[]>({
+        await:()=>[
+            this.samples
+        ],
+        invoke:async()=>{
+            const studyToSamples = _.groupBy(this.samples.result!, s=>s.studyId);
+            return _.flatten(await Promise.all(
+                this.studyIds.map(studyId=>{
```
kalletlak: instead of looping `studyIds` may be better to loop `mutationProfiles`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -482,4 +492,78 @@ export class StudyViewPageStore {
...
+        }
+    });
+
+    readonly fractionGenomeAltered = remoteData<FractionGenomeAltered[]>({
+        await:()=>[
+            this.samples
+        ],
+        invoke:async()=>{
+            const studyToSamples = _.groupBy(this.samples.result!, s=>s.studyId);
+            return _.flatten(await Promise.all(
```
kalletlak: may be
```
return _.flatten(await Promise.all(
    _.map(studyToSamples, (samples, studyId) => {
        return internalClient.fetchFractionGenomeAlteredUsingPOST({
            studyId,
            fractionGenomeAlteredFilter: {
                sampleIds: samples.map(s => s.sampleId)
            } as FractionGenomeAlteredFilter
        });
    })
));
```

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: it seems equivalent to me, whats your thought process?

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: sure

src/pages/studyView/StudyViewPageStore.ts
kalletlak: something like
```
return _.flatten(await Promise.all(
    _.map(this.mutationProfiles, (profile) => {
        const samples = studyToSamples[profile.studyId];
        return defaultClient.fetchMutationCountsInMolecularProfileUsingPOST({
            molecularProfileId: profile.molecularProfileId,
            sampleIds: samples.map(s=>s.sampleId)
        });
    })
));
```

adamabeshouse: no I understand but why do you prefer that?

kalletlak: I prefer this way instead of using null and filtering them

adamabeshouse: oh I see. that makes sense to me, i will change

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -482,4 +488,77 @@ export class StudyViewPageStore {
...
+                _.map(studyToSamples, (samples, studyId)=>{
+                    if (samples && samples.length) {
```
kalletlak: I think this condition is not necessary as `studyToSamples` comes from `groupBy`

adamabeshouse: thats true but id rather not couple those

PR #1287, opened 2018-07-13
Create mutational signature tab and table in patient view

jxu8:
New files include clinicalmutationalsignaturetable.ts, mutationalsignature.ts

Modified patientviewpage.tsx, storedutil.ts


src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+        interface IAttrData {
+            [attrId:string]: {
+                [sampleId: string]: ClinicalAttribute | string;
+                clinicalAttribute: ClinicalAttribute;
+                id: string;
+            };
+        };
+
+        interface IConvertedSamplesData {
```
adamabeshouse: pull these interfaces out of the function, so that they can be referenced by other code

src/shared/model/MutationalSignature.ts
```diff
@@ -0,0 +1,8 @@
+export interface IMutationalSignature {
+    sampleId: string;
```
adamabeshouse: this should have `uniqueSampleKey:string` and `uniquePatientKey:string` too

src/shared/model/MutationalSignature.ts
```diff
@@ -0,0 +1,8 @@
...
+    patientId: string;
+    studyId: string;
+    mutationalSignatureId: string;
+    value: string;
```
adamabeshouse: why isnt `value` a number?

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+            columns: IColumn[];
+            items: IAttrData;
+        }
+
+
+        let sampleInvertedDataByMutationalSignature = _(sampleData)
+            .groupBy(x => x.sampleId)
+            .map((value, key)  => ({id: key, clinicalData: value}))
```
adamabeshouse: can you name (value, key) w more descriptive names, or at least type them so its clear what the data is being passed around here?

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+    }
+
+    public invertData(sampleData: IMutationalSignature[]): IConvertedSamplesData{
+        interface IColumn {
+            id:string;
+        };
+
+        interface IAttrData {
+            [attrId:string]: {
+                [sampleId: string]: ClinicalAttribute | string;
```
adamabeshouse: please see my comment below, its messy to mix indexing and named members, they should be separated in the data structure

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+                output.items[clinicalData.mutationalSignatureId].id = clinicalData.mutationalSignatureId;
+            }))
+        })
+
+        return output;
+    }
+
+    public prepareData(sampleData: IMutationalSignature[]) :IMutationalSignatureRow[] {
+
+        interface IMutationalSignatureRow {
```
adamabeshouse: its a little messy to have named members and indexing, even though in this case it will be fine because no samples will have id `mutationalSignatureId`. I still think its cleaner to separate them in the data structure, eg `{ mutationalSignatureId:string, sampleValues:{[uniqueSampleKey:string]:number}`, 

we need `uniqueSampleKey` because `sampleId` may not be unique across studies, only `studyId`,`sampleId` pairs are unique, but pairs are hard to deal with, so we have `uniqueSampleKey` which is unique

also please pull this interface out of the function

zhx828:
@jxu8 I created a mutation signature branch so we can do incremental merges. Once the feature is ready, then we can send a pull request to the rc branch.

src/shared/lib/StoreUtils.ts
```diff
@@ -492,6 +493,12 @@ export function fetchMyCancerGenomeData(): IMyCancerGenomeData
     return geneToMyCancerGenome(data);
 }
 
+export function fetchMutationalSignatureData(): IMutationalSignature[]
+{
+    const data: IMutationalSignature[] = require('../../../resources/samplemutsigdata.json');
```
adamabeshouse: small nitpick, you can save a line and just do `return require(..)`

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.spec.tsx
```diff
@@ -0,0 +1,21 @@
+import * as ClinicalInformationMutationalSignatureTable from './ClinicalInformationMutationalSignatureTable';
+import React from 'react';
+import { assert } from 'chai';
+import { shallow, mount } from 'enzyme';
+import sinon from 'sinon';
+
+describe('ClinicalInformationMutationalSignatureTable', () => {
```
zhx828: Could you add some tests?

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+import styles from './style/patientTable.module.scss';
+import {SHOW_ALL_PAGE_SIZE} from "../../../shared/components/paginationControls/PaginationControls";
+import {IMutationalSignature} from "../../../shared/model/MutationalSignature";
+import convertSamplesData, {IConvertedSamplesData} from "./lib/convertSamplesData";
+import {ISampleRow} from "./ClinicalInformationSamplesTable";
+import _ from 'lodash';
+
+export interface IClinicalInformationMutationalSignatureTableProps {
+    data: IMutationalSignature;
+    showTitleBar?: boolean;
```
zhx828: I think those options should have a default? And should be passed to MutationalSignatureTable.

src/shared/model/MutationalSignature.ts
zhx828: The mock data is string at this moment. @jxu8 we should define the API soon.

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
zhx828: Agreed, I asked James follows the implementation from ClinicalInformationSamplesTable. We probably should do the same thing over there. @jxu8 could you fix ClinicalInformationSamplesTable as well?

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,119 @@
...
+
+interface IMutationalSignatureRow {
+    mutationalSignatureId:string;
+    [key:string]:string;
+};
+
+export default class ClinicalInformationMutationalSignatureTable extends React.Component<{}> {
+    public render() {
+        const sampleData: IMutationalSignature[] = fetchMutationalSignatureData();
+        const sampleInvertedData =  this.invertData(sampleData);
```
zhx828: I don't see a reason to get sampleInvertedData if you just want to get the columns info which is gonna just be  _.map(_.uniq(_.map(mut, 'sampleId')), sampleId=>({id:sampleId}))

package.json
```diff
@@ -43,7 +43,7 @@
     "@types/chai-enzyme": "^0.6.2",
     "@types/chart.js": "^2.4.6",
     "@types/classnames": "0.0.32",
-    "@types/d3-scale": "^2.0.0",
+    "@types/d3-scale": "^2.0.1",
```
zhx828: @adamabeshouse when we do a fresh node install, these dependencies will get updated automatically. This is ok?

src/pages/patientView/clinicalInformation/lib/convertSamplesData.ts
```diff
@@ -7,7 +7,7 @@ export interface IColumn {
 
 export interface IAttrData {
     [attrId:string]: {
-        [sampleId: string]: ClinicalAttribute | string;
+        [sampleId: string]: ClinicalAttribute | string | number;
```
zhx828: Why we change the type?

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,93 @@
...
+        for (const sample of mutationalSignature.samples){
+            mutationalSignatureRowForTable.sampleValues[sample.sampleId] = Number(sample.value.toFixed(4));
+        }
+        tableData.push(mutationalSignatureRowForTable);
+    }
+    return tableData;
+}
+
+export function prepareDataForColumns(mutationalSignatureData: IMutationalSignature[]): IColumns[]{
+    return _.map(_.uniqBy(mutationalSignatureData,"sampleId"), (uniqSample) => ({id: uniqSample.sampleId}));
```
zhx828: This now needs to be grouped by unique sample id. And also other places using sample id as the key.

package.json
adamabeshouse: i think so... as long as tests pass i guess its fine

src/shared/lib/FormatUtils.ts
```diff
@@ -45,3 +45,15 @@ export function toFixedWithThreshold(value: number, digits: number): string
 export function getPercentage(proportion: number, digits: number = 1) {
     return `${toFixedWithThreshold(100 * proportion, digits)}%`;
 }
+
+/* difference between this function and the previous one is it will display
+percentages less than 1% as <1%
+*/
+export function getMutSigPercentage(proportion: number, digits: number = 0) {//0.003 -> 0.3
```
adamabeshouse: MutSig is actually a separate thing, I think its an algorithm or a suite of algorithms... for clarity write out 'Signature'

src/pages/patientView/clinicalInformation/ClinicalInformationMutationalSignatureTable.tsx
```diff
@@ -0,0 +1,94 @@
...
+    return tableData;
+}
+
+export default class ClinicalInformationMutationalSignatureTable extends React.Component<IClinicalInformationMutationalSignatureTableProps, {}> {
+
+    public render() {
+        const uniqueSamples = _.map(_.uniqBy(this.props.data, "uniqueSampleKey"), (uniqSample => ({id: uniqSample.uniqueSampleKey})));
+        const tableData = prepareDataForTable(this.props.data);
+        const firstCol = 'mutationalSignatureId';
+        const columns:Column<IMutationalSignatureRow>[] = [{
```
adamabeshouse: its better to set these columns up outside the `render` method, for two reasons:

(1) (less important here bc its not computationally expensive, but in general something to watch out for) so that you avoid unnecessarily recomputing things on every render. theres nothing in this object that actually needs to be recomputed on every render, these can be a constant and in fact it could really be declared outside the class.

(2) (this is the main reason) react decides whether to rerender things based on if their props have changed. this way, the props to `MutationalSignatureTable` will change every time, because a new javascript object for columns is passed every time. this will lead to unnecessary renders.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -173,6 +174,10 @@ export class PatientViewPageStore {
         return fetchMyCancerGenomeData();
     }
 
+    mutationalSignatureData(): IMutationalSignature[]{
```
adamabeshouse: this should be a `remoteData` - look to the others for example, but it should look like 
`readonly mutationalSignatureData = remoteData({
    invoke:async()=>fetchMutationalSignatureData()
})`

later on when you actually call the API you'll be using this format, so better to just do it now

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -174,9 +174,10 @@ export class PatientViewPageStore {
         return fetchMyCancerGenomeData();
     }
 
-    mutationalSignatureData(): IMutationalSignature[]{
-        return fetchMutationalSignatureData();
-    }
+    readonly mutationalSignatureData = remoteData({
+        invoke: async() => fetchMutationalSignatureData(),
+        default: fetchMutationalSignatureData()
```
adamabeshouse: default isnt necessary here. see comment above for how to use it without default

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -504,7 +504,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                         <div className="clearfix">
                             <FeatureTitle title="Mutational Signatures" isLoading={ patientViewPageStore.clinicalDataGroupedBySample.isPending } className="pull-left" />
                             {  (patientViewPageStore.clinicalDataGroupedBySample.isComplete) && (
```
adamabeshouse: add a check for `mutationalSignatureData` here, so `(patientViewPageStore.clinicalDataGroupedBySample.isComplete && patientViewPageStore.mutationalSignatureData.isComplete) && ...`

PR #1296, opened 2018-07-18
Plots data avail

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -456,6 +455,65 @@ export class ResultsViewPageStore {
...
+            this.molecularProfilesInStudies,
+            this.nonMutationMolecularProfileDataAvailability
+        ],
+        invoke:()=>{
+            const ret:{[entrezGeneId:number]:MolecularProfile[]} = {};
+            const molecularProfilesInStudies = this.molecularProfilesInStudies.result!;
+            const nonMutationMolecularProfileDataAvailability = this.nonMutationMolecularProfileDataAvailability.result!;
+            for (const gene of this.genes.result!) {
+                ret[gene.entrezGeneId] = molecularProfilesInStudies.filter(p=>{
+                    if (p.molecularProfileId in nonMutationMolecularProfileDataAvailability) {
```
alisman: i think you could just return nonMutationMolecularProfileDataAvailability[p.molecularProfileId]===true ?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -186,11 +190,15 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                 this._dataType = t;
             },
             get dataSourceId() {
+                if (this.entrezGeneId === undefined) {
```
alisman: this could all use some commenting to explain relatioship between gene and datasourceId.    in general we need to preserve domain knowledge better in comments.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -427,45 +435,51 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
         }
     }
 
-    @computed get dataTypeOptions() {
-        let dataTypeIds:string[] = [];
+    @computed get geneToDataTypeOptions():{[entrezGeneId:number]:{value:string, label:string}[]} {
```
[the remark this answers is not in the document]
alisman: it's a slipperly slope.  if people see this as example, they will start doing it more in scenarios where it's more inappropriate.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -427,45 +435,51 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
-            ret = Object.assign(ret, _.mapValues(
-                _.groupBy(this.props.store.molecularProfilesInStudies.result, profile=>profile.molecularAlterationType),
-                profiles=>profiles.map(p=>({value:p.molecularProfileId, label:p.name}))
-            ));
-        }
-
-        if (this.clinicalAttributeOptions.length) {
-            ret[CLIN_ATTR_DATA_TYPE] = this.clinicalAttributeOptions;
+    @computed get geneToTypeToDataSourceOptions():{[entrezGeneId:number]:{[type:string]:{value:string, label:string}[]}} {
+        if (this.props.store.nonMutationMolecularProfilesWithData.isComplete) {
```
alisman: ditto above

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -417,6 +417,7 @@ export default class BoxScatterPlot<D extends IBaseBoxScatterPlotPoint> extends
 
     render() {
         if (!this.props.data.length) {
+            setTimeout(()=>this.svgRef(null), 0);
```
alisman: could you just put this in componentDidUpdate hook?

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: that's not the data structure but you're right it would be clearer to return nonMutation... > 0

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -167,13 +167,17 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
             },
             get dataType() {
```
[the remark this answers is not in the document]
adamabeshouse: I feel like a test would basically just mimic the same branching logic. i'd rather just cover it with screenshots

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: good point

src/pages/resultsView/plots/PlotsTab.tsx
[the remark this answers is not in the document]
adamabeshouse: I dont see how the memoized method/pure function doesnt do the same? either way we need to funnel the mobxpromise into the @computed

I dont see it as slippery slope, this is a particular case where status doesnt matter to the consumer, and code would be a little more complicated if we put the status checks elsewhere

src/shared/components/plots/BoxScatterPlot.tsx
adamabeshouse: I think I prefer this way for the reason that otherwise we'd have to duplicate the check `if(!this.props.data.length)` and maintain that check if the render logic changes at all. or otherwise create a new variable like "lastRenderWasEmpty" and use that in `componentDidUpdate` to be completely semantically robust.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: actually I think we want to keep this check but I'll add a comment to explain it

src/shared/components/plots/BoxScatterPlot.tsx
alisman: i'm really against this kind of thing when it can be achieved using the normal lifecycle.  what about just checking the data in getter for svgRef?

adamabeshouse: svgRef needs to be triggered exactly when it changes. I'm not sure how to accomplish that cleanly with manual checks. can you explain more?

src/pages/resultsView/plots/PlotsTab.tsx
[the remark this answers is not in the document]
alisman: memoized function is different because result is passed in directly (de-referenced).  status is checked outside in view, which is where the component can properly be hidden if its data is pending.  the basic principle we are trying to enforce is, don't show UI when its data dependencies are loading.   making a rule about the above is a good way of enforcing that

PR #1297, opened 2018-07-19
Collapse DownloadControls into hoverable icon

adamabeshouse:


alisman:
looks good to me.    my only wonder is if it would be better to implement as dropdown.  the way you've done this (and i suggested this), if this element is not absolutely positioned, it will bump other elements.   a dropdown (or flyout) avoids this problem because the dropdown itself is postioned absolutely and so can't affect position of other elements.

adamabeshouse:
I dont understand what you're saying - the expansion is in a tooltip, which doesnt bump other elements

src/shared/components/downloadControls/DownloadControls.css
```diff
@@ -0,0 +1,16 @@
+.menu-item {
```
alisman: these need to be wrapped in .download-controls b/c they are global now

src/shared/components/downloadControls/DownloadControls.css
```diff
@@ -0,0 +1,16 @@
...
+    padding:5px;
+    user-select:none;
+}
+
+.menu-item.menu-item-enabled:hover {
+    cursor:pointer;
+    text-decoration:underline;
+}
+
+.menu-item.menu-item-disabled {
```
alisman: what's disabled for?

adamabeshouse: disabled is one of the options you can set

adamabeshouse:
@alisman which tabs should we collapse for?

PR #1298, opened 2018-07-19
Include grid layout library and layout pie chart based on priority [ do not merge ]

zhx828:
There are few things need to be done after this pull request:
1. Include all charts under ChartContainer
2. Update the layout algorithm based on [here](https://github.com/cBioPortal/cbioportal/blob/master/docs/Study-View.md)


src/shared/components/window/WindowStore.ts
```diff
@@ -0,0 +1,32 @@
...
+        this.size = {
+            width: this.window.innerWidth,
+            height: this.window.innerHeight
+        };
+    }
+
+    private handleWindowResize = _.debounce(this.getWindowSize, 1000);
+
+    @observable size: WindowSize
+    @observable window: any;
```
alisman: i wouldn't do this because you don't want to let mobx loose on actually window object

alisman: i don't think there's a reason to make it observable

src/shared/components/window/WindowStore.ts
```diff
@@ -0,0 +1,32 @@
...
+    constructor() {
+        if (typeof window === 'object') {
+            this.window = window;
+            this.getWindowSize();
+            this.window.addEventListener("resize", this.handleWindowResize)
+        }
+    }
+
+    @action
+    private getWindowSize() {
```
alisman: this probably needs to be bound (@auto-bind) to work.  my guess is that's what's wrong.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -102,6 +103,140 @@ export function mutationCountVsCnaTooltip(d:{ data:Pick<IStudyViewScatterPlotDat
...
+ * Calculate the layout used by react-grid-layout
+ *
+ * @param {ChartMeta[]} visibleAttributes
+ * @param {number} cols number of grids per row, 6 cols will be the stander when using 13 inch laptop
+ * @param chartsDimension
+ * @returns {ReactGridLayout.Layout[]}
+ */
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, chartsDimension: any): Layout[] {
+    let matrix: layoutMatrixItem[] = [];
+    _.sortBy(visibleAttributes, [(attr: ChartMeta) => attr.clinicalAttribute.priority])
```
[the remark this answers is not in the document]
zhx828: True, but when resizing the browser, it's quite difficult to preserve the previous position.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -77,8 +132,24 @@ export class StudyViewPageStore {
 
     @observable private _chartVisibility = observable.map<boolean>();
 
+    @observable private chartsDimension = observable.map<ChartDimension>();
```
alisman: should this have _?   seems there's a convention here for private

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -102,6 +109,140 @@ export function mutationCountVsCnaTooltip(d:{ data:Pick<IStudyViewScatterPlotDat
...
+
+/**
+ * Calculate the layout used by react-grid-layout
+ *
+ * @param {ChartMeta[]} visibleAttributes
+ * @param {number} cols number of grids per row, 6 cols will be the stander when using 13 inch laptop
+ * @param chartsDimension
+ * @returns {ReactGridLayout.Layout[]}
+ */
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, chartsDimension: any): Layout[] {
```
alisman: seems wouldn't be hard to type chartDimension

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -102,6 +109,140 @@ export function mutationCountVsCnaTooltip(d:{ data:Pick<IStudyViewScatterPlotDat
...
+    let matrix: LayoutMatrixItem[] = [];
+    _.sortBy(visibleAttributes, [(attr: ChartMeta) => attr.clinicalAttribute.priority])
+        .forEach((attr, index) => {
```
alisman: _.reduce because then it's more clear what's going on.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -102,6 +109,140 @@ export function mutationCountVsCnaTooltip(d:{ data:Pick<IStudyViewScatterPlotDat
...
+/**
+ * Calculate the layout used by react-grid-layout
+ *
+ * @param {ChartMeta[]} visibleAttributes
+ * @param {number} cols number of grids per row, 6 cols will be the stander when using 13 inch laptop
+ * @param chartsDimension
+ * @returns {ReactGridLayout.Layout[]}
+ */
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, chartsDimension: any): Layout[] {
+    let matrix: LayoutMatrixItem[] = [];
```
alisman: matrixGroups?  matrices?

src/pages/studyView/styles.module.scss
```diff
@@ -18,4 +18,14 @@
...
 .studyViewSurvivalPlotBody {
     position: relative
+}
+
+.studyViewChartContainer {
+    border: 1px solid #d3d3d3
+}
+
+.studyViewGridItem {
+    .react-grid-item:hover {
```
alisman: fyi, i don't think nesting of classes will work the way you might expect in css modules.   react-grid-item will also get "uniqueified" and need to be applied in js.  it will not apply contextually.   if i'm right, you may want to use :global() to achieve contextual application.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -258,13 +339,24 @@ export class StudyViewPageStore {
...
-                    acc.push(chartMeta)
+                    acc.push(chartMeta);
+                    this.chartsDimension.set(chartMeta.clinicalAttribute.clinicalAttributeId, chartMeta.dimension);
                 }
             }
             return acc;
         }, []);
     }
 
+    public changeChartType(attr: ChartMeta, newChartType: ChartType) {
```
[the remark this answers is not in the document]
zhx828: Calling an action outside of the class?

PR #1299, opened 2018-07-19
Adjust tooltips in survival and expression to be moused over more easily

adamabeshouse:


alisman:
@adamabeshouse can you describe the solution here?

adamabeshouse:
@alisman sure: as I said in the linked issue, I think the tooltips are only unusable in those tabs because the tooltip is not showing up close enough to the mouse, so its not possible to roll the mouse over the tooltip because theres too much space in between for a mouseout to register. so all I'm doing in this PR is moving those tooltips closer to the mouse (and making an adjustment to survival tab to make it work aesthetically with the hover circles)

alisman:
@adamabeshouse nice work.  should we do this as a hotfix (to master)?

PR #1302, opened 2018-07-20
Enable selecting custom backend on heroku

inodb:
Enable selection of a custom backend on heroku by changing env variables. Also, heroku was broken in general (backend api url was always unknown). This makes the standalone work again


Procfile
```diff
@@ -1 +1 @@
-web: http-server --cors dist/ -p ${PORT}
+web: ls dist.backup || cp -a dist/ dist.backup; cp -a dist.backup dist; for f in $(find dist/); do sed -i "s|replace_me_env_cbioportal_url|${CBIOPORTAL_URL:-http\://www.cbioportal.org}|g" $f; sed -i "s|replace_me_env_genome_nexus_url|${GENOME_NEXUS_URL:-https\://www.genomenexus.org}|g" $f; done && http-server --cors dist/ -p ${PORT}
```
inodb: make copy of dist folder if it doesn't exist yet. On every run copy that dist.backup folder back and replace the variable `replace_me_env_cbioportal_url` with CBIOPORTAL_URL. Same for GENOME_NEXUS_URL. This allows us to change the backend w/o recompiling

inodb: it's still a bit funky sometimes b/c there's no cache buster in place with build time of frontend, so need to do hard refresh after PR is updated

PR #1319, opened 2018-07-27
Include Significant Mutational Signatures Component in Patient View Header

jxu8:
New component is based on pull request for GDD: 
https://github.com/cBioPortal/cbioportal-frontend/pull/724

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -252,7 +252,10 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                             }
                         </span>
 
-                        <SignificantMutationalSignatures data={patientViewPageStore.mutationalSignatureData.result}/>
+                        {patientViewPageStore.clinicalDataGroupedBySample.isComplete && patientViewPageStore.mutationalSignatureData.isComplete
+                        && patientViewPageStore.mutationalSignatureMetaData.isComplete && patientViewPageStore.hasMutationalSignatureData.isComplete &&
+                        (<SignificantMutationalSignatures data={patientViewPageStore.mutationalSignatureData.result}
+                        metadata={patientViewPageStore.mutationalSignatureMetaData.result} uniqueSampleKey={sample.id}/>)}
 
                     </div>
                 );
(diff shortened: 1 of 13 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -247,6 +251,12 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                 )
                             }
                         </span>
+
+                        {patientViewPageStore.clinicalDataGroupedBySample.isComplete && patientViewPageStore.mutationalSignatureData.isComplete
```
adamabeshouse: maybe a LoadingIndicator component could be useful here?

src/pages/patientView/patientHeader/SignificantMutationalSignatures.tsx
```diff
@@ -0,0 +1,143 @@
...
+                        <td style={{paddingTop: 3}}>{this.progressBar(significantMutationalSignaturesForSample.significantSignatures[significantSignatureValue].toString(), false)}</td>
+                        </tr>)}
+                    </table>
+                }
+
+            </div>
+        );
+    }
+
+    public progressBar(confidence:string, determineColor:boolean) {
```
adamabeshouse: this doesnt need anything in the class right? like it doesn't reference `this`. so I dont think it needs to be an instance method. you can move it out as a helper method, maybe into a file called `MutationalSignaturesUtils.tsx`

src/pages/patientView/patientHeader/SignificantMutationalSignatures.tsx
```diff
@@ -0,0 +1,143 @@
...
+                    textAlign:"center"}}
+                >{confidencePerc}%</div>
+            </div>
+        );
+    }
+
+}
+
+export function prepareMutationalSignaturesForHeader(mutationalSignatureData: IMutationalSignature[], mutationalSignatureMetaData: IMutationalSignatureMeta[],
+                                                     uniqueSampleKey: string): ISignificantMutationalSignaturesForSample{
```
adamabeshouse: the way we've been organizing code in the frontend is to move stuff like this into `MutationalSignaturesUtils`

jxu8: Ok, sounds good! I can move the equivalent function for the mutational signature table in there as well for organization.

PR #1330, opened 2018-07-30
new oncoprint setting - only show clinical track legends for altered cases

adamabeshouse:
still need to update package.json reference to upcoming new oncoprintjs version


adamabeshouse:
replaced https://github.com/cBioPortal/cbioportal-frontend/pull/1366

PR #1335, opened 2018-07-31
Only show No Oql warning banner with complex oql query

adamabeshouse:


src/shared/components/NoOqlWarning.tsx
```diff
@@ -1,16 +1,23 @@
...
-        This tab does not reflect the OQL specification from your query.
-    </span>
-);
-
-export default NoOqlWarning;
\ No newline at end of file
+@observer
+export default class NoOqlWarning extends React.Component<{store:ResultsViewPageStore}, {}> {
+    render() {
+        if (this.props.store.queryContainsOql) {
```
alisman: would rather see this outside in parent view, but no biggie

adamabeshouse: just wanted to simplify use since it would always be the same

PR #1340, opened 2018-07-31
plots tab - Fix bug when selecting a gene with profile that its not available for

adamabeshouse:


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -383,19 +389,21 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
             otherSelection = this.horzSelection;
         } else {
             targetSelection = this.horzSelection;
             otherSelection = this.vertSelection;
         }
-        targetSelection.entrezGeneId = entrezGeneId;
-        if (this.geneLock) {
-            otherSelection.entrezGeneId = entrezGeneId;
-        }
+        runInAction(()=>{
```
alisman: this is already an action, so why why runInAction?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -184,7 +184,13 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                 }
             },
             set entrezGeneId(e:number|undefined) {
-                this._entrezGeneId = e;
+                runInAction(()=>{
+                    if (e !== this._entrezGeneId) {
```
alisman: my guess about what's happening here is, if you change gene, you want to clear the other selections?   if that's case, maybe what we need is "clear" routine, which gets called when you change gene.  this would encapsulate the logic for clearing, make it reusable, more readable.   i know we're working fast now but this might be worthy of test?

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: is every set in an observable automatically an action?

adamabeshouse: oh oops i got confused

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: I see what you're saying. but I don't think this would be reused and its too simple to test. do you think a comment be enough?

alisman: sure

adamabeshouse:
we're not filtering by data availability now so this bug wont happen anymore

PR #1342, opened 2018-07-31
Update react-select-checked, and use fontawesome in NoOqlWarning

adamabeshouse:


alisman:
@adamabeshouse looks like this needs screenshots updated

PR #1348, opened 2018-08-01
Dirty fix to get 3d structures to work again

inodb:
Somehow the width doesn't get updated properly - hardcoding the width solves it for now, but need to get a proper solution


inodb:
Related to: https://github.com/cBioPortal/cbioportal/issues/4561

jjgao:
@inodb thanks. We can go with this for the release and @onursumer can fix it properly later.

PR #1361, opened 2018-08-07
Structural variant data format

rnugraha:
# What? Why?
This PR will display a new tab called Fusion tab to display structural variant (fusion) data also to visualise the queried gene model and the different breakpoints found. Changes are based on RFC 31_b and have a dependency on PR [cbioportal#4057](https://github.com/cBioPortal/cbioportal/pull/4057)
 
- Display Fusion tab when queried studies contain fusion data
- Display the data in the table and also visualise the breakpoints and its pfam domain
- Visualisation is interactive, enable the user to filter the data via the pie chart and get additional information from the tooltips.

**RFC**: [RFC 31_b](https://docs.google.com/document/d/1tzSW2iB59bcns4xNCSJjMTxaX2uavCILDD4lo3HwcE4/edit#heading=h.t10nr9te3qvu)

# Checks
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation/data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!
- [x] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases, multiple topical commits can be useful. During the review, it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
  the `-s` option)

# Any screenshots or GIFs?
fusiongif

# Notify reviewers
@cBioPortal/frontend

and notify them here:


alisman:
@rnugraha @oplantalech what's status of this PR?   i'm going to mark it "do not merge" 
if in fact, it's just waiting on review from my team, we can remove that tag and schedule release.  thanks!

oplantalech:
@alisman we are waiting until the backend PR is merged (hopefully soon), then I will rebase this one and ask for reviews.

Backend PR is in: cbioportal#4057

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -935,7 +938,7 @@ export class ResultsViewPageStore {
                         }
                     }));*/
                     ret.push(profile);
-                } else {
+                } else if (profile.molecularAlterationType !== AlterationTypeConstants.STRUCTURAL_VARIANT) { //All studies except structural variants
```
alisman: i think this comment is mean to be refer to "profile" not study?

alisman: we should probably refactor this whole if statement as switch.  would be more readable.  not too much work.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+        ],
+        invoke: async () => {
+            let molecularProfiles: Array<string> = [];
+            let samples: SampleMolecularIdentifier[] = [];
+            let params: StructuralVariantFilterExt = {entrezGeneIds:[]};
+            if (this.genes.result) {
+                params.entrezGeneIds = _.map(this.genes.result, (gene: Gene) => gene.entrezGeneId);
+            }
+            // compose request parameters
+            if (this.samples.result) {
```
alisman: we should put this logic in helper function so it can be tested (and also so it won't bloat the store.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+    readonly fusions = remoteData<StructuralVariant[]>({
+        await: () => [
+            this.genes,
+            this.selectedMolecularProfiles,
+            this.samples,
+            this.studyIdToStudy
+        ],
+        invoke: async () => {
+            let molecularProfiles: Array<string> = [];
+            let samples: SampleMolecularIdentifier[] = [];
```
alisman: i don't think we need initliazation of samples up here (nor empty array assignment) because it's only used inside if statement and is assigned to result of map.

alisman: same is true of molecularProfiles.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+            let params: StructuralVariantFilterExt = {entrezGeneIds:[]};
+            if (this.genes.result) {
+                params.entrezGeneIds = _.map(this.genes.result, (gene: Gene) => gene.entrezGeneId);
+            }
+            // compose request parameters
+            if (this.samples.result) {
+                // get sample molecular identifiers
+                samples = this.samples.result.map(sample => {
+                    return {
+                        molecularProfileId: sample.studyId + '_structural_variants',
```
alisman: we should put this in helper function so that the suffix (_structural_variants) is in once place.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+                if (this.selectedMolecularProfiles.result) {
+                    molecularProfiles = this.selectedMolecularProfiles.result.filter(profile => {
+                        return profile.molecularAlterationType === AlterationTypeConstants.STRUCTURAL_VARIANT;
+                    }).map(profile => {
+                        return profile.molecularProfileId;
+                    });
+                    params.molecularProfileIds = molecularProfiles;
+                }
+            }
+            return await client.fetchStructuralVariantsUsingPOST({'structuralVariantFilter':<StructuralVariantFilter> params});
```
alisman: await is only necessary when you want to assign the asyncronous result to a local variable and "pretend" to program sychronously.   here, you are just returning a promise, so you can take the async tag off the invoke function and just return client.fetch___

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+        }
+    });
+
+    /**
```
alisman: these interface comments seem redundant given that the typescript annotation says the same thing.  are the comments being auto-generated?  textual comments that explain the function are great though.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+     * Group fusion data by genes.
+     * @returns {{[p: string]: StructuralVariant[]}}
+     */
+    @computed
+    get fusionsByGene(): { [hugeGeneSymbol: string]: StructuralVariant[] } {
+        let genes:Array<string> = [];
```
alisman: not necessary

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+            }
+        }
+    }, {});
+
+    /**
+     * Get fusion mapper store by gene
+     * @param {string} hugoGeneSymbol
+     * @returns {ResultViewFusionMapperStore}
+     */
+    public getFusionMapperStore(hugoGeneSymbol: string): ResultViewFusionMapperStore | undefined {
```
alisman: this function hides the fact that stores may not exist until genes are loaded.   might be clearer just to reference the fusionMapperStores.result directly in code because developer will then understand they need to handle the loading status.  i don't think this function does enough to warrant the potential confusion it could cause.  (this is common issue in our code base)

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,106 @@
...
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ **/
+
+import * as React from 'react';
+import { observer } from "mobx-react";
+import { ResultsViewPageStore } from '../ResultsViewPageStore';
+import { observable } from 'mobx';
+import { MSKTab, MSKTabs } from '../../../shared/components/MSKTabs/MSKTabs';
+import Loader from '../../../shared/components/loadingIndicator/LoadingIndicator';
```
alisman: I think we call it LoadingIndicator elsewhere so lets keep that convention so people know it's the the same thing

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,106 @@
...
+}
+
+@observer
+export default class Fusions extends React.Component<IFusionPageProps, {}> {
+
+    @observable fusionGeneTab: string;
+
+    constructor(props: IFusionPageProps) {
+        super(props);
+        this.handleTabChange.bind(this);
```
alisman: use autobind decorator

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,106 @@
...
+        this.fusionGeneTab = this.props.store.hugoGeneSymbols[0];
+    }
+
+    public render() {
+        // use routing if available, if not fall back to the observable variable
+        const activeTabId = this.props.routing ?
+            this.props.routing.location.query.fuionsGeneTab : this.fusionGeneTab;
```
alisman: would probably be better to keep the routing store out of the component.  you could just make a property activeTabId and set it to routing store at page component.

alisman: we should also probably keep the selected tab OUT of the session, which is unfortunately kind of confused with the "query string" right now on the resultsView page.  I.e. we don't want it memoized in the session because that would mean creating a new session every time user switches tab.  the right solution is probably to have deeper level path.   e.g. results/fusion/:fusionTab.    lets discuss how to achieve this.

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+export interface IFusionMapperProps {
+    store: ResultViewFusionMapperStore;
+}
+
+@observer
+export default class ResultViewFusionMapper extends React.Component<IFusionMapperProps, {}> {
+
+    private renderComponents(studyResult: { [studyId: string]: CancerStudy },
+                             molecularProfile: { [molecularProfileId: string]: MolecularProfile }) {
+        const exonChartStore = new ExonsChartStore(
```
alisman: We want to try to limit stores at "local" level, kept in component heirarchy because it creates possibility of duplicating state higher up in the application.  if local state is necessary, better to do some with observable local members/props of components, which is analagous to React's component state paradigm.

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+            this.props.store.ensemblTranscripts.isPending
+        );
+        const store = this.props.store.dataStore;
+        const occurence = this.props.store.fusionOccurrence;
+
+        return (
+            <div>
+                <div style={{display: 'flex', flexWrap: 'wrap'}}>
+                    <ExonsChart store={exonChartStore}/>
+                    <FusionPieChart dataStore={store} fusionCounts={occurence}/>
```
alisman: "dataStore" is kind of redundant.  store = data.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -223,6 +225,23 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                 }
             },
 
+            {
+                id:ResultsViewTab.FUSION,
+                hide:()=>{
```
adamabeshouse: !this.resultsViewPageStore.fusions.isComplete || this.resultsViewPageStore.fusions.result.length === 0

src/pages/resultsView/ResultsViewPageStore.spec.ts
```diff
@@ -158,3 +160,134 @@ describe('extendSamplesWithCancerType', () => {
...
+        store = new ResultsViewPageStore(appStore, '');
+        sinon.stub(store, 'fusions').callsFake(function fakeFn() {
+            return dummyFusions;
+        });
+        sinon.stub(store, 'genes').callsFake(function fakeFn() {
+            return dummyGenes;
+        });
+    });
+
+    it('should grouped fusions by queried gene');
```
adamabeshouse: needs tests

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: I think you mean `getStructuralVariantProfile(studyId)` and I agree

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: yes I agree. I think you don't even need the variables because they're immediately used and not used again - just set params.X = ...

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: I agree with @alisman I personally think it's harder to read and clutters the code and it creates maintenance burden, mostly duplicates work that should properly be done by the param names and type names. Any further function documentation can be put in a standard comment inside the body of the function

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+            // get the genes
+            genes = this.genes.result.map(gene => gene.hugoGeneSymbol);
+            // group fusion data by the genes from input query
+            genes.forEach(gene => {
+                let tmpFusions:Array<any> = [];
+                if (this.fusions.result) {
+                    tmpFusions = tmpFusions.concat(this.fusions.result.filter((fusion:StructuralVariant) => {
+                        return fusion.site1HugoSymbol === gene || fusion.site2HugoSymbol === gene;
+                    }));
+                    fusionsByGene = Object.assign(fusionsByGene, { [gene]: tmpFusions });
```
adamabeshouse: this seems like a complicated way of writing `fusionsByGene[gene] = tmpFusions`, unless I'm missing something?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+            return await client.fetchStructuralVariantsUsingPOST({'structuralVariantFilter':<StructuralVariantFilter> params});
+        }
+    });
+
+    /**
+     * Group fusion data by genes.
+     * @returns {{[p: string]: StructuralVariant[]}}
+     */
+    @computed
+    get fusionsByGene(): { [hugeGeneSymbol: string]: StructuralVariant[] } {
```
adamabeshouse: Often it makes most sense in a case like this, where the result depends on a promise being ready, that you turn this into a mobxpromise

```
readonly fusionsByGene = remoteData({
    await:()=>[this.genes],
   ...
```

adamabeshouse: based on use below it seems this would be the best way to do it

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+        return fusionsByGene;
+    }
+
+    /**
+     * Assign a FusionMapperStore per genes
+     * @type {MobxPromiseUnionTypeWithDefault<{[p: string]: FusionMapperStore}>}
+     */
+    readonly fusionMapperStores = remoteData<{[hugoGeneSymbol: string]: ResultViewFusionMapperStore}>({
+        invoke: () => {
+            if (this.genes.result && this.fusionsByGene) {
```
adamabeshouse: this is a big no-no - if you use a mobxpromise inside another mobxpromise you must await it.

`await:()=>[this.genes]` (and `this.fusionsByGene` when you change it)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1630,6 +1633,112 @@ export class ResultsViewPageStore {
...
+                return Promise.resolve(
+                    this.genes.result.reduce((map: { [hugoGeneSymbol: string]: ResultViewFusionMapperStore }, gene: Gene) => {
+                        map[gene.hugoGeneSymbol] = new ResultViewFusionMapperStore(
+                            gene,
+                            this.studyIdToStudy,
+                            this.molecularProfileIdToMolecularProfile,
+                            this.samples,
```
adamabeshouse: all of these need to be awaited - samples, molecularProfileIdToMolecularProfile, studyIdToStudy.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: I agree but also see that it's confusing from @rnugraha POV because this is how we do the mutation tab (from earlier in our work when we didn't have the idioms nailed down)

src/pages/resultsView/fusion/Fusions.tsx
adamabeshouse: agreed. also theres a typo `fuionsGeneTab`

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,106 @@
...
+
+        return (
```
adamabeshouse: in keeping with new conventions in our codebase, lets use `MakeMobxView` for this. it's straightforward to use, look for example at `groupComparison/MutationEnrichments.tsx`

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+    private renderComponents(studyResult: { [studyId: string]: CancerStudy },
+                             molecularProfile: { [molecularProfileId: string]: MolecularProfile }) {
+        const exonChartStore = new ExonsChartStore(
+            this.props.store.gene,
+            this.props.store.dataStore,
+            this.props.store.processedEnsembleTranscripts,
+            studyResult,
+            this.props.store.processedPfamDomains,
+            this.props.store.ensemblTranscripts.isPending
```
adamabeshouse: seems weird that you would want to pass in the pending state of this promise, rather than just the promise itself and then react to its status inside?

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+                />
+            </div>);
+    }
+
+    /**
+     * Render fusion table but display loading indicator when fusion data is still pending
+     * @returns {any}
+     */
+    public render() {
+        const studyResult = this.props.store.studyIdToStudy.result;
```
adamabeshouse: lets use `MakeMobxView` here too

src/pages/resultsView/fusion/ResultViewFusionMapperDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+type ImmutablePositionAttr = PositionAttr & Immutable.ImmutableObject<PositionAttr>;
+
+export default class FusionMapperDataStore extends SimpleLazyMobXTableApplicationDataStore<StructuralVariant[]> {
+    @observable.ref private selectedPositions: ImmutablePositionAttr;
+    @observable.ref private highlightedPositions: ImmutablePositionAttr;
+
+    @action
+    public setPositionSelected(position: number, newVal: boolean) {
+        const toMerge: PositionAttr = {};
+        toMerge[position + ""] = newVal;
```
adamabeshouse: I like this trick as well but @alisman convinced me `.toString()` is cleaner

src/pages/resultsView/fusion/ResultViewFusionMapperDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+        this.selectedPositions = this.selectedPositions.merge(toMerge) as ImmutablePositionAttr;
+    }
+
+    @action
+    public setPositionHighlighted(position: number, newVal: boolean) {
+        const toMerge: PositionAttr = {};
+        toMerge[position + ""] = newVal;
```
adamabeshouse: `toString`

src/pages/resultsView/fusion/ResultViewFusionMapperDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+import { StructuralVariant } from "../../../shared/api/generated/CBioPortalAPI";
+import { action, observable, computed } from "mobx";
+import Immutable from "seamless-immutable";
+import { SimpleLazyMobXTableApplicationDataStore } from '../../../shared/lib/ILazyMobXTableApplicationDataStore';
+
+type PositionAttr = { [position: string]: boolean };
+type ImmutablePositionAttr = PositionAttr & Immutable.ImmutableObject<PositionAttr>;
+
+export default class FusionMapperDataStore extends SimpleLazyMobXTableApplicationDataStore<StructuralVariant[]> {
+    @observable.ref private selectedPositions: ImmutablePositionAttr;
```
adamabeshouse: instead of using immutables, a good way (and more typical in our codebase) of doing this is using mobx `observable.map`

adamabeshouse: oh I see now that you just copied the way we do it in MutationMapperDataStore. Okay I think you should just leave it then, sorry.

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+
+    /**
+     * Returns fusion data with label added as new attribute composed of site1 and site2 genes.
+     * @type {MobxPromiseUnionTypeWithDefault<StructuralVariantExt[]>}
+     */
+    readonly fusionData = remoteData({
+        invoke: async () => {
+            return (<StructuralVariantExt[]> this.fusions || []).map(f => {
+                // generate fusion label made of site1 and site2 genes
+                f.label = `${f.site1HugoSymbol}-${f.site2HugoSymbol}`;
```
adamabeshouse: I think its better to use an object for this because gene names can have hyphens in them. its the most edgey of edge case but I don't think its a big deal and its better safe than sorry. what do you think?

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+        }
+    }, []);
+
+    /**
+     * Apply color scale to the transcript
+     * @param {EnsemblTranscriptExt} d
+     * @param {boolean} isRefGene
+     * @param {EnsemblTranscriptExt[]} tempArr
+     */
+    private applyColor(d: EnsemblTranscriptExt, isRefGene: boolean, tempArr: EnsemblTranscriptExt[]) {
```
adamabeshouse: lets use a more descriptive name for `tempArr`

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+    readonly ensemblTranscripts = remoteData({
+        await: () => [
+            this.fusionData,
+        ],
+        invoke: async () => {
+            // Collect all transcript id's from the fusion data
+            let transcriptIds: any = [];
+            if (this.fusionData.result) {
+                this.fusionData.result.forEach((fusion: StructuralVariant) => {
+                    if (transcriptIds.indexOf(fusion.site1EnsemblTranscriptId) < 0) {
```
adamabeshouse: in this case we should really use a set, because the repeated indexOfs gives quadratic time

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+                        transcriptIds.push(fusion.site1EnsemblTranscriptId);
+                    }
+                    if (transcriptIds.indexOf(fusion.site2EnsemblTranscriptId) < 0) {
+                        transcriptIds.push(fusion.site2EnsemblTranscriptId);
+                    }
+                });
+            }
+            // Fetch ensembl transcript data with transcript ids as request params
+            return await fetchEnsemblTranscriptsByEnsemblFilter({'transcriptIds': transcriptIds})
```
adamabeshouse: as mentioned by @alisman above, async and await are not necessary here

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
adamabeshouse: `applyColor` is also a bit too vague, how about `assignColorScaleToTranscript`

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+        ],
+        invoke: async () => {
+            if (this.ensemblTranscripts.result) {
+                // get all pfam domain ids from requested transcripts
+                let pFamDomainIds = this.ensemblTranscripts.result
+                    .map(t => t.pfamDomains)
+                    .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
+                    .map(t => t.pfamDomainId);
+                // fetch the data with pfam domain ids as request params
+                return await fetchPfamDomainData(pFamDomainIds);
```
adamabeshouse: async and await are not necessary

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+    /**
+     * Fetch pfam domain data
+     * @type {MobxPromiseUnionTypeWithDefault<PfamDomain[]>}
+     */
+    readonly pfamDomainData = remoteData<PfamDomain[] | undefined>({
+        await: () => [
+            this.ensemblTranscripts
+        ],
+        invoke: async () => {
+            if (this.ensemblTranscripts.result) {
```
adamabeshouse: you know ensemblTranscripts is ready because you've awaited it, so this if is not necessary

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,212 @@
...
+        return _.countBy(this.fusionData.result, (d) => d.label);
+    }
+
+    /**
+     * Get pfam domain data
+     * @returns {PfamDomain[]}
+     */
+    @computed
+    get processedPfamDomains(): PfamDomain[] {
+        return this.pfamDomainData.result || [];
```
adamabeshouse: whats the purpose of this? users should await pfamDomainData, or pfamDomainData should have default value set of []

adamabeshouse: same goes for `processedEnsembleTranscripts`

src/pages/resultsView/fusion/ResultsViewFusionTable.tsx
```diff
@@ -0,0 +1,148 @@
...
+
+export interface IResultsViewFusionTableProps extends IFusionTableProps {
+    // add results view specific props here if needed
+}
+
+@observer
+export default class ResultsViewFusionTable extends FusionTable<IResultsViewFusionTableProps> {
+
+    constructor(props: IResultsViewFusionTableProps) {
+        super(props);
```
adamabeshouse: unnecessary constructor

src/shared/components/fusionTable/IFusionTableProps.tsx
```diff
@@ -0,0 +1,43 @@
...
+ **/
+
+import { FusionTableColumnType } from './FusionTable';
+import { CancerStudy, MolecularProfile, StructuralVariant } from '../../api/generated/CBioPortalAPI';
+import { SortDirection } from '../lazyMobXTable/LazyMobXTable';
+import { IPaginationControlsProps } from '../paginationControls/PaginationControls';
+import { ILazyMobXTableApplicationLazyDownloadDataFetcher } from '../../lib/ILazyMobXTableApplicationLazyDownloadDataFetcher';
+import { ILazyMobXTableApplicationDataStore } from '../../lib/ILazyMobXTableApplicationDataStore';
+
+export interface IFusionTableProps {
```
adamabeshouse: we don't put props in different files, just put it in FusionTable.tsx

src/shared/components/exonsCharts/ExonsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+            let leftLabel = {
+                label: fusion.site1HugoSymbol + '-' + fusion.site2HugoSymbol,
+                isReference: false
+            };
+            let _barPlotStore =  new ExonsBarPlotStore(leftLabel, fusion, referenceGeneWidth);
+            return <ExonBarPlot store={_barPlotStore}/>;
+        });
+    }
+
+    renderReferenceGenes() {
```
adamabeshouse: `@computed get` so it doesnt recompute unnecessarily

src/shared/components/exonsCharts/ExonsChart.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { EnsemblTranscriptExt} from '../../model/Fusion';
+import { StructuralVariantExt } from '../../model/Fusion';
+
+export interface IExonsChartProps {
+    store: ExonsChartStore;
+}
+
+@observer
+export default class ExonsChart extends React.Component<IExonsChartProps, {}> {
+    renderBarPlots(fusions: StructuralVariantExt[], referenceGeneWidth: number) {
```
adamabeshouse: `@computed get`

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+    getExonsBySite(siteId: number, transcriptId: string, breakpoint: number): Array<ExonRangeExt> {
+        let exons: Array<ExonRangeExt> = [];
+        if (siteId === 1 || siteId === 2) {
+            let _t = this.getTranscriptById(transcriptId)[0];
+            let _sortedExons = <ExonRangeExt[]> _t.exons.sort(
+                // sort exons by rank
+                (exon1: ExonRangeExt, exon2: ExonRangeExt) => exon1.rank - exon2.rank
+            );
+            exons =
+                siteId < 2
```
adamabeshouse: you're inside `siteId === 1 || siteId === 2`, so i think this condition should be `siteId === 1`

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+                    ? _sortedExons.filter(exon => exon.rank <= breakpoint)
+                    : _sortedExons.filter(exon => exon.rank >= breakpoint);
+        }
+        return exons;
+    }
+
+    getTranscriptById(transcriptId: string): Array<EnsemblTranscriptExt> {
```
adamabeshouse: getTranscript*s*ById

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+        return (this.transcripts || []).filter(d => {
+            return d.transcriptId === transcriptId;
+        });
+    }
+
+    @computed
+    get computedFusions(): StructuralVariantExt[] {
+        return (this.fusionDataStore || {tableData: []}).tableData.map((f: StructuralVariantExt[]) => {
```
adamabeshouse: why would fusionDataStore be undefined? its given in the constructor

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+            });
+    }
+
+    /**
+     * get longest site 1 fusion
+     * @param {StructuralVariantExt[]} previousFusions
+     * @returns {StructuralVariantExt}
+     */
+    getSite1LongestFusion(previousFusions: StructuralVariantExt[]): StructuralVariantExt | undefined {
+        const maxTotalWidth = Math.max.apply(Math, previousFusions
```
adamabeshouse: this should be unit tested

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+            .filter(fusion => !fusion.isLeftAligned)
+            .map(fusion => fusion.totalWidth));
+        return previousFusions.find(fusion => {
+            return fusion.totalWidth === maxTotalWidth && !fusion.isLeftAligned;
+        })
+    }
+
+    updateDeltaX(prevFusions: StructuralVariantExt[], t: EnsemblTranscriptExt) {
+        return prevFusions.map(prefFusion => {
```
adamabeshouse: this should  be unit tested

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+                prefFusion.deltaX = t.deltaX;
+            }
+            return prefFusion
+        });
+    }
+
+    hasLongestSite1(fusion: StructuralVariantExt,
+                    t: EnsemblTranscriptExt,
+                    prevLongest: StructuralVariantExt | undefined): boolean {
+        let longerThanPreviousLongest = true;
```
adamabeshouse: this should be unit tested

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+                pfam.fillColor = 'orange';
+                // calculate width of pfam
+                pfam.width = pfam.pfamDomainEnd - pfam.pfamDomainStart;
+                return pfam;
+            });
+    }
+
+    @computed
+    get referenceTranscripts(): EnsemblTranscriptExt[] {
+        return (this.computedTranscripts || [])
```
adamabeshouse: you're changing computedTranscripts in place

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+    getTotalWidth(exons: ExonRangeExt[]): number {
+        return exons.reduce((accumulator, e) => {
+            e.width = e.width ? e.width : 0;
+            return accumulator + e.width;
+        }, 0);
+    }
+
+    @computed
+    get computedTranscripts(): EnsemblTranscriptExt[] {
+        return (this.transcripts || []).map(t => {
```
adamabeshouse: you're changing transcripts in place

src/shared/components/exonsCharts/ExonsChartStore.ts
adamabeshouse: you're changing tableData in place.

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,252 @@
...
+                return fusion;
+            });
+    }
+
+    /**
+     * Associate fusions with their reference gene transcipts
+     * @returns {EnsemblTranscriptExt[]}
+     */
+    @computed
+    get fusionsByReferences(): EnsemblTranscriptExt[] {
```
adamabeshouse: this name doesn't make any sense. this also changes referenceTranscripts in place

src/shared/components/fusionCharts/FusionPieChart.tsx
```diff
@@ -0,0 +1,119 @@
...
+}
+
+@observer
+export default class FusionPieChart extends React.Component<IFusionPieChartProps, {}> {
+
+    geneFilters: string[];
+
+    constructor(props: IFusionPieChartProps) {
+        super(props);
+        this.geneFilters = []; // initialise
```
adamabeshouse: you can just initialize it on line 45 and delete the constructor

src/shared/components/fusionCharts/FusionPieChart.tsx
```diff
@@ -0,0 +1,119 @@
...
+                }, false);
+            });
+        } else {
+            // empty filter when there're no selected genes
+            this.props.dataStore.setFilter((d => true));
+        }
+    }
+
+    @action
+    onClickHandler() {
```
adamabeshouse: this is not an @action, @actions are imperative functions that make a change to an @observable property

src/shared/components/fusionCharts/FusionPieChart.tsx
```diff
@@ -0,0 +1,119 @@
...
+                colorScale={FUS_GENE_COLORSCALE}
+                style={{
+                    parent: {width: '15%', stroke: 'red'}, labels: {
+                        fontSize: 25,
+                        fontFamily: '"Helvetica Neue", Helvetica, Arial, sans-serif;',
+                        fill: '#FFF'
+                    }
+                }}
+                data={this.getFusionCounts()}
+                labels={(d: any) => `${d.y/this.props.dataStore.tableData.length > 0.1 ? d.y : ''}`}
```
adamabeshouse: can you put parens around d.y/this....tableData.length, would make it easier to read

src/shared/components/fusionCharts/FusionPieChart.tsx
```diff
@@ -0,0 +1,119 @@
...
+                events={[{
+                    target: 'data',
+                    eventHandlers: {
+                        onClick: () => this.onClickHandler()
```
adamabeshouse: annotate onClickHandler iwith @autobind, then just pass it in here `onClick: this.onClickHandler`

src/shared/components/fusionCharts/FusionPieChartLegend.tsx
```diff
@@ -0,0 +1,56 @@
...
+
+export interface IFusionPieChartLegendProps {
+    fusionCounts: {[fusedGene:string]: number}
+}
+
+@observer
+export default class FusionPieChartLegend extends React.Component<IFusionPieChartLegendProps, {}> {
+
+    constructor(props: IFusionPieChartLegendProps) {
+        super(props);
```
adamabeshouse: unnecessary constructor

src/shared/components/fusionCharts/FusionPieChartLegend.tsx
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    public render() {
+        let fusionGenes = Object.keys(this.props.fusionCounts);
+        let data =  fusionGenes.map(d => {
+            return {name: d};
+        });
```
adamabeshouse: this is a good candidate for @computed

```
@computed get data() {
    return Object.keys(this.props.fusionCounts).map(name=>({name}));
}
```

src/shared/components/fusionCharts/FusionPieChartLegend.tsx
```diff
@@ -0,0 +1,56 @@
...
+        // legend height depends on how many genes are fused to the reference gene
+        let legendHeight = fusionGenes.length > 9 ? (fusionGenes.length/9 * 400) : 400;
```
adamabeshouse: variables which are never reassigned should be declared as `const`

src/shared/components/fusionTable/FusionTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+                if (indexLocation > -1) {
+                    linkToPatientView = window.location.href.substring(0, indexLocation) + 'case.do' + linkToPatientView;
+                }
+                // END HACK
+
+                return <a href={linkToPatientView} target='_blank'>{sampleId}</a>
+            }
+        }
+
+        if (label.columnType === FusionTableColumnType.STUDY && this.props.studyIdToStudy) {
```
adamabeshouse: `else if` would make the logic here clearer

src/shared/components/fusionTable/FusionTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+            FusionTableColumnType.SITE2_HUGO_SYMBOL,
+            FusionTableColumnType.TUMOR_READ_COUNT,
+            FusionTableColumnType.TUMOR_VARIANT_COUNT,
+            FusionTableColumnType.ANNOTATION
+        ];
+
+        this._columns = {};
+
+        if (this.props.studyIdToStudy) {
+            let studyObject = Object.assign({}, this.props.studyIdToStudy);
```
adamabeshouse: theres no purpose for `studyObject`, just use Object.keys(this.props.studyIdToStudy)

src/shared/components/fusionTable/FusionTable.tsx
```diff
@@ -0,0 +1,335 @@
...
+    COMMENTS,
+    EXTERNAL_ANNOTATION,
+
+    DRIVER_FILTER,
+    DRIVER_FILTER_ANNOTATION,
+    DRIVER_TIERS_FILTER,
+    DRIVER_TIERS_FILTER_ANNOTATION,
+}
+
+export type FusionTableColumnLabel = {
```
adamabeshouse: this isn't a column label, its a column specification. naming matters.

src/pages/resultsView/ResultsViewPageStore.ts
oplantalech: I think we added it to have an explicit definition of every function, I don't think that the comments were auto-generated. I will remove them because indeed with the function names and the types of the parameters you can easily understand what is going on.

src/shared/components/exonsCharts/ExonsChartStore.spec.ts
```diff
@@ -0,0 +1,1451 @@
...
+    });
+
+    describe("getTotalWidth", () => {
+        it("should return total width of exons", () => {
+            const total = ExonsChartStore.getTotalWidth(allTranscriptsRaw[0].exons);
+            assert.equal(total, 1400);
+        });
+    });
+
+    describe("getLongestSite1Fusion", () => {
```
adamabeshouse: `getSite1LongestFusion`

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,242 @@
...
+                    )
+                    .map((e: ExonRangeExt) => {
+                        // add color
+                        e.fillColor = e.fillColor ? e.fillColor : t.fillColor;
+                        // calculate width of exon
+                        e.width = e.exonEnd - e.exonStart;
+                        return e;
+                    });
+                t.totalWidth = ExonsChartStore.getTotalWidth(<ExonRangeExt[]> t.exons);
+                t.deltaX = 0; // initialise delta for the x axis
```
adamabeshouse: you're modifying the actual object transcripts.result

what you're probably trying to do is:

`return Object.assign({}, t, { totalWidth: ..., deltaX: ... });`

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,242 @@
...
+            // retrieve label and desc from pfamDomains
+            const pfamDomain = (this.pfamDomains.result || []).find(
+                domain => domain.pfamAccession === pfam.pfamDomainId);
+            // add name and descriptions
+            pfam.name = pfamDomain ? pfamDomain.name : '';
+            pfam.description = pfamDomain ? pfamDomain.description : '';
+            // add color
+            pfam.fillColor = 'orange';
+            // calculate width of pfam
+            pfam.width = pfam.pfamDomainEnd - pfam.pfamDomainStart;
```
adamabeshouse: again, you are modifying these objects in-place, which is not a good way of doing things. probably it's not what you're intending either. please see my other comments for instructions on how to improve this

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,242 @@
...
+                const prevSite1Longest = this.getSite1LongestFusion(prevFusions);
+                const isCurrentFusionHasLongestSite1 = ExonsChartStore.hasLongestSite1(fusion, t, prevSite1Longest);
+
+                prevSite1LongestVal = 0; // re-initialize prev site1 longest val
+
+                // initialize total width
+                fusion.totalWidth = fusion.totalWidth ? fusion.totalWidth : 0;
+
+                // initialize delta-x for both ref gene and fusion
+                t.deltaX = isCurrentFusionHasLongestSite1 ? fusion.totalWidth - referenceTotalWidth : t.deltaX;
```
adamabeshouse: again you are modifying things in-place

src/shared/components/fusionTable/FusionTable.tsx
```diff
@@ -0,0 +1,338 @@
...
+    COMMENTS,
+    EXTERNAL_ANNOTATION,
+
+    DRIVER_FILTER,
+    DRIVER_FILTER_ANNOTATION,
+    DRIVER_TIERS_FILTER,
+    DRIVER_TIERS_FILTER_ANNOTATION,
+}
+
+export type FusionTableColumnLabelProp = {
```
adamabeshouse: `FusionTableColumnProps` works better

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,81 @@
...
+export default class ResultViewFusionMapper extends React.Component<IFusionMapperProps, {}> {
+
+    private renderComponents() {
+        const {
+            dataStore, fusionOccurrence, exonsChartStore, fusionMolecularProfile,
+            studyIdToStudy, molecularProfileIdToMolecularProfile
+        } = this.props.store;
+
+        const studyMap = new Map(Object.entries(studyIdToStudy));
+        const molecularProfileMap = new Map(Object.entries(molecularProfileIdToMolecularProfile));
```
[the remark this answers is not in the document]
rnugraha: Ah ya, should be .result

src/shared/components/exonsCharts/ExonsChart.tsx
```diff
@@ -0,0 +1,71 @@
...
+const containerStyle =
+{
+    width: '70vw',
+    overflow: 'auto',
+    marginRight: '5px'
+};
+
+@observer
+export default class ExonsChart extends React.Component<IExonsChartProps, {}> {
+    renderFusionExons(fusions: StructuralVariantExt[] | undefined, referenceGeneWidth: number) {
```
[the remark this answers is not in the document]
rnugraha: I don't really think t needs the `@compute get ` here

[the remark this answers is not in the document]
adamabeshouse: oops sorry, I misread this

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,137 @@
...
+    constructor(public gene: Gene,
+                public studyIdToStudy: MobxPromise<{ [studyId: string]: CancerStudy }>,
+                public molecularProfileIdToMolecularProfile: MobxPromise<{ [molecularProfileId: string]: MolecularProfile }>,
+                public samples: MobxPromise<SampleIdentifier[]>,
+                public fusions: StructuralVariantExt[]) {
+        labelMobxPromises(this);
+    }
+
+    private applyColor (result: EnsemblTranscriptExt[]) {
+        let fusionTranscripts: EnsemblTranscriptExt[] = [];
```
adamabeshouse: this could really be a util since it never uses `this`, should probably be unit tested as well. I still think `applyColor` is not a good name, `apply` is not the right verb. `addColor` works better.

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,137 @@
...
+                .map(t => t.pfamDomains)
+                .reduce((accumulator, currentValue) => accumulator.concat(currentValue), [])
+                .map(t => t.pfamDomainId);
+            // fetch the data with pfam domain ids as request params
+            return fetchPfamDomainData(pFamDomainIds);
+        }
+    }, []);
+
+    @computed
+    get fusionOccurrence(): { [fusionLabel: string]: number } {
```
adamabeshouse: I still think `fusionCounts` is a better name. Occurence is vague to me, unless thats a biological term I don't know. what do you think?

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,260 @@
...
+
+export class ExonsChartStore {
+
+    constructor(public gene: Gene,
+                public transcripts: MobxPromise<EnsemblTranscriptExt[]>,
+                public pfamDomains: MobxPromise<PfamDomain[]>,
+                public fusions: StructuralVariantExt[]
+    ) {}
+
+    readonly transcriptById = remoteData({
```
adamabeshouse: small but `transcripts`, with an s to make it plural

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,260 @@
...
+    static getTotalWidth(exons: ExonRangeExt[]): number {
+        return exons.reduce((accumulator, e) => {
+            e.width = e.width ? e.width : 0;
+            return accumulator + e.width;
+        }, 0);
+    }
+
+    assignExonsToFusions (fusion: StructuralVariantExt): StructuralVariantExt {
+        let site1Exons = this.getExonsBySite(1, fusion.site1EnsemblTranscriptId, fusion.site1Exon);
+        let site2Exons = this.getExonsBySite(2, fusion.site2EnsemblTranscriptId, fusion.site2Exon);
```
adamabeshouse: nitpick: variables that are never reassigned (`site1Exons`, `site2Exons`) should be `const`

src/shared/components/exonsCharts/ExonsChartStore.ts
```diff
@@ -0,0 +1,260 @@
...
+        const exons = site1Exons.concat(site2Exons);
+        const totalWidth = ExonsChartStore.getTotalWidth(site1Exons.concat(site2Exons) || []);
+        return Object.assign({}, fusion, {
+            isLeftAligned: (fusion.site1HugoSymbol === this.gene.hugoGeneSymbol),
+            exons: exons,
+            totalWidth: totalWidth
```
adamabeshouse: small thing: when the variable name is the same as the object name, you dont' have to repeat it.
i.e. this can be written
```
{
  isLeftAligned: ..,
  exons, 
  totalWidth
}
```

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
rnugraha: sounds good

rnugraha:
> thank you so much for your hard work making the changes!

No problem, tx for reviewing this giant PR. Anyway pls don't merge, unit testings aren't updated yet. Will let you guys know once updated.

rnugraha:
Hey guys, could you be kind to have a look again? Code has been rebased and broken unit testing has been fixed.

inodb:
@rnugraha 	I'm thinking of merging this directly to master as part of `3.1.1` after `release-3.1.0` branch has been merged, since it's a frontend only change. Once 3.1.0 is deployed we can import some data to production and test the PR with real data.

Wondering if it would be a good idea to add localdb e2e tests to this PR?

rnugraha:
@inodb Hi Ino, yes that would be great. I agree it is good idea to add localdb e2e to this PR.

inodb:
I think this is superseded by one of Kelsey's PRs

inodb:
closing this as superseded by #3023

PR #1366, opened 2018-08-07
new oncoprint setting - only show clinical track legends for altered cases

adamabeshouse:


src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -768,6 +775,16 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
                         onClick={this.onInputClick}
                     /> Show legends for clinical tracks
                 </label></div>
+                <div className="checkbox" style={{marginLeft:20}}><label>
+                    <input
+                        data-test="onlyShowClinicalLegendsForAltered"
+                        type="checkbox"
+                        value={EVENT_KEY.onlyShowClinicalLegendForAlteredCases}
+                        checked={this.props.state.onlyShowClinicalLegendForAlteredCases}
+                        onClick={this.onInputClick}
```
alisman: Would onChange be slightly more semantic?

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -662,6 +662,7 @@ function transitionClinicalTrack(
             },
             sort_direction_changeable: true,
             tooltipFn: makeClinicalTrackTooltip(nextSpec, true),
+            important_ids: nextSpec.altered_uids,
```
alisman: maybe a comment why they are important?  Could it be for reason other than they are altered?

src/shared/components/oncoprint/controls/OncoprintControls.tsx
adamabeshouse: I've always been happy with onClick, I think I've had issues with onChange with controlled components

src/shared/components/oncoprint/DeltaUtils.ts
adamabeshouse: in the oncoprint library I have a comment explaining this - important means they determine which rules show up in the legend. it could be any list of ids at all

adamabeshouse: ill add a comment explaining that

PR #1369, opened 2018-08-09
Move vertical box plot legends to the top

adamabeshouse:


alisman:
@does this take legend width into account?  or does the wrapping somehow prevent legend from ever overflowing?

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -182,16 +183,30 @@ export default class BoxScatterPlot<D extends IBaseBoxScatterPlotPoint> extends
...
     private get legend() {
         const x = this.legendX;
         if (this.props.legendData && this.props.legendData.length) {
+            // all the props conditional on this.props.horizontal are because
+            //  we want to put the legend on top of the plot, since the plot
+            //  could be arbitrarily wide with a lot of boxes
             return (
                 <VictoryLegend
-                    orientation="vertical"
+                    orientation={this.props.horizontal ? "vertical" : "horizontal"}
```
alisman: so there is still a case where legend will be vertical?

adamabeshouse:
@alisman heuristically the wrapping will prevent the legend from overflowing

src/shared/components/plots/BoxScatterPlot.tsx
adamabeshouse: yes, if the box plot is oriented so that more boxes go down instead of to the right. what do you think?

PR #1372, opened 2018-08-09
Filter mutationMapperTool by transcriptid

inodb:
Only show mutations for the canonical transcriptId being displayed. Related to:

https://github.com/genome-nexus/genome-nexus/issues/205

Try with:

```
Sample_ID	Cancer_Type	Chromosome	Start_Position	End_Position	Reference_Allele	Variant_Allele
TCGA-49-4494-01	Lung Adenocarcinoma	6	27819890	27819890	A	G
```

Now doesn't show any mutations.

Discuss: might be better to not show any genes at all in this case, but might be ok for now.

Also this should at some point be applied to mutations tab, but we should figure out what the best way to do it is.

### DIFF
--- a/src/shared/lib/MutationAnnotator.spec.ts
+++ b/src/shared/lib/MutationAnnotator.spec.ts
@@ -2,9 +2,10 @@ import {assert} from "chai";
 import sinon from 'sinon';
 import * as _ from 'lodash';
 
-import {annotateMutations, resolveMissingProteinPositions} from "./MutationAnnotator";
+import {annotateMutations, resolveMissingProteinPositions, fetchVariantAnnotationsIndexedByGenomicLocation} from "./MutationAnnotator";
 import {Mutation} from "shared/api/generated/CBioPortalAPI";
 import {initMutation} from "test/MutationMockUtils";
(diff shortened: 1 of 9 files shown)


src/shared/lib/MutationAnnotator.spec.ts
```diff
@@ -352,7 +352,7 @@ describe("MutationAnnotator", () => {
         ];
     });
 
-    describe('annotateMutations', () => {
+    /*describe('annotateMutations', () => {
```
inodb: this should be updated to reflect new methods

src/shared/lib/MutationAnnotator.ts
```diff
@@ -63,12 +63,46 @@ export async function annotateMutations(mutations: Mutation[],
         }
     ): [];
 
-    const indexedVariantAnnotations = indexAnnotationsByGenomicLocation(variantAnnotations);
+    return indexAnnotationsByGenomicLocation(variantAnnotations);
```
inodb: needed this result in the mutation mapper store generator, such that the mutations could be filtered by transcript. I think it's good to keep both the mutations and the annotations available such that we can add the dropdown box for transcript selection at some point

PR #1377, opened 2018-08-13
Cna and putative driver coloring in expression tab

adamabeshouse:


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -260,13 +260,14 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
 
                             const store = this.resultsViewPageStore;
 
-                            if (store.rnaSeqMolecularData.isComplete && store.studyIdToStudy.isComplete
-                                && store.mutations.isComplete && store.genes.isComplete && store.coverageInformation.isComplete) {
+                            if (store.studyIdToStudy.isComplete
+                                && store.putativeDriverAnnotatedMutations.isComplete && store.genes.isComplete && store.coverageInformation.isComplete) {
```
alisman: putative driver contains rnaseq and mutations data?

adamabeshouse: what?

adamabeshouse: oh, no I just rewrote it to use numericGeneMolecularDataCache as a data source

adamabeshouse:
- [ ] reverse tick label angle
- [ ] cbioTooltip class needs to be applied on Popover
- [ ] reduce right padding when no legend, but still leave room for angled tick labels
- [ ] move '* Driver annotations ... ' message outside borderedChart

PR #1378, opened 2018-08-13
fix error popup on patient page when service fails

alisman:
See attached issue.  This PR prevents queryStore from unnecessarily instantiating on patient page and subsequently calling virtual study service.  Also disables error popup since it isn't wired properly to errors

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/appBootstrapper.jsx
+++ b/src/appBootstrapper.jsx
@@ -103,7 +103,7 @@ switch (window.defaultRoute) {
 const syncedHistory = syncHistoryWithStore(history, routingStore);
 
 // lets make query Store since it's used in a lot of places
-const queryStore = new QueryStore(window, window.location.href);
+const queryStore = (/\/patient$/.test(window.location.pathname)) ? {} : new QueryStore(window, window.location.href);
 
 const stores = {
     // Key can be whatever you want
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -292,12 +292,12 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         return (
             <div className="patientViewPage">
 
-                <AjaxErrorModal
-                    show={(patientViewPageStore.ajaxErrors.length > 0)}
-                    onHide={()=>{ patientViewPageStore.clearErrors() }}
-                    title={`Can't find ${patientViewPageStore.pageMode} ${patientViewPageStore.caseId} in study ${patientViewPageStore.studyId}.`}
-                    troubleshooting={["Check that your URL parameters are valid.", "Try refreshing the page.", "Make sure you are connected to the internet."]}
-                />
+                {/*<AjaxErrorModal*/}
+                    {/*show={(patientViewPageStore.ajaxErrors.length > 0)}*/}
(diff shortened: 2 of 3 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -492,7 +493,9 @@ export class QueryStore
 		default: {},
 	});
 
-	readonly virtualStudies = remoteData(sessionServiceClient.getUserVirtualStudies(), []);
+	readonly virtualStudies = remoteData(async ()=>{
+		return sessionServiceClient.getUserVirtualStudies();
```
pieterlukasse: I think this will still trigger an exception in the backend and http 500 error in the frontend if no session service is configured. As a side note, I also think that sessionService is an implementation detail that perhaps does not belong here. In this layer, all we want is the list of user virtual studies, perhaps from a VirtualStudiesService. Where the data is stored should be irrelevant here. VirtualStudiesService could then have the appropriate logic to return an empty list, e.g. if not session service is found, and do the exception handling/throwing if necessary.

alisman: Hey Peter, you're right that we should't be calling the service at all if it's not configured. I will follow up with that.  Less sure on your second point.  The store is where we load data and handle the asynchronous behavior of doing so.   If there was a way of storing user virtual studies without session service, then we could further abstract, but it's the instance of remoteData that allows us to handle async, so this would more or less look like it does.   What I think is elegant about the remoteData (wrapper for mobx promise) paradigm is that it allows us to store both the data and the loading state of the data (and the potential error state) in one "node".

src/appBootstrapper.jsx
```diff
@@ -103,7 +103,7 @@ switch (window.defaultRoute) {
 const syncedHistory = syncHistoryWithStore(history, routingStore);
 
 // lets make query Store since it's used in a lot of places
-const queryStore = new QueryStore(window, window.location.href);
+const queryStore = (window.location.pathname === "/patient") ? {} : new QueryStore(window, window.location.href);
```
alisman: this isn't strictly necessary anymore, but it will become irrelevant in SPA

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -292,12 +292,12 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         return (
             <div className="patientViewPage">
 
-                <AjaxErrorModal
-                    show={(patientViewPageStore.ajaxErrors.length > 0)}
-                    onHide={()=>{ patientViewPageStore.clearErrors() }}
-                    title={`Can't find ${patientViewPageStore.pageMode} ${patientViewPageStore.caseId} in study ${patientViewPageStore.studyId}.`}
-                    troubleshooting={["Check that your URL parameters are valid.", "Try refreshing the page.", "Make sure you are connected to the internet."]}
-                />
+                {/*<AjaxErrorModal*/}
```
pieterlukasse: maybe remove this code instead of commenting out?

pieterlukasse:
any news on this one?

PR #1379, opened 2018-08-13
Study view : show user selections

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/4153

### DIFF
--- a/src/pages/studyView/UserSelections.tsx
+++ b/src/pages/studyView/UserSelections.tsx
@@ -1,0 +1,160 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import { observer } from "mobx-react";
+import { computed } from 'mobx';
+import styles from "./styles.module.scss";
+import { StudyViewFilter } from 'shared/api/generated/CBioPortalAPIInternal';
+import { ChartMeta, UniqueKey } from 'pages/studyView/StudyViewPageStore';
+import { isFiltered } from 'pages/studyView/StudyViewUtils';
+import autobind from 'autobind-decorator';
+
+export interface IUserSelectionsProps {
+    filter: StudyViewFilter;
+    attributesMetaSet: { [id: string]: ChartMeta };
+    updateClinicalDataEqualityFilter: (chartMeta: ChartMeta, value: string[]) => void;
+    clearGeneFilter: () => void;
+    clearCNAGeneFilter: () => void;
+    clearCustomCasesFilter: () => void;
+    clearAllFilters: () => void
+}
+
+
+@observer
+export default class UserSelections extends React.Component<IUserSelectionsProps, {}> {
+
+    constructor(props: IUserSelectionsProps) {
+        super(props);
+    }
+
+    @computed get showFilters() {
+        return isFiltered(this.props.filter)
+    }
+
+    @computed private get clinicalDataEqualityFilters() {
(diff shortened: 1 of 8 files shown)


src/pages/studyView/UserSelections.tsx
```diff
@@ -0,0 +1,98 @@
...
+
+@observer
+export default class UserSelections extends React.Component<IUserSelectionsProps, {}> {
+
+    constructor(props: IUserSelectionsProps) {
+        super(props);
+    }
+
+    @computed get showFilters() {
+        return !_.isEmpty(this.props.filter) && (
```
zhx828: Wondering whether this should be a util function? This could be used in some other places too? Can't think about a use case though.

src/pages/studyView/styles.module.scss
```diff
@@ -18,4 +18,29 @@
 
 .studyViewSurvivalPlotBody {
     position: relative
+}
+
+.selectedFilters {
```
zhx828: Let's always fixed a `studyView` string before all classes

src/pages/studyView/UserSelections.tsx
onursumer: @kalletlak I also noticed a similar if condition in the cache, but actually it is not logically equivalent to this one. Is this is intentional or is it a bug? 

https://github.com/cBioPortal/cbioportal-frontend/blob/090546ff443214e716ab484273beca20dd625d4f/src/shared/cache/StudyViewClinicalDataCountsCache.ts#L45

At some point, we will also need to add `clinicalDataIntervalFilters` (which will be introduced by bar chart API) both here and in the cache.

kalletlak: It looks like same. I'll move it utils

src/pages/studyView/UserSelections.tsx
```diff
@@ -0,0 +1,94 @@
...
+    }
+
+    render() {
+        if (this.showFilters) {
+            return (
+                <div className={styles.studyViewUserSelections}>
+                    <span>Your selection:</span>
+                    <div>
+                        {
+                            _.map((this.props.filter.clinicalDataEqualityFilters || []), clinicalDataEqualityFilter => {
```
alisman: this code is too deeply nested.  i would remove it to computed getter or even some nested components.  look at stateless components for example

src/pages/studyView/UserSelections.tsx
```diff
@@ -0,0 +1,94 @@
...
+                                let chartMeta = this.props.attributesMetaSet[clinicalDataEqualityFilter.clinicalDataType + '_' + clinicalDataEqualityFilter.attributeId];
+                                return (
+                                    <span className={styles.filter}>
+                                        <span className={styles.name}>{chartMeta.displayName}</span>
+                                        {_.map(clinicalDataEqualityFilter.values, value => {
+                                            return (
+                                                <span className={styles.value}>
+                                                    <span className={styles.label}>{value}</span>
+                                                    <i className="fa fa-times" style={{ cursor: "pointer" }} onClick={event => {
+                                                        this.props.updateClinicalDataEqualityFilter(chartMeta, _.filter(clinicalDataEqualityFilter.values, _value => _value !== value))
```
alisman: you should make this onClick a member of the component (autobind it).   Try to avoid using anonymous callbacks to handle events in jsx.  It is not performance because diffing algorithm sees it as a new function every render cycle and does unnecessary DOM manipulation.

PR #1380, opened 2018-08-14
plots tab - filter out nan data

adamabeshouse:


pieterlukasse:
maybe cover this with a regression test?

PR #1398, opened 2018-08-20
Functional Impact Column, query all mutations

inodb:
Was only querying missense. This uses genomic location endpoint instead to
query all of them


src/shared/cache/GenomeNexusEnrichment.ts
```diff
@@ -37,7 +36,14 @@ export function fetch(queries: Mutation[]):Promise<VariantAnnotationEnriched[]>
 
 export default class GenomeNexusCache extends LazyMobXCache<VariantAnnotationEnriched, Mutation> {
     constructor() {
-        super((m:Mutation) => (generateGenomeNexusQuery(m)), // queryToKey
+        super((m:Mutation) => {
+                  const genomicLocation = extractGenomicLocation(m);
+                  if (genomicLocation) {
+                    return genomicLocationString(genomicLocation); // queryToKey
+                  } else {
+                      return "";
```
inodb: @onursumer can u check if this is correct, not sure what key value should be returned for the cache if the input format is incorrect.

onursumer: @inodb Actually, not sure what to do in case of a mutation with invalid genomic location. Maybe we should filter out mutations with invalid genomic locations before invoking any cache method?

PR #1399, opened 2018-08-20
Oncoprint - only determine ProfiledIn tracks based on selected profiles

adamabeshouse:
Fixes bug where it was recorded that a sample was not profiled for CNA because it was not profiled for a profile that was not in the query.


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -759,22 +759,27 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         invoke:()=>{
             const groupedSelectedMolecularProfiles:{[alterationType:string]:MolecularProfile[]} =
                 _.groupBy(this.props.store.selectedMolecularProfiles.result!, "molecularAlterationType");
+            const selectedMolecularProfilesMap = _.keyBy(this.props.store.selectedMolecularProfiles.result!, p=>p.molecularProfileId);
```
alisman: is groupedSelectedMolecularProfiles used anymore below?

alisman:
I think this could use a test and a comment explaining what's going on.  The names are pretty descriptive but it's still hard to grok.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -759,22 +759,27 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
 
             const existsUnprofiled:{[alterationType:string]:boolean} = {};
             const coverageInfo = this.props.store.coverageInformation.result!.samples;
             const molecularProfileIdToMolecularProfile = this.props.store.molecularProfileIdToMolecularProfile.result!;
             for (const uniqueSampleKey of Object.keys(coverageInfo)) {
                 for (const gpData of coverageInfo[uniqueSampleKey].notProfiledAllGenes) {
-                    existsUnprofiled[
-                        molecularProfileIdToMolecularProfile[gpData.molecularProfileId].molecularAlterationType
-                    ] = true;
+                    if (gpData.molecularProfileId in selectedMolecularProfilesMap) {
```
alisman: for the loops here, i'm in favor of using reduce because it would outline the purpose of this code. i.e. to produce the output: existsUnprofiled.

src/shared/components/oncoprint/ResultsViewOncoprintUtils.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    // If its a single study query (isSingleStudyQuery), and there is only one profile of a particular alteration type
+    //  in the study, then an attribute will be generated specifically for that profile. Otherwise, the attribute
+    //  will represent an entire alteration type.
+
+    const groupedSelectedMolecularProfiles:{[alterationType:string]:MolecularProfile[]} =
+        _.groupBy(selectedMolecularProfiles, "molecularAlterationType");
+    const selectedMolecularProfilesMap = _.keyBy(selectedMolecularProfiles, p=>p.molecularProfileId);
+
+    const existsUnprofiled:{[alterationType:string]:boolean} = _.reduce(coverageInformation, (map, sampleCoverage)=>{
```
alisman: i believe if you type the argument "map" in the reducer callback, you can avoid having to to type both the variable and then cast the initial memo (third arg of reduce)

PR #1400, opened 2018-08-20
New 'profiled in' tracks better-compatible with multiple study queries

adamabeshouse:


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -1089,6 +1116,14 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         return "";
     }
 
+    @computed get alterationTypesInQuery() {
```
alisman: I know we've talked about this before but I really prefer @memoized and then you pass in the dereferenced result.  That we you deal with the completion status outside

adamabeshouse: you mean make a function wrapping line 1121?

adamabeshouse:
https://github.com/cBioPortal/cbioportal-frontend/pull/1418

PR #1401, opened 2018-08-21
Fix mutation mapper tool example protein change

inodb:
This was broken in the PR that filtered mutation mapper tool by transcript id.
When gene/protein change is given, we don't know the transcript id. Therefore
check whether the annotations based on genomic positions are actually being
used.

### DIFF
--- a/src/pages/tools/mutationMapper/MutationMapperToolStore.ts
+++ b/src/pages/tools/mutationMapper/MutationMapperToolStore.ts
@@ -171,7 +171,7 @@ export default class MutationMapperToolStore
                         this.canonicalTranscriptsByHugoSymbol.result[gene.hugoGeneSymbol].transcriptId : undefined;
                     
                     const getMutations = () => {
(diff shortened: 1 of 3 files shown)


src/pages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -171,7 +171,7 @@ export default class MutationMapperToolStore
                         this.canonicalTranscriptsByHugoSymbol.result[gene.hugoGeneSymbol].transcriptId : undefined;
                     
                     const getMutations = () => {
-                        if (canonicalTranscriptId && this.indexedVariantAnnotations.result) {
+                        if (canonicalTranscriptId && this.indexedVariantAnnotations.result && !_.isEmpty(this.indexedVariantAnnotations.result)) {
```
inodb: it might be nice to use some flag that the input annotations are protein changes - but this works too i guess

inodb:
It's not entirely clear how the combined input should work, filed an issue: https://github.com/cBioPortal/cbioportal/issues/4687

PR #1403, opened 2018-08-22
SV: Download clinical data

kalletlak:
Implement download clinical data feature in https://github.com/cBioPortal/cbioportal/issues/4154

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -109,9 +109,7 @@ export class StudyViewPageStore {
 
     @observable private queriedGeneSet = observable.map<boolean>();
 
-    private _fetchedClinicalDataSamples: string[] = [];
-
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -491,12 +493,23 @@ export class StudyViewPageStore {
...
+        await: () => [this.clinicalAttributes],
         invoke: () => {
             return internalClient.fetchFilteredSamplesUsingPOST({
                 studyViewFilter: this.emptyFilter
             })
         },
-        default: []
+        default: [],
+        onResult: (result) => {
+            this._sampleClinicalDataSet = _.reduce(result, (acc, next) => {
```
alisman: seems that this is a computation of samples, so why not just make it a remoteData that awaits samples and then spits back the transformed value?  or even just a helper that consuming code can use to transform the data?   maybe there's a reason here, but we want to avoid using onResult to set data except when absolutely necessary.  it's an antipattern.

PR #1405, opened 2018-08-23
SV: query button

kalletlak:
# What? Why?
Fix query button in https://github.com/cBioPortal/cbioportal/issues/4154


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -744,4 +748,63 @@ export class StudyViewPageStore {
             return Promise.resolve(data);
         }
     });
+
```
alisman: fyi keep it in mind that we will need to adapt this to new SPA submission.  no change necessary now

PR #1410, opened 2018-08-24
Update chart loading

kalletlak:


src/pages/studyView/charts/StudyViewComponentLoader.tsx
```diff
@@ -1,30 +1,27 @@
...
-                <If condition={this.props.promise.isPending}>
-                    <LoadingIndicator
-                        isLoading={!!this.props.promise.isPending}
-                    />
-                </If>
-                <If condition={this.props.promise.isError}>
-                    <div>Error when loading data.</div>
-                </If>
-                <If condition={this.props.promise.isComplete}>
+            <div style={{ position: "relative", height: "100%"}}>
```
zhx828: Why don't we use LoadingIndicator? And this style is lost `styles.studyViewAutoMargin`

kalletlak: I wasn't able to apply styles to that loader

alisman: LoadingIndicator should have a style property which will apply styles directly to loader.  Please do use it.  Check definition.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -258,7 +267,7 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
                     chartControls={this.chartControls}
                     changeChartType={this.changeChartType}
                 />
-                <StudyViewComponentLoader promise={this.props.promise}>
+                <StudyViewComponentLoader promise={this.props.promise} isInitialLoaded={!_.isUndefined(this.chartData)}>
```
alisman: You can now just use === undefined.     The problem with testing against undefined is old and has been solved in modern browsers.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -154,6 +156,13 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
         this.chartType = chartType;
     }
 
+    @computed get chartData(){
```
alisman: how come we are caching data?

kalletlak:
Closing this as we no longer require this functionality

PR #1411, opened 2018-08-24
Add FixedHeaderTable from react-virtualized and use in study view summary tab

zhx828:
Features included:
- Fixed header
- Sort, filter
- Select Samples button (multi-selection without rerender)
- typescript upgraded to 2.8

This is the part of changes in https://github.com/cBioPortal/cbioportal/issues/4158

# Checks
  the `-s` option)

# Any screenshots or GIFs?
screen shot 2018-08-24 at 4 01 02 pm


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -143,18 +141,13 @@ export class StudyViewPageStore {
...
             //TODO: all elements instead of one
             mutatedGeneFilter = { entrezGeneIds: [] };
         }
-        let _index = mutatedGeneFilter.entrezGeneIds.indexOf(entrezGeneId);
-        if (_index === -1) {
-            mutatedGeneFilter.entrezGeneIds.push(entrezGeneId);
-        } else {
-            mutatedGeneFilter.entrezGeneIds.splice(_index, 1);
-        }
+        mutatedGeneFilter.entrezGeneIds = _.xor(mutatedGeneFilter.entrezGeneIds, entrezGeneIds);
```
alisman: you might want to consider making a new gene filter object each time instead of treating the old one as mutable.  is there anything else on the filter object besides the list of geneIds?  if not, then maybe just the gene collection instead of nested object.

alisman: this is just a general best practice because then you don't have to use deep observation on the collection.  there probably won't be a ton of genes here so it may be insignificant, but still

alisman: then you could make the array observable ref

zhx828:
@alisman updated based on the suggestion.

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -1,106 +1,219 @@
...
+            },
+            width: 100
         },
         {
             name: 'CNA',
             render: (data: CopyNumberCountByGene) =>
-                <span>{data.alteration === -2 ? 'DEL' : 'AMP'}</span>,
-            width: "20%"
+                <span
+                    className={classnames(data.alteration === -2 ? styles.del : styles.amp)}>{data.alteration === -2 ? 'DEL' : 'AMP'}</span>,
```
onursumer: Minor formatting suggestion:
```
<span className={classnames(data.alteration === -2 ? styles.del : styles.amp)}>
    {data.alteration === -2 ? 'DEL' : 'AMP'}
</span>
```

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+    @action
+    sort({sortBy}: any) {
+        this._store.defaultHeaderClick(this.getColumn(sortBy));
+        this._sortBy = sortBy;
+        this._sortDirection = this._store.dataStore.sortAscending ? 'asc' as 'asc' : 'desc' as 'desc';
+    }
+
+    @bind
+    @action
+    onFilterTextChange() {
```
onursumer: I think this code is very similar (if not identical) to the one we have in `LazyMobXTable`, it would be great to have a shared utility function for this.

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -0,0 +1,161 @@
...
+                    sortDirection={RVSDTtoStrType[this._sortDirection]}
+                    sortBy={this._sortBy}
+                >
+
+                    {
+                        this.props.columns.map(column => {
+                            return <RVColumn label={column.name} dataKey={column.name} width={Number(column.width)}
+                                             cellRenderer={function (props: TableCellProps) {
+                                                 return column.render(props.rowData);
+                                             }.bind(this)}/>;
```
onursumer: Is there a specific reason here to use a `function` keyword and manually bind instead of using an arrow function like this:
`(props: TableCellProps) => column.render(props.rowData)`

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -0,0 +1,153 @@
...
+            columns: this.props.columns,
+            data: this.props.data,
+            initialSortColumn: this._sortBy,
+            initialSortDirection: this._sortDirection
+        });
+    }
+
+    @bind
+    rowClassName({index}: any) {
+        if (_.includes(this.props.selectedRows, index)) {
```
kalletlak: changing sorting order highlights different records as the index of the selected record changes

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -85,9 +85,11 @@ class CNASegmentIframe extends React.Component<CNSegmentsIframeProps,{}>{
            height:630,
            border:'none'
         }).appendTo($(this.iframeDiv).empty());
-        iframe.contentWindow.document.open();
-        iframe.contentWindow.document.write(this.getIframeBody());
-        iframe.contentWindow.document.close();
+        if (iframe.contentWindow) {
```
alisman: how come?

zhx828: The 2.8 typescript complains about it may be undefined

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -1,106 +1,228 @@
...
+            return this.selectedRows.length > 0 && !_.isUndefined(_.find(this.selectedRows, (row: CNAGenesTableUserSelectionWithIndex) => row.entrezGeneId === entrezGeneId && row.alteration === alteration));
+        } else {
             return true;
-        });
-        return flag;
+        }
+    }
+
+    @bind
+    @action
```
alisman: i don't see any mutation (setting/changing of state) here so does this need to be @action?

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -35,89 +89,35 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
...
+    @action
+    private tooltipLabelMouseEnter(value: string): void {
         if (this.props.highlightedRow) {
             this.props.highlightedRow(value);
         }
     }
 
     @bind
-    @action private tooltipLabelMouseLeave(): void {
+    @action
```
alisman: action should probably be in the highlightedRow callback itself or whatever actually directly mutates the mobx observable

alisman:
Seems like some unecessary use of @action.

src/pages/resultsView/cnSegments/CNSegments.tsx
alisman: gotcha.

PR #1417, opened 2018-08-28
CircleCI: show package lock url before diff

inodb:
as per Adam's request - should show URL before the diff, so it's easier to download the new package lock file


inodb:
@adamabeshouse first one is to get the exit code of git diff (you check whether or not the file actually changed). Then run the part behind `||` if the exit code is non-zero, that's bash's OR. Normally that command outputs the diff, but i just send it to /dev/null. Could prolly also use `--silent` parameter or something else instead. The second one is to actually show the result, we could drop `--exit-code` parameter there actually, but whatevs

adamabeshouse:
@inodb oh so for circleCI the first output has to be the exit code, you cant just use the exit code at the end. that makes sense

PR #1423, opened 2018-08-29
SV: Add chart dropdown

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4154

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -21,7 +21,8 @@ import {
     Gene,
     MolecularProfile,
     MolecularProfileFilter,
-    Patient, PatientFilter
+    Patient, PatientFilter,
+    ClinicalAttributeFilter
 } from 'shared/api/generated/CBioPortalAPI';
 import {PatientSurvival} from 'shared/model/PatientSurvival';
 import {getPatientSurvivals} from 'pages/resultsView/SurvivalStoreHelper';
@@ -394,8 +395,13 @@ export class StudyViewPageStore {
         this._cnaGeneFilter = [];
     }
 
-    @action changeChartVisibility(uniqueKey:string, visible: boolean) {
-        this._chartVisibility.set(uniqueKey, visible);
+    @action
+    changeChartVisibility(uniqueKey:string, visible: boolean) {
+        if(visible){
+            this._chartVisibility.set(uniqueKey, true);
+        } else {
+            this._chartVisibility.delete(uniqueKey);
+        }
     }
 
     @action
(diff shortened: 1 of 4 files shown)


src/pages/studyView/styles.scss
```diff
@@ -18,6 +18,38 @@
     .btn:active:focus {
         box-shadow: none; 
     }
+
+    .Select {
```
zhx828: Doesn't the style come with the component already?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -306,26 +325,76 @@ export class StudyViewPageStore {
...
+            // TODO: this is a temorary solution util we figure out the chartVisibility logic
+            let osStatusFlag = false;
+            let osMonthsFlag = false;
+            let dfsStatusFlag = false;
+            let dfsMonthsFlag = false;
+            let mutationCountFlag = false;
+            let fractionGenomeAlteredFlag = false;
+
+            clinicalAttributes.forEach(obj => {
+                if (obj.clinicalAttributeId === 'OS_STATUS') {
```
zhx828: Let's make all these strings to const, similarly what we have for MUTATION_COUNT

src/pages/studyView/styles.scss
kalletlak: seems not

kalletlak:
@zhx828 
1. I dont see any console errors
2. Cannot reproduce it
3. I don't see any error message
4. I don't think we can add loading gif, we may need to change in the library

zhx828:
@jjgao @kalletlak should we add an option `Add the default charts` after clicking on the `Clear`?

zhx828:
@kalletlak Another issue: the percentage of Mutated Genes/CNA Genes table are way high
screen shot 2018-08-30 at 10 45 06 am

Reproduce:
- Open study: http://www.cbioportal.org/newstudy?id=acc_tcga_pan_can_atlas_2018#summary
- Clear all charts
- Add Ethnicity Category from the dropdown
- Select the 8 slice
- Take a look the drop down

package.json
```diff
@@ -139,7 +139,7 @@
     "mobx-react-devtools": "^4.2.11",
     "mobx-react-router": "^3.1.2",
     "mobx-utils": "^2.0.1",
-    "mobxpromise": "^1.3.1",
+    "mobxpromise": "github:cbioportal/mobxpromise#master",
```
zhx828: Why?

zhx828: Just saw the message in the channel

zhx828:
Created another issue https://github.com/thehyve/react-select-checked/issues/23

src/pages/studyView/styles.scss
zhx828: The style needs to be updated, especially the line hight. Please see the style in coExpression folder under resultsView as example.

PR #1427, opened 2018-08-30
Add Clinical Data Tab

zhx828:
This is a part of fixes for https://github.com/cBioPortal/cbioportal/issues/4734

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -716,6 +716,27 @@ export class StudyViewPageStore {
         }
     });
 
+    readonly getDataForClinicalDataTab = remoteData({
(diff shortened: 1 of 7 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -716,6 +716,27 @@ export class StudyViewPageStore {
         }
     });
 
+    readonly getDataForClinicalDataTab = remoteData({
+        await: () => [this.clinicalAttributes, this.selectedSamples],
```
kalletlak: are we just showing selected/filtered samples clinical data?

zhx828: Yup, JJ asked before to synchronize the tab with the current selection. So I added it in.

PR #1429, opened 2018-08-31
Add Mutation Mapper Tool Dropdown

inodb:
Adds transcript dropdown for mutation mapper tool

- [x] Add dropdown
- [x] Change lollipop plot when switching transcripts
- [x] Only list mutations in mutations table that actually target that mutation
- [x] Disable for now on mutations tab in results view

Genome Nexus
- [x] Add TranscriptConsequenceSummary for all transcript consequences produced by VEP (so we can get protein start / end)
- [x] make backwards compatible

CBioPortal Frontend
- [x] Keep track of all mutations (now number of transcripts in dropdown changes depending on currently selected transcript)
- [x] re-annotate mutations once genome nexus endpoint has been extended
- [x] show 3d structures for those transcripts that have one
- [x] add refseqid
- [x] add ccdsid

UX questions:
- [x] should we list mutations not appearing on the active transcript? Showing in the dropdown header how many mutations there are in total
- [x] should the drop down show # mutations and protein length for a given transcript?
- [x] show maybe a faded hotspot for transcripts that were not computed for the active transcript (Not fixing for now: https://github.com/cBioPortal/cbioportal/issues/4913)

Oddities:
- [x] If the transcript does not have protein length we ignore it. We wouldn't be able to draw lollipop mutations anyway. Another solution would be to only show the mutations table, but not the lollipop.
- [x] Make sure the pfam domain locations are for the given transcript
- [x] some amino acid changes are bigger than the transcript selected (see EGFR)

Incomplete data handling:
- [x] Make mutation mapper tool work for protein changes only. Also addresses https://github.com/cBioPortal/cbioportal/issues/4806
- [x] For a study w/o genomic positions we should show all the mutations
- [x] list which mutations have been filtered out

Improve data fetching:
- [x] getMutationAlginer.json is called every time when switching. We could save that result (are people using this? Maybe track) (Not fixing for now: https://github.com/cBioPortal/cbioportal/issues/4909)


inodb:
Blocked by: https://github.com/genome-nexus/genome-nexus/pull/221

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -147,6 +171,23 @@ export default class MutationMapperStore
...
     }, undefined);
 
+    @computed get transcriptsWithAnnotations(): string[] {
+        if (this.indexedVariantAnnotations.result && this.mutationData.result && this.allTranscripts.result) {
+            // ignore transcripts without protein length
+            // TODO: better solution is to show only mutations table, not
+            // lollipop plot for those transcripts
+            const transcriptsForThisGeneWithProteinLength: string[] = _.compact(this.allTranscripts.result.map((et: EnsemblTranscript) => et.proteinLength && et.transcriptId));
+
+            const rv:string[] = _.uniq([].concat.apply([], (uniqueGenomicLocations(this.getMutations()).map((gl:GenomicLocation) => {
```
onursumer: Maybe spell out rv, what's rv? :)

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -177,6 +218,25 @@ export default class MutationMapperStore
...
+
+    @computed get mutationsByTranscriptId(): {[transcriptId:string]: Mutations[]} {
+        if (this.indexedVariantAnnotations.result) {
+            return _.fromPairs(this.transcriptsWithAnnotations.map((t:string) => (
+                                  [t,
+                                   getMutationsToTranscriptId(this.getMutations(),
+                                                              t,
+                                                              this.indexedVariantAnnotations.result!!)
+                                  ]
+                               ))
```
onursumer: looks like we have too much indentation here

src/shared/lib/MutationAnnotator.ts
```diff
@@ -99,6 +100,68 @@ export function filterMutationByTranscriptId(mutation:Mutation,
...
+        return getMutationFromSummary(mutation, variantAnnotation.annotation_summary, transcriptConsequenceSummary, true);
+    } else {
+        return undefined;
+    }
+}
+
+export function getMutationFromSummary(mutation:Mutation,
+                                       annotationSummary:VariantAnnotationSummary,
+                                       transcriptConsequenceSummary:TranscriptConsequenceSummary,
+                                       overwrite:boolean)
```
onursumer: maybe add a default value `false` for overwrite?

env/master.sh
```diff
@@ -1,2 +1,2 @@
 export CBIOPORTAL_URL="http://www.cbioportal.org"
-export GENOME_NEXUS_URL="https://www.genomenexus.org"
+export GENOME_NEXUS_URL="https://genomenexus-pr-221.herokuapp.com"
```
onursumer: I guess this is a temporary change to make the pr work, let's not forget to change it back before merging.

src/shared/api/genomeNexusClientInstance.ts
```diff
@@ -8,7 +8,7 @@ async function checkVersion(client:GenomeNexusAPI) {
     }
 }
 
-const client = new GenomeNexusAPI(getGenomeNexusApiUrl());
+const client = new GenomeNexusAPI("https://genomenexus-pr-221.herokuapp.com/");
```
onursumer: I guess this is a temporary change to make the pr work, let's not forget to change it back before merging.

jjgao:
@inodb great documentation 👍

jjgao:
and functionality :)

src/shared/lib/MutationAnnotator.ts
inodb: i think it's good to be explicit in this case, it's called once with false, once with true, so there's not really a clear preference over which should be the default.

`overwrite` parameter might be slightly confusing, cause initAnnotatedMutation always deepclones the entire mutation. It's either using the existing values or overwriting them in the new object

Procfile
```diff
@@ -1 +1 @@
-web: ls dist.backup || cp -a dist/ dist.backup; cp -a dist.backup dist; for f in $(find dist/); do sed -i "s|replace_me_env_cbioportal_url|${CBIOPORTAL_URL:-http\://www.cbioportal.org}|g" $f; sed -i "s|replace_me_env_genome_nexus_url|${GENOME_NEXUS_URL:-https\://www.genomenexus.org}|g" $f; done && http-server --cors dist/ -p ${PORT}
+web: ls dist.backup || cp -a dist/ dist.backup; cp -a dist.backup dist; for f in $(find dist/); do sed -i "s|replace_me_env_cbioportal_url|${CBIOPORTAL_URL:-http\://www.cbioportal.org}|g" $f; sed -i "s|replace_me_env_genome_nexus_url|${GENOME_NEXUS_URL:-https\://www.genomenexus.org}|g" $f; done && pushstate-server dist ${PORT}
```
inodb: to handle routing in frontend on heroku (already in rc)

src/pages/tools/mutationMapper/resources/standaloneMutationDataExample.txt
```diff
@@ -119,11 +119,8 @@ TCGA-S3-A6ZG-01	Breast Invasive Carcinoma, NOS	E928V	Missense_Mutation	7	5526649
 TCGA-17-Z026-01	Lung Adenocarcinoma	G721V	Missense_Mutation	7	55241714	55241714	G	T	5	21		46
 TCGA-50-5066-01	Lung Adenocarcinoma	L210=	Splice_Region	7	55220240	55220240	G	T	30	279		302
 TCGA-33-4566-01	Lung Squamous Cell Carcinoma	X1039_splice	Splice_Region	7	55269425	55269425	C	T	38	154		180
-TCGA-O2-A52N-01	Lung Squamous Cell Carcinoma	EGFR-CHODL	Fusion	7	-1	-1	NA
```
inodb: removed incorrect entries from the example, included one in the tests instead

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -147,6 +177,37 @@ export default class MutationMapperStore
         }
     }, undefined);
 
+    @computed get transcriptsWithAnnotations(): string[] {
```
adamabeshouse: why not make this and transcriptsWithProteinLength mobxpromises?

inodb: @adamabeshouse 	i guess im a bit confused about when to use computed over a promise. Could you clarify?

adamabeshouse: the advantages to using a promise are that you can await it, you can check its status when using it. so you could show a loading indicator for the dropdown until these results were loaded. but if thats not an issue, then @computed should be fine

inodb: i see thanks!

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -147,6 +177,66 @@ export default class MutationMapperStore
...
+                    getMutationsToTranscriptId(this.getMutations(),
+                                            t,
+                                            this.indexedVariantAnnotations.result!!).length > 0
+                ));
+            } else {
+                return [];
+            }
+        },
+        onError: (err: Error) => {
+            throw new Error("Failed to get transcriptsWithAnnotations");
```
adamabeshouse: Im pretty sure this will only show up in the console. If you want to return this particular error from the mobxpromise, for use in the code, then i think you would have to wrap the `invoke` body with `try..catch` and throw this error manually. Im not 100% on that but I am pretty sure

inodb: Thanks! Not doing much error handling in current production either atm, console message is ok for now i think. Filed an issue to improve error handling: https://github.com/cBioPortal/cbioportal/issues/5018

PR #1430, opened 2018-08-31
SV scatter plot improvements, and other improvements

adamabeshouse:


zhx828:
I will keep testing, but there are two changes you can make now
- [ ] Remove the multip group analysis when the chart is deleted
- [ ] The selected cases(red dots) should be shown above the blue dots,

PR #1432, opened 2018-09-05
Implement gene set option in plots tab

oplantalech:
Add gene sets in the Plots tab. You test the functionality here: https://cbioportal-test.thehyve.net:8081/cbioportal/

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -54,10 +54,16 @@ import GeneMolecularDataCache from "../../shared/cache/GeneMolecularDataCache";
 import GenesetMolecularDataCache from "../../shared/cache/GenesetMolecularDataCache";
 import GenesetCorrelatedGeneCache from "../../shared/cache/GenesetCorrelatedGeneCache";
 import GeneCache from "../../shared/cache/GeneCache";
+import GenesetCache from "../../shared/cache/GenesetCache";
 import {IHotspotIndex} from "../../shared/model/CancerHotspots";
 import {IOncoKbData} from "../../shared/model/OncoKB";
 import {generateQueryVariantId} from "../../shared/lib/OncoKbUtils";
-import {CosmicMutation, AlterationEnrichment, ExpressionEnrichment} from "../../shared/api/generated/CBioPortalAPIInternal";
+import {
+    CosmicMutation,
+    AlterationEnrichment,
+    ExpressionEnrichment,
+    Geneset
+} from "../../shared/api/generated/CBioPortalAPIInternal";
 import internalClient from "../../shared/api/cbioportalInternalClientInstance";
 import {IndicatorQueryResp} from "../../shared/api/generated/OncoKbAPI";
 import {getAlterationString} from "../../shared/lib/CopyNumberUtils";
@@ -1674,6 +1680,13 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly genesets = remoteData<Geneset[]>({
+        invoke: async () => internalClient.fetchGenesetsUsingPOST({genesetIds: this.genesetIds.slice()}),
+        onResult:(genesets:Geneset[])=>{
+            this.genesetCache.addData(genesets);
+        }
+    });
+
     readonly entrezGeneIdToGene = remoteData<{[entrezGeneId:number]:Gene}>({
         await: ()=>[this.genes],
         invoke: ()=>Promise.resolve(_.keyBy(this.genes.result!, gene=>gene.entrezGeneId))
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -918,7 +1006,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                             /> Apply Log Scale
                         </label></div>
                     )}
-                    <div className="form-group" style={{opacity:(axisSelection.dataType === CLIN_ATTR_DATA_TYPE ? 0 : 1)}}>
+                    {(axisSelection.dataType !== CLIN_ATTR_DATA_TYPE && axisSelection.dataType !== GENESET_DATA_TYPE) && (<div className="form-group">
```
adamabeshouse: I was using `opacity` because that way the size of the panel doesn't change, which may be smoother UX. @alisman what do you think?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2454,6 +2467,10 @@ export class ResultsViewPageStore {
         return new GeneCache();
     }
 
+    @cached get genesetCache() {
```
alisman: where is this being used (the cache)?  We are actually moving away from using caches like this so I wouldn't bother with this unless there's something i'm misunderstanding.    We have implemented a client level cache (not merged yet) that will take care of this in a more abstract way.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -493,13 +536,36 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
+        // computed instead of remoteData in order to make the rerender synchronous when the 'Same gene set (GENE SET)'
+        //  option changes. if its remoteData, theres setTimeout(0)'s in the way and it causes unnecessarily an extra
+        //  render which leads to a flash of the loading icon on the screen
+        let sameGenesetOption = undefined;
+        if (this.horzSelection.selectedGenesetOption && this.horzSelection.dataType === GENESET_DATA_TYPE) {
+            // show "Same gene set" option as long as horzSelection has a selected option, and horz is gene set attribute, bc
+            //  in that case theres no selected gene displayed so its confusing UX to have "Same gene" as an option
+            sameGenesetOption = [{ value: SAME_GENESET_OPTION_VALUE, label: `Same gene set (${this.horzSelection.selectedGenesetOption.label})`}];
+        }
+        return (sameGenesetOption || []).concat((this.horzGenesetOptions.result || []) as any[]);
```
alisman: seems like should be typed with same as horzGenesetOptions.result

alisman: i.e. not any[]

src/pages/resultsView/ResultsViewPageStore.ts
oplantalech: It is used in line 1686 of this same file to retrieve the gene sets, similarly as done with genes.

src/pages/resultsView/plots/PlotsTab.tsx
oplantalech: If you prefer, I can keep the "gene" selection box hidden when selecting the clinical attribute instead of removing it to keep the size of the panel, as you implemented. Be aware though that now there are have four selection boxes (gene and gene set boxes are independent).

alisman: @adamabeshouse I agree we want to avoid jumpiness.   Oleg, I'm not sure how to ansewr question.  I think I need to see it to undersatnd.  Can you put screenshot or do demo for me?

src/pages/resultsView/ResultsViewPageStore.ts
alisman: but where is it actually consumed by view layer?  like where is cache data actually accessed?

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: @oplantalech would there ever be the gene and geneset box visible at the same time?

oplantalech: @adamabeshouse No. Geneset box only appears when the geneset datatype is selected

adamabeshouse: @oplantalech would it be possible to reuse the same component, and just insert different options and a different label and a different onChange method?

oplantalech: @adamabeshouse I guess so, but I think it is more logical to be a separate component. If you want to keep the same size for the selection boxes, I can keep the gene box in place without the option to see it when clinical attributes are selected (what you implemented).

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -477,7 +477,7 @@ function makeAxisDataPromise_Geneset(
                     const value = d.value;
                     return {
                         uniqueSampleKey: d.uniqueSampleKey,
-                        value
+                        value: Number(value)
```
adamabeshouse: wow good find :D

PR #1434, opened 2018-09-06
SV query update

kalletlak:
# What? Why?
1. able to query virtual study from new SV, https://github.com/cBioPortal/cbioportal/issues/4463


src/shared/components/query/QueryStore.ts
```diff
@@ -275,40 +242,18 @@ export class QueryStore
 	@computed get selectableSelectedStudyIds():string[]
 	{
 		let ids:string[] = this._allSelectedStudyIds.keys();
-		const selectableStudies = this.selectableStudiesSet;
+		const selectableStudies = this.selectableStudiesSet.result;
```
alisman: check to make sure it's not undefined (what is default of this promise?)

alisman:
There's a lot of logic that we're going to need to test. Might as well at least start by putting this in helper files

src/shared/components/query/QueryStore.ts
```diff
@@ -563,6 +504,194 @@ export class QueryStore
...
+							acc[studyId] = _.keys(next)
+							return acc;
+						},{} as {[id: string]: string[]})
+					}
+
+					let isStudySampleMapEmpty = true;
+					_.each(studySampleMap,(samples) => {
+						if(samples.length > 0 ){
+							isStudySampleMapEmpty = false;
+							return false;
```
alisman: returning false from _.each won't break looping.    this looks to me like a case for using _.some

src/shared/components/query/QueryStore.ts
```diff
@@ -563,6 +504,194 @@ export class QueryStore
...
+
+		result = _.reduce(this.sharedVirtualStudiesSet.result,(acc, next, key)=>{
+			acc[key] = next.data.studies.map(study=> study.id)
+			return acc;
+		},result)
+		return  result;
+		},
+		default: {},
+		onResult:(selectableStudiesSet)=>{
+			if(this.selectedSampleListId !== CUSTOM_CASE_LIST_ID) {
```
alisman: lets make a practice of always putting a comment where we use onResult explaining why we need to use onResult.   Basically, my understanding is that it's because we want to set a default user selection, right?

src/shared/components/query/QueryStore.ts
kalletlak: default is `{}`

src/shared/components/query/QueryStore.ts
```diff
@@ -275,40 +242,18 @@ export class QueryStore
...
-			} else {
-				studyToSampleSet[id] = {} //an empty list indicates all to select all samples
-			}
-		});
-		return studyToSampleSet;
-	}
-
 	set selectableSelectedStudyIds(val:string[]) {
 		//surrounded with action block to indicate that state is going to be modified
 		action(()=>{
```
alisman: seems weird that we do filtering on set.  seems to me filtering should take place on reading or in code which calls set.  i.e it should not be hidden here.   also, we should not call action inside of a setter because again that hides the fact that it's an action from "controlling" code.  also, developer may double wrap with action because they will think they need to.  there's nothing to let them know there's intervening code when then set the property.

kalletlak:
@alisman looks like there is lot redundant/unnecessary code. I am going to reuse gene box and custom case box which were written for study-view and add tests in another pr

src/shared/components/query/QueryContainer.tsx
```diff
@@ -69,7 +69,7 @@ export default class QueryContainer extends React.Component<QueryContainerProps,
         return (
 			<FlexCol padded overflow className={styles.QueryContainer}>
                 {
-                    <UnknownStudiesWarning ids={this.store.unknownStudyIds} />
+                    <UnknownStudiesWarning ids={this.store.unknownStudyIds.result} />
```
alisman: this is a problem i already fixed in the SPA (waiting for unknonwStudyIds) but any time you check result you should really be first checking status.   We should check status wherever we deference the mobxpromise (i.e. refer to result property)

src/shared/components/query/QueryStore.ts
```diff
@@ -563,6 +499,181 @@ export class QueryStore
...
+			let result:{[studyId:string]:VirtualStudy} = {}
+			
+			await Promise.all(unknownQueriedIds.map(id =>{
+				return new Promise((resolve, reject) => {
+					sessionServiceClient.getVirtualStudy(id).then((virtualStudy)=>{
+						//physical study ids iin virtual study
+						let ids = virtualStudy.data.studies.map(study=>study.id);
+						//unknown/unauthorized studies within virtual study
+						let unKnownPhysicalStudyIds = ids.filter(id => !physicalStudiesIdsSet[id]);
+						if(_.isEmpty(unKnownPhysicalStudyIds)){
```
alisman: lets not overuse isEmpty.   i believe filter will return empty array if no matches, so just check for length of the array > 0

src/shared/components/query/QueryStore.ts
```diff
@@ -563,6 +499,181 @@ export class QueryStore
...
+			return studyToSampleSet;
+		},
+		default: {}
+	})
+
+	private readonly selectableStudiesSet = remoteData<{[studyId:string]:string[]}>({
+		await: ()=>[this.physicalStudiesSet, this.selectedStudyToSampleSet],
+		invoke: async ()=>{
+			let result:{[studyId:string]:string[]} = {}
+		result = _.reduce(this.physicalStudiesSet.result,(acc, next, key)=>{
```
alisman: this is a case where it's confusing to use reduce because you have to pass in a previously assigned variable.   here it's better to use each.   you don't have to change this i'm just pointing it out.

src/shared/components/query/QueryContainer.tsx
alisman: Just to be clear, you can keep what you've done, just do check for status

PR #1435, opened 2018-09-07
study view summary header updates

kalletlak:
1. Hide gene box validation message
2. Show separate breadcrumb filters for scatter plot and custom case selection

### DIFF
--- a/src/pages/studyView/UserSelections.tsx
+++ b/src/pages/studyView/UserSelections.tsx
@@ -4,19 +4,19 @@ import { observer } from "mobx-react";
 import { computed } from 'mobx';
 import styles from "./styles.module.scss";
 import {ClinicalDataIntervalFilterValue, StudyViewFilter} from 'shared/api/generated/CBioPortalAPIInternal';
-import { ChartMeta, UniqueKey } from 'pages/studyView/StudyViewPageStore';
+import { ChartMeta, UniqueKey, StudyViewFilterWithSampleIdentifierFilters} from 'pages/studyView/StudyViewPageStore';
 import {intervalFiltersDisplayValue, isFiltered} from 'pages/studyView/StudyViewUtils';
 import autobind from 'autobind-decorator';
 
 export interface IUserSelectionsProps {
-    filter: StudyViewFilter;
+    filter: StudyViewFilterWithSampleIdentifierFilters;
     attributesMetaSet: { [id: string]: ChartMeta };
     updateClinicalDataEqualityFilter: (chartMeta: ChartMeta, value: string[]) => void;
     updateClinicalDataIntervalFilter: (chartMeta: ChartMeta, values: ClinicalDataIntervalFilterValue[]) => void;
     clearGeneFilter: () => void;
     clearCNAGeneFilter: () => void;
-    clearCustomCasesFilter: () => void;
-    clearAllFilters: () => void;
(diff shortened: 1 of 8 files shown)


src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
```diff
@@ -55,7 +56,7 @@ export default class GeneSymbolValidator extends React.Component<IGeneSymbolVali
 				</div>
 			);
 
-		if (this.props.genes.isPending && this.props.genes.result!.suggestions.length == 0)
+		if (this.props.genes.isPending && this.props.genes.result!.suggestions.length == 0 && !this.props.hideValidatingMessage)
```
alisman: lets use either just the false (this.props.genes.result!.suggestions.length) or === but not double equal sign

src/pages/studyView/UserSelections.tsx
```diff
@@ -108,50 +109,77 @@ export default class UserSelections extends React.Component<IUserSelectionsProps
...
     chartMeta: ChartMeta;
-    values: string[];
-    updateClinicalDataEqualityFilter: (chartMeta: ChartMeta, value: string[]) => void;
+    labelValueSet?: { [id: string]: any };
+    updateFilter?: (chartMeta: ChartMeta, value: any[]) => void;
+    clearFilter?: (chartMeta: ChartMeta) => void; // could be used if we want to remove all filters for that particular chart
 }
 
-class ClinicalDataEqualityFilter extends React.Component<IClinicalDataEqualityFilterProps, {}> {
+@observer
```
alisman: can you fix the indentation of this?

src/pages/studyView/UserSelections.tsx
```diff
@@ -64,26 +66,26 @@ export default class UserSelections extends React.Component<IUserSelectionsProps
...
-            return (
-                <span className={styles.filter}>
-                    <span className={styles.name}>{chartMeta.displayName}</span>
-                    <i className="fa fa-times" style={{ cursor: "pointer" }} onClick={this.props.clearCustomCasesFilter}></i>
-                </span>
-            )
-        }
-        return null;
+    @computed private get sampleIdentifiersFilters() {
+        return _.map(this.props.filter.sampleIdentifiersSet, (sampleIdentifiers, chartUniqKey) => {
```
alisman: lets use reduce instead of returning null values.  or filter then man if there aren't going to be that many items in this list and we don't care too much about performance.

src/pages/studyView/UserSelections.tsx
```diff
@@ -31,31 +30,31 @@ export default class UserSelections extends React.Component<IUserSelectionsProps
...
-                return (
-                    <ClinicalDataEqualityFilter
-                        chartMeta={chartMeta}
-                        values={clinicalDataEqualityFilter.values}
-                        updateClinicalDataEqualityFilter={this.props.updateClinicalDataEqualityFilter}
-                    />
-                )
-            } else {
-                return null;
+                let labelValueSet = _.reduce(clinicalDataEqualityFilter.values, (acc, next) => {
```
alisman: no big deal but you could use _.keyBy.    i would say use _.reduce only when there's some kind of statement involved, like you're not operating for every element of input array.    you don't not need to change this.

PR #1438, opened 2018-09-10
Expression tab fixes

adamabeshouse:
https://github.com/cBioPortal/cbioportal/issues/4720
https://github.com/cBioPortal/cbioportal/issues/4755


src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -616,6 +616,23 @@ export default class BoxScatterPlot<D extends IBaseBoxScatterPlotPoint> extends
         );
     }
 
+    @autobind
+    private getTooltip() {
```
alisman: how come not computed?  i guess observer will prevent re render?

adamabeshouse: @alisman right. observer needs a function ()=>ReactElement

PR #1446, opened 2018-09-12
Co-expression table use scientific notation

inodb:
- [x] conditional significance formatting (bold for q-value < 0.05, same as enrichments)
- [x] enrichments sorts by q value, then p value


src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -96,6 +96,10 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
     @computed get molecularProfileIdToMolecularProfile() {
         return _.keyBy(this.profiles, profile=>profile.molecularProfileId);
     }
+    
+    private sortCoExpressionData(data: any[]): any[] {
+        return _.sortBy(data, ["pValue", "hugoGeneSymbol"]);
```
adamabeshouse: another way to do this, which may be a little cleaner, is to adjust the `sortBy` function for the qValue column to be d=>[d.qValue, d.pValue]

PR #1450, opened 2018-09-13
Update study name for multi study query

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/3624


kalletlak:
seems not, we just have
screen shot 2018-09-14 at 11 10 37 am

PR #1454, opened 2018-09-13
Added missing download functionalities for survival and scatter charts

onursumer:
# Checks
  the `-s` option)


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -141,6 +146,48 @@ export function makeMutationCountVsCnaTooltip(sampleToAnalysisGroup?:{[sampleKey
...
+
+        if (analysisGroups !== undefined) {
+            header.push("Color");
+            valueToGroup = _.keyBy(analysisGroups, 'value');
+        }
+    }
+
+    const rows = data.map(datum => {
+        const row = [
+            `${datum.studyId}`,
```
alisman: curious why you are wrapping in string this way.  why not just plain variable or var.toString?

onursumer: `undefined.toString()` gives an error, but this doesn't. Though, not sure if we really want to see `undefined` as an output.

PR #1458, opened 2018-09-17
Fix two issues with study view involving unncessary reloading

adamabeshouse:


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -104,15 +105,24 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
         reaction(
             () => props.routing.location.query,
             query => {
+                let newStudyIds:string[]|null = null;
```
adamabeshouse: this changes would be replaced by @kalletlak

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -104,15 +105,24 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
-                    this.store.sampleAttrIds = ('sampleAttrIds' in query ? (query.sampleAttrIds as string).split(",") : []);
-                    this.store.patientAttrIds = ('patientAttrIds' in query ? (query.patientAttrIds as string).split(",") : []);
+                if (newStudyIds !== null) {
+                    const newStudyIdsMap = stringListToSet(newStudyIds);
+                    const existingStudyIdsMap = stringListToSet(this.store.studyIds);
+                    if (!_.isEqual(newStudyIdsMap, existingStudyIdsMap)) {
+                        // if new study ids in routing location, update store
+                        runInAction(()=>{
+                            this.store.studyIds = newStudyIds!;
+                            this.store.sampleAttrIds = ('sampleAttrIds' in query ? (query.sampleAttrIds as string).split(",") : []);
```
alisman: could we change sampleAttrIds independent of studyIds?  If so, this wouldn't work right?   Also, we should perform same deep equality check on these and only update if unequal

kalletlak: I got rid of `sampleAttrIds` and `patientAttrIds` as we are not using this feature in production.

PR #1459, opened 2018-09-17
fix bug where profiled in tracks were appearing when they shouldnt

adamabeshouse:


src/shared/components/oncoprint/ResultsViewOncoprintUtils.ts
```diff
@@ -52,11 +56,26 @@ export function makeProfiledInClinicalAttributes(
                 }
             }
         });
+
```
alisman: i think this whole thing needs testing.  i mean, if anything does, it's this, right?

adamabeshouse: good point, I covered it with a screen shot but Ill add a case for this to the spec file

PR #1461, opened 2018-09-18
Add samplesPerPatient, withMutationData, withCnaData charts

kalletlak:
# What? Why?
Adds 3 custom pie charts (https://github.com/cBioPortal/cbioportal/issues/4701)
1. No. of Samples Per Patient
2. With Mutation Data
3. With CNA Data pie charts

And
CSS updates in custom case selection box and virtual study box

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -5,7 +5,7 @@ import {ChartContainer, IChartContainerProps} from 'pages/studyView/charts/Chart
 import { MSKTab, MSKTabs } from "../../shared/components/MSKTabs/MSKTabs";
 import { reaction, observable } from 'mobx';
 import { If } from 'react-if';
-import {ChartMeta, ChartType, ChartTypeEnum, DataBinMethodConstants, StudyViewPageStore, AnalysisGroup} from 'pages/studyView/StudyViewPageStore';
+import {ChartMeta, ChartType, ChartTypeEnum, DataBinMethodConstants, StudyViewPageStore, AnalysisGroup, UniqueKey, CUSTOM_CHART_KEYS} from 'pages/studyView/StudyViewPageStore';
 import SummaryHeader from 'pages/studyView/SummaryHeader';
 import {Gene, SampleIdentifier, ClinicalAttribute} from 'shared/api/generated/CBioPortalAPI';
 import { SingleGeneQuery } from 'shared/lib/oql/oql-parser';
@@ -101,7 +101,10 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps, 
             },
             updateChartsVisibility: (visibleChartIds: string[]) => {
                 this.store.updateChartsVisibility(visibleChartIds);
-            }
+            },
+            setCustomChartFilters: (chartMeta: ChartMeta, values: string[]) => {
+                this.store.setCustomChartFilters(chartMeta, values);
+            },
 
         }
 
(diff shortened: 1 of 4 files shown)


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -139,13 +142,29 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
-                props.promise = this.store.studyViewClinicalDataCountsCache.get({
-                    attribute: chartMeta.clinicalAttribute!,
-                    filters: this.store.filters
-                });
-                props.filters = this.store.getClinicalDataFiltersByUniqueKey(chartMeta.uniqueKey);
-                props.onValueSelection = this.handlers.onValueSelection;
-                props.onResetSelection = this.handlers.onValueSelection;
+
+                //custom charts
+                if (_.includes([UniqueKey.SAMPLES_PER_PATIENT, UniqueKey.WITH_CNA_DATA, UniqueKey.WITH_MUTATION_DATA], chartMeta.uniqueKey)) {
```
alisman: give a name to that array of UniqueKeys so we can understand what we're matching against. and a comment here to explain

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1750,6 +1819,27 @@ export class StudyViewPageStore {
                     ret[UniqueKey.MUTATION_COUNT_CNA_FRACTION] = (this.mutationCountVsFractionGenomeAlteredData.result || []).length;
                 }
 
+                if (UniqueKey.SAMPLES_PER_PATIENT in this.chartMetaSet) {
+                    ret[UniqueKey.SAMPLES_PER_PATIENT] = _.reduce(this.samplesPerPatientData.result, (acc, next) => {
```
alisman: i think you could use lodash _.sum

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1844,6 +1934,182 @@ export class StudyViewPageStore {
 
         formSubmit('index.do', formOps, "_blank", 'post');
     }
+
+    @action
+    setCustomChartFilters(chartMeta: ChartMeta, values: string[]) {
```
alisman: this method is too long.  impossible to read/understand/test.   i'm pretty sure this should be a switch(chartMeta.uniqueKey).  then call nicely named methods to reduce sample collections

src/pages/studyView/StudyViewUtils.spec.ts
```diff
@@ -1303,4 +1305,75 @@ describe('StudyViewUtils', () => {
...
+            getSamplesByExcludingFiltersOnChart(
+                'WITH_MUTATION_DATA',
+                emptyStudyViewFilter,
+                { 'WITH_MUTATION_DATA': [{ sampleId: 'sample1', studyId: 'study1' }], 'WITH_CNA_DATA': [{ sampleId: 'sample1', studyId: 'study1' }] },
+                [{ sampleId: 'sample1', studyId: 'study1' }, { sampleId: 'sample2', studyId: 'study1' }],
+                ['study1']
+            ).then(() => {
+                assert.isTrue(fetchStub.calledWith({ studyViewFilter: { ...emptyStudyViewFilter, sampleIdentifiers: [{ sampleId: 'sample1', studyId: 'study1' }] } }));
+                done();
+            }).catch(done);
```
alisman: nice tests!  how come you want to catch though?  i think this would suppress a runtime error

kalletlak: yes

PR #1477, opened 2018-09-21
plots tab - if sample has multiple mutation types, give it type Multiple

adamabeshouse:


adamabeshouse:
@alisman theres going to be screenshot updates for this so lets wait to merge it until e2e is fixed

PR #1481, opened 2018-09-24
Improved FixedHeaderTable store updates

onursumer:
# What? Why?
This fixes re-rendering issues of `FixedHeaderTable`.

# Checks
  the `-s` option)


src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -60,6 +54,11 @@ export default class FixedHeaderTable<T> extends React.Component<IFixedHeaderTab
         this.initDataStore();
     }
 
+    componentWillReceiveProps() {
+        // TODO do not re-initialize the whole store, just update the changed props within the store
+        this.initDataStore();
+    }
+
     @action
     initDataStore() {
```
onursumer: This function was tagged as `@action`, but it was not changing any observable properties.

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -22,10 +22,6 @@ export type IFixedHeaderTableProps<T> = {
     sortBy?: string;
     width?: number;
     height?: number;
-    // TODO: These properties are just used to force to rerender the table when selectedGenes is modified.
-    // We don't actually use the property anywhere in the table.
-    selectedGenes?: string[];
-    selectedRows?: number[];
```
onursumer: We don't need these anymore, since now we observe the store.

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -38,14 +34,12 @@ const RVSDTtoStrType = {
 
 @observer
 export default class FixedHeaderTable<T> extends React.Component<IFixedHeaderTableProps<T>, {}> {
-    private _store: LazyMobXTableStore<T>;
+    @observable private _store: LazyMobXTableStore<T>;
```
onursumer: Making the whole `_store` observable may not be the best idea, but this is a quick fix for the re-rendering issues and helps removing the unused properties.

zhx828: Make the whole store observable?

onursumer: @zhx828 store is not `observable` anymore, instead we are calling `store.updateProps` function which is an `action` itself.

PR #1503, opened 2018-09-26
Merge master to rc

zhx828:


inodb:
Not sure why the dots moved in the plots tab: https://16078-66571349-gh.circle-artifacts.com/0/imageCompare.html. We also should prolly exclude the footer from the study view screenshot

PR #1510, opened 2018-09-28
coexpression and enrichment tabs - use same molecular profile options and component

adamabeshouse:


src/pages/resultsView/enrichments/EnrichmentsDataSetDropdown.tsx
```diff
@@ -17,33 +20,32 @@ export interface IEnrichmentsDataSetDropdownProps {
...
-                <Form inline>
-                    <FormGroup controlId="formControlsSelect" bsSize="small">
-                        <ControlLabel>Data Set:</ControlLabel>
-                        <FormControl componentClass="select" style={{marginLeft:5}} onChange={this.change} value={this.props.selectedValue}>
-                            {options}
-                        </FormControl>
-                    </FormGroup>
-                </Form>
+                <div style={{display:"flex", alignItems:"center"}}>
+                    <span style={{fontWeight:"bold"}}>Data Set:</span>
```
alisman: You can use <strong /> tag which bootstrap renders as bold

src/shared/components/MolecularProfileSelector.tsx
```diff
@@ -0,0 +1,75 @@
...
+    onChange:(option:{label:string, value:string})=>void;
+    molecularProfiles:MolecularProfile[] | MobxPromise<MolecularProfile[]>;
+    molecularProfileIdToProfiledSampleCount:{[molecularProfileId:string]:number} | MobxPromise<{[molecularProfileId:string]:number}>;
+}
+
+@observer
+export default class MolecularProfileSelector extends React.Component<IMolecularProfileSelector, {}> {
+    readonly resolvedPromise = remoteData(()=>Promise.resolve());
+
+    readonly options = remoteData({
```
alisman: why not just deal with async loading of options outside of component?

adamabeshouse: the main functionality here is to let anyone use the same component when dealing w molecular profiles and available sample data counts. we want to encapsulate that logic

PR #1516, opened 2018-10-02
Updated visualize.tsx with the latest tools.jsp content

onursumer:
# What? Why?
Fix  https://github.com/cBioPortal/cbioportal/issues/4904

visualize-tools

# Checks
  the `-s` option)


end-to-end-tests/specs/mutationMapperTool.spec.js
```diff
@@ -14,7 +14,7 @@ describe('Mutation Mapper Tool', function() {
 
     describe('example genomic changes input', ()=>{
         beforeEach(()=>{
-            var url = `${CBIOPORTAL_URL}/mutation_mapper.jsp`;
+            var url = `${CBIOPORTAL_xURL}/s/mutation_mapper`;
```
onursumer: @alisman is this the right way to fix this route?

src/routes.jsx
```diff
@@ -81,7 +81,7 @@ export const makeRoutes = (routing) => {
                 <Route path="/newstudy" onEnter={()=>{$(document).scrollTop(0)}} getComponent={lazyLoadComponent(StudyViewPage)} />
                 <Route path="/study" component={getBlankPage()} />
 
-                <Route path="/mutation_mapper" getComponent={lazyLoadComponent(MutationMapperTool)} />
+                <Route path="/s/mutation_mapper" getComponent={lazyLoadComponent(MutationMapperTool)} />
```
onursumer: Do we still need to keep the original path `/mutation_mapper`? I think the only place where we refer this route is in `mutation_mapper.jsp` which we can retire now.

end-to-end-tests/specs/mutationMapperTool.spec.js
adamabeshouse: is there a typo CBIOPORTAL_*x*URL

onursumer: oops, yes :)

PR #1519, opened 2018-10-02
Updates to study view

kalletlak:
Fixes 
1. https://github.com/cBioPortal/cbioportal/issues/4899
2. tweaks to clinical data tab

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -50,7 +50,6 @@ import {
     isLogScaleByDataBins,
     isPreSelectedClinicalAttr,
     makePatientToClinicalAnalysisGroup,
-    EXPONENTIAL_FRACTION_DIGITS,
     generateScatterPlotDownloadData,
     NA_DATA, ONE_GRID_TABLE_ROWS, PIE_TO_TABLE_LIMIT,
     COLORS, NA_COLOR,
@@ -58,7 +57,8 @@ import {
     getFilteredSampleIdentifiers,
     UNSELECTED_GROUP_COLOR, SELECTED_GROUP_COLOR,
     showOriginStudiesInSummaryDescription,
-    getFilteredStudiesWithSamples
+    getFilteredStudiesWithSamples,
+    submitToPage
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
 import {SingleGeneQuery} from 'shared/lib/oql/oql-parser';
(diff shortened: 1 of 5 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2041,12 +2040,14 @@ export class StudyViewPageStore {
...
 
+        let url = '/';
         if (!_.isEmpty(this.geneQueries)) {
             formOps.Action = 'Submit';
             formOps.gene_list = this.geneQueries.map(query => unparseOQLQueryLine(query)).join('\n');
+            url = '/results/legacy_submission';
         }
-
-        formSubmit('index.do', formOps, "_blank", 'post');
+        window.localStorage.setItem("legacyStudySubmission",JSON.stringify(formOps))
```
alisman: could you do me a favor and wrap this in a function that we import and which abstracts away the actual submission method?  you could call it, submitToPage

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1121,4 +1121,18 @@ export function clinicalDataCountComparator(a: ClinicalDataCount, b: ClinicalDat
...
     }
+}
+
+export function submitToPage(url:string, params: { [id: string]: string }, target?: string) {
+    window.localStorage.removeItem("legacyStudySubmission");
+    try {
+        window.localStorage.setItem("legacyStudySubmission", JSON.stringify(params))
+    } catch (e) {
+        // in case of this QuotaExceededError clear case_ids filed as this is the only one which could be longest 
+        console.log(e);
```
alisman: hmmm. i think you need to clear localStorage, not clear the params you are trying to set.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1121,4 +1121,13 @@ export function clinicalDataCountComparator(a: ClinicalDataCount, b: ClinicalDat
     else {
         return b.count - a.count;
     }
+}
+
+export function submitToPage(url:string, params: { [id: string]: string }, target?: string) {
+    try {
+        window.localStorage.setItem("legacyStudySubmission", JSON.stringify(params))
+    } catch (e) {
+        window.localStorage.clear();
```
alisman: this doesn't try to set the submission again after clearing the localStorage!??

alisman: the whole point is the clear localstorage so we can set it.  if we just submit then what are we doing?????

PR #1522, opened 2018-10-03
fixing e2e tests - make sure to update and add screenshots as necessary

adamabeshouse:


adamabeshouse:
@inodb thanks for checking it out!

-`session_cancer_type_summary_element_chrome_1600x1000.png` this was an intentional change

- `profiled_in_tracks_in_multiple_study_with_sos1_element_chrome_1600x1000.png` I had to change the query because the existing session/URL was broken for some reason https://github.com/cBioPortal/cbioportal-frontend/pull/1522/files#diff-7c425d5af090d3e6f4bb964b1a83b98bR196

-`patient_view_lgg_ucsf_2014_p04_element_chrome_1600x1000.png` good catch, I thought it was okay because it seemed to be because of setting the viewport size, but I think Im setting the viewport size wrong.

-`msk_impact_2017_query_stk11:homdel_mut_element_chrome_1600x1000.png` this is intentional, it's just checking a different element, and I have separate tests which cover the oql banner

PR #1523, opened 2018-10-04
Study View updates

kalletlak:
# What? Why?
Fix
1. https://github.com/cBioPortal/cbioportal/issues/4900
2. remove recurrent api call when there are no filter applied
3. https://github.com/cBioPortal/cbioportal/issues/4944
4. https://github.com/cBioPortal/cbioportal/issues/4928

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -1948,6 +1948,8 @@ export class StudyViewPageStore {
                 }
 
                 if (UniqueKey.MUTATION_COUNT_CNA_FRACTION in this.chartMetaSet) {
+                    // number of samples containing mutationCountVsFractionGenomeAlteredData should be 
+                    // calculated from the selected samples
                     const mutationCountVsFractionGenomeAlteredDataSet = this.mutationCountVsFractionGenomeAlteredDataSet.result;
                     const selectedSamplesMap = _.keyBy(this.selectedSamples.result!, s => s.uniqueSampleKey);
                     const filteredData = _.reduce(selectedSamplesMap, (acc, next) => {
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1107,15 +1107,23 @@ export class StudyViewPageStore {
                 } : DEFAULT_LAYOUT_PROPS.dimensions[chartType]);
             });
 
+            const cancerTypeIds = _.uniq(this.queriedPhysicalStudies.result.map(study=>study.cancerTypeId));
```
zhx828: @jjgao we use the `mixed` in the cancer type id field to determine whether it is a mixed study. This should cover most of the use cases. But when user select one cancer type from a mixed study and saved as VS. Next time, when they open the VS, the survival plot should be shown by default?

jjgao: That's a good point. When opening a VS, is there a way to look at the main cancer type, if there is only one, show survival?

Alternatively, we don't have to worry about it. We will move the survival plots to comparisons tab later anyway.

zhx828: Let's postpone it.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1930,7 +1948,15 @@ export class StudyViewPageStore {
                 }
 
                 if (UniqueKey.MUTATION_COUNT_CNA_FRACTION in this.chartMetaSet) {
-                    ret[UniqueKey.MUTATION_COUNT_CNA_FRACTION] = (this.mutationCountVsFractionGenomeAlteredData.result || []).length;
+                    const mutationCountVsFractionGenomeAlteredDataSet = this.mutationCountVsFractionGenomeAlteredDataSet.result;
```
alisman: would be nice to have a comment here explaining this

PR #1525, opened 2018-10-04
survival tab - handle case when no case is altered

adamabeshouse:


src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -107,11 +107,14 @@ export function getScatterDataWithOpacity(patientSurvivals: PatientSurvival[], e
...
-    return [patientSurvivals.length,
-    patientSurvivals.filter(patientSurvival => patientSurvival.status === true).length,
-    getMedian(patientSurvivals, estimates)];
+export function getStats(patientSurvivals?: PatientSurvival[], estimates?: number[]): [number, number, string] {
+    if (patientSurvivals && estimates) {
+        return [patientSurvivals.length,
+        patientSurvivals.filter(patientSurvival => patientSurvival.status === true).length,
+        getMedian(patientSurvivals, estimates)];
+    } else {
+        return [0, 0, "N/A"];
```
alisman: seems like a weird data type?

PR #1527, opened 2018-10-04
SPA enrichments - fix Add Genes to Query

adamabeshouse:


src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -44,6 +45,15 @@ function normalizeLegacySession(sessionData:any){
     return sessionData;
 }
 
+export enum QueryParameter {
```
alisman: i think there is at least one missing.  priority. lets discuss how to get full list

PR #1528, opened 2018-10-04
Hide loader in IFrameLoader when iframe is loaded

adamabeshouse:


alisman:
I actually intended the loader to be hidden by the iframe itself.      Due to a recent z-index change this stopped working but it should be fixed with z-index instead.  So iframe needs higher zindex that the loader in this special case.

adamabeshouse:
ah okay, I'll do that

PR #1529, opened 2018-10-05
Config

alisman:
Adapted application use new configuration service and properties piped directly from portal.properties

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/appBootstrapper.jsx
```diff
@@ -212,4 +194,21 @@ if (__DEBUG__ && module.hot) {
     module.hot.accept('./routes', () => render());
 }
 
-$(document).ready(() => render());
+$(document).ready(async () => {
+
+    // we use rawServerConfig (written by JSP) if it is present
+    // or fetch from config service if not
+    // need to use jsonp, so use jquery
+    let config = window.rawServerConfig || await $.ajax({
```
alisman: Since we still have jsp (and need it for authorization) we might as well write config directly to it rather than fetching config async unnecessarily.   this is trivial to achieve in jsp

src/appBootstrapper.jsx
```diff
@@ -212,4 +194,21 @@ if (__DEBUG__ && module.hot) {
     module.hot.accept('./routes', () => render());
 }
 
-$(document).ready(() => render());
+$(document).ready(async () => {
+
+    // we use rawServerConfig (written by JSP) if it is present
```
alisman: i decided to fetch configuration before we even mount the react application so that react app doesn't have to deal at all with async configuration.  it can assume config is loaded no matter what.

alisman: i could/should probably do this by adding a top layer React component above even the router.  TODO

alisman:
With a couple exceptions, there are two types of configuration values:
boolean and string

for boolean values, we also need to handle case where property is missing (null) or unassigned "" in portal.properties.  For this reason, all booleans should have a default and it should resort to default if value is "" or null.   I checked the logic in methods of global.properties to make sure this is valid.

src/config/config.ts
```diff
@@ -135,4 +144,44 @@ export function initializeAPIClients(){
...
+    // we want to respect frontUrl if it is already set (case where localdist is true)
+    // @ts-ignore: ENV_* are defined in webpack.config.js
+    const frontendUrl = config.frontendUrl || (/\/\/localhost:3000/.test(win.location.href)) ? "//localhost:3000/" : `//${ENV_CBIOPORTAL_URL}/`;
+
+    const configServiceUrl = config.configurationServiceUrl || APIROOT;
+
+    const envConfig: Partial<IAppConfig> = {
+        apiRoot:APIROOT,
+        frontendUrl:frontendUrl,
+        configurationServiceUrl:APIROOT,
```
inodb: should that `configurationServiceUrl:APIROOT` be `configurationServiceUrl: configServiceUrl `?

src/config/IAppConfig.ts
```diff
@@ -1,86 +1,91 @@
...
-    genomespaceEnabled: boolean;
-    skinExampleStudyQueries: string[]; // in query the example searches
-    priorityStudies: PriorityStudies;
-
-    googleAnalyticsProfile?: string;
-
     authLogoutUrl?: string;
     authUserName?: string;
     authGoogleLogin?: string;
+    configurationServiceUrl?: string;
```
inodb: the `auth` stuff above is not `serverConfig` ?

src/config/IAppConfig.ts
```diff
@@ -1,86 +1,91 @@
...
+    //oncoprintHideVUSDefault?:boolean;
+    //sessionServiceIsEnabled?:boolean;
+    //sessionServiceUrl?:string;
+
+    //skinIsMarkdownDocumentation?:boolean;
+    //skinDocumentationBaseUrl?: string;
+    //skinFaqSourceURL?: string;
+    //skinAboutSourceURL?: string;
+    //skinNewsSourceURL?: string;
+    //skinOQLSourceURL?: string;
```
inodb: would prefer to remove all these lines over commenting out

PR #1532, opened 2018-10-05
Mutation lollipop plot label improvements 

onursumer:
# What? Why?
- Improved lollipop plot label positioning (fix https://github.com/cBioPortal/cbioportal/issues/4922)
- Limited number of protein changes to the top 3 when constructing the lollipop label

# Examples
- Lolipop close to origin (right aligned):

right_aligned_longer

right_aligned_4_more

right_aligned_9_more

- Lollipop close to the protein end (left aligned):

left_aligned

# Checks
  the `-s` option)


src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -80,27 +79,59 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
         }))
     }, {});
 
-    private lollipopLabel(mutationsAtPosition:Mutation[]):string {
-        let proteinChanges = _.uniq(mutationsAtPosition.map(m=>m.proteinChange));
-        proteinChanges.sort();
+    private lollipopLabelText(mutationsAtPosition:Mutation[], size?: number): string {
```
onursumer: Ideally we should isolate & test this function and other functions similar to this one in this class. Maybe in a new `LollipopUtils` class?

src/shared/components/SVGAxis.tsx
```diff
@@ -88,7 +88,7 @@ export default class SVGAxis extends React.Component<SVGAxisProps, {}> {
             let y:number;
             let transform:string;
             if (this.props.vertical) {
-                x = this.props.x - this.props.tickLength - 20;
+                x = this.props.x - this.props.tickLength - 30;
```
onursumer: This pushes the y-axis label of the mutation diagram further left, so that it doesn't overlap with 4-digit mutation counts.

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -80,27 +79,59 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
...
+        if (proteinChanges.length < mutationCountsByProteinChange.length) {
+            label = `${label} and ${mutationCountsByProteinChange.length - proteinChanges.length} more`;
+        }
+
+        return label;
+    }
+
+    private lollipopLabelTextAnchor(labelText: string, codon: number, fontSize: number): string {
+        let anchor = "middle";
+        const approxLabelWidth = labelText.length * fontSize;
```
onursumer: For simplicity using an approximate width, instead of calculating the actual (computed) width after rendering.

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -86,27 +83,11 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
         }))
     }, {});
 
-    private lollipopLabel(mutationsAtPosition:Mutation[]):string {
```
onursumer: moved to a utility class

src/shared/lib/LollipopPlotUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+export function lollipopLabelText(mutationsAtPosition:Mutation[], size?: number): string {
+    const mutationCountsByProteinChange = countMutationsByProteinChange(mutationsAtPosition);
+
+    // only pick specified number of protein change values
+    const proteinChanges = mutationCountsByProteinChange
+        .map(m => m.proteinChange)
+        .slice(0, size && size > 0 ? size : undefined);
+
+    // sort alphabetically (to make it easier to find longest common starting substring)
+    const proteinChangesSorted = proteinChanges.slice(0).sort();
```
alisman: didn't you already call slice above?

src/shared/lib/LollipopPlotUtils.ts
```diff
@@ -0,0 +1,60 @@
...
+    let label = startStr + proteinChangesTrimmed.join("/");
+
+    if (proteinChanges.length < mutationCountsByProteinChange.length) {
+        label = `${label} and ${mutationCountsByProteinChange.length - proteinChanges.length} more`;
+    }
+
+    return label;
+}
+
+export function lollipopLabelTextAnchor(labelText: string,
```
alisman: Anchor sounds like it's going to return <a>

src/shared/lib/LollipopPlotUtils.ts
onursumer: this one is to make another copy to store the sorted array without mutating `proteinChanges`. so `slice()` here is only used for the purpose of cloning, may not be the best way to do it though

PR #1538, opened 2018-10-09
oncoprint - use max for MUTATION_COUNT attribute instead of average

adamabeshouse:


src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -352,21 +352,27 @@ export function fillClinicalTrackDatum(
...
+            if (values.length === 0) {
                 fillNoDataValue(trackDatum, attribute);
             } else {
-                // average
-                trackDatum.attr_val = numValSum / numValCount;
-                trackDatum.attr_val_counts[trackDatum.attr_val] = 1;
+                switch (attribute.clinicalAttributeId) {
+                    case "MUTATION_COUNT":
+                        // max
+                        trackDatum.attr_val = values.reduce((max, nextVal)=>Math.max(max, nextVal), Number.NEGATIVE_INFINITY);
```
alisman: is this different from just Math.max(values) ?

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -352,21 +352,27 @@ export function fillClinicalTrackDatum(
...
+                        break;
+                    default:
+                        // average
+                        trackDatum.attr_val = values.reduce((sum, nextVal)=>sum+nextVal, 0) / values.length;
```
alisman: _.mean(values) ?

src/shared/components/oncoprint/DataUtils.ts
adamabeshouse: was not aware of this

src/shared/components/oncoprint/DataUtils.ts
adamabeshouse: didn't know about that

PR #1541, opened 2018-10-09
Remove study view level caching

zhx828:
We have API level caching capacity now, study view level caching is no longer needed;
This solves issues
https://github.com/cBioPortal/cbioportal/issues/4827
https://github.com/cBioPortal/cbioportal/issues/4915
https://github.com/cBioPortal/cbioportal/issues/4966


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -310,7 +313,7 @@ export class StudyViewPageStore {
         };
     }
 
-    private clinicalDataCache:ClinicalDataCache = new ClinicalDataCache()
+    private clinicalDataPromises:{[id:string]:MobxPromise<any>} = {};
```
alisman: i would think the typing on that is pretty important.  should just be like, ClinicalAttribute[] ??

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1317,90 +1422,25 @@ export class StudyViewPageStore {
...
                 return 0;
             }).forEach(attribute => {
-                const uniqueKey = getClinicalAttributeUniqueKey(attribute);
                 const priority = Number(attribute.priority);
-                if (priority === 0){
-                    this.changeChartVisibility(uniqueKey, false);
-                }else if (attribute.patientAttribute) {
-                    if (patientAttributeCount < 10 || priority > 1) {
+                if (attribute.patientAttribute) {
+                    if (patientAttributeCount < 5 || priority > 1) {
```
alisman: maybe make these constants?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1317,90 +1422,25 @@ export class StudyViewPageStore {
...
+        default: [],
         onResult: (result) => {
-            _.forEach(result, (obj, uniqueKey) => {
-                if (obj.length < 2) {
-                    this.changeChartVisibility(uniqueKey, false);
-                } else {
-                    this.changeChartVisibility(uniqueKey, true);
-                }
+            _.each(result, (attribute) => {
+                this.changeChartVisibility(getClinicalAttributeUniqueKey(attribute), true);
```
alisman: gonna guess that changeChartVisibility is an "action" that updates observable.  If so, we will get as many redundant render cycles as items in result.   you should put the loop inside a method tagged with action to avoid this.

src/pages/studyView/StudyViewPageStore.ts
zhx828: This is a mix proxy, but definitely can be updated

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -810,9 +827,114 @@ export class StudyViewPageStore {
...
+        });
+
+        const patientSamplesMap = _.groupBy(samples, sample => sample.uniquePatientKey);
+
+        _.forEach(patientClinicalData, item => {
+            (patientSamplesMap[item.uniquePatientKey] || []).forEach(sample => {
+                clinicalData[sample.uniqueSampleKey] = { ...(clinicalData[sample.uniqueSampleKey] || {}), ['PATIENT_' + item.clinicalAttributeId]: item.value };
+            });
+        });
+        return _.reduce(samples, (acc, sample) => {
```
kalletlak: why not just return `clinicalData`

zhx828: Good point, probably missed when moving the function. Done.

PR #1547, opened 2018-10-09
Various study view fixes

zhx828:
# What? Why?
Fix cBioPortal/cbioportal#4959 cBioPortal/cbioportal#4958 cBioPortal/cbioportal#4964 cBioPortal/cbioportal#4956 cBioPortal/cbioportal#4955

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -500,7 +500,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                         </div>
                     </MSKTab>
 
-                    {/*<MSKTab key={6} id="mutationalSignatureTab" linkText="Mutational Signature Data" hide={true}>*/}
+                    {/*<MSKTab key={6} id="mutationalSignatures" linkText="Mutational Signature Data" hide={true}>*/}
                         {/*<div className="clearfix">*/}
                             {/*<FeatureTitle title="Mutational Signatures" isLoading={ patientViewPageStore.clinicalDataGroupedBySample.isPending } className="pull-left" />*/}
                             {/*<LoadingIndicator isLoading={patientViewPageStore.mutationalSignatureData.isPending}/>*/}
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -243,10 +243,10 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                 )
                             }
                         </span>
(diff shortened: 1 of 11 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -192,6 +192,7 @@ export class PatientViewPageStore {
 
     readonly hasMutationalSignatureData = remoteData({
         invoke: async() => false,
```
kalletlak: seems odd. its `false` all the time

zhx828: Yes, we are temporary disable the feature until the backend is ready

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -243,10 +243,10 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                 )
                             }
                         </span>
-                        {patientViewPageStore.hasMutationalSignatureData.isComplete &&
+                        {patientViewPageStore.hasMutationalSignatureData.result &&
```
alisman: why change from isComplete to result?

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -30,6 +30,16 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
         super(props);
     }
 
+    static readonly defaultProps = {
+        width: 300
+    };
+
+    @computed
+    get columnWidth() {
+        let lastTwoColumns = [80, 60];
```
alisman: could you abstract this to a method named add80Plus60() and then write a few tests for it?  ; )

src/pages/patientView/PatientViewPage.tsx
zhx828: The info should only be shown when the result is true. Not when isComplete

src/pages/studyView/table/ClinicalTable.tsx
zhx828: [freaking out]

src/pages/patientView/PatientViewPage.tsx
alisman: I see, so it's boolean.

alisman: maybe add ===true just to make it super clear.  otherwise, it seems like we're using it as a proxy for isComplete

zhx828: Updated

PR #1551, opened 2018-10-10
Authentication: page should NOT reload after user logs in

alisman:
When user logs in, we need to call configuration service to see if login has established session and get username etc.   Previously we had been reloading page to update user state but we can do this without reloading page

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -1,0 +1,23 @@
+import {computed, observable} from "mobx";
+import {initializeAPIClients} from "./config/config";
+import * as _ from 'lodash';
+
+export class AppStore {
+
+    @observable userName:string | undefined;
+
+    @observable authMethod:string | undefined;
+
+    @computed get isLoggedIn(){
+        return _.isString(this.userName) && this.userName !== "anonymousUser"
+    }
+
+    @computed get logoutUrl(){
+        if (this.authMethod === "saml") {
+            return "/saml/logout";
+        } else {
+            return "j_spring_security_logout"
+        }
+    }
+
+}
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -30,6 +30,7 @@ import {setServerConfig, updateConfig} from "../../config/config";
 import {embedGoogleAnalytics} from "../../shared/lib/tracking";
 import {computed} from "mobx";
 import { If, Else } from 'react-if';
+import {AppStore} from "../../AppStore";
 
 interface IContainerProps {
     location: Location;
@@ -47,6 +48,10 @@ export default class Container extends React.Component<IContainerProps, {}> {
 
     private get routingStore(){
         return getBrowserWindow().routingStore;
+    }
+
+    private get appStore(){
+        return getBrowserWindow().globalStores.appStore;
     }
 
     renderChildren() {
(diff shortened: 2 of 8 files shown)


src/appShell/App/Container.tsx
```diff
@@ -49,6 +50,10 @@ export default class Container extends React.Component<IContainerProps, {}> {
         return getBrowserWindow().routingStore;
     }
 
+    private get appStore(){
```
alisman: @adamabeshouse at some point i would like to use <Provider> to allow child components access to these global stores without resorting to window object.  But i just don't quite understand how it works with typing yet.  Not too important

src/config/config.ts
```diff
@@ -192,3 +192,18 @@ export function initializeConfiguration(){
     }
 
 }
+
+export function fetchServerConfig(){
```
alisman: we're using "config" a little loosely now.  The config service returns not just configuration data but user's authentication state.   Maybe we should at least split it into nodes.  @inodb what do you think?

src/AppStore.ts
```diff
@@ -0,0 +1,23 @@
+import {computed, observable} from "mobx";
+import {initializeAPIClients} from "./config/config";
+import * as _ from 'lodash';
+
+export class AppStore {
```
alisman: We need an "AppStore" because configuration object is not observable and probably shouldn't be as it would not be expected to update I don't think.

src/config/config.ts
```diff
@@ -192,3 +192,18 @@ export function initializeConfiguration(){
...
+
+    return $.ajax({
```
alisman: i'm using jquery ajax because superagent doesn't supply jsonp support out of box (seems crazy) . you have to install a plugin.

src/AppStore.ts
inodb: should we call it `AuthStore` to reflect it's relation to authentication, or is there a plan to extend its purpose?

src/appShell/App/PortalHeader.tsx
```diff
@@ -102,31 +105,25 @@ export default class PortalHeader extends React.Component<{}, {}> {
...
                     </ul>
                 </nav>
             </div>
 
             <div id="rightHeaderContent">
-                <If condition={!_.isEmpty(AppConfig.authUserName)}>
+                <If condition={this.props.appStore.isLoggedIn}>
                     <Then>
-                        <div className="identity">Logged in as <span dangerouslySetInnerHTML={{__html:AppConfig.authUserName!}}></span>
+                        <div className="identity">Logged in as {this.props.appStore.userName}
```
adamabeshouse: what on earth was going on with the way it used to be and dangerouslySetInnerHTML

src/config/config.ts
inodb: yeah might be nice to move the auth out, but fine for now imo

src/config/IAppConfig.ts
```diff
@@ -28,6 +28,8 @@ export type VirtualCohort = {
 export interface IServerConfig {
 
     "app_version": string|null;   // default: "1.0"
+    "authenticationMethod": string | undefined;
```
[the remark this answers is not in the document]
alisman: yeah, this has to do with mixing of "state" stuff and prop names from portal.properties.

src/appShell/App/PortalHeader.tsx
alisman: not sure.  i'm hoping that there would never be any HTML in there.  but who the heck knows!

PR #1555, opened 2018-10-11
Upgrade to node LTS 8.12.0 / npm 6.4.1

inodb:
This allows us to use `npm ci` (https://docs.npmjs.com/cli/ci), which doesn't change package.json or
package-lock.json and fails when installation can't be completed with the
specified dependencies. Also ran `npm audit` which upgraded a few dependencies.
Furthermore upgrade `node-sass`, because the binary from the old version is gone.

**UPDATE**
Needed to update the screenshots. There is some shifting of the fonts and the scrollbar looks a bit different, because we are using a new circleci image with prolly a slightly different chrome version.

- [x] works on jitpack
- [x] works on heroku


.travis.yml
```diff
@@ -1,85 +0,0 @@
-language: node_js
```
inodb: not using travis anymore

pom.xml
```diff
@@ -36,7 +36,7 @@
                         </goals>
                         <!-- Optional configuration which provides for running any npm command -->
                         <configuration>
-                            <arguments>install --production</arguments>
+                            <arguments>ci --only=production</arguments>
```
inodb: use npm ci for jitpack as well (jitpack is actually for java, which is why we are using a pom file)

package.json
```diff
@@ -139,12 +139,12 @@
     "mobx-react-router": "^3.1.2",
     "mobx-utils": "^2.0.1",
     "mobxpromise": "^1.3.1",
-    "node-sass": "^4.3.0",
+    "node-sass": "^4.9.3",
```
inodb: this upgrade was necessary because the binary was missing for 4.3.0

pieterlukasse:
Nice! Thanks @inodb :+1:

PR #1560, opened 2018-10-11
mutations tab - use just one big loader

adamabeshouse:


src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -50,6 +50,13 @@ export default class ResultsViewMutationMapper extends MutationMapper<IResultsVi
         }
     }
 
+    protected get isMutationTableDataLoading() {
+        return (this.props.store.clinicalDataForSamples &&
```
alisman: lets use my new function getMobxPromiseGroupStatus.  Also, why check for existence of promise?

adamabeshouse: good point, I dont know

PR #1563, opened 2018-10-15
Reference the promise only if the Added Chart is clicked

zhx828:

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4991

### DIFF
--- a/src/pages/studyView/SummaryHeader.tsx
+++ b/src/pages/studyView/SummaryHeader.tsx
@@ -1,32 +1,37 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import {
-    Sample,
     ClinicalDataIntervalFilterValue,
-    SampleIdentifier,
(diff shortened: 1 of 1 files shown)


src/pages/studyView/SummaryHeader.tsx
```diff
@@ -137,21 +183,48 @@ export default class SummaryHeader extends React.Component<ISummaryHeaderProps,
...
+                    freq: formatFrequency(freq)
+                };
+                if (sampleCount === 0) {
+                    newOption.disabled = true;
+                }
+                options.push(newOption);
+                return options;
+            }, [] as { label: string, value: string, freq: number, disabled?: boolean }[]);
+        } else {
+            options = _.reduce(Object.keys(this.props.attributesMetaSet), (options, key: string) => {
```
kalletlak: why not hide options until `this.cachedClinicalAttributesWithCount` is set

PR #1564, opened 2018-10-15
studyIds needs to come from original query (cannot derive from sample…

alisman:
…s) b/c a multi study query may actually have samples from only a single study

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -53,9 +53,13 @@ function initStore() {
             // normalize cancer_study_list this handles legacy sessions/urls where queries with single study had different param name
             const cancer_study_list = query.cancer_study_list || query.cancer_study_id;
 
+            const cancerStudyIds = cancer_study_list.split(",");
```
alisman: @zhx828 @kalletlak  we also have cohortIds parameter and this maybe the same as cancer study list?  When we submit a virtual study, does the cancer_study_list param contain all physical studies or just virtual study?  and what about cohortIdsList?

kalletlak: @alisman it contains virtual study.

adamabeshouse: anyone else think this is getting confusing and needs some documentation

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1123,40 +1132,96 @@ export class ResultsViewPageStore {
...
+               const allSampleLists = await Promise.all(this.samplesSpecification.map((spec)=>{
+                   return client.getAllSampleListsInStudyUsingGET({
+                       studyId: spec.studyId,
+                       projection: 'SUMMARY'
+                   })
+               }));
+
+               const specs = allSampleLists.reduce((aggregator:SamplesSpecificationElement[], sampleLists)=>{
+                   //find the sample list matching the selected category using the map from shortname to full category name :(
+                   const matchingList = _.find(sampleLists,(list)=>list.category===SampleListCategoryMap[this.sampleListCategory!]);
```
adamabeshouse: pull `SampleListCategoryMap[this.sampleListCategory!]` out of the loops since it doesnt change

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1123,40 +1132,96 @@ export class ResultsViewPageStore {
...
+
+           // is this a sample list category query?
+           // if YES, we need to derive the sample lists by:
+           // 1. looking up all sample lists in selected studies
+           // 2. using those with matching category
+           if (!this.sampleListCategory) {
+               return this.samplesSpecification
+           } else {
+               const allSampleLists = await Promise.all(this.samplesSpecification.map((spec)=>{
+                   return client.getAllSampleListsInStudyUsingGET({
```
adamabeshouse: your loop on line 1208+ could end up taking quite a long time this way if theres a long custom case list. instead you could collect the unique study ids then query for those

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1175,20 +1240,23 @@ export class ResultsViewPageStore {
     }, {});
 
     readonly studyIds = remoteData({
-        await: ()=>[this.studyToSampleIds],
         invoke: ()=>{
-            return Promise.resolve(Object.keys(this.studyToSampleIds.result));
+            return Promise.resolve(this.cohortIdsList.slice());
```
adamabeshouse: what's the point of `studyIds` anymore?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1123,40 +1132,96 @@ export class ResultsViewPageStore {
...
+                if (next.sampleListId) {
+                    map[next.studyId] = next.sampleListId;
+                }
+                return map;
+            }, {} as {[studyId: string]: string});
+        }
+    })
+
+
+    readonly _sampleSpecification = remoteData({
```
adamabeshouse: missing an `s` at the end of `sample`. Also I think this one should be without underscore, the current member should be with an underscore in my opinion, since this is the one that wraps and sanitizes it.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1123,40 +1132,96 @@ export class ResultsViewPageStore {
...
 
-    @computed get studyToSampleListId(): { [studyId: string]: string } {
-        return this.samplesSpecification.reduce((map, next) => {
-            if (next.sampleListId) {
-                map[next.studyId] = next.sampleListId;
-            }
-            return map;
-        }, {} as {[studyId: string]: string});
-    }
+    // @computed get studyToSampleListId(): { [studyId: string]: string } {
```
adamabeshouse: delete comment

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -118,6 +118,12 @@ export const DataTypeConstants = {
     LOG2VALUE:"LOG2-VALUE"
 };
 
+export const SampleListCategoryMap = {
+    w_mut:"all_cases_with_mutation_data",
```
adamabeshouse: w_mut, w_cna, w_mut_cna would be great to be values of a string-valued enum

```
enum SampleListCategoryType [
    w_mut="w_mut",
   etc
}

export const SampleListCategoryToSampleListId = {
    [SampleListCategoryType.w_mut]:"all_cases_with_mutation_data",
    etc
}

@observable sampleListCategory:SampleListCategoryType | undefined;
```

src/pages/resultsView/ResultsViewPageStore.ts
alisman: it's true but we know that if there's a category at play, then there can only ever be as many spec elements as studies and they will be unique

adamabeshouse: ah okay, fair enough. maybe add a comment about this?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -118,6 +118,18 @@ export const DataTypeConstants = {
     LOG2VALUE:"LOG2-VALUE"
 };
 
+enum SampleListCategoryType {
+    "w_mut"="w_mut",
+    "w_cna"="w_cna",
+    "w_mut_cna"="w_cna_mut"
+}
+
+export const SampleListCategoryMap = {
```
adamabeshouse: I still think this should have a better name. `Map` is so nondescript

PR #1566, opened 2018-10-16
Hide all charts with 0 priority

zhx828:


zhx828:
@inodb this is the study view configuration in frontendConfig. Equivalent to what we use in current configs.json. Let's update those portals once the pr is merged.

PR #1567, opened 2018-10-16
Handle virtual studies in results view

alisman:
# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -29,7 +29,10 @@ import Helmet from "react-helmet";
 import {createQueryStore} from "../home/HomePage";
 import {ServerConfigHelpers} from "../../config/config";
 import {showCustomTab} from "../../shared/lib/customTabs";
-import { buildResultsViewPageTitle} from "./ResultsViewPageStoreUtils";
+import {buildResultsViewPageTitle} from "./ResultsViewPageStoreUtils";
+import {
+    updateStoreFromQuery
+} from "./ResultsViewPageHelpers";
 
 function initStore() {
 
@@ -39,12 +42,22 @@ function initStore() {
         getBrowserWindow().currentQueryStore = createQueryStore();
     }
 
-    const reaction1 = reaction(
+    let lastQuery:any;
+
+    const queryReactionDisposer = reaction(
         () => {
             return getBrowserWindow().globalStores.routing.query
         },
         (query) => {
 
+            // escape from this if queryies are deeply equal
+            // TODO: see if we can figure out why query is getting changed and
+            // if there's any way to do shallow equality check to avoid this expensive operation
+            if (_.isEqual(lastQuery, query)) {
+                return;
+            } else {
+                lastQuery = query;
+            }
 
             if (!getBrowserWindow().globalStores.routing.location.pathname.includes("/results")) {
                return;
@@ -53,12 +66,9 @@ function initStore() {
             // normalize cancer_study_list this handles legacy sessions/urls where queries with single study had different param name
             const cancer_study_list = query.cancer_study_list || query.cancer_study_id;
 
-            const cancerStudyIds = cancer_study_list.split(",");
+            const cancerStudyIds: string[] = cancer_study_list.split(",");
 
             const oql = decodeURIComponent(query.gene_list);
-
-
-
 
             let samplesSpecification: SamplesSpecificationElement[];
 
@@ -89,7 +99,7 @@ function initStore() {
                             sampleId: undefined
                         };
                     });
-            } else {
+            } else if (query.case_set_id === "all") { // case_set_id IS equal to all
                 samplesSpecification = cancerStudyIds.map((studyId:string)=>{
                     return {
                         studyId,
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -0,0 +1,141 @@
+import * as _ from "lodash";
+import {getMolecularProfiles} from "./ResultsViewPageStoreUtils";
+import {runInAction} from "mobx";
+import {ResultsViewPageStore, SamplesSpecificationElement} from "./ResultsViewPageStore";
+import client from "../../shared/api/cbioportalClientInstance";
+import sessionServiceClient from "../../shared/api/sessionServiceInstance";
+import {CancerStudy} from "../../shared/api/generated/CBioPortalAPI";
+import {VirtualStudy} from "../../shared/model/VirtualStudy";
+
+export function updateStoreFromQuery(resultsViewPageStore:ResultsViewPageStore, query:any,
```
alisman: will add some tests soon

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -89,94 +105,54 @@ function initStore() {
...
+                // fetch summary projection because it's likely to be already used (an in cache) or will be used later
+                runInAction(()=>{
+                    updateStoreFromQuery(resultsViewPageStore, query, samplesSpecification, physicalStudies, oql, cancerStudyIds);
+                });
 
-                if (!_.isEqual(query.RPPA_SCORE_THRESHOLD, resultsViewPageStore.rppaScoreThreshold)) {
-                    resultsViewPageStore.rppaScoreThreshold = parseFloat(query.RPPA_SCORE_THRESHOLD);
-                }
+                // now update state so view can render result
+                resultsViewPageStore.checkingVirtualStudies = false;
```
alisman: @adamabeshouse i hate this pattern here but am not sure yet how else to do this.

alisman: Basically, we when we detect virtual studies in query, we need to fetch them and fixup the query specification.  The reaction makes this kind of awkward.    Maybe if instead of reaction we were getting "query" as a property, there would a more idiomatic way to to do this.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -39,12 +48,22 @@ function initStore() {
...
+    const queryReactionDisposer = reaction(
         () => {
             return getBrowserWindow().globalStores.routing.query
         },
         (query) => {
 
+            // escape from this if queryies are deeply equal
+            // TODO: see if we can figure out why query is getting changed and
+            // if there's any way to do shallow equality check to avoid this expensive operation
+            if (_.isEqual(lastQuery, query)) {
```
adamabeshouse: you can use ExtendedRouterStore::queryHash :)

but I just realized that if you're having any weirdness with the insertion order of the query object, then you may have to update that method to use some package that does a "stable stringify" eg https://www.npmjs.com/package/json-stable-stringify

zhx828:
http://www.cbioportal.org/beta/results?Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=all&gene_list=BRAF&geneset_list=%20&tab_index=tab_visualize&Action=Submit&cancer_study_list=5bc657f5498eb8b3d5685b33 says there is no BRAF mutated samples. But if you look at the study, there are 121 sample having BRAF mutations. http://www.cbioportal.org/beta/newstudy?id=5bc657f5498eb8b3d5685b33&filters={%22mutatedGenes%22:[{%22entrezGeneIds%22:[673]}]} 
 cc @jjgao

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -89,94 +105,54 @@ function initStore() {
                             sampleId: undefined
                         };
                     });
-            } else {
+            } else if (query.case_set_id === "all") { // case_set_id IS equal to all
```
adamabeshouse: I feel like this comment must signify something deeper than just restating the condition, could you use that specific instead?

src/pages/resultsView/ResultsViewPage.tsx
adamabeshouse: what about making the fetch into a mobxpromise and using .isPending? you could still reference the result (and the status) in the reaction

src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -0,0 +1,141 @@
...
+            } else {
+                resolve([]);
+            }
+        });
+    });
+    return prom;
+
+}
+
+export function substitutePhysicalStudiesForVirtualStudies(cancerStudyIds:string[], virtualStudies:VirtualStudy[]){
```
adamabeshouse: just so Im clear on terminology - what constitutes a virtual study? is a multiple-study query a virtual study? or is a virtual study something the user has *saved*. and is there any other kind of study besides physical and virtual?

zhx828:
When open URL includes VS created by others, the page tries to fetch `http://www.cbioportal.org/beta/api/sample-lists/5bc657f5498eb8b3d5685b33_all/sample-ids` but it should not. It causes the page not rendering anything. http://www.cbioportal.org/beta/results?Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&case_set_id=all&gene_list=BRAF&geneset_list=%20&tab_index=tab_visualize&Action=Submit&cancer_study_list=5bc657f5498eb8b3d5685b33%2Cmsk_impact_2017

src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    let physicalStudies:string[] = [];
+
+    //if a study is a virtual study, substitute its physical study ids
+    const virtualStudiesKeyedById = _.keyBy(virtualStudies,(virtualStudy)=>virtualStudy.id);
+    cancerStudyIds.forEach((studyId)=>{
+        if (studyId in virtualStudiesKeyedById) {
+            const virtualStudy = virtualStudiesKeyedById[studyId];
+            physicalStudies = physicalStudies.concat(virtualStudy.data.origin);
```
zhx828: The origin could be a VS as well

zhx828: Instead of recursively fetching VS info,  we can use all the keys in virtualStudy.data.studies

zhx828: @kalletlak  correct me if I'm wrong

src/pages/resultsView/ResultsViewPage.tsx
alisman: good point

src/pages/resultsView/ResultsViewPageHelpers.ts
alisman: A virtual study is one that has been saved using the vs service.   It has  a study ID that looks like a guid and we use the vs service to retrieve it's properties.  Right now, it's just a set of studies and lists of samples in those studies.

src/pages/resultsView/ResultsViewPageHelpers.ts
kalletlak: yes, origin could be VS.

src/pages/resultsView/ResultsViewPage.tsx
alisman: I don't think insertion order matters because we are doing deep comparison of object.  nothing stringified

src/pages/resultsView/ResultsViewPageHelpers.ts
adamabeshouse: got it thanks

PR #1568, opened 2018-10-16
Move all study view configurations to AppConfig

zhx828:
This solves issue https://github.com/cBioPortal/cbioportal/issues/4778 and https://github.com/cBioPortal/cbioportal/issues/5016

### DIFF
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -25,6 +25,11 @@ export type VirtualCohort = {
     constituentStudyIds:string[]
 };
 
+export type StudyView = {
+    tableAttrs: string[]
+    priority: { [id: string]: number }
+}
+
 export interface IServerConfig {
 
     "app_version": string|null;   // default: "1.0"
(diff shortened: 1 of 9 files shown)


zhx828:
@inodb this is the study view configuration in frontendConfig. Equivalent to what we use in current configs.json. Let's update those portals once the pr is merged.

zhx828:
@inodb yeah it works as is. As long as we add study_view in the json file, it will overwrite the front-end default. No backend changes needed

alisman:
This seems like a whole lot of configuration.  Is it realistic that someone would want to change this on a instance basis?   Another issue is that with deep nested config structures, you can't configure just one property and fallback on defaults for everything else.  We could improve this.  We can merge this and then discuss later.  Food for thought.

zhx828:
@alisman good point. Let's just keep whatever available in the current configs.json. Let's revisit before the release.

zhx828:
@alisman  @kalletlak updated. In this round, let the backend supports table `tableAttrs` and `priority`. @inodb could you show me where you update the frontendConfig.json? I can update for the genie portal.

PR #1569, opened 2018-10-16
Hide tabs based on data availability

adamabeshouse:


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -355,7 +379,11 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
             {
                 id:"network",
                 hide:()=>{
-                    return this.resultsViewPageStore.studies.result!.length > 1;
+                    if (!this.resultsViewPageStore.studies.isComplete) {
```
alisman: maybe reverse sign so that we ONLY evaluate result if it's complete (not error)

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -330,7 +346,8 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
             {
                 id:"survival",
                 hide:()=>{
-                    return this.resultsViewPageStore.studies.result!.length > 1;
+                    return !this.resultsViewPageStore.survivalClinicalData.isComplete ||
+                        this.resultsViewPageStore.survivalClinicalData.result.length === 0;
```
ersinciftci: Won't this load the whole survival data? I thought we were going to call the `META` projection and check the `total-count` header so that the data itself is still lazy loaded when the tab is clicked.

PR #1578, opened 2018-10-18
Fix tests

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5028

### DIFF
--- a/end-to-end-tests/specs/home.spec.js
+++ b/end-to-end-tests/specs/home.spec.js
@@ -640,7 +640,7 @@ describe('oncoprint', function() {
             };
         }
 
-        it.skip("should sort patients and samples by custom case list order correctly", ()=>{
+        it("should sort patients and samples by custom case list order correctly", ()=>{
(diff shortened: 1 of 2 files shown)


end-to-end-tests/specs/home.spec.js
```diff
@@ -654,13 +654,13 @@ describe('oncoprint', function() {
             checkBox.waitForExist(10000);
             browser.click('[data-test="StudySelect"] input');
 
-            browser.setValue(inputSelector, '');
+            for (let i = 0; i < browser.getValue(inputSelector).length; i++) { browser.setValue(inputSelector, "\uE003").pause(100); };
```
alisman: interesting!  thank you for fixing this test.  how did you know about this trick?  maybe we need to standarize this

kalletlak: figured out while I was testing locally. somehow `setValue()` isn't working.
found this solution at here https://github.com/webdriverio/webdriverio/issues/1140

PR #1587, opened 2018-10-19
E2E tests remove redundant viewport size

inodb:
We are setting this in goToUrlAnSetLocalStorage, so need to call multiple times

### DIFF
--- a/end-to-end-tests/specs/screenshot.spec.js
+++ b/end-to-end-tests/specs/screenshot.spec.js
@@ -246,7 +246,6 @@ describe('study view screenshot test', function(){
         mutatedGenesHeader.waitForExist(30000);
 
         // give charts time to render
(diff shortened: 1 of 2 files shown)


end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -218,7 +218,6 @@ describe('patient view page screenshot test', function(){
     before(function(){
         var url = `${CBIOPORTAL_URL}/patient?studyId=lgg_ucsf_2014&caseId=P04`;
         goToUrlAndSetLocalStorage(url);
-        browser.setViewportSize({ height: 1600, width: 1600 });
```
adamabeshouse: we need to leave this one because otherwise the genomic overview doesnt expand correctly

PR #1591, opened 2018-10-19
fix heroku

inodb:


src/config/config.ts
```diff
@@ -177,14 +177,17 @@ export function initializeConfiguration() {
...
+    const frontendUrl = config.frontendUrl || `//${win.location.host}/`;
 
-    const configServiceUrl = config.configurationServiceUrl || APIROOT;
+    const configServiceUrl = config.configurationServiceUrl || `${APIROOT}config_service.jsp`;
 
+    // should override both when in dev mode and when serving compiled source
+    // code outside of legacy project
     // @ts-ignore: ENV_* are defined in webpack.config.js
-    if (IS_DEV_MODE) {
+    if (IS_DEV_MODE || !(config.frontendUrl)) {
```
inodb: this could prolly only check `config.frontendUrl` or anything else that checks we're not getting things passed from the legacy project

PR #1595, opened 2018-10-22
Fix compile errors

adamabeshouse:


inodb:
only patient view e2e fails - seems to fail intermittently (old viewport issue)

PR #1600, opened 2018-10-23
enrichments tab - handle data loading failures more gracefully

adamabeshouse:
image


alisman:
At some point we should make a component which handles the switch statement.

alisman:
replaced with https://github.com/cBioPortal/cbioportal-frontend/pull/1605

PR #1604, opened 2018-10-23
Dont use oncokb in oncoprint, mutations, patient view if !show_oncokb

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -313,7 +313,7 @@ export class ResultsViewPageStore {
                 this._oncoKb = val;
             },
             get oncoKb() {
-                return this._oncoKb && !store.didOncoKbFailInOncoprint;
+                return AppConfig.serverConfig.show_oncokb && this._oncoKb && !store.didOncoKbFailInOncoprint;
```
alisman: what that didOncoKbFail thing?

jjgao:
@inodb @onursumer are we using genome nexus for hotspots annotation for some part? Should we switch to GN for all?

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: checking if oncokb promise has failed, but I cant use isError because of the way we did it with returning an Error object as result so that the whole app doesnt fail via error result causing all the dependent promises to error.

adamabeshouse:
seems like screenshots failing because oncokb is not turned on in circleCI? or maybe hotspots is not? how do we set the default config

PR #1625, opened 2018-10-25
Make batch API calls to reduce number of calls

adamabeshouse:


ersinciftci:
500 error should be fixed now.

PR #1636, opened 2018-10-26
enrichments tooltip - undo log transformation to show pvalue

adamabeshouse:


src/pages/resultsView/enrichments/MiniScatterChart.tsx
```diff
@@ -114,7 +114,7 @@ export default class MiniScatterChart extends React.Component<IMiniScatterChartP
                         positionTop={this.tooltipModel.y - 33}>
                         Gene: {this.tooltipModel.datum.hugoGeneSymbol}<br/>
                         Log Ratio: {formatLogOddsRatio(this.tooltipModel.datum.logRatio)}<br/>
-                        p-Value: {toConditionalPrecision(this.tooltipModel.datum.y, 3, 0.01)}<br/>
+                        p-Value: {toConditionalPrecision(Math.pow(10, -this.tooltipModel.datum.y), 3, 0.01)}<br/>
```
ersinciftci: Would be cleaner if you modify `getAlterationScatterData` and `getExpressionScatterData` to add a pValue field to the data, and use that value in here, just like it's done for qValue. Because if we change the -log10 formula to something else in the future, this line will probably be forgotten, and would produce the wrong value.

adamabeshouse: good point

PR #1639, opened 2018-10-26
oncoprint - change heatmap UI to make buttons more clear

adamabeshouse:
we need to work on the UI more


adamabeshouse:
these changes arent adequate, just going to close this and we'll revisit it later

PR #1654, opened 2018-10-30
Fix pfam domains undef

inodb:
Fix cBioPortal/cbioportal#5143

The problem is that allTranscripts sometimes did not have the canonicalTranscript. This is b/c they are accessing different tables in the db

It might be good to improve user feedback when not all the required data is there, e.g. some msg indicating that there is no transcript for the given hugo symbol. This should not happen, but it's possible b/c the hugo symbol checking in the query selector differs from the canonical transcript assignment in genome nexus

I guess a cleaner solution would be to fix the genome nexus data

### DIFF
--- a/src/shared/components/mutationMapper/MutationMapperStore.ts
+++ b/src/shared/components/mutationMapper/MutationMapperStore.ts
@@ -174,7 +174,7 @@ export default class MutationMapperStore
         }
     }, undefined);
 
-    readonly allTranscripts = remoteData<EnsemblTranscript[] | undefined>({
+    readonly transcriptsByHugoSymbol = remoteData<EnsemblTranscript[] | undefined>({
         invoke: async()=>{
             if (this.gene) {
                 return fetchEnsemblTranscriptsByEnsemblFilter({"hugoSymbols":[this.gene.hugoGeneSymbol]});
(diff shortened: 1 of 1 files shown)


src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -203,6 +203,27 @@ export default class MutationMapperStore
...
+            this.canonicalTranscript
+        ],
+        invoke: async()=> {
+            let transcripts: EnsemblTranscript[] = [];
+            if (this.transcriptsByHugoSymbol.result) {
+                transcripts = this.transcriptsByHugoSymbol.result;
+            }
+            // add canonical transcript if it wasn't in response of transcriptsByHugoSymbol
+            if (this.canonicalTranscript.result && !_.some(transcripts, (et => et.transcriptId === this.canonicalTranscript.result!.transcriptId))) {
+                transcripts = transcripts.concat(this.canonicalTranscript.result);
```
alisman: this kind of logic makes me want a test.  but i think you could maybe compress all 15 lines of this to one line: _.unionBy(this.transcriptsByHugoSymbol.result,[this.cannonicalTrancript.result],(t)=>t.transcriptId)

inodb: @alisman updated as suggested

PR #1655, opened 2018-10-30
Fix geneset bugs in SPA

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -654,6 +655,23 @@ export class ResultsViewPageStore {
                             ret.push(profile);
                         }
                     }));
+                } else if (profile.molecularAlterationType === AlterationTypeConstants.GENESET_SCORE) {
+                    // geneset profile, we dont have the META projection for geneset data, so just add it
```
alisman: kill comment?

adamabeshouse: no I think leave because we'll want to change that if we ever add that projection

PR #1657, opened 2018-10-30
New Study View Bug fixes

zhx828:
cBioPortal/cbioportal#5146
cBioPortal/cbioportal#5002
cBioPortal/cbioportal#5019

### DIFF
--- a/src/pages/studyView/charts/barChart/BarChart.tsx
+++ b/src/pages/studyView/charts/barChart/BarChart.tsx
@@ -111,20 +111,27 @@ export default class BarChart extends React.Component<IBarChartProps, {}> implem
         return values;
     }
 
-    @computed get tickFormat() : string[] {
+    @computed get tickFormat() {
         // copy non-numerical categories as is
         return [
             ...this.numericalTickFormat,
             ...this.categories
-        ] as string[];
+        ];
     }
 
     @computed get tilted() {
         return this.tickValues.length > STUDY_VIEW_CONFIG.thresholds.escapeTick;
     }
 
-    @computed get bottomPadding() :number {
-        return this.tilted ? adjustedLongestLabelLength(this.tickFormat) * 7 : 20;
+    @computed
+    get bottomPadding(): number {
+        return this.tilted ? adjustedLongestLabelLength(this.tickFormat.map((tick: string | string[]) => {
+            if (_.isArray(tick)) {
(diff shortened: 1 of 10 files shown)


src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -110,12 +111,20 @@ export default class BarChart extends React.Component<IBarChartProps, {}> implem
         return values;
     }
 
-    @computed get tickFormat() {
+    @computed get tickFormat() : string[] {
```
onursumer: This function does not always return an array of strings, its actual type is `(string|string[])[]`. See https://github.com/cBioPortal/cbioportal-frontend/blob/d635d95a416725a98f1ef1f628fe2c29a55ea88f/src/pages/studyView/charts/barChart/BarChart.tsx#L86

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -110,12 +111,20 @@ export default class BarChart extends React.Component<IBarChartProps, {}> implem
...
         // copy non-numerical categories as is
         return [
             ...this.numericalTickFormat,
             ...this.categories
-        ];
+        ] as string[];
```
onursumer: This is not a safe cast, since the function returns `(string|string[])[]`

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -118,6 +119,21 @@ export default class BarChart extends React.Component<IBarChartProps, {}> implem
...
 
+    @computed get tilted() {
+        return this.tickValues.length > STUDY_VIEW_CONFIG.thresholds.escapeTick;
+    }
+
+    @computed
+    get bottomPadding(): number {
+        return this.tilted ? adjustedLongestLabelLength(this.tickFormat.map((tick: string | string[]) => {
+            if (_.isArray(tick)) {
+                return tick.join('^');
```
onursumer: Maybe join with an empty string? We never actually show the character `^` in the label. For example, 10^-2 is displayed as 10<sup>-2</sup>. So, I think `10-2` is a better approximation than `10^-2` when calculating the label length.

zhx828: Oh ok, let me do that

PR #1664, opened 2018-10-31
Add max POST cache size with FIFO clearing

adamabeshouse:


src/shared/lib/proxyPost.ts
```diff
@@ -19,6 +20,39 @@ const postCacheStore:any = {};
 
 getBrowserWindow().postCacheStore = postCacheStore;
 
+// Cache clearing for memory management
+const sizeQueue:{methodName:string, storeKey:number, size:number}[] = [];
+const maxCacheSize = 1000000000; // set to Number.POSITIVE_INFINITY to disable cache clearing
```
adamabeshouse: @alisman major question is what to set this limit to

src/shared/lib/proxyPost.ts
```diff
@@ -30,9 +64,27 @@ export function proxyPost(targetObj:any, methodName:string){
     targetObj[methodName] = function(arg:any){
         //console.log('posted data',arg);
         const hash = getHash(arg);
+        // remove this from sizeQueue, if it exists, because we just used it so it should get bumped back in line,
+        //  away from deletion
+        const sizeQueueIndex = sizeQueue.findIndex(o=>((o.methodName === methodName) && (o.storeKey === hash)));
+        if (sizeQueueIndex > -1) {
+            sizeQueue.splice(sizeQueueIndex, 1);
+        }
+        // add new element to size queue- push it here so that its place in line depends on when it was called, not
```
adamabeshouse: just realized I might as well reuse the element I just spliced

src/shared/lib/proxyPost.ts
adamabeshouse: discussed on slack - decided to try to clear it in response to catching out of memory error

I think we can try to clear it half at a time

adamabeshouse: as discussed on slack - this is not possible bc out of memory is a fatal error - theres no memory left to do any more javascript - cache clearing, GC, etc. so we have to set some reasonable max limit

adamabeshouse: all studies w 30 genes crashes bc result from server is too large for a single allocation - we get "allocation size error"

all studies with 3 genes succeeds, w total memory usage of 424 MB

adamabeshouse: as discussed on slack, theres no way to determine how much memory is available and make the decision based on that. we also dont want to make the limit lower than a big study would cause.

the decision for now is to just not cache sufficiently large results - let them linger in the mobxpromise they called from, and naturally get GC'd when that mobxpromise no longer in use. I'll determine the threshold

alisman:
See #1712

PR #1665, opened 2018-10-31
Use isInitialLoading computed to differentiate the initial load or user interaction

zhx828:
This solves 
https://github.com/cBioPortal/cbioportal/issues/5119
https://github.com/cBioPortal/cbioportal/issues/4969

@onursumer could you take a look the bar chart part?
@kalletlak this is the solution after discussing with @alisman

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -279,7 +279,7 @@ export class StudyViewPageStore {
         reaction(()=>this.filters, ()=>this.clearAnalysisGroupsSettings()); // whenever any data filters change, reset survival analysis settings
     }
 
-    @observable private defaultFilters: Partial<StudyViewFilter> = {};
+    @observable private initialFilters: Partial<StudyViewFilter> = {};
 
     @observable studyIds: string[] = [];
 
@@ -310,13 +310,13 @@ export class StudyViewPageStore {
 
     @action
     updateStoreFromURL(query: Partial<StudyViewURLQuery>) {
-        let studyIdsString;
+        let studyIdsString:string = '';
         let studyIds: string[] = [];
         if (query.studyId) {
-            studyIdsString = (query.studyId as string);
+            studyIdsString = query.studyId;
         }
         if (query.id) {
-            studyIdsString = (query.id as string);
+            studyIdsString = query.id;
         }
         if (studyIdsString) {
             studyIds = studyIdsString.trim().split(",");
@@ -387,12 +387,12 @@ export class StudyViewPageStore {
                 return acc;
             }, [] as CopyNumberGeneFilter[]);
         }
-        this.defaultFilters = filters;
+        this.initialFilters = filters;
     }
 
     @computed
     get isInitialFilterState(): boolean {
-        return _.isEqual(toJS(this.defaultFilters), toJS(this.filters));
+        return _.isEqual(toJS(this.initialFilters), toJS(this.filters));
     }
 
     @computed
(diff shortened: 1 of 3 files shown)


zhx828:
This PR needs to be updated with once the instance is updated after merging https://github.com/cBioPortal/cbioportal/pull/5164

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -372,6 +375,13 @@ export class StudyViewPageStore {
                 return acc;
             }, [] as CopyNumberGeneFilter[]);
         }
+        this.defaultFilters = filters;
+    }
+
+    @computed
+    get isInitialLoading(): boolean {
+        // TODO: figure out why isEqual cannot be applied to these two observables
+        return _.isEqual(JSON.parse(JSON.stringify(this.defaultFilters)), JSON.parse(JSON.stringify(this.filters)));
```
alisman: it's probably because stuff mobx adds to make arrays and objects observable.   look at mobx "toJS"  https://mobx.js.org/refguide/tojson.html
will accomplish the same as what you are doing above but in more sane fashion

alisman: should we rename defaultFilters = initialFilters 
and filter = currentFilters?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -372,6 +375,13 @@ export class StudyViewPageStore {
                 return acc;
             }, [] as CopyNumberGeneFilter[]);
         }
+        this.defaultFilters = filters;
+    }
+
+    @computed
+    get isInitialLoading(): boolean {
```
alisman: i would name this more clearly:   isInitialFilterState

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1665,16 +1650,87 @@ export class StudyViewPageStore {
...
+                        return {
+                            attributeId: attr.clinicalAttributeId,
+                            clinicalDataType: attr.patientAttribute ? 'PATIENT' : 'SAMPLE'
+                        } as ClinicalDataFilter
+                    }),
+                    studyViewFilter: this.defaultFilters
+                } as ClinicalDataCountFilter
+            });
+        },
+        onResult: (data) => {
```
alisman: lets evaluate whether we really need these onResult handlers, which are kind of an anti-pattern

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1665,16 +1650,87 @@ export class StudyViewPageStore {
...
+            this.initialClinicalDataBinCountCharts(data);
+        },
+        default: []
     });
 
+    @action
+    initialClinicalDataCountCharts(data: ClinicalDataCountItem[]) {
```
alisman: i think all action methods should have verbs in name.  like, initializeClinicalDataCountCharts

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -996,24 +1006,27 @@ export class StudyViewPageStore {
         let uniqueKey:string = getClinicalAttributeUniqueKey(chartMeta.clinicalAttribute!);
```
alisman: we should see if we can avoid having to call this method to populate the clinicalDataCountPromises.  clinicalDataCountPromises seems like it should be computed.     if we must then method should be named something like addChartData

src/pages/studyView/StudyViewPageStore.ts
zhx828: I tried toJS yesterday. It's surprisingly not working here

src/pages/studyView/StudyViewPageStore.ts
zhx828: I actually did. Most of unnecessary has been removed or grouped together. The ones left are relying on different endpoints. And I think this is the proper place to update some states.

src/pages/studyView/StudyViewPageStore.ts
zhx828: This is a good point. I agree a computed map seems a fair solution. Let me reevaluate.

zhx828:
Seems toJS does not work well for computed. https://github.com/mobxjs/mobx/issues/1532

src/pages/studyView/StudyViewPageStore.ts
zhx828: I changed the defaultFilters to initialFilters. Not the filters yet. It will create too many conflicts with my other pr. I will do a cleanup in a separate pr. Will follow up when working on this issue https://github.com/cBioPortal/cbioportal/issues/5197

PR #1667, opened 2018-11-01
Compact oncoprint

adamabeshouse:


jjgao:
@adamabeshouse could you add a screenshot here?

adamabeshouse:
@jjgao 
image

image


image

jjgao:
@alisman @adamabeshouse please prioritize this since it also fixes some issue.

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -39,24 +73,31 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
...
-        return <div>
+    readonly singleStudyUI = remoteData({
+        await:()=>[
+            this.props.store.queriedStudies,
+            this.props.store.sampleLists,
+            this.props.store.samples,
+            this.props.store.patients,
+            this.props.store.genes
+        ],
+        invoke:()=>Promise.resolve(<div>
```
alisman: oh boy.  JSX returned from invoke of remoteData?   I don't think I like this.   I think remoteData should return data

adamabeshouse: why not? this way the caller only has to do myJSXElement.isComplete which is much more direct

alisman: I would like to see a component which accepts mobxPromise[] and renders either success, pending or callback UI depending on status.   kid of extension of aggregate thing i did a while back

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -77,8 +119,37 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
...
+
+    readonly alterationSummary = remoteData({
+        await:()=>[this.props.store.samples, this.props.store.patients,
+            this.props.store.alteredSampleKeys, this.props.store.alteredPatientKeys],
+        invoke:()=>{
+            const numSamples = this.props.store.samples.result.length;
+            const numPatients = this.props.store.patients.result.length;
+            const numAlteredSamples = this.props.store.alteredSampleKeys.result!.length;
+            const numAlteredPatients = this.props.store.alteredPatientKeys.result!.length;
+            const sampleSummary = `${numSamples} (${getPercentage(numAlteredSamples/numSamples, 0)}) of queried samples`;
```
alisman: it think this should all be in pure functions (construction of the string)

src/pages/resultsView/querySummary/QuerySummary.tsx
adamabeshouse: but then the caller still has to know about all its dependencies and do more work. this enables encapsulating each thing with its dependencies so the caller doesnt need to know as much. I dont get the problem

src/pages/resultsView/querySummary/QuerySummary.tsx
adamabeshouse: I dont see the point here, bc its only one line and only called once

adamabeshouse: sure :+1:

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -154,6 +138,11 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                 this.selectedClinicalAttributeIds.set(SpecialAttribute.StudyOfOrigin, true);
             }
         });
+        onMobxPromise([props.store.samples, props.store.patients], (samples:any[], patients:any[])=>{
```
alisman: do these get disposed?    i had the idea that component methods could be decorated as reactions.   That would make it possible to dispose them on destruction of component.  it looks like there is some support for this in mobx, but prob not in our version 
https://github.com/mobxjs/mobx-react/issues/181

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -1045,6 +987,20 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         }
     }
 
+    @bind
```
alisman: autobind?

adamabeshouse: for some reason I thought we were using bind

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: good point - if they dont resolve then they're not disposed

adamabeshouse: Ill log an issue and lets come back to it

PR #1674, opened 2018-11-02
Dont hide tabs that are routed to

adamabeshouse:


src/config/config.ts
```diff
@@ -128,11 +126,6 @@ export class ServerConfigHelpers {
         }
     }
 
-    @memoize static parseDisabledTabs(str:string){
```
adamabeshouse: I was having a lot of weird issues having to do w/ circular dependency or some kind of dependency timing issue, having to do with config. So I had to move this function out of config once it referenced the ResultsViewTab enum

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -156,21 +157,6 @@ export interface IResultsViewPageProps {
     params: any; // from react router
 }
 
-export enum ResultsViewTab {
```
adamabeshouse: I moved this out in order to solve circular dependency issues

src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -8,15 +8,54 @@ import {CancerStudy} from "../../shared/api/generated/CBioPortalAPI";
...
+    NETWORK="network",
+    EXPRESSION="expression",
+    DOWNLOAD="download"
+}
+
 export function getTabId(pathname:string) {
     const match = pathname.match(/results\/([^\/]+)/);
     if (match) {
-        return match[1];
+        return match[1] as ResultsViewTab;
```
alisman: seems like it might be safer to just return pathname itself?   also, just fyi, i like the regexp test method b/c it return boolean.   /results\/.test(pathname)

adamabeshouse: pathname is like `/results/plots`, this method extracts 'plots' from that

PR #1677, opened 2018-11-03
Move selected samples and sequence/cna profiled samples to study summary

zhx828:
This solves https://github.com/cBioPortal/cbioportal/issues/5069

screen shot 2018-11-09 at 5 28 46 pm
screen shot 2018-11-09 at 5 29 04 pm


jjgao:
@zhx828 maybe we can use something like the buttons below?

image

@alisman any styling magic we can do here?

kalletlak:
@zhx828 
- [ ] custom case selection box is missing

zhx828:
@kalletlak we will add the custom case selection box in another way. Will send a pr later

zhx828:
@alisman @kalletlak this pr is now read to be reviewed.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -47,16 +42,34 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
     private handleTabChange(id: string) {
-        this.props.routing.updateRoute({ tab: id });
+        this.props.routing.updateRoute({tab: id});
     }
 
-    content(){
+    @computed
+    get numberOfSamplesWithMutation() {
+        if (this.store.molecularProfileSampleCounts.isComplete &&
+            this.store.molecularProfileSampleCounts.result.numberOfMutationProfiledSamples) {
```
kalletlak: any reason for checking `numberOfMutationProfiledSamples`

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -1,27 +1,23 @@
...
-import { StudySummaryRecord } from 'pages/studyView/virtualStudy/VirtualStudy';
 import {PageLayout} from "../../shared/components/PageLayout/PageLayout";
 import IFrameLoader from "../../shared/components/iframeLoader/IFrameLoader";
-import { StudySummaryTab } from 'pages/studyView/tabs/SummaryTab';
-import {StudyViewFilter} from "../../shared/api/generated/CBioPortalAPIInternal";
+import {StudySummaryTab} from 'pages/studyView/tabs/SummaryTab';
+import StudyPageHeader from "./studyPageHeader/StudyPageHeader";
+import "./styles.scss";
+import styles from './styles.module.scss';
+import SelectedInfo from "./SelectedInfo/SelectedInfo";
```
kalletlak: may be combining both the style files

zhx828: We cannot yet. There is a global style class, this is the only way to overwrite some styles.

src/pages/studyView/StudyViewPage.tsx
zhx828: The endpoint does not always have numberOfMutationProfiledSamples

src/pages/studyView/studyPageHeader/rightPanel/RightPanel.tsx
```diff
@@ -0,0 +1,201 @@
...
+        this.props.store.getDownloadDataPromise().then(text => {
+            this.downloadingData = false;
+            fileDownload(text, 'data.tsv');
+        }).catch(() => {
+            this.downloadingData = false;
+            this.showDownloadErrorMessage = true;
+        });
+    }
+
+    @bind
```
alisman: lets use autobind instead of bind

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -47,16 +42,34 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
+    get numberOfSamplesWithMutation() {
+        if (this.store.molecularProfileSampleCounts.isComplete &&
+            this.store.molecularProfileSampleCounts.result.numberOfMutationProfiledSamples) {
+            return this.store.molecularProfileSampleCounts.result.numberOfMutationProfiledSamples;
+        } else {
+            return 0;
+        }
+    }
 
+    @computed
```
alisman: You should avoid referencing mobx promises inside computed functions.  The problem is that they hide load status.  Instead we should accept the results of the promises as arguments to a pure function and handle loading status outside.   For example, this function will return 0 inaccurately.  It's not 0, it just hasn't loaded yet.   I know this is a pain in the ass but it's just not a good practice.

zhx828: Makes sense, will update

PR #1681, opened 2018-11-05
Hide virtual studies section in download data tab in query page

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5188


src/shared/components/query/QueryAndDownloadTabs.tsx
```diff
@@ -32,6 +32,7 @@ export default class QueryAndDownloadTabs extends React.Component<IQueryAndDownl
 	onSelectTab = (eventKey:string) =>
 	{
 		this.store.forDownloadTab = eventKey === DOWNLOAD;
+		this.store.selectableSelectedStudyIds = [];
```
alisman: won't this undo selection when you move between tabs?  maybe that desirable?

kalletlak: it shows `x studies selected (xxx samples)` in header if there are any virtual studies selected from query tab

PR #1683, opened 2018-11-06
Change footer to app store

dippindots:
Put fetch of version info inside of mobxpromise in AppStore and adjust footer view to use it.
Fix #5180
@alisman

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -1,6 +1,9 @@
 import {computed, observable} from "mobx";
+import {remoteData} from "shared/api/remoteData";
 import {initializeAPIClients} from "./config/config";
 import * as _ from 'lodash';
+import internalClient from "shared/api/cbioportalInternalClientInstance";
+
 
 export class AppStore {
 
@@ -20,4 +23,13 @@ export class AppStore {
         }
     }
 
+    readonly portalVersion = remoteData<string | undefined>({
+        invoke:async()=>{
+            const portalVersionResult = await internalClient.getInfoUsingGET({});
+            if (portalVersionResult && portalVersionResult.portalVersion) {
+                return Promise.resolve("v" + portalVersionResult.portalVersion.split('-')[0]);
+            }
+            return undefined; 
+        }
+    });
 }
--- a/src/appShell/App/PortalFooter.tsx
+++ b/src/appShell/App/PortalFooter.tsx
@@ -9,35 +9,25 @@ import {
     default as CBioPortalAPIInternal
 } from "shared/api/generated/CBioPortalAPIInternal";
 import internalClient from "shared/api/cbioportalInternalClientInstance";
+import {AppStore} from "../../AppStore";
+import getBrowserWindow from "../../shared/lib/getBrowserWindow";
+import {observer} from "mobx-react";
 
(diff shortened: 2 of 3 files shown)


src/AppStore.ts
```diff
@@ -1,13 +1,18 @@
...
+import internalClient from "shared/api/cbioportalInternalClientInstance";
+
 
 export class AppStore {
 
     @observable userName:string | undefined;
 
     @observable authMethod:string | undefined;
 
+    @observable portalVersion:string | undefined;
```
alisman: we don't need this portalVersion property

src/AppStore.ts
```diff
@@ -20,4 +25,13 @@ export class AppStore {
         }
     }
 
+    readonly getPortalVersion = remoteData<string>({
```
alisman: instead of calling this "getPortalVersion" lets call it just portalVersion ( since it's not a method)

src/AppStore.ts
```diff
@@ -20,4 +25,13 @@ export class AppStore {
...
+        invoke:async()=>{
+            const portalVersionResult = await internalClient.getInfoUsingGET({});
+            if (portalVersionResult && portalVersionResult.portalVersion) {
+                return Promise.resolve("v" + portalVersionResult.portalVersion.split('-')[0]);
+            }
+            return "";
```
alisman: lets return undefined instead of empty string

src/AppStore.ts
```diff
@@ -25,13 +23,13 @@ export class AppStore {
         }
     }
 
-    readonly getPortalVersion = remoteData<string>({
+    readonly portalVersion = remoteData<string | undefined>({
```
alisman: nice work!

alisman:
for some there are 861 files getting in this commit, almost all of which are not changed.  probably something your IDE is doing.  please have a look

alisman:
Also, you'll want to swash your commits to one.  In general we swash commits unless there's a good logical reason to keep changes discrete in separate commits.

src/appShell/App/PortalFooter.tsx
```diff
@@ -33,11 +35,15 @@ export default class PortalFooter extends React.Component<{}, IPortalFooterState
     }
 
     render() {
+            var version = "--";
```
alisman: One more thing: if you look at PortalHeader and Container.tsx, you'll see how  appStore is passed in as a property.  This is better than having global references everywhere so lets adopt that here too

src/appShell/App/PortalFooter.tsx
```diff
@@ -33,11 +27,18 @@ export default class PortalFooter extends React.Component<{}, IPortalFooterState
     }
 
     render() {
+            var version;
+            if (this.props.appStore.portalVersion.isComplete && !this.props.appStore.portalVersion.isError && getBrowserWindow().globalStores.appStore.portalVersion.result){
```
alisman: @dippindots one more thing i just noticed: you're still using the global reference to appStore in last item here.

dippindots: Sorry about that, I just fixed it.

PR #1686, opened 2018-11-06
Use new router store in oncoprint

adamabeshouse:


src/pages/resultsView/ResultsViewQuery.ts
```diff
@@ -0,0 +1,118 @@
+import {SamplesSpecificationElement, SampleListCategoryType} from "./ResultsViewPageStore";
+import {computed, observable} from "mobx";
+import {CancerStudyQueryUrlParams} from "../../shared/components/query/QueryStore";
+import {getMolecularProfiles} from "./ResultsViewPageStoreUtils";
+import hashString from "../../shared/lib/hashString";
+import _ from "lodash";
+
+export class ResultsViewQuery {
+    // TODO: make sure i am initializing these the exact same way as they were in resultsviewpagestore to avoid any weird bugs w assuming undefined at start
+    @observable.ref public samplesSpecification:SamplesSpecificationElement[] = [];
```
alisman: I'm not sure these have to be observable because this is only a transfer object.   Maybe all properties should be readonly?

PR #1697, opened 2018-11-09
Tooltip for alteration % in oncoprint tracks

adamabeshouse:
image

image

image

### DIFF
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "node-sass": "^4.9.3",
     "numeral": "^2.0.6",
     "object-sizeof": "^1.2.0",
-    "oncoprintjs": "^1.1.11",
+    "oncoprintjs": "^1.1.12",
     "parameter-validator": "^1.0.2",
     "pdfobject": "^2.0.201604172",
     "plotly.js": "^1.41.3",
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1063,28 +1063,26 @@ export class ResultsViewPageStore {
     readonly sequencedSampleKeys = remoteData<string[]>({
         await:()=>[
             this.samples,
-            this.coverageInformation
+            this.coverageInformation,
+            this.selectedMolecularProfiles
         ],
         invoke:()=>{
             const genePanelInformation = this.coverageInformation.result!;
-            return Promise.resolve(this.samples.result!.map(s=>s.uniqueSampleKey).filter(k=>{
-                const sequencedInfo = genePanelInformation.samples[k];
-                return !!sequencedInfo.allGenes.length || !!Object.keys(sequencedInfo.byGene).length;
-            }));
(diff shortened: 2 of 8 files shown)


jjgao:
@adamabeshouse maybe change "altered/sequenced" to "altered/profiled"?

package.json
```diff
@@ -147,7 +147,7 @@
     "mobxpromise": "github:cbioportal/mobxpromise#v1.0.2",
     "node-sass": "^4.9.3",
     "numeral": "^2.0.6",
-    "oncoprintjs": "^1.1.9",
+    "oncoprintjs": "github:cbioportal/oncoprintjs#61c9f7196262082635ed16f51abc5798bc313edb",
```
adamabeshouse: @alisman can we do this?

alisman: I do believe this works based in a little googlery.  Try it! If it works ... it works!!!

adamabeshouse: it does seem to work... i was just wondering if you had had issues w it with jitpack or something bc i remember there was some reason you preferred to use tags

alisman: most of the examples seem to use tags. probably more kosher to use tags but not important to me.

pieterlukasse: I would only use tags if you can guarantee they are reliable. E.g. 1.1.10 should be backwards compatible with 1.1.9 since it is a minor release. If this is not the goal, then what you propose above should be fine.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1019,6 +1023,34 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly areSequencedSamplesDifferentPerGene = remoteData<boolean>({
```
pieterlukasse: why is this check needed? Why not let the tooltip always appear?

adamabeshouse: I thought to only put that, along with the asterisk, if it's needed. What do you think? Always tooltip? Always asterisk? Always both?

jjgao: I think we can always display the tooltip.

adamabeshouse: what about the asterisk?

jjgao: I think asterisk should only be there when there is `not profiled` samples in the row.

adamabeshouse: regardless of whether the not profiled samples are different in each track, i.e. the denominator is the same in each track

jjgao: Yes, I think so. It's a useful feature to indicate there are `not profiled` samples in the row (or all rows) since they are at the end of the oncoprint.

adamabeshouse: kk, that simplifies things

PR #1700, opened 2018-11-09
Mutationstab: change to banner alert

dippindots:
Change to alert to the uniform banner alert.
Fix cBioPortal/cbioportal/#5183


src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -51,6 +53,12 @@ export default class Mutations extends React.Component<IMutationsPageProps, {}>
                         onToggle={this.onToggleOql}
                     />
                 </div>
+                
+                {this.props.store.mutationMapperStores.isComplete && ! this.props.store.mutationMapperStores.result[this.mutationsGeneTab].dataStore.showingAllData &&
+                    <div>
```
alisman: you probably don't need extra div around this

alisman: also, try putting this above inside the tabMessageContainer div.

dippindots: Thanks, I just fixed it.

PR #1702, opened 2018-11-12
DataAccessTokens download token details on auth. portals

angelicaochoa:
# What? Why?
Added "Download token" link which generates a new token for a user on an authenticated portal. If user is on portal with authentication method "social_auth" then this link is suppressed.

* Users on authenticated portals can now download API access tokens from username dropdown on portal home page. 
* Added DAT endpoint calls to `CBioPortalAPI.ts`
* Added new react component for DataAccessTokens
* Moved sign out link to dropdown
* Add DAT file contents build and download link
* Suppress DAT download if auth method is social_auth

**Dropdown for user on authenticated portal:**
screen shot 2018-11-12 at 12 18 58 pm

**Dropdown for user on portal with authentication method = "social_auth"**
dat-dropdown-social_auth

[Sample DAT download file](https://github.com/cBioPortal/cbioportal-frontend/files/2577354/cbioportal_data_access_token.7.txt)

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -207,3 +207,7 @@ export function getDocsUrl(sourceUrl:string,docsBaseUrl?:string): string {
         return docsBaseUrl + "/" + sourceUrl;
     }
 }
+
+export function getLegacyDataAccessTokenUrl(){
+    return buildCBioPortalAPIUrl("api-legacy/dataAccessToken");
+}
--- a/src/shared/components/dataAccessTokens/DataAccessTokens.tsx
+++ b/src/shared/components/dataAccessTokens/DataAccessTokens.tsx
@@ -1,0 +1,109 @@
+import * as React from "react";
+import AppConfig from "appConfig";
+import {If, Then, Else} from 'react-if';
+import client from "shared/api/cbioportalClientInstance";
+import {AppStore} from "../../../AppStore";
+import LoadingIndicator from "../loadingIndicator/LoadingIndicator";
+import {observer} from "mobx-react";
+import {observable} from "mobx";
+import {buildCBioPortalPageUrl} from "../../api/urls";
+import {isNullOrUndefined} from "util";
+import fileDownload from 'react-file-download';
+
+export class UserDataAccessToken {
+    @observable token : string;
+    @observable creationDate: string;
+    @observable expirationDate: string;
+    @observable username: string;
+    constructor(token: string, creationDate: string, expirationDate: string, username: string) {
+        this.token = token;
+        this.creationDate = creationDate;
+        this.expirationDate = expirationDate;
+        this.username = username;
+    }
+}
+
+export interface IDataAccessTokensProps {
+    token?:string;
+    creationDate?:string;
+    expirationDate?:string;
+    loadingComponent?:JSX.Element;
+    appStore:AppStore;
+}
+
+function buildDataAccessTokenFileContents(dat:UserDataAccessToken | undefined) {
+    if (!isNullOrUndefined(dat)) {
+        const fileContents = "token: " + dat!.token + "\n" +
+            "creation_date: " + dat!.creationDate + "\n" +
+            "expiration_date: " + dat!.expirationDate;
+        return fileContents;
+    } else {
+        alert("Cannot create Data Access Token file for user with non-existent tokens.");
+        return null;
+    }
+}
+
+@observer
+export class DataAccessTokens extends React.Component<IDataAccessTokensProps, {}> {
+    public static defaultProps: Partial<IDataAccessTokensProps> = {
+        loadingComponent:<LoadingIndicator isLoading={true}/>
+    };
+
+    constructor(props: IDataAccessTokensProps) {
+        super(props);
+    }
+
+    getDatDropdownList() {
+        const listItems = [
+            {
+                id:"signout",
+                action:<a href={buildCBioPortalPageUrl(this.props.appStore.logoutUrl)}>Sign out</a>,
+                hide:false
+            },
+            {
+                id:"datDownload",
+                action:<a onClick={() => this.downloadDataAccessTokenFile()}>Download token</a>,
+                hide:AppConfig.serverConfig.authenticationMethod === "social_auth"
+            }
+        ];
+
+        const shownListItems = listItems.filter((l)=>{
+            return !l.hide;
+        });
+
+        return shownListItems.map((l)=>{
+            return <li>{l.action}</li>;
+        });
+    }
+
+    async downloadDataAccessTokenFile() {
+        const dat = await this.generateNewDataAccessToken();
+        if (!isNullOrUndefined(dat)) {
+            const fileContents = buildDataAccessTokenFileContents(dat);
+            fileDownload(fileContents, "cbioportal_data_access_token.txt");
+        }
+    }
+
+    async generateNewDataAccessToken() {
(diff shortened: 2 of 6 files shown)


sheridancbio:
I've marked this DO NOT MERGE for now because we have not completed/merged the backend API development feature branch to RC yet. We could merge this into RC ahead of time if we added code to detect whether or not the backend has token support or not (whether the DAT endpoints are present/functional) ... if they are non-functional then then front end could suppress the menu option "Download Token" and *only* show the "Sign out" option. Maybe we want that behavior anyway. Maybe for non-authenticated portals, the GET /dataAccessTokens endpoint would return a status code such as "403 Forbidden" to indicate that tokens only work on an authenticated portal. And this reminds me that another frontend feature is needed ... some users are prohibited from receiving tokens even if they have authenticated. If such a user logs in, it would be best not to even show them the "Download Token" option in the menu. The backend also could (maybe does) return a 403 Forbidden status to the GET request above ... if the frontend receives such a status, it should hide (or at least inactivate) the menu choice.

src/shared/components/dataAccessTokens/DataAccessTokens.tsx
```diff
@@ -0,0 +1,102 @@
...
+            userDataAccessToken: new UserDataAccessToken('','','','')
+        } as IDataAccessTokensState;
+    }
+
+     async generateNewDataAccessToken() {
+        if (this.props.appStore.isLoggedIn) {
+            let _token = await Promise.resolve(
+                client.createDataAccessTokenUsingPOST(
+                    {'allowRevocationOfOtherTokens':AppConfig.serverConfig.dat_revoke_other_tokens}))
+            const dat = new UserDataAccessToken(_token.token, _token.creation, _token.expiration, _token.username);
```
sheridancbio: I'm not able to work out in my mind whether this code will be executed before the user clicks the "Download Token" menu choice. It is important that the frontend *not* call the API and generate the token before the user chooses the menu choice, because every time a token is generated, the previous token will be automatically revoked and this would revoke a user's current token just for signing in to the portal. Also, it would be undesirable to generate a token for every user who logs into the portal -- most likely very few users will ever request a token. So I think it is important for us to test whether an existing token for a user gets revoked when that user simply logs in to the portal again with a different browser (i.e. with no cookie and where they authenticate)
My guess is that when the component (DataAccessTokens.tsx) becomes part of the DOM, it will call generateNewDataAccessToken(), and that will instantiate UserDataAccessToken with the *resolved* promise that yields _token. So I think the token will be generated when the DOM element is realized ... perhaps immediately ... or perhaps it becomes part of the DOM after triggering the dropdown menu? (So when testing revocation of prior tokens in an incognito window, maybe we should drop the menu down but not choose "Download Token" ... and then see if the revocation occurred)

src/shared/api/urls.ts
```diff
@@ -178,3 +178,7 @@ export function getDocsUrl(sourceUrl:string,docsBaseUrl?:string): string {
         return docsBaseUrl + "/" + sourceUrl;
     }
 }
+
+export function getLegacyDataAccessTokenUrl(){
```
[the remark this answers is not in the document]
angelicaochoa: @sheridancbio yes, once we move it to `api/` then the domain + path can be referenced the same way as the other endpoints are being referenced here 👍

angelicaochoa:
@sheridancbio you're right. It will generate a new token whenever someone signs in. I intended on fetching a user's token but during dev/testing I just wanted to create a token for a user. I'll implement that change soon

angelicaochoa:
@sheridancbio I've updated the PR to create a new token on the Download Token link click instead of on render of the drop down list component

angelicaochoa:
@sheridancbio @jjgao I changed the base branch to `dat-feature-branch` since this feature will require more development alongside development going into the `cbioportal/feature-data-access-tokens` branch as well

angelicaochoa:
@alisman Would you mind reviewing as well? This PR will be going into a development branch for the Data Access Token UI

angelicaochoa:
@sheridancbio re your comment 

> I've made a couple of suggestions already, and here I only make one real request ... that we test whether a previously issued token for a user will be revoked automatically when the user logs in with a different browser / incognito window and then drops down the menu with the "Download Token" option without choosing it (such as when Signing out). I suspect it will, but I am not sure.

With this current implementation and how we (our portal instance) will always pass `allowRevocationOfOtherTokens = true`, the user will be assigned a new token when they click the download token link.

src/shared/components/dataAccessTokens/DataAccessTokens.tsx
```diff
@@ -0,0 +1,109 @@
...
+        const listItems = [
+            {
+                id:"signout",
+                action:<a href={buildCBioPortalPageUrl(this.props.appStore.logoutUrl)}>Sign out</a>,
+                hide:false
+            },
+            {
+                id:"datDownload",
+                action:<a onClick={() => this.downloadDataAccessTokenFile()}>Download token</a>,
+                hide:AppConfig.serverConfig.authenticationMethod === "social_auth"
```
sheridancbio: This covers our main use cases. For future enhancement, the backend API for DAT could be probed to see whether the individual user has permission to use data access tokens on that portal. Then the menu item could be disabled/hidden for individual users who are prohibited from using tokens. I see that you have dropped the Store for holding state for the user's tokens, which makes sense. But we may want to revive that soon to hold user state information ... such as whether they have permissions to use data access tokens, and maybe other user-specific state such as available sessions they have saved through the session-service.

src/shared/components/dataAccessTokens/DataAccessTokens.tsx
```diff
@@ -0,0 +1,109 @@
...
+    getDatDropdownList() {
+        const listItems = [
+            {
+                id:"signout",
+                action:<a href={buildCBioPortalPageUrl(this.props.appStore.logoutUrl)}>Sign out</a>,
+                hide:false
+            },
+            {
+                id:"datDownload",
+                action:<a onClick={() => this.downloadDataAccessTokenFile()}>Download token</a>,
```
sheridancbio: Here the click triggers the download directly, which is what we want.

src/shared/components/dataAccessTokens/DataAccessTokens.tsx
sheridancbio: This has been resolved by directly calling the download function in the click handler for the menu item .. rather than in a separate data access token component / observer / store.

src/shared/components/dataAccessTokens/DataAccessTokens.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }
+    }
+
+    async generateNewDataAccessToken() {
+        if (this.props.appStore.isLoggedIn) {
+            let _token = await Promise.resolve(
+                client.createDataAccessTokenUsingPOST(
+                    {'allowRevocationOfOtherTokens':AppConfig.serverConfig.dat_revoke_other_tokens}))
+            const dat = new UserDataAccessToken(_token.token, _token.creation, _token.expiration, _token.username);
+            this.setState({userDataAccessToken : dat})
```
sheridancbio: I'm not sure we still need this call to setState() ... maybe yes because async downloadDataAccessTokenFile() is waiting for the promise to be fulfilled and the setState() triggers the notification?

PR #1713, opened 2018-11-13
Replaced the iframe with igv 2.0 in CNSegments tab

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4670.

Notes:
- ~~This version fetches all the segment data in advance before displaying the `CN Segments` tab. For bigger data sets it takes a while for the `CN Segments` tab to appear. In order to make the initial loading faster we may want to modify the API to take an additional chromosome parameter and initially only query the selected chromosomes (chromosomes for query genes only)~~. Now using the new endpoint to query chromosome specific segment data only for the gene tabs.
- We may need to tweak the IGV browser options to fix performance issues for bigger data sets.
- In addition to the list of genes, there is also an explicit `Whole Genome` option right next to the gene list at the top.

Single gene selection:
cn_segments_single_gene

Whole Genome:
cn_segments_all

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -5,6 +5,7 @@ import {
     ClinicalData,
     ClinicalDataMultiStudyFilter,
     ClinicalDataSingleStudyFilter,
+    CopyNumberSeg,
     Gene,
     GenePanel,
     GenePanelData,
@@ -39,6 +40,7 @@ import PdbHeaderCache from "shared/cache/PdbHeaderCache";
 import {
     cancerTypeForOncoKb,
     fetchCnaOncoKbDataWithNumericGeneMolecularData,
+    fetchCopyNumberSegmentsForSamples,
     fetchGenes,
     fetchGermlineConsentedSamples,
     fetchMyCancerGenomeData,
@@ -635,6 +637,34 @@ export class ResultsViewPageStore {
             return ret;
         }
     });
+
+    readonly cnSegments = remoteData<CopyNumberSeg[]>({
+        await: () => [
+            this.samples
+        ],
+        invoke: () => fetchCopyNumberSegmentsForSamples(this.samples.result)
+    }, []);
+
+    readonly cnSegmentsByChromosome = remoteData<{ [chromosome: string]: MobxPromise<CopyNumberSeg[]> }>({
+        await: () => [
+            this.genes,
+            this.samples
+        ],
+        invoke: () => {
+            if (this.genes.result) {
+                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
+                    map[chromosome] = remoteData<CopyNumberSeg[]> ({
+                        invoke: () => fetchCopyNumberSegmentsForSamples(this.samples.result, chromosome)
+                    });
+
+                    return map;
+                }, {}));
+            } else {
+                return Promise.resolve({});
+            }
+        }
+    }, {});
+
 
     readonly molecularData = remoteData<NumericGeneMolecularData[]>({
         await: () => [
--- a/src/pages/resultsView/cnSegments/CNSegments.tsx
+++ b/src/pages/resultsView/cnSegments/CNSegments.tsx
@@ -1,139 +1,180 @@
 import * as React from 'react';
-import $ from 'jquery';
 import {observer} from "mobx-react";
-import {MSKTabs, MSKTab} from "../../../shared/components/MSKTabs/MSKTabs";
-import {Gene, Sample} from "../../../shared/api/generated/CBioPortalAPI";
-import {unescapeTabDelimited, unquote} from "shared/lib/StringUtils";
+import {action, computed, observable} from "mobx";
+import MobxPromise from "mobxpromise";
+import autobind from "autobind-decorator";
+import {Nav, NavItem} from "react-bootstrap";
+
 import {ResultsViewPageStore} from "../ResultsViewPageStore";
-import {observable} from "mobx";
-import AppConfig from "appConfig";
-import fileDownload from 'react-file-download';
-import {buildCBioPortalPageUrl, getLegacyCopyNumberUrl} from '../../../shared/api/urls';
-
-interface CNSegmentsIframeProps {
-    sampleIds:string[];
-    studyId:string;
-    gene:Gene
-}
-
-class CNASegmentIframe extends React.Component<CNSegmentsIframeProps,{}>{
-    
-    iframeDiv:HTMLDivElement;
-
-    constructor(){
-        super();
-        this.downloadSegmentData = this.downloadSegmentData.bind(this);
-    }
-    
-    private getIframeBody(){
-
-        var segmentTrackHeight = 1.5*this.props.sampleIds.length;
-        var height = 300 + 2*this.props.sampleIds.length;
-        height = Math.min(height, 800);
-        var headerContent = '<head>    <link rel="stylesheet" type="text/css"  href="https://ajax.googleapis.com/ajax/libs/jqueryui/1.11.2/themes/smoothness/jquery-ui.css"/>'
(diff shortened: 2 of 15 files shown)


src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -1,139 +1,114 @@
...
-            return null
+            return <LoadingIndicator isLoading={true} center={true} size="big" />;
         }
-
-
     }
 
+    @autobind
+    private handleDownload() {
+        onMobxPromise(
```
alisman: should we have an alias for onMobxPromise called onMobxPromiseOnce?  This is probably the default use case for it (times=1) and the idea that it disposes after one call is key.  i had to check to make sure that it was disposing and not just firing every time the observed promise resolved.  @adamabeshouse

src/shared/components/igv/IntegrativeGenomicsViewer.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+    private igvDiv: HTMLDivElement|undefined;
+    private igvBrowser: any;
+
+    @observable private locus: string|string[]|undefined;
+
+    constructor(props: IGVProps) {
+        super(props);
+
+        reaction(
```
alisman: this reaction isn't disposed so it will live (and react forever).  we should only use reactions as absolute last resort.  i think we can accomplish the same with react lifecycle hooks.   you could put this in componentDidUpdate.   or make componentSHouldUpdate= false and use componentWillRecieveProps to call igvBrowser.search (no locus observable prop necessary)

src/shared/lib/StoreUtils.ts
```diff
@@ -252,6 +252,24 @@ export async function fetchCopyNumberSegments(studyId:string,
     }
 }
 
+export async function fetchCopyNumberSegmentsForSamples(samples: Sample[],
```
alisman: I'm not sure it make sense to hide the promise inside the utility function in this way.  If we're going to move the fetch to a utility then it should return promise explicitly (typing).   The importance of doing this is 1. if there's logic we really need to test in the handling of the response, 2. If the logic is lengthy and bloats the store.   Neither of those conditions is true here so I don't see the point.

alisman: Also, in the case of #1, it's the sychronous logic of handling the response that we really want to decouple for the purpose of testing.

alisman:
@onursumer will this fix https://github.com/cBioPortal/cbioportal/issues/5162 ?

onursumer:
@alisman there is something wrong with the igv module. I think it is not a true es6 module. I tried different ways of importing it, works fine in development mode, but `npm build` fails with an error message:

```
ERROR in reactapp/0.495168c8aeaf9f0b713d.chunk.js from UglifyJs
Unexpected token: name (igv) [./node_modules/source-map-loader!./node_modules/igv/dist/igv.esm.min.js:1,0][reactapp/0.495168c8aeaf9f0b713d.chunk.js:94150,4]
```

On https://www.npmjs.com/package/igv its example usage is like this: `import igv from 'igv.esm.min.js'`

onursumer:
@alisman yes it will fix cBioPortal/cbioportal#5162. I relocated the download button to the top right though.

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -1,139 +1,114 @@
...
+            if (this.props.store.studies.result.length > 1) {
+                prefix = "multi_study_";
+            }
+            else if (this.props.store.studies.result.length === 1) {
+                prefix = `${this.props.store.studies.result[0].studyId}_`;
+            }
+        }
 
-    render(){
+        return `${prefix}segments.seg`;
```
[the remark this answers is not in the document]
adamabeshouse: this one is fine I think because its kind of just a weak dependence. I don't think I understand what you're saying/how it fits into our larger discussion.

src/pages/resultsView/cnSegments/CNSegments.tsx
adamabeshouse: you mean just for readability? bc the default param value is already 1

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -507,7 +507,7 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                         onMobxPromise(
                             [this.props.store.sampleKeyToSample,
                                 this.props.store.patientKeyToPatient],
-                            (sampleKeyToSample:{[sampleKey:string]:Sample}, patientKeyToPatient)=>{
+                            (sampleKeyToSample:{[sampleKey:string]:Sample}, patientKeyToPatient: any)=>{
```
onursumer: this is to fix typescript errors due to API update

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -529,7 +529,7 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
```
onursumer: this is to fix typescript errors due to API update

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -636,6 +638,34 @@ export class ResultsViewPageStore {
...
+
+    readonly cnSegmentsByChromosome = remoteData<{ [chromosome: string]: MobxPromise<CopyNumberSeg[]> }>({
+        await: () => [
+            this.genes,
+            this.samples
+        ],
+        invoke: () => {
+            if (this.genes.result) {
+                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
+                    map[chromosome] = remoteData<CopyNumberSeg[]> ({
```
alisman: sorry i didn't notice this before, do we have to return a map of remoteDatas?  why not just Promise.all of fetchCopyNumberSegmentsForSamples?

src/shared/components/cnSegments/CNSegmentsDownloader.tsx
```diff
@@ -0,0 +1,74 @@
...
+            </DefaultTooltip>
+        );
+    }
+
+    @autobind
+    private handleDownload() {
+        if (this.downloadFileContent) {
+            fileDownload(this.downloadFileContent, this.props.filename);
+        }
+        else if (this.props.promise.isComplete) {
```
alisman: i think you could just use onMobxPromise once and it would handle case where it promise is resolved or still pending

onursumer: If the promise is already complete I didn't want to set `this.downloading` to true upon download button click, otherwise the `else` block below is already using `onMobxPromise`. I'll try to simplify & remove the additional `else if` block.

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: Actually, this is for the new endpoint where we fetch CN segments by a single chromosome. If we use `Promise.all`, once this promise (`cnSegmentsByChromosome`) is accessed, wouldn't it start fetching all the data for all chromosomes?

alisman: are you accessing the data on a per chromosome basis?

onursumer: Yes, each tab is only fetching a single chromosome (except the whole genome tab, which uses a different promise), and also checking if the data for that specific chromosome is pending/complete. In the initial implementation, we were just fetching the whole genome data and loading it all at once into the IGV browser. Now, loading only partial data for genes.

src/shared/components/cnSegments/CNSegmentsDownloader.tsx
onursumer: Just tried to remove this `else if` block, it has virtually no effect on the loading state of the button. I think it is safe to remove it.

PR #1718, opened 2018-11-14
Only fetch oncogenicity for oncoprint

zhx828:
This is a part of fixes for https://github.com/cBioPortal/cbioportal/issues/5167


zhx828:
Something to notice for this change, since we only fetch the ONCOGENIC, the rest of the data returned from oncokb currently is not reliable except the `query`, `oncogenic` filed. And the data cannot be shared with Mutations tab because it needs all evidences from OncoKB. @onursumer

PR #1719, opened 2018-11-14
Optimization: only query for hotspot genes

inodb:
Currently all mutations are queried to check for hotspots. We might as well
only query those that are on a hotspot gene similar to the way this is done for
oncokb. Since the hotspot data is static I figured we can include this as part
of the source code instead of developing a new API endpoint in Genome Nexus
just for this purpose.


src/shared/lib/CancerHotspotsUtils.ts
```diff
@@ -23,6 +23,13 @@ export async function fetchHotspotsData(mutationData: MobxPromise<Mutation[]>,
     return await client.fetchHotspotAnnotationByGenomicLocationPOST({genomicLocations: genomicLocations});
 }
 
+export function filterMutationsOnNonHotspotGenes(mutationData: Mutation[]) {
+    const hotspotGenes = new Set(require("shared/static-data/hotspotGenes.json"));
```
alisman: support for Set is still iffy and no reason to use it.   it's an array, so just arr.includes(geneSybmol) will work

alisman: you can just as it as string[];

PR #1728, opened 2018-11-16
More density plot fixes

adamabeshouse:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2290,17 +2290,28 @@ export class StudyViewPageStore {
...
                     yAxisAttributeId: MUTATION_COUNT,
                     xAxisStart:0, xAxisEnd:1, // FGA always goes 0 to 1
                     yAxisStart:0, // mutation always starts at 0
+                    xAxisBinCount,
                     yAxisBinCount,
                     clinicalDataType: "SAMPLE",
                     studyViewFilter
                 })).filter(bin=>(bin.count > 0));// only show points for bins with stuff in them
+                const xBinSize = 1/xAxisBinCount;
+                const yBinSize = Math.max(...bins.map(bin=>bin.binY)) / (yAxisBinCount - 1);
```
ersinciftci: I think you need to do `const yBinSize = (Math.max(...bins.map(bin=>bin.binY)) - Math.min(...bins.map(bin=>bin.binY))) / (yAxisBinCount - 1);`, because Y axis doesn't always start from 0.

ersinciftci: Sorry, actually it seems like it starts from zero, didn't see the `yAxisStart=0` parameter, I thought we were only showing the actual range.

PR #1729, opened 2018-11-16
Download Tab Genetic Alteration Table improvements

dippindots:
Generate the new download tab: combine gene and alteration in one column.


src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,58 +39,78 @@ export interface ICaseAlterationTableProps {
...
-        const alterationData = generator.getAlterationData(data);
+    if (data.alterationTypes.length === 0 || !data.alterationTypes.includes(alterationType)){
+        oqlValue = "no alteration";
+    }
+    switch (alterationType){
+        case "MUT":
+            generator = {
+                label: "MUT",
+                getAlterationData: (oqlData: IOqlData) => oqlData.mutation,
+                getProfiledData: (oqlData: IOqlData) => oqlData.isMutationProfiled,
```
onursumer: maybe rename to `isProfiled`?

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,58 +39,78 @@ export interface ICaseAlterationTableProps {
...
 
-        if (alterationData.length > 0) {
-            oqlValue.push(`${generator.label}: ${generator.getValues(alterationData).join(",")};`);
+    if (generator){
+        const alterationData = generator.getAlterationData(data);
+        if (alterationData.length > 0){
+            oqlValue = generator.getValues(alterationData).join(",");
         }
-    });
+        const ProfiledData = generator.getProfiledData(data);
```
onursumer: never start a variable name with an uppercase character. Actually I don't think you even need a variable here. Just do this:

```suggestion
        if (generator.isProfiled(data)) {
           ...
        }
```

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,58 +39,78 @@ export interface ICaseAlterationTableProps {
...
+        if (ProfiledData === false){
```
onursumer: leave a space between `)` and `{`

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -93,13 +119,28 @@ export function generateOqlDisplayValue(oqlData: {[oqlLine: string]: IOqlData},
...
+            oqlValue = datum.sequenced ? generateOqlValue(oqlData[oqlLine], alterationType) : "No alteration";
         }
     }
 
     return oqlValue;
 }
 
+export function findAlterationType(oql: OQLLineFilterOutput<AnnotatedExtendedAlteration>, alterationData: ICaseAlteration[]): string[]
+{
+    let types = new Set<string>();
```
onursumer: use const here instead of let

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -93,13 +119,28 @@ export function generateOqlDisplayValue(oqlData: {[oqlLine: string]: IOqlData},
...
+    alterationData.forEach(element => {
+            for (var key in element.oqlData) {
```
onursumer: never use the keyword `var`

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -93,13 +119,28 @@ export function generateOqlDisplayValue(oqlData: {[oqlLine: string]: IOqlData},
...
+                if (element.oqlData.hasOwnProperty(key)) {
+                    for (var type in element.oqlData[key].alterationTypes){
```
onursumer: never use the keyword `var`, leave a space between `)` and `{`

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -216,6 +220,11 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
                         .map(data => data.oql))
     });
 
+    readonly molecularProfileIdToMolecularProfile = remoteData<{[molecularProfileId:string]:MolecularProfile}>({
```
onursumer: This should probably be implemented inside the store.

src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -22,20 +21,54 @@ export interface IDownloadFileRow {
...
+    let isMutationProfiled = false;
+    let isFusionProfiled = false;
+    let isCnaProfiled = false;
+    let isMrnaExpProfiled = false;
+    let isProteinLevelProfiled = false;
+
+    //record the profile information
+    if (datum.profiled_in)
+    {
+        for (const profile of datum.profiled_in)
```
onursumer: This logic is assuming that if it is not in the `profiled_in` list that it is not profiled. But that may not always be the case. I would instead go over the `not_profiled_in` array to find out the not profiled ones. So maybe also rename your flags to `isMutationNotProfiled`, `isCnaNotProfiled`, etc.

dippindots: @onursumer Thanks Onur, I just updated my commit. Please take a look, appreciated.

onursumer: @dippindots Thanks for the changes, I just found one more thing, can you also update that one?

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -144,16 +184,20 @@ export default class CaseAlterationTable extends React.Component<ICaseAlteration
...
-                sortBy: (data: ICaseAlteration) => generateOqlDisplayValue(data.oqlData, oql.oql_line),
-                filter: (data: ICaseAlteration, filterString: string, filterStringUpper: string) =>
-                    generateOqlDisplayValue(data.oqlData, oql.oql_line).toUpperCase().includes(filterStringUpper)
+            const alterationTypes = findAlterationType(oql, this.props.caseAlterationData);
+            alterationTypes.forEach(alterationType => {
+                columns.push({
+                    name: `${oql.gene} ${alterationType}`,
+                    tooltip: <span>{oql.oql_line}</span>,
+                    headerDownload: (name: string) => oql.oql_line,
+                    render: (data: ICaseAlteration) => generateOqlDisplayValue(data.oqlData, oql.oql_line, alterationType) == "no alteration" || generateOqlDisplayValue(data.oqlData, oql.oql_line, alterationType) == "not profiled" ?
```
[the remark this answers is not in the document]
dippindots: Updated. Thanks for suggesting this change, it makes the code clearly and tidier than the original way.

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -93,13 +118,28 @@ export function generateOqlDisplayValue(oqlData: {[oqlLine: string]: IOqlData},
         const datum = oqlData[oqlLine];
 
         if (datum) {
-            oqlValue = datum.sequenced ? generateOqlValue(oqlData[oqlLine]) : "N/P";
+            oqlValue = datum.sequenced ? generateOqlValue(oqlData[oqlLine], alterationType) : "no alteration";
```
adamabeshouse: I find it a little confusing to call this "oql" since its not.. maybe something like "alteration summary". or pseudoOqlSummary or something. Just makes it more readable IMO

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -93,13 +118,28 @@ export function generateOqlDisplayValue(oqlData: {[oqlLine: string]: IOqlData},
...
         }
     }
 
     return oqlValue;
 }
 
+export function findAlterationType(oql: OQLLineFilterOutput<AnnotatedExtendedAlteration>, alterationData: ICaseAlteration[]): string[]
```
adamabeshouse: I dont like the word "find" here, it's not really finding anything is it? It's more like, computeAlterationTypes (since its finding more than one type).

also I don't see that the parameter oql is used?

src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -22,20 +21,54 @@ export interface IDownloadFileRow {
...
+                        isMrnaExpNotProfiled = true;
+                        break;
+                    case AlterationTypeConstants.PROTEIN_LEVEL:
+                        isProteinLevelNotProfiled = true;
+                        break;
+                    case AlterationTypeConstants.MUTATION_EXTENDED:
+                        isMutationNotProfiled = true;
+                    case AlterationTypeConstants.FUSION:
+                        isFusionNotProfiled = true;
+                        break;
```
adamabeshouse: would be much simpler to store this data as a map `const isNotProfiled ={[alterationTypeConstant:string]:boolean}`. Then lines 47-63 can be compressed to `isNotProfiled[molecularAlterationType] = true`

src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -22,20 +21,54 @@ export interface IDownloadFileRow {
 }
 
 export function generateOqlData(datum: GeneticTrackDatum,
```
onursumer: we may want to add more unit tests for this function to cover the latest changes

src/pages/resultsView/download/DownloadUtils.spec.ts
```diff
@@ -664,4 +664,252 @@ describe('DownloadUtils', () => {
...
+                    "datatype": "MAF",
+                    "name": "Mutations",
+                    "description": "Mutation data from whole exome sequencing. Mutation packager: LGG/20160128/gdac.broadinstitute.org_LGG.Mutation_Packager_Calls.Level_3.2016012800.0.0.tar.gz.",
+                    "showProfileInAnalysisTab": true,
+                    "molecularProfileId": "lgg_tcga_mutations",
+                    "studyId": "lgg_tcga"
+                } as MolecularProfile
+            };
+
+            const oqlDataZero = updateOqlData(
```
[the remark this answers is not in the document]
dippindots: Thanks! Updated.

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,71 +39,105 @@ export interface ICaseAlterationTableProps {
...
         }
     }
 
-    return oqlValue;
+    return pseudoOqlSummary;
+}
+
+export function computeAlterationTypes(alterationData: ICaseAlteration[]): string[]
+{
+    const types = new Set<string>();
```
alisman: we shouldn't use Set yet as it isn't fully supported.  Just use object

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,71 +39,105 @@ export interface ICaseAlterationTableProps {
...
+    alterationData.forEach(element => {
```
alisman: this code could be clarified with use of lodash.  you're basically trying to merge the values of alterationTypes collection for teach item in oqlData.   you should be using _.map and _.values and also look at _.chain to make things more succinct and easy to ready.

src/pages/resultsView/download/DownloadUtils.spec.ts
```diff
@@ -664,4 +664,252 @@ describe('DownloadUtils', () => {
                 "protein data is correct for the sample key VENHQS1FRS1BMjBDLTA2OnNrY21fdGNnYQ");
         });
     });
+
+    describe('updateOqlData', () => {
+        it('properly updates oql data', () => {
+            const sampleOql = {
```
alisman: lets call this a sampleOqlReport.   Try to include the type/identify of the entity in its name.  For example, this variable does not contain oql, it contains a report about oql.

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -33,71 +39,102 @@ export interface ICaseAlterationTableProps {
...
     }
 
-    return oqlValue;
+    return pseudoOqlSummary;
+}
+
+export function computeAlterationTypes(alterationData: ICaseAlteration[]): string[]
+{
+    const types = _.chain(alterationData)
+        .map((el) => _.values(el.oqlData))
```
onursumer: `alteration` might be a better variable name choice than `e1`

PR #1738, opened 2018-11-19
Fix GraphML and SIF download in network tab

alisman:
These links are written by server.   We now have to submit all case ids so need to do so via post.  

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/resultsView/network/network.htm
+++ b/src/pages/resultsView/network/network.htm
@@ -7,7 +7,6 @@
 <link rel="stylesheet" href="//maxcdn.bootstrapcdn.com/font-awesome/4.3.0/css/font-awesome.min.css"></link>
 <script src="https://cdnjs.cloudflare.com/ajax/libs/js-url/2.5.3/url.min.js"></script>
 
-
 <script>
 
     var queryParams = url("?");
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/network/network.htm
```diff
@@ -89,13 +88,61 @@
...
     };
+
+    // calls to network do must now include all case ids
+    // so we need to hijack click and form submit
+    $(document).ready(function(){
+        $("body").on("click","a",function(e){
+            e.preventDefault();
+
+            switch(this.innerText) {
+                case "GraphML":
```
adamabeshouse: man this is really rough.... capturing *every* `<a>` click? Is it possible to make it more specific?

alisman: it's only aesthetically unpleasing.   it's better to bind to all clicks than to have to worry about rendering timeline.  i think we should just leave it.  it's harmless.

alisman: i can easily limit it to the netmsg div.  this is actually not different because jquery is doing the same operation on every click, but whatever

adamabeshouse: the click delegation is fine, but how about using a class selector? "a" is so broad. How is it safe - wouldn't this apply always once this page is loaded?

adamabeshouse: the issue is in case any future link shows up with innerText = "GraphML" its unlikely but it still makes me uneasy

alisman: this is better.

PR #1740, opened 2018-11-19
Update sv density plot tooltip for clarity

adamabeshouse:
image


tmazor:
@adamabeshouse can you put the number of samples in the first row of the tooltip? I think the ranges make more sense if the user is first explicitly reminded that there are multiple samples.

And also for consistent capitalization, can you make it "Number of Samples" (capital "S")

adamabeshouse:
good points @tmazor

PR #1741, opened 2018-11-19
Add Download To Timeline

dippindots:
Add download options to timeline(SVG, PNG, PDF).

### DIFF
--- a/src/pages/patientView/timeline/Timeline.tsx
+++ b/src/pages/patientView/timeline/Timeline.tsx
@@ -12,6 +12,8 @@ import SampleManager from "../sampleManager";
 
 import {PatientViewPageStore} from "../clinicalInformation/PatientViewPageStore";
 import {ClinicalEvent, ClinicalEventData} from "../../../shared/api/generated/CBioPortalAPI";
+import DownloadControls from "../../../shared/components/downloadControls/DownloadControls";
+import autobind from "autobind-decorator";
 
 interface ITimelineProps {
     sampleManager:SampleManager;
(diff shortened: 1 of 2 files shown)


src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -112,10 +113,22 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
 
     }
 
+    private getSvg() {
+        return document.getElementsByClassName("genomicOverviewTracksContainer")[0].firstChild as SVGElement | null;
```
alisman: you should learn about React's ref pattern to avoid this global lookup.  See this documentation and look at "callback ref"  
https://reactjs.org/docs/refs-and-the-dom.html#callback-refs

alisman: i'm sure you found the above code elsewhere so there may be some reason we're not doing it there.  maybe just laziness!

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -112,10 +114,24 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
 
     }
 
+    private svgContainer: any;
```
alisman: it's a div so, you can type this as HTMLDivElement.    try to avoid using any.  it's not that big a deal here but good practice in general

dippindots: Thanks, Aaron. I just updated this. And I also changed the position of the download button.

PR #1747, opened 2018-11-21
move data set dropdown under the gene tab

dippindots:


src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -204,6 +204,9 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
                             );
                         })}
                     </MSKTabs>
+                    <Observer>
```
alisman: i see.  i thought it would be harder because it would need to go inside the tab, but it turns out we are only using MSK tabs for the buttons.

PR #1749, opened 2018-11-21
oncoprint progress indicator improvements

adamabeshouse:
(1) Add "clinical" to the loading message, when clinical data loading is required (when there are selected clinical attributes which are not locally computed)
(2) Querying x samples and y genes
(3) Loading message indicating expected length of query

@schultzn @tmazor @jjgao @alisman 
the thresholds and loading messages are determined in: https://github.com/cBioPortal/cbioportal-frontend/pull/1749/files#diff-dcf685585bcddc86c05877659a76496eR401

image

image

### DIFF
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -27,7 +27,7 @@ import tabularDownload from "./tabularDownload";
 import * as URL from "url";
 import classNames from 'classnames';
 import FadeInteraction from "shared/components/fadeInteraction/FadeInteraction";
-import {SpecialAttribute} from "../../cache/OncoprintClinicalDataCache";
+import {clinicalAttributeIsLocallyComputed, SpecialAttribute} from "../../cache/OncoprintClinicalDataCache";
 import OqlStatusBanner from "../oqlStatusBanner/OqlStatusBanner";
 import autobind from "autobind-decorator";
 import {getAnnotatingProgressMessage} from "./ResultsViewOncoprintUtils";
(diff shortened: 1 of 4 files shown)


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -911,9 +912,25 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
+
+            ret.push({
+                label: `Querying ${numGenes} genes in ${numSamples} samples - ${getLoadingMessage(numGenes, numSamples)}`,
+                promises: [], // empty promises means insta-complete
+                hideIcon: true, // dont show any icon, this is just a message
+                style:{ fontWeight:"bold" }
+            });
+        }
+
+        const areClinicalAttributesSelected = this.selectedClinicalAttributeIds.keys()
```
alisman: seems like a good case for _.some

src/shared/components/progressIndicator/ProgressIndicator.tsx
```diff
@@ -59,10 +63,13 @@ export default class ProgressIndicator extends React.Component<IProgressIndicato
             }
         }
         return (
-            <span className={classnames(styles["item-row"], {[styles["item-row-not-first"]]:(index > 0), [styles["not-invoked-item"]]:notInvoked})}>
-            {item.label}
+            <span
+                className={classnames(styles["item-row"], {[styles["item-row-not-first"]]:(index > 0), [styles["not-invoked-item"]]:notInvoked})}
```
alisman: you could accomplish this with nth-child?

adamabeshouse: thats true but i do feel that selector is a bit hacky/coupling view structure w logic. better IMO to just tag it manually. what do you think?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: good point

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -396,4 +396,15 @@ export function doesQueryHaveCNSegmentData(
...
         return _.some(detailedSamples, s=>!!s.copyNumberSegmentPresent);
     }
+}
+
+export function getLoadingMessage(numGenes:number, numSamples:number){
+    const numDataPoints = numGenes * numSamples;
+    if (numDataPoints < 1500) {
+        return "this shouldn't take long.";
+    } else if (numDataPoints < 50000) {
+        return "this can take several seconds.";
```
jjgao: this also depends on the network -- maybe we should make it vaguer?

adamabeshouse: would it be worth doing some empirical data collection to see if we can draw an approximate curve, then we can say "this usually takes about X seconds"?

alisman: I think use only really needs messaging fgor something we know will take a fair amount of time.  Like > 5 seconds.  That's why I kind of like just setting a timeout and turing on the message after the timeout.

src/shared/components/progressIndicator/ProgressIndicator.tsx
alisman: hmmm.  if it's just style, it seems opposite of hacky to me.  in there words, logic/function world doesn't care at all about the fact that the row is first.  it's just a style thing so it should be unobtrusively applied via selector.

src/pages/resultsView/ResultsViewPageStoreUtils.ts
adamabeshouse: what should the message say?

adamabeshouse:
@alisman thoughts?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -932,10 +959,12 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
 
     public render() {
         return (
-            <div className="posRelative">
+            <div>
```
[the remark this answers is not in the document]
adamabeshouse: yes AFAI can tell its right

PR #1751, opened 2018-11-26
fix sort by allele freq not working properly on patient view

dippindots:
Put the NaN raws to the bottom all the time.

### DIFF
--- a/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
+++ b/src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
@@ -98,10 +98,9 @@ export default class AlleleFreqColumnFormatter {
         }
 
         // as long as we have tooltip lines, show tooltip in either cases (single or multiple)
-        if (tooltipLines.length > 0)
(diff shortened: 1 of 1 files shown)


alisman:
@dippindots if you hover over a empty row, you'll still see a tooltip with NaN displayed.   @inodb confirm but I think we should not show any tooltip, right?
http://cbioportal-frontend-pr-1751.herokuapp.com/patient?studyId=lgg_ucsf_2014&tab=summaryTab&caseId=P04

src/pages/patientView/mutation/column/AlleleFreqColumnFormatter.tsx
```diff
@@ -98,10 +98,9 @@ export default class AlleleFreqColumnFormatter {
         }
 
         // as long as we have tooltip lines, show tooltip in either cases (single or multiple)
-        if (tooltipLines.length > 0)
+        if (tooltipLines.length > 0 && freqs.filter((freq) => freq !== undefined).length > 0)
```
alisman: optimization: use _.some when you are checking for "at least one" because it will stop iteration after it find a single element that complies with condition.

PR #1752, opened 2018-11-26
Custom tabs: invoke callback on unmount event of tab

alisman:
Custom tabs need to implement an unmount event in order to dispose (cleanup) reactions that they created on mounting.  This in turn will allow gc of observables to which the tabs were bound.  MSKTabs needs to call any configured mount and unmount handlers passed as properties.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -181,8 +181,8 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
     }
 
     @autobind
-    private customTabMountCallback(div:HTMLDivElement,tab:any){
-        showCustomTab(div, tab, this.props.routing.location, this.resultsViewPageStore);
+    private customTabCallback(div:HTMLDivElement,tab:any, isUnmount = false){
+        showCustomTab(div, tab, this.props.routing.location, this.resultsViewPageStore, isUnmount);
     }
 
     componentWillUnmount(){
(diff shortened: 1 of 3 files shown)


oplantalech:
@alisman I did some performance tests with interesting results. I basically create a query with 10 genes and I changed the query 10 times, either adding a gene or removing one.

This is the "basic" performance test, in cbioportal.org:
pt cbioportal org

Then, I tested the performance of your code in (1) Docker installation with custom tabs defined in portal.properties and (2) Docker installation without custom tabs defined in portal.properties; in both cases there is a similar increase of memory:
pt docker custom tab active
pt docker no custom tab active

I did the same (also in Docker) with the current rc code, again (1) with active custom tabs and (2) without custom tabs, with similar results:
pt docker custom tab active no pr
pt docker no custom tab no pr

Finally, I tested your PR code locally, just to ensure that the problem was not created by Docker, and again I see the same results with (1) custom tabs, and (2) no custom tabs:
pt no docker custom tab active
pt no docker no custom tab active

pieterlukasse:
Interesting. On cbioportal.org's version, the memory seems to get released, but in the figure from `rc` this seems not really to be the case.

alisman:
Hi guys, I realized I forgot to tell you a key detail : (   

We our api client caches the api results in memory.  So if you change the query, it keeps previous results in memory at the level of the request.   By default it does this up to 450MB, at which point it clears space as necessary.  There is a way to turn this off.  Run this in console:
localStorage.frontendConfig = JSON.stringify({"serverConfig":{ "api_cache_limit":0} });

image

This is a better result.  Still look like there's a creep upward each time which bears looking into.  Oleguer, could you repeat your experiment with the cache limit set to zero?

PR #1755, opened 2018-11-26
oql filtering - add _GERMLINE and _SOMATIC mutation modifiers

adamabeshouse:


src/shared/lib/oql/oql-parser.pegjs
```diff
@@ -19,7 +19,7 @@ zmbs = zero_or_more_breaks_and_spaces:[; \t\r\n]*
 ombs = one_or_more_breaks_and_spaces:[; \t\r\n]+
 
 StartMergedGenes
-	= "[" zmbs "\"" StringExceptQuotes "\"" {return {"label": label, "list":[]};}
+	= "[" zmbs "\"" label:StringExceptQuotes "\"" {return {"label": label, "list":[]};}
```
adamabeshouse: this bug was introduced before. @inodb seems like oql-parser.spec.ts isn't running with the build?

adamabeshouse:
@jjgao same way we determine if it is germline - according to the mutationStatus field on the API response

jjgao:
@adamabeshouse I see. One problem is that not all MAFs are properly filled for mutationStatus. (Germline mutations are usually probably filled.) Maybe we can count all non-germline as somatic? @schultzn

schultzn:
Yes, good idea.

adamabeshouse:
@jjgao @schultzn okay I've updated the code to consider every mutation that doesnt explicitly have a status, or has a status that is not "germline" (case insensitive), as somatic

tmazor:
Nice! A few thoughts from testing:

1) All the symbols are broken but I think that's a heroku thing?
image

2) The mutations tab doesn't seem to respect GERMLINE/SOMATIC filtering even though it generally does respect other OQL specifications. Is this intended?

3) Testing with BRCA1 in ov_tcga_pub, these work:
MUT.GERMLINE
MUT=FRAMESHIFT.GERMLINE
These don't work:
MUT=E23.GERMLINE
MUT=E23VFS*17.GERMLINE
Is there a reason not to support specific mutations? I think we should support those last 2 use cases.

4) We need to update the OQL page along with this feature. I'm happy to be responsible for this once we've sorted out the functionality listed above.

5) If I add ".SOMATIC" to a non-mutation OQL term, the current error message is about an invalid gene symbol but should be about invalid OQL
image

6) Did we consider using an ampersand rather than a period to connect the terms? It might be a bit more intuitive to write "MUT&SOMATIC" rather than "MUT.SOMATIC".

adamabeshouse:
thanks @tmazor for the detailed feedback. 
(1) yeah most likely due to heroku, I'm not seeing that on my local instance
(2) good catch
(3) (and (6)) Ah so those specs aren't working because the ".GERMLINE" is being considered part of the protein change code. To fix this, we need to exclude "." from protein change codes (when I search the DB I don't find any with ".", so that seems it would be okay). Or if we change characters it would need to be another which isn't in protein change codes. & works but personally I find it too visually cluttered. I liked "." because it reminds me of the R programming conventions. I think "/" could also work?
(4) Sure, thanks!
(5) This looks like a bug introduced with merged query. With a fix, the query will just say it's a failure. But it won't give more specific debug info. Is that something we want to invest some time into? It would involve adding certain common OQL mistakes into the parsing grammar and then being more specific with the error we return.

jjgao:
@tmazor great list. Thanks!

Somehow I like `missense.germline` more than `missense&germline`, but it's probably just my engineer's brain. @schultzn thoughts?

(@adamabeshouse I don't hink missense/germline` because I would interpret it as OR.)

adamabeshouse:
good point @jjgao . I think I don't like & because it's the same height as letters, so theres less of a visual break. what about "-" ?

tmazor:
What about a plus sign: MUT=MISSENSE+GERMLINE ?

Also since we use a non-fixed-width font, the period can get a bit lost, which is another reason to try a different character.

In regards to (5), I think it's ok to have a generic "invalid OQL" statement -- but also, I only just realized now that we have any specificity in the OQL error messages. Is this new?

adamabeshouse:
@tmazor no, we have always had no specificity in the OQL error messages. Im not sure whether it would be worth investing time into that. The best we have is invalid gene and the warning that something is being interpreted as a protein change code.

yeah I like + !

tmazor:
@adamabeshouse I also thought there was no specificity in the error message, but I was just playing around on production and I get this:
image

And then I can click Submit and I get this:
image

Interestingly, I can do this without having selected a study. And also the cursor actually doesn't go to the location of the OQL error. But it's sounds like it's trying to? I don't remember having seen this before? Is this all new?

adamabeshouse:
@tmazor the failure to move the cursor is new AFAIK, but that functionality has been there for a while

tmazor:
Ok, so I've just been oblivious to it previously :)  Thanks!  I'll make an issue about the failure to move the cursor since that sounds like a new bug then.

In any case, now that I've semi-hijacked this PR, I'm fine with a generic OQL error comment. I don't think it's worth investing time on more specific error messages at this point.

adamabeshouse:
@jjgao @schultzn what do you think about using + as the delimiter?

jjgao:
@adamabeshouse I like `+` 👍

adamabeshouse:
I just pushed the update using +. This fixes the OQL parsing errors. But there are still some bugs (mainly making note for myself here)
(1) using + causes some issues with URI encoding and it doesnt make it through to the query
(2) Mutation tab still not working right
(3) The oql grammar is currently broken, it was broken from before from the changes made to support merged gene tracks, but we didn't see it until now. And it's going to take a little more work than expected to fix it.

schultzn:
I don't like "+" - it looks like the specification of two separate mutations. How about a hypen or an underscore?

adamabeshouse:
@schultzn out of those two I prefer a hyphen. @jjgao @tmazor?

tmazor:
I like the hyphen also

adamabeshouse:
the only problem with - which I just realized is that it conflicts with our use in the study search bar, which is "exclude". so maybe underscore is better

schultzn:
I had that thought too… (but in the search, doesn’t there have to a space before the “-“? But I don’t mind using underscore.

adamabeshouse:
@tmazor @jjgao feelings about underscore?

tmazor:
@adamabeshouse underscore is fine by me!

adamabeshouse:
@schultzn agreed "-" is less unclear without a space in front, but I have it that "_GERMLINE" alone is valid oql, equivalent to all germline mutations. so it could be very confusing in that case

adamabeshouse:
@tmazor @jjgao @schultzn its updated and ready to be tested again :+1:

jjgao:
@adamabeshouse this is pretty cool! http://cbioportal-frontend-pr-1755.herokuapp.com/results/oncoprint?session_id=5c0af1b8e4b024a8f3efa7cd

Should we also support `BRCA1: GERMLINE` (w/o underscore) and `BRCA1: GERMLINE_TRUNC`? @schultzn @tmazor

schultzn:
Yes, we should definitely support BRCA1: GERMLINE

GERMLINE_TRUNC might be too special a case - at least until someone uses cBioPortal with all germline variants, including known or likely benign SNPs.

adamabeshouse:
@schultzn @jjgao Im confused -- isnt GERMLINE_TRUNC equivalent to TRUNC_GERMLINE?

jjgao:
@adamabeshouse yes, `GERMLINE_TRUNC` should be equivalent to `TRUNC_GERMLINE`. Currently, only `TRUNC_GERMLINE` is supported.

adamabeshouse:
okay @jjgao that change has been made

PR #1756, opened 2018-11-26
Fix tests

adamabeshouse:
(1) A test that breaks with diff sample counts
(2) Make oncoprint clinical track selector not load until click - this fixes tests that are breaking depending on how long this takes.


adamabeshouse:
@jjgao can you please confirm that it's okay that, in the All case list, constructed from all studies selected that match `tcga -provisional`, the number of samples changes as data gets updated. So we don't have to tether our tests to this number (which is what this PR does)

jjgao:
@adamabeshouse this is fine.

PR #1761, opened 2018-11-27
Add quick search to the home page

ersinciftci:


src/appBootstrapper.jsx
```diff
@@ -36,7 +36,7 @@ import getBrowserWindow from "shared/lib/getBrowserWindow";
 import {getConfigurationServiceApiUrl} from "shared/api/urls";
 import {AppStore} from "./AppStore";
 
-superagentCache(superagent);
+//superagentCache(superagent);
```
ersinciftci: Caching causes a bug in the search bar (if the load more option clicked for a term for the second time, the dropdown disappears), so I'm turning it off for now until I find a fix, so the heroku demo wouldn't be buggy.

alisman:
replaced with https://github.com/cBioPortal/cbioportal-frontend/pull/1966

jjgao:
@alisman is this replaced with #1966?

inodb:
This is in release-2.2.0 branch now

PR #1763, opened 2018-11-28
Format DAT dates for download token file

angelicaochoa:
Signed-off-by: Angelica Ochoa <aochoa4230@gmail.com>


src/shared/components/dataAccessTokens/DataAccessTokens.tsx
```diff
@@ -32,10 +33,11 @@ export interface IDataAccessTokensProps {
 }
 
 function buildDataAccessTokenFileContents(dat:UserDataAccessToken | undefined) {
+    const dateFormat = "YYYY/MM/DD HH:MM:SS";
```
sheridancbio: Should we include a timezone? Do we know if our expiration dates are represented in GMT or in ET? Maybe there is a simple dateFormat that will attach the correct timezone for the user locale. It looks like timezone can be added to the format with "Z", like : dateFormat = "YYYY/MM/DD HH:MM:SS Z"

PR #1767, opened 2018-11-28
redirect from default results page route to real route (for tracking …

alisman:
…purposes)

# What? Why?
Fix # .

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -26,17 +26,13 @@ import {ITabConfiguration} from "../../shared/model/ITabConfiguration";
 import getBrowserWindow from "../../shared/lib/getBrowserWindow";
 import CoExpressionTab from "./coExpression/CoExpressionTab";
 import Helmet from "react-helmet";
-import {createQueryStore} from "../home/HomePage";
-import {ServerConfigHelpers} from "../../config/config";
 import {showCustomTab} from "../../shared/lib/customTabs";
 import {
     getTabId, parseConfigDisabledTabs, ResultsViewTab,
     updateStoreFromQuery
 } from "./ResultsViewPageHelpers";
 import {buildResultsViewPageTitle, doesQueryHaveCNSegmentData} from "./ResultsViewPageStoreUtils";
-import {filterAndSortProfiles} from "./coExpression/CoExpressionTabUtils";
 import {AppStore} from "../../AppStore";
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -482,13 +484,21 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
 
     public render() {
-        return (
-            <PageLayout noMargin={true}>
-                {
-                    this.pageContent
-                }
-            </PageLayout>
-        )
+
```
alisman: we should probably handle studies loading here, but we don't.  don't really want to worry about it now

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -482,13 +485,21 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+        if (this.resultsViewPageStore.studies.isComplete && !this.resultsViewPageStore.tabId) {
+            setTimeout(()=>{
```
alisman: i would be nicer to handle this redirect in router but we need resultsViewStore data to determine where to send user.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -437,22 +437,25 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
         return isRoutedTo || (!isExcludedInList && !isExcluded);
     }
 
-    // if it's undefined, MSKTabs will default to first
-    public currentTab(tabId:string|undefined):string | undefined {
+    public currentTab(tabId:string|undefined):string {
```
alisman: gonna write a test for this

src/pages/resultsView/ResultsViewPage.tsx
adamabeshouse: theres no need to debounce/clear this bc its idempotent right?

alisman: Yes, if the tabId route was set multiple times, it wouldn't have any affect.  
It's also highly unlikely (maybe impossible) that render would be called twice in same frame.  And the code run in the timeout sets tabId (next frame).  So only change this could run twice is if another render was able to execute in same frame.

PR #1773, opened 2018-11-29
Stacked bar chart in plots tab for discrete v discrete data

adamabeshouse:
The colors are chosen by following process:
- If it is a value that we have a color for, e.g. one of the study view clin attr colors, or oncoprint mutation/cna colors, then we use it,
- Otherwise, we select from a preset list of colors (same process as in oncoprint)

image

image

image

image

image

image

image


src/shared/components/plots/PlotUtils.ts
```diff
@@ -107,6 +107,57 @@ export function scatterPlotSize(
     }
 }
 
+function darkenHexChannel(c:string) {
```
alisman: should we be using a library for this?   open for discussion

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -0,0 +1,630 @@
...
+            return 23.7*numRows;
+        }
+    }
+
+    @computed get legendData() {
+        return this.data.map(obj=>({
+            name: obj.minorCategory,
+            symbol:{
+                type: "square",
+                fill: this.getColor(obj.minorCategory),
```
alisman: this seems slightly weird.  see above

src/shared/components/plots/PlotUtils.ts
adamabeshouse: for which part?

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -0,0 +1,630 @@
...
+        this.container = container;
+    }
+
+    @computed get getColor() {
+        const uniqueColorGetter = makeUniqueColorGetter(_.values(this.props.categoryToColor));
+        const categoryToColor:{[category:string]:string} = {};
+        _.forEach(this.props.categoryToColor, (color, category)=>{
+            categoryToColor[category.toLowerCase()] = color;
+        });
+        return function(category:string) {
```
alisman: i think you could make this function a method of the class the references a computed value categoryToColor.   it's maybe a little clearer?

adamabeshouse: i dont understand your sentence, maybe some typos?

src/shared/components/plots/PlotUtils.ts
alisman: all of this color manipulation stuff

adamabeshouse:
@alisman approve?

PR #1775, opened 2018-11-30
new study view preliminary test

dippindots:
A preliminary test for new study view


end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -462,3 +462,24 @@ describe('result page tabs, loading from session id', function(){
...
 });
+
+
+describe('new study view screenshot test', function(){
+    before(function(){
+        var url = `${CBIOPORTAL_URL}/study?id=laml_tcga`;
+        goToUrlAndSetLocalStorage(url);
+    });
+     it('new study view laml_tcga', function() {
+        waitForNetworkQuiet();
```
alisman: I would think that mainColumn would be visible before networkQuiet, no?

PR #1778, opened 2018-12-01
oncoprint - add option to show oql filters as "sublabels" in tracks

adamabeshouse:
image

image

Im not showing them in merged tracks, but they do show in the expansions as seen in screenshot


schultzn:
Looks good! I assume the full OQL shows up upon mouseover?

pieterlukasse:
Nice :+1:

jjgao:
somehow I couldn't deploy to heroku...

@adamabeshouse when a user is explicitly using oql, are you displaying them by default?

adamabeshouse:
no, the default is off. but we can do that instead - if user specifies OQL then we'll turn it on by default

adamabeshouse:
no, the default is off. but we can do that instead - if user specifies OQL then we'll turn it on by default

jjgao:
@adamabeshouse I think that would be good.

pieterlukasse:
@adamabeshouse should this have gone to `master` instead of `rc`?

adamabeshouse:
@pieterlukasse ah maybe. i thought it was small enough to go to master

PR #1779, opened 2018-12-03
Added CN segments tab for the study view page

onursumer:
# What? Why?
- Partial fix for cBioPortal/cbioportal#5221
- Clinical attribute tracks are still missing: https://github.com/cBioPortal/cbioportal/issues/4410
- Optimized the rendering w.r.t. to visibility and track configuration changes

# Checks
  the `-s` option)

### DIFF
--- a/src/pages/resultsView/cnSegments/CNSegments.tsx
+++ b/src/pages/resultsView/cnSegments/CNSegments.tsx
@@ -6,6 +6,7 @@ import autobind from "autobind-decorator";
 import {Nav, NavItem} from "react-bootstrap";
 
 import {ResultsViewPageStore} from "../ResultsViewPageStore";
+import {ResultsViewTab} from "../ResultsViewPageHelpers";
 import {CopyNumberSeg, Gene} from "shared/api/generated/CBioPortalAPI";
 import IntegrativeGenomicsViewer from "shared/components/igv/IntegrativeGenomicsViewer";
 import CNSegmentsDownloader from "shared/components/cnSegments/CNSegmentsDownloader";
@@ -52,18 +53,7 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
 
     @computed get filename()
     {
-        let prefix = "";
-
-        if (this.props.store.studies.result) {
-            if (this.props.store.studies.result.length > 1) {
-                prefix = "multi_study_";
-            }
-            else if (this.props.store.studies.result.length === 1) {
-                prefix = `${this.props.store.studies.result[0].studyId}_`;
-            }
-        }
-
-        return `${prefix}segments.seg`;
+        return `${this.props.store.downloadFilenamePrefix}segments.seg`;
     }
 
     @computed get isHidden() {
@@ -107,7 +97,7 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
     @computed get progressItems(): IProgressIndicatorItem[] {
         return [
             {
-                label: "Downloading copy number segments data...",
+                label: "Loading copy number segments data...",
                 promises: this.indicatorPromises
             },
             {
@@ -143,7 +133,7 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
                         ))
                     }
                 </Nav>
-                {!this.isLoading && (
+                <div style={this.isHidden ? {opacity: 0} : undefined}>
                     <IntegrativeGenomicsViewer
                         tracks={[
                             {
@@ -156,8 +146,9 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
                         onRenderingStart={this.onIgvRenderingStart}
                         onRenderingComplete={this.onIgvRenderingComplete}
                         disableSearch={this.activeLocus !== WHOLE_GENOME}
+                        isVisible={this.props.store.tabId === ResultsViewTab.CN_SEGMENTS && !this.isHidden}
                     />
-                )}
+                </div>
             </div>
         );
     }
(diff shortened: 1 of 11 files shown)


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -93,6 +94,9 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                                         <IFrameLoader height={700}
                                                       url={`//bioinformatics.mdanderson.org/TCGA/NGCHMPortal/?${this.store.MDACCHeatmapStudyMeta.result[0]}`}/>
                                     </MSKTab>
+                                    <MSKTab key={3} id={"cnSegments"} linkText={"CN Segments"} unmountOnHide={true}>
```
zhx828: The `id` and `linktext` are moved to Enums. Please take a look other tabs.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2236,6 +2238,13 @@ export class StudyViewPageStore {
         default: []
     });
 
+    readonly cnSegments = remoteData<CopyNumberSeg[]>({
+        await: () => [
+            this.samples
```
zhx828: Shouldn't you wait for `this.selectedSamples`?

src/pages/studyView/tabs/CNSegments.tsx
```diff
@@ -0,0 +1,113 @@
...
+        return this.isLoading || !this.renderingComplete;
+    }
+
+    @computed get isLoading()
+    {
+        return this.activePromise ? this.activePromise.isPending : true;
+    }
+
+    @computed get activePromise() {
+        return this.props.store.cnSegments;
```
zhx828: The promise will be lost? If not, I don't see the benefit of checking and assigning the default in `features`

src/pages/studyView/tabs/CNSegments.tsx
```diff
@@ -0,0 +1,113 @@
...
+    }
+
+    @computed get progressItems(): IProgressIndicatorItem[] {
+        return [
+            {
+                label: "Downloading copy number segments data...",
```
zhx828: Let's use `Loading` instead of `Downloading` to keep the consistency with other tabs.

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -143,7 +144,7 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
                         ))
                     }
                 </Nav>
-                {!this.isLoading && (
+                <div style={this.isHidden ? {opacity: 0.1} : undefined}>
```
onursumer: hiding the actual component causes problems with rendering, so changing the opacity instead

src/shared/components/cnSegments/CNSegmentsDownloader.tsx
```diff
@@ -27,7 +27,6 @@ export default class CNSegmentsDownloader extends React.Component<CNSegmentsDown
     };
 
     @observable downloading = false;
-    private downloadFileContent: string|undefined;
```
onursumer: removing this because caching download file content is problematic when the promise is updated due to filters

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -156,8 +157,9 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
                         onRenderingStart={this.onIgvRenderingStart}
                         onRenderingComplete={this.onIgvRenderingComplete}
                         disableSearch={this.activeLocus !== WHOLE_GENOME}
+                        isVisible={this.props.store.tabId === ResultsViewTab.CN_SEGMENTS && !this.isHidden}
```
onursumer: Component needs to be aware of its own visibility in order to prevent rendering when hidden

src/pages/studyView/tabs/CNSegments.tsx
onursumer: Probably we don't need it here. I just wanted this class to have the same structure/design with the results view `CNSegments` class.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -270,6 +275,8 @@ export class StudyViewPageStore {
        });
     }
 
+    @observable tabId: StudyViewPageTabKeyEnum|undefined = undefined;
```
zhx828: There is already a observable `currentTab`

src/pages/studyView/tabs/CNSegments.tsx
```diff
@@ -0,0 +1,114 @@
...
+        const segments: CopyNumberSeg[] = this.activePromise ? this.activePromise.result || [] : [];
+
+        return generateSegmentFeatures(segments);
+    }
+
+    @computed get filename()
+    {
+        let prefix = "";
+
+        if (this.props.store.studyIds.length > 0) {
```
zhx828: Can we consolidate the logics of generating the download file name? We have a very similar logic in `clinicalDataDownloadFilename`, but the name is bit different.

onursumer: added a new computed value named `downloadFilenamePrefix`

onursumer: also added new utility functions in the new `FilenameUtils`

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -188,6 +189,9 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                                         <IFrameLoader height={700}
                                                       url={`//bioinformatics.mdanderson.org/TCGA/NGCHMPortal/?${this.store.MDACCHeatmapStudyMeta.result[0]}`}/>
```
alisman: could we just the urls method to get the url for iframe?

onursumer: we should probably fix this in a separate PR

PR #1783, opened 2018-12-03
onMobxPromise - automatically dispose when a component unmounts

adamabeshouse:
do not merge reason: should try using it before merging
just want to wait for @alisman to sign off on the interface before going forward with that


alisman:
Another thought:  shouldn't the onMobxPromise actually be a member of the mobxpromise is observers?  That way, the promise could dispose its own handlers.  i.e.    store.genes.onComplete( )
I realize this would require a lot of refactoring : (

alisman:
Also, I don't believe that right now any of our mobxpromises that are store members every dispose anyway.   They just don't fire anymore because the stores become defunct every time a new query happens.

alisman:
Hi, sorry about sloppy review.   Another idea: what about a decorator for react components which would dispose instances of onMobxPromise on dismount?  Then user doesn't have to do anything except make the onMobxPromise an instance.   We could have a similar thing for stores.

adamabeshouse:
I like the UX of the idea. But the problem is the invokation of onMobxPromise doesnt receive the caller as context. `this` is undefined. so Im not sure how youd accomplish the registration of the onMobxPromise calls. no matter what the user has to do something to notify onMobxPromise where its calling from, and I think simply passing in the argument is simpler than asking them to deal w some factory initiailization or .bind or anything

adamabeshouse:
ooh wait... just had a brain blast. the decorator can add a private member this.onMobxPromise = onMobxPromise.bind(this)

adamabeshouse:
and the user would use that

adamabeshouse:
better idea: https://github.com/cBioPortal/cbioportal-frontend/pull/1792

PR #1786, opened 2018-12-04
Port gene sets to co-expression tab

oplantalech:
This PR adds gene sets to the co-expression tab and also extends the functionality of the current co-expression tab. Thanks to a new endpoint (https://github.com/cBioPortal/cbioportal/pull/5354) it is possible to calculate the co-expression (correlation) of different profiles, e.g. BRCA1 from mRNA profile against genes from RPPA profile (or gene sets from GSVA profile). Therefore, there are two dropdowns for profiles, the "subject" profile (specifying the profile for the queried gene / gene set) and the "query" profile (the profile for the correlated genes / gene sets).

Example of co-expression between mRNA expression from SOX9 and GSVA scores of gene sets:
captura de pantalla 2019-01-17 a les 15 15 37

Example of GSVA score of BIOCARTA_ASBCELL_PATHWAY against RPPA scores of genes:
captura de pantalla 2019-01-17 a les 15 16 14

You can try it yourself here: https://cbioportal-test.thehyve.net:8081/cbioportal/


oplantalech:
I added the DO NOT MERGE tab because first the backend endpoint needs to be merged, but you can already start reviewing the code.

adamabeshouse:
very cool! Ill look when I get in, but one thing I notice is that the phrasing Subject and Query are a little confusing to me. What about instead of Subject, you put the actual subject gene/geneset name? eg BRAF Profile:

adamabeshouse:
very cool! Ill look when I get in, but one thing I notice is that the phrasing Subject and Query are a little confusing to me. What about instead of Subject, you put the actual subject gene/geneset name? eg BRAF Profile:

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,53 +142,112 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
         );
     }
 
+    @bind
+    private isSelectedGeneticEntityAGeneset(geneticEntity:GeneticEntity) {
```
adamabeshouse: - this is a setter but its named like a getter
- it doesnt use the parameter
- it has the same name as the @computed getter above, really confusing for me

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,53 +142,112 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
+                        <span>Subject Profile:</span>
+                        <div style={{display:"inline-block", width:376, marginLeft:4, marginRight:4, zIndex:15 /* so that on top when opened*/}}>
+                            <Select
+                                name="subject-profile-select"
+                                value={this.isSelectedGeneticEntityAGeneSet ? this.selectedSubjectProfileGeneset.molecularProfileId : this.selectedSubjectProfileGene.molecularProfileId}
+                                onChange={this.isSelectedGeneticEntityAGeneSet ? this.onSelectSubjectProfileGeneset : this.onSelectSubjectProfileGene}
+                                options={this.isSelectedGeneticEntityAGeneSet ? subjectOptionsGeneset : subjectOptionsGene}
+                                searchable={false}
+                                clearable={false}
+                                disabled={this.isSelectedGeneticEntityAGeneSet ? true : false}
```
adamabeshouse: you dont need the ternary here, `this.isSelectedGeneticEntityAGeneset` is already boolean :), so just `disabled={this.isSelectedGeneticEntityAGeneset}`

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -149,56 +260,69 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
+                        arrowStyle={{'line-height':0.8}}
                     >
-                        {this.props.store.genes.result!.map((gene:Gene, i:number)=>{
+                        {this.props.store.geneticEntities!.map((geneticEntity:GeneticEntity, i:number)=>{
                             return (
                                 <MSKTab
                                     key={i}
-                                    id={gene.entrezGeneId+""}
-                                    linkText={gene.hugoGeneSymbol}
+                                    id={String(geneticEntity.geneticEntityId)}
```
adamabeshouse: I think we've been using `toString()`

src/pages/resultsView/coExpression/CoExpressionTabUtils.ts
```diff
@@ -6,8 +6,71 @@ export const correlationInformation = "Pearson correlations are computed first.
...
+            const profileId = profile.molecularProfileId.toLowerCase();
+            good = profileId.indexOf("gsva_pvalues") === -1;
+        }
+        return good;
+    });
+    return profs;
+}
+
+export function sortProfiles(profs: MolecularProfile[]) {
+    profs.sort(function(profA, profB) {
```
adamabeshouse: I think its clearer to use `_.sortBy`, that way it's very easy to see what you're sorting by, without needing to check out logic. It's also more concise, and it doesn't modify the input array in place, which saves some potential bugs down the line. In this case you could do `_.sortBy(profs, profile=>(profile.molecularProfileId.toLowerCase().indexOf("rna_seq") > -1) ? 0 : 1)`

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -295,7 +325,7 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
                     { (this.plotData.isComplete && this.highlightedCoExpression) && (
                         <div style={{marginLeft:10}}>
                             <CoExpressionPlot
-                                xAxisGene={this.props.gene}
+                                xAxisGene={this.props.geneticEntity}
```
adamabeshouse: should we change the prop name to `xAxisGeneticEntity`? etc

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -86,15 +92,17 @@ export class CoExpressionDataStore extends SimpleGetterLazyMobXTableApplicationD
 @observer
 export default class CoExpressionViz extends React.Component<ICoExpressionVizProps, {}> {
 
-    @observable.ref highlightedCoExpression:CoExpression|undefined; // only undefined initially, before data loaded
+    @observable.ref highlightedCoExpression:MolecularProfileCorrelation|undefined; // only undefined initially, before data loaded
```
adamabeshouse: should it be `highlightedCorrelation`?

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -86,15 +92,17 @@ export class CoExpressionDataStore extends SimpleGetterLazyMobXTableApplicationD
...
     @observable allDataRequested:boolean = true; // set to true to request all data by default
 
-    private lastCoExpressionData:CoExpression[];
+    private lastCoExpressionData:MolecularProfileCorrelation[];
```
adamabeshouse: should it be `lastCorrelationData`?

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -304,7 +334,8 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
                                 showMutations={this.plotShowMutations}
                                 logScale={this.plotLogScale}
                                 handlers={this.props.plotHandlers}
-                                molecularProfile={this.props.molecularProfile}
+                                queryMolecularProfile={this.props.queryProfile}
+                                subjectMolecularProfile={this.props.subjectProfile}
```
adamabeshouse: I'm not really getting the conceptual difference between query and subject, besides as explaining to the user. Seems to me it could more clearly be `xMolecularProfile` and `yMolecularProfile`? Could you please change it to that everywhere, unless I'm misunderstanding?

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
```diff
@@ -54,16 +55,23 @@ export function computePlotData(
     };
     for (const datum of mutationData) {
         if (datum.proteinChange) {
-            const targetData = (datum.entrezGeneId === xEntrezGeneId ? xMutations : yMutations);
+            const targetData = (datum.entrezGeneId === xGeneticEntityId ? xMutations : yMutations);
```
adamabeshouse: wont this code break when genetic entities are genesets?

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
```diff
@@ -73,16 +81,16 @@ export function computePlotData(
         const yDatum = yData[uniqueSampleKey];
         if (xDatum && yDatum) {
             // only add data if data for both axes
-            const xVal = xDatum.value;
-            const yVal = yDatum.value;
+            const xVal = Number(xDatum.value);
```
adamabeshouse: we typically use parseFloat

src/shared/components/lazyMobXTable/styles.scss
```diff
@@ -27,3 +27,14 @@ th.sort-asc:after {
     margin-bottom: 5px;
   }
 }
+
+.table td.text {
```
adamabeshouse: are we sure we want to make this change for *every* LazyMobxTable? sure you dont want to just put this into the coexpression tab tables css?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,53 +142,112 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
-                <div
+                <div>
+                    <div
+                        style = {{
+                            display: "flex",
+                            alignItems: "center",
+                            padding: "5px"
+                        }}
+                    >
+                        <span>Subject Profile:</span>
```
adamabeshouse: I think `Subject` is a little unclear, should we show this to the user as `${xGeneticEntity.name} Profile` (or y, not sure which it corresponds to)?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -227,6 +230,14 @@ interface IQueriedMergedTrackCaseData {
     list?: IQueriedCaseData<object>[];
 }
 
+export type GeneticEntity = {
```
adamabeshouse: can you add comments next to the type def lines saying what each value is in the case of gene vs geneset?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -21,28 +26,64 @@ export interface ICoExpressionTabProps {
...
+            && this.props.store.coexpressionTabMolecularQueryProfiles.result.length > 0) {
+            return this.props.store.coexpressionTabMolecularQueryProfiles.result[0];
+        } else {
+            return this._selectedQueryProfile;
+        }
+    }
+
+    @computed get selectedGeneticEntityId():string | undefined {
+        if (!this._selectedGeneticEntityId && this.props.store.geneticEntities !== undefined) {
+            return String(this.props.store.geneticEntities[0].geneticEntityId);
```
adamabeshouse: and please change all instances of String() to .toString() and Number() to parseFloat()

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,53 +142,112 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
-                        <MolecularProfileSelector
-                            name="data-set-select"
-                            value={this.selectedMolecularProfile.molecularProfileId}
-                            onChange={this.onSelectDataSet}
-                            molecularProfiles={this.props.store.coexpressionTabMolecularProfiles.result}
-                            molecularProfileIdToProfiledSampleCount={this.props.store.molecularProfileIdToProfiledSampleCount.result}
-                            className="coexpression-select-profile"
-                        />
+                    >
+                        <span>Query Profile:</span>
```
adamabeshouse: I'm more okay with using the word Query here, as displayed to the user. though in the code I think we should not have this concept, just stick with x and y

pieterlukasse: I think x and y will be confusing if the api endpoint is again again subject and query profiles... @oplantalech how is this named in the api now?

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
oplantalech: @adamabeshouse No, because gene sets do not contain mutation data.

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
oplantalech: but parseFloat expects to receive a string, and xDatum.value and xDatum.value sometimes they are strings and sometimes, already numbers.

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: I've removed this part, it is not used

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: @adamabeshouse good catch 👍

src/pages/resultsView/coExpression/CoExpressionViz.tsx
oplantalech: I've changed all the references to 'subjectProfiles' into 'xProfiles', and 'queryProfiles' to 'yProfiles' - that's how they are displayed in the plot.

src/shared/components/lazyMobXTable/styles.scss
oplantalech: Agreed, this shouldn't be here, and it is not necessary, so I've removed it.

src/pages/resultsView/ResultsViewPageStore.ts
oplantalech: I don't exactly understand the question: do you want me to comment, e.g., for `geneticEntityId: string|number` number for genes and string for gene sets?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: I've changed all `String()` to `.toString()` and the `Number()` that were refering to a string to `parseFloat()`.

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: @pieterlukasse @adamabeshouse Now the API has `profileA` for what it was the `subjectProfile` (the profile of the queried genetic entity) and `profileB` for the `queryProfile`. Here (frontend), `profileA` is `profileX` and `profileB` is `profileY`, in accordance with the position of the profiles in the plots.

adamabeshouse: I think then we should go with X and Y it makes the code a *lot* more readable

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: yeah exactly, like "entrezGeneId in gene, geneSetId for geneset"

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
adamabeshouse: don't you know by this point in the code what type of data you're looking at? and can cast as a string if necessary

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
adamabeshouse: ah okay. can you add a comment about that?

src/pages/resultsView/coExpression/CoExpressionPlot.tsx
```diff
@@ -28,11 +28,12 @@ import {scatterPlotSize} from "../../../shared/components/plots/PlotUtils";
...
 export interface ICoExpressionPlotProps {
-    xAxisGene:GeneId;
-    yAxisGene:GeneId;
+    xAxisGene:GeneticEntity;
+    yAxisGene:MolecularProfileCorrelation;
     data:CoExpressionPlotData[];
-    coExpression:CoExpression;
-    molecularProfile:MolecularProfile;
+    coExpression:MolecularProfileCorrelation;
+    queryMolecularProfile:MolecularProfile;
```
adamabeshouse: x/y please

src/pages/resultsView/coExpression/CoExpressionPlot.tsx
```diff
@@ -28,11 +28,12 @@ import {scatterPlotSize} from "../../../shared/components/plots/PlotUtils";
 type GeneId = { hugoGeneSymbol:string, entrezGeneId: number, cytoband:string};
 
 export interface ICoExpressionPlotProps {
-    xAxisGene:GeneId;
-    yAxisGene:GeneId;
+    xAxisGene:GeneticEntity;
```
adamabeshouse: xAxisGeneticEntity since its not always a gene

jjgao:
Thanks, @oplantalech. This feature can be very useful. I'll comment base on the screenshots above and would be happy to test once the demo site is up.

- Are we support all profiles with continuous data, ie, mRNA, protein, cna, methylation, and gene sets?
- Can we also support continuous clinical data?
- I am not sure if the mean of "Subject Profile" and "Query Profile" is clear. Maybe we can turn the two drop-downs into a sentence? "Looking for genes in [drop-down of Query Profile] that are correlated with SOX9 in [drop-down of Subject Profile]."

cc'ing @cBioPortal/product

jjgao:
@oplantalech I am testing on https://cbioportal-test.thehyve.net:8081/cbioportal/. 

- I think it's too much white space above the table and chart. 
- Can we have the two dropdown in the same line? 
- I like that you changed "Subject Profile" to "EGFR Profile", but maybe we can even make it more clear by change to "EGFR expression in".
- And we can change "Query Profile" (Query is confusing b/c we used Query on the page level) to "Find EGFR-correlated genes in".


image

https://bit.ly/2HgG9M3

pieterlukasse:
hi @jjgao regarding clinical attributes: we did discuss this on October 12 (coordination meeting). From what I remember (and from notes there), the co-expression tab was not really appropriate for the clinical attribute use case when starting with a clinical attribute since all tabs in this page are separated by query gene. I like the clinical attributes idea, but I think a different UI would be needed. Let's discuss again. Anyway, I think the current implementation without clinical attributes is already an improvement and could be merged as such.

pieterlukasse:
I like @jjgao sentence idea :+1: , i.e.:
 "Find correlated genes in [drop-down of Query Profile] that are correlated with SOX9 in [drop-down of Subject Profile]."

oplantalech:
I've updated the demo and the screenshots.

jjgao:
@pieterlukasse agreed that clinical data integration can be done later.

jjgao:
@pieterlukasse thanks for updating the pr. I'm uploading the screenshot you slacked me.

image

jjgao:
@oplantalech @pieterlukasse I am not sure if I like `item`, but couldn't think of a better one.

@tmazor @schultzn @inodb comments?

pieterlukasse:
@jjgao maybe something like:

"Find in [mrna expression ...] what correlates to SOX9 in [rppa ...]"

jjgao:
but we also have protein and we could also do methylation.

How about: "Find genes" when there is no gene sets, and "Find genes/gene sets" when there are gene sets?

schultzn:
I like that idea.

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
adamabeshouse: It seems like you could avoid this loop if xGeneticEntityId is a string.

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
```diff
@@ -54,16 +55,23 @@ export function computePlotData(
...
             targetData[datum.uniqueSampleKey].push(datum);
             addSampleInfo(datum);
         }
     }
+
     for (const datum of molecularData) {
-        const targetData = (datum.entrezGeneId === xEntrezGeneId ? xData : yData);
-        targetData[datum.uniqueSampleKey] = datum;
-        addSampleInfo(datum);
+        if ((datum as NumericGeneMolecularData).entrezGeneId !== undefined) {
```
adamabeshouse: it would be a lot more efficient (and readable IMO) to check once outside the loop whether this is a geneset or gene situation, rather than checking once per datum in the loop.

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
```diff
@@ -73,16 +81,16 @@ export function computePlotData(
...
+            const yVal = Number(yDatum.value);
             if (!isNaN(xVal) && !isNaN(yVal)) {
                 ret.push({
                     x: xVal,
                     y: yVal,
                     mutationsX: dispMut(xMutations[uniqueSampleKey] || []),
                     mutationsY: dispMut(yMutations[uniqueSampleKey] || []),
-                    profiledX: isProfiled(uniqueSampleKey, studyId, xHugoGeneSymbol, coverageInformation, studyToMutationMolecularProfile),
-                    profiledY: isProfiled(uniqueSampleKey, studyId, yHugoGeneSymbol, coverageInformation, studyToMutationMolecularProfile),
+                    profiledX: isProfiled(uniqueSampleKey, studyId, xGeneticEntityName, coverageInformation, studyToMutationMolecularProfile),
```
adamabeshouse: coverageInformation only works for genes, not for genesets, so profiledX/Y should probably always be true if X/Y is a geneset, rather than false which is what this will give I'm pretty sure.

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -284,17 +294,31 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
-                <CoExpressionTable
-                    referenceGene={this.props.gene}
-                    dataStore={this.dataStore}
-                    tableMode={this.dataStore.tableMode}
-                    onSelectTableMode={this.onSelectTableMode}
-                />
-                {this.requestAllDataButton}
-            </div>
-        );
+        if (this.props.profileY.molecularAlterationType !== 'GENESET_SCORE'){
```
adamabeshouse: AlterationTypeConstants.GENESET_SCORE - lets always use constants rather than "magic strings", because you can't catch magic string typos with typescript

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -262,7 +272,7 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
     }
 
     @computed get showLogScaleControls() {
-        const profileId = this.props.molecularProfile.molecularProfileId.toLowerCase();
+        const profileId = this.props.profileY.molecularProfileId.toLowerCase();
```
adamabeshouse: why does this only depend on the Y profile?

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -225,11 +235,11 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
             }
 
             const promises = this.getPlotDataPromises(this.highlightedCoExpression);
-            let numericGeneMolecularData:NumericGeneMolecularData[] = [];
+            let numericGeneMolecularData:NumericGeneMolecularData[]|GenesetMolecularData[] = [];
             if (promises.molecularX && promises.molecularX.isComplete)
-                numericGeneMolecularData = numericGeneMolecularData.concat(promises.molecularX.result!);
+                numericGeneMolecularData = (numericGeneMolecularData as any).concat(promises.molecularX.result!);
```
adamabeshouse: small nitpick but in this case I would do `as any[]`

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -225,11 +235,11 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
             if (promises.molecularY && promises.molecularY.isComplete)
-                numericGeneMolecularData = numericGeneMolecularData.concat(promises.molecularY.result!);
+                numericGeneMolecularData = (numericGeneMolecularData as any).concat(promises.molecularY.result!);
```
adamabeshouse: `any[]`

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -160,31 +162,39 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
+                    molecularProfileId: this.props.profileY.molecularProfileId
             });
         }
 
         if (this.props.mutationCache) {
-            ret.mutationX = this.props.mutationCache.get({entrezGeneId: this.props.gene.entrezGeneId});
+            ret.mutationX = this.props.mutationCache.get({entrezGeneId: Number(this.props.geneticEntity.geneticEntityId)});
             if (yAxisCoExpression) {
-                ret.mutationY = this.props.mutationCache.get({entrezGeneId: yAxisCoExpression.entrezGeneId});
+                ret.mutationY = this.props.mutationCache.get({entrezGeneId: parseFloat(yAxisCoExpression.geneticEntityId)});
```
adamabeshouse: entrezGeneId is an integer, so it should be `parseInt(..., 10)` - the 10 is important

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -160,31 +162,39 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
+                molecularProfileId: this.props.profileY.molecularProfileId
+            }) : this.props.numericGeneMolecularDataCache.get({
+                    entrezGeneId: parseFloat(yAxisCoExpression.geneticEntityId),
+                    molecularProfileId: this.props.profileY.molecularProfileId
             });
         }
 
         if (this.props.mutationCache) {
-            ret.mutationX = this.props.mutationCache.get({entrezGeneId: this.props.gene.entrezGeneId});
+            ret.mutationX = this.props.mutationCache.get({entrezGeneId: Number(this.props.geneticEntity.geneticEntityId)});
```
adamabeshouse: `parseInt(..., 10)`

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -160,31 +162,39 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
-            molecularY:MobxPromise<NumericGeneMolecularData[]>|undefined,
+            molecularX:MobxPromise<NumericGeneMolecularData[]|GenesetMolecularData[]>,
+            molecularY:MobxPromise<NumericGeneMolecularData[]|GenesetMolecularData[]>|undefined,
             mutationX:MobxPromise<Mutation[]>|undefined,
             mutationY:MobxPromise<Mutation[]>|undefined
         } = {
-            molecularX: this.props.numericGeneMolecularDataCache.get({
-                entrezGeneId: this.props.gene.entrezGeneId,
-                molecularProfileId: this.props.molecularProfile.molecularProfileId
+            molecularX: this.props.profileX.molecularAlterationType === 'GENESET_SCORE' ?
```
adamabeshouse: AlterationTypeConstants.GENESET_SCORE

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -160,31 +162,39 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
 
     private getPlotDataPromises(yAxisCoExpression?:CoExpressionWithQ) {
         const ret:{
-            molecularX:MobxPromise<NumericGeneMolecularData[]>,
-            molecularY:MobxPromise<NumericGeneMolecularData[]>|undefined,
+            molecularX:MobxPromise<NumericGeneMolecularData[]|GenesetMolecularData[]>,
```
adamabeshouse: as far as I understand, the type should really be `MobxPromise<NumericGeneMolecularData[]> | MobxPromise<GenesetMolecularData[]>`. and for molecularY too

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2126,6 +2145,23 @@ export class ResultsViewPageStore {
         }
     });
 
+    @computed get geneticEntities():GeneticEntity[]|undefined {
+        if (this.genes.isComplete && this.genesets.isComplete) {
+            let res: GeneticEntity[] = [];
+            for (const gene of this.genes.result) {
+                res.push({geneticEntityName: gene.hugoGeneSymbol, geneticEntityType: "gene",
```
adamabeshouse: always enums are better than strings, enums keep things clear, organized, and protect against typos. so lets do `enum GeneticEntityType { GENE, GENESET}` and use that for GeneticEntity.geneticEntityType

src/pages/resultsView/coExpression/CoExpressionTabUtils.ts
```diff
@@ -28,17 +28,60 @@ export function filterAndSortProfiles(profiles:MolecularProfile[]) {
...
+    // sort rna seq to the top
+    return _.sortBy(profs, profile=>(profile.molecularProfileId.toLowerCase().indexOf("rna_seq") > -1) ? 0 : 1);
+}
+
+export function filterAndSortProfilesY(profiles:MolecularProfile[]) {
+    const allProfs = getGeneProfiles(profiles).concat(getGenesetProfiles(profiles));
+    return sortProfiles(allProfs);
+}
+
+export function filterAndSortProfilesX(geneticEntity:"gene"|"geneset", profiles:MolecularProfile[]) {
```
adamabeshouse: for example,`GeneticEntityType` enum should be type of `geneticEntity` parameter (probably rename the parameter to `geneticEntityType` too

src/pages/resultsView/coExpression/CoExpressionTabUtils.ts
```diff
@@ -28,17 +28,60 @@ export function filterAndSortProfiles(profiles:MolecularProfile[]) {
...
+
+            const profileId = profile.molecularProfileId.toLowerCase();
+            good = profileId.indexOf("gsva_pvalues") === -1;
         }
+        return good;
     });
     return profs;
+}
+
+export function sortProfiles(profs: MolecularProfile[]) {
```
adamabeshouse: since this is single purpose, and it's being used for that single purpose according to the comments, I would just name this what it does: `sortRnaSeqProfilesToTop`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -242,6 +243,14 @@ export interface IQueriedMergedTrackCaseData {
     mergedTrackOqlList?: IQueriedCaseData<object>[];
 }
 
+export type GeneticEntity = {
+    geneticEntityName: string,
+    geneticEntityType: "gene"|"geneset",
```
adamabeshouse: `GeneticEntityType` enum

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -242,6 +243,14 @@ export interface IQueriedMergedTrackCaseData {
     mergedTrackOqlList?: IQueriedCaseData<object>[];
 }
 
+export type GeneticEntity = {
+    geneticEntityName: string,
```
adamabeshouse: please add comment that says `// hugo gene symbol for gene, gene set name for geneset`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -748,9 +757,19 @@ export class ResultsViewPageStore {
         }
     });
 
-    readonly coexpressionTabMolecularProfiles = remoteData<MolecularProfile[]>({
+    readonly coexpressionTabMolecularProfilesXGene = remoteData<MolecularProfile[]>({
+        await:()=>[this.molecularProfilesWithData],
+        invoke:()=>Promise.resolve(filterAndSortProfilesX("gene", this.molecularProfilesWithData.result!))
```
adamabeshouse: why do you have a separate function for filter and sort x axis profiles vs y profiles?

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -263,12 +262,16 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
             {
                 id:ResultsViewTab.COEXPRESSION,
                 hide:()=>{
-                    if (!this.resultsViewPageStore.isThereDataForCoExpressionTab.isComplete ||
```
adamabeshouse: why have you eliminated this condition?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -15,34 +16,88 @@ import {ICoExpressionPlotProps} from "./CoExpressionPlot";
...
+export class CoExpressionCache extends MobxPromiseCache<{profileX: MolecularProfile, 
+    profileY: MolecularProfile, geneticEntityId: string,
+    geneticEntityType: string, allData:boolean}, CoExpression[]> {}
 
 @observer
 export default class CoExpressionTab extends React.Component<ICoExpressionTabProps, {}> {
-    @observable _selectedMolecularProfile:MolecularProfile|undefined; // only undefined initially, until molecular profiles downloaded
-    @observable _selectedEntrezGeneId:number | undefined; // only undefined initially, until genes downloaded
+    @observable _selectedProfileXGene:MolecularProfile|undefined; // only undefined initially, until molecular profiles downloaded
+    @observable _selectedProfileXGeneset:MolecularProfile|undefined; // only undefined initially, until molecular profiles downloaded
```
adamabeshouse: can you ever select a gene profile and a geneset profile at the same time? if not then these shouldnt be handled as two separate things.

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -15,34 +16,88 @@ import {ICoExpressionPlotProps} from "./CoExpressionPlot";
...
+    @observable _selectedProfileY:MolecularProfile|undefined; // only undefined initially, until molecular profiles downloaded
+    @observable _selectedGeneticEntityId:string | undefined; // only undefined initially, until genes and gene sets downloaded
+    @observable _geneticEntities:GeneticEntity[] | undefined; // only undefined initially, until genes and gene sets downloaded
```
adamabeshouse: is `_geneticEntities` used? i dont see it

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
oplantalech: No, because we need to check explicitly if it is a particular Entrez Gene Id. It is not about `string` or `number` here (all `datum.entrezGeneId` are of type `number`, and here we are distinguishing if the mutation is coming from the gene in X or the gene in Y).

adamabeshouse: my point is that if xGeneticEntity is a gene, then wouldnt xGeneticEntityId be a number (the entrez gene id)? so if xGeneticEntityId is a string, then you know it is a geneset id and it seems like this whole loop is irrelevant, or redundant.

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -262,18 +276,21 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
     }
 
     @computed get showLogScaleControls() {
-        const profileId = this.props.molecularProfile.molecularProfileId.toLowerCase();
+        const profileXId = this.props.profileX.molecularProfileId.toLowerCase();
+        const profileYId = this.props.profileY.molecularProfileId.toLowerCase();
         return (
-            (profileId.indexOf("rna_seq") > -1) &&
-            (profileId.indexOf("zscore") === -1)
+            ((profileXId.indexOf("rna_seq") > -1) &&
```
adamabeshouse: we have a helper method that does this called `logScalePossible`

src/pages/resultsView/coExpression/CoExpressionTab.tsx
adamabeshouse: but there is only one profile ever selected, so why should you store them separately?

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
```diff
@@ -54,16 +56,25 @@ export function computePlotData(
...
             targetData[datum.uniqueSampleKey].push(datum);
             addSampleInfo(datum);
         }
     }
-    for (const datum of molecularData) {
-        const targetData = (datum.entrezGeneId === xEntrezGeneId ? xData : yData);
-        targetData[datum.uniqueSampleKey] = datum;
-        addSampleInfo(datum);
+
+    if ((molecularData[0] as NumericGeneMolecularData).entrezGeneId !== undefined) { //Check if molecularData contains gene or gene set data
```
adamabeshouse: error if molecularData is empty

src/pages/resultsView/coExpression/CoExpressionTabUtils.ts
```diff
@@ -28,17 +28,60 @@ export function filterAndSortProfiles(profiles:MolecularProfile[]) {
...
+
+export function filterAndSortProfilesX(geneticEntityType:GeneticEntityType, profiles:MolecularProfile[]) {
+    if (geneticEntityType === GeneticEntityType.GENE) {
+        const profs = getGeneProfiles(profiles);
+        return sortRnaSeqProfilesToTop(profs);
+    } else {
+        return getGenesetProfiles(profiles);
+    }
+}
+
```
adamabeshouse: theres a lot of code duplication in this file. could you please try to DRY this up? that would have the added benefit of making the logical dependencies between all of these methods more clear.

the names of the functions could also be clearer - say what they do. not just `filterAndSortProfilesX`, but `filterAndSortProfilesForOneGeneticEntityType` or something like that

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -15,34 +16,87 @@ import {ICoExpressionPlotProps} from "./CoExpressionPlot";
...
+
+    @computed get selectedGeneticEntityName(): string | undefined {
+        if (this.props.store.geneticEntities !== undefined) {
+            if (this._selectedGeneticEntityId !== undefined) {
+                for (const geneticEntity of this.props.store.geneticEntities) {
+                    if (this._selectedGeneticEntityId === geneticEntity.geneticEntityId.toString()) {
+                        return geneticEntity.geneticEntityName;
+                    }
+                }
+                return this._selectedGeneticEntityName; //Expected not to reach this code, if so, it will return undefined
```
adamabeshouse: if it's not set then why hold onto the variable just for it to be `undefined`? just return `undefined`

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -15,34 +16,87 @@ import {ICoExpressionPlotProps} from "./CoExpressionPlot";
...
-            return this._selectedMolecularProfile;
+            return this._selectedProfileXGene;
         }
     }
 
-    @computed get selectedEntrezGeneId():number | undefined {
-        if (this._selectedEntrezGeneId === undefined && this.props.store.genes.isComplete &&
-                this.props.store.genes.result.length > 0) {
-            return this.props.store.genes.result[0].entrezGeneId;
+    @computed get selectedProfileXGeneset():MolecularProfile|undefined {
```
adamabeshouse: like Ive said in other places, Im really not convinced we need to handle these cases separately - it's always a MolecularProfile, and always exactly one of them is used. logically it occupies one space, not two

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
oplantalech: Here you know that `xGeneticEntity` is a gene (for genesets, `mutationData.length` is 0), and what you are trying to do here is splitting which mutations correspond to the `xGeneticEntity` and which mutations correspond to the genetic entities from `y`. This is important when `xGeneticEntity` is a gene and the `yProfile` is not GSVA (so comparing against genes). Of course, if we are comparing a gene `xGeneticEntity` against GSVA, `datum.entrezGeneId === xGeneticEntityId` will always be `true` so all mutations will go to `xMutations`, and a similar thing will also happen if we are comparing a gene set `xGeneticEntity` against a "genetic" profile like mRNA (`datum.entrezGeneId === xGeneticEntityId` will always be `false` so all mutations will go to `yMutations`).

oplantalech: By loop do you refer to `for (const datum of mutationData)`? If so, I get your point now

src/pages/resultsView/coExpression/CoExpressionViz.tsx
oplantalech: `logScalePossible` is from the Plots tab and requires an element of type `AxisMenuSelection`, I don't think it can be reused here.

Sjoerd-van-Hagen:
@adamabeshouse Hi Adam, could you review Oleguer's changes based on your comments?

src/pages/resultsView/coExpression/CoExpressionPlot.tsx
```diff
@@ -1,38 +1,25 @@
...
-
 export interface ICoExpressionPlotProps {
-    xAxisGene:GeneId;
-    yAxisGene:GeneId;
+    xAxisGeneticEntity:GeneticEntity;
+    yAxisGeneticEntity:CoExpressionWithQ;
     data:CoExpressionPlotData[];
-    coExpression:CoExpression;
-    molecularProfile:MolecularProfile;
+    coExpression:CoExpressionWithQ;
```
adamabeshouse: It seems like the `coExpression` prop is redundant/unused

adamabeshouse: I also think that yAxisGeneticEntity should be typed as its used: GeneticEntity

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,48 +1,88 @@
...
+import {IDataQueryFilter} from "shared/lib/StoreUtils";
 
 export interface ICoExpressionTabProps {
     store:ResultsViewPageStore;
 }
 
-export class CoExpressionCache extends MobxPromiseCache<{entrezGeneId:number, molecularProfile:MolecularProfile, allData:boolean}, CoExpression[]> {};
+export class CoExpressionCache extends MobxPromiseCache<{profileX: MolecularProfile, 
+    profileY: MolecularProfile, geneticEntityId: string,
+    geneticEntityType: string, allData:boolean}, CoExpression[]> {}
```
adamabeshouse: `geneticEntityType:GeneticEntityType`

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,55 +134,120 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
+                const profiledSampleCount = this.props.store.molecularProfileIdToProfiledSampleCount.result![profile.molecularProfileId];
+                return {
+                    label: `${profile.name} (${profiledSampleCount} sample${profiledSampleCount !== 1 ? "s" : ""})`,
+                    value: profile.molecularProfileId
+                };
+            });
+        }
+    }
+
+    @computed get yOptions() {
```
adamabeshouse: please turn all @computeds that wait for promises to be complete into remoteDatas - @alisman and I have discussed this and decided its the best solution even though sometimes it adds a bit of boilerplate

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,55 +134,120 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
         );
     }
 
+    @computed get xOptions() {
```
adamabeshouse: I think xOptions is too vague a name - what kind of options are they? `xProfileOptions` would be better

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -68,19 +108,23 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
-            return (option:any)=>{};
+            this._selectedProfileX = option.value;
         }
     }
 
     @bind
-    private onSelectGene(entrezGeneId:string) {
-        this._selectedEntrezGeneId = parseInt(entrezGeneId, 10);
+    public onSelectProfileY(option:any) {
+        if (this.props.store.molecularProfileIdToMolecularProfile.isComplete) {
```
adamabeshouse: theres no need to wait for this promise to be ready - you're not using it's result on the inside of the if statement

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -68,19 +108,23 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
         };
     }
 
-    @computed public get onSelectDataSet() {
+    @bind
+    public onSelectProfileX(option:any) {
         if (this.props.store.molecularProfileIdToMolecularProfile.isComplete) {
```
adamabeshouse: theres no longer a need to wait for this promise to be ready - you're not using it's result on the inside of the if statement

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,48 +1,88 @@
...
-    @observable _selectedEntrezGeneId:number | undefined; // only undefined initially, until genes downloaded
+    @observable _selectedProfileX:string|undefined; // only undefined initially, until molecular profiles downloaded
+    @observable _selectedProfileY:string|undefined; // only undefined initially, until molecular profiles downloaded
+    @observable _selectedGeneticEntityId:string|undefined; // only undefined initially, until genes and gene sets downloaded
 
-    @computed get selectedMolecularProfile():MolecularProfile|undefined {
-        if (!this._selectedMolecularProfile && this.props.store.coexpressionTabMolecularProfiles.isComplete
-            && this.props.store.coexpressionTabMolecularProfiles.result.length > 0) {
-            return this.props.store.coexpressionTabMolecularProfiles.result[0];
+    @computed get selectedProfileX():string|undefined {
```
adamabeshouse: all of these @computed wrappers around `_selectedSomething`s should be `remoteData`

src/pages/resultsView/coExpression/CoExpressionViz.tsx
adamabeshouse: good point. but I really don't want to duplicate this logic again. could you please add a util function somewhere, might as well just put it in PlotsTabUtils, like
```
    export function logScalePossibleForProfile(profileId:string) {
        return !(/zscore/i.test(profileId)) &&
            /rna_seq/i.test(profileId);
    }
```
And use that here, and in logScalePossible so it would be `...&& logScalePossibleForProfile(axisSelection.dataSourceId)`?

src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -262,18 +276,21 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
...
+            (profileXId.indexOf("zscore") === -1)) ||
+            ((profileYId.indexOf("rna_seq") > -1) &&
+            (profileYId.indexOf("zscore") === -1))
         );
     }
 
     private get requestAllDataButton() {
         if (this.plotData.isComplete && !this.dataStore.allData.length && !this.allDataRequested) {
```
adamabeshouse: please leave all the rendering functions that depend on `somepromise.isComplete` alone (as you've done) - it's just the @computed's that we want to change

src/pages/resultsView/coExpression/CoExpressionVizUtils.ts
adamabeshouse: yes that's what I mean - there's no point going through this loop if you already know that xGeneticEntityId is not an entrezGeneId. for a large query this could cause a significant slowdown. can you please add logic here to skip the loop in this case?

src/pages/resultsView/coExpression/CoExpressionTabUtils.ts
```diff
@@ -28,17 +33,22 @@ export function filterAndSortProfiles(profiles:MolecularProfile[]) {
...
-        } else if (rnaSeqA) {
-            return -1;
-        } else {
-            return 1;
+
+    // sort rna seq to the top
+    return sortRnaSeqProfilesToTop(profs);
+}
+
+export function getGenesetProfiles(profiles:MolecularProfile[]) {
```
adamabeshouse: GREAT work revising your changes on this file! thank you

src/pages/resultsView/coExpression/CoExpressionTab.tsx
adamabeshouse: sorry I missed that until now

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,50 +1,108 @@
...
+    @observable _selectedProfileY:string|undefined; // only undefined initially, until molecular profiles downloaded
+    @observable _selectedGeneticEntityId:string|undefined; // only undefined initially, until genes and gene sets downloaded
 
-    @computed get selectedMolecularProfile():MolecularProfile|undefined {
-        if (!this._selectedMolecularProfile && this.props.store.coexpressionTabMolecularProfiles.isComplete
-            && this.props.store.coexpressionTabMolecularProfiles.result.length > 0) {
-            return this.props.store.coexpressionTabMolecularProfiles.result[0];
-        } else {
-            return this._selectedMolecularProfile;
+    readonly selectedProfileX = remoteData<string|undefined>({
```
adamabeshouse: perfect, thank you

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,50 +1,108 @@
...
+            this.props.store.geneticEntities
+        ],
+        invoke: () => {
+            if (this._selectedGeneticEntityId !== undefined) {
+                for (const geneticEntity of this.props.store.geneticEntities.result!) {
+                    if (this._selectedGeneticEntityId === geneticEntity.geneticEntityId.toString()) {
+                        return Promise.resolve(geneticEntity.geneticEntityName);
+                    }
+                }
+                return Promise.resolve(undefined); //Expected not to reach this code, if so, it will return undefined
```
adamabeshouse: would this indicate something has gone wrong, e.g. with data? if so maybe we should actually throw an error rather than return undefined. @alisman what do you think?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -90,55 +148,131 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
...
+            this.isSelectedGeneticEntityAGeneSet
+        ],
+        invoke: () => {
+            let filteredProfiles: MolecularProfile[] = [];
+            if (this.isSelectedGeneticEntityAGeneSet.result!) {
+                filteredProfiles = getGenesetProfiles(this.props.store.coexpressionTabMolecularProfiles.result!);
+            } else {
+                filteredProfiles = filterAndSortProfiles(this.props.store.coexpressionTabMolecularProfiles.result!);
+            }
+            return Promise.resolve(filteredProfiles.map(profile=>{
```
adamabeshouse: this part is duplicated between x and y profile options, can you pull it out into a helper method in CoExpressionUtils, e.g. `export function getProfileOptions(profiles:MolecularProfile[], profileToProfiledSampleCount:{[id:string]:number}){`

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: Yes, it would mean that the promise `this.props.store.geneticEntities` has failed. I'm not sure if I need to throw an error here or this `return` can be just removed. @adamabeshouse @alisman what's your opinion on this?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,50 +1,108 @@
...
-    }
+    });
+
+    readonly selectedGeneticEntityId = remoteData<string>({
+        await: ()=>[
+            this.props.store.geneticEntities
+        ],
+        invoke: () => {
+            if (!this._selectedGeneticEntityId) {
+                return Promise.resolve(this.props.store.geneticEntities.result![0].geneticEntityId.toString());
```
alisman: this assumes there will ALWAYS be geneticEntites returned.  Is that true?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
alisman: it seems to me the promise selectedGeneticEntityId already handles the case where there's no _seletedGeneticIdentityId be resolving with the first one, so it would seem that the selected name is just the name property of whatever corresponding entity is return by that promise.   So, this promise seems redundant.   Also, see my comment about that promise above.

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -1,50 +1,108 @@
...
+            if (!this._selectedProfileY && this.yProfileOptions.result!.length > 0) {
+                return Promise.resolve(this.yProfileOptions.result![0].value);
+            } else {
+                return Promise.resolve(this._selectedProfileY);
+            }
         }
-    }
+    });
+
+    readonly selectedGeneticEntityId = remoteData<string>({
```
alisman: why not just have a promise that returns the selectedEntity and consumers can take whatever property they need of that object?   Is it true that the _selectedGeneticEntityId will always be represented in the geneticEntities collection?

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -138,7 +122,13 @@ export default class CoExpressionTab extends React.Component<ICoExpressionTabPro
 
     @bind
     private onSelectGeneticEntity(geneticEntityId:string) {
-        this._selectedGeneticEntityId = geneticEntityId;
+        if (this.props.store.geneticEntities.isComplete) {
```
alisman: you might consider throwing an error here because the UI should never allow a user to selected a geneticEntity before geneticEntites are complete.  but not critical

alisman:
this one needs rebase on rc and use of yarn.  should be no package-lock.json

alisman:
@jjgao this PR involves a fair amount of refactoring of the co-expression tab to generalize for genesets and so should be tested carefully by our team to make sure existing co-expression functionality hasn't regressed

src/pages/resultsView/coExpression/CoExpressionTab.tsx
oplantalech: @alisman Done

adamabeshouse: its better to throw an error object so that the stack trace is available which i believe its not if you just throw a string

adamabeshouse: its better to throw an error object so that the stack trace is available which i believe its not if you just throw a string

Sjoerd-van-Hagen:
@jjgao Hi JJ, when can we expect this PR to be merged, assuming there are no big issues?

alisman:
@pieterlukasse @oplantalech we'll merge this today when @adamabeshouse gets in.  sorry for the delay.

Sjoerd-van-Hagen:
@alisman Do not forget to merge the backend PR too https://github.com/cBioPortal/cbioportal/pull/5354

alisman:
@oplantalech there is a conflict on the backend PR so couldn't merge.  as you guys pointed out, we need to merge that first so that the frontend api sync test will pass.

alisman:
Oleg, I deployed backend changes to rc.  Now we have to update the js client so that api sync test will pass.   It's the command "yarn run updateAPI" which should regenerate the api client based on swagger doc.  Let me know if you have questions.

Sjoerd-van-Hagen:
@alisman @adamabeshouse Thanks for your review and feedback!

PR #1792, opened 2018-12-04
cbioObserver decorator - applies observer and auto disposes reactions on unmount

adamabeshouse:
do not merge: lets confirm the interface, then ill try to use it

### DIFF
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -1,3 +1,4 @@
+import cbioObserver from "shared/lib/decorators/cbioObserver/cbioObserver";
 import * as _ from 'lodash';
 import * as React from 'react';
 import UnsupportedBrowserModal from "shared/components/unsupportedBrowserModal/UnsupportedBrowserModal";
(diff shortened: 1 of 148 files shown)


alisman:
this looks fine to me.

adamabeshouse:
remoteData doesnt need a disposal because we took out the reactions from it - now that theyre computed its GC-able

adamabeshouse:
reactiondisposer will get any mobx reaction, autorun as well

adamabeshouse:
@alisman on further thought I dont think this is the best interface... bc the user has to create and assign member properties for each disposer

What I think would be better is to have the decorator create functions on the component: this.addReaction and this.addAutorun, which add the results to an internal list which componentWillUnmount goes through

alisman:
"bc the user has to create and assign member properties for each disposer"  not sure i understand this. 
it seems reasonable to me that reactions should be made members of the class so they can be disposed.    lets remember that we should be using reactions very sparingly in the first place.  

also, if you wanted to add these extra methods, i think we would really have to create a subclass of React.Component to get the typing correct.  i would rather not go down this route.

adamabeshouse:
@alisman sounds good.

are there any other things we want to roll into cbioObserver?

PR #1793, opened 2018-12-05
Study view fixes

zhx828:
This solves
cBioPortal/cbioportal#5349
cBioPortal/cbioportal#5273
cBioPortal/cbioportal#5310

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -269,6 +269,10 @@ export class StudyViewPageStore {
 
     @observable private chartsType = observable.map<ChartType>();
 
+    // We highlight newly added charts by comparing the updated time.
+    // Chart visibility is computed and does not carry any status to indicate a chart is newly added except creating
+    // a new observable map to record the time. The highlight style will no be updated until user interects with the
+    // chart.
     @autobind
     isChartHighlighted(uniqueKey: string): boolean {
         // Highlight survival analysis chart
@@ -279,7 +283,7 @@ export class StudyViewPageStore {
         const date = this._chartToBeHighlighted.get(uniqueKey);
         if (date === undefined) {
             return false;
-        } else if (new Date().getTime() - date.getTime() < 5000) {
+        } else if (new Date().getTime() - date.getTime() < STUDY_VIEW_CONFIG.thresholds.chartHighlight) {
             return true;
         } else {
             return false;
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -255,6 +255,8 @@ export class StudyViewPageStore {
     // ChartContainer should be modified, instead of accepting a promise, it should accept data and loading state.
     @observable private _chartVisibility = observable.map<boolean>();
 
+    @observable private _chartToBeHighlighted = observable.map<Date>();
+
     @observable geneQueryStr: string;
 
     @observable private geneQueries: SingleGeneQuery[] = [];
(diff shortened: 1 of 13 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -267,6 +269,23 @@ export class StudyViewPageStore {
...
+    isChartHighlighted(uniqueKey: string): boolean {
+        // Highlight survival analysis chart
+        if (this.analysisGroupsSettings.clinicalAttribute &&
+            (getClinicalAttributeUniqueKey(this.analysisGroupsSettings.clinicalAttribute) === uniqueKey)) {
+            return true;
+        }
+        const date = this._chartToBeHighlighted.get(uniqueKey);
+        if (date === undefined) {
+            return false;
+        } else if (new Date().getTime() - date.getTime() < 5000) {
```
onursumer: I think this time sensitive condition deserves a comment why we are doing it this way

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -86,9 +86,9 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
             <span>{getFrequencyStr((data.count / this.totalCount) * 100)}</span>,
         tooltip: (
             <span>Percentage of {this.props.patientAttribute ? 'patients' : 'samples'}</span>),
-        filter: (d: ClinicalDataCountWithColor, f: string, filterStringUpper: string) => {
+        filter: (d: ClinicalDataCountWithColor, f: string) => {
             let freq = getFrequencyStr((d.count / this.totalCount) * 100);
-            return (freq.indexOf(filterStringUpper) > -1)
+            return (freq.indexOf(f) > -1)
```
alisman: we now have String.includes (and Array.includes) which is a little nicer than indexOf.  you don't need to fix all these, but going forward ...

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -125,13 +125,26 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
         return _.sumBy(this.props.data, obj => obj.count);
     }
 
+    @autobind
```
alisman: not 100% sure but i don't think you need autobind on these.  you should only need autobind when a DOM callback is changing the context (this) when executing callback.    i.e. onclick={this.someCallback} unless you bind someCallback to the desired instance, it will be the clicked element.

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -125,13 +125,26 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
...
+    addAll(selectedRows: ClinicalDataCountWithColor[]) {
+        this.props.onUserSelection(selectedRows.map(row => row.value));
+    }
+
+    @autobind
+    removeAll(deselectedRows: ClinicalDataCountWithColor[]) {
+        const deselectRows = deselectedRows.map(row => row.value);
+        this.props.onUserSelection(this.props.filters.filter(filter => !_.includes(deselectRows, filter)));
```
alisman: you can use deselectRows.incldues(filter)

src/pages/studyView/table/tables.module.scss
```diff
@@ -9,7 +9,7 @@
 // The current portal is wrapped a class .cbioportal-frontend to overwrite the legacy styles
```
alisman: you can get rid of this comment

zhx828: Don't we still have `.cbioportal-frontend`?

PR #1809, opened 2018-12-07
fix back button bug in results view caused by redirect

adamabeshouse:
The bug:
(1) Go to cbioportal.org
(2) Submit a query to go to results page
(3) Wait for page to finish loading
(4) Click browser back button

We can't go back, we're stuck on the results page.

The cause of the bug was that when we go to the "results" route, without a tab specified (i.e. as opposed to  "results/oncoprint"), we auto-redirect to oncoprint or cancer types summary, depending on the query. But in the redirection, we *add* the redirected URL to history, we do not replace the original "results" route. Thus, clicking on "back" will just go back to "results" where we get redirected back to "results/oncoprint".

This PR makes it so that the redirection replaces the "results" URL in history, with e.g. "results/oncoprint", rather than adding "results/oncoprint" to the history as a new entry.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -176,8 +176,8 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
         getBrowserWindow().resultsViewPageStore = this.resultsViewPageStore;
     }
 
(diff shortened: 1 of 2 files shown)


adamabeshouse:
@alisman me and @jjgao both encountered it

PR #1810, opened 2018-12-07
Q value mutex

dippindots:
Change Mutex to use Benjamini–Hochberg to calculate q-value.


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+    await: () => [
+        this.filteredAlterations,
+        this.annotatedMolecularData,
+        this.selectedMolecularProfiles,
+        this.defaultOQLQuery,
+        this.samples,
+        this.patients
+    ],
+    invoke: () => {
+        const data = [...(this.filteredAlterations.result!), ...(this.annotatedMolecularData.result!)];
```
alisman: are parens necessary?  also, i would just use _.concat for readability

alisman: this is just a minor quibble.  i.e. lets not overuse the ... notation

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+                    accessorsInstance,
+                    defaultOQLQuery
+                )
+            );
+
+            return Promise.resolve(filteredAlterationsByOQLLine.map(
+                (oql) => ({
+                    cases: groupDataByCase(oql, samples, patients),
+                    oql,
+                    list: filterSubQueryData(
```
alisman: improve the name "list"   list of what?  alterations?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+                filterCBioPortalWebServiceDataByUnflattenedOQLLine(
+                    this.oqlQuery,
+                    data,
+                    accessorsInstance,
+                    defaultOQLQuery
+                )
+            );
+
+            return Promise.resolve(filteredAlterationsByOQLLine.map(
+                (oql) => ({
```
alisman: this param is probably an oqlLine?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+        }
+    }
+});
+
+    readonly isSampleAlteredMap = remoteData({
+        await: () => [this.filteredAlterationsDataByUnflattenedOQLLine, this.samples],
+        invoke: async() => {          
+            const result : {[x: string]: boolean[]} = {};
+            this.filteredAlterationsDataByUnflattenedOQLLine.result!.forEach((element, key) => {
+                //1: is not group(1.1, 1.2)
```
alisman: not sure what 1.1, 1.2 means.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+                if (element.list === undefined) {
+                    const notGroupedOql = element.oql as OQLLineFilterOutput<AnnotatedExtendedAlteration>;                    
+                    const sampleKeys = _.map(notGroupedOql.data, (data) => data.uniqueSampleKey);
+                    const isOnlyGene = (oql_parser.parse(this.oqlQuery)![key] as oql_parser.SingleGeneQuery).alterations === false;
```
alisman: strange that you have to cast this

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: Based on the user input query. There is two main type of query: Gene(like EGFR / EGFR: MUT) and Gene group(like ["RAF" BRAF RAF1 ARAF]). I am using 1.1 to represent is gene like EGFR, which don't need to trim the last character of the oqlString. The 1.2 represents a gene with mutation type like EGFR: MUT, which need to remove the last punctuation from the oqlString.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+                    //1.1: gene
+                    if (isOnlyGene) {
+                        result[notGroupedOql.gene] = this.samples.result.map((sample: Sample) => {
+                            return _.includes(sampleKeys, sample.uniqueSampleKey);
```
alisman: you can use sampleKeys.includes  method of native js array

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+            ));
+        }
+    }
+});
+
+    readonly isSampleAlteredMap = remoteData({
+        await: () => [this.filteredAlterationsDataByUnflattenedOQLLine, this.samples],
+        invoke: async() => {          
+            const result : {[x: string]: boolean[]} = {};
+            this.filteredAlterationsDataByUnflattenedOQLLine.result!.forEach((element, key) => {
```
alisman: there is a lot of logic here inside this invoke.  it should be unit tested, which will require that you put the function into a "pure" helper function.   what you could do is make this into this.filterAlterationDataByUnglattendOQLine.result.map(   someTestableExternalFunction(this.samples.result) );

dippindots: Ok! I will add unit tests for this function!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,6 +981,91 @@ export class ResultsViewPageStore {
...
+                            return _.includes(sampleKeys, sample.uniqueSampleKey);
+                        });
+                    }
+                }
+                //2: is group
+                else {
+                    const groupedOql = element.oql as MergedTrackLineFilterOutput<AnnotatedExtendedAlteration>;
+                    const oqlData = _.flatten(_.map(groupedOql.list, (list) => list.data));
+                    const sampleKeys = _.map(oqlData, (data) => data.uniqueSampleKey);
+                    const resultKey = groupedOql.label! === undefined ? _.map(groupedOql.list, (data) => data.gene).join(' / ') : groupedOql.label!;
```
alisman: i don't think .label! makes sense because the ! = NOT undefined, but you're testing here whether it's undefined or not

src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.tsx
```diff
@@ -41,6 +41,18 @@ export function calculateAdjustedPValue(pValue: number, count: number): number {
     return value > 1 ? 1 : value;
 }
 
+export function calculateQValue(sortedData: MutualExclusivity[]): MutualExclusivity[] {
+    const DataLength = sortedData.length;
+    for (let i = DataLength - 1; i >= 0; i--) {
```
alisman: for loop is fast but makes code hard to read and error prone.   you should use array.forEach unless there is a demonstrable performance issue.   if you are dealing with a lot of elements (e.g. > 10,000) and you can see rendering performance is suffering, then optimize loops, otherwise, code for readability

PR #1816, opened 2018-12-10
oncoprint - Show/hide germline mutations, and corresponding UI changes

adamabeshouse:


src/shared/lib/MutationUtils.ts
```diff
@@ -175,6 +176,7 @@ export function getProteinStartPositionsByRange(data: Mutation[][], start: numbe
     return _.uniq(positions);
 }
 
+const GERMLINE_REGEXP = new RegExp(MUTATION_STATUS_GERMLINE, "i");
```
alisman: seems weird that the RegExp instance is capitalized?

adamabeshouse: hmm yeah I'm not sure what would be more idiomatic. I think I went cap because its a constant and used for checking strings so its kind of like a string..

PR #1820, opened 2018-12-11
Error page for network errors

alisman:
When we encounter 400 error on network, show an app error screen like so
image

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -1,23 +1,34 @@
-import {computed, observable} from "mobx";
-import {addErrorHandler, remoteData} from "shared/api/remoteData";
+import {action, computed, observable} from "mobx";
+import {addServiceErrorHandler, remoteData} from "shared/api/remoteData";
 import {initializeAPIClients} from "./config/config";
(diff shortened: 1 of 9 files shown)


src/AppStore.ts
```diff
@@ -35,6 +46,23 @@ export class AppStore {
         }
     }
 
+    @computed get undismissedSiteErrors(){
+        const me =  _.filter(this.siteErrors.slice(), (err)=>!err.dismissed);
```
adamabeshouse: just return result, idk what `me` means

src/AppStore.ts
```diff
@@ -35,6 +46,23 @@ export class AppStore {
...
+        return me;
+    }
+
+    @computed get isErrorCondition(){
+        return this.undismissedSiteErrors.length > 0;
+    }
+
+    @action
+    public dismissErrors(){
+        this.siteErrors = this.siteErrors.map((err)=>{
```
adamabeshouse: `this.siteErrors.forEach` would be more idiomatic, since you're actually making the changes in place.

PR #1822, opened 2018-12-11
OQL filtering driver mutations

adamabeshouse:
supporting `DRIVER` `GERMLINE_DRIVER` `DRIVER_GERMLINE` `TRUNC_DRIVER` `DRIVER_MISSENSE` `INFRAME_DRIVER_GERMLINE` `DRIVER_GERMLINE_INFRAME`


adamabeshouse:
@cBioPortal/product Please see this link for example of use of DRIVER in different ways, together with GERMLINE and mutation types

http://cbioportal-frontend-pr-1822.herokuapp.com/results/oncoprint?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=ov_tcga_pub&case_set_id=ov_tcga_pub_cna_seq&data_priority=0&gene_list=BRCA1%253AGERMLINE_DRIVER%250ABRCA1%253ADRIVER_GERMLINE%250ABRCA1%253ATRUNC_DRIVER_GERMLINE%250ABRCA1%253ADRIVER%250ABRCA1%253AGERMLINE%250ABRCA1%253ATRUNC_DRIVER%250ABRCA1%253ADRIVER_MISSENSE%250ABRCA1%253ASOMATIC_DRIVER_MISSENSE%250ABRCA1%253AMISSENSE_DRIVER_GERMLINE&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=ov_tcga_pub_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ov_tcga_pub_mutations&tab_index=tab_visualize

jjgao:
Thanks, @adamabeshouse. It works great for mutations!

I think we should also apply DRIVER to other types (fusions and CNAs).

http://cbioportal-frontend-pr-1822.herokuapp.com/results/oncoprint?session_id=5c1138b7e4b024a8f3efacef

adamabeshouse:
@cBioPortal/product @jjgao that link is working now 👍 :D

tmazor:
It looks like the % altered in the header isn't working correctly - in the example @adamabeshouse linked to above, the header says 8% but one of the tracks is at 12%
image

The OncoPrint will update in realtime if I change the definition of a driver (e.g. un-click OncoKB). Is this reflected on the rest of the site? 

Also, if I do un-select OncoKB, then only CancerHotspots is selected to define drivers -- but the "BRCA1: DRIVER" track still shows one sample with a deep deletion. Isn't CancerHotspots only for mutations? How can a deep deletion be a driver according to CancerHotspots?
image

schultzn:
OncoKB annotations are currently only applied to mutations.

jjgao:
@adamabeshouse let's also support `EGFR:DRIVER_MUT` / `EGFR:MUT_DRIVER`.

http://bit.ly/2UValie

image

jjgao:
@schultzn @tmazor I am confused... @adamabeshouse's new code applies `driver` to CNAs and fusions as well.

@adamabeshouse when OncoKB is deselected, CNA driver events are still shown.

@adamabeshouse after changing the driver sources, is it possible to propagate the oncoprint change to other tabs, e.g. Mutex, Enrichments, Survival, etc?

schultzn:
Sorry, my bad - I was not aware.

adamabeshouse:
@jjgao CNA driver events are annotated regardless, as long as oncokb is available. turning oncokb off in the mutations menu only applies to mutations and fusions. should we change that?

adamabeshouse:
@jjgao @tmazor just pushed MUT_DRIVER, DRIVER_MUT, CNA_DRIVER, DRIVER_CNA :+1:

@tmazor in doing this I came across the reason why the OQL error selection stopped working - the pegjs error interface changed. I updated it and it seems to be working now.

tmazor:
Looking good! And hooray for finding that other OQL bug fix in the process :)

A few things:
1) DRIVER doesn't work with specific protein changes (e.g. EGFR:MUT=A289T_DRIVER) shows nothing. GERMLINE also doesn't seem to work with specific protein changes (see http://bit.ly/2Lq43TC)
2) Something else that may have gotten lost above - the % altered in the header isn't working correctly - in the BRCA1 example linked above, the header says 8% altered but one of the OncoPrint tracks is 12% altered.
3) I think it's important to propagate changes made to the definition of driver to the rest of the site. I can't speak to whether that should be a different PR or not, but I think it's an important thing to implement.

jjgao:
> "@jjgao CNA driver events are annotated regardless, as long as oncokb is available. turning oncokb off in the mutations menu only applies to mutations and fusions. should we change that?"

Yes, please change that and also apply to  CNAs.

> "propagating to other tabs is possible but I think that's a separate task from this PR, right?"

OK to do it as a separate task (https://github.com/cBioPortal/cbioportal/issues/5499), but maybe we can add some warning in relevant tabs before propagate?

adamabeshouse:
@jjgao @tmazor addressed all your comments, including being able to specify modifiers and mutations in any order.

but I think using same data in other tabs should be a separate task, as should turning off oncokb annotation for CNA, since they're both existing issues in production.

jjgao:
@adamabeshouse thanks! I logged the CNA issue here https://github.com/cBioPortal/cbioportal/issues/5513. Please do a follow-up fix soon.

tmazor:
Nice @adamabeshouse! But I'm having trouble with "DRIVER_MUT_GERMLINE" - seems like OQL accepts mutation type in the middle (e.g. DRIVER_TRUNC_GERMLINE) but not MUT in the middle.

Also the OQL can get so long that it overruns the box in OncoPrint:
image

adamabeshouse:
good catch, Ill make another PR

adamabeshouse:
@tmazor https://github.com/cBioPortal/cbioportal-frontend/pull/1871

the tooltip issue is already logged https://github.com/cBioPortal/cbioportal/issues/4810

tmazor:
@adamabeshouse I just tried a different query and now the header says altered in 0% which is definitely not right:
image

Does this not work with gene sets?

I don't know how to make a link that works with heroku, but the query is:
DATATYPES: DRIVER
[EGFR ERBB2]
["RAS" KRAS NRAS HRAS]
["RAF" ARAF BRAF RAF1]
["MEK" MAP2K1 MAP2K2]
in TCGA PanCan Colorectal

tmazor:
@adamabeshouse Also it's looking like other tabs don't see any alterations when DRIVER is used.

Given this OncoPrint:
image

Cancer Types Summary sees no EGFR alterations:
image

And neither does MutEx:
image

And the mutations tab just ignores OQL:
image

I haven't figured out what's happening on enrichments yet, and the download tab just doesn't seem to load:
image

adamabeshouse:
altered percent in the header is also broken because EGFRs only OQL is driver.... jesus this is messy

image

adamabeshouse:
I think theres no way around it - we have to use OQL everywhere now. I actually dont know why I didnt do it initially, I forgot that these tabs are already using OQL

adamabeshouse:
okay, I think all issues are fixed by this PR: https://github.com/cBioPortal/cbioportal-frontend/pull/1877

tmazor:
looks good to me now @adamabeshouse !

PR #1828, opened 2018-12-12
Add EllipsisTextTooltip component and use in ClinicalTable

zhx828:
Inspired by
https://stackoverflow.com/questions/21064101/understanding-offsetwidth-clientwidth-scrollwidth-and-height-respectively/21064102#21064102
https://stackoverflow.com/questions/7738117/html-text-overflow-ellipsis-detection

This solves issue https://github.com/cBioPortal/cbioportal/issues/5427

### DIFF
--- a/src/shared/components/ellipsisTextTooltip/EllipsisTextTooltip.tsx
+++ b/src/shared/components/ellipsisTextTooltip/EllipsisTextTooltip.tsx
@@ -1,0 +1,33 @@
+import * as React from 'react';
+import {observable} from 'mobx';
+import {observer} from "mobx-react";
+import styles from "./EllipsisTextTooltip.module.scss";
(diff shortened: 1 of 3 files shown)


src/shared/components/ellipsisTextTooltip/EllipsisTextTooltip.tsx
```diff
@@ -0,0 +1,33 @@
...
+export default class EllipsisTextTooltip extends React.Component<{ content: string }, {}> {
+    @observable disabled = false;
+    private contentRef: any;
+
+    constructor(props: { content: string }) {
+        super(props);
+    }
+
+    componentDidMount(): void {
+        if (this.contentRef.offsetWidth === this.contentRef.scrollWidth) {
```
alisman: is this responsive though?  i mean, if user adjusts browser width, then  ...

PR #1830, opened 2018-12-13
Add parameter to remove What's New section

oplantalech:
Add boolean property in `portal.properties` to be able to remove the "What's New" section from the right bar. Associated backend PR: https://github.com/cBioPortal/cbioportal/pull/5441

### DIFF
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -61,49 +61,53 @@ export default class RightBar extends React.Component<IRightBarProps, IRightBarS
     }
 
     private getWhatsNew() {
-        if (!_.isEmpty(AppConfig.serverConfig.skin_right_nav_whats_new_blurb)) {
-            return (
-                <div className="rightBarSection">
-                    <h3>What's New</h3>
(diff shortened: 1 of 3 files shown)


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -61,49 +61,53 @@ export default class RightBar extends React.Component<IRightBarProps, IRightBarS
...
-                            }}
-                            options={{
-                                username: 'cbioportal',
-                                height: '200',
-                                chrome: 'noheader%20nofooter',
-                            }}
-                            onLoad={() => this.setState({twitterLoading:false})}
-                        />
+        if (AppConfig.serverConfig.skin_right_nav_show_whats_new) {
+            if (!_.isEmpty(AppConfig.serverConfig.skin_right_nav_whats_new_blurb)) {
```
alisman: why not just get rid of skin_right_nav_whats_new_blurb?  Isn't this redundant?

oplantalech: @alisman No, the new property is a boolean which allows to show or not the "What's new" section, while the current property (`skin_right_nav_whats_new_blurb`) defines the content of the section.

alisman: i see.  ok, we essentially have a default (the twitter thing)

PR #1833, opened 2018-12-13
Add EllipsisTextTooltip component and use in ClinicalTable

alisman:
When text in table cells in study view overflow, we need a tooltip
This solves issue https://github.com/cBioPortal/cbioportal/issues/5427

### DIFF
--- a/src/pages/studyView/table/ClinicalTable.tsx
+++ b/src/pages/studyView/table/ClinicalTable.tsx
@@ -13,6 +13,7 @@ import {
     getFrequencyStr
 } from "../StudyViewUtils";
 import {SortDirection} from "../../../shared/components/lazyMobXTable/LazyMobXTable";
+import EllipsisTextTooltip from "../../../shared/components/ellipsisTextTooltip/EllipsisTextTooltip";
 
 export interface IClinicalTableProps {
     data: ClinicalDataCountWithColor[];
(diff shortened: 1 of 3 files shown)


src/shared/components/ellipsisTextTooltip/EllipsisTextTooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+export default class EllipsisTextTooltip extends React.Component<{ text:string },{}> {
+
+    @observable tooltipVisible = false;
+
+    el:HTMLSpanElement;
+
+    @autobind
+    @action
+    onVisibleChange(isVisible:boolean){
+        const isOverflowed = (this.el.scrollWidth - $(this.el).innerWidth()) > 1;
```
zhx828: We could use this.el. clientWidth to replace innerWidth from jquery

src/pages/studyView/table/ClinicalTable.tsx
```diff
@@ -112,7 +113,7 @@ export default class ClinicalTable extends React.Component<IClinicalTableProps,
                             <rect x="0" y="0" width="12" height="12" fill={data.color}/>
                         </g>
                     </svg>
-                    <span className={styles.ellipsisText} title={data.value}>{data.value}</span>
+                    <EllipsisTextTooltip text={data.value}></EllipsisTextTooltip>
```
adamabeshouse: self close tag

src/shared/components/ellipsisTextTooltip/EllipsisTextTooltip.tsx
adamabeshouse: I feel like isOverflowed should be updated during componentDidMount, isVisible should be updated by onVisibleChange, and then tooltipVisible should be a computed that combines isVisible and isOverflowed

PR #1834, opened 2018-12-13
use log2 odds ratio in Mutex

dippindots:
change from log to log2, add the tooltip.


alisman:
this PR is dependent on 1791, correct?

dippindots:
@alisman Yes, this is dependent on 1791.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -302,6 +305,44 @@ export function extendSamplesWithCancerType(samples:Sample[], clinicalDataForSam
 
 }
 
+export function getSampleAlteredMap(filteredAlterationData: IQueriedMergedTrackCaseData[], samples: Sample[], oqlQuery: string){
```
alisman: sorry this should go in ResultsViewPageStoreUtils

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -302,6 +305,44 @@ export function extendSamplesWithCancerType(samples:Sample[], clinicalDataForSam
...
+        return notGroupedOql.gene;
+    }
+    //gene with alteration type
+    else {
+        return notGroupedOql.oql_line.slice(0, -1);
+    }
+}
+
+export function getMultipleGeneResultKey(groupedOql: MergedTrackLineFilterOutput<AnnotatedExtendedAlteration>){
+    return groupedOql.label! === undefined ? _.map(groupedOql.list, (data) => data.gene).join(' / ') : groupedOql.label!;
```
alisman: you shouldn't need the ! if you are checking whether it's defined b/c typscript should not compain

src/pages/resultsView/ResultsViewPageStore.ts
alisman: and this one should have a unit test.

PR #1835, opened 2018-12-13
Scatter plot regression improvements

adamabeshouse:
(1) Add Pearson p-value to scatter plot, in coexpression and plots tabs
(2) Add ability to show linear regression line in plots tab scatter plots, and in coexpression tab, as seen in screenshots:

image

image


jjgao:
@adamabeshouse it's really cool! Two small comments:

- is it possible to use superscript for R^2?
- the downloaded PDF seems not working for the line and equation.

adamabeshouse:
@jjgao there seems to be some problem with the svg to pdf conversion when handling the path. Our svg to pdf conversion gives us a lot of problems. Is it okay to just take out pdf download for plots? Or can we update our svg to pdf conversion library (is it outdated, possibly bugfixes available in newer version?)

jjgao:
@adamabeshouse  good idea to update the convertor. Do you want to give it a try? https://github.com/cBioPortal/cbioportal/blob/master/core/pom.xml#L221

I am wondering if @onursumer has tried that?

adamabeshouse:
@jjgao ill open a new ticket for the pdf issues, and can we merge it as-is (pending review, and need to update screenshots)

jjgao:
@adamabeshouse that sounds good. I was also thinking about another solution: We can open a new tab for svg and the browser print to pdf is usually good.

adamabeshouse:
@jjgao do you mean showing the svg on a new cbioportal-wrapped page? bc if not, then they can just do that by opening the svg download in the browser.

src/shared/components/plots/ScatterPlot.tsx
```diff
@@ -48,14 +50,13 @@ export interface IScatterPlotProps<D extends IBaseScatterPlotData> {
 
 const DEFAULT_FONT_FAMILY = "Verdana,Arial,sans-serif";
 const CORRELATION_INFO_Y = 100; // experimentally determined
-export const LEGEND_Y = CORRELATION_INFO_Y + 30 /* approximate correlation info height */ + 30 /* top padding*/
+export const LEGEND_Y = CORRELATION_INFO_Y + 60 /* approximate correlation info height */ + 30 /* top padding*/
```
alisman: semicolon

src/shared/components/plots/ScatterPlot.tsx
```diff
@@ -329,6 +340,44 @@ export default class ScatterPlot<D extends IBaseScatterPlotData> extends React.C
...
+    @computed get regressionLineComputations() {
+        return regression.linear(this.props.data.map(
+            // perform same transformations on data for this calculation as we do for plot (i.e. log if necessary)
+            d=>([this.x(d), this.y(d)]))
+        );
+    }
+
+    private get regressionLine() {
+        if (this.props.showRegressionLine) {
+            const coeffs = this.regressionLineComputations.equation;
```
alisman: it seems like there's some actual computation here that ought to be unit tested.

alisman:
@jjgao i just want to double check that you've validated the output here.  I can see we are using a library (regression) but there is still some math written.  i've asked Adam to write a couple more unit tests.

jjgao:
@alisman I didn't check the math part. It would be a good idea to compare to another source (e.g. R or web tools) add a few unit tests.

jjgao:
@adamabeshouse I am thinking of utilizing window.print(): https://github.com/cBioPortal/cbioportal/issues/5738. Let's see what others think.

PR #1841, opened 2018-12-14
StudyView: The study summary info should wait until charts are render…

alisman:
…ed before fading in

Product team wants the summary info to show after charts are rendered, to make sure user's eye is drawn to totals

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
-import {inject, observer} from "mobx-react";
+import * as _ from 'lodash';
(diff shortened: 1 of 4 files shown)


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -152,19 +174,29 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                                     disableAddGenomicButton={this.props.routing.location.query.tab === StudyViewPageTabKeys.CLINICAL_DATA}
                                 />}
 
-                                <div className={styles.selectedInfo}>
+
+                                <Observer>
```
alisman: i should probably nest this in another component.  Observer is here so that page component will no re-render on all the chart data promises

PR #1852, opened 2018-12-18
Q value mutex

dippindots:


src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.tsx
```diff
@@ -41,6 +41,22 @@ export function calculateAdjustedPValue(pValue: number, count: number): number {
     return value > 1 ? 1 : value;
 }
 
+export function calculateQValue(sortedData: MutualExclusivity[]): MutualExclusivity[] {
```
alisman: @jjgao can you review this logic for calculating q-value?

adamabeshouse: agreed, it should be in its own file, and there should be more tests, including one with a longer input list.

I also think its crucial to link to a source in a comment.

jjgao: I realized that the q-value calcuation code was in Java for Enrichments (#5515). How about Co-exp, @adamabeshouse?

adamabeshouse: coexp q-value is also on the server.

I still think its worthwhile to put it in its own file for future use

src/shared/lib/calculation/qValuesCalculation.ts
```diff
@@ -0,0 +1,12 @@
+export function calculateQValues(sortedPValues: number[]): number[] {
```
ersinciftci: I think the file name should be qValueCalculator.ts, or maybe it should specify what kind of qValue calculation it is, such as BenjaminiHochbergFDRCalculator.ts

dippindots: Thanks, Ersin! Yes, Maybe I use "BenjaminiHochbergFDRCalculator.ts" is much better.

src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.tsx
jjgao: @adamabeshouse agreed. We need to move the code to frontend soon for Enrichments. Maybe we could do it for Co-exp when we do that too.

src/pages/resultsView/mutualExclusivity/MutualExclusivityTable.tsx
```diff
@@ -132,17 +132,17 @@ export default class MutualExclusivityTable extends React.Component<IMutualExclu
...
-        this._columns[MutualExclusivityTableColumnType.ADJUSTED_P_VALUE] = {
-            name: "Adjusted p-Value",
-            render: (d: MutualExclusivity) => formatPValueWithStyle(d.adjustedPValue),
-            tooltip: <span>Bonferroni adjusted p-Value</span>,
-            sortBy: (d: MutualExclusivity) => d.adjustedPValue,
-            download: (d: MutualExclusivity) => formatPValue(d.adjustedPValue)
+        this._columns[MutualExclusivityTableColumnType.Q_VALUE] = {
+            name: "q-Value",
+            render: (d: MutualExclusivity) => formatQValueWithStyle(d.qValue),
+            tooltip: <span>Derived from Benjamini-Hochberg procedure</span>,
```
adamabeshouse: might be good to say "Benjamini-Hochberg FDR correction procedure", because not everyone will know by looking immediately

src/shared/lib/calculation/BenjaminiHochbergFDRCalculator.ts
```diff
@@ -0,0 +1,12 @@
+export function calculateQValues(sortedPValues: number[]): number[] {
```
adamabeshouse: decided to leave it this way? it's probably okay

dippindots: Yes Adam, trying to keep the same pace with the backend implementation. So decided to use the array input and output.

dippindots: Thanks for your idea for the method, I am sure I will use it later!

adamabeshouse: 👍

dippindots:
@adamabeshouse @ersinciftci Could you help me to review the update of this pr one more time. Thank you so much!

PR #1853, opened 2018-12-18
Study view fixes

zhx828:
This PR fixes broken e2e tests;
This solves issue https://github.com/cBioPortal/cbioportal/issues/5469


src/pages/studyView/filters/groupLogic/GroupLogic.tsx
```diff
@@ -21,7 +21,7 @@ export class GroupLogic extends React.Component<IGroupLogicProps, {}> {
         }
         _.reduce(this.props.components, (acc, next, index): any => {
                 if (index > 0) {
-                    acc.push(<span key={this.props.operation} className={styles.text}>{this.props.operation}</span>);
+                    acc.push(<span key={index} className={styles.text}>{this.props.operation}</span>);
```
alisman: how come you switched from operation?  you should not use index as key.  it's probably ok here, but could be dangerous elsewhere

PR #1855, opened 2018-12-18
Eliminate references to legacy services

alisman:
Eliminate references to legacy services


end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -234,26 +234,6 @@ describe('patient view page screenshot test', function(){
     });
 });
 
-describe('study view screenshot test', function(){
```
alisman: there's a new spec in release-1.19.0

PR #1864, opened 2018-12-19
Add Custom Selection button

zhx828:
# What? Why?
This solves https://github.com/cBioPortal/cbioportal/issues/5492
https://github.com/cBioPortal/cbioportal/issues/5516


jjgao:
updated https://github.com/cBioPortal/cbioportal/issues/5492. We may want to close this pr.

zhx828:
@jjgao I will update this PR to reflect the latest requirement.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -126,6 +132,17 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
         }
     });
 
+    @computed
+    get addChartButtonText() {
+        if (!this.props.routing.location.query.tab || this.props.routing.location.query.tab === StudyViewPageTabKeys.SUMMARY) {
```
alisman: since i see we are making a lot of reference to this, lets make getter on the store, currentTab, which then references this.  That way we won't end up with tons of referenences to router everywhere.   This will be actually be close to the ultimate solution.

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -130,74 +165,98 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+        }
     }
 
     @autobind
     private onToggleOption(key: string) {
-        if (this.selectedAttrs.includes(key)) {
-            this.props.store.addCharts(_.reduce(this.selectedAttrs, (acc, next) => {
-                if (next !== key) {
-                    acc.push(next);
+        const option = _.find(this.clinicalDataOptions.concat(this.genomicDataOptions), option => option.key === key);
```
alisman: method should probably @action?

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -191,14 +208,42 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
-                                        store={this.store}
-                                        currentTab={this.props.routing.location.query.tab ? this.props.routing.location.query.tab : ''}
-                                        addChartOverlayClassName='studyViewAddChartOverlay'
-                                        disableAddGenomicButton={this.props.routing.location.query.tab === StudyViewPageTabKeys.CLINICAL_DATA}
-                                    />}
+                                                        style={{marginLeft: '10px'}}>Custom Selection
+                                                </button>
+                                            </DefaultTooltip>
+                                            <AddChartButton
+                                                key={'addChart'}
```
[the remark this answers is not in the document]
zhx828: There were two AddChartButton in the previous implementation, thought to use that to differentiate the components, but it would not make much sense since they are not under a for loop. I will remove this.

src/pages/studyView/StudyViewPage.tsx
zhx828: Sounds good, will do.

src/pages/studyView/addChartButton/AddChartButton.tsx
zhx828: Yeah, good call.

PR #1869, opened 2018-12-20
modify query - dont close form if submit fails

adamabeshouse:


tmazor:
Looks good, but the cursor isn't jumping to the location of the error - just wanted to confirm that's because the fix just isn't included here yet, and not that the bug reappeared? @adamabeshouse

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -136,9 +136,11 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
     }
 
     @autobind
-    onSubmit(){
-        this.closeQueryForm();
-        getGAInstance()('send', 'event', 'resultsView', 'query modified');
+    onSubmit(submitSucceeded:boolean){
```
alisman: not important but would it make more sense not to call onSubmit in case of failure?

PR #1870, opened 2018-12-20
oncoprint - dont annotate CNA with oncokb if oncokb is off

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2178,7 +2178,7 @@ export class ResultsViewPageStore {
         onResult:result=>{
             initializeCustomDriverAnnotationSettings(
                 result!,
-                this.mutationAnnotationSettings,
+                this.driverAnnotationSettings,
```
alisman: driver annotation settings can be changed by user input?

adamabeshouse: yeah, thats the mutations menu in the oncoprint 
image

see the Putative Driver vs VUS section

PR #1872, opened 2018-12-21
select visual study

dippindots:
fix the visual study is not be selected.


dippindots:
@alisman There are two commits right now.
The first one is: keep the URL with the cancer_study_id.
The second one is the default URL, the backward is when we refresh the page, the study will deselect.

PR #1877, opened 2018-12-21
Fix remaining issues caused by OQL DRIVER keyword. This includes making all tabs use same data as oncoprint

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -873,14 +873,16 @@ export class ResultsViewPageStore {
...
         await:()=>[
-            this.unfilteredAlterations,
+            this.putativeDriverAnnotatedMutations,
+            this.annotatedMolecularData,
             this.selectedMolecularProfiles,
             this.defaultOQLQuery
         ],
         invoke:()=>{
             if (this.rvQuery.oqlQuery.trim() != "") {
+                const data = [...(this.putativeDriverAnnotatedMutations.result!), ...(this.annotatedMolecularData.result!)];
```
alisman: just curious why these are in parens

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -978,55 +982,10 @@ export class ResultsViewPageStore {
...
-                    )
-                })
-            ));
-        }
-    }
-});
-
     readonly isSampleAlteredMap = remoteData({
-        await: () => [this.filteredAlterationsDataByUnflattenedOQLLine, this.samples],
+        await: () => [this.putativeDriverFilteredCaseAggregatedDataByUnflattenedOQLLine, this.samples],
```
alisman: i wonder if this is getting TOO descriptive?   do we ever have this data NOT putative driver filtered?

alisman:
@adamabeshouse lets go over this one together.  looks fine but just want to make sure i understand.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: good point.. with this pull request, all our data is now putative driver filtered. ill remove that from the names.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: hmm i guess no reason, spread operator wouldnt apply to `this` right?

PR #1882, opened 2018-12-27
add tooltip for study name in main query page

dippindots:
fix the issue "Full study names not always visible on the main query page" by adding tooltip when the user hover over the study name.


jjgao:
@dippindots Thanks for the pull request. I am wondering if it is possible to display more text when the window is large. @alisman

dippindots:
@jjgao Yes, we can do that.

dippindots:
@jjgao Just updated the css to display more text.

PR #1884, opened 2018-12-28
Add w/mutation and w/cna data charts

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5464

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -68,7 +68,8 @@ import {
     MutationCountVsCnaYBinsMin,
     NA_DATA,
     showOriginStudiesInSummaryDescription,
-    submitToPage
+    submitToPage,
+    getClinicalDataCountWithColorByCategoryCounts
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
 import {SingleGeneQuery} from 'shared/lib/oql/oql-parser';
@@ -108,6 +109,8 @@ export enum UniqueKey {
     CANCER_STUDIES = 'CANCER_STUDIES',
     MUTATION_COUNT = "SAMPLE_MUTATION_COUNT",
     FRACTION_GENOME_ALTERED = "SAMPLE_FRACTION_GENOME_ALTERED",
+    WITH_MUTATION_DATA = "WITH_MUTATION_DATA",
+    WITH_CNA_DATA = "WITH_CNA_DATA"
 }
 
 export enum StudyViewPageTabKeyEnum {
@@ -193,6 +196,33 @@ export const SPECIAL_CHARTS: ChartMeta[] = [{
     },
     renderWhenDataChange: false,
     priority: 70
+},
+{
+    uniqueKey: UniqueKey.WITH_MUTATION_DATA,
+    displayName: 'With Mutation Data',
+    description: 'With Mutation Data',
+    chartType: ChartTypeEnum.PIE_CHART,
+    dataType: ChartMetaDataTypeEnum.GENOMIC,
+    patientAttribute:false,
+    dimension: {
+        w: 1,
+        h: 1
+    },
+    priority: 0,
+    renderWhenDataChange: false
+},{
+    uniqueKey: UniqueKey.WITH_CNA_DATA,
+    displayName: 'With CNA Data',
+    description: 'With CNA Data',
+    chartType: ChartTypeEnum.PIE_CHART,
+    dataType: ChartMetaDataTypeEnum.GENOMIC,
+    patientAttribute:false,
+    dimension: {
+        w: 1,
+        h: 1
+    },
+    priority: 0,
+    renderWhenDataChange: false
 }];
 
 export type CustomGroup = {
@@ -244,6 +274,12 @@ export type StatusMessage = {
     status: 'success' | 'warning' | 'danger' | 'info',
     message: string
 };
+
+export enum Datalabel {
+    YES = 'YES',
+    NO = 'NO',
+    NA = "NA"
+}
 
 export class StudyViewPageStore {
     private reactionDisposers:IReactionDisposer[] = [];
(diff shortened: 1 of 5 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2946,17 +2983,43 @@ export class StudyViewPageStore {
         submitToPage(url, formOps, '_blank');
     }
 
+    //
```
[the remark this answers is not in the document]
kalletlak: forgot to clean it

jjgao:
@kalletlak @zhx828 agreed. I think the check box and the pie chart should be linked. The same filter should be used. (currently, there are two: `Samples with CNA data` & `With CNA data: YES` -- let's use the latter)

Selecting the checkbox `w/ CNA data` should be the same as selecting the `Yes` slice (If the `No` slice is currently selected, it should be reset or switched).

Unselecting the checkbox `w/ CNA data` should be the same as reset the pie chart.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3023,28 +3114,50 @@ export class StudyViewPageStore {
...
+                            result['NO'] = {
+                                'count': noCount,
+                                'value': 'NO'
+                            }
+                        }
+                    } else {
+                        result = _.reduce(this.selectedSamples.result, (acc, sample) => {
+                            const findCase = _.find(this._customChartsSelectedCases.get(uniqueKey), (selectedCase:CustomChartIdentifierWithValue) => selectedCase.sampleId === sample.sampleId);
+                            let value =  'NA';
+                            if(findCase !== undefined) {
```
alisman: "foundCase" might be better name

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3023,28 +3114,50 @@ export class StudyViewPageStore {
...
+                    return _.includes([UniqueKey.WITH_MUTATION_DATA, UniqueKey.WITH_CNA_DATA],uniqueKey) ? [this.molecularProfileSampleCounts, this.selectedSamples] : [this.selectedSamples];
+                },
                 invoke: async () => {
-                    const result = _.reduce(this.selectedSamples.result, (acc, sample) => {
-                        const findCase = _.find(this._customChartsSelectedCases.get(uniqueKey), (selectedCase:CustomChartIdentifierWithValue) => selectedCase.sampleId === sample.sampleId);
-                        let value =  'NA';
-                        if(findCase !== undefined) {
-                            value =  findCase.value;
-                        }
+                    let result: { [id: string]: ClinicalDataCount } = {};
```
alisman: we should extract the code in this invoke to pure helper function and have a test for it.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3023,28 +3114,50 @@ export class StudyViewPageStore {
...
-                    }, {} as { [id: string]: ClinicalDataCount });
-
+                        if(noCount > 0) {
+                            result['NO'] = {
+                                'count': noCount,
+                                'value': 'NO'
+                            }
+                        }
+                    } else {
+                        result = _.reduce(this.selectedSamples.result, (acc, sample) => {
```
alisman: i would avoid calling this the "result" because it's not actually the output of the code here.  give a more descriptive name

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3023,28 +3114,50 @@ export class StudyViewPageStore {
...
-                            acc[value].count = acc[value].count + 1
-                        } else {
-                            acc[value] = {
-                                value: value,
-                                count: 1
+                        let yesCount = UniqueKey.WITH_MUTATION_DATA === uniqueKey ? this.molecularProfileSampleCounts.result.numberOfMutationProfiledSamples : this.molecularProfileSampleCounts.result.numberOfCNAProfiledSamples;
+                        let noCount = UniqueKey.WITH_MUTATION_DATA === uniqueKey ? this.molecularProfileSampleCounts.result.numberOfMutationUnprofiledSamples : this.molecularProfileSampleCounts.result.numberOfCNAUnprofiledSamples;
+
+                        if(yesCount > 0) {
+                            result['YES'] = {
```
alisman: does result ever have any other keys besides Yes/No?   If not, then maybe we just declare those keys above or just have local variables yes, no.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -236,6 +266,12 @@ export type CopyNumberAlterationIdentifier = CopyNumberGeneFilterElement & {
     hugoGeneSymbol: string
 }
 
+export enum DataEntity {
```
alisman: This needs a better name.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1252,4 +1255,50 @@ export function submitToPage(url:string, params: { [id: string]: string }, targe
             alert('Sorry, the query is too large to submit');
         }
     }
+}
+
+export function getCustomChartClinicalDataCountWithColor(
+    uniqueKey: string,
```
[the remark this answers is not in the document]
kalletlak: @alisman `uniqueKey` is for chart. How about `chartId`?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: any suggestions?

src/pages/studyView/StudyViewUtils.tsx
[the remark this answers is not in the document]
alisman: well, if i understand correctly, it's not really chartId, which would mean a particular instance of a chart, but chartTypeId, right?

zhx828:
@kalletlak also when you convert the pie chart to table, the chart size is not updated.

kalletlak:
@zhx828 fixed the behavior of newly added pie chart. And it seems like pie to table conversion is an existing issue. should fix that in different pr

zhx828:
@kalletlak I thought I fixed the pie-2-table issue. Did you see same issue in production?

kalletlak:
yes, for custom added charts

PR #1886, opened 2018-12-28
fix download for mutation table

dippindots:
Fix the issue "Allele frequencies and Copy # not downloading from mutations table" by add download function for columns.

### DIFF
--- a/src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
+++ b/src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
@@ -248,7 +248,7 @@ export default class MutationMapperToolStore
     }
 
     @cached get downloadDataFetcher() {
-        return new MutationTableDownloadDataFetcher(this.mutations, () => this.genomeNexusCache);
+        return new MutationTableDownloadDataFetcher(this.mutations, undefined, () => this.genomeNexusCache);
     }
 
     @action public clearCriticalErrors() {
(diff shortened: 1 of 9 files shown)


alisman:
@onur can you take another look at this?

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -248,7 +248,7 @@ export default class MutationMapperToolStore
     }
 
     @cached get downloadDataFetcher() {
-        return new MutationTableDownloadDataFetcher(this.mutations, () => this.genomeNexusCache);
+        return new MutationTableDownloadDataFetcher(this.mutations, {}, () => this.genomeNexusCache);
```
onursumer: Is it possible to set the second parameter to `undefined`, instead of an empty object? Empty object as a parameter is not very intuitive.

src/shared/lib/MutationTableDownloadDataFetcher.ts
```diff
@@ -81,4 +91,24 @@ export class MutationTableDownloadDataFetcher implements ILazyMobXTableApplicati
             return undefined;
         }
     }
+
+    private async fetchAllfetchDiscreteCNAData()
```
onursumer: Did you mean `fetchAllDiscreteCNAData` ?

dippindots: Updated! Thanks for your time!

PR #1888, opened 2019-01-02
Upgrade react 15 to 16

alisman:
Upgrade React and associated deps.  Enzyme introduces breaking changes


src/shared/components/MSKTabs/MSKTabs.spec.tsx
```diff
@@ -22,153 +22,175 @@ describe('MSKTabs', () => {
...
 
     after(()=>{
 
     });
 
-    it('initial render only mounts first tab', async ()=>{
-        await sleep(0);
-        assert.equal(tabs.find('.msk-tab').length, 1);
+    it('initial render only mounts first tab', (done)=>{
+        setTimeout(function(){
```
adamabeshouse: does the upgrade make `async` tests not work?

alisman: so, i think it's more of a question of why the tests used to work (without done ever being called).  i'm not sure why since MSKTabs has a deferred rendering cycle.   this may have to do with React 16 changes to sync/async rendering.  anyway, this is actually more in line with mocha documentation so it's probably better.

adamabeshouse: you dont have to return or call anything for an `async` test https://jasmine.github.io/tutorials/async

PR #1891, opened 2019-01-03
study view - optimize calculation for Add Chart for Custom Groups

adamabeshouse:
Improve code which computes intersection of sample identifier selections
over charts.

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -1018,15 +1018,14 @@ export class StudyViewPageStore {
             filters.mutationCountVsCNASelection = this._mutationCountVsCNAFilter;
         }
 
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1018,15 +1018,15 @@ export class StudyViewPageStore {
...
-            let [chartKey,sampleIdentifiers] = next
-            if(key === 0){
-                acc = sampleIdentifiers
-            } else {
-                acc = _.intersectionWith(acc,sampleIdentifiers, _.isEqual) as SampleIdentifier[];
-            }
-            return acc
-        },[] as SampleIdentifier[]);
+        let _sampleIdentifiers:SampleIdentifier[] =
+            _.chain(this._chartSampleIdentifiersFilterSet.entries())
```
zhx828: I think we can directly use . values to avoid additional map?

PR #1892, opened 2019-01-03
Dispose study view page reactions; Only visualize tables when priority is not 0

zhx828:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -270,6 +271,12 @@ export class StudyViewPageStore {
        });
     }
 
+    componentWillUnmount() {
```
alisman: The store is not a React component so has no lifecycle, unless this method is being called by a component somewhere.    Lets discuss how this should happen (@adamabeshouse has views as well)

alisman:
this should probably be two prs?

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: theres no way around it, there has to be a destroy method which we call when we switch away from the study view page

alisman: agreed.   would it make sense to set up tear down the reactions in the Page component itself, since that owns the lifecycle?      or have a store super class with handlers for this kind of thing.

adamabeshouse: I think the clearest thing to do would be to destroy it in the same place we instantiate it. If we put the destroy() call in StudyViewPage.componentWillUnmount, we have to go into StudyViewPage to see it, which to me is not ideal. What do you think about adding a prop to StudyViewPage: `onUnmount`, and then putting the call to store.destroy() in that? That way that code is in the same zone as where we instantiate the store.

alisman: @zhx828 @adamabeshouse picking this up again.    i think we're in agreement?   in Page component, we instantiate store and associated reactions in constructor of onComponentDidMount and then destroy them in componentWillUnmount.     I think this makes sense because these reactions are tied to the lifecycle of the page.

adamabeshouse: that sounds good to me :+1:

zhx828: @alisman @adamabeshouse thank you guys. Updated the pr.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -270,6 +266,23 @@ export class StudyViewPageStore {
        });
     }
 
+    // store reactions should be registered here.
+    onComponentDidMount() {
```
alisman: this should really be in the same place where store is instantiated.   constructor of view. as  it actually has nothing to do with component mounting.

adamabeshouse: you're right. I think it's my fault because I suggested it - I was confused and thought that the study view store was instantiated elsewhere, so that to put the call to destroy() in the same place as its instantiated would require something like this.

PR #1894, opened 2019-01-04
Handle invalid genes error by directing use to correct query in query…

alisman:
We get a handful of errors each day where users have outdated links with invalid genes.  This PR directs users back to query page to fix things.


src/shared/components/query/QueryAndDownloadTabs.tsx
```diff
@@ -14,21 +15,24 @@ const QUERY = 'query';
...
 
-	get store()
-	{
-		return this.props.store;
+	constructor(props:IQueryAndDownloadTabsProps){
+		super();
+		this.store = props.getQueryStore();
 	}
 
+	@observable.ref store: QueryStore;
```
adamabeshouse: you only want to set this on init?

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -141,6 +140,16 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
         getGAInstance()('send', 'event', 'resultsView', 'query modified');
     }
 
+    @computed get queryForm(){
+        return <div style={{marginTop:10}}>
```
adamabeshouse: welcome to the dark side 😈

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2051,12 +2053,29 @@ export class ResultsViewPageStore {
...
         onResult:(genes:Gene[])=>{
             this.geneCache.addData(genes);
+        },
+        onError:(err)=>{
+            // throwing this allows sentry to report it
+            throw(err);
         }
     });
 
+    @computed get genesInvalid(){
```
adamabeshouse: should we also catch invalid profiles/studies/overall query, like `QueryStore.submitError`?

alisman: lets discuss at standup.  i happen to know from sentry that we often get invalid genes.  not sure if that's true for those other params.

src/pages/resultsView/querySummary/QuerySummary.tsx
alisman: why is this the dark side?  jsx from @computed?  if so, i've long been guilty.  the darkside is jsx from mobxpromise.

src/shared/components/query/QueryAndDownloadTabs.tsx
alisman: the basic idea is that when i show the the query component, i want to create a store for that session/lifecycle of the component.    it cannot be replaced/updated by anything.  and it gets throw out on close.   i could model this above in the page, but this seems cleaner to me.  let me know what you think.

PR #1896, opened 2019-01-04
study view grid layout algorithm change

dippindots:
change the layout based on the previous layout.


dippindots:
@zhx828 Please help me to take a look at this approach. I have disabled the even layout right now because it is much better to me. If that is not good, we can discuss another approach. Thank you!

dippindots:
@zhx828 I have updated this pr, please feel free to leave any comments here. Then I am going to create and update the unit tests and end-to-end tests.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -949,11 +949,24 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
...
 
-export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): Layout[] {
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, currentGridLayout?: Layout[], currentGridElement?: ChartMeta): Layout[] {
     let layout: Layout[] = [];
     let matrix = [new Array(cols).fill('')] as string[][];
+    let sortedVisibleAttributes = visibleAttributes.sort((a, b) => b.priority - a.priority);
 
-    _.forEach(visibleAttributes.sort((a, b) => b.priority - a.priority), (chart: ChartMeta) => {
+    if (currentGridLayout && currentGridLayout.length > 0) {
+        if (currentGridElement){
```
zhx828: I would suggest to separate out the currentGridElement out from sortedVisibleAttributes if available. Would be easier to deal with later on when currentGridElement becomes an array.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -437,11 +437,20 @@ export class StudyViewPageStore {
...
     }
 
+    @autobind
+    @action
+    updateCurrentGridLayout(currentGridLayout: ReactGridLayout.Layout[]) {
+        this.currentGridLayout = currentGridLayout;
+    }
+
+    private currentGridLayout: ReactGridLayout.Layout[] | undefined = undefined;
+    private currentGridElement: ChartMeta | undefined = undefined;
```
zhx828: This is great. Now you have a clear picture how this `currentGridElement` is used. Let's name this variable more clearly. `currentFocusedChartByUser`? We don't use `element` concept in anywhere yet, so let's avoid this.

src/pages/studyView/StudyViewUtils.tsx
zhx828: Then you also avoid to check `currentGridElement` multiple times.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -977,6 +990,50 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
     return layout;
 }
 
+export function populateFixedPositionLayout(layout: Layout[], matrix: string[][], currentGridLayout: Layout, cols: number, currentGridElement: ChartMeta){
```
zhx828: Since this is an util function, it should be separated to few steps otherwise you are losing the context.
- Populate matrix by using currentGridElement (This probably can be an util function)
- Add the currentGridElement to layout array(this can just be added in calculateLayout method)

dippindots: All comment has been applied to the updated code in the second commit. Thanks!

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -977,6 +986,55 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
     return layout;
 }
 
+export function calculateNewLayoutForFocusedChart(previousLayout: Layout, currentFocusedChartByUser: ChartMeta, cols: number): Layout {
+    const initialX = previousLayout.x;
+    const initialY = previousLayout.y;
+    const dimensionX = currentFocusedChartByUser.dimension.w;
```
zhx828: dimensionX probably should be named `dimensionWidth`?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -977,6 +986,55 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+        }
+    }
+}
+
+export function isFocusedChartShrinked(oldDimension: ChartDimension, newDimension: ChartDimension): boolean {
+    return (oldDimension.w + oldDimension.h) > (newDimension.w + newDimension.h) ? true : false;
+}
+
+export function getPositionXByUniqueKey(layout: Layout[], uniqueKey: string): number {
+    return (_.find(layout, {'i': uniqueKey}) as Layout).x;
```
zhx828: These two functions have to do undefined check. The layout does not necessary have uniqueKey. This is why you need to do a force type here I think.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -977,6 +988,55 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+        }
+    }
+}
+
+export function isFocusedChartShrinked(oldDimension: ChartDimension, newDimension: ChartDimension): boolean {
+    return (oldDimension.w + oldDimension.h) > (newDimension.w + newDimension.h) ? true : false;
+}
+
+export function getPositionXByUniqueKey(layout: Layout[], uniqueKey: string): number | undefined {
+    return _.find(layout, {'i': uniqueKey}) ? (_.find(layout, {'i': uniqueKey}) as Layout).x : undefined;
```
zhx828: You should avoid to find the item twice.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -437,11 +437,20 @@ export class StudyViewPageStore {
...
             gridMargin: STUDY_VIEW_CONFIG.layout.gridMargin,
-            layout: calculateLayout(this.visibleAttributes, cols),
+            layout: calculateLayout(this.visibleAttributes, cols, this.currentGridLayout, this.currentFocusedChartByUser),
             dimensions: STUDY_VIEW_CONFIG.layout.dimensions
         };
     }
 
+    @autobind
+    @action
+    updateCurrentGridLayout(currentGridLayout: ReactGridLayout.Layout[]) {
```
alisman: newGridLayout?

dippindots:
@alisman @zhx828 the unit tests has been updated!

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -953,11 +953,22 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
     return occupied;
 }
 
-export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): Layout[] {
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, previousGridLayout?: Layout[], currentFocusedChartByUser?: ChartMeta): Layout[] {
```
alisman: make this currentGridLayout

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -953,11 +953,22 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
...
     let layout: Layout[] = [];
     let matrix = [new Array(cols).fill('')] as string[][];
+    let sortedVisibleAttributesByPriority = _.orderBy(visibleAttributes, ['priority'], ['desc']);
+
+    if (previousGridLayout && previousGridLayout.length > 0 && currentFocusedChartByUser) {
+        const previousLayout = previousGridLayout.find((layout) => layout.i === currentFocusedChartByUser.uniqueKey)!;
```
alisman: previousChartLayout

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -953,11 +953,22 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
...
+        if (previousLayout) {
```
alisman: handle negative case with throw(new Error(""))

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -953,11 +953,22 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
...
+            const newLayout = calculateNewLayoutForFocusedChart(previousLayout, currentFocusedChartByUser, cols);
+            layout.push(newLayout);
+            populateFixedPositionLayout(newLayout, matrix, cols);
```
alisman: this function should return matrix and it's name should reflect that purpose

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -981,6 +992,57 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
 }
 
+export function calculateNewLayoutForFocusedChart(previousLayout: Layout, currentFocusedChartByUser: ChartMeta, cols: number): Layout {
+    const initialX = previousLayout.x;
+    const initialY = previousLayout.y;
+    const dimensionWidth = currentFocusedChartByUser.dimension.w;
+    let x = initialX;
+    let y = initialY;
+
+    if (isFocusedChartShrinked({w: previousLayout.w, h: previousLayout.h}, currentFocusedChartByUser.dimension)) {
```
alisman: isFocusedChartShrunk

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -981,6 +992,57 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+        h: currentFocusedChartByUser.dimension.h,
+        isResizable: false
+    };
+}
+
+export function populateFixedPositionLayout(layout: Layout, matrix: string[][], cols: number) {
+    const xMax = layout.x + layout.w;
+    const yMax = layout.y + layout.h;
+    while (yMax >= matrix.length) {
+        matrix.push(new Array(cols).fill(''));
```
alisman: lets not use fill method

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -981,6 +992,57 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+    }
+    for (let i = layout.y; i < yMax; i++) {
+        for (let j = layout.x; j < xMax; j++) {
+            matrix[i][j] = layout.i!;
+        }
+    }
+}
+
+export function isFocusedChartShrinked(oldDimension: ChartDimension, newDimension: ChartDimension): boolean {
+    return (oldDimension.w + oldDimension.h) > (newDimension.w + newDimension.h) ? true : false;
```
alisman: don't need to return true/false

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -981,6 +992,57 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+}
+
+export function getPositionXByUniqueKey(layout: Layout[], uniqueKey: string): number | undefined {
+    const findLayoutResult = _.find(layout, {'i': uniqueKey});
```
alisman: make arrow function

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -981,6 +992,57 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): L
...
+    return findLayoutResult ? findLayoutResult.x : undefined;
+}
+
+export function getPositionYByUniqueKey(layout: Layout[], uniqueKey: string): number | undefined {
+    const findLayoutResult = _.find(layout, {'i': uniqueKey});
```
alisman: make arrow function

src/pages/studyView/StudyViewUtils.tsx
dippindots: We may don't need to change this since babel_polyfill handled this function.

zhx828: Just curious why we should not use fill method? @alisman

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1237,6 +1315,32 @@ export function clinicalDataCountComparator(a: ClinicalDataCount, b: ClinicalDat
     }
 }
 
+// Descent sort priority then ascent sort by display name
+export function chartMetaComparator(a: ChartMeta, b: ChartMeta): number
+{
+    if (a.priority !== b.priority) {
+        if (a.priority > b.priority) {
```
zhx828: could just return  b.priority - a.priority

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1237,6 +1315,32 @@ export function clinicalDataCountComparator(a: ClinicalDataCount, b: ClinicalDat
...
+            return -1;
+        }
+        else {
+            return 1;
+        }
+    }
+    else {
+        if (a.displayName !== b.displayName) {
```
zhx828: Could just do a.displayName. localeCompare(b.displayName)

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -953,11 +953,28 @@ export function isOccupied(matrix: string[][], position: Position, chartDimensio
     return occupied;
 }
 
-export function calculateLayout(visibleAttributes: ChartMeta[], cols: number): Layout[] {
+export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, currentGridLayout?: Layout[], currentFocusedChartByUser?: ChartMeta): Layout[] {
     let layout: Layout[] = [];
     let matrix = [new Array(cols).fill('')] as string[][];
+    // sort the visibleAttributes by priority, then make a copy to calculate layout
+    visibleAttributes.sort(chartMetaComparator);
+    let visibleAttributesClone = _.clone(visibleAttributes);
```
zhx828: Why not just a filter instead of a clone?

PR #1898, opened 2019-01-08
Move enrichments queryGenes filtering and qValue computation to front end

adamabeshouse:


src/pages/resultsView/coExpression/CoExpressionViz.tsx
```diff
@@ -99,6 +107,19 @@ export default class CoExpressionViz extends React.Component<ICoExpressionVizPro
         });
     }
 
+    readonly coExpressionsWithQValues = remoteData<(CoExpression & { qValue:number})[]>({
+        await:()=>[this.coExpressionDataPromise],
+        invoke:()=>{
+            const coexpressions = this.coExpressionDataPromise.result!;
+            const sortedByPvalue = _.sortBy(coexpressions, c=>c.pValue);
+            const qValues = calculateQValues(sortedByPvalue.map(c=>c.pValue));
+            for (let i=0; i<qValues.length; i++) {
```
alisman: curious why you like to use for loops rather than forEach?   the only reason i see is optimization

alisman: which is a fine reason if there is actually a need

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -41,9 +42,19 @@ export default class AlterationEnrichmentContainer extends React.Component<IAlte
             this.props.store.hugoGeneSymbols);
     }
 
+    @computed get excludeGenes():string[]|null {
```
alisman: this could use a comment and maybe a better name.

PR #1900, opened 2019-01-08
Update to typescript 3.2.2

adamabeshouse:


src/shared/components/copyDownloadControls/CopyDownloadControls.tsx
```diff
@@ -185,9 +186,10 @@ export class CopyDownloadControls extends React.Component<IAsyncCopyDownloadCont
     {
         // async update of the copy text, if the copy text ends up to be different than the previous one,
         // we show a modal to force a second copy action.
-        setTimeout(this.initCopyProcess(), 1);
+        setTimeout(this.initCopyProcess, 1);
```
adamabeshouse: @onursumer

onursumer: good catch, we should remove the `setTimeout` function, and just call `this.initCopyProcess()`.

PR #1903, opened 2019-01-09
Use AppStore to understand whether user is logged in

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5537
https://github.com/cBioPortal/cbioportal/issues/5517


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -108,6 +109,10 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
         );
     }
 
+    private get appStore(){
```
alisman: you can now get the appStore at the page level by using @injection.   see the ResultsViewPage

zhx828:
@alisman updated

PR #1906, opened 2019-01-09
Results refac

alisman:
This is a just a cleanup of logic in ResultsViewPage.  As we start to include routing on other pages, we may want to share logic/pattern.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -27,18 +27,23 @@ import CoExpressionTab from "./coExpression/CoExpressionTab";
 import Helmet from "react-helmet";
 import {showCustomTab} from "../../shared/lib/customTabs";
 import {getTabId, parseConfigDisabledTabs, ResultsViewTab} from "./ResultsViewPageHelpers";
-import {buildResultsViewPageTitle, doesQueryHaveCNSegmentData} from "./ResultsViewPageStoreUtils";
+import {buildResultsViewPageTitle} from "./ResultsViewPageStoreUtils";
 import {AppStore} from "../../AppStore";
-import {bind} from "bind-decorator";
 import {updateResultsViewQuery} from "./ResultsViewQuery";
 import {trackQuery} from "../../shared/lib/tracking";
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -118,6 +124,7 @@ function initStore() {
                         }
 
                         const changes = updateResultsViewQuery(resultsViewPageStore.rvQuery, query, samplesSpecification, cancerStudyIds, oql);
+
```
alisman: @adamabeshouse this initDriverAnnotations thing is yucky to me.  Idea: should we just reinstantiate the results store every time the query changes?  That way we don't have to any of this equality checking or re-init.  The client level caching will prevent refetching of data.   We just start fresh store every change to query.  It would simplify.

adamabeshouse: yeah I completely agree

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -187,220 +194,13 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
-                id:ResultsViewTab.DOWNLOAD,
-                getTab: () => {
-                    return <MSKTab key={11} id={ResultsViewTab.DOWNLOAD} linkText={'Download'}>
-                        <DownloadTab store={store}/>
-                    </MSKTab>
-                }
-            }
-
-        ];
+        const tabMap = getResultsViewTabMap(this.resultsViewPageStore, this.props.appStore, this.props.routing);
```
adamabeshouse: its actually a tab list not a map

alisman: true!

src/pages/resultsView/ResultsViewPage.tsx
alisman: on second thought, what if we do sometimes want to keep some state in the store but not in the url. it would be lost.  hmmmm.

adamabeshouse: but when would the query change but the state not be reset?

PR #1909, opened 2019-01-10
StudyView updates

kalletlak:
# What? Why?
Fixes
1. https://github.com/cBioPortal/cbioportal/issues/5568
2. https://github.com/cBioPortal/cbioportal/issues/5584

### DIFF
--- a/src/pages/studyView/virtualStudy/VirtualStudy.tsx
+++ b/src/pages/studyView/virtualStudy/VirtualStudy.tsx
@@ -90,16 +90,12 @@ export default class VirtualStudy extends React.Component<IVirtualStudyProps, {}
                     return acc;
                 }, []);
 
-                let filters = { patients: {}, samples: {} };
+                let { sampleIdentifiersSet, ...studyViewFilter } = this.props.filter;
 
-                /* 
-                    TODO: this is to support existing virtual study feature.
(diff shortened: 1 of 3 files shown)


src/pages/studyView/virtualStudy/VirtualStudy.tsx
```diff
@@ -89,16 +89,10 @@ export default class VirtualStudy extends React.Component<IVirtualStudyProps, {}
...
-
-                /* 
-                    TODO: this is to support existing virtual study feature.
-                    but eventually we need to save StudyViewFilter
-                 */
                 let parameters = {
                     name: this.name || this.namePlaceHolder,
                     description: this.description,
-                    filters: filters,
+                    studyViewFilter: _.omit(this.props.filter, ['sampleIdentifiersSet']),
```
zhx828: What about the custom charts?

kalletlak: Not sure. these are actually not captured in `studyViewFilter` class in backend. however I assume `sampleIdentifiers` would be updated accordingly for those charts

alisman: is sampleIdentifiersSet a type property?  If so, it would be best to avoid the above because putting property name as string is not typesafe.

PR #1915, opened 2019-01-11
Mutation enrichments tab for group comparison

adamabeshouse:


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -3,29 +3,64 @@ import {SampleGroup} from "./GroupComparisonUtils";
...
-            this.resultsViewPageStore.samples,
+    @observable private _sampleGroup1Id:string;
+    @observable private _sampleGroup2Id:string;
+
+    @action public setSampleGroups(id1:string, id2:string) {
+        this._sampleGroup1Id = id1;
+        this._sampleGroup2Id = id2;
+    }
+
+    @computed get selectedGroupIds() {
```
kalletlak: there is a possibility for selected groups to be more than 2

adamabeshouse: according to @jjgao we are only doing pairwise comparisons for enrichments. so i guess the group 1 group 2 is enrichment specific

kalletlak: true, for enrichments we require two. but there is a possibility that user can select more than groups(if present)

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -2,30 +2,139 @@ import {ResultsViewPageStore} from "../resultsView/ResultsViewPageStore";
...
+} from "../resultsView/enrichments/EnrichmentsUtil";
+import MobxPromiseCache from "../../shared/lib/MobxPromiseCache";
+import {AlterationEnrichment} from "../../shared/api/generated/CBioPortalAPIInternal";
+import {makeEnrichmentDataPromise} from "../resultsView/ResultsViewPageStoreUtils";
+import internalClient from "../../shared/api/cbioportalInternalClientInstance";
 
 export default class GroupComparisonStore {
     constructor(
-        private resultsViewPageStore:ResultsViewPageStore
+        private __TEMP_SAMPLEGROUPSPROMISE:any
```
kalletlak: Any specific reason for not passing `resultsViewPageStore`

adamabeshouse: @alisman and I spoke and were saying that we don't want GroupComparisonStore to depend on ResultsViewPageStore anymore

kalletlak: ok

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -233,3 +236,27 @@ export function getBoxPlotScatterData(molecularData: NumericGeneMolecularData[],
...
+export function pickCopyNumberEnrichmentProfiles(profiles:MolecularProfile[]) {
+    return _.filter(profiles, (profile: MolecularProfile) =>
+        profile.molecularAlterationType === AlterationTypeConstants.COPY_NUMBER_ALTERATION && profile.datatype === "DISCRETE");
+}
+
+export function pickMRNAEnrichmentProfiles(profiles:MolecularProfile[]) {
+    const mrnaProfiles = profiles.filter(p=>{
+        return p.molecularAlterationType === AlterationTypeConstants.MRNA_EXPRESSION
+    });
+    return Promise.resolve(filterAndSortProfiles(mrnaProfiles));
```
kalletlak: may be just send the result instead of promise

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -233,3 +236,27 @@ export function getBoxPlotScatterData(molecularData: NumericGeneMolecularData[],
...
+}
+
+export function pickProteinEnrichmentProfiles(profiles:MolecularProfile[]) {
+    const protProfiles = profiles.filter(p=>{
+        return p.molecularAlterationType === AlterationTypeConstants.PROTEIN_LEVEL;
+    });
+    return Promise.resolve(filterAndSortProfiles(protProfiles));
```
kalletlak: even here

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
adamabeshouse: oops good catch

PR #1920, opened 2019-01-11
oncoprint - fix case where profiled in tracks wouldnt appear

adamabeshouse:


adamabeshouse:
@tmazor looks working now!

PR #1923, opened 2019-01-14
Survival tab from group comparison

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/5553

### DIFF
--- a/src/pages/groupComparison/GroupComparisonStore.ts
+++ b/src/pages/groupComparison/GroupComparisonStore.ts
@@ -1,25 +1,26 @@
-import {ResultsViewPageStore} from "../resultsView/ResultsViewPageStore";
 import {SampleGroup, TEMP_localStorageGroupsKey} from "./GroupComparisonUtils";
 import {remoteData} from "../../shared/api/remoteData";
-import ListIndexedMap from "../../shared/lib/ListIndexedMap";
 import {
     MolecularProfile,
     MolecularProfileFilter,
-    Sample, SampleFilter,
-    SampleIdentifier
+    SampleFilter,
+    ClinicalDataMultiStudyFilter,
+    ClinicalData
 } from "../../shared/api/generated/CBioPortalAPI";
-import {action, computed, observable} from "mobx";
+import { computed, observable } from "mobx";
 import client from "../../shared/api/cbioportalClientInstance";
 import _ from "lodash";
 import {
     pickCopyNumberEnrichmentProfiles, pickMRNAEnrichmentProfiles,
     pickMutationEnrichmentProfiles, pickProteinEnrichmentProfiles
 } from "../resultsView/enrichments/EnrichmentsUtil";
-import MobxPromiseCache from "../../shared/lib/MobxPromiseCache";
-import {AlterationEnrichment} from "../../shared/api/generated/CBioPortalAPIInternal";
 import {makeEnrichmentDataPromise} from "../resultsView/ResultsViewPageStoreUtils";
 import internalClient from "../../shared/api/cbioportalInternalClientInstance";
 import autobind from "autobind-decorator";
+import { PatientSurvival } from "shared/model/PatientSurvival";
+import request from "superagent";
+import { getPatientSurvivals } from "pages/resultsView/SurvivalStoreHelper";
+import { SURVIVAL_CHART_ATTRIBUTES } from "pages/resultsView/survival/SurvivalChart";
 
 export default class GroupComparisonStore {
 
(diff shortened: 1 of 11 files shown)


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -21,11 +24,57 @@ export default class GroupComparisonStore {
...
+
+    public readonly patientToAnalysisGroup = remoteData({
+        await: () => [
+            this.sampleGroups
+        ],
+        invoke: () => {
+            return Promise.resolve(_.reduce(this.sampleGroups.result, (acc, next) => {
+                next.samples.forEach(sample => {
+                    acc[sample.uniquePatientKey] = next.name
+                })
```
adamabeshouse: missing semicolon (just happened to see this, not something i usually scan for)

src/pages/groupComparison/GroupComparisonTab.tsx
```diff
@@ -2,14 +2,35 @@ import * as React from "react";
...
+import Survival from "./Survival";
 
 export interface IGroupComparisonTabProps {
     store:GroupComparisonStore;
 }
 
 @observer
 export default class GroupComparisonTab extends React.Component<IGroupComparisonTabProps, {}> {
+
+    @observable currentTabId:string;
```
adamabeshouse: 👍

src/pages/groupComparison/Survival.tsx
```diff
@@ -0,0 +1,102 @@
...
+                            xLabelWithoutEventTooltip="Time of Last Observation"
+                            fileName="Disease_Free_Survival" />
+                    </div>
+                </div>
+            );
+        } else {
+            diseaseFreeNotAvailable = true;
+        }
+
+        if (overallNotAvailable && diseaseFreeNotAvailable) {
```
adamabeshouse: wouldnt you want to split these cases up?

adamabeshouse: or put the `push` statements instead of where you set `..NotAvailable = true`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2877,24 +2877,36 @@ export class ResultsViewPageStore {
...
     });
 
-    public readonly sampleGroups = remoteData({
+    public readonly sampleGroups = remoteData<{
+        name: string;
+        sampleIdentifiers: {
+            sampleId: string;
+            studyId: string;
+        }[];
+        color?: string;
```
adamabeshouse: can you add these properties to the `SampleGroup` type?

src/pages/groupComparison/Survival.tsx
kalletlak: I was using same code from survivalTab from resultView. I'll split it

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -20,12 +23,59 @@ export default class GroupComparisonStore {
...
+                this.resultsViewPageStore.sampleGroups.result!.map((group)=>({
                     name: group.name,
-                    samples: group.sampleIdentifiers.map((si:SampleIdentifier)=>sampleIdentifierToSample.get(si.sampleId, si.studyId))
+                    samples: _.reduce(group.sampleIdentifiers,(samples:Sample[], si:SampleIdentifier)=>{
+                        if(sampleIdentifierToSample.get(si.sampleId, si.studyId)){
+                            samples.push(sampleIdentifierToSample.get(si.sampleId, si.studyId)!);
+                        }
+                        return samples;
+                    },[]),
+                    color: group.color ? group.color : colors.shift()!,
```
adamabeshouse: on second thought, if you had a ton of sample groups then you could run out of colors right? so maybe
```
let colorIndex = 0;
return Promise.resolve(
...
color: group.color ? group.color : COLORS[(colorIndex++) % COLORS.length]
...
)
```

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2877,24 +2878,28 @@ export class ResultsViewPageStore {
...
                 },
                 {
                     name: "Unaltered Samples",
-                    sampleIdentifiers: this.unalteredSamples.result!.map(s=>({ sampleId: s.sampleId, studyId: s.studyId }))
+                    sampleIdentifiers: this.unalteredSamples.result!.map(s=>({ sampleId: s.sampleId, studyId: s.studyId })),
+                    color: "blue",
+                    legendText: "Cases without Alteration(s) in Query Gene(s)"
                 },
-                {
+                /* {
```
adamabeshouse: wait why have you commented this out?

adamabeshouse:
@alisman is this still DO NOT MERGE?

src/pages/resultsView/ResultsViewPageStore.ts
kalletlak: this group contains all the samples from first two groups. With the existing logic in SurvivaChart only this group would be drawn. i.e, it doesn't handle overlapping samples/patients. Not sure how to deal with it.

@jjgao How should we deal with groups with overlapping samples in survival plots?

adamabeshouse: that seems like a code obstacle not a conceptual obstacle, right?

kalletlak: even conceptual obstacle. whether to show same sample in multiple lines.

kalletlak: may be for now i'll show it

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -0,0 +1,84 @@
...
+        ],
+        invoke:()=>{
+            const sampleIdentifierToSample = new ListIndexedMap<Sample>();
+            for (const sample of this.resultsViewPageStore.samples.result!) {
+                sampleIdentifierToSample.set(sample, sample.sampleId, sample.studyId);
+            }
+
+            let colorIndex = 0;
+            return Promise.resolve(
+                this.resultsViewPageStore.sampleGroups.result!.map((group)=>({
```
alisman: i think we decided that resultsViewPageStore should not maintain sampleGroups collection.  These groups should be in the groupComparison store

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -0,0 +1,84 @@
...
+                    groups.push(next.name);
+                    acc[sample.uniquePatientKey] = groups;
+                })
+                return acc;
+            }, {} as { [id: string]: string []}));
+        }
+    });
+
+    @computed get showSurvivalTab() {
+        return this.resultsViewPageStore.survivalClinicalDataExists.isComplete && this.resultsViewPageStore.survivalClinicalDataExists.result;
```
alisman: the store should get it's own survival data, not rely on resultsViewPageStore to get it

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -0,0 +1,84 @@
...
+    }
+
+    readonly overallPatientSurvivals = remoteData<PatientSurvival[]>({
```
alisman: the groups should be generic, not based on groups from resultsViewPageStore

kalletlak: this is not group specific. its complete data for `overall survival` plot

adamabeshouse: since we're moving away from depending on resultsviewpagestore we'll have to fetch it separately anyway

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -137,4 +137,117 @@ export default class GroupComparisonStore {
...
+    });
+
+    public readonly patientToAnalysisGroups = remoteData({
+        await: () => [
+            this.sampleGroups,
+            this.sampleSet
+        ],
+        invoke: () => {
+            let sampleSet = this.sampleSet.result!
+            return Promise.resolve(_.reduce(this.sampleGroups.result, (acc, next) => {
```
alisman: lets set this to a local var and then just resolve the local var

src/pages/groupComparison/GroupComparisonStore.ts
alisman: ok gotcha

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -137,4 +137,117 @@ export default class GroupComparisonStore {
...
+    });
+
+    readonly survivalClinicalDataExists = remoteData<boolean>({
+        await:()=>[
+            this.studyIds,
+            this.samples
+        ],
+        invoke:async()=>{
+            const filter: ClinicalDataMultiStudyFilter = {
+                attributeIds: ["OS_STATUS", "OS_MONTHS", "DFS_STATUS", "DFS_MONTHS"],
```
alisman: we should make sure these attribute ids are not duplicated in resultsViewPageStore (put them in a some config file)

PR #1925, opened 2019-01-16
Import twitter library only when it is used

oplantalech:

### DIFF
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -7,7 +7,6 @@ import Testimonials from "../testimonials/Testimonials";
 import {ThreeBounce} from 'better-react-spinkit';
 import AppConfig from "appConfig";
 import {QueryStore} from "shared/components/query/QueryStore";
-import { Timeline } from 'react-twitter-widgets';
 import { Link } from 'react-router';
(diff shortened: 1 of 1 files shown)


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -69,6 +68,7 @@ export default class RightBar extends React.Component<IRightBarProps, IRightBarS
                 </div>
             );
         } else {
+            let Timeline = require('react-twitter-widgets').Timeline;
```
alisman: are you sure this works?   code importing should happen regardless

jjgao:
Maybe we should just remove twitter as a whole? @alisman

Sjoerd-van-Hagen:
@jjgao don't you think the Twitter feed is useful?

jjgao:
@Sjoerd-van-Hagen I thought so but no one used it after a few months. We have removed it.

alisman:
This script is for the twitter feed, not the twitter post.

jjgao:
oh.. sorry. We need twitter feed.

Sjoerd-van-Hagen:
@alisman can you test if it works for you?

src/shared/components/rightbar/RightBar.tsx
alisman: it does seem to work.  very surprised!

PR #1930, opened 2019-01-16
fix select samples button text

dippindots:
change text for select samples button.
fix https://github.com/cBioPortal/cbioportal/issues/5612


zhx828:
@dippindots @alisman we don't need to consider the row only has one sample?

dippindots:
Thanks @zhx828 .Yes, we probably need to consider the case when the user just selects one sample. @alisman I just did the update to cover this case. Can I still push to the original branch?

jjgao:
@zhx828 @dippindots let's not worry about that. Since it's not in a sentence, I don't think we have to be so strick.

PR #1932, opened 2019-01-17
Fix for url hash links in SPA

alisman:
In SPA, links like /tutorials#oql don't work anymore because the anchor corresponding to hash doesn't exist at time url changes (it's part of async loaded content).

This is a hack to address the problem that polls for existence of corresponding DOM element whenever page changes (and url hash exists)

https://github.com/cBioPortal/cbioportal/issues/4554


alisman:
I get where you're coming from but the tutorials page doesn't actually use the StaticContent component because it's not getting markdown  (StaticContent is not a great name).    The async is due to loading of Page bundle (created by code splitting in router).   

To me, the router is a pretty natural place to do this because "on enter" is basically a page load event.  It's quite possible we might want to do this on other pages.   

If you still object, i could see just throwing this in componentDidMount of the Tutorials page.

PR #1933, opened 2019-01-17
Add a download button after study name

zhx828:

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5622#issuecomment-455324823

### DIFF
--- a/src/pages/staticPages/datasetView/DatasetList.tsx
+++ b/src/pages/staticPages/datasetView/DatasetList.tsx
@@ -4,7 +4,9 @@ import {CancerStudy} from 'shared/api/generated/CBioPortalAPI';
 import {ThreeBounce} from 'better-react-spinkit';
 import request from 'superagent';
 import LazyMobXTable from "shared/components/lazyMobXTable/LazyMobXTable";
-import {getStudyDownloadListUrl, getStudySummaryUrl} from "../../../shared/api/urls";
+import {getStudyDownloadListUrl} from "../../../shared/api/urls";
+import {StudyLink} from "../../../shared/components/StudyLink/StudyLink";
+import {StudyDataDownloadLink} from "../../../shared/components/StudyDataDownloadLink/StudyDataDownloadLink";
 
 
 interface IDataTableRow {
@@ -51,14 +53,7 @@ class CancerStudyCell extends React.Component<ICancerStudyCellProps,{}> {
 
     render() {
         return (
-            <span>
-                <a
-                    href={getStudySummaryUrl(this.props.studyId)}
-                    target='_blank'
-                >
-                    {this.props.name}
-                </a>
-            </span>
+            <StudyLink studyId={this.props.studyId}>{this.props.name}</StudyLink>
         );
 
     }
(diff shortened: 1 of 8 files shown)


src/pages/staticPages/datasetView/DatasetList.tsx
```diff
@@ -136,17 +132,17 @@ export default class DataSetsPageTable extends React.Component <IDataSetsTablePr
...
-                                           href={'http://download.cbioportal.org/' + data.studyId + '.tar.gz'} download>
-                                            <i className='fa fa-download'/>
-                                        </a>
-                                    );
-                                }},
                                 {
-                                    name:'Reference',
+                                    name: '', sortBy: false, togglable: false, download: false, type: 'download',
+                                    render: (data: IDataTableRow) => {
+                                        const download = this.state.downloadable.indexOf(data.studyId) > -1;
```
alisman: instead of indexOf, we can now use .includes()

src/pages/studyView/studyPageHeader/studySummary/StudySummary.tsx
```diff
@@ -70,10 +74,37 @@ export default class StudySummary extends React.Component<IStudySummaryProps, {}
...
     }
 
+    @action
+    updateDataDownloadable(content: boolean) {
+        this.dataDownloadable = content;
+    }
+
+    componentDidMount(): void {
+        if (this.props.studies.length === 1) {
+            request(getStudyDownloadListUrl()).then(
```
alisman: i this will automatically fetch all study data on page load (not when user actually requests it)?

src/pages/studyView/studyPageHeader/studySummary/StudySummary.tsx
```diff
@@ -70,10 +74,37 @@ export default class StudySummary extends React.Component<IStudySummaryProps, {}
         return [];
     }
 
+    @action
+    updateDataDownloadable(content: boolean) {
+        this.dataDownloadable = content;
+    }
+
+    componentDidMount(): void {
+        if (this.props.studies.length === 1) {
```
alisman: this should probably be in store so that it's properly responsive.   By putting in in componentDidMount we introduce a dependency on the view lifecycle which is unnecessary

zhx828: Can we talk a bit about this more? The component does not have store. Only the list of studies are required by the component. This blob of the code should not belong here. The code should be executed when `this.props.studies` is updated. A reaction for this?

zhx828:
@alisman this PR is updated.

src/pages/studyView/studyPageHeader/studySummary/StudySummary.tsx
```diff
@@ -73,7 +77,21 @@ export default class StudySummary extends React.Component<IStudySummaryProps, {}
     render() {
         return (
             <div className={classnames(styles.summary)}>
-                <h3 style={{marginBottom:3}}>{this.name}</h3>
+                <h3 style={{marginBottom:3, display: 'flex', alignItems: 'baseline'}}>
```
alisman: Are you sure you want empty <h3> tag to render if conditions are negative?

alisman:
@zhx828 please rebase and merge

PR #1935, opened 2019-01-17
Support simple filter in study view

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5621


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -321,7 +331,7 @@ export class StudyViewPageStore {
     }
 
     @action
-    updateStoreFromURL(query: Partial<StudyViewURLQuery>) {
+    async updateStoreFromURL(query: Partial<StudyViewURLQuery>) {
```
alisman: async actions need special treatment.   also, there is no way to reflect the loading state of this behavior in the UI.

alisman: also, we need tests for this stuff.  lot of logic in here.  it's pretty important because it determines the results the user sees.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -411,17 +452,28 @@ export class StudyViewPageStore {
...
         let initialFilter = {} as StudyViewFilter;
-        if(_.isEmpty(this.queriedSampleIdentifiers.result)){
+        if (_.isEmpty(this.queriedSampleIdentifiers.result)) {
             initialFilter.studyIds = this.queriedPhysicalStudyIds.result;
         } else {
             initialFilter.sampleIdentifiers = this.queriedSampleIdentifiers.result;
         }
+
+        if (this.simpleFilter !== undefined && this.simpleFilterType !== undefined) {
+            if (this.simpleFilterType == "NUMBER") {
```
alisman: double equals

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -411,17 +452,28 @@ export class StudyViewPageStore {
...
+        }
+
         return Object.assign({}, this.initialFiltersQuery, initialFilter);
     }
 
     @computed
     get isInitialFilterState(): boolean {
-        return _.isEqual(toJS(this.initialFilters), toJS(this.filters));
+        const result = _.isEqual(toJS(this.initialFilters), toJS(this.filters));
+        console.log(result);
```
alisman: console

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -108,7 +109,12 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                 }
 
                 this.store.updateCurrentTab(props.routing.location.query.tab);
-                this.store.updateStoreFromURL(query);
+                const newStudyViewFilter:StudyViewURLQuery = _.pick(props.routing.location.query, ['id', 'studyId', 'cancer_study_id', 'filters', 'filterAttributeId', 'filterValues']);
```
zhx828: @alisman do we have a way to only get properties exist in the type `StudyViewURLQuery`?

alisman: I think if you use destructuring notation you will get typechecking

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -108,7 +109,12 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
+
+                if (!_.isEqual(newStudyViewFilter, this.studyViewQueryFilter)) {
+                    this.store.updateStoreFromURL(newStudyViewFilter);
+                    this.studyViewQueryFilter = newStudyViewFilter;
```
alisman: seems like this property "studyViewQueryFilter" isn't used anywhere?

zhx828: It's used in the if condition to check whether the filter is the same as before. Don't update the store is the filter hasn't change.

jjgao:
@zhx828 please let me know once it's release.

PR #1936, opened 2019-01-17
rest of group comparison enrichments tabs

adamabeshouse:


src/pages/resultsView/enrichments/ExpressionEnrichmentsContainer.tsx
```diff
@@ -40,14 +47,18 @@ export default class ExpressionEnrichmentContainer extends React.Component<IExpr
...
+        return getExpressionRowData(this.props.data, this.props.store ? this.props.store.hugoGeneSymbols : []);
     }
 
     @computed get excludeGenes():string[]|null {
-        if (this.props.store.selectedMolecularProfiles.isComplete &&
-            this.props.store.selectedMolecularProfiles.result
-                .findIndex(x=>x.molecularProfileId === this.props.selectedProfile.molecularProfileId) > -1) {
-            return this.props.store.hugoGeneSymbols;
+        if (this.props.store) {
+            if (this.props.store.selectedMolecularProfiles.isComplete &&
```
kalletlak: may be combine both if conditions

adamabeshouse: good catch. ill do the same in AlterationEnrichmentContainer

env/group-comparison.sh
```diff
@@ -0,0 +1 @@
+./rc.sh
```
alisman: probably shouldn't be committed, right?  you can use custom btw

src/pages/groupComparison/CopyNumberEnrichments.tsx
```diff
@@ -0,0 +1,61 @@
...
+    private onChangeProfile(m:MolecularProfile) {
+        this.props.store.copyNumberEnrichmentProfile = m;
+    }
+
+    readonly tabUI = MakeMobxView({
+        await:()=>[
+            this.props.store.copyNumberHomdelEnrichmentData,
+            this.props.store.copyNumberAmpEnrichmentData
+        ],
+        forcePending:()=>!this.props.store.copyNumberEnrichmentProfile,
```
alisman: what is forcePending about?

adamabeshouse: its like await, but with any predicate. forcePending might not be the best name.

alisman: but what's the basic idea?   like, don't render this if there is no profile?

adamabeshouse: yeah, consider it pending if theres no profile

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -113,28 +123,169 @@ export default class GroupComparisonStore {
...
+            if (!this._mutationEnrichmentProfile && this.mutationEnrichmentProfiles.result!.length > 0) {
+                return Promise.resolve(this.mutationEnrichmentProfiles.result![0]);
+            } else {
+                return Promise.resolve(this._mutationEnrichmentProfile);
+            }
         }
-    }
-    public set mutationEnrichmentProfile(profile:MolecularProfile|undefined) {
+    });
+    public setMutationEnrichmentProfile(profile:MolecularProfile|undefined) {
```
alisman: @action?   i think it's good to keep up the convention even if when there's only one observable set it means nothing

PR #1937, opened 2019-01-18
Trigger calls to retrieve study tags only when hovering over the info component

oplantalech:
Fix cbioportal/cbioportal#5606

### DIFF
--- a/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
+++ b/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
@@ -24,11 +24,11 @@
 import * as React from 'react';
 import DefaultTooltip from '../defaultTooltip/DefaultTooltip';
 import { observer } from 'mobx-react';
+import { observable, computed } from "mobx";
 import { JsonToTable } from 'react-json-to-table';
 import './StudyTagsTooltip.scss';
 import { remoteData } from '../../api/remoteData';
 import client from "shared/api/cbioportalClientInstance";
-import { MobxPromiseUnionType } from 'mobxpromise';
 import Loader from '../loadingIndicator/LoadingIndicator';
 
 export type StudyTagsTooltipProps = {
@@ -40,10 +40,14 @@ export type StudyTagsTooltipProps = {
     children: any;
 };
 
-@observer
-export default class StudyTagsTooltip extends React.Component<StudyTagsTooltipProps, {}> {
+export type BuildOverlayTooltipProps = {
+    studyDescription: string;
+    studyId: string;
+};
 
-    readonly studyMetadata = remoteData({
+@observer
+default class BuildOverlay extends React.Component<BuildOverlayTooltipProps, {}> {
+    @observable readonly studyMetadata = remoteData({
         invoke: async () => {
             return client.getTagsUsingGET({studyId: this.props.studyId});
         },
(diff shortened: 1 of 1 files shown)


src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -56,31 +60,41 @@ export default class StudyTagsTooltip extends React.Component<StudyTagsTooltipPr
...
-            const resultKeyLength = Object.keys(meta.result).length;
+        else if (this.studyMetadata.isComplete) {
+            const resultKeyLength = Object.keys(this.studyMetadata.result).length;
             const description = <div dangerouslySetInnerHTML={this.addHTMLDescription(this.props.studyDescription)}/>;
-            overlay = resultKeyLength > 0 ? ([description, <br/>, <div className="studyTagsTooltip"> <JsonToTable json={meta.result}/></div>]) : description;
+            overlay = resultKeyLength > 0 ? ([description, <br/>, 
+                <div className="studyTagsTooltip"> <JsonToTable json={this.studyMetadata.result}/></div>]) : description;
         }
-        if (meta.isError) {
+        else if (this.studyMetadata.isError) {
```
alisman: how about "Error retrieving study meta data"

src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -56,31 +60,41 @@ export default class StudyTagsTooltip extends React.Component<StudyTagsTooltipPr
         return {__html: description};
     }
 
-    renderTooltip = (meta: MobxPromiseUnionType<any>) => {
+    render() {
         let overlay:any = '';
-        if (meta.isPending) {
+        if (this.studyMetadata.isPending) {
```
alisman: instead of having a local variable "overlay" with no typing, how about just use return statements below

src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -56,31 +60,41 @@ export default class StudyTagsTooltip extends React.Component<StudyTagsTooltipPr
...
             overlay = <Loader isLoading={true}/>;
         }
-        if (meta.isComplete) {
-            const resultKeyLength = Object.keys(meta.result).length;
+        else if (this.studyMetadata.isComplete) {
+            const resultKeyLength = Object.keys(this.studyMetadata.result).length;
             const description = <div dangerouslySetInnerHTML={this.addHTMLDescription(this.props.studyDescription)}/>;
-            overlay = resultKeyLength > 0 ? ([description, <br/>, <div className="studyTagsTooltip"> <JsonToTable json={meta.result}/></div>]) : description;
+            overlay = resultKeyLength > 0 ? ([description, <br/>,
```
alisman: you can now use <React.Fragment> as wrapper instead of the array

src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
oplantalech: I thought about that, but since the if-else doesn't cover all cases there is a problem. I don't want to replace one of the "if else" cases by "else" because then you don't see explicitly all the cases and it is more difficult to understand the code.

PR #1938, opened 2019-01-18
search by study id in the home page

dippindots:
search by study id in the home page.
solved issue: https://github.com/cBioPortal/cbioportal/issues/3993

### DIFF
--- a/src/shared/components/query/StudyListLogic.ts
+++ b/src/shared/components/query/StudyListLogic.ts
@@ -35,8 +35,14 @@ export default class StudyListLogic
 		// first compute individual node match results
 		let parsedQuery = parse_search_query(this.store.searchText);
 		let map_node_searchResult = new Map<CancerTreeNode, SearchResult>();
-		for (let [node, meta] of this.store.treeData.map_node_meta.entries())
-			map_node_searchResult.set(node, perform_search_single(parsedQuery, meta.searchTerms));
+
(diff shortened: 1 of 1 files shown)


src/shared/components/query/StudyListLogic.ts
```diff
@@ -35,8 +35,14 @@ export default class StudyListLogic
 		// first compute individual node match results
 		let parsedQuery = parse_search_query(this.store.searchText);
 		let map_node_searchResult = new Map<CancerTreeNode, SearchResult>();
-		for (let [node, meta] of this.store.treeData.map_node_meta.entries())
-			map_node_searchResult.set(node, perform_search_single(parsedQuery, meta.searchTerms));
+
+		for (let [node, meta] of this.store.treeData.map_node_meta.entries()){
+			let searchTerms = meta.searchTerms;
+			if (!meta.isCancerType){
```
alisman: could we/should we determine that is IS a study rather than it's NOT a cancer type?

dippindots: @alisman In my observation, the node should be either cancer type or the study. @kalletlak could you please correct me if I am wrong on this point? Thanks!

alisman: that is probably the case, but still what i'm saying is, why not test it positive rather than negative? in other words, meta.isStudy?

dippindots: @alisman Thanks! Updated!

PR #1940, opened 2019-01-18
Group comparison - overlap tab

kalletlak:
Fixes https://github.com/cBioPortal/cbioportal/issues/5552


src/pages/groupComparison/GroupComparisonUtils.ts
```diff
@@ -8,4 +9,23 @@ export type SampleGroup = {
...
+            let newSet = {
+                sets: [...res.sets, currentSet.name],
+                cases: commonCases
+            }
+            combinations.push(newSet);
+            f(newSet, sets.slice(i + 1));
+        }
+    }
+    f({ sets: [], cases: [] }, sets);
+    return combinations;
```
adamabeshouse: while I appreciate the recursive solution, I think we should simplify this by hardcoding the combinations for up to three groups, since that's all we need. To reflect this more in the code, the function parameter could also take a tuple type of two elements with an optional third `(sets: [SampleGroup, SampleGroup, SampleGroup?])` 

what do you think?

adamabeshouse: also, why have you not taken SampleGroup[] as the argument here?

kalletlak: there is a possibility to list all the combinations in a table. I would prefer to leave as is and may be send only 3 groups while invoking it. 
and `SampleGroup` has `SampleIdentifier` which doesn't have any information about patient

adamabeshouse: ah, I see. sounds good

adamabeshouse: I think it deserves a unit test

adamabeshouse:
I think it really needs unit tests for the methods you added in GroupComparisonUtils

PR #1941, opened 2019-01-21
change url for whole slide viewer

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/5630.

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -33,7 +33,8 @@ import "./patient.scss";
 import IFrameLoader from "../../shared/components/iframeLoader/IFrameLoader";
 import {
     getDigitalSlideArchiveIFrameUrl,
-    getSampleViewUrl
+    getSampleViewUrl,
+    getWholeSlideViewerUrl
 } from "../../shared/api/urls";
 import {PageLayout} from "../../shared/components/PageLayout/PageLayout";
 import Helmet from "react-helmet";
@@ -43,6 +44,9 @@ import {showCustomTab} from "../../shared/lib/customTabs";
 import {StudyLink} from "../../shared/components/StudyLink/StudyLink";
 import WindowStore from "shared/components/window/WindowStore";
 import {QueryParams} from "url";
+import {AppStore} from "../../AppStore";
+import request from 'superagent';
+import {remoteData} from "../../shared/api/remoteData";
 
 const patientViewPageStore = new PatientViewPageStore();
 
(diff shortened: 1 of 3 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -158,6 +159,10 @@ export class PatientViewPageStore {
         return this.pageMode === 'sample' ? this.sampleId : this.patientId;
     }
 
+    @computed get userName(){
+		return getBrowserWindow().globalStores.appStore.userName;
```
jjgao: This this the full email address?

dippindots: @jjgao  Yes, it is the whole email address. The format is like username@mskcc.org.

dippindots: Maybe I should change the name to the userEmail or something?

alisman:
@dippindots, the patient view store should probably not have a reference to appStore if possible.  Instead, in the view, we should have a @computed which produces the url.  Also, you'll see in both ResultsViewPage and StudyViewPage how to use @inject decorator to get access to appStore.   Use this method.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
jjgao: that's ok. Maybe add a comment.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -193,6 +199,18 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         showCustomTab(div, tab, this.props.routing.location, patientViewPageStore);
     }
 
+    private wholeSlideViewerUrl = remoteData({
+        await: () => [patientViewPageStore.getWholeSlideViewerIds],
+        invoke: async() => {
+            const url = getWholeSlideViewerUrl(patientViewPageStore.getWholeSlideViewerIds.result!, this.props.appStore.userName!);
+            await request.get(url);
```
alisman: you should probably put a comment to explain this logic.   i.e. if request succeeds then we return the url because we know request works.   but if request fails, we will return undefined.    
in fact, a more clear way of doing this would be simple to wrap request in try {} catch {}  rather than use the onError handler, which slightly confuses things.

dippindots: Thanks for your help, @alisman. I just updated this pr. It is ready to review now.

PR #1945, opened 2019-01-23
improve the load behavior of the download page

dippindots:
improve the load behavior of the download page.
Fix https://github.com/cBioPortal/cbioportal/issues/5124

### DIFF
--- a/src/pages/resultsView/download/DownloadTab.tsx
+++ b/src/pages/resultsView/download/DownloadTab.tsx
@@ -25,6 +25,8 @@ import {remoteData} from "../../../shared/api/remoteData";
 import LoadingIndicator from "shared/components/loadingIndicator/LoadingIndicator";
 import onMobxPromise from "shared/lib/onMobxPromise";
 import {MolecularProfile} from "shared/api/generated/CBioPortalAPI";
+import {getMobxPromiseGroupStatus} from "../../../shared/lib/getMobxPromiseGroupStatus";
+import ErrorMessage from "../../../shared/components/ErrorMessage";
 
 export interface IDownloadTabProps {
     store: ResultsViewPageStore;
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -247,42 +248,49 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
-                        title="Type of Genetic Alterations Across All Samples"
-                        isLoading={this.caseAlterationData.isPending || this.oqls.isPending}
-                        className="pull-left forceHeaderStyle h4"
-                    />
-                    {this.oqls.isComplete && this.caseAlterationData.isComplete && this.props.store.alterationsBySelectedMolecularProfiles.isComplete && (
-                        <CaseAlterationTable
-                            caseAlterationData={this.caseAlterationData.result}
-                            oqls={this.oqls.result}
-                            alterationTypes={this.props.store.alterationsBySelectedMolecularProfiles.result}
+            <If condition={this.downloadableFilesTable.isPending || this.geneAlterationData.isPending || this.caseAlterationData.isPending || this.oqls.isPending}>
```
alisman: promises can have isError = true, so this would result in "data" view being shown for error state. 
You should use getMobxPromiseGroupStatus to obtain status for all the deps and then show separate state for pending, complete, error.    

We should probably have a component that just accepts a group of promises and lets dev specify the different handlers corresponding to above states.  But we don't have one.

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -221,70 +223,56 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
-                </div>
-            </WindowWidthBox>
-        );
+        const status = getMobxPromiseGroupStatus((this.downloadableFilesTable, this.geneAlterationData, this.caseAlterationData, this.oqls));
+
+        switch (status) {
+            case "pending":
+                return <LoadingIndicator isLoading={true} center={true} size={"big"}/>;
+            case "error":
+                return <ErrorMessage/>;
```
onursumer: This looks simpler but now, for example, a single error for downloadable files table will prevent rendering of error-free gene alteration and case alteration tables. I don't know if it is a big deal though.

dippindots: Thanks Onur, that's a good point. @alisman Do we still want to show the rest of good tables if we got the error.

PR #1955, opened 2019-01-25
Add genome build info to mutation mapper tool

inodb:
We currently don't indicate which version we are using. Text becomes:

screen shot 2019-01-26 at 1 57 52 pm


schultzn:
Maybe also add genome build information in the Data Format description?

Mutation files should be tab delimited, and should at least have the genomic location headers in the first line for a successful annotation. Note that all variants have to be reported for genome build GRCh37 (hg19).

inodb:
@schultzn Thanks! Updated

PR #1959, opened 2019-01-28
Fix two serious bugs in plots stacked bar

adamabeshouse:
Unfortunately these stacked plots have just been completely wrong. Bars from various categories have been mixed together. I discovered this when I added more info to the tooltip, showing which x-axis category each bar-segment belongs to, and found they were all shuffled around. I should have done more careful testing to make sure the plots were right.

This PR also fixes the issue where the plot layout would be out of whack.

image

Note that there is still an outstanding bug with this plot:
https://github.com/cBioPortal/cbioportal/issues/5569


src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -358,11 +312,12 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
 
     @computed get chartExtent() {
         const miscPadding = 100; // specifying chart width in victory doesnt translate directly to the actual graph size
-        const numBars = this.data.length;
-        return this.categoryCoord(numBars - 1) + 2*this.categoryAxisDomainPadding + miscPadding;
-        //return 2*this.domainPadding + numBoxes*this.barWidth + (numBoxes-1)*this.barSeparation;
-        //const ret = Math.max(computedExtent, this.props.chartBase);
-        //return ret;
+        if (this.data.length > 0) {
+            const numBars = this.data[0].counts.length;
```
adamabeshouse: I don't think testing this makes sense - it is subject to the specific hardcoded aesthetic values for padding. I think this is the kind of thing that validation testing should catch - whether the chart looks right. Our existing screenshot tests catch this wonky behavior but I/we didn't pay close enough attention. Our updated screenshots should catch it.

jjgao:
@adamabeshouse thanks for the fix. Can we reduce the space in x-axis. Ideally using the exactly same layout and look&fill (even code) as Cancer Types Summary.

http://bit.ly/2W2bJjF

Plots:
image

Cancer Types Summary:
image

src/shared/components/plots/StackedBarPlot.tsx
alisman: sounds good

PR #1965, opened 2019-01-30
Group creation interface in study view

adamabeshouse:
To review: go to a study view page, and click the button labeled "Groups" on the top right


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -256,7 +264,8 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
                                             >
                                                 <button className='btn btn-primary btn-xs'
                                                         data-test='custom-selection-button'
-                                                        onClick={() => {
+                                                        onClick={(e)=>{
+                                                            e.stopPropagation();
```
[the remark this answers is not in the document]
adamabeshouse: this is another way to handle the issue of clicking outside the button should close the menu. see on line 164.

src/pages/groupComparison/ComparisonGroupManager.tsx
```diff
@@ -0,0 +1,285 @@
...
+    private cancelAddGroup() {
+        this.addGroupPanelOpen = false;
+        this._inputGroupName = "";
+        this.addSamplesTargetGroupId = "";
+    }
+
+    private get header() {
+        return (
+            <div>
+                <h5 style={{position:"absolute", top:7}}>Groups</h5>
```
alisman: curious why h5 is position absolute.  If it's just to bump it down, better to do with margin-top so it stays in flow.

src/pages/groupComparison/ComparisonGroupManager.tsx
```diff
@@ -0,0 +1,285 @@
...
+                    placeholder="Search.."
+                    value={this.groupNameFilter}
+                    onChange={this.onChangeGroupNameFilter}
+                />
+            </div>
+        );
+    }
+    private get groupCheckboxes() {
+        return (
+            <div className="group-checkboxes">
```
alisman: group-checkboxes is a tiny bit generic.  Maybe css-modules?

src/pages/groupComparison/ComparisonGroupManager.tsx
```diff
@@ -0,0 +1,285 @@
...
+                {this.filteredGroups.length > 0 && (
+                    this.filteredGroups.map(group=>(
+                        <div className="checkbox"><label>
+                            <input
+                                type="checkbox"
+                                value={group.id}
+                                checked={this.props.store.isComparisonGroupSelected(group.id)}
+                                onClick={()=>{
```
alisman: repeated elements are one place we should habitually use autobound methods as handlers.  i know it's annyoing because we want to use the group from the closure.   maybe research some solution.  not urgent.

src/pages/groupComparison/ComparisonGroupManager.tsx
```diff
@@ -0,0 +1,285 @@
...
+                >Delete</button>
+            );
+        }
+    }
+
+    private get actionButtons() {
+        return (
+            <div
+                style={{
+                    position:"absolute",
```
alisman: lets discuss why position:absolute.    It should generally be avoided except when you need to position something outside the "flow" of the DOM.

src/pages/groupComparison/GroupComparisonUtils.ts
```diff
@@ -2,14 +2,17 @@ import ListIndexedMap from 'shared/lib/ListIndexedMap';
...
 
 export type SampleGroup = {
     // mandatory:
     id:string, // unique identifier
     sampleIdentifiers:SampleIdentifier[], // samples in the group
+    name:string, // display name
 
     // optional:
-    name?:string, // display name
     color?: string; // color for charts
```
alisman: maybe slight more descrtive name for color?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -276,6 +279,34 @@ export class StudyViewPageStore {
...
+    public isComparisonGroupSelected(groupId:string) {
+        return !!this._selectedComparisonGroups.get(groupId);
+    }
+
+    @action public removeComparisonGroupSelectionEntry(groupId:string) {
+        this._selectedComparisonGroups.delete(groupId);
+    }
+
+    @computed get comparisonGroups() {
+        const allGroups = getLocalStorageGroups();
```
alisman: i imagine you'll do this when you integrate Karthik's services, but would be nice to have abstraction of persistance so we don't have any implementation specific names here (e.g. localStorage).

alisman:
@adamabeshouse @zhx828 @jjgao would it make sense to have group managment in a modal to give us more space and avoid come complexity of the dropdown?  just an idea.

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: I agree, but let's leave it until later when we know exactly how everything fits together. I don't think we're adding any real technical debt since in the end all the users are just asking for the total list `comparisonGroups`

adamabeshouse:
@alisman my feeling is that modals are visually jarring and it's not complex enough to need a modal.

One disadvantage I can think of is that it might be nice to be able to look at the state of the study view for context when interacting with the groups, such as deciding which to view/select, and deciding on a name for the new group.

src/pages/groupComparison/ComparisonGroupManager.tsx
adamabeshouse: interesting, good point

src/pages/groupComparison/GroupComparisonUtils.ts
adamabeshouse: I'll do this in a new PR where I also address some refactoring to be done around colors

src/pages/groupComparison/GroupComparisonUtils.ts
```diff
@@ -122,4 +125,18 @@ export function getPatientIdentifiers(
 export const ENRICHMENTS_TOO_MANY_GROUPS_MSG = "Can't show enrichments for more than two groups - deselect some from the area at the top of the page.";
 export function ENRICHMENTS_TOO_MANY_STUDIES_MSG(enrichmentsType:string) {
     return `Selected comparison groups span more than one study, so we can't show ${enrichmentsType} enrichments. Please deselect groups from the top of the page, or try a different set of groups.`;
+}
+
+export function getDefaultGroupName(filters:StudyViewFilter) {
+    // TODO: generate describe name from filters
+    return "<DESCRIPTIVE GENERATED NAME FROM FILTERS>";
```
adamabeshouse: @jjgao @alisman do we actually want to build this out? I thought it could be cool to autogenerate "Lung cancer" if you have lung cancer selected, but then I realized it would be useless/cluttered if the selection was any more complicated than that. So in the current implementation I leave the name box empty. Thoughts?

jjgao: @adamabeshouse that would be cool, but yeah if it can be a masters thesis project :D

For virtual study, we just put the time as default.

adamabeshouse: this is the name which would be exposed to the user, so I dont think time would look very nice right? Ill change the uid to be based on the time

PR #1966, opened 2019-01-30
Quick search

alisman:
Implement phase one of quick search


inodb:
This is in release-2.2.0 branch now

PR #1950, opened 2019-01-24
change text for mrna and protein

dippindots:
change the text for mRNA and protein from up/down-regulation to high/low.
Fix https://github.com/cBioPortal/cbioportal/issues/5529.

PR #1968, opened 2019-01-31
Improve download tab

dippindots:
fix https://github.com/cBioPortal/cbioportal/issues/5659
- [x] Looking for all different rows for data for a gene (a gene may present in multiple rows)
- [x] Add links to study, sample, and patient
- [x] Change colors of `no alteration` and `not profiled` to gray and add a strikethrough to ~~not profiled~~
- [x] Add columns for each gene that combines all alterations (hidden by default but included in download.


src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -141,14 +144,18 @@ export function computeAlterationTypes(alterationData: ICaseAlteration[]): strin
 export function getDisplayStyle(value: string): object {
     switch(value) {
         case "no alteration":
-            return {whiteSpace: "nowrap" , color: "orange"};
+            return {whiteSpace: "nowrap" , color: "gray"};
```
alisman: this should really be in css and applied via classname.   This is especially important for elements which are repeated many times in DOM as their is a performance cost.  Just as a general rule.

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -192,29 +199,53 @@ export default class CaseAlterationTable extends React.Component<ICaseAlteration
             }
         ];
 
-        const typeSet : {[x: string]: boolean} = {};
+        const geneSet : {[x: string]: OQLLineFilterOutput<AnnotatedExtendedAlteration>} = {};
         this.props.oqls.forEach(oql => {
+            if (oql.gene in geneSet) {
+                geneSet[oql.gene].oql_line += oql.oql_line;
```
alisman: does the oql_lin value have a delimiter in it so that it can be just concatenated in this way?

src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -192,29 +199,53 @@ export default class CaseAlterationTable extends React.Component<ICaseAlteration
...
+            }
+            else {
+                geneSet[oql.gene] = oql;
+            }
+        });
+        
+        _.values(geneSet).forEach(oql => {
```
alisman: i don't think you need to first call _.values.  _.forEach(geneSet, ()=>)  will do it without extra iteration cycle.  here, impact is negligible but in general ...

src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -371,6 +372,8 @@ export function generateCaseAlterationData(
                 // that's why we need a map here
                 caseAlterationData[key].oqlData[data.oql.oql_line] = generateOqlData(datum, geneAlterationDataByGene, molecularProfileIdToMolecularProfile);
                 updateOqlData(datum, caseAlterationData[key].oqlData[data.oql.oql_line], molecularProfileIdToMolecularProfile);
+                caseAlterationData[key].oqlDataByGene[data.oql.gene] = caseAlterationData[key].oqlDataByGene[data.oql.gene] !== undefined ? _.merge(generateOqlData(datum, geneAlterationDataByGene, molecularProfileIdToMolecularProfile), caseAlterationData[key].oqlDataByGene[data.oql.gene]) : generateOqlData(datum, geneAlterationDataByGene, molecularProfileIdToMolecularProfile);
```
alisman: this code here is too dense.  very hard to read.   i would avoid using ternary when there's so much going on.  if statement would be clearer.  or you could spread ternary across several lines with linebreaks

src/pages/resultsView/download/CaseAlterationTable.tsx
dippindots: Yes, the oql_line comes with delimiter at the end.

src/pages/resultsView/download/DownloadUtils.ts
dippindots: Thank Aaron, I will update and clean the code then let you know!

dippindots:
@cBioPortal/product the improvement of download tab is done, could you help to review this pr when you get a chance? Thanks!
The patient view link and sample link is not working on Heroku instance, but it is working on the on our server.
Here are some test examples:
1. Single gene (alteration/ no alteration/ not profiled): http://bit.ly/2UHhv8U
2. Overlapped gene: http://bit.ly/2GeiUAL

jjgao:
Thanks, @dippindots. Could you also change UP/Down to High/Low. It wasn't handled by https://github.com/cBioPortal/cbioportal-frontend/pull/1950

dippindots:
@jjgao Thank you for review my pr.
1. I will fix UP/Down to High/Low in this pr. 
2. These two links are not working on Heroku instance, but they are working on the on our portal with local storage setting. Maybe @alisman or @inodb know the reason? I noticed that these links are produced by `buildCBioPortalPageUrl` function.

jjgao:
I am signing off from the product perspective.

dippindots:
Thank you @jjgao. I will ask Aaron about the link issue and give this pr another tech review.

PR #1972, opened 2019-02-01
Support backslash comma in URL filterValues when value has comma in it

zhx828:


zhx828:
@jjgao escape backslash is now supported and deployed.  http://www.cbioportal.org/study?id=msk_impact_2017&filterAttributeId=CANCER_TYPE&filterValues=Skin%20Cancer\,%20Non-Melanoma,Breast%20Cancer

PR #1973, opened 2019-02-01
When oncoprint is zoomed out, show multiple columns in tooltip

adamabeshouse:


jjgao:
I think it looks good.

image

tmazor:
I agree, this looks good!

One thought - should the first line of the tooltip make clear that multiple samples/patients are included? The current tooltip has that information as the last line, but I wonder if having at the beginning of the tooltip will make it clearer?

Also, when there are gene panels listed, the sample count (e.g. "(7)") is included as part of the link. This actually confused me a little - can the link be just the gene panel name, and the count for each gene panel is not part of the link?

adamabeshouse:
@tmazor good catch, I just updated to put the gene panel sample count outside the link. I agree with you that it would be nice to show the total at the top. Let me know what you and @jjgao think the best way to do that would be.

jjgao:
> One thought - should the first line of the tooltip make clear that multiple samples/patients are included?

Good idea. @tmazor are you suggesting moving `View these 7 patients` to the top? I think that may work.

adamabeshouse:
I think it looks quite a lot better that way anyway

image

image

adamabeshouse:
@jjgao @tmazor just updated it. Please let me know if it looks good, and if so I'll write tests and merge it.

jjgao:
@adamabeshouse looks good!  👍

PR #1976, opened 2019-02-04
Fix redirect on 401s

alisman:
Redirection broken since SPA but has gone unnoticed because so few people access site logged out or lose their session.


alisman:
@n1zea144 this should fix issue on frontend, though as discussed, we should figure out why it's not handled by server redirection.   The above fix should only have to handle the rare case where someone loses a session while the application is open.

PR #1977, opened 2019-02-05
Merge master to 2.1.0

zhx828:


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -70,6 +69,7 @@ export default class RightBar extends React.Component<IRightBarProps, IRightBarS
                     </div>
                 );
             } else {
+                let Timeline = require('react-twitter-widgets').Timeline;
```
zhx828: @adamabeshouse @alisman this is one of the conflicts I'm not so sure about. Another one study view is resolved.

zhx828: It's introduced in https://github.com/cBioPortal/cbioportal-frontend/pull/1925.

zhx828: I was trying to test whether it still works, but seems `What's New` section is hidden on cbioporta.org/beta/ portal. Is this intended?

PR #1979, opened 2019-02-05
Handle broken virtual study fetching for heroku

alisman:
On heroku we try to use proxy to load virtual studies. It breaks. We need to fail gracefully.


alisman:
approve by @inodb  as he walks out door

PR #1981, opened 2019-02-06
Fix restoration of urls after user loses session 

alisman:
…t page without session)

### DIFF
--- a/src/appBootstrapper.jsx
+++ b/src/appBootstrapper.jsx
@@ -110,12 +110,13 @@ superagent.Request.prototype.end = function (callback) {
         }
         if (response && response.statusCode === 401) {
             var storageKey = `redirect${Math.floor(Math.random() * 1000000000000)}`
-            localStorage.setItem(storageKey, window.location.hash);
+            localStorage.setItem(storageKey, window.location.href);
 
+            // build URL with a reference to storage key so that /restore route can restore it after login
             const loginUrl = buildCBioPortalPageUrl({
-               query: {
-                   "spring-security-redirect":`${window.location.pathname}${window.location.search}${'#/restore?key=' + storageKey}`
-               }
+                query: {
+                    "spring-security-redirect":buildCBioPortalPageUrl({ pathname:"restore", query: { key: storageKey} })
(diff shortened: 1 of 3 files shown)


src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -78,9 +78,11 @@ export default class MutationMapperStore
                     return this.getMutations();
                 }
             } else {
-                if (this.activeTranscript === undefined) {
-                    this.activeTranscript = canonicalTranscriptId;
-                }
+                try {
+                    if (this.activeTranscript === undefined) {
+                        this.activeTranscript = canonicalTranscriptId;
```
alisman: @inodb this code (setting instance prop inside of getter) is an antipattern and quite possibly a bug.  when this method is accessed inside a computed property or observer component (which does in fact happen), mobx throws an error in order to avoid endless loop.   we should looking into this

src/appBootstrapper.jsx
```diff
@@ -110,12 +110,13 @@ superagent.Request.prototype.end = function (callback) {
...
-            localStorage.setItem(storageKey, window.location.hash);
+            localStorage.setItem(storageKey, window.location.href);
 
+            // build URL with a reference to storage key so that /restore route can restore it after login
             const loginUrl = buildCBioPortalPageUrl({
-               query: {
-                   "spring-security-redirect":`${window.location.pathname}${window.location.search}${'#/restore?key=' + storageKey}`
-               }
+                query: {
+                    "spring-security-redirect":buildCBioPortalPageUrl({ pathname:"restore", query: { key: storageKey} })
```
alisman: this was never updated when we changed routing from hash to html5 and so restoration of route after login was broken

src/shared/components/mutationMapper/MutationMapperStore.ts
inodb: @alisman 	i see - thanks! Will revisit when fixing this https://github.com/cBioPortal/cbioportal/issues/3522

PR #1982, opened 2019-02-06
make iframes responsive on patient view page

dippindots:
make iframes on patient view page responsive.
fix https://github.com/cBioPortal/cbioportal/issues/5692

### DIFF
--- a/src/pages/patientView/pathologyReport/PathologyReport.tsx
+++ b/src/pages/patientView/pathologyReport/PathologyReport.tsx
@@ -8,6 +8,7 @@ import {observer} from "mobx-react";
 export type IPathologyReportProps = {
 
     pdfs:PathologyReportPDF[];
+    iframeHeight: number;
     iframeStyle?:{[styleProp:string]:any};
 
 }
@@ -35,9 +36,9 @@ export default class PathologyReport extends React.Component<IPathologyReportPro
 
     }
 
-    shouldComponentUpdate(nextProps: IPathologyReportProps){
-        return nextProps === this.props;
-    }
+    // shouldComponentUpdate(nextProps: IPathologyReportProps){
+    //     return nextProps === this.props;
+    // }
 
     handleSelection(){
         this.setState({ pdfUrl:this.buildPDFUrl(this.pdfSelectList.options[this.pdfSelectList.selectedIndex].value) });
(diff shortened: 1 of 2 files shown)


dippindots:
@alisman I found that in our WindowStore, the window size is the window's inner size. But sometimes the content in the window will larger than the window, so I added the scroll size into it. But Maybe it is better to create a DocumentStore to?

src/pages/patientView/pathologyReport/PathologyReport.tsx
```diff
@@ -35,26 +36,25 @@ export default class PathologyReport extends React.Component<IPathologyReportPro
...
+    // }
 
     handleSelection(){
         this.setState({ pdfUrl:this.buildPDFUrl(this.pdfSelectList.options[this.pdfSelectList.selectedIndex].value) });
     }
 
     render(){
 
-        return (<div>
+        return (<div style={{position: "relative"}}>
```
alisman: what's reason for position:relative?

dippindots: hmm, there is no special reason for that. I can clean the unneeded style.

dippindots:
Here are two examples URL:
1. https://cbioportal.mskcc.org/patient/pathologyReport?caseId=TCGA-OR-A5J1&studyId=acc_tcga_pan_can_atlas_2018
2. https://cbioportal.mskcc.org/patient?studyId=mskimpact&caseId=P-0022312

jjgao:
Thanks, @dippindots .

I think it's probably good idea to remove the footer now? @alisman

dippindots:
@jjgao @alisman Sorry I forgot to ask. Do we want to hide the footer for all tabs in patient view page or just for these three tabs?

jjgao:
@dippindots I think we should remove it from patient view if @alisman agrees.

src/shared/components/window/WindowStore.ts
```diff
@@ -20,13 +21,14 @@ class WindowStore {
     @action
     private setWindowSize() {
         this.size = {
-            width: this.windowObj.innerWidth,
-            height: this.windowObj.innerHeight
+            width: Math.max(this.windowObj.innerWidth, this.documentObj.documentElement.scrollWidth),
```
alisman: I'm curious why we do this.  Let me know when you get in.  It does appear to be working well!

PR #1984, opened 2019-02-06
Use string replace over regex

zhx828:
The Lookbehinds is only supported in ECMA2018 which ony chrome supports it at this moment. https://stackoverflow.com/questions/50011366/javascript-regex-negative-lookbehind-not-working-in-firefox

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5697  https://github.com/cBioPortal/cbioportal/issues/5696

### DIFF
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -1255,5 +1255,5 @@ export function submitToPage(url:string, params: { [id: string]: string }, targe
 }
 
 export function getClinicalEqualityFilterValuesByString(filterValues: string):string[] {
-    return filterValues.split(/(?<!\\),/).map(val=>val.trim().replace('\\,',','));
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1255,5 +1255,5 @@ export function submitToPage(url:string, params: { [id: string]: string }, targe
 }
 
 export function getClinicalEqualityFilterValuesByString(filterValues: string):string[] {
-    return filterValues.split(/(?<!\\),/).map(val=>val.trim().replace('\\,',','));
+    return filterValues.replace('\\,','$@$').split(",").map(val=>val.trim().replace('$@$',','));
```
alisman: i'm pretty sure these replaces will only replace single instance, not global.   to make it globak, you need to use regexp with //g

PR #1985, opened 2019-02-06
Fix patient page when navCaseIds is very long (google analytics regex…

alisman:
We use url hash to transmit very large list of patients in cohort to patient page.  Google Analytics runs a regexp on entire url that crashes.  This fixes that by stashing list in local storage when list is too long.


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -172,7 +173,24 @@ export class PatientViewPageStore {
...
+    @observable.ref
+    private _patientIdsInCohort: string[] = [];
+
+    public set patientIdsInCohort(patientIds:string[]){
+        runInAction(()=>this._patientIdsInCohort = patientIds);
+    }
+
+    @computed
+    public get patientIdsInCohort(){
+        if (getBrowserWindow().localStorage.navCaseIds) {
```
adamabeshouse: as we discussed, put these branches in opposite order

PR #1987, opened 2019-02-07
Keep the parent tooltip alive when mousing over child tooltip

zhx828:


zhx828:
Seems the issue is automatically solved after upgrading to the latest library in https://github.com/cBioPortal/cbioportal-frontend/pull/1991/files

PR #1990, opened 2019-02-07
Homepage should ignore broken session service (virtual studies)

alisman:
Homepage should not show error message when virtual study service is down


alisman:
@inodb i reformatted this whole file, which has been messed up since days of Andy.

PR #1992, opened 2019-02-07
Do not fetch oncokb genes info when show_oncokb is false

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5643


zhx828:
@alisman I thought about that. But all other references need to properly handle the await. That seems too much at all places calling for this promise. And giving the promise is only initialized once in a page, do that in here is cleaner to keep the logic at one place.

PR #1993, opened 2019-02-07
OncoKB : Remove the navCaseIds for the feedback

zhx828:
Fix https://github.com/cBioPortal/cbioportal/issues/3754

### DIFF
--- a/src/shared/components/annotation/OncoKB.tsx
+++ b/src/shared/components/annotation/OncoKB.tsx
@@ -18,6 +18,7 @@ import OncoKbTooltip from "./OncoKbTooltip";
 import OncokbPubMedCache from "shared/cache/PubMedCache";
 import {default as TableCellStatusIndicator, TableCellStatus} from "shared/components/TableCellStatus";
 import AppConfig from "appConfig";
+import {getCurrentURLWithoutHash} from "../../api/urls";
 
 export interface IOncoKbProps {
     status: "pending" | "error" | "complete";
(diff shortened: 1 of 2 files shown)


src/shared/components/annotation/OncoKB.tsx
```diff
@@ -170,7 +170,7 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}>
         const geneParam = `entry.1744186665=${hugoSymbol || ''}`;
         const alterationParam = `entry.1671960263=${alteration || ''}`;
         const userParam = `entry.1381123986=${this.props.userEmailAddress || ''}`;
-        const uriParam = `entry.1083850662=${encodeURIComponent(window.location.href)}`;
+        const uriParam = `entry.1083850662=${encodeURIComponent(window.location.href.replace(/navCaseIds=[^&]*/, ''))}`;
```
alisman: probably better to just build window.location.hostname+window.location.path+window.location.search. but i'll approve and you make change if you want

zhx828: @alisman I'm thinking to keep other parameters after the pound sign

alisman: i just don't want to have that parameter hardcoded.  i don't think the hash will be important

zhx828: Thanks. Updated the url.

PR #1994, opened 2019-02-08
Make wider heatmap menu

pvannierop:
# What? Why?
This PR implements a wider heatmap menu  in order to accommodate menu items with longer names.

- The heatmap menu is assigned an additional `heatmap` className.
- Css for oncoprint controls set this menu to 400px. 
- The selection menu is set to fill available real estate (100%).
- The text area for genes can only be resized in vertical direction. 

# Before
screenshot from 2019-02-08 15-11-09
screenshot from 2019-02-08 15-11-31
## Genes text box
screenshot from 2019-02-09 08-48-36

# After
screenshot from 2019-02-08 15-11-45
screenshot from 2019-02-08 15-11-48
## Genes text box:
screenshot from 2019-02-09 08-47-39

# Note
The other menu's of the oncoprint controls keep their original width.

### DIFF
--- a/src/pages/resultsView/plots/PlotsTab.spec.tsx
+++ b/src/pages/resultsView/plots/PlotsTab.spec.tsx
@@ -4,7 +4,6 @@ describe("PlotsTab", ()=>{
     describe("controls", ()=>{
         describe("gene lock", ()=>{
             it("genes can be freely selected when gene lock is off", ()=>{
-                assert.equal(1,1);
             });
             it("when gene lock is selected, both axes gene selection should be the one from the axis where lock was clicked", ()=>{
             });
--- a/src/pages/resultsView/plots/PlotsTab.spec.tsx
+++ b/src/pages/resultsView/plots/PlotsTab.spec.tsx
@@ -4,6 +4,7 @@ describe("PlotsTab", ()=>{
     describe("controls", ()=>{
         describe("gene lock", ()=>{
             it("genes can be freely selected when gene lock is off", ()=>{
+                assert.equal(1,1);
             });
             it("when gene lock is selected, both axes gene selection should be the one from the axis where lock was clicked", ()=>{
             });
--- a/src/shared/components/oncoprint/controls/CustomDropdown.tsx
+++ b/src/shared/components/oncoprint/controls/CustomDropdown.tsx
@@ -6,7 +6,7 @@ import {observable} from "mobx";
 
 export interface ICustomDropdownProps extends ButtonProps {
     title: string;
-    buttonProps?: ButtonProps;
+    classType?: string;
 }
 
 class CustomButton extends React.Component<any,{}> { // cant type this more specifically because of some typing issues w ES6 classes not having component.replaceState
@@ -22,10 +22,10 @@ class CustomButton extends React.Component<any,{}> { // cant type this more spec
 class CustomMenu extends React.Component<any,{}> {
 
     render() {
-        const { children } = this.props;
+        const { classType, children } = this.props;
 
         return (
-            <div className="dropdown-menu" style={{ padding: '6px' }}>
+            <div className={`dropdown-menu ${classType}`} style={{ padding: '6px' }}>
                 {children}
             </div>
         );
@@ -46,12 +46,12 @@ export default class CustomDropdown extends React.Component<ButtonProps, {}> {
     }
 
     render() {
-        const {children, id, ...props} = this.props;
+        const {children, id, classType, ...props} = this.props;
         return (
             <RootCloseWrapper onRootClose={this.hide}>
                 <Dropdown id={id+""} open={this.open}>
                     <CustomButton bsStyle="default" bsRole="toggle" title="Custom Toggle" onClick={this.toggle} {...props}/>
-                    <CustomMenu bsRole="menu">
+                    <CustomMenu bsRole="menu" classType={classType}>
                         {children}
                     </CustomMenu>
                 </Dropdown>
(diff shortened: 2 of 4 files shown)


src/pages/resultsView/plots/PlotsTab.spec.tsx
```diff
@@ -4,6 +4,7 @@ describe("PlotsTab", ()=>{
     describe("controls", ()=>{
         describe("gene lock", ()=>{
             it("genes can be freely selected when gene lock is off", ()=>{
+                assert.equal(1,1);
```
alisman: hmm!

src/shared/components/oncoprint/controls/CustomDropdown.tsx
```diff
@@ -6,7 +6,7 @@ import {observable} from "mobx";
 
 export interface ICustomDropdownProps extends ButtonProps {
     title: string;
-    buttonProps?: ButtonProps;
+    classType?: string;
```
alisman: I would keep things consistent/generic and make this "className".  You can then use the classNames library to apply both dropdown-menu and anything passed in through this prop.   Search for "classNames" and you'll find it used a lot in our project

src/shared/components/oncoprint/controls/styles.scss
```diff
@@ -37,6 +36,10 @@
     }
   }
 
+  .dropdown-menu.heatmap {
```
alisman: use sass & syntax to nest this in the .dropdown-menu class above.

src/shared/components/oncoprint/controls/styles.scss
```diff
@@ -37,6 +36,10 @@
...
+    width:450px;
```
alisman: would max-width work and still allow it to shrink to size when necessary?

src/shared/components/oncoprint/controls/styles.scss
```diff
@@ -84,7 +87,7 @@
     }
 
     .Select-menu-outer {
-      width: 300px;
+      width: 100%;
```
alisman: do you need this class at all?  possibly not.

src/pages/resultsView/plots/PlotsTab.spec.tsx
pvannierop: Wooops, leftover from other code changes. Sorry for this. File is reverted back to its original state.

src/shared/components/oncoprint/controls/CustomDropdown.tsx
pvannierop: I have rename this to `classsName`. I have used the _classNames_ lib as per your suggestion. I am unfamiliar with React and your suggestion:
>You can then use the classNames library to apply both dropdown-menu and anything passed in through this prop.

may mean there are more clever ways of class propagation to children with _classNames_. If so, I am more than interested to be pointed in the right direction.

src/shared/components/oncoprint/controls/styles.scss
pvannierop: Implemented your change.

src/shared/components/oncoprint/controls/styles.scss
pvannierop: No max-width does not work. Width of the menu is not affected by the menu entries. Can you live with 450px fixed?

src/shared/components/oncoprint/controls/styles.scss
pvannierop: Nom its not needed. I removed this css entry.

pvannierop:
I have updated the PR to make the genes text area only resizeble in vertical direction (see updated PR message).

src/shared/components/oncoprint/controls/CustomDropdown.tsx
```diff
@@ -22,10 +23,10 @@ class CustomButton extends React.Component<any,{}> { // cant type this more spec
 class CustomMenu extends React.Component<any,{}> {
 
     render() {
-        const { children } = this.props;
+        const { className, children } = this.props;
```
alisman: i'm not sure why the original dev saved these to local variables.  There's no reason and in general it's better just to reference the property on the props object b/c it's clearer (e.g. this.props.className.   This is not a big deal but since you're new, just want to make sure you know.

src/shared/components/oncoprint/controls/CustomDropdown.tsx
```diff
@@ -46,12 +47,12 @@ export default class CustomDropdown extends React.Component<ButtonProps, {}> {
...
 
     render() {
-        const {children, id, ...props} = this.props;
+        const {children, id, className, ...props} = this.props;
         return (
             <RootCloseWrapper onRootClose={this.hide}>
                 <Dropdown id={id+""} open={this.open}>
                     <CustomButton bsStyle="default" bsRole="toggle" title="Custom Toggle" onClick={this.toggle} {...props}/>
-                    <CustomMenu bsRole="menu">
+                    <CustomMenu bsRole="menu" className={classNames(className)}>
```
alisman: no reason to use classNames here since you have just one thing.

src/shared/components/oncoprint/controls/CustomDropdown.tsx
alisman: no, you have it right now.

alisman:
@adamabeshouse if you look at failed screenshot on this on https://27959-66571349-gh.circle-artifacts.com/0/imageCompare.html
the color of the indicators on oncoprint seem to have changed.  is this expected?

src/shared/components/oncoprint/controls/CustomDropdown.tsx
adamabeshouse: I think it was probably me bc I wrote this class... I think probably it was legacy from when I was trying to extract the rest props in order to pass them through, but then abandoned that or something. anyway you're right

adamabeshouse:
@alisman no, that's not expected, and I'm not able to reproduce that in my browser... :/

PR #2000, opened 2019-02-08
Saving tracks selection state across mutation mapper tabs

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5674.

# Checks
  the `-s` option)

### DIFF
--- a/src/shared/components/mutationMapper/MutationMapper.tsx
+++ b/src/shared/components/mutationMapper/MutationMapper.tsx
@@ -23,6 +23,7 @@ import ProteinChainPanel from "shared/components/proteinChainPanel/ProteinChainP
 import TrackPanel from "../tracks/TrackPanel";
 import {TrackDataStatus, TrackNames, TrackVisibility} from "../tracks/TrackSelector";
 import MutationMapperStore from "./MutationMapperStore";
+import {initDefaultTrackVisibility} from "./MutationMapperUserSelectionStore";
 import { EnsemblTranscript } from 'shared/api/generated/GenomeNexusAPI';
 import Mutations from 'pages/resultsView/mutation/Mutations';
 import {IServerConfig} from "../../../config/IAppConfig";
@@ -31,6 +32,7 @@ import {getNCBIlink} from "../../api/urls";
 
 export interface IMutationMapperProps {
     store: MutationMapperStore;
+    trackVisibility?: TrackVisibility;
     config: IServerConfig;
     studyId?: string;
     myCancerGenomeData?: IMyCancerGenomeData;
@@ -46,12 +48,23 @@ export interface IMutationMapperProps {
 export default class MutationMapper<P extends IMutationMapperProps> extends React.Component<P, {}>
 {
     @observable protected lollipopPlotGeneX = 0;
-    @observable protected trackVisibility: TrackVisibility = {
-        [TrackNames.OncoKB]: 'hidden',
-        [TrackNames.CancerHotspots]: 'hidden',
-        [TrackNames.PDB]: 'hidden'
-    };
+    @observable private _trackVisibility: TrackVisibility|undefined;
     //@observable protected geneWidth = 665;
+
+    @computed
+    protected get trackVisibility(): TrackVisibility
+    {
+        if (this.props.trackVisibility) {
+            return this.props.trackVisibility!;
+        }
+        else {
+            if (!this._trackVisibility) {
+                this._trackVisibility = initDefaultTrackVisibility();
+            }
+
+            return this._trackVisibility;
+        }
+    }
 
     protected handlers:any;
 
@@ -352,7 +365,7 @@ export default class MutationMapper<P extends IMutationMapperProps> extends Reac
 
     protected proteinChainPanel(): JSX.Element|null
     {
-        return (
+        return this.is3dPanelOpen ? (
             <ProteinChainPanel
                 store={this.props.store}
                 pdbHeaderCache={this.props.pdbHeaderCache}
@@ -360,7 +373,7 @@ export default class MutationMapper<P extends IMutationMapperProps> extends Reac
                 geneXOffset={this.lollipopPlotGeneX}
                 maxChainsHeight={200}
             />
-        );
+        ): null;
     }
 
     protected trackPanel(): JSX.Element|null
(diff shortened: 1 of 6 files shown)


src/shared/components/mutationMapper/MutationMapperUserSelectionStore.ts
```diff
@@ -0,0 +1,11 @@
+import {observable} from "mobx";
+import {TrackNames, TrackVisibility} from "../tracks/TrackSelector";
+
+export default class MutationMapperUserSelectionStore
```
onursumer: We can add more states (like column visibility) into this store to shared the state of other types of use selection. For now only supporting track visibility.

src/shared/components/proteinChainPanel/ProteinChainPanel.tsx
```diff
@@ -206,6 +206,10 @@ export default class ProteinChainPanel extends React.Component<ProteinChainPanel
         );
     }
 
+    componentDidMount() {
+        onNextRenderFrame(() => this.props.store.pdbChainDataStore.selectFirstChain());
```
onursumer: Probably there is a better solution to select the first chain at init, but for now keeping the existing logic.

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -487,14 +494,12 @@ export default class MutationMapper<P extends IMutationMapperProps> extends Reac
     @action
     protected open3dPanel() {
         this.trackVisibility[TrackNames.PDB] = 'visible';
-        this.props.store.pdbChainDataStore.selectFirstChain();
```
onursumer: For simplicity, this logic is now moved into `ProteinChainPanel`, we only set the visibility here.

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -487,14 +494,12 @@ export default class MutationMapper<P extends IMutationMapperProps> extends Reac
...
     }
 
     @autobind
     @action
     protected close3dPanel() {
         this.trackVisibility[TrackNames.PDB] = 'hidden';
-        this.props.store.pdbChainDataStore.selectUid();
```
onursumer: We don't need this, since the `ProteinChainPanel` is not rendered anymore when the visibility is set to `hidden`.

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -44,12 +46,17 @@ export interface IMutationMapperProps {
 @observer
 export default class MutationMapper<P extends IMutationMapperProps> extends React.Component<P, {}>
 {
+    public static defaultProps: Partial<IMutationMapperProps> = {
+        userSelectionStore: new MutationMapperUserSelectionStore()
```
onursumer: Using the default `MutationMapperUserSelectionStore` value actually shares an implicit singleton instance for all instances of `MutationMapper`. It might be clearer/safer to explicitly create an instance of `MutationMapperSelectionStore` in the parent class, and pass it to the `MutationMapper` instances.

adamabeshouse: I agree this doesnt seem safe.

another option would be to create one in the constructor of the `MutationMapper` instance, and wrap it in a @computed to use the instance version if the prop is undefined

alisman: I think the best thing would be in contstructor to instantiate a store if one is not passed in props.

src/shared/components/mutationMapper/MutationMapperUserSelectionStore.ts
alisman: Could we just keep these properties on view itself?

src/shared/components/proteinChainPanel/ProteinChainPanel.tsx
alisman: what's reason to use onNextRenderFrame?

alisman: i can see it's been this way, but just curious.

src/shared/components/mutationMapper/MutationMapper.tsx
onursumer: sounds good, I'll change this to explicitly pass a shared store where we instantiate `MutationMapper`s.

adamabeshouse: @alisman @onursumer is edge case relevant where at one point you pass in a shared store then later on you dont, or vice versa? these types of things can happen at various points as the rendered state stabilizes to the declarative spec

src/shared/components/mutationMapper/MutationMapperUserSelectionStore.ts
onursumer: Do you mean have an explicit `trackVisibility` property for `MutationMapper` component, and then instantiate an observable `trackVisibility` in the parent components `Mutations` and `MutationMapperTool`? In that case we need to do the same instantiation for `trackVisibility` (and for some possible future shared user selection observables) in multiple places, that's why I wanted to wrap these shared states inside a separate class.

@alisman We can still instantiate this shared store in the parent components (`Mutations` and `MutationMapperTool`) and explicitly pass the properties (currently only `trackVisibilitly`) instead of the store instance if you think it is a better way to do it.

src/shared/components/proteinChainPanel/ProteinChainPanel.tsx
onursumer: For some reason, it doesn't re-render when we call `selectFirstChain` without `onNextRenderFrame`. Ideally we shouldn't be calling `selectFirstChain` at init at all, but I think that should be fixed in another PR.

src/shared/components/mutationMapper/MutationMapperUserSelectionStore.ts
alisman: The basic idea is that I don't want to have lots of different stores floating nested at various layers of view heirarchy.   I can stomach multiple stores but at top level.   If there is state which is truly local to features/views then i think they can be stored on view.

onursumer: @alisman removed the store from the `MutationMapper` component props. Keeping it only at top level (in `Mutations` and `MutationMapperTool`) and explicitly passing a `TrackVisibility` prop to `MutationMapper`.

PR #2024, opened 2019-02-13
Change to new oncokb route in the OncoKB card link out

zhx828:


src/shared/components/annotation/OncoKbCard.tsx
```diff
@@ -44,9 +44,9 @@ export default class OncoKbCard extends React.Component<OncoKbCardProps>
     get oncokbLinkOut() {
         let link: string | undefined = undefined;
         if (this.props.gene) {
-            link = `http://oncokb.org/#/gene/${this.props.gene}`;
+            link = `https://oncokb.org/gene/${this.props.gene}`;
```
alisman: this should be from urls file/configuration

zhx828: We actually don't want to lead the internal portal user to internal OncoKB instance yet. Will need to be updated when that's available.

PR #2026, opened 2019-02-13
Do not allow select all studies.  Introduce Select pan can button

alisman:
We no longer want to allow user to select all studies.  We need a button that allows them to select all pan can atlas ONLY when no filters are in affect

### DIFF
--- a/src/shared/components/query/CancerStudySelector.tsx
+++ b/src/shared/components/query/CancerStudySelector.tsx
@@ -8,7 +8,7 @@ import classNames from 'classnames';
 import ReactSelect from 'react-select';
 import StudyList from "./studyList/StudyList";
 import {observer, Observer} from "mobx-react";
-import {action, expr} from 'mobx';
+import {action, expr, runInAction} from 'mobx';
 import memoize from "memoize-weak-decorator";
 import {If, Then, Else} from 'react-if';
 import {QueryStore, QueryStoreComponent} from "./QueryStore";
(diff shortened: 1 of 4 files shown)


src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -78,6 +83,7 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
     constructor(props: ICancerStudySelectorProps) {
         super(props);
         this.store = this.props.queryStore;
+        getBrowserWindow().moo = this;
```
alisman: will get rid

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -330,18 +342,28 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
...
-                                           checked={shownAndSelectedStudies.length === shownStudies.length}
-                                    />
-                                    <strong>{(shownAndSelectedStudies.length === shownStudies.length) ?
-                                        `Deselect all listed studies ${(shownStudies.length < this.store.cancerStudies.result.length) ? "matching filter" : ""} (${shownStudies.length})` :
-                                        `Select all listed studies ${(shownStudies.length < this.store.cancerStudies.result.length) ? "matching filter" : ""}  (${shownStudies.length})`}
-                                    </strong>
-                                </label>
+                                <If condition={this.logic.mainView.showQuickSelect}>
+                                    <div className={styles.quickSelect}>
+                                        Quick select: <button onClick={this.selectTCGAPanAtlas} className={"btn btn-default btn-xs"}>TCGA Pancancer Atlas studies</button> <button className={"hidden btn btn-default btn-xs"}>All TCGA PanCan</button>
```
adamabeshouse: whats this second button about? "All TCGA PanCan"

alisman: nice spot.   it's hidden.  i thought there might be more buttons.  will kill it.

PR #2033, opened 2019-02-15
Adding a display name to allow users to visualize on their choice of ID

dippindots:
Changing to "id (display name)" in the header and make it in the first row in various mouse over tables.
Fix https://github.com/cBioPortal/cbioportal/issues/5711.

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -260,7 +260,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                             target="_blank"
                                             onClick={(e: React.MouseEvent<HTMLAnchorElement>) => this.handleSampleClick(sample.id, e)}
                                         >
-                                            {sample.id}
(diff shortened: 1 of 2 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -104,6 +104,21 @@ function transformClinicalInformationToStoreShape(patientId: string, studyId: st
         clinicalData: clinicalDataPatient
     };
     const samples = groupBySampleId(sampleIds, clinicalDataSample);
+    samples.forEach((sample, index) => {
+        let attr = _.clone(sample.clinicalData[1]);
+        attr.clinicalAttribute = {
```
alisman: is this draft code?

dippindots: @alisman Yes, it is the dummy data I created. I will remove it later when we have the attribute in the data.

dippindots: I should add comments for the dummy data next time.

dippindots:
@jjgao Yes, thanks! Just updated.

PR #2035, opened 2019-02-16
add rsid column in mutation table

leexgh:
Add rsid column in mutation table. Return with links.

### DIFF
--- a/src/shared/components/mutationTable/column/RsidColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/RsidColumnFormatter.tsx
@@ -1,0 +1,100 @@
+import * as React from 'react';
+import {Circle} from "better-react-spinkit";
+import 'rc-tooltip/assets/bootstrap_white.css';
+import {Mutation} from "shared/api/generated/CBioPortalAPI";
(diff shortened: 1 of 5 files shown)


src/shared/components/mutationTable/column/RsidColumnFormatter.tsx
```diff
@@ -0,0 +1,99 @@
...
+import GenomeNexusCache, { GenomeNexusCacheDataType } from "shared/cache/GenomeNexusCache";
+
+export default class RsidColumnFormatter {
+
+    public static renderFunction(data:Mutation[],
+                                 genomeNexusCache:GenomeNexusCache|undefined) {
+        const genomeNexusCacheData = RsidColumnFormatter.getGenomeNexusDataFromCache(data, genomeNexusCache);
+        return (
+            <div>
+                <span>{RsidColumnFormatter.getRsidViz(genomeNexusCacheData)}</span>
```
onursumer: do we really need the `<span>` around this component?

leexgh: Yes I think you are right, we don't need it. already removed it.

leexgh: @onursumer

PR #2044, opened 2019-02-19
use gene validation box for oncoprint heatmap genes

adamabeshouse:


src/shared/components/GeneSelectionBox/styles.module.scss
```diff
@@ -4,6 +4,11 @@
 		resize: vertical;
 		border-radius: 2px;
 	}
+	textarea.oncoprintHeatmap {
```
alisman: lets take textarea out since css modules seems to ignore it

adamabeshouse: why do you say it ignores it? it only gives a dynamic classname for the classname, but are you sure it doesnt still only apply the styles to textarea?

alisman: ah.  i didn't think about that.  i just figured it stripped it out.  but it probably does work how you say.

PR #2053, opened 2019-02-20
Integration of treatment response data in OncoPrint tab

pvannierop:
# What? Why?
This PR will implement handling of the new `treatment response` (genetic entity) data type by the OncoPrint component (see [this backend PR](https://github.com/cBioPortal/cbioportal/pull/5460)). The selection of treatments to be shown on the oncoprint does not occur on the Query page but in the heatmap menu. For this reason identifier validation as used for gene names on the query page does not support the user in entering valid identifiers. This PR will improve selection of treatments in the heatmap menu by allowing selection of treatments from a select box and by parsing identifiers typed by the user.

## Changes
- Treatment response profiles are shown in the heatmap menu (when available in cBioPortal db) (see Fig.1). When a treatment response value is selected, treatments can be added as a heatmap track by pressing the `Add Treatments to Heatmap` button.
- Treatment data points can be _truncated_ (real value lies beyond a certain value) which can be indicated by `>` or `<` prefixes to values. Heatmap track for treatment profiles are represented with a new legend type that shows continuous values (gradient) and truncated values (categories) (see Fig.2). Clustering of treatment heatmap tracks that contain truncated values used the truncated value for respective samples (for example _8_ represents truncated value _>8_). 
- Tooltips on a treatment heatmap track contain an url to external websites (when defined in cbioportal db) (see Fig.2).
- Entries in the treatment select box can be filtered for key-words (looks in _stable_id_, _name_ and _description_ fields for treatments) (see Fig.3). Individual treatments can be selected with check boxes. All treatments can be (de-)selected using buttons (see Fig.3).
- Selected treatments are displayed in the menu as icons (see Fig.4). Treatments can be removed by pressing the X-button in respective icons.
- Treatment identifiers can be typed in the free text area. Recognized treatments are added as icon above, and selected in the treatment selection menu (see Fig.4).

## Notes
- This PR should be merged after [this frontend PR](https://github.com/cBioPortal/cbioportal-frontend/pull/2051).
- A similar menu for genes can be implemented, but is not included in this PR.

### Figure 1
- Selection box for treatments is displayed when treatment profile is selected.
screenshot from 2019-02-20 13-46-05

### Figure 2
1. Treatment profile shown as heatmap track
2. New legend type with continuous and truncated values
3. URL in tooltip of treatment heatmap tracks
screenshot from 2019-02-20 13-16-24
 
### Figure 3
- Filtering of treatments in the selection box
- Selection of treatment via check box
- Buttons for selection and deselection of all treatments
screenshot from 2019-02-20 13-46-53

### Figure 4
- Selected treatments are show as icons
- Icons have a X-button used to remove a selected treatment
- A recognized treatment identifier in the text-area is added as a  
screenshot from 2019-02-20 13-47-43
screenshot from 2019-02-20 13-57-53

BACKEND_BRANCH=thehyve:treatment_study_implementation_rebased


src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -555,36 +648,60 @@ export function makeClinicalTracksMobxPromise(oncoprint:ResultsViewOncoprint, sa
 }
 
 export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sampleMode:boolean) {
```
adamabeshouse: why didn't you want to go the route that we did with genesets, and make it a separate category - treatmentHeatmapTracks, then have a different mobxpromise, etc? why do you want to conceptually tie gene heatmaps and treatment heatmaps?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -58,14 +58,16 @@ export interface IGenesetExpansionRecord {
...
 export const HEATMAP_TRACKS_URL_PARAM = "heatmap_track_groups";
+export const TREATMENT_LIST_URL_PARAM = "treatment_list";
 
 const CLINICAL_TRACK_KEY_PREFIX = "CLINICALTRACK_";
 
 type HeatmapTrackGroupRecord = {
     trackGroupIndex:number,
-    genes:ObservableMap<boolean>,
+    molecularAlterationType:string,
+    entities:ObservableMap<boolean>,
```
adamabeshouse: I really dont like storing genes and treatments in the same map - better to have one map for genes and one for treatments

src/shared/components/oncoprint/TooltipUtils.ts
```diff
@@ -97,18 +98,21 @@ export function makeHeatmapTrackTooltip(genetic_alteration_type:MolecularProfile
...
+            case AlterationTypeConstants.METHYLATION:
                 data_header = 'METHYLATION: ';
                 break;
+            case AlterationTypeConstants.TREATMENT_RESPONSE:
+                data_header = 'TREATMENT: ';
+                break;
         }
         if ((d.profile_data !== null) && (typeof d.profile_data !== "undefined")) {
-            profile_data = d.profile_data.toFixed(2);
+            profile_data = d.category || d.profile_data.toFixed(2);
```
adamabeshouse: this is really specific to TREATMENT_RESPONSE - you should do a check `if(TREATMENT_RESPONSE) { // what you have (line 115) } else {  // what it was before (line 111) }`

src/shared/components/textIconArea/TextIconArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+import { observable } from 'mobx';
+import autobind from "autobind-decorator";
+
+export interface ITextIconAreaProps {
+    elements: ITextIconAreaItemProps[];
+    text:string;
+    placeholder?: string;
+    classNames?: string[];
+    onItemRemove?: (d:string) => void;
+    onChange?: (s:string) => string;
```
adamabeshouse: can you make the parameters to onChange and onItemRemove more descriptive? is `d` the value of ITextIconAreaItemProps? (it should be) is `s` the text in the textarea?>

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -381,17 +404,118 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
         }
     }
 
+    @autobind
+    private onChangeEditableDiv(text:string):string {
+
+        // get values from input string
+        const elements = this.splitTextField(text);
```
adamabeshouse: I think this logic deserves unit tests - on how `text` is processed, and on `splitTextField`/which treatments are detected

env/custom.sh
```diff
@@ -1,2 +1,2 @@
-# set e.g.
-# export CBIOPORTAL_URL="http://localhost:8080"
\ No newline at end of file
+export CBIOPORTAL_URL="http://localhost:3000"
+export GENOME_NEXUS_URL="https://www.genomenexus.org"
```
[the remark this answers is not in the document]
pvannierop: It is in my .gitignore. I must have accidentally added custom.sh to the tracked files. I will remove it in my next commit.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1998,6 +2010,8 @@ export class ResultsViewPageStore {
                                 return 4;
```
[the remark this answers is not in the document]
pvannierop: All unselected values were increased by one so that there is never overlap between selected and unselected profiles.

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1104,7 +1104,8 @@ export function logScalePossible(
         // molecular profile
         return !!(
             axisSelection.dataSourceId &&
-            logScalePossibleForProfile(axisSelection.dataSourceId)
```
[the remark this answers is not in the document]
pvannierop: Ok, this relates to another PR that I have filed for changes to the PlotsTab that includes a rework of log-scale evaluation (treatment profiles can be log-transformed). Since I have split the project in multiple PRs this is one of the consequences of this. Would it be better to revert these lines into the original state?

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -280,7 +280,7 @@ export function fillHeatmapTrackDatum<T extends IBaseHeatmapTrackDatum, K extend
     featureKey: K,
     featureId: T[K],
     case_:Sample|Patient,
-    data?: {value: number}[]
+    data?: {value: number, truncation?: string}[]
```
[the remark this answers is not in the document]
pvannierop: Truncation is a (possibly poorly) chosen name for values that are larger or smaller than a given value (e.g, `>8`). The truncation value on data points represent the `>` or `<` symbols. Since I realize that truncation is a confusing word when regarded in genomics context, I am currently looking for a replacement term. When a more suitable term has been agreed on I am planning to rename all the `truncation` references in the code. A possibility might be `value_prefix_symbol` might be more suitable?

src/shared/components/oncoprint/DeltaUtils.spec.ts
```diff
@@ -512,4 +514,170 @@ describe("Oncoprint DeltaUtils", ()=>{
...
+
+            let oncoprint: OncoprintJS<any> = createStubInstance(OncoprintJS);
+            
+            // deacitvated because private method
+            //transitionTracks(prevProps, prevProps, oncoprint, trackSpecKeyToTrackId, molecularProfileMap);
+            
+            assert.equal(prevProps.heatmapTracks[0].maxProfileValue, 3);
+            assert.equal(prevProps.heatmapTracks[0].ruleSetTrackId, 1);
+            
+            // oncoprint = createStubInstance(OncoprintJS);
```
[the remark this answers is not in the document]
pvannierop: I am currently writing tests for this PR (see out teleconference of last week). These dead lines of code in the test file will be replaced when I revisit this code.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -388,9 +391,6 @@ function transitionTracks(
             } else if (heatmap === undefined) {
                 heatmap = trackSpecKeyToTrackId[spec.key];
             }
-            if (heatmap01 !== undefined && heatmap !== undefined) {
-                break;
```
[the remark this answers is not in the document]
pvannierop: Wow well spotted, I think this is a remnant of the rework efforts at some time. Indeed this loop may terminate when both rule sets were identified. I have re-added the deleted lines.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -454,17 +471,30 @@ function transitionTracks(
     }
 
     // Transition heatmap tracks
-    const prevHeatmapTracks = _.keyBy(prevProps.heatmapTracks || [], track=>track.key);
-    for (const track of nextProps.heatmapTracks) {
+    const prevHeatmapTracks = _.keyBy(prevProps.heatmapTracks || [], (track:IHeatmapTrackSpec)=>track.key);
+    for (let track of nextProps.heatmapTracks) {
+
+        // add treatment layout/formatting information to the track specs
+        track['maxProfileValue'] = treatmentProfileMaxValue[track.molecularProfileId];
```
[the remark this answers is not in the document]
pvannierop: Ha ha, I started learning Python at the same time as JS (3 months ago). I guess I have mixed things up here. I have replaced the string indexes with dot notations.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -869,28 +901,57 @@ function transitionHeatmapTrack(
...
+        // register new track in oncoprint
+        const newTrackId:number = oncoprint.addTracks([heatmapTrackParams])[0];
+        // store relation between React heatmap track specs and OncoprintJS trackIds
         trackSpecKeyToTrackId[nextSpec.key] = newTrackId;
 
-        let trackIdForRuleSetSharingKey:"heatmap"|"heatmap01" = "heatmap";
-        if (nextSpec.molecularAlterationType === "METHYLATION") {
+        let trackIdForRuleSetSharingKey:"heatmap"|"heatmap01"|undefined;
+
+        if (nextSpec.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE) {
```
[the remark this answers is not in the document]
pvannierop: Implemented as per your suggestion.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -869,28 +901,57 @@ function transitionHeatmapTrack(
...
+        // register the new data points in oncoprint
         if (nextSpec.data !== prevSpec.data) {
             // shallow equality check
             oncoprint.setTrackData(trackId, nextSpec.data, "uid");
         }
         if (nextSpec.info !== prevSpec.info && nextSpec.info !== undefined) {
             oncoprint.setTrackInfo(trackId, nextSpec.info);
         }
+        // re-register the shared rule set (when defined) when reloading the page 
+        // if (nextSpec.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE) {
```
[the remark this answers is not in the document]
pvannierop: Dead line of code was removed. The comment was changed to:

```
// re-register the shared rule set
```
Can you live with this?

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -392,4 +392,27 @@ describe('OncoprintUtils', () => {
...
+                    {profile_data: 1, study: "study1", uid: "uid"}, 
+                    {profile_data: 2, study: "study1", uid: "uid"}, 
+                    {profile_data: 3, study: "study1", uid: "uid"}
+                ],
+                datatype: "TREATMENT_RESPONSE",
+                trackGroupIndex: 1,
+                onRemove: () => {}
+            };
+
+            // const ruleSetParams = getTreatmentTrackRuleSetParams(treatmentTracSpec);
```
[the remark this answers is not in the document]
pvannierop: I am currently writing tests for this PR (see out teleconference of last week). These dead lines of code in the test file will be replaced when I revisit this code.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -139,12 +146,17 @@ export function doWithRenderingSuppressedAndSortingOff(oncoprint:OncoprintJS<any
...
-export function getHeatmapTrackRuleSetParams(molecularAlterationType: string) {
+export function getHeatmapTrackRuleSetParams(trackSpec: IHeatmapTrackSpec):RuleSetParams {
     let value_range:[number, number];
     let legend_label:string;
-    let colors:number[][];
+    let colors:[number, number, number, number][];
     let value_stop_points:number[];
-    if (molecularAlterationType === "METHYLATION") {
+
+    if (trackSpec.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE) {
```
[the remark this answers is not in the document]
pvannierop: Implemented as per your suggestion.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: This is something we have considered. For us treatment profiles are just another layer of numerical properties that can be represented for sampes/patients in the oncoprint. We appreciate that treatments do not relate to genes. What we wanted to achieve, however, is a seamless mix of molecular and treament profiles that can be selected in sequence by the user and are represented in this order. This is different behavior than that of geneset tracks that are 1) selected from the query page and 2) represented always at the bottom of the heatmap. In the backend, molecular and treatment profiles are all stored in the `genetic_profile` table. All together having one mobxpromise for all profile data in the selected studies seemed like the most straight forward approach.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: I personally disagree. The scope of `genetic_profiles` is expanding as cBioPortal evolves (genes->mRNA->protein, gene sets, treatments, ...). Heatmap tracks can be any of these so I think it is better to abandon the separation based on type. This also prevents the rather arbitrary priority that is imposed by cBioPortal relating to representation of heatmap tracks that are of different type w/o much complication. I assume that when selecting a treatment profile then a molecular profile that the user expects the treatment is show on top. When separating all profiles into different track group maps this is not the behavior that is designed for. When allowing other profile types besides molecular, I would argue to combine them.

src/shared/components/oncoprint/TooltipUtils.ts
pvannierop: Ok, added a condition for treatment data.

src/shared/components/textIconArea/TextIconArea.tsx
pvannierop: I have made the callback names and arguments more clear:
>    onIconClicked?: (itemId:string) => void;
>    onChangeTextArea?: (textAreaContents:string) => string;

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: I fully agree. However, these methods are private and not accessible to any test that I write. Can you provide a suggestion on how to do this?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: In this particular case, the issue is that gene identifiers and treatment identifiers don't live in the same universe. So I find it hacky to use them as keys in the same map.

src/shared/components/textIconArea/TextIconArea.tsx
adamabeshouse: could you make it `itemValue` so it corresponds with the member name on  `ITextIconAreaItemProps` ? That's what it is, right?

src/shared/components/oncoprint/controls/OncoprintControls.tsx
adamabeshouse: The way I usually do this is pull it out into a helper function which we put in a file called, say, OncoprintControlsUtils.ts. That function is exported, and would take the minimum possible data as arguments, for ease of testing. So for example, it wouldn't take the whole OncoprintControls object, it would just take the value of `this._selectedTreatmentOptions`, `this.treatmentOptionsByValueMap`.

splitTextField is already a pure function so you could just move that into utils entirely and not put it as a class method.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -156,11 +170,15 @@ const EVENT_KEY = {
...
     horzZoomSlider:"30",
+    addTreatmentsToHeatmap: "32"
 };
 
 @observer
 export default class OncoprintControls extends React.Component<IOncoprintControlsProps, {}> {
+
     @observable horzZoomSliderState:number;
+    @observable _selectedTreatmentOptions:ISelectOption[] = [];
+    textareaTreatmentText:string = "";
```
adamabeshouse: could you make `private` explicit?

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
adamabeshouse: Yes please, since logScalePossibleForProfile literally does the exact computation.

src/shared/components/oncoprint/DataUtils.ts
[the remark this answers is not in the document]
adamabeshouse: ah, okay. How about `thresholdType: ">" | "<"`. That would make me fully understand if I was reading it.

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: To me, the main reason to keep them separate is to not separate them into different track groups - personally, I don't see a reason to tie our frontend ontology to the way the data happens to be kept in the database. I think that makes sense. 

So I think what I would want is for you to please put the functionality in `makeHeatmapTracksMobxPromise` into helper functions - one that handles genes, one that handles treatments. It's a little confusing as it is. What do you think?

src/shared/components/textIconArea/TextIconArea.tsx
```diff
@@ -0,0 +1,112 @@
...
+import ReactDOM from 'react-dom';
+import { observable } from 'mobx';
+import autobind from "autobind-decorator";
+
+export interface ITextIconAreaProps {
+    elements: ITextIconAreaItemProps[];
+    text:string;
+    placeholder?: string;
+    classNames?: string[];
+    onIconClicked?: (itemId:string) => void;
```
adamabeshouse: I left a comment on the previous commit, I'll add it here again because Im not sure how github will handle it/if it will be easy to see:

Could you change `itemId` to `itemValue` so that it's more clearly indicates what it is? It corresponds to `ITextIconAreaItemProps.value`, right?

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
pvannierop: I looked into the logScalePossibleForProfile() function. The reason for isolating the logTransformation logic into a separate function is that the CoExpressionViz.tsx component duplicated this logic in the showLogScaleControls() function. I think therefore having an isolated  logScalePossibleForProfile function is justified, also considering that treatment profiles will be added to this logic. Possible the logScalePossibleForProfile() function should move to a more suited class? Could you propose one?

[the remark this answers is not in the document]
adamabeshouse: Sorry, I'm confused -- I don't understand why we'd make this change here since it's literally the same logic. Are you planning to change the functionality of logScalePossibleForProfile?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: I have refactored the MobXPromisses for molecular profiles and treatment profiles. However, I am not quite understanding what your proposal is regarding your last comment:
>In this particular case, the issue is that gene identifiers and treatment identifiers don't live in the same universe. So I find it hacky to use them as keys in the same map.
The `HeatmapTrackGroupRecord` type is used by the molecularProfileIdToHeatmapTracks observable map of ResultsViewOncoprint. This map is keyed by molecular profile id:

`this.molecularProfileIdToHeatmapTracks.set(molecularProfileId, trackGroup);`

Although the molecular and treatment identifiers live in separate spaces, the molecular profile identifiers that are used for keying do live in the same space. I would be more that happy to accomodate your whish, but I am uncertain what your conceptual problem is and how I should solve it.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: I have refactored the MobXPromisses for molecular profiles and treatment profiles. Now there are two methods for molecular and treatment heatmap tracks:
- ```function makeMolecularProfileHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sampleMode:boolean) {```
and 
- ```function makeTreatmentProfileHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sampleMode:boolean) {```

The `makeHeatmapTracksMobxPromise` has been refactored to use the two new partial heatmap track methods:

```
export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sampleMode:boolean) {

    const moleculePromise = makeMolecularProfileHeatmapTracksMobxPromise(oncoprint, sampleMode);
    const treatmentPromise = makeTreatmentProfileHeatmapTracksMobxPromise(oncoprint, sampleMode);

    return remoteData<IHeatmapTrackSpec[]>({
        await:()=>[
            moleculePromise,
            treatmentPromise
        ],
        invoke:async()=>{
            return (moleculePromise.result!).concat(treatmentPromise.result!);
        },
        default: []
    });
}
```

src/shared/components/oncoprint/DataUtils.ts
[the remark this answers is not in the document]
pvannierop: The new name thresholdType has been applied throughout the code as well as the type restriction (>" | "<").

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
pvannierop: What I mean is that the logic of logScalePossible():

```
return !(/zscore/i.test(profileId)) &&
        /rna_seq/i.test(profileId);
```
is duplicated in the CoExpressionViz.tsx component. What I did was isolate the logic into the logScalePossibleForProfile() helper function and used this helper function in CoExpressionViz.tsx:

```    
@computed get showLogScaleControls() {
        const profileXId = this.props.profileX.molecularProfileId.toLowerCase();
        const profileYId = this.props.profileY.molecularProfileId.toLowerCase();
        return (
            (logScalePossibleForProfile(profileXId) || logScalePossibleForProfile(profileYId))
        );
    }
```
The is refactoring is even more important considering that will update the logScalePossible() logic for treatments in the plots tab. So the center of the debate is that your statement:

>...and this function and logScalePossibleForProfile are both in the same file. 

is incorrect and that I should answer your question:
>Are you planning to change the functionality of logScalePossibleForProfile?

with "yes".

src/shared/components/textIconArea/TextIconArea.tsx
pvannierop: You are correct in that it points to the `value` field of the `ITextIconAreaItemProps` interface. I have Updated the name to refect this. Good suggestion.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: Extraction of helper functions was performed and unit test are included. Thanks for the push in the right direction :)

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: Done...

src/shared/components/textIconArea/TextIconArea.tsx
pvannierop: Done...

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
adamabeshouse: I see `logScalePossibleForProfile` in this file on line 1095. You may not have realized that was there, which is why there is confusion. In light of that, do you still want to write out the logic this way?

src/shared/components/oncoprint/Oncoprint.tsx
```diff
@@ -108,16 +114,24 @@ interface IBaseHeatmapTrackSpec {
...
+    data: IBaseHeatmapTrackDatum[]; // can be IGeneHeatmapTrackDatum or ITreatmentHeatmapTrackDatum
     info?: string;
     labelColor?: string;
+    trackLinkUrl?: string | undefined;
+    onRemove: () => void;
+    molecularProfileName?: String
+    pivotThreshold?: number;
+    sortOrder?: SortOrder;
+    maxProfileValue?: number;
+    ruleSetTrackId?: number;
```
adamabeshouse: sorry I just noticed this. but ruleSetTrackId doesn't belong here. Users of the oncoprint shouldn't have to specify that, it should be automatically handled. Maybe you are trying to spedcify wheher its a gene heatmap or a treatment heatmap? In that case please add an enum for that and have the user specify that, then we'll handle the ruleset sharing the same way we have been in existing code.

src/shared/components/oncoprint/Oncoprint.tsx
```diff
@@ -108,16 +114,24 @@ interface IBaseHeatmapTrackSpec {
...
+    category?:string;
```
adamabeshouse: what's the purpose of this prop?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: Sorry, I should be more specific: it's the `entities` property that bothered me. I would prefer if you split that up into `genes` and `treatments`. However, I'm now understanding that for a single HeatmapTrackGroupRecord you would only have genes, or treatments, which I didn't understand before. In that case, it's fine this way, just please add a comment next to it indicating what specifically it's a map of - for genes, hugo gene symbols?  for treatments, treatment id?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -607,28 +625,48 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
-    private addHeatmapTracks(molecularProfileId:string, genes:string[]) {
-        let trackGroup = this.molecularProfileIdToHeatmapTracks.get(molecularProfileId);
-        if (!trackGroup) {
-            let newTrackGroupIndex = 2;
-            for (const group of this.molecularProfileIdToHeatmapTracks.values()) {
-                newTrackGroupIndex = Math.max(newTrackGroupIndex, group.trackGroupIndex + 1);
+    private addHeatmapTracks(molecularProfileId:string, entities:string[]) {
+        
+        let profile:MolecularProfile = this.props.store.molecularProfileIdToMolecularProfile.result![molecularProfileId];
+        // TODO fix the Promise dependency in this function (remove if profile statement)
```
adamabeshouse: TODO?

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
pvannierop: I have reverted the log-scale evaluation to the rc state.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: Comment added.

src/shared/components/oncoprint/Oncoprint.tsx
pvannierop: I have removed the `ruleSetTrackId` member of the `IHeatmapTrackSpec` interface. I have replaced this with an extension of the `trackIdForRuleSetSharing` map in `transitionTracks()`:
```
    const trackIdForRuleSetSharing = {
        genetic: undefined as undefined|TrackId,
        genesetHeatmap: undefined as undefined|TrackId,
        heatmap: undefined as undefined|TrackId,
        heatmap01:undefined as undefined|TrackId,
        treatment: {} as any as TreatmentProfileToTrackIdMap  <----
    };
```
I think this solves your problem with attaching the ruleSetTrackId to Oncoprint.

src/shared/components/oncoprint/Oncoprint.tsx
pvannierop: `sortOrder` controls the direction of the track legend. With SortOrder.ASC the smallest value is shown at the left side and gets the 'significant' color. With SortOrder.DESC the larges  value is shown at the left side and gets the 'significant' color.

`maxProfileValue` and `minProfileValue` (recently added in new commit) represent the max and min values all tracks of a trackGroup. This information is added because only a single track controls the legend, but the legend is dynamically drawn based on all tracks in the group.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: No, this was a leftover. I have removed the TODO comment.

src/shared/components/oncoprint/Oncoprint.tsx
adamabeshouse: that's exactly what I was hoping for, thanks

src/shared/components/oncoprint/Oncoprint.tsx
adamabeshouse: sorry, I should have been more specific: I meant the `category` prop

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -405,10 +414,30 @@ function transitionTracks(
...
-
+    
+    // collect max trackId (last assigned one) for each treatment profile
+    // the last assigned track id is used to reference the current layout of the track group
+    // Note: the resolution of `trackIds for ruleset sharing` is different from
+    // the section above because different formatting is applied to each treatment profile (molecularProfileId)
+    trackIdForRuleSetSharing.treatment = _.chain(prevProps.heatmapTracks)
+                                    .filter((s:IHeatmapTrackSpec) => s.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE)
+                                    .groupBy((track:IHeatmapTrackSpec) => track.molecularProfileId)
+                                    .mapValues( (o:IHeatmapTrackSpec[]) => _(o).map('key').map((key:string) => trackSpecKeyToTrackId[key]).max() )
```
adamabeshouse: we typically dont like to use the syntax of `.map("key")`, because I don't think typescript catches it. We use `.map(d=>d.key)`. Also on line 431

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -870,28 +912,62 @@ function transitionHeatmapTrack(
...
+
+        switch (nextSpec.molecularAlterationType) {
+            case AlterationTypeConstants.TREATMENT_RESPONSE:
+
+                trackIdForRuleSetSharingKey = undefined;
+        
+                // register the new track id so that it will determine formatting of the track group
+                const rulesetTrackId = trackIdForRuleSetSharing.treatment![nextSpec.molecularProfileId];
+                if (rulesetTrackId) {
+                    oncoprint.shareRuleSet(newTrackId, rulesetTrackId);
```
adamabeshouse: reverse arguments - the source track id goes first
 `shareRuleSet:(source_track_id:TrackId, target_track_id:TrackId)=>void;`

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -870,28 +912,62 @@ function transitionHeatmapTrack(
...
             // shallow equality check
             oncoprint.setTrackData(trackId, nextSpec.data, "uid");
         }
         if (nextSpec.info !== prevSpec.info && nextSpec.info !== undefined) {
             oncoprint.setTrackInfo(trackId, nextSpec.info);
         }
+        // re-register the shared rule set
+        const rulesetTrackId = trackIdForRuleSetSharing.treatment![nextSpec.molecularProfileId];
+        if (rulesetTrackId) {
+            oncoprint.shareRuleSet(rulesetTrackId, trackId);
```
adamabeshouse: (1) why?
(2) this code applies for gene heatmaps as well - methylation, mrna, protein, so it's going to break non-treatment heatmaps

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: thats perfect, thanks

end-to-end-tests/create_localdb.sh
```diff
@@ -0,0 +1,170 @@
+#!/usr/bin/env bash
```
adamabeshouse: please make sure you get a review from someone else about all the non-js build related stuff, because I have no idea

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
adamabeshouse: I still see it expanded here on lines 1107 and 1108

src/shared/api/generated/CBioPortalAPI.ts
```diff
@@ -2357,6 +2436,11 @@ export default class CBioPortalAPI {
         }): Promise < Array < MolecularProfile >
         > {
             return this.fetchMolecularProfilesUsingPOSTWithHttpInfo(parameters).then(function(response: request.Response) {
+                response.body = _(response.body).each((d:MolecularProfile) => {
+                    if (d.sortOrder !== undefined) {
+                        d.sortOrder = d.sortOrder as any === "ASC"? SortOrder.ASC : SortOrder.DESC;
```
adamabeshouse: did you hand write this?

adamabeshouse: seems like you might want to move these and your changes above into another file in /api/ but not /generated/, like `TreatmentAPI` or something

src/shared/api/generated/CBioPortalAPIInternal.ts
```diff
@@ -270,6 +274,50 @@ export type GenesetMolecularData = {
 
         'value': string
 
+};
+export type Treatment = {
+    'treatmentId': string
+
+    'name': string
```
adamabeshouse: sorry I'm actually not sure whether these changes are handwritten or not. if they are, you'll need to move them into another file as I mentioned above. if not, all good

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -166,6 +182,86 @@ export function getHeatmapTrackRuleSetParams(molecularAlterationType: string) {
...
+    const dataPoints = trackSpec.data;
+    const pivotThreshold = trackSpec.pivotThreshold!;
+    const sortOrder = trackSpec.sortOrder!;
+    
+    const colorWorse  = [0,0,255,1] as [number, number, number, number];
+    const colorPivot  = [0,0,0,1] as [number, number, number, number];
+    const colorBetter = [255,0,0,1] as [number, number, number, number];
+    const categoryColorOptions = [ 'rgba(230,159,0,1)', 'rgba(0,158,115,1)', 'rgba(204,121,167,1)', 'rgba(240,228,66,1)', 'rgba(86,180,233,1)' ];
+
+    const maxValue:number = trackSpec.maxProfileValue || _(dataPoints as ITreatmentHeatmapTrackDatum[]).filter((d:ITreatmentHeatmapTrackDatum) => !d.category).map('profile_data').max()!;
```
adamabeshouse: would you ever get here and not have it already calculated?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -166,6 +182,86 @@ export function getHeatmapTrackRuleSetParams(molecularAlterationType: string) {
...
+    const minValue:number = trackSpec.minProfileValue || _(dataPoints as ITreatmentHeatmapTrackDatum[]).filter((d:ITreatmentHeatmapTrackDatum) => !d.category).map('profile_data').min()!;
+    const categories = _(dataPoints as ITreatmentHeatmapTrackDatum[]).filter((d:ITreatmentHeatmapTrackDatum) => !!d.category).map('category').uniq().value();
+
+    // when all observed values are negative or positive
+    // assume that 0 should be used in the legend
+    const rightBoundaryValue = maxValue <= 0? 0 : maxValue;
```
adamabeshouse: Math.max(0, maxValue)

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -166,6 +182,86 @@ export function getHeatmapTrackRuleSetParams(molecularAlterationType: string) {
...
+    const leftBoundaryValue = minValue >= 0? 0 : minValue;
```
adamabeshouse: Math.min(0, minValue)

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
pvannierop: I am losing track of your intentions with this comment. These lines are what is currently in rc branch. How can your comment on the expanded line relate to my PR?

src/shared/components/oncoprint/Oncoprint.tsx
pvannierop: Ah I see. Treatment profile data points in essence consist of a value (e.g. '8.00') and an optional  'thresholdType' (e.g. '>'). For Oncoprint trackGroups data points that have a thresholdType will have a `category` member that is a concatenation between the rounded value prefixed with the thresholdType (e.g., '>8.00'). This `category` member is used by oncoprintJS to draw the recently implemented 'gradient+categorical' legend type.

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: In my dev setup it works as specified on the lodash docs. You probably are aware that this `map()` function is the lodash map and not the standard JS version? Should I go over all my changes and revert back all similar cases of `map('key')`?

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Ah yes, you are right. This reverse argument does not do what I thought it would do. I would like to point out that there is a reverse logic for treatment profile tracks compared to others. I have looked at the code again and refactored to make this logic more clear. In essence the difference is:
- molecular profiles: a new molecular profile track is linked with the rule set of an existing track via the trackid of the existing set. If its the first track it is added to `trackIdForRuleSetSharing` so that others can link to it. The ruleSet association of existing tracks is never updated (link to the first track that was added). 
- treatment profiles: a new treatment profile track is always added to `trackIdForRuleSetSharing` so that others can link to it. The ruleSet association of existing treatment tracks is always updated to the track that was added last.

The reason for this different logic is that legends/ruleSets for treatments are based on an aggregation of data in all tracks in the trackGroup. The last added track holds up-to-date information on all tracks (includes itself). When this track is added the styling of all existing tracks are updated to point to the ruleSet defined by the last added track.

I think that this reverse logic in not necessary. When all track types would link to the ruleSet of the last added track irrespective of the data type this logic can be removed and the functionality is identical.

What do you think?

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: (see comment above)

> Ah yes, you are right. This reverse argument does not do what I thought it would do. I would like to point out that there is a reverse logic for treatment profile tracks compared to others. I have looked at the code again and refactored to make this logic more clear. In essence the difference is:
> 
> molecular profiles: a new molecular profile track is linked with the rule set of an existing track via the trackid of the existing set. If its the first track it is added to trackIdForRuleSetSharing so that others can link to it. The ruleSet association of existing tracks is never updated (always link to the first track that was added).
> treatment profiles: a new treatment profile track is always added to trackIdForRuleSetSharing so that others can link to it. The ruleSet association of existing treatment tracks is always updated to the track that was added last.
> The reason for this different logic is that legends/ruleSets for treatments are based on an aggregation of data in all tracks in the trackGroup (data is not standardized f.i. between -3 and 3 Z-scores). The last added track holds up-to-date information on all tracks in the group (includes information on itself). When this track is added the styling of all existing tracks are updated to point to the ruleSet defined by the last added track.
> 
> I think that this reverse logic in not necessary. When all track types would link to the ruleSet of the last added track irrespective of the data type this logic can be removed and the functionality is identical.
> 
> What do you think?

end-to-end-tests/create_localdb.sh
pvannierop: Sigh, I have no idea how this piece of code ended up here. I still mess up quite regularly with git. I have removed this code since not relevant for this PR.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: Yes! Much more elegant.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: Yes! Much more elegant.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: Eh no, that never happens in the current code. I have removed this construct since I guess that is what you would suggest if I answer this.

src/shared/api/generated/CBioPortalAPIInternal.ts
pvannierop: Oh really? I did not know. Since these files are in src/ I did not suspect this. I will rewrite all the code.

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
adamabeshouse: Look at lines 1107 and 1108 - they are green, indicating they have been added by this PR. Im asking that you undo all changes in this file `PlotsTabUtils.tsx`

src/shared/components/oncoprint/DeltaUtils.ts
adamabeshouse: image

image

adamabeshouse: the functionality works, but typescript wont catch misspelled keys

src/shared/components/oncoprint/Oncoprint.tsx
adamabeshouse: I understand. But why is the category prop in the HeatmapTrackSpec?

src/shared/components/oncoprint/TooltipUtils.ts
```diff
@@ -123,34 +124,72 @@ export function makeClinicalTrackTooltip(track:ClinicalTrackSpec, link_id?:boole
...
+            } else if (profileDataCount > 0 && categoryCount === 1) {
+                categoryTextElement = `${profileCategories[0]}`;
+            } else {
+                categoryTextElement = `${_.uniq(profileCategories).join(", ")} (${categoryCount} data points)`;
             }
         }
-        let ret = data_header + '<b>' + profile_data + '</b><br/>';
+
+        let ret = data_header;
+        if (valueTextElement !== 'N/A' || categoryCount === 0) {
```
adamabeshouse: it would be better to pull "N/A" out into a constant and use it both here and in the place where you construct your data. It gets messy to deal with checking for string equality everywhere.

src/pages/resultsView/plots/PlotsTabUtils.tsx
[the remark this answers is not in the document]
pvannierop: Sorry for this ... I do not know why this was the case. Lines reverted back  to rc state.

src/shared/components/oncoprint/Oncoprint.tsx
pvannierop: Was a remnant of earlier attempts. Field was removed.

src/shared/api/generated/CBioPortalAPI.ts
pvannierop: Treatment API was replaced by swagger CodeGen output.

src/shared/api/generated/CBioPortalAPIInternal.ts
pvannierop: Treatment API was replaced by swagger CodeGen output.

scripts/env_vars.sh
```diff
@@ -5,16 +5,28 @@ RED='\033[0;31m'
 NC='\033[0m'
 
 if [[ "$CIRCLECI" ]]; then
-    # on circle ci determine env variables based on branch or in case of PR
-    # what branch the PR is pointing to
-    if [[ "$CIRCLE_PR_NUMBER" ]] && ! [[ $CIRCLE_BRANCH == "release-"* ]]; then
-        BRANCH=$(curl "https://github.com/cBioPortal/cbioportal-frontend/pull/${CIRCLE_PR_NUMBER}" | grep -oE 'title="cBioPortal/cbioportal-frontend:[^"]*' | cut -d: -f2 | head -1)
-    elif [[ "$CIRCLE_PULL_REQUEST" ]] && ! [[ $CIRCLE_BRANCH == "release-"* ]]; then
-        BRANCH=$(curl "${CIRCLE_PULL_REQUEST}" | grep -oE 'title="cBioPortal/cbioportal-frontend:[^"]*' | cut -d: -f2 | head -1)
+
```
adamabeshouse: should this file be here?

src/shared/components/oncoprint/DeltaUtils.spec.ts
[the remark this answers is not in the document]
adamabeshouse: just making sure this is still on the agenda

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -601,8 +715,90 @@ export function makeHeatmapTracksMobxPromise(oncoprint:ResultsViewOncoprint, sam
...
+                const molecularProfileId = query.molecularProfileId;
+                const profile = molecularProfileIdToMolecularProfile[molecularProfileId];
+                const dataCache = oncoprint.props.store.treatmentMolecularDataCache.result!;
+
+                const treatmentId = query.treatmentId;
+                const pivotThreshold = profile.pivotThreshold;
+                const sortOrder = profile.sortOrder;
+
+                return {
+                    key: `HEATMAPTRACK_${molecularProfileId},${treatmentId}`,
```
adamabeshouse: would be better to prefix with `TREATMENTHEATMAPTRACK` in case theres some unlikely overlap between treatment ids and gene ids

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: I have changed all occurences of the .map('key') pattern.

scripts/env_vars.sh
pvannierop: No, reverted back to rc.

src/shared/components/oncoprint/DeltaUtils.spec.ts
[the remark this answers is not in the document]
pvannierop: If you agree that transitionTracks will be exported from DeltaUtils.ts I can implement the unit test. Or is there a reflection method that I can use to access it?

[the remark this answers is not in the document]
adamabeshouse: yes, fine to export transitionTracks

src/shared/components/oncoprint/DeltaUtils.ts
adamabeshouse: Thank you for the clear explanation. Yes, I agree with your conclusion.

adamabeshouse: So I'm still confused why you're only aggregating max/min over each profile, is that how it's done? You have a separate ruleset per-profile?

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -405,10 +415,30 @@ function transitionTracks(
...
-
+    
+    // collect max trackId (last assigned one) for each treatment profile
+    // the last assigned track id is used to reference the current layout of the track group
+    // Note: the resolution of `trackIds for ruleset sharing` is different from
+    // the section above because different formatting is applied to each treatment profile (molecularProfileId)
+    trackIdForRuleSetSharing.treatment = _.chain(prevProps.heatmapTracks)
+                                    .filter((s:IHeatmapTrackSpec) => s.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE)
+                                    .groupBy((track:IHeatmapTrackSpec) => track.molecularProfileId)
+                                    .mapValues( (o:IHeatmapTrackSpec[]) => _(o).map((d)=>d.key).map((key:string) => trackSpecKeyToTrackId[key]).max() )
```
adamabeshouse: It's not technically part of the oncoprint API that more recent tracks are assigned higher ids

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -870,28 +913,51 @@ function transitionHeatmapTrack(
...
 
-        let trackIdForRuleSetSharingKey:"heatmap"|"heatmap01" = "heatmap";
-        if (nextSpec.molecularAlterationType === "METHYLATION") {
-            trackIdForRuleSetSharingKey = "heatmap01";
-        }
-        if (typeof trackIdForRuleSetSharing[trackIdForRuleSetSharingKey] !== "undefined") {
-            oncoprint.shareRuleSet(trackIdForRuleSetSharing[trackIdForRuleSetSharingKey]!, newTrackId);
+        if (nextSpec.molecularAlterationType !== AlterationTypeConstants.TREATMENT_RESPONSE) {
+            let rulesetTrackIdKey:"heatmap"|"heatmap01";
+            if (nextSpec.molecularAlterationType !== AlterationTypeConstants.METHYLATION) {
```
adamabeshouse: heatmap01 should be for methylation, you have it reversed

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -870,28 +913,51 @@ function transitionHeatmapTrack(
...
+            } else {
+                rulesetTrackIdKey = "heatmap";
+            }
+            // if the track is a molecular profile, check whether there is an existing ruleset that applies to this track
+            if (trackIdForRuleSetSharing[rulesetTrackIdKey] !== undefined) {
+                // if so, associate the is of the new track so that its is formatted by the existing ruleset
+                oncoprint.shareRuleSet(trackIdForRuleSetSharing[rulesetTrackIdKey]!, newTrackId);
+            } else {
+                // if not, make this track available to other new tracks for ruleset sharing
+                trackIdForRuleSetSharing[rulesetTrackIdKey] = newTrackId;
```
adamabeshouse: there is no `else` in the original logic, have you added it here on purpose? this line should happen unconditionally

src/shared/components/oncoprint/DeltaUtils.ts
adamabeshouse: It seems like you shouldn't bother setting these inside each transitionTrack as well, you should probably set them all at once at the end, right?

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: No, the max and min values are evaluated per all tracks in a molecular profile (in DeltaUtils.ts):
```
    const treatmentProfilesMap = _.chain(nextProps.heatmapTracks)
                                    .filter((s:IHeatmapTrackSpec) => s.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE)
                                    .groupBy((track:IHeatmapTrackSpec) => track.molecularProfileId)
                                    .mapValues( (o:IHeatmapTrackSpec[]) => _(o).flatMap((d)=>d.data).filter((d:IBaseHeatmapTrackDatum) => ! d.category).map((d)=>d.profile_data).value() )
                                    .value();

    // find the max and min treatment profile value in the next heatmap track group
    // max and min value is used to create a custom legend for the track group
    const treatmentProfileMaxValues = _.mapValues(treatmentProfilesMap, (profile_data:number[]) => { return _.max(profile_data) });
    const treatmentProfileMinValues = _.mapValues(treatmentProfilesMap, (profile_data:number[]) => { return _.min(profile_data) });
```
These values are attached to each trackspec and passed on to `transitionHeatmapTrack`:
```
        // add treatment layout/formatting information to the track specs
        track.maxProfileValue = treatmentProfileMaxValues[track.molecularProfileId];
        track.minProfileValue = treatmentProfileMinValues[track.molecularProfileId];

        transitionHeatmapTrack(track, prevHeatmapTracks[track.key], getTrackSpecKeyToTrackId,
            () => undefined, oncoprint, nextProps, {}, trackIdForRuleSetSharing, 
            undefined);
        delete prevHeatmapTracks[track.key];
```

src/shared/components/oncoprint/TooltipUtils.ts
pvannierop: Ok, changed

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: Yes, prefix added.

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Well spotted! the !== was changed in ===.

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Hmm, the code in the most recent version is as this:
```
        if (nextSpec.molecularAlterationType !== AlterationTypeConstants.TREATMENT_RESPONSE) {
            let rulesetTrackIdKey:"heatmap"|"heatmap01";
            if (nextSpec.molecularAlterationType === AlterationTypeConstants.METHYLATION) {
                rulesetTrackIdKey = "heatmap01";
            } else {
                rulesetTrackIdKey = "heatmap";
            }
            // if the track is a molecular profile, check whether there is an existing ruleset that applies to this track
            if (trackIdForRuleSetSharing[rulesetTrackIdKey] !== undefined) {
                // if so, associate the is of the new track so that its is formatted by the existing ruleset
                oncoprint.shareRuleSet(trackIdForRuleSetSharing[rulesetTrackIdKey]!, newTrackId);
            } else {
                // if not, make this track available to other new tracks for ruleset sharing
                trackIdForRuleSetSharing[rulesetTrackIdKey] = newTrackId;
            }
        } else {
            // if the track is a treatment profile, always add to trackIdForRuleSetSharing
            // this makes the trackId available for existing tracks fir ruleset sharing
            trackIdForRuleSetSharing.treatment![nextSpec.molecularProfileId] = newTrackId;
        }
```

You can see that this `else` is for treatment tracks. I did add a comment for this to clarify this point.

adamabeshouse: In `rc`, `trackIdForRuleSetSharing[trackIdForRuleSetSharingKey] = newTrackId` runs unconditionally, as I indicated in my original comment.

src/shared/components/oncoprint/DeltaUtils.spec.ts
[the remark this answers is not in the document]
pvannierop: Unit test for DeltaUtils --> transitionHeatmapTrack are now included.

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Ok, I think I forgot to fetch. My change is now updated to reflect current `rc`. Moreover, this logic is now covered by `DeltaUtil.spec.ts`.

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Ok, I have refactored the evaluation of the trackID for the last added track. Instead of the max value I find the last added in the sequence:
```
    trackIdForRuleSetSharing.treatment = _.chain(prevProps.heatmapTracks)
                                    .filter((s:IHeatmapTrackSpec) => s.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE)
                                    .groupBy((track:IHeatmapTrackSpec) => track.molecularProfileId)
                                    .mapValues( (o:IHeatmapTrackSpec[]) => _.last(o) )
                                    .mapValues( (o:IHeatmapTrackSpec) => trackSpecKeyToTrackId[_.get(o, 'key')] )
                                    .value();
```
This implementation relies on preservation of the order in the lodash `groupBy` function. This is a safe assumption because part of the function description:
```
Creates an object composed of keys generated from the results of running each element of collection thru iteratee. The order of grouped values is determined by the order they occur in collection. The corresponding value of each key is an array of elements responsible for generating the key. The iteratee is invoked with one argument: (value).
```

pvannierop: I do not really understand this comment:

>It seems like you shouldn't bother calling shareRuleSet these inside each transitionTrack as well, you should probably set them all at once at the end, right?

Do you mean that I should suppress calling `shareRuleSet` for treatment tracks in the `transitionHeatmapTrack()` function and call id after the `// Transition heatmap tracks` block in `transitionTracks`?

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -416,14 +416,14 @@ function transitionTracks(
...
-    // the last assigned track id is used to reference the current layout of the track group
+    // collect trackId of last assigned track for each treatment profile
     // Note: the resolution of `trackIds for ruleset sharing` is different from
     // the section above because different formatting is applied to each treatment profile (molecularProfileId)
     trackIdForRuleSetSharing.treatment = _.chain(prevProps.heatmapTracks)
                                     .filter((s:IHeatmapTrackSpec) => s.molecularAlterationType === AlterationTypeConstants.TREATMENT_RESPONSE)
                                     .groupBy((track:IHeatmapTrackSpec) => track.molecularProfileId)
-                                    .mapValues( (o:IHeatmapTrackSpec[]) => _(o).map((d)=>d.key).map((key:string) => trackSpecKeyToTrackId[key]).max() )
+                                    .mapValues( (o:IHeatmapTrackSpec[]) => _.last(o) )
+                                    .mapValues( (o:IHeatmapTrackSpec) => trackSpecKeyToTrackId[_.get(o, 'key')] )
```
adamabeshouse: why _.get? just `o.key`

src/shared/components/oncoprint/DeltaUtils.ts
adamabeshouse: @pvannierop sorry for the confusion. this is what we talked about the other day and I agreed the way you're doing it makes sense

src/shared/components/oncoprint/DeltaUtils.ts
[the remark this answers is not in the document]
adamabeshouse: comments are good, I just meant remove the commented out code

src/shared/components/oncoprint/OncoprintUtils.spec.ts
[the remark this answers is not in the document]
adamabeshouse: update?

src/shared/components/oncoprint/DataUtils.spec.ts
```diff
@@ -1287,6 +1288,24 @@ describe("DataUtils", ()=>{
...
+            partialTrackDatum,
+            "treatment_id",
+            "TREATMENT_ID_1",
+            {patientId:"patient", studyId:"study"} as Sample,
+            data
+           )
+           assert.deepEqual(
+            partialTrackDatum,
+            {treatment_id: "TREATMENT_ID_1", study_id:"study", profile_data:-10, thresholdType: '>', category: '>8.00'}
+           );
```
adamabeshouse: can you more exhaustively test this, with different threshold types and edge cases eg value right on the threshold, different sort orders

src/shared/components/oncoprint/DeltaUtils.spec.ts
```diff
@@ -513,4 +517,172 @@ describe("Oncoprint DeltaUtils", ()=>{
             assert.equal(oncoprint.setSortConfig.callCount, 0);
         });
     });
+
+    describe('transitionHeatmapTrack() for molecular profile', () => {
```
adamabeshouse: I don't understand this message - "for molecular profile". can you explain more?

adamabeshouse: nevermind, I understand now.

src/shared/components/oncoprint/DeltaUtils.spec.ts
```diff
@@ -513,4 +517,172 @@ describe("Oncoprint DeltaUtils", ()=>{
...
+            const trackIdForRuleSetSharing = {treatment: { profile_1: 2 }};
+            const prevSpec = nextSpec;
+
+            transitionHeatmapTrack(nextSpec, prevSpec, trackspec2trackId, ()=>undefined, oncoprint, nextProps, prevProps, trackIdForRuleSetSharing);
+            
+            assert.isTrue((oncoprint.shareRuleSet as SinonStub).called);
+            assert.equal(trackIdForRuleSetSharing.treatment['profile_1'], 2);
+        });
+
+    });
```
adamabeshouse: this testing is great!! thank you for this thoroughness

src/shared/components/oncoprint/OncoprintUtils.spec.ts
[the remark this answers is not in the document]
adamabeshouse: update on this?

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -392,4 +395,80 @@ describe('OncoprintUtils', () => {
...
+        C`);
+        assert.equal(elements.length, 3);
+    });
+    it("Removes duplicate entries", () => {
+        const elements = splitHeatmapTextField("A B B");
+        assert.equal(elements.length, 2);
+    });
+});
+
+describe('Treatment selections extracted from text area', () => {
```
adamabeshouse: we do `describe("functionName")` rather than `describe("feature")` for unit tests
so `describe("extractTreatmentSelections")`

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -392,4 +395,80 @@ describe('OncoprintUtils', () => {
...
+                onRemove: () => {}
+            };
+
+            // const ruleSetParams = getTreatmentTrackRuleSetParams(treatmentTracSpec);
+
+        });
+    });
+});
+
+describe('Split heatmap text field', () => {
```
adamabeshouse: `describe("splitHeatmapTextField")`

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -392,4 +395,80 @@ describe('OncoprintUtils', () => {
...
+    it('Splits around spaces', () => {
```
adamabeshouse: for these tests, it's even better to do 
`assert.deepEqual(elements, ["A", "B", "C"])` etc rather than `assert.equal` on the length

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -392,4 +395,80 @@ describe('OncoprintUtils', () => {
...
+describe('Treatment selections extracted from text area', () => {
+
+    const selectedTreatments:ISelectOption[] = [];
+    const treatmentMap = {
+        'treatmentA': {value: "valueA", label: "labelA"}
+    };
+
+    it('Adds recognized treaments to selection', () => {
+        extractTreatmentSelections("treatmentA treatmentB", selectedTreatments, treatmentMap);
+        assert.equal(selectedTreatments.length, 1);
```
adamabeshouse: use `assert.deepEqual` for more specific test

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -395,17 +418,73 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
         }
     }
 
+    @autobind
+    private onChangeTreamentTextArea(text:string):string {
```
adamabeshouse: typo - missing `t` in `Treatment`

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -752,3 +949,37 @@ export function makeGenesetHeatmapTracksMobxPromise(
...
+
+    // remove valid treatment ids from the input string
+    if (detectedTreatments.length > 0) {
+        _.each(detectedTreatments, (d:string) => {
+            text = text.replace(d, "");
+        });
+    }
+
+    // return the input string
+    return text.trim().replace("\t+","\t").replace(" +"," ");
```
adamabeshouse: can you explain what the purpose is of these changes, and of returning this string?

src/shared/components/oncoprint/DeltaUtils.ts
pvannierop: Changed.

src/shared/components/oncoprint/DataUtils.spec.ts
pvannierop: Added tons of test testing all kinds of edge cases. Also updated DataUtils to meet test requirements ;).

src/shared/components/oncoprint/DeltaUtils.spec.ts
pvannierop: Thnx :)

src/shared/components/oncoprint/OncoprintUtils.spec.ts
pvannierop: Updated.

src/shared/components/oncoprint/OncoprintUtils.spec.ts
pvannierop: Updated.

src/shared/components/oncoprint/OncoprintUtils.spec.ts
pvannierop: Good point. Changed.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: Updated.

src/shared/components/oncoprint/OncoprintUtils.ts
pvannierop: Ok. The text area component has editable text in its text area element. This text is updated when a word identical to a treatment identifier is recognized and the word is removed from the text. Normally I would do this with a observable MobX field in OncoprintControls. However, I was unable to get this to work for reasons unknown to me. Instead the onChange handler function in TextIconArea component can handle a return value:

```
 onChangeTextArea?: (textAreaContents:string) => string;
```

The TextIconArea uses the string returned by:
```
 return text.trim().replace("\t+","\t").replace(" +"," ");
```

to update the text area.

src/shared/components/oncoprint/OncoprintUtils.spec.ts
adamabeshouse: this is important, I think

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: Ah okay thanks, I think thats a good solution

src/shared/components/oncoprint/DataUtils.spec.ts
adamabeshouse: great, thank you

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -166,6 +181,86 @@ export function getHeatmapTrackRuleSetParams(molecularAlterationType: string) {
     };
 }
 
+function getTreatmentTrackRuleSetParams(trackSpec: IHeatmapTrackSpec):RuleSetParams {
```
adamabeshouse: could you just add a test for this?

pvannierop: Ok, added tons of unit tests that test all edge cases. Again, updated the function under test to make all tests work.

src/shared/components/oncoprint/OncoprintUtils.spec.ts
pvannierop: Good point. All tests updated.

src/shared/components/oncoprint/OncoprintUtils.spec.ts
```diff
@@ -396,6 +400,29 @@ describe('OncoprintUtils', () => {
         })
     });
 
+    describe('Treatment ruleset params', () => {
+
+        it('Is created from Track Spec param', () => {
```
adamabeshouse: this whole section seems should be deleted right?

pvannierop:
@adamabeshouse 
I have squashed all commits. Thnx for the patience. I learned a lot.

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -60,6 +60,11 @@ const protRenderPriority = {
     'down': 0
 };
 
+type InternalCaseDatum = {
```
adamabeshouse: vague name

pvannierop:
I have added the DO NOT MERGE label. This PR depends on a backend PR that should be merged before this PR. After merge of backend changes the API code can be re-generated so that the API test pass.

PR #2055, opened 2019-02-20
Integration of treatment response data in PlotsTab (incl. new Waterfall plot)

pvannierop:
# What? Why?
This PR will implement handling of the new `treatment response` (genetic entity) data type by the PlotsTab component. A new plot type _Watefall plot_ is added.

## Changes
- `Treatment` is shown in data type select boxes of the horz. and vert. axis menu's (when available in cBioPortal db). When selected treatment response genetic profiles are shown in the data source select menu's (see Fig.1). Only treatments selected in the oncoprint heatmap menu are available for selection. 
- Treatment data points can be _limited_ (real value lies beyond a certain value) which can be indicated by `>` or `<` prefixes to values (for example _>8.00_, see [this backend PR](https://github.com/cBioPortal/cbioportal/pull/5460)). Bar, scatter and waterfall (see below) plot types contain a legend option `Limit Value` when one or more data points in the plot are limited. Enabling this legend option represents limited data points as diamonds (see Fig.1). This option is enabled by default and is re-enabled after each axis menu change. When __enabled__ limited data values are __excluded__ from calculations of correlation coefficients in the scatter plot legend. 
- When `Treatment` data type is selected a `--`option is shown in the other data type menu that represents no data type selected for the corresponding axis (see Fig.1). For treatment data the `Apply log scale` option is displayed in the axis menu's. When the `Treatment` and `--` combination is selected by the user, a new plot type `Waterfall plot` is displayed.
- A new plot type `Waterfall plot` is implemented (see Fig.2). This plot type represents a sorted (asc or desc) bar graph of response values expressed relative to a given threshold value a.k.a. `pivot_threshold `. The `pivot_threshold` is set on a profile level scope and is passed from the cbioportal backend. When no `pivot_threshold` is passed unmodified values are represented. When a waterfall plot is shown `Log-scale` and `Sort Order` controls are shown in the corresponding axis menu that control log10-transformation and asc-desc sorting, respectively. The legend options for this plot type include a new gene selection box that controls for which gene the mutation or CNA information is represented in the plot. For the waterfall plot the mutation and CNA styling options are mutually exclusive. The selected gene is represented in the title of the plot. When legend option `Limit Value` is enabled limited data points are marked by a black diamond above respective bars. Sample and mutation queries are represented by red plus signs below respective bars. The width of the waterfall plot is dynamic with a max-width of 1000px. Download of waterfall plot data includes raw and `pivot_threshold`-adjusted values and the sort order.

### Notes
- Because PlotsTab only shows treatments selected in Oncoprint functionality in this PR for treatment response data is dependent on [this PR](https://github.com/cBioPortal/cbioportal-frontend/pull/2054) that implements functionality for treatment response data in the OncoprintTab. For review of the Plotstab functionality a treatment list must be added to the URL (f.i. http://localhost:8081/cbioportal/results/plots? ... &treatment_list=**17-AAG**%3B**AEW541**).

### Figure 1
1. Treatment data type
2. `--` data type option in other menu
3. `Limit Value` legend option
4. Limited data points styled as diamonds
5. `Log scale` control for treatment profile values

Screenshot from 2019-04-04 09-32-03

### Figure 2
1. Waterfall plot (at max width)
2. `Limit Value` legend option
3. Limited data points styled as diamonds above bars
4. Gene selection box for mutation or CNA styling
5. Selected gene represented in title
6. `Log Scale` and `Sort Order` controls for treatment profile values
7. Sample search as red plus sign

Screenshot from 2019-04-04 09-32-51


adamabeshouse:
a lot of files in this PR that should not be included - build files, and that yarn-error.log

adamabeshouse:
I don't think waterfall plot is a good name, it doesn't match the use of waterfall plot that I find by googling?

This symbol isn't in the legend 
image

adamabeshouse:
have you thought about renaming "--" to be more explicit, like "None (List of samples)" or "Samples" or something? I find "--" confusing

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2189,6 +2203,27 @@ export class ResultsViewPageStore {
                     geneticEntityId: geneset.genesetId, cytoband: "-", geneticEntityData: geneset});
             }
             return Promise.resolve(res);
+
+        }
+    });
+
+    readonly treatmentsInStudies = remoteData<Treatment[]>({
```
adamabeshouse: werent these changes in your previous PR?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1,49 +1,42 @@
...
+import { stringListToIndexSet } from "../../../shared/lib/StringUtils";
+import { AlterationTypeConstants, AnnotatedMutation, AnnotatedNumericGeneMolecularData } from "../ResultsViewPageStore";
+import { CoverageInformation } from "../ResultsViewPageStoreUtils";
+import { AxisMenuSelection, MutationCountBy, NONE_SELECTED_OPTION_NUMERICAL_VALUE, NONE_SELECTED_OPTION_STRING_VALUE, ViewType, PlotType, UtilitiesMenuSelection } from "./PlotsTab";
+import { IBaseWaterfallPlotData } from "shared/components/plots/WaterfallPlot";
 
 export const CLIN_ATTR_DATA_TYPE = "clinical_attribute";
+export const BOUNDARY_VALUE_NAME = "Limit"; // "... value": name to indicate values that are > or < than a certain value (e.g. "">8.00")
 export const GENESET_DATA_TYPE = "GENESET_SCORE";
+export const TREATMENT_DATA_TYPE = "TREATMENT_RESPONSE";
```
adamabeshouse: AlterationTypeConstants

adamabeshouse: oh looks like you used it below, so just delete this

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -79,7 +73,8 @@ export function sortMolecularProfilesForDisplay(profiles:MolecularProfile[]) {
 
 export const CNA_STROKE_WIDTH = 1.8;
 export const PLOT_SIDELENGTH = 650;
-
+export const WATERFALLPLOT_BASE_SIDELENGTH = 500;
+export const WATERFALLPLOT_SIDELENGTH_SAMPLE_MULTIPLATION_FACTOR = 1.6;
```
adamabeshouse: like I said before, waterfall plot doesn't seem the right name

adamabeshouse: also multiplication is misspelled

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -130,24 +129,42 @@ export interface IScatterPlotSampleData {
     profiledMutations?:boolean;
     mutations: AnnotatedMutation[];
     copyNumberAlterations: AnnotatedNumericGeneMolecularData[];
+    xThresholdType?: '>'|'<'|undefined; // threshold symbol for x coordinate in scatter plot
```
adamabeshouse: please remove xT and yT, move them into a new interface IScatterPlotSampleData that extends IPlotSampleData

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -130,24 +129,42 @@ export interface IScatterPlotSampleData {
...
+    yThresholdType?: '>'|'<'|undefined; // threshold symbol for y coordinate in scatter plot
+    thresholdType?: '>'|'<'|undefined; // threshold symbol for continuous axis in box-scatter and waterfall plot
```
adamabeshouse: please put this separately into IBoxScatterPlotPoint and IWaterfallPlotSampleData

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -130,24 +129,42 @@ export interface IScatterPlotSampleData {
...
-export interface IScatterPlotData extends IScatterPlotSampleData, IBaseScatterPlotData {};
-
-export interface IBoxScatterPlotPoint extends IScatterPlotSampleData {
+export interface IBoxScatterPlotPoint extends IPlotSampleData {
     category:string;
     value:number;
     jitter:number;
 }
 
+export interface IWaterfallPlotSampleData extends IPlotSampleData {
```
adamabeshouse: I don't understand what this interface is. the `SampleData` indicates to me you're going to have one of these per sample, as with the `SampleData` interfaces above, but this seems to hold plot-level attributes. Do you need to store sortOrder and pivotThreshold on a per-sample basis?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1297,6 +1543,132 @@ export function makeScatterPlotData(
     return expandedData;
 }
 
+export function makeWaterfallPlotData(
+    axisData: INumberAxisData,
```
adamabeshouse: please add unit tests for this method

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -117,14 +137,22 @@ export enum MutationCountBy {
...
     selectedGenesetOption?:{value:string, label:string};
+    selectedTreatmentOption?:{value:string, label:string};
     dataType?:string;
     dataSourceId?:string;
     mutationCountBy:MutationCountBy;
     logScale: boolean;
 };
 
+export type UtilitiesMenuSelection = {
+    entrezGeneId?:number;
```
adamabeshouse: Can you adjust the name to describe what the purpose is of selecting a gene in the utilities menu? e.g. highlightedEntrezGeneId or something like that?

I don't also see the purpose of storing both selectedGeneOption and entrezGeneId, seems like you could compute entrezGeneId from the selectedGeneOption, and indeed you could do that (see how getters are handled in AxisMenuSelection)

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1387,6 +1771,45 @@ export function getScatterPlotDownloadData(
...
 
+export function getWaterfallPlotDownloadData(
+    data:IWaterfallPlotData[],
+    sortOrder:string,
+    pivotThreshold:number,
+    axisLabel:string,
+    entrezGeneIdToGene:{[enstrezGeneId:number]:Gene}
+) {
+
+    let dataPoints = _.cloneDeep(data);
```
adamabeshouse: please add test for this method

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -244,8 +247,8 @@ export default class BoxScatterPlot<D extends IBaseBoxScatterPlotPoint> extends
             }
         }
         if (this.props.logScale) {
-            min = this.logScale(min);
-            max = this.logScale(max);
+            min = this.props.logScale.fLogScale(min, 0);
```
adamabeshouse: props.logScale might be undefined, typescript should catch that

src/shared/components/plots/PlotUtils.ts
```diff
@@ -247,4 +253,10 @@ export function computeCorrelationPValue(correlation:number, numSamples:number)
     } else {
         return null;
     }
-}
\ No newline at end of file
+}
+
+export function dataPointIsLimited(point:IPlotSampleData) {
+    const o =  (point.xThresholdType !== undefined || point.yThresholdType !== undefined || point.thresholdType !== undefined);
```
adamabeshouse: no biggie but could just return without saving `o`

src/shared/components/plots/ScatterPlot.tsx
```diff
@@ -41,8 +45,9 @@ export interface IScatterPlotProps<D extends IBaseScatterPlotData> {
         spearman: number;
     };
     showRegressionLine?:boolean;
-    logX?:boolean;
-    logY?:boolean;
+    logX?:IAxisLogScaleParams|undefined;
+    logY?:IAxisLogScaleParams|undefined;
+    excludeLimitValuesFromCalculation?:boolean;
```
adamabeshouse: could you add a comment for this?

src/shared/components/plots/ScatterPlot.tsx
```diff
@@ -200,9 +204,17 @@ export default class ScatterPlot<D extends IBaseScatterPlotData> extends React.C
     }
 
     @computed get splitData() {
+
+        // when limit values are shown in the legend, exclude
+        // these points from calculations of correlation coefficients
+        let data = this.props.excludeLimitValuesFromCalculation?
+            _.filter(this.props.data, dataPointIsLimited)
+            :
```
adamabeshouse: could you move the `:` onto line 211

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -130,24 +129,42 @@ export interface IScatterPlotSampleData {
...
+    sortOrder:string;
+    pivotThreshold:number;
+}
+
+export interface IScatterPlotData extends IPlotSampleData, IBaseScatterPlotData {};
+export interface IWaterfallPlotData extends IWaterfallPlotSampleData, IBaseWaterfallPlotData {};
+
+export interface IAxisLogScaleParams {
+    label:string;
+    fLogScale:((x:number, offset:number)=>number);
```
adamabeshouse: what does `offset` mean? can you add a comment

src/shared/components/plots/ScatterPlot.tsx
```diff
@@ -311,34 +315,36 @@ export default class ScatterPlot<D extends IBaseScatterPlotData> extends React.C
...
 
     @bind
     private tickFormatY(t:number, i:number, ticks:number[]) {
-        return this.tickFormat(t, ticks, !!this.props.logY);
+        return this.tickFormat(t, ticks, this.props.logY);
     }
 
     @computed get data() {
-        return separateScatterDataByAppearance(
+        let o = separateScatterDataByAppearance(
```
adamabeshouse: no need to save `o`

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,182 @@
+import React from 'react';
+import { assert } from 'chai';
+import { IBaseWaterfallPlotData, IWaterfallPlotProps } from './WaterfallPlot';
+import WaterfallPlot from 'shared/components/plots/WaterfallPlot';
+import { shallow } from 'enzyme';
+import labelMobxPromises from 'mobxpromise';
+
+describe('', () => {
```
adamabeshouse: `describe('WaterfallPlot', ()=>{`

src/shared/components/plots/WaterfallPlot.tsx
```diff
@@ -0,0 +1,545 @@
...
+
+    @bind
+    private tickFormatY(t:number, i:number, ticks:number[]) {
+        if (this.props.horizontal) {
+            return undefined;
+        }
+        return this.tickFormat(t, ticks, this.props.log);
+    }
+
+    @computed get internalData() {
```
adamabeshouse: more descriptive name please

src/shared/components/plots/WaterfallPlot.tsx
```diff
@@ -0,0 +1,545 @@
...
+const DEFAULT_FONT_FAMILY = "Verdana,Arial,sans-serif";
+export const LEGEND_Y = 30
+const RIGHT_PADDING = 120; // room for correlation info and legend
+const NUM_AXIS_TICKS = 8;
+const PLOT_DATA_PADDING_PIXELS = 50;
+const MIN_LOG_ARGUMENT = 0.01;
+const LEFT_PADDING = 25;
+const LABEL_OFFSET_FRACTION = .02;
+const LABEL_SIZE_MULTIPLIER = 1.5;
+const labelStyle = {
```
adamabeshouse: why is this called labelStyle it doesn't seem to describe labels

src/shared/components/plots/WaterfallPlot.tsx
```diff
@@ -0,0 +1,545 @@
...
+                                tickFormat={this.tickFormatY}
+                                dependentAxis={true}
+                                axisLabelComponent={<VictoryLabel dy={-35}/>}
+                                label={this.props.axisLabel}
+                            />}
+                            <VictoryBar
+                                // barRatio={1} // removes spaces between bars
+                                style={{
+                                    data: {
+                                        fill: (d:D) => d.fill,
```
adamabeshouse: it's significantly more efficient with Victory to make a separate VictoryBar element for each style settings, rather than using functions. You can see an example of how I do this in `ScatterPlot.tsx` in the getChart() method with VictoryScatter elements

src/shared/components/plots/WaterfallPlot.tsx
```diff
@@ -0,0 +1,545 @@
...
+                                data={this.limitValueLabels}
+                                x={this.datumAccessorLabelX}
+                                y={this.datumAccessorLabelY}
+                                />
+                            <VictoryScatter
+                                style={{
+                                    data: {
+                                        fill: "white",
+                                        stroke: "red",
+                                        strokeWidth: 1*LABEL_SIZE_MULTIPLIER,
```
adamabeshouse: 1* ?

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+            datum2 = plot.internalData[1];
+            assert.equal(datum1.order, 1);
+            assert.equal(datum1.value, 4);
+            assert.equal(datum2.order, 2);
+            assert.equal(datum2.value, 2);
+        });
+        
+    });
+    
+    describe('Plot domain', () => {
```
adamabeshouse: `describe('#plotDomain', ()=>{`

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+        testData = [
+            {value: 2},
+            {value: 4}
+        ];
+        
+        plot = new WaterfallPlot<IBaseWaterfallPlotData>(testProps);
+
+    });
+
+    describe('Data points', () => {
```
adamabeshouse: `describe('#internalData', ()=>{`

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+            assert.isDefined(datum.fill);
+            assert.isDefined(datum.stroke);
+            assert.isDefined(datum.strokeOpacity);
+            assert.isDefined(datum.strokeWidth);
+            assert.isDefined(datum.symbol);
+            assert.isDefined(datum.labelVisibility);
+            
+        });
+
+        // it('Negative values are be log-transformed', () => {
```
adamabeshouse: delete commented code

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,182 @@
...
+            assert.equal(plot.limitValueLabels.length, 2);
+        });
+
+        it('Limit vlaue label data points have x/y coordinates', () => {
+            testProps.labelVisibility = () => true;
+            plot = new WaterfallPlot<IBaseWaterfallPlotData>(testProps);
+            assert.isDefined(plot.limitValueLabels[0].labelx);
+            assert.isDefined(plot.limitValueLabels[0].labely);
+        });
+
```
adamabeshouse: overall these are great tests

adamabeshouse: @alisman what do you think of this approach to testing (checking public properties on the instantiated component)

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1415,6 +2045,54 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
+
+        // log-transformation parameters for non-treatment reponse 
+        // profile data Note: log2-transformation is used by default
+        if(axisSelection.dataType !== AlterationTypeConstants.TREATMENT_RESPONSE) {
+            
+            label = "log2";
+            fLogScale = (x:number, offset:number) => Math.log2(Math.max(x, MIN_LOG_ARGUMENT));
+            fInvLogScale = (x:number) => Math.pow(2, x);
+            
+        // log-transformation parameters for treatment reponse
```
adamabeshouse: I think it's clearer to put these comments inside the branches, this one in particular I find really confusing

pvannierop:
> I don't think waterfall plot is a good name, it doesn't match the use of waterfall plot that I find by googling? It's just a bar chart, isn't it?

I disagree (see f.i. [this ncbi page](https://www.ncbi.nlm.nih.gov/pmc/articles/PMC4093310/))

pvannierop:
> have you thought about renaming "--" to be more explicit, like "None (List of samples)" or "Samples" or something? I find "--" confusing

`--` means as a much as `-- no genetic profile selected for this axis --`. I do not see how  "None (List of samples)" or "Samples" would describe this better. I also wanted to make clear in one glance that this entry does not represent a profile, hence the `--` to make it stand out. How about `-- none --`?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Removed.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Spelling error corrected. Awaiting outcome of waterfall plot name.

adamabeshouse:
> I disagree (see f.i. this ncbi page)

Ah, okay, thanks.

> --

It's just that I think the value of Data Type for each axis should represent what the data is along that axis. So the data along the axis is a list of samples, it's not nothing. That's why I find `--` and `none` and `no profile` confusing.

@schultzn @jjgao can you weigh in?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Ok, I have made two attempts in this direction and both were unsuccessful. The reason for this is that I cannot reconcile 1) specific data types accepted by the different plots and 2) the type requirements of many functions in PlotsTabUtils.

I think the problem lies in the fact that in the current state `sample data` is the only property on which data points are styled in plots. As such all styling related functions in PlotsTabUtils are typed for the `IPlotSampleData` interface. With the introduction on a value-dependent qualifier such as the thresholdType symbol, the description of `IPlotSampleData` as a stand-in for what in effect is a `IPlotStylingData` interface is not longer sufficient.

I am afraid that splitting up of `IPlotSampleData` into different types for scatterplot (two-dimensional) and boxscatter/water (one-dimensional) as per your suggestion is something I tried two times, but failed. As stated in the paragraph above the reason for this is that `IPlotSampleData` is treated as a `IPlotStylingData` interface. The bad naming of this interface suggests that sample attributes are incorrectly mixed with value attributes in my PR. These attributes seem apples and pears, but are similar in the fact that these are attributes used to provide styling to datapoints. For this reason I added this comment:
```
// this interface contains all attributes needed to provide correct styling of 
// graph elements (points, bars, ...)
```

I my assessment correct so-far?

What do you think of this proposition:
```
export interface IPlotSampleData {}

export interface ITwoDimentsionPlotStyling {
    xThresholdType?: '>'|'<'|undefined; // threshold symbol for x coordinate in scatter plot
    yThresholdType?: '>'|'<'|undefined; // threshold symbol for y coordinate in scatter plot
}

export interface IOneDimensionPlotStyling {
    thresholdType?: '>'|'<'|undefined; // threshold symbol for x coordinate in scatter plot
}

export interface IPlotStylingData extends IPlotSampleData, IOneDimensionPlotStyling, ITwoDimentsionPlotStyling {}
```
After this rework, we can remap all styling related function to the new `IPlotStylingData` interface. Does this rework make sense?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: See discussion in the comment above.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: I have removed this interface.

src/shared/components/plots/PlotUtils.ts
pvannierop: Updated.

src/shared/components/plots/ScatterPlot.tsx
pvannierop: Added.

src/shared/components/plots/ScatterPlot.tsx
pvannierop: Updated.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Comment added.

src/shared/components/plots/ScatterPlot.tsx
pvannierop: Updated.

src/shared/components/plots/WaterfallPlot.spec.tsx
pvannierop: Corrected.

src/shared/components/plots/WaterfallPlot.tsx
pvannierop: What do you think of the name `waterfallPlotdata`?

src/shared/components/plots/WaterfallPlot.tsx
pvannierop: I agree. In the current state `label` means `searchIndicator`. Do you agree with renaming all occurences of `label` into `searchIndicator`?

src/shared/components/plots/WaterfallPlot.tsx
pvannierop: In the case of scatterplot the approach is to partition the datapoints into groups of identical styling and then consecutively plot different groups of points with hard coded styling information. For bar plots such as the waterfallplot this is not possible because the bar widths and inter-bar widths should be evaluated with all bars being present. When taking the scatterplot approach leads to malformed graphs. This was in fact my first implementation that I had to change to make it work.

src/shared/components/plots/WaterfallPlot.tsx
pvannierop: :) removed .... Math is hard.

src/shared/components/plots/WaterfallPlot.spec.tsx
pvannierop: Updated.

src/shared/components/plots/WaterfallPlot.spec.tsx
pvannierop: Updated.

src/shared/components/plots/WaterfallPlot.spec.tsx
pvannierop: No, it should be removed here since it is a left over. I have moved the formulation of log-scaling functions to the PlotsTab component (`Plotst=Tab` line 2055). Reason is that we required data type-dependent rather than plot type dependent application of data transformation types (treatment response data is log10 transformed, molecular data log2-transformed). I will move and implement the test to PlotsTab.spec.ts. Agree?

src/shared/components/plots/WaterfallPlot.spec.tsx
pvannierop: I used this approach since this component is just a layer between Plotstab and Victory chart. I just test what data this component passes onto Victory. I realize this is very implementation dependent. But I saw no other way to test the functionality of WaterfallPlot.tsx. Feedback would be much appreciated.

src/pages/resultsView/plots/PlotsTab.tsx
pvannierop: Comments moved. Do you also object to the content of the comment in the sense that it is unclear?

schultzn:
I agree with Adam. I think the proper date type would be “Sample IDs” or “Samples”. 

And on the question of whether this is a waterfall plot or a bar chart: A proper waterfall plot will show positive (tumor growth) and negative (shrinking) values. Does this implementation do that?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Tests were added.

src/pages/resultsView/plots/PlotsTab.tsx
pvannierop: How about `geneForMutationAndCnaStyling`?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Tests were added.

pvannierop:
@schultzn @adamabeshouse 
How about `Ordered samples` or `Sample order` to better reflect what is on the axis?

pvannierop:
>This symbol isn't in the legend

You are correct. This symbol appears when the user entered a search term in the sample or mutation search boxes. The red cross is equivalent to the enlarged data points that are also not reflected in the legend. I agree that a new character appearing in the plot may require a different approach to making a plot element larger in terms of readability. However, the user story is that the user enters a search term and that the symbol appears to indicate the search result. The user is conscious of this symbol. When using the figure for publication the red cross requires explanation. But again, the meaning of a larger data point is not explained by the legend either. So, I'd rather come up with a general solution to this problem and a solid plan (what info should the legend show, should the legend show the info on the samples that match, how should the legend respond to sample searches with more than one sample, ...).

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: thats great :+1:

src/shared/components/plots/ScatterPlot.tsx
adamabeshouse: Thanks. now that I understand what it is, I think it would be better to change 'Calculation' to 'Correlation' in the name

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: perfect, thanks

src/shared/components/plots/WaterfallPlot.spec.tsx
adamabeshouse: the other option would be to move the logic to a helper function. In a way I prefer that since it makes the test not dependent on more arbitrary types of implementation details. But in this case it doesn't bother me and I see the benefits of doing it like this because it's almost like an end to end test. I'm cool with it.

src/shared/components/plots/WaterfallPlot.spec.tsx
```diff
@@ -0,0 +1,167 @@
...
+        testData = [
+            {value: 2},
+            {value: 4}
+        ];
+        
+        plot = new WaterfallPlot<IBaseWaterfallPlotData>(testProps);
+
+    });
+
+    describe('#internalData', () => {
```
adamabeshouse: `#waterfallPlotData`

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -130,24 +128,55 @@ export interface IScatterPlotSampleData {
     profiledMutations?:boolean;
     mutations: AnnotatedMutation[];
     copyNumberAlterations: AnnotatedNumericGeneMolecularData[];
+    thresholdType?: '>'|'<'|undefined; // threshold symbol for continuous axis in box-scatter and waterfall plot
+    xThresholdType?: '>'|'<'|undefined; // threshold symbol for x coordinate in scatter plot
+    yThresholdType?: '>'|'<'|undefined; // threshold symbol for y coordinate in scatter plot
+}
+
+export interface ITwoDimentsionPlotStyling extends IPlotSampleData {
```
adamabeshouse: typo/grammar, should be `Dimensional`

adamabeshouse:
@pvannierop I like "Ordered samples" and let's put in the tooltip of the info icon "Sample order determined by value of data on ${other} axis". 

What happens if you choose "Ordered samples" for both axes? I think it should forbid that from happening - remove that option from the list of one axis if it's chosen in the other axis.

adamabeshouse:
@pvannierop I think that the values don't need to be explained by the legend, that's self-explanatory by the axis. But the symbol is unusual and I think it would definitely be good to put in the legend, along with all the color explanations.

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: No, I think the content of the comment is good. Thanks!

src/shared/components/plots/WaterfallPlot.spec.tsx
adamabeshouse: that sounds good :+1:

src/shared/components/plots/WaterfallPlot.tsx
adamabeshouse: :) indeed

src/shared/components/plots/WaterfallPlot.tsx
adamabeshouse: Can you calculate the bar widths using all the data together, then separate them out for input to victory?

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: Sorry but I'm confused. Can we talk on the phone about this? Can you speak for a half hour on Monday at 10 AM my time?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -132,22 +130,47 @@ export interface IScatterPlotSampleData {
...
+    label:string;
+
+    // Function for transforming `number` into its log-scale representation
+    // `offset` is a constant added to the value before transformation (used
+    // for transformation of data sets with negative numbers)
+    fLogScale:((x:number, offset:number)=>number);
+
+    // Function for back-transforming `number` transformed with fLogScale
+    // function into its linear-scale representation
+    fInvLogScale:((x:number)=>number);
```
adamabeshouse: just occurred to me, dont you need offset to be a part of this too

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -132,22 +130,47 @@ export interface IScatterPlotSampleData {
     copyNumberAlterations: AnnotatedNumericGeneMolecularData[];
 }
 
-export interface IScatterPlotData extends IScatterPlotSampleData, IBaseScatterPlotData {};
+export interface IThresholdDataPoint {
+    thresholdType?:'>'|'<'|undefined;
+}
+export interface IThresholdsDataPoint {
```
adamabeshouse: good but threshold and thresholds are too similar to the eye. better 1DThreshold and 2DThreshold

pvannierop:
>@pvannierop I like "Ordered samples" and let's put in the tooltip of the info icon "Sample order determined by value of data on ${other} axis".

>What happens if you choose "Ordered samples" for both axes? I think it should forbid that from happening - remove that option from the list of one axis if it's chosen in the other axis.

Implemented as per your suggestion. Selection of `Ordered samples` in the dataType menu shows a info icon with the message. Note that this is the first dataType selection to provide information this way.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Note: after discussion with @adamabeshouse we decided to refactor the interface and add  'thresholdType` type checking in the function in PlotsTabUtils that accept data points that conform to the less specific IPlotsTabData interface.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Resoved in rework of interfaces.

src/shared/components/plots/ScatterPlot.tsx
pvannierop: Name was updated to `excludeLimitValuesFromCalculation`.

src/shared/components/plots/WaterfallPlot.tsx
pvannierop: Maybe, I don't know. But how much of a problem is the current implementation? Will this performance difference be noticeable by the user? To be honest we do not have budget left to explore alternative ways that may or may not work, and I suspect that your proposed change is not guaranteed to be successful . Can't we not postpone this mod to a later time point?

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Added offset parameters as optional parameters to the interfaces. For the log10-backtransformations this parameter is now also used. Unit tests were added.

src/pages/resultsView/plots/PlotsTabUtils.tsx
pvannierop: Interfaces were renamed to `IDataPoint1D` and `IDataPoint2D`.

pvannierop:
>@pvannierop I think that the values don't need to be explained by the legend, that's self-explanatory by the axis. But the symbol is unusual and I think it would definitely be good to put in the legend, along with all the color explanations.

Search results symbols are added to the legend with label `Sample match`.
Screenshot from 2019-05-06 18-24-36

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -132,22 +130,47 @@ export interface IScatterPlotSampleData {
...
 
-export interface IBoxScatterPlotPoint extends IScatterPlotSampleData {
+export interface IBoxScatterPlotPoint extends IPlotSampleData, IThreshold1D {
     category:string;
     value:number;
     jitter:number;
 }
 
+export interface IScatterPlotData extends IPlotSampleData, IBaseScatterPlotData, IThreshold2D {};
+export interface IWaterfallPlotData extends IPlotSampleData, IBaseWaterfallPlotData, IThreshold1D {};
```
adamabeshouse: all the interfaces and inheritance look very clear to me now. thank you!

src/shared/components/plots/WaterfallPlot.spec.tsx
adamabeshouse: lets do this otherwise it is confusing

pvannierop:
I have added the `DO NOT MERGE` label. This PR depends on a backend PR that should be merged before this PR. After merge of backend changes the API code can be re-generated so that the API test pass.

PR #2056, opened 2019-02-20
Add group and percentage stacked bar chart in Plots Tab

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5755

screen shot 2019-03-05 at 11 31 38 pm
screen shot 2019-03-05 at 11 32 11 pm

screen shot 2019-02-22 at 11 56 51 am
screen shot 2019-02-22 at 11 57 00 am


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1152,6 +1157,18 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                             /> Horizontal Bars
                         </label></div>
                     )}
+                    {showStackedBarHorizontalOption && (
+                        <div className="checkbox"><label>
```
adamabeshouse: want to move this up into line 1149, since its the same condition?

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -215,6 +216,9 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
     }
 
     @computed get maxMajorCount() {
+        if(!!this.props.percentage){
```
adamabeshouse: this is a nitpick but since percentage is boolean, !!percentage will be true iff percentage is true. so you can just check `if(this.props.percentage)`

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -336,7 +340,7 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
         //  this axis is for numbers, not categories
         const label = [this.props.axisLabelX];
         if (this.props.horizontalBars) {
-            label.unshift(COUNT_AXIS_LABEL);
+            label.unshift(`${COUNT_AXIS_LABEL}${!!this.props.percentage ? " (%)": ""}`);
```
adamabeshouse: same as above, just check for this.props.percentage

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -360,7 +364,7 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
     @computed get vertAxis() {
         const label = [this.props.axisLabelY];
         if (!this.props.horizontalBars) {
-            label.push(COUNT_AXIS_LABEL);
+            label.push(`${COUNT_AXIS_LABEL}${!!this.props.percentage ? " (%)": ""}`);
```
adamabeshouse: remove `!!`

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -483,7 +487,8 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
             this.majorCategoryOrder,
             this.getColor,
             this.categoryCoord,
-            !!this.props.horizontalBars
+            !!this.props.horizontalBars,
+            !!this.props.percentage
```
adamabeshouse: I guess here we need it for typescript

src/shared/components/plots/StackedBarPlotUtils.spec.ts
```diff
@@ -629,7 +629,7 @@ describe("StackedBarPlotUtils", ()=>{
             }], "given major order (alphabetical minor)");
 
             assert.deepEqual(makeBarSpecs(
-                multipleValueHorzBarsData, categoryOrder, categoryOrder, getColor, categoryCoord, true
+                multipleValueHorzBarsData, categoryOrder, categoryOrder, getColor, categoryCoord, true, false
```
adamabeshouse: it would be good to add tests of data created with percentage calculation as well - ensure its values for `y` and `percentage` are correct

Im surprised these unit tests would even be passing since it looks like you're adding the `percentage` member to the data

kalletlak: will add test once we are ok with styles and everything

jjgao:
Should it be a dropdown menu: Bar chart, Stacked bar chart, 100% stacked bar chart?

image

kalletlak:
@jjgao do we want `Horizontal Bars` checkbox for all the bar chart types?

jjgao:
@kalletlak I think that'd good if possible.

jjgao:
I guess the first one (grouped bar chart) is usually vertical. But for consistency, I think it's ok to always have the "horizontal bar" option.

@adamabeshouse @inodb : thoughts?

adamabeshouse:
yeah I dont see why not

kalletlak:
> Should it be a dropdown menu: Bar chart, Stacked bar chart, 100% stacked bar chart?
> 
> image

@jjgao should we also move `Show Table Plot` to dropdown

screen shot 2019-02-22 at 11 30 16 am

kalletlak:
@jjgao @adamabeshouse updated screenshots in description

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -78,8 +78,10 @@ enum EventKey {
     utilities_viewMutationType,
     utilities_viewCopyNumber,
     utilities_discreteVsDiscreteTable,
-    utilities_stackedBarHorizontalBars,
-    utilities_showRegressionLine
+    utilities_horizontalBars,
+    utilities_showRegressionLine,
+    utilities_percentageBars,
+    utilities_stackedBars
```
kalletlak: Code is yet to be cleaned up and tests needs to be updated/added. I'll work on it once we are good with layout.

jjgao:
> @jjgao should we also move `Show Table Plot` to dropdown

good idea 👍

jjgao:
Not sure if I like the gaps within a group... 

Any solutions: maybe a thin line at the bottom for count of 0?

screen shot 2019-02-22 at 11 56 20 am

adamabeshouse:
dropdown is a great idea but 100% stacked bar chart is sort of weird wording.. percentage stacked bar chart would be better IMO

jjgao:
@adamabeshouse 100% stacked bar chart was copied from google sheets. I think it's a common name: https://www.google.com/search?q=100%25+stacked+bar+chart

kalletlak:
@jjgao @adamabeshouse how about this?
screen shot 2019-02-22 at 1 31 56 pm

adamabeshouse:
@jjgao ah okay, I accept it :D 

@kalletlak i think it looks good. however I think the small line doesn't look like zero. Maybe to compensate, you can bump up the 0 axis tick by that same amount? Im not sure how easy that would be in victory...

jjgao:
> @kalletlak i think it looks good. however I think the small line doesn't look like zero.

I feel the same actually. Maybe we can center the rectangle around 0 and make it thinner if necessary?

And why x-axis is dashed line?

adamabeshouse:
@jjgao the x-axis is dashed because of victory placing a grid line on 0. theres probably a way to get rid of that.

kalletlak:
@jjgao @adamabeshouse what about this. reduced the size of the bar with 0 count.
screen shot 2019-03-02 at 11 04 50 am

kalletlak:
@adamabeshouse could you review this pr again, added grouped bar chart and screenshots from previous review.

Sjoerd-van-Hagen:
Sorry for entering this discussion this late. I was not aware of this work.

I would prefer we change the way the plot is set up because what this does is basically coloring by clinical attribute, and grouping by another. We already have a way to color based on mutations and CNV, we should use the same mechanism for coloring by clinical variable, especially because this could also allow you to color by clinical variable in other plots types. Now it is counterintuitieve because you ask the user to select data for an axis but he will get a coloring.

For the waterfall plot, we basically let the user select a variable on one axis, and not having anything on the other axis. The plot then switches to waterfall where you have one variable and the data is sorted across the other axis. We use the standard coloring functionality for the coloring. It would be nice if your plot would do the same.

The choice between horizontal and vertical can be removed because it is implemented by swap axis.

The only thing that you would need on top of our solution is a menu to aggregate on counts, stacked counts or percentage counts.

Does that make sense?

jjgao:
@Sjoerd-van-Hagen thanks for the comments. The stacked bar chart is already in production. This pull request is to add the grouped bar chart and 100% stacked bar chart. 

This one is also is a little different from waterfall plot. Currently we only use Mutations and CNA to color samples (i.e. waterfall plot and scatter plot are sample level plots). In this bar chart, we are plot summary data, the clinical data is used as both y-axis (or x-axis) and as coloring groups, so I don't think it's counterintuitive (Google sheets or MS excel use it in the same way).

I think we should merge this one and follow up with the idea you mentioned (using clinical data for coloring). We hope that a GSoC student could pick it up: https://github.com/cBioPortal/GSoC/issues/63.

About horizontal vs vertical, I think it doesn't hurt to give the user the option. Some data look better with horizontal bar charts and some with vertical.

Sjoerd-van-Hagen:
@jjgao I never tried the stacked bar chart and agree that this is good to be merged. I was just thinking of ways to make it easier and in line with the waterfall plot.

end-to-end-tests/specs/screenshot.spec.js
```diff
@@ -565,56 +565,95 @@ describe("plots tab screenshot tests", function() {
...
     });
     it("plots tab mutations wildtype mode vs clinical table plot", function() {
-        browser.click('input[data-test="StackedBarHorizontalBars"]');
-        browser.click('input[data-test="DiscreteVsDiscreteTable"]');
+        browser.execute(function() { resultsViewPlotsTab.onDiscreteVsDiscretePlotTypeSelect({ value: "Table" }); });
         waitForAndCheckPlotsTab();
     });
     it("plots tab mutations vs clinical stacked bar plot", function() {
-        browser.click('input[data-test="DiscreteVsDiscreteTable"]');
+        browser.execute(function() { resultsViewPlotsTab.onDiscreteVsDiscretePlotTypeSelect({ value: "StacketBar" }); });
```
adamabeshouse: typo?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -104,8 +104,10 @@ export enum PlotType {
 }
 
 export enum DiscreteVsDiscretePlotType {
-    Table,
-    StackedBar
+    Bar = "Bar",
+    StackedBar = "StacketBar",
```
adamabeshouse: `StackedBar` with a `d`

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -156,8 +165,11 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
     @observable viewCopyNumber:boolean = false;
     @observable showRegressionLine = false;
     // discrete vs discrete settings
-    @observable discreteVsDiscretePlotType = DiscreteVsDiscretePlotType.StackedBar;
-    @observable stackedBarHorizontalBars = false;
+    @observable discreteVsDiscretePlotType:DiscreteVsDiscretePlotType = DiscreteVsDiscretePlotType.StackedBar;
+    //@observable discreteVsDiscretePlotType = DiscreteVsDiscretePlotType.StackedBar;
```
adamabeshouse: remove commented line?

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -29,11 +29,13 @@ export interface IStackedBarPlotProps {
     axisLabelX?: string;
     axisLabelY?: string;
     legendLocationWidthThreshold?: number;
+    percentage?:boolean;
+    stacked?:boolean;
```
adamabeshouse: I think the component instead should be called `MultipleCategoryBarPlot` since it no longer is only stacked. what do you think? as it is now, the `stacked` prop confuses me, because I'm like "isnt the component stacked?"

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -215,11 +222,18 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
...
+        }
         const majorCategoryCounts:{[major:string]:number} = {};
         for (const d of this.data) {
             for (const c of d.counts) {
                 majorCategoryCounts[c.majorCategory] = majorCategoryCounts[c.majorCategory] || 0;
-                majorCategoryCounts[c.majorCategory] += c.count;
+                if (this.props.stacked) {
+                    majorCategoryCounts[c.majorCategory] += c.count;
+                } else {
+                    majorCategoryCounts[c.majorCategory] = c.count > majorCategoryCounts[c.majorCategory] ? c.count : majorCategoryCounts[c.majorCategory];
```
adamabeshouse: `Math.max(c.count, majorCategoryCounts[c.majorCategory]` :)

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -245,8 +259,12 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
         return { x, y };
     }
 
+    @computed get offset() {
+        return this.barWidth;
+    }
+
     @computed get categoryAxisDomainPadding() {
-        return this.domainPadding;
+        return this.domainPadding + (this.props.stacked ? 0 : this.categoryCoord(this.data.length / 2));
```
adamabeshouse: How did you get to this formula? It seems like a lot of padding.

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -285,10 +303,16 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
...
+        let miscPadding = 100; // specifying chart width in victory doesnt translate directly to the actual graph size
         if (this.data.length > 0) {
-            const numBars = this.data[0].counts.length;
-            return this.categoryCoord(numBars - 1) + 2*this.categoryAxisDomainPadding + miscPadding;
+            let numBars = 0;
+            if (this.props.stacked) {
+                numBars = this.data[0].counts.length;
+            } else {
+                numBars = _.sumBy(this.data, x => x.counts.length);
+                miscPadding = miscPadding - this.categoryCoord(this.data.length / 2);
```
adamabeshouse: this is probably related to the formula in categoryAxisDomainPadding? can you explain? it might be helpful to pull out `@computed get SOMENAME() { return this.categoryCoord(this.data.length / 2); }` and reuse it and `SOMENAME` explains something about the rationale behind it

src/shared/components/plots/StackedBarPlot.tsx
```diff
@@ -483,13 +512,18 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
...
         )
         return barSpecs.map(spec=>(
             <VictoryBar
-                horizontal={this.props.horizontalBars}
                 style={{ data: { fill: spec.fill, width:this.barWidth } }}
-                data={ spec.data }
+                data={
+                    _.map(spec.data, datum => ({
+                        ...datum,
+                        y: datum.y + this.zeroCountOffset
```
adamabeshouse: I think this was the right way to handle it 👍

adamabeshouse:
@kalletlak left some more comments

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -394,15 +406,16 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
             case EventKey.utilities_showRegressionLine:
                 this.showRegressionLine = !this.showRegressionLine;
                 break;
-            case EventKey.utilities_discreteVsDiscreteTable:
+            case EventKey.utilities_discreteVsDiscretePlotType:
+                //TODO: update
```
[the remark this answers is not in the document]
kalletlak: ah, forgot to cleanup. this case is never invoked.

src/shared/components/plots/StackedBarPlot.tsx
kalletlak: yeah, I'll rename it

src/shared/components/plots/StackedBarPlot.tsx
kalletlak: I have done some trial and error, this one was giving perfect plot

adamabeshouse: basically you're making padding take about half the available space, right?

kalletlak: yes, adding half the width for each category

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1379,6 +1391,14 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
     }
 
+    @computed get isPercentage() {
+        return this.discreteVsDiscretePlotType === DiscreteVsDiscretePlotType.PercentageStackedBar;
+    }
+
+    @computed get isStacked() {
+        return this.discreteVsDiscretePlotType === DiscreteVsDiscretePlotType.StackedBar || this.discreteVsDiscretePlotType === DiscreteVsDiscretePlotType.PercentageStackedBar;
+    }
+
```
adamabeshouse: I was drawn to this because I think the names are confusing because they specifically refer to stacked bar plots, but they're in the general plots tab component. But on reading closely I think they should both just be consts declared in the plot() getter, in the MultipleCategoryBarPlot part, since they're only used in that one place.

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -280,15 +298,25 @@ export default class StackedBarPlot extends React.Component<IStackedBarPlotProps
...
         return {
             [this.countAxis]:this.countAxisDomainPadding,
-            [this.categoryAxis]:this.categoryAxisDomainPadding
+            [this.categoryAxis]:this.categoryAxisDomainPadding+this.additionalPadding
         };
     }
 
+    @computed get additionalPadding() {
+        //if not stacked add padding to move the bars right
+        return this.props.stacked ? 0 : this.categoryCoord(this.data.length / 2);
```
adamabeshouse: much better, thanks

src/pages/resultsView/plots/PlotsTab.tsx
kalletlak: updated

PR #2061, opened 2019-02-21
front-end token based authorization for whole slide viewer

dippindots:
Partial fix https://github.com/cBioPortal/cbioportal/issues/5624.

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -219,7 +219,16 @@ export function getDocsUrl(sourceUrl:string,docsBaseUrl?:string): string {
 }
 
 export function getWholeSlideViewerUrl(ids: string[], userName: string): string {
-    return ids.length > 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids.join('.svs;') + '.svs'}&user=${userName}&annotation=off` : ids.length === 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids[0] + '.svs'}&user=${userName}&annotation=off&filetree=off` : "";
+    try {
+        const tokenInfo = JSON.parse(AppConfig.serverConfig.mskWholeSlideViewerToken);
+        const token = `&token=${tokenInfo.token}`;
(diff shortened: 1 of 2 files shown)


src/shared/api/urls.ts
```diff
@@ -219,9 +219,11 @@ export function getDocsUrl(sourceUrl:string,docsBaseUrl?:string): string {
 }
 
 export function getWholeSlideViewerUrl(ids: string[], userName: string): string {
-    return ids.length > 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids.join('.svs;') + '.svs'}&user=${userName}&annotation=off` : ids.length === 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids[0] + '.svs'}&user=${userName}&annotation=off&filetree=off` : "";
+    const tokenInfo = JSON.parse(AppConfig.serverConfig.mskWholeSlideViewerToken);
```
alisman: you should always wrap a parsing operation in try{} catch {} in case the input is invalid.

dippindots: Yes! Thanks for the reminder.

src/shared/api/urls.ts
```diff
@@ -219,9 +219,11 @@ export function getDocsUrl(sourceUrl:string,docsBaseUrl?:string): string {
...
+    const token = `&token=${tokenInfo.token}`;
+    const time = `&t=${tokenInfo.time}`
+    return ids.length > 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids.join('.svs;') + '.svs'}&user=${userName}${time}${token}&annotation=off` : ids.length === 1 ? `https://slides.mskcc.org/cbioportal?ids=${ids[0] + '.svs'}&user=${userName}${time}${token}&annotation=off&filetree=off` : "";
```
alisman: i think this is a nested ternary operation, which should be avoided.  

overall seems over complicated.  if first condition is > 1 then why do you also need a subsequent ===1 ?
 
how about:  
map((name)=>name+".svs").join(";")

PR #2065, opened 2019-02-22
disable mutsig and gistic buttons

dippindots:
disable mutsig and gistic buttons currently. 
Fix https://github.com/cBioPortal/cbioportal/issues/5768.

### DIFF
--- a/src/shared/components/query/GeneSetSelector.tsx
+++ b/src/shared/components/query/GeneSetSelector.tsx
@@ -94,7 +94,8 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
 					onChange={option => this.store.geneQuery = option ? option.value : ''}
 				/>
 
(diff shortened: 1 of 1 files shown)


src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -94,7 +94,8 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
 					onChange={option => this.store.geneQuery = option ? option.value : ''}
 				/>
 
-				{!!(this.store.mutSigForSingleStudy.result.length || this.store.gisticForSingleStudy.result.length) && (
+				{/* currently disable the buttons */}
```
alisman: i'm not sure what this comment means.  currently?   i would just comment out the whole thing

alisman: and put a comment in explaining that we are hiding these buttons on ___ date

PR #2066, opened 2019-02-22
CircleCI: Persist frontend artifacts

inodb:
Instead of building them in each job. Build them only in install phase. Then
attach that volume again in the other jobs.

Solves this issue: https://discuss.circleci.com/t/saving-cache-stopped-working-warning-skipping-this-step-disabled-in-configuration/24423

You can see unit tests now take 2m45s vs 8m50s in master

https://circleci.com/workflow-run/35a12c59-5369-4b9c-806f-0fb60d283474

vs

https://circleci.com/workflow-run/ee731dd1-c57b-4184-b331-1001d0a43a0a

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,7 +3,7 @@
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
 defaults: &defaults
-  working_directory: ~/repo
+  working_directory: /tmp/repo
   docker:
     # specify the version you desire here
     - image: circleci/node:8.12.0-browsers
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -17,13 +17,17 @@ jobs:
...
           command:
             sudo npm install --global yarn@1.13.0
-
       - run:
           name: "Show yarn and node versions"
           command: |
             node --version
             yarn --version
       - checkout
+      - restore_cache:
```
inodb: this restores node_modules/ if it exists in the cache, the hash key for the cache is the `yarn.lock` file

.circleci/config.yml
```diff
@@ -36,31 +40,28 @@ jobs:
       - run:
           name: "Run build"
           command: 'yarn run build'
+      - persist_to_workspace:
```
inodb: this stores everything in the repo folder in the workspace volume, such that other jobs in the workflow can reuse the node_modules + build artifacts in dist/. This way they don't have to rebuild the whole thing themselves

PR #2069, opened 2019-02-26
change the exon style 

leexgh:
- [x] hide `/ x` in result view and appear in patient view.
- [x] show exon number on the top right summary.


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -449,6 +449,7 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                                 columnVisibilityProps={{
                                                     onColumnToggled: this.onMutationTableColumnVisibilityToggled
                                                 }}
+                                                showAllExon={true}
```
inodb: @leexgh i think showTotalNrOfExons might be better to understand as variable name

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -92,6 +92,7 @@ export default class ResultsViewMutationMapper extends MutationMapper<IResultsVi
                 enableHotspot={this.props.config.show_hotspot}
                 enableMyCancerGenome={this.props.config.mycancergenome_show}
                 enableCivic={this.props.config.show_civic}
+                showAllExon={false}
```
inodb: @leexgh 	no need to add this as a prop on resultsView. In other cases where we have a difference between the visualization of results and patient view, we hardcode in the patient view mutation table what this property should be (in the patient view table case `true`)

inodb: see e.g.: https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/mutation/PatientViewMutationTable.tsx#L92-L93

and the one you added yourself: 

https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/mutation/PatientViewMutationTable.tsx#L112

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -86,6 +86,7 @@ export interface IMutationTableProps {
     initialSortDirection?:SortDirection;
     paginationProps?:IPaginationControlsProps;
     showCountHeader?:boolean;
+    showAllExon?:boolean;
```
inodb: same here, no need to add a prop. This ofc is also a good way to do it, but just because we've solved this using customizations in PatientViewMutationTable before it's better to follow the same pattern

src/shared/components/mutationTable/column/ExonColumnFormatter.tsx
```diff
@@ -41,11 +42,12 @@ export default class ExonColumnFormatter {
                 return exonData;
             }
             else {
-                return <span style = {{display:"inline-block", float:"right"}}>            
-                    <span style = {{float:"left",width:"24px", textAlign:"right"}}> {exonData.split("/")[0]} </span>  
-                    <span style = {{float:"left", marginLeft:"4px", marginRight:"4px"}}>/</span>
-                    <span style = {{float:"left",width:"24px", textAlign:"left"}}> {exonData.split("/")[1]} </span>
-                </span>
+                let displayStyle = showAllExon?"inline":"none";
```
inodb: Although this works I personally think it's a bit hard to read it this way. Cause you have to read the HTML to figure out which things are not being displayed. I think it's easier to read to do something like (not 100% sure if this works, :

```
class Example extends Component {
  render() {
    const showTotalNrOfExons = false;
    return (
      <span>
      5
       {showTotalNrOfExons && (
         <span> / 5</span>
       )}
      </span>
    )
  }
}
```
see e.g.: https://codesandbox.io/s/ovwl49xw05

inodb: Makes sense?

src/shared/components/mutationTable/MutationTable.tsx
leexgh: yes i see, if i don't add the new prop here, do you think it's ok to have `false` here by default?
```
render: (d:Mutation[]) => (this.props.genomeNexusCache
       ? ExonColumnFormatter.renderFunction(d, this.props.genomeNexusCache, false)
       : <span></span>),
```

inodb: @leexgh yes fine to use false, or don't pass a third argument. Since it's an optional argument, it will be undefined and will work as expected

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -99,5 +101,9 @@ export default class ResultsViewMutationTable extends MutationTable<IResultsView
         this._columns[MutationTableColumnType.NUM_MUTATIONS].shouldExclude = ()=>{
             return !this.props.mutationCountCache;
         };
+
+        // customization for columns
+        this._columns[MutationTableColumnType.EXON].render = 
+        (d:Mutation[]) => (ExonColumnFormatter.renderFunction(d, this.props.genomeNexusCache, false));
```
onursumer: indentation here would be useful

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -196,6 +196,13 @@ export default class MutationMapper<P extends IMutationMapperProps> extends Reac
                         ) : '-'}
                     </span>
                 </div>
+                <div style={{paddingTop:"1em"}}>
+                    <span data-test="GeneSummaryNrOfExons">{'Number of Exons: '}
```
onursumer: I don't think you need a string variable here, you can simply do something like this:
```suggestion
                    <span data-test="GeneSummaryNrOfExons">
                        Number of Exons: {transcript && transcript.exons && transcript.exons.length > 0? (
                            transcript.exons.length
                        ) : '-'}
                            
```

src/shared/components/mutationTable/column/ExonColumnFormatter.tsx
```diff
@@ -10,11 +10,12 @@ import styles from "./exon.module.scss";
 export default class ExonColumnFormatter {
     
     public static renderFunction(data:Mutation[],
-                                 genomeNexusCache:GenomeNexusCache|undefined) {
+                                 genomeNexusCache:GenomeNexusCache|undefined,
+                                 showTotalNrOfExons?:boolean) {
```
onursumer: don't be afraid of spelling out the variables, you can say `showTotalNumberOfExons` :)

leexgh:
@inodb @onursumer thanks! all updated

PR #2071, opened 2019-02-26
replace the back-end pdf generate function with window.print()

dippindots:
replace the back-end pdf generate function with window.print().
Fix https://github.com/cBioPortal/cbioportal/issues/5738.

### DIFF
--- a/src/shared/components/downloadControls/DownloadControls.tsx
+++ b/src/shared/components/downloadControls/DownloadControls.tsx
@@ -51,6 +51,25 @@ function makeMenuItem(spec:ButtonSpec) {
     );
 }
 
+function openPrintTab(svg: SVGElement) {
+    const mywindow = window.open('', 'Print', 'height=600,width=800');
+    if (mywindow) {
+        // write the element to the new tab
+        mywindow.document.write("<?xml version='1.0'?>"+replaceUnicodeChars((new XMLSerializer()).serializeToString(svg)));    
+        mywindow.document.close();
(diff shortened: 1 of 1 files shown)


inodb:
This works nice in Chrome and Brave, but in Firefox and Safari it's not super obvious that you have to click on the bottom left button, which might get people to accidentally print the plot:

screen shot 2019-02-26 at 6 25 47 pm

dippindots:
@inodb Yes, that's true, thanks for the feedback!

src/shared/components/downloadControls/DownloadControls.tsx
```diff
@@ -51,6 +51,25 @@ function makeMenuItem(spec:ButtonSpec) {
     );
 }
 
+function openPrintTab(svg: SVGElement) {
+    const mywindow = window.open('', 'Print', 'height=600,width=800');
+    if (mywindow) {
+        // write the element to the new tab
+        mywindow.document.write("<?xml version='1.0'?>"+replaceUnicodeChars((new XMLSerializer()).serializeToString(svg)));
```
alisman: it is possible that the issues Tali found are due to the browser handling the xml file differently.   one idea would be to try opening an HTML file with just the SVG inside of it.  if all the browsers can successfully print the current page (with visualization inside layout) but some fail with xml file, then this would be a good route to try.  document.write may work, or look at something called a "data uri", which allows you to set the HTML in the url itself.

PR #2073, opened 2019-02-26
fix mutex not shown when only one gene with multiple alteration type

dippindots:
fix mutex not shown when only one gene with multiple alteration type
Fix https://github.com/cBioPortal/cbioportal/issues/5791.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -192,7 +192,8 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                     </MSKTab>
                 },
                 hide:()=>{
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -192,7 +192,12 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                     </MSKTab>
                 },
                 hide:()=>{
-                    return this.resultsViewPageStore.hugoGeneSymbols.length < 2;
+                    if (!this.resultsViewPageStore.isSampleAlteredMap.isComplete) {
```
alisman: i think an "or statement" might make this more clear.  also put a comment here.   we are using the size of isSampleAlteredMap as a proxy for the number of things we have to compare.   that's business logic we should make that more clear.

PR #2074, opened 2019-02-27
Fix the properties to set defaults for Oncoprint mutation annotations

oplantalech:
The current properties `oncoprint.custom_driver_annotation.default`, `oncoprint.custom_driver_tiers_annotation.default` and `oncoprint.oncokb_hotspots.default` do not work as expected in the [documentation](https://github.com/cBioPortal/cbioportal/blob/master/docs/portal.properties-Reference.md). 

This PR proposes to fix this and change some of the name properties so that it is more clear what they do. All properties are boolean, if they are set to `true`, then the checkbox in the "Mutations" menu of the Oncoprint will be checked by default:

- `oncoprint.custom_driver_annotation.binary.default`: referring to the "binary" custom annotations.
- `oncoprint.custom_driver_annotation.tiers.default`: referring to the tiers custom annotations.
- `oncoprint.oncokb.default`: referring to the OncoKB annotations.
- `oncoprint.hotspots.default`: referring to Hotspots annotations.

This PR goes together with cBioportal/cbioportal#5795.

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -345,7 +345,8 @@ export type DriverAnnotationSettings = {
     cbioportalCountThreshold:number;
     cosmicCount:boolean;
     cosmicCountThreshold:number;
-    driverFilter:boolean;
+    customBinary:boolean;
+    customTiersDefault: boolean;
     driverTiers:ObservableMap<boolean>;
     hotspots:boolean;
     oncoKb:boolean;
@@ -372,9 +373,9 @@ export class ResultsViewPageStore {
             cbioportalCountThreshold: 0,
             cosmicCount: false,
             cosmicCountThreshold: 0,
-            driverFilter: false,
             driverTiers: observable.map<boolean>(),
 
+            _customBinary: undefined,
             _hotspots:false,
             _oncoKb:false,
             _ignoreUnknown: false,
@@ -402,12 +403,20 @@ export class ResultsViewPageStore {
                     this.hotspots ||
                     this.cbioportalCount ||
                     this.cosmicCount ||
-                    this.driverFilter ||
+                    this.customBinary ||
                     this.driverTiers.entries().reduce((oneSelected:boolean, nextEntry:[string, boolean])=>{
                         return oneSelected || nextEntry[1];
                     }, false);
 
                 return anySelected;
+            },
+
+            get customBinary() {
+                return this._customBinary === undefined ? AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default : this._customBinary;
+            },
+
+            get customTiersDefault() {
+                return AppConfig.serverConfig.oncoprint_custom_driver_annotation_tiers_default;
             }
         });
 
@@ -500,11 +509,9 @@ export class ResultsViewPageStore {
         this.driverAnnotationSettings.cbioportalCountThreshold = 10;
         this.driverAnnotationSettings.cosmicCount = false;
         this.driverAnnotationSettings.cosmicCountThreshold = 10;
-        this.driverAnnotationSettings.driverFilter = !!AppConfig.serverConfig.oncoprint_custom_driver_annotation_default;
         this.driverAnnotationSettings.driverTiers = observable.map<boolean>();
-
-        this.driverAnnotationSettings.hotspots = !AppConfig.serverConfig.oncoprint_oncokb_hotspots_default;
-        (this.driverAnnotationSettings as any)._oncoKb = !AppConfig.serverConfig.oncoprint_oncokb_hotspots_default;
+        (this.driverAnnotationSettings as any)._oncoKb = !!AppConfig.serverConfig.oncoprint_oncokb_default;
+        this.driverAnnotationSettings.hotspots = !!AppConfig.serverConfig.oncoprint_hotspots_default;
         (this.driverAnnotationSettings as any)._ignoreUnknown = !!AppConfig.serverConfig.oncoprint_hide_vus_default;
     }
 
@@ -2220,8 +2227,9 @@ export class ResultsViewPageStore {
             initializeCustomDriverAnnotationSettings(
                 result!,
                 this.driverAnnotationSettings,
-                !(_.isEmpty(AppConfig.serverConfig.oncoprint_custom_driver_annotation_tiers_menu_label)),
-                AppConfig.serverConfig.oncoprint_oncokb_hotspots_default === "custom"
+                this.driverAnnotationSettings.customTiersDefault,
+                this.driverAnnotationSettings.oncoKb,
+                this.driverAnnotationSettings.hotspots
             );
         }
     });
@@ -2365,7 +2373,7 @@ export class ResultsViewPageStore {
                     this.getCosmicCount.result!(mutation) >= this.driverAnnotationSettings.cosmicCountThreshold);
 
                 const customDriverBinary:boolean =
-                    (this.driverAnnotationSettings.driverFilter &&
+                    (this.driverAnnotationSettings.customBinary &&
                         mutation.driverFilter === "Putative_Driver") || false;
 
                 const customDriverTier:string|undefined =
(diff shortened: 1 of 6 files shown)


alisman:
@oplantalech unless i'm confused (very possible!)  i want to try to avoid having properties called "defaults."   we had a practice in the past of having a config property and then a default property to specify the default for that property.  i don't think that makes sense.  the default should simple be the value we assign in serverConfigDefaults.   Does that make sense?   That does mean killing these "default" props in the old configuration.  But @jjgao seemed to suggest a while ago that was safe.  Why would a user specify a default instead of just specifying the value itself.  The defaults are essentially internal configuration under our control (not instance user).

adamabeshouse:
@alisman as in the other PR, default is an actual config https://github.com/cBioPortal/cbioportal/blob/master/docs/portal.properties-Reference.md#custom-annotation-of-driver-and-passenger-mutations

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -352,6 +353,7 @@ export class ResultsViewPageStore {
             cosmicCount: false,
             cosmicCountThreshold: 0,
             driverFilter: false,
+            selectedDriverTiers: false,
```
adamabeshouse: why do we need this on top of the driverTiers object?

alisman: @oplantalech can you respond to this here?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -388,6 +390,22 @@ export class ResultsViewPageStore {
...
                 return anySelected;
+            },
+
+            get driverFilter() {
+                return AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default;
+            },
+
+            set driverFilter(val:boolean) {
+                this._driverFilter = val;
+            },
```
adamabeshouse: I'm not completely sure what you're going for but I think what you want is to change the initial value of _driverFilter to be undefined on line 359. Then `get driverFilter()` should return ` this._driverFilter === undefined ? AppConfig... : this._driverFilter`.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -388,6 +390,22 @@ export class ResultsViewPageStore {
...
+
+            get isDriverTiersSelected() {
+                return AppConfig.serverConfig.oncoprint_custom_driver_annotation_tiers_default;
+            },
+
+            set isDriverTierSelected(val:boolean) {
+                this._isDriverTiersSelected = val;
```
adamabeshouse: I don't see the point of this `set` on line 407, you don't use it anywhere. I think we had a misunderstanding with my last comment, but I think I misunderstood so we should forget about it. Let's just get rid of this.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -388,6 +390,22 @@ export class ResultsViewPageStore {
...
+
+            get driverFilter() {
+                return AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default;
+            },
+
+            set driverFilter(val:boolean) {
+                this._driverFilter = val;
+            },
+
+            get isDriverTiersSelected() {
```
adamabeshouse: let's rename this. It's `defaultDriverTierSelected`

src/pages/resultsView/ResultsViewPageStore.ts
oplantalech: @alisman I just fixed it, and made a getter instead

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -388,6 +390,14 @@ export class ResultsViewPageStore {
                     }, false);
 
                 return anySelected;
+            },
+
+            get driverFilter() {
```
adamabeshouse: I just realized that driverFilter is too general a name. It should be customBinaryDriverFilter, and for tiers lets change it to customTiersDriverFilter.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -352,9 +353,10 @@ export class ResultsViewPageStore {
             cbioportalCountThreshold: 0,
             cosmicCount: false,
             cosmicCountThreshold: 0,
-            driverFilter: false,
             driverTiers: observable.map<boolean>(),
 
+            _defaultDriverTierSelected: false,
```
adamabeshouse: there's no need for this. It's not used, right?

Sjoerd-van-Hagen:
@oplantalech everything fixed in this PR?

oplantalech:
@Sjoerd-van-Hagen yes! I have rebased the PR to keep it up-to-date.

Sjoerd-van-Hagen:
@adamabeshouse can you re-review?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -396,12 +397,20 @@ export class ResultsViewPageStore {
...
                     }, false);
 
                 return anySelected;
+            },
+
+            get customBinaryDriverFilter() {
+                return this._customBinaryDriverFilter === undefined ? AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default : this._customBinaryDriverFilter;
+            },
+
+            get customTiersDriverFilter() {
```
adamabeshouse: Reading the name you've chosen, I would think that it has the same sort of function as `customBinaryDriverFilter`, when in fact it's completely different - `customBinaryDriverFilter` is whether binary driver annotation is enabled, whereas this here is just the default value of tiers selected. Please try to write the names of functions and variables to reflect the functionality of how they are used. This is critically important when working on a large codebase like cbioportal, or else little by little over time the code will become unreadable. Can you please rename this to reflect better how it is used?

oplantalech: @adamabeshouse I have re-evaluated the names of the two variables that I introduce and I have changed their name. I think now they reflect what they contain.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -397,19 +397,19 @@ export class ResultsViewPageStore {
...
             },
 
-            get customBinaryDriverFilter() {
-                return this._customBinaryDriverFilter === undefined ? AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default : this._customBinaryDriverFilter;
+            get isCustomBinaryDriverFilterEnabled() {
+                return this._isCustomBinaryDriverFilterEnabled === undefined ? AppConfig.serverConfig.oncoprint_custom_driver_annotation_binary_default : this._isCustomBinaryDriverFilterEnabled;
             },
 
-            get customTiersDriverFilter() {
+            get iscustomTiersDriverFilterEnabled() {
```
adamabeshouse: this name still isn't right. this is the *default*, right?

oplantalech: First of all, I have fixed the names, I agree with you. Just to clarify why "filter" is here, I think I wrote that name because what the annotations are doing is _filtering_ the mutations in the Oncoprint by type - but of course there is no point on having this name here and it just adds more confusion.

adamabeshouse: Thanks very much, it looks great now. Sorry if I was unpleasant.

oplantalech: No problem, the code needs to be clear and understandable for everybody. Thanks a lot for spending time on it 😄

adamabeshouse:
@oplantalech looks like there are some test failures, are they related?

oplantalech:
@adamabeshouse In my PR there are 3 screenshot tests failing coming from the Study View, they don't seem related to my PR since they also fail in rc: the points on the plot Mutation Count vs Fraction Genome Altered are missing. Interestingly though, the tables Mutated Genes and CNA Genes report different content in rc but they look the same in my PR (that might explain why there are more screenshot tests failing in rc than in my PR).

The failing tests in master and rc should be fixed... are you working on that?

PR #2076, opened 2019-02-28
add clinical tab for single sample on patient view page

dippindots:
show clinical tab for single sample on the patient view page.
Fix https://github.com/cBioPortal/cbioportal/issues/3280.


dippindots:
@zhx828 Yes, the fix is just removing the check. @alisman One thing is not sure, do we still show patient table when we are looking at a single sample?
image

PR #2079, opened 2019-03-01
change exon column header

leexgh:
change exon column header and delete exon in summary. 
Header style e.g.: ```Exon\n(11 in total)```

result view:
image


alisman:
@onursumer can you take a look at this?

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -65,6 +65,9 @@ export default class ResultsViewMutationMapper extends MutationMapper<IResultsVi
 
     protected mutationTableComponent(): JSX.Element|null
     {
+        const canonicalTranscriptId = this.props.store.canonicalTranscript.result && this.props.store.canonicalTranscript.result.transcriptId;
+        const transcript = this.props.store.activeTranscript && (this.props.store.activeTranscript === canonicalTranscriptId)? this.props.store.canonicalTranscript.result : this.props.store.transcriptsByTranscriptId[this.props.store.activeTranscript!!];
```
alisman: what's the double exclamation mark for?

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -65,6 +65,9 @@ export default class ResultsViewMutationMapper extends MutationMapper<IResultsVi
 
     protected mutationTableComponent(): JSX.Element|null
     {
+        const canonicalTranscriptId = this.props.store.canonicalTranscript.result && this.props.store.canonicalTranscript.result.transcriptId;
```
alisman: This seems to introduce some new asyc dependicies for the table.  It could cause cost re-render cycles. @onursumer should we just wait until these resolve before rendering the table?

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
leexgh: sorry my mistake, it should be a single mark

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
onursumer: @alisman good point. We need to make sure that this wouldn't significantly increase the initial load time though.

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -65,6 +65,9 @@ export default class ResultsViewMutationMapper extends MutationMapper<IResultsVi
...
+        const transcript = this.props.store.activeTranscript && (this.props.store.activeTranscript === canonicalTranscriptId)? this.props.store.canonicalTranscript.result : this.props.store.transcriptsByTranscriptId[this.props.store.activeTranscript!];
```
onursumer: This line is really long, it's a good practice to divide it into multiple lines. Maybe something like this:

```suggestion
        const transcript = this.props.store.activeTranscript && (this.props.store.activeTranscript === canonicalTranscriptId) ?
            this.props.store.canonicalTranscript.result : this.props.store.transcriptsByTranscriptId[this.props.store.activeTranscript!];
```

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -10,6 +10,7 @@ import ExonColumnFormatter from "shared/components/mutationTable/column/ExonColu
 
 export interface IResultsViewMutationTableProps extends IMutationTableProps {
     // add results view specific props here if needed
+    totalExon?:string;
```
onursumer: `totalExonNumber` or `totalNumberOfExons` would be better. 

Also, can this number be a `not a number`? If not, maybe declare it as a `number` instead of a `string`.

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -105,5 +106,7 @@ export default class ResultsViewMutationTable extends MutationTable<IResultsView
         // customization for columns
         this._columns[MutationTableColumnType.EXON].render = 
             (d:Mutation[]) => (ExonColumnFormatter.renderFunction(d, this.props.genomeNexusCache, false));
+        this._columns[MutationTableColumnType.EXON].headerRender = 
+            () => <span style = {{display:'inline-block'}}>Exon <div>{this.props.totalExon} in total</div></span>;
```
onursumer: Is the `<div></div>` added to render the number in a new line?

alisman:
Hi @leexgh i'm going to close this for now.  if you want to response to feedback, you can do so and re-open.  thanks!

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
leexgh: @onursumer could you explain more? what to do here?

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
leexgh: @onursumer I think `totalNumberOfExons` should be number when it has exon data, but I'm not sure whether it always has data, so I put a `None` here as well. In this case maybe we could keep using string?

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
leexgh: @onursumer yes, I change to <br/> instead

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
onursumer: @leexgh you can try updating the `isMutationTableDataLoading` with the promise `this.props.store.canonicalTranscript`, and see if there is any significant initial load time increase.

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
onursumer: let's keep it as a `string` for now

jjgao:
Maybe put it in parentheses: `(11 in total)` ?

leexgh:
@jjgao done

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
leexgh: @onursumer and I test it, looks no significant initial load time increase.

PR #2083, opened 2019-03-04
Show description over refs when available

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4365

### DIFF
--- a/src/shared/components/annotation/OncoKbCard.tsx
+++ b/src/shared/components/annotation/OncoKbCard.tsx
@@ -95,7 +95,8 @@ export default class OncoKbCard extends React.Component<OncoKbCardProps>
                         cancerType: string,
                         pmidData: ICache<any>,
                         pmids:number[],
-                        abstracts:ArticleAbstract[])
+                        abstracts:ArticleAbstract[],
+                        description:string)
     {
         const levelTooltipContent = () => (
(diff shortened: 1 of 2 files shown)


src/shared/components/annotation/OncoKbCard.tsx
```diff
@@ -95,23 +95,31 @@ export default class OncoKbCard extends React.Component<OncoKbCardProps>
...
-                <div style={{maxWidth: "400px", maxHeight: "400px", overflowY: "auto"}}>
-                    <ul className="list-group" style={{margin: '0 5px'}}>
-                        {this.abstractList(abstracts)}
-                        {this.pmidList(pmids, pmidData)}
-                    </ul>
-                </div>
-            ) : <span/>;
+        const treatmentTooltipContent = (abstracts.length > 0 || pmids.length > 0 || description) ?
+            () => {
+                return <div style={{maxWidth: "400px", maxHeight: "400px", overflowY: "auto"}}>
```
onursumer: do we need to have an explicit return statement here? why not keep the original style (i.e: no curly brackets no `return` keyword, just the `JSX` content)?

zhx828: I was having some troubles using that, but seems fine know. Let me revert the change.

PR #2084, opened 2019-03-04
enrichments - frequency scatter chart

adamabeshouse:
the issue is posted for comparison, but @jjgao asked that we merge it into master for the results page


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -41,6 +42,19 @@ export function getAlterationScatterData(alterationEnrichments: AlterationEnrich
     });
 }
 
+export function getAlterationFrequencyScatterData(alterationEnrichments: AlterationEnrichmentRow[], queryGenes: string[]): IMiniFrequencyScatterChartData[] {
+    return alterationEnrichments.filter(a => !queryGenes.includes(a.hugoGeneSymbol)).map((alterationEnrichment) => {
```
adamabeshouse: I don't think this needs testing since theres no logic

src/pages/resultsView/enrichments/MiniFrequencyScatterChart.tsx
```diff
@@ -0,0 +1,219 @@
...
+                        }
+                    ];
+                }
+            }
+        }];
+
+        return (
+            <div className="posRelative">
+                <div className="borderedChart inlineBlock" style={{position:"relative"}}>
+                    <VictoryChart containerComponent={<VictorySelectionContainer containerRef={this.svgRef}
```
alisman: i think we should make these attributes referesnces to methods or variables that hold the components.  kind of a lot of nesting.

PR #2087, opened 2019-03-04
filter nonexistent cases from groups, and show warning icon

adamabeshouse:
for the screenshots, I just considered samples nonexistent at random

image

image

image

### DIFF
--- a/src/pages/groupComparison/GroupComparisonUtils.tsx
+++ b/src/pages/groupComparison/GroupComparisonUtils.tsx
@@ -196,18 +196,19 @@ export function getNumSamples(
     return _.sum(group.studies.map(study=>study.samples.length));
 }
 
-export function filterNonExistentSamples(
+export function finalizeStudiesAttr(
     groupData:Pick<SessionGroupData, "studies">,
     sampleSet:ListIndexedMap<Sample> // key: [studyId, sampleId]
 ) {
-    // keep track of nonexisting samples, existing patients
+    // (1) filter out, and keep track of nonexisting samples
+    // (2) add `patients` object
     const nonExistentSamples = [];
     const studies = [];
 
     for (const study of groupData.studies) {
         const studyId = study.id;
         const samples = [];
-        const patients = [];
+        let patients = [];
         for (const sampleId of study.samples) {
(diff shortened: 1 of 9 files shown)


src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -279,4 +326,17 @@ export function getPieChartGroupFilters(
         });
     }
     return newFilters;
+}
+
+export function getMissingSamplesMessage(
```
alisman: we should learn signature of stateless components and use it

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -53,7 +59,7 @@ export default class ComparisonGroupManager extends React.Component<IComparisonG
         await:()=>[this.props.store.comparisonGroups],
         invoke:()=>Promise.resolve(
             // TODO: fuzzy string search?
-            this.props.store.comparisonGroups.result!.filter(group=>group.data.name.toLowerCase().indexOf(this.groupNameFilter.toLowerCase()) > -1)
+            this.props.store.comparisonGroups.result!.filter(group=>group.name.toLowerCase().indexOf(this.groupNameFilter.toLowerCase()) > -1)
```
alisman: lets use RegExp(blah,"i").test  and get rid of all the lowercase and indexOf stuff

src/pages/groupComparison/styles.module.scss
```diff
@@ -14,14 +14,9 @@
...
-        :global(div.groupItem):first-child {
-            margin-top:0px;
-        }
-        :global(div.groupItem):last-child {
+        :global(div.checkbox) {
             margin-bottom:0px;
+            margin-top:0px;
         }
 
         :global(.markedForDeletion) {
```
alisman: i know this isn't part of this PR but seems like maybe "markedForDeletion" would be a good to not have global here.    I guess maybe groupItem is another.

adamabeshouse: kk. I actually got rid of groupItem altogether

PR #2089, opened 2019-03-05
study view - allow bar charts to open comparison sessions

adamabeshouse:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -415,9 +415,124 @@ export class StudyViewPageStore {
...
+                    // create groups using data
+                    let groups:SessionGroupData[];
+                    let patientToSamples:{[uniquePatientKey:string]:SampleIdentifier[]} = {};
+                    if (clinicalAttribute.patientAttribute) {
+                        patientToSamples = _.groupBy(selectedSamples, s=>s.uniquePatientKey);
+                    }
+                    switch (distinctValues.length) {
+                        case 1:
+                        case 2:
+                        case 3:
```
zhx828: We have a threshold set for bar chart at backend if the unique value number is 5 then we don't group. We could use that here as well.

adamabeshouse: since we're looking at quartiles, I think 4 makes sense, no?

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -344,4 +408,71 @@ export function MissingSamplesMessage(
...
+            naCases = _.uniqWith(naCases, (a,b)=>(a.studyId === b.studyId && a.patientId === b.patientId));
+        } else {
+            naCases = _.uniqWith(naCases, (a,b)=>(a.studyId === b.studyId && a.sampleId === b.sampleId));
+        }
+        naBin.data = naCases;
+        sortedBins.push(naBin);
+    }
+}
+
+export function getDataBinGroupName(
```
adamabeshouse: needs tests

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -344,4 +408,71 @@ export function MissingSamplesMessage(
...
+
+export type DataBinWithData = DataBin & { data:Pick<ClinicalData, "studyId"|"sampleId"|"patientId"|"uniquePatientKey">[] };
+
+export function putDataIntoSortedBins(
+    data:Pick<ClinicalData, "value"|"studyId"|"sampleId"|"patientId"|"uniquePatientKey"|"uniqueSampleKey">[],
+    sortedBins: DataBinWithData[],
+    selectedSamples:Pick<Sample, "uniquePatientKey"|"uniqueSampleKey"|"studyId"|"sampleId"|"patientId">[],
+    isPatientAttribute:boolean,
+    naBin?:DataBinWithData
+):void {
```
adamabeshouse: needs tests

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -423,9 +426,91 @@ export class StudyViewPageStore {
...
+                    });
+
+                    // copy bins, take out NA bin if there is one
+                    let binsWithData = _.cloneDeep(bins)
+                        .map(bin=>{
+                            (bin as any).data = [];
+                            return bin as DataBinWithData;
+                        });
+
+                    const naBinIndex = binsWithData.findIndex(bin=>bin.specialValue === "NA");
```
zhx828: You need to find all nonnumerical values. The bar chart supports string.
For instance, Tumor Purity in https://cbioportal.mskcc.org/study?id=mskimpact
Also some special ranges, for instance '<18'  in genie `Age at Which Sequencing was Reported`. http://genie.cbioportal.org/study?id=genie_public
@jjgao @adamabeshouse this is basically redoing all the logics at the server side. @onursumer can we make a API call for this? Give a clinical attribute id, study view filter, and return the sample ids in every data bin.

adamabeshouse: that would be much much much better. sample ids as well as the name, so frontend doesnt have to combine "specialValue" and endpoint etc.

zhx828:
I believe it should be the case. Otherwise we need to implement the same
logic front and back.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -423,9 +425,124 @@ export class StudyViewPageStore {
...
+                                    studies,
+                                    origin: this.studyIds,
+                                    studyViewFilter: getNumberAttributeGroupFilters(this.filters, clinicalAttribute, [value, value])
+                                };
+                            });
+                            break;
+                        default:
+                            // set up groups for quartiles
+                            // first, get the limit for each quartile
+                            const quartileTops:number[] = [];
```
adamabeshouse: test this part

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -309,7 +309,40 @@ export function getTabId(pathname:string) {
...
+                end:range[1]
+            } as ClinicalDataIntervalFilterValue]
+        ), (a:ClinicalDataIntervalFilterValue, b:ClinicalDataIntervalFilterValue)=>{
+            return (a.start === b.start && a.end === b.end);
+        });
+    } else {
+        newFilters.clinicalDataIntervalFilters.push({
+            attributeId: clinicalAttribute.clinicalAttributeId,
+            clinicalDataType,
+            values:[{start:range[0], end:range[1], value:"what"}]
```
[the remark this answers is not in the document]
adamabeshouse: oops, that was placeholder :D good catch. yeah I think it would be unnecessary to have value here

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -423,9 +425,124 @@ export class StudyViewPageStore {
...
+                        clinicalDataMultiStudyFilter: {
+                            attributeIds: [clinicalAttribute.clinicalAttributeId],
+                            identifiers: selectedSamples.map(s=>({ studyId: s.studyId, entityId: s[entityIdKey] }))
+                        }
+                    });
+
+                    // sort values
+                    // group into halves, thirds, or fourths, depending on how many distinct values there are
+                    const groupedByValue = _.groupBy(data, d=>parseFloat(d.value));
+                    const distinctValues = _.sortBy(Object.keys(groupedByValue).map(v=>parseFloat(v)));
```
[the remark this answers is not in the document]
adamabeshouse: good idea

src/pages/studyView/StudyViewPageStore.ts
zhx828: It does, but when there are only five distinct values probably makes more sense showing five groups. But again, back to original discussion which we vote down. So no need to make it complicated. What you are doing right now is correct.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -309,7 +324,40 @@ export function getTabId(pathname:string) {
...
+    const newFilters = _.cloneDeep(baseFilters);
+
+    newFilters.clinicalDataIntervalFilters = newFilters.clinicalDataIntervalFilters || [];
+    const existingFilter = newFilters.clinicalDataIntervalFilters.find(f=>{
+        return f.attributeId === clinicalAttribute.clinicalAttributeId &&
+            f.clinicalDataType === clinicalDataType;
+    });
+
+    if (existingFilter) {
+        existingFilter.values = _.uniqWith(
```
alisman: i think this could use a comment explaining why we have to edit existing filter

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -347,4 +395,23 @@ export function MissingSamplesMessage(
             </div>
         </div>
     );
+}
+
+export function sortDataIntoQuartiles<D>(
```
alisman: I think jstat has utilties for generating quartiles. could use it?

src/shared/lib/sortedFindWith.ts
```diff
@@ -0,0 +1,31 @@
+// Use binary search to find the element in `sortedArray` for which `oracle` returns 0.
+// It's assumed that sortedArray is sorted, i.e. `sortedArray.map(oracle) = [-1,-1,...,-1,0,1,1,...,1]`
+// Behavior undefined on non-sorted arrays.
+
+export function sortedFindWith<T>(
```
alisman: i don't see where this is being used in code.  when i search for "sortedFindWith" i only get this file and spec file.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -424,9 +427,119 @@ export class StudyViewPageStore {
         this._comparisonGroupsChangeCount += 1;
     }
 
-    private createPieChartComparisonSession(
+    private async createNumberAttributeComparisonSession(
         clinicalAttribute:ClinicalAttribute,
-        clinicalAttributeValues:{ value:string, color:string}[],
+        statusCallback:(phase:LoadingPhase)=>void
```
alisman: probably really late to the game here, but is there no way we could accomplish this with our usual status tracking mechanism (networks for mobx promises).  just wanted to discuss.

src/pages/groupComparison/GroupComparisonUtils.tsx
adamabeshouse: I'm using the jstat library for calculating the quartile limits, but I didn't see a method for sorting data into those quartiles

src/shared/lib/FormatUtils.tsx
```diff
@@ -19,6 +19,19 @@ export function toPrecision(value:number, precision:number, threshold:number)
     return ret;
 }
 
+export function toFixedWithoutTrailingZeros(value:number, digits:number):string {
```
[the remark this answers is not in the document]
adamabeshouse: oh okay

src/shared/lib/sortedFindWith.ts
adamabeshouse: I had originally written it for changes which I then removed. But I thought it was a useful function, and I wrote it and tested it, so I decided to leave it in.

PR #2092, opened 2019-03-06
group comparison page - allow moving overlap into new group

adamabeshouse:
image

image

image

### DIFF
--- a/src/pages/groupComparison/GroupComparisonPage.tsx
+++ b/src/pages/groupComparison/GroupComparisonPage.tsx
@@ -22,7 +22,7 @@ import autobind from "autobind-decorator";
 import {AppStore} from "../../AppStore";
 import _ from "lodash";
 import ClinicalData from "./ClinicalData";
-import ReactSelect from "react-select";
+import ReactSelect from "react-select2";
 
 export enum GroupComparisonTab {
     OVERLAP = "overlap",
@@ -197,15 +197,16 @@ export default class GroupComparisonPage extends React.Component<IGroupCompariso
         } 
     });
 
-    @computed get overlapStrategySelector() {
-        if (this.store._selectionInfo.isComplete) {
-            if (this.store._selectionInfo.result.overlappingSamples.length === 0 &&
-            this.store._selectionInfo.result.overlappingPatients.length === 0) {
+    readonly overlapStrategySelector = MakeMobxView({
+        await:()=>[this.store._selectionInfo],
+        render:()=>{
+            if (this.store._selectionInfo.result!.overlappingSamples.length === 0 &&
+                this.store._selectionInfo.result!.overlappingPatients.length === 0) {
                 return null;
             } else {
                 // (${caseCountsInParens(this.store._selectionInfo.result.overlappingSamples, this.store._selectionInfo.result.overlappingPatients)})
                 return (
-                    <div style={{width:333, zIndex:20}}>
+                    <div style={{width:355, zIndex:20}}>
                         <ReactSelect
                             name="select overlap strategy"
                             onChange={(option:any|null)=>{
@@ -214,19 +215,19 @@ export default class GroupComparisonPage extends React.Component<IGroupCompariso
                                 }
                             }}
                             options={[
-                                { label: "Include overlapping samples", value: OverlapStrategy.INCLUDE},
-                                { label: `Exclude overlapping samples`, value: OverlapStrategy.EXCLUDE},
-                                { label: "Treat overlapping samples as a separate group", value: OverlapStrategy.GROUP }
+                                { label: OverlapStrategy.INCLUDE, value: OverlapStrategy.INCLUDE},
+                                { label: OverlapStrategy.EXCLUDE, value: OverlapStrategy.EXCLUDE},
+                                { label: OverlapStrategy.GROUP, value: OverlapStrategy.GROUP }
                             ]}
                             clearable={false}
                             searchable={false}
-                            value={this.store.overlapStrategy}
+                            value={{ label: this.store.overlapStrategy, value: this.store.overlapStrategy}}
                         />
                     </div>
                 );
             }
         }
-    }
+    });
 
     render() {
         if (!this.store) {
@@ -243,7 +244,7 @@ export default class GroupComparisonPage extends React.Component<IGroupCompariso
                                 <GroupSelector
                                     store = {this.store}
                                 />
-                                {this.overlapStrategySelector}
+                                {this.overlapStrategySelector.component}
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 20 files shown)


src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -196,42 +197,42 @@ export default class GroupComparisonPage extends React.Component<IGroupCompariso
         } 
     });
 
+    @computed get overlapStrategySelector() {
+        if (this.store._selectionInfo.isComplete) {
```
alisman: should probably return null if not complete.  i'm surprised this passes ts.   shouldn't this be your MobxPromiseView?

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -196,42 +197,42 @@ export default class GroupComparisonPage extends React.Component<IGroupCompariso
...
+            if (this.store._selectionInfo.result.overlappingSamples.length === 0 &&
+            this.store._selectionInfo.result.overlappingPatients.length === 0) {
+                return null;
+            } else {
+                // (${caseCountsInParens(this.store._selectionInfo.result.overlappingSamples, this.store._selectionInfo.result.overlappingPatients)})
+                return (
+                    <div style={{width:333, zIndex:20}}>
+                        <ReactSelect
```
alisman: you might as well use new ReactSelect2 to avoid having to update later.  probably not a big deal to change it.

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -39,38 +39,44 @@ import {getPatientSurvivals} from "pages/resultsView/SurvivalStoreHelper";
...
+        return this._currentTabId;
+    }
+
     @autobind
     public setTabId(id:GroupComparisonTab) {
-        this.currentTabId = id;
+        this._currentTabId = id;
+    }
+
+    @computed get overlapStrategy() {
```
alisman: super minor but i actually think these don't have to be computed since they just return observable values and don't do any work

src/pages/groupComparison/GroupComparisonPage.tsx
adamabeshouse: oh you're right

src/pages/groupComparison/GroupComparisonPage.tsx
adamabeshouse: kk

src/pages/groupComparison/GroupComparisonStore.ts
adamabeshouse: good point

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -89,7 +95,7 @@ export default class GroupComparisonStore {
         }
     }
 
-    readonly unfilteredGroups = remoteData<ComparisonGroup[]>({
+    readonly _remoteGroups = remoteData<ComparisonGroup[]>({
```
alisman: why the underscore?

adamabeshouse: just to encourage that its not to be used, its just part of the computation of the groups. what should be used is availableGroups and activeGroups. let me know if you have a better way to do this, I couldn't figure one out.

src/pages/groupComparison/GroupComparisonUtils.ts
```diff
@@ -192,38 +194,43 @@ export function getNumSamples(
...
-                }),
-                patients: study.patients.filter(patientId=>{
-                    if (overlappingPatientsSet.has(studyId, patientId)) {
-                        hasOverlappingPatients = true;
-                        return false;
-                    } else {
-                        return true;
-                    }
-                }),
+            const samples = study.samples.filter(sampleId=>{
```
alisman: maybe filteredSamples? or something that indicates subset?

src/pages/groupComparison/GroupComparisonUtils.ts
```diff
@@ -192,38 +194,43 @@ export function getNumSamples(
...
+                if (overlappingSamplesSet.has(studyId, sampleId)) {
```
alisman: i was looking at documenation for has and can't find anything about multiple arguments.  what does this do?

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManagerUtils.ts
```diff
@@ -11,14 +11,23 @@ export function getSelectedGroups(
...
-        (sampleIdentifiers, studyId)=>({
+export function getStudiesAttr(
+    sampleIdentifiers:SampleIdentifier[],
+    patientIdentifiers?:PatientIdentifier[]
+) {
+    const samples = _.groupBy(sampleIdentifiers, id=>id.studyId);
+    let patients = patientIdentifiers ? _.groupBy(patientIdentifiers, id=>id.studyId) : {};
+    const studies = _.uniq(Object.keys(samples).concat(Object.keys(patients)));
+    return studies.map(studyId=>{
+        const ret:any = {
```
alisman: any

src/pages/groupComparison/GroupComparisonUtils.ts
adamabeshouse: this is a set class that I wrote, ListIndexedSet, that uses rest args as the key

src/shared/lib/ListIndexedMap.ts
```diff
@@ -61,4 +61,32 @@ export default class ListIndexedMap<R> {
         this.map[folderKey] = this.map[folderKey] || [];
         return this.map[folderKey];
     }
+}
+
+export class ListIndexedSet {
```
alisman: makes me nervous because it's library code without tests or documentation.  Also, i'm not sure what "ListIndexed" means.    If I understand correctly this keeping track of a set of keys?

adamabeshouse: yeah. I think it's not an ideal system. I think Im going to go in and refactor this so it accepts an object rather than an amorphous list of strings. ListIndexedMap, which it depends on, is thoroughly tested, I'll add tests for ListIndexedSet specifically

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -39,38 +39,44 @@ import {getPatientSurvivals} from "pages/resultsView/SurvivalStoreHelper";
...
+}
 
 export default class GroupComparisonStore {
 
-    @observable currentTabId:GroupComparisonTab|undefined = undefined;
-    @observable excludeOverlapping:boolean = false;
-    @observable sessionId:string;
+    @observable private _currentTabId:GroupComparisonTab|undefined = undefined;
+    @observable private _overlapStrategy:OverlapStrategy = OverlapStrategy.INCLUDE;
+    @observable private sessionId:string;
```
zhx828: Seems should be `_sessionId`?

adamabeshouse: since its just private and theres no setter, I feel like it makes sense to leave it like this

PR #2094, opened 2019-03-06
move download chart PDF function to the front-end

dippindots:
Create SVG to PDF using the library: [svg2pdf.js](https://github.com/yWorks/svg2pdf.js)
Fix https://github.com/cBioPortal/cbioportal/issues/5738.


src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -171,9 +171,10 @@ export default class Oncoprinter extends React.Component<IOncoprinterProps, {}>
             onClickDownload:(type:string)=>{
                 switch(type) {
                     case "pdf":
-                        if (!svgToPdfDownload("oncoprint.pdf", this.oncoprint.toSVG(true))) {
-                            alert("Oncoprint too big to download as PDF - please download as SVG.");
-                        }
+                        svgToPdfDownload("oncoprint.pdf", this.oncoprint.toSVG(false));
```
alisman: maybe wrap this in try catch and notify user if there's an error

src/shared/lib/svgToPdfDownload.ts
```diff
@@ -13,49 +12,65 @@ function base64ToArrayBuffer(base64:string) {
...
+    const width = svg.scrollWidth || parseInt((svg.attributes.getNamedItem('width') as Attr).nodeValue!), height = svg.scrollHeight || parseInt((svg.attributes.getNamedItem('height') as Attr).nodeValue!);
 
-    if (!req) {
-       return false;
-    }
-
-    req.end((err, res)=>{
-        if (!err && res.ok) {
-            fileDownload(base64ToArrayBuffer(res.text), filename);
+    // dealing with oncoprint svg (temporarily)
```
alisman: this is temporary b/c @adamabeshouse will fix in actual library?

src/shared/lib/svgToPdfDownload.ts
```diff
@@ -13,49 +12,65 @@ function base64ToArrayBuffer(base64:string) {
...
+        yOffset: 0,
+        scale: 1
+    });
+    // return the svg data, we don't need the header
+    return pdf.output('dataurlstring').split(',').length >= 2 ? pdf.output('dataurlstring').split(',')[1] : "";
 }
 
-export async function svgToPdfPromise(svg:Element, servletUrl?: string) {
-    const res = await svgToPdfRequest(svg, servletUrl);
+export async function svgToPdfPromise(svg:Element) {
```
alisman: can we get rid of this?

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
dippindots: Sure I can do that. But should we have a limitation on the file size of SVG? For a very large oncoprint SVG, it may take half a minute for downloading.

src/shared/lib/svgToPdfDownload.ts
dippindots: Yes, correct. when we have the new oncoprint.js, these lines can be removed from here.

src/shared/lib/svgToPdfDownload.ts
dippindots: It is required by the charts in study view page, may not easy to change.

tmazor:
So far this looks good except when I downloaded the enrichments volcano plot as a PDF something went very wrong with the text....
image

tmazor:
A few issues on study view:

The mutation vs CNA plot doesn't have a PDF download option.

The "<=" sign doesn't come through correctly:
image

Superscript gets lost - these numbers are 10, 10^2, 10^3 and 10^4:
image

dippindots:
@tmazor Thank you so much for helping to find these issues. The issues have been solved by changing to a custom font(FreeSerif), this font can handle most of the special characters. Could you help me to review my pr again? thanks in advance! Here is the link: https://cbioportal-frontend-pr-2094.herokuapp.com/

tmazor:
Thanks @dippindots !

I don't really love the fact that we now have a mix of fonts in some of the PDF downloads (I'm having trouble with uploading images, but take a look at the PDF download of the plot on the co-expression tab and note how the title is a different font than everything else), and the predominant font in the PDF downloads is now a serif font when the website (and png/svg downloads) all use a sans serif font.

I realize this is a consequence of using FreeSerif to fix the problems with the special characters. Is there a sans serif option that we could use instead?

This isn't a deal breaker for me - so if freeSerif if the best option to get PDF downloads working again, I can live with this. But I'd be happier with more consistency in the fonts :)

src/pages/studyView/chartHeader/ChartHeader.tsx
```diff
@@ -78,8 +79,10 @@ export class ChartHeader extends React.Component<IChartHeaderProps, {}> {
                                         const fileName = `${this.fileName}.${props.type.substring(0, 3).toLowerCase()}`;
                                         if (props.type === "PNG") {
                                             saveSvgAsPng(data, fileName, {backgroundColor:"#ffffff"});
-                                        } else if ((props.type === "PDF" || typeof data === "string") && data.length > 0) {
+                                        } else if (typeof data === "string" && data.length > 0) {
```
alisman: what type is this?

dippindots: @alisman It is for SVG type, SVG will be serialized to string.

alisman:
There are three failing e2e tests that fail with timeout.  They are not connected with this PR.

PR #2097, opened 2019-03-06
Modularize OncoKB related components and styles

zhx828:
- Move most of the oncokb global styles to modules
- Remove not used styles
- Separate out AbstractItem, PmidItem, ReferenceList, OncoKbTreatmentTable, SummaryWithRefs, RefComponent to components and reuse
- Fix issues https://github.com/cBioPortal/cbioportal/issues/5807 https://github.com/cbioportal/cbioportal/issues/5724


src/globalStyles/global.scss
```diff
@@ -27,11 +27,11 @@ button:active, button:focus, div:focus, div:active {
 
 a {
   cursor:pointer;
-  text-decoration:none !important;
+  text-decoration:none;
```
zhx828: @alisman not sure why these styles are important. Can we remove?

zhx828:
@alisman @onursumer Sorry about the big PR. The merge commit seems ruined history of the few files. Now in history, for instance OncoKbCard, it shows the file was removed and added. So you cannot see what actually modified. Here are some info about these files. 

- AbstractItem, PmidItem, ReferenceList, SummaryWithRefs, RefComponent are all coming from OncoKbCard. I didn't change much of the logic, just straight moving for components sharing. I did some clean up to these components.
- OncoKbTreatmentTable is new, you probably want to look at it.
- In OncoKB, the main thing is how we calculate the background image position. We used to predefined all classnames with x,y predefined. But with the latest oncokb image updated, that would be too much to add. Since the logic of the positions are predictable, so I put them into util functions.
- Most of the global styles are removed except the one for the tooltip

src/globalStyles/global.scss
alisman: this is legitimate.  the style of our links is no underline until hover.   how come you needed to remote !important?  it may not be necessary now that we don't have legacy styles to compete with. but just curious.   we would need to take a look through site to see that things are still consistent.

src/shared/components/annotation/oncokb/AbstractItem.tsx
```diff
@@ -0,0 +1,19 @@
+import * as React from 'react';
+import styles from '../styles/oncokb/listGroupItem.module.scss';
+import {ArticleAbstract} from "../../../api/generated/OncoKbAPI";
+
+export default class AbstractItem extends React.Component<ArticleAbstract> {
```
alisman: I think AbstractItem is, to put it mildly, too vague.  We shouldn't have to look at folder path to gain specifity.

src/globalStyles/global.scss
zhx828: @alisman because `!important` is overwriting the component style when I don't want to have `text-decoration` for the component when hover. I understand you want to standardize the styles, but shouldn't it work already without the `!important`? I cannot overwrite it since it's a global css style on the particular html element.

src/shared/components/annotation/oncokb/AbstractItem.tsx
zhx828: Would be clarify the purpose of the component after renaming to `ArticleAbstractItem`? I think the only confusing part is the component using a type from OncoKB?

src/globalStyles/global.scss
alisman: i'm open to taking !important off if it turns out it's not important anymore.  but you should be able to override locally.   you just need to use a more specific selector.  
".someClass a"   will win over "a"  if both have !important

src/shared/components/annotation/styles/oncokb/collapsible.module.scss
```diff
@@ -0,0 +1,18 @@
+@import "vars.module";
+
+$collapsible-height: 30px;
+$collapsible-header-color: white;
+
+.collapsible-header {
```
alisman: when these are applied, i would think that the hyphen in style.collapsible-height would cause javascript to error, no?

src/shared/components/annotation/oncokb/AbstractItem.tsx
alisman: Ah, I see, it's "astract" in a sense of a "scientific abstract".    Yeah, I think ArticleAbstractItem would be better.

src/shared/components/annotation/styles/oncokb/collapsible.module.scss
zhx828: Yeah, I need to use style['collapsible-height'] which was ok because the IDE autogenerated. But I would open to change to style.collapsibleHeight if this is how we should do it and I love to keep the consistency.

src/shared/components/annotation/oncokb/ReferenceList.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+        if (this.props.pmidData) {
+            this.props.pmids.forEach((uid: number) => {
+                const cacheData = this.props.pmidData![uid.toString()];
+                const articleContent = cacheData ? cacheData.data : null;
+
+                if (articleContent) {
+                    list.push(
+                        <PmidItem title={articleContent.title}
+                                  author={(_.isArray(articleContent.authors) && articleContent.authors.length > 0) ? (articleContent.authors[0].name + ' et al.') : 'Unknown'}
```
onursumer:
```suggestion
                                  author={
                                      _.isArray(articleContent.authors) && articleContent.authors.length > 0 ?
                                          articleContent.authors[0].name + ' et al.' : 'Unknown'
                                  }
```

src/shared/components/annotation/oncokb/RefComponent.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import {getNCBIlink} from "../../../api/urls";
+import mainStyles from "../styles/oncokb/main.module.scss";
+import ReferenceList from "./ReferenceList";
+import DefaultTooltip from "../../defaultTooltip/DefaultTooltip";
+import {ICache} from "../../../lib/SimpleCache";
+
+export default class RefComponent extends React.Component<{ str: string, componentType: 'tooltip' | 'linkout', pmidData:ICache<any>}> {
```
onursumer: `str` doesn't sound very informative as a property name, can we call this something else?

src/shared/components/annotation/oncokb/SummaryWithRefs.tsx
```diff
@@ -0,0 +1,34 @@
+import * as React from 'react';
+import RefComponent from "./RefComponent";
+import {ICache} from "../../../lib/SimpleCache";
+
+export default class SummaryWithRefs extends React.Component<{ str: string | undefined, type: 'tooltip' | 'linkout', pmidData: ICache<any> }> {
```
onursumer: same as previous `str`, can we rename?

src/shared/components/annotation/oncokb/RefComponent.tsx
zhx828: Changed to `content`

src/shared/components/annotation/oncokb/SummaryWithRefs.tsx
zhx828: Changed to `content`

src/globalStyles/global.scss
zhx828: I'm not so confident removing it without testing all components. Gonna revert it.

zhx828:
@alisman @onursumer this pr is updated.

src/shared/lib/OncoKbUtils.ts
```diff
@@ -184,8 +185,52 @@ export function normalizeLevel(level:string):string|null
...
+export function oncogenicYPosition(oncogenicity: string, isVUS: boolean, resistanceLevel: string | null) {
+    const oncogenicityMap: { [id: string]: number } = {
+        'Oncogenic': 0,
+        'Predicted Oncogenic': 0,
+        'Likely Oncogenic': 0,
+        'Neutral': 1,
+        'Likely Neutral': 1,
+        'Unknown': 2,
+        'Inconclusive': 2,
+        '': 4
```
alisman: wow, empty string can be used as key?

src/shared/lib/OncoKbUtils.ts
```diff
@@ -184,8 +185,52 @@ export function normalizeLevel(level:string):string|null
...
+
+    let oncogenicityIndex = oncogenicityMap[oncogenicity];
+    if (oncogenicityIndex === undefined) {
+        oncogenicityIndex = 4;
+    }
+    if (oncogenicityIndex > 1 && isVUS) {
+        oncogenicityIndex = 3;
+    }
+
+    let resistanceLevelIndex = resistanceLevel === null ? 0 : (resistanceLevelMap[resistanceLevel] || 0);
```
alisman: seems like maybe you don't need that '' : 0 in the resistanceLevelMap because it's a default.   so, i think all you need is just `resistanceLevelMap[resistanceLevel] || 0`

src/shared/lib/OncoKbUtils.ts
```diff
@@ -184,8 +185,52 @@ export function normalizeLevel(level:string):string|null
...
+    };
+    const resistanceLevelMap: { [id: string]: number } = {
+        '': 0,
+        'R1': 1,
+        'R2': 2,
+        'R3': 3
+    };
+
+    let oncogenicityIndex = oncogenicityMap[oncogenicity];
+    if (oncogenicityIndex === undefined) {
```
alisman: just as below, i don't think you need '' : 0

src/shared/components/annotation/styles/oncokb/vars.module.scss
```diff
@@ -0,0 +1,6 @@
+$darker-blue: #064785;
```
alisman: i think these variables will become global, so you probably want to, 1. give them domain specific names, 2. maybe place them in the global variables file

src/shared/lib/OncoKbUtils.ts
zhx828: Yeah, it was designed that way, but latter on forgot to do some clean up. Will do as suggested below. Thanks.

src/shared/lib/OncoKbUtils.ts
zhx828: Still need to do a null check though

zhx828:
@alisman this pr is updated.

src/globalStyles/global.scss
```diff
@@ -31,7 +31,7 @@ a {
 }
 
 a:hover {
-  text-decoration:underline !important;
+  text-decoration:underline !important;;
```
alisman: double ;

alisman:
@zhx828 i think we should ask for some product testing to make sure everything still works properly

zhx828:
@alisman I will ask some feedbacks from OncoKB team since this is only has impact on OncoKB Card.

zhx828:
@alisman got approval from OncoKB team. Gonna merge the PR when I figure out the e2e.

zhx828:
The only test failing is not related to this PR.

PR #2098, opened 2019-03-08
oncoprinter - fix and enhance system of exposing data error to user

adamabeshouse:
prompted by this issue: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/cbioportal/BPaRVsFxsyQ/of0lyUXPCQAJ

image


src/pages/staticPages/tools/oncoprinter/OncoprinterStore.ts
```diff
@@ -121,6 +122,7 @@ export default class OncoprinterStore {
             // parse input
             const parsed = parseInput(this._dataInput);
             if (parsed.status === "error") {
+                await sleep(0); // this gives mobxpromise time to register a catch handler so that the error doesnt throw uncaught promise rejection in the console
```
alisman: i think this gets into the semantics of the remoteData again.  error === http error, not a parsing error.    i understand the temptation to do this but i think it should be handled in a different way.   like, return error object as result.  or just do the parsing outside of the remote data.     lets discuss.

adamabeshouse: ok, that makes sense

PR #2101, opened 2019-03-08
Update how study view sorts the clinical attributes

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5815  
https://github.com/cBioPortal/cbioportal/issues/3800


src/config/serverConfigDefaults.ts
```diff
@@ -90,21 +90,40 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
         tableAttrs: ['SAMPLE_CANCER_TYPE', 'SAMPLE_CANCER_TYPE_DETAILED'],
         priority: {
             "SAMPLE_CANCER_TYPE": 3000,
+            "PATIENT_CANCER_TYPE": 3000,
```
zhx828: Please ignore these duplicate default settings. These will be cleaned up once the backend does not require to define which group(sample/patient) the clinical attribute in. For now, we need to define them separately due the the same attribute could be defined as patient/sample differently on study level

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1322,12 +1317,31 @@ export function clinicalDataCountComparator(a: ClinicalDataCount, b: ClinicalDat
...
+        displayName: a.displayName
+    }, {
+        priority: Number(b.priority),
+        displayName: b.displayName
+    });
+}
+
+// Descent sort priority then ascent sort by display name
+export function chartMetaComparator(a: ChartMeta, b: ChartMeta): number {
+    return clinicalAttributeSortingComparator(_.pick(a, ['priority', 'displayName']), _.pick(b, ['priority', 'displayName']))
```
alisman: not typesafe b/c props passed as string.   porbably just just do   ` {  priority:a.priority, displayName:a.displayName  } `
but actually, why not just pass the whole object?  why do you need to get rid of the other things?

zhx828: The function is used in two places but the models are different. They just have common properties priority and displayName which are only required component.

zhx828:
@alisman this pr is updated

PR #2109, opened 2019-03-12
Filtering MutatedGenes/CNA Genes tables with cancer gene list

zhx828:
This fixes https://github.com/cBioPortal/cbioportal/issues/5716 cBioPortal/cbioportal#5504


src/shared/components/annotation/images/oncogenic-black.svg
```diff
@@ -0,0 +1,11 @@
+<?xml version="1.0" encoding="utf-8"?>
```
zhx828: @alisman this is a copy of the oncogenic-only.svg except changing the color. This is not the best implementation. The optional is making the icon as font and included?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2499,14 +2554,23 @@ export class StudyViewPageStore {
...
             if (!_.isEmpty(this.mutationProfiles.result!)) {
                 // TODO: get data for all profiles
-                return internalClient.fetchMutatedGenesUsingPOST({
+                let mutatedGenes = await internalClient.fetchMutatedGenesUsingPOST({
                     studyViewFilter: this.filters
                 });
+                return mutatedGenes.map(item => {
+                    return {
+                        ...item,
+                        oncokbAnnotated: this.oncokbAnnotatedGeneEntrezGeneIds.includes(item.entrezGeneId),
```
alisman: @adamabeshouse @zhx828 i know i always harp on this and it hasn't seemed to matter much, but it always bothers me when i see a computed which is just a filtering of mobxpromise.  this is a good example because this promise awaits oncokbCancerGenes, but then uses oncokbAnnotatedGeneIds which is a computation of the former.   It would be more straightforward if oncokbTsgEntrezGeneIds was just a utility function that accepted oncokbGenes.   But I know people like to encapsulate stuff.   I wonder if there is a way for mobxpromises to expose methods that would their results.   Just food for thought.

alisman: I guess you do get a benefit @computed being cached. So it will only compute once per oncokbgenes result.  That's a pretty big win.  But I think the same could be accomplished with @memoized decorator on a utility function.

src/pages/studyView/oncokb/OncokbIconLinkImg.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    @computed
+    get geneLinkout() {
+        return this.props.oncokbAnnotated ? `https://oncokb.org/gene/${this.props.hugoSymbol}` : 'https://oncokb.org/cancerGenes';
+    }
+
+    getOncokbIcon() {
+        return this.props.isCancerGene ? <a className={styles.cancerGenesRow} href={this.geneLinkout}
+                                            target="_blank">{getOncoKbImage()}</a> : <span></span>;
```
alisman: null instead of span?

src/pages/studyView/oncokb/OncokbIconLinkImg.tsx
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    getGeneTypeText() {
+        let content = [];
+        if (this.props.oncokbOcg) {
+            content.push(<span style={{color: STUDY_VIEW_CONFIG.colors.deletion, marginLeft:2, fontWeight: 'bold'}}>O</span>);
```
alisman: if this is something repeated a number of times, we should avoid using inline css, at least for margin and fontWeight.    Actually fontWeight you could accomplish by using <strong> instead of <span>.  Color is ok.

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: > I wonder if there is a way for mobxpromises to expose methods that would their results.
 
would what their results?

adamabeshouse: also I agree that those should be mobxpromise

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2274,6 +2286,49 @@ export class StudyViewPageStore {
...
+            return oncoKBClient.genesGetUsingGET({});
+        },
+        onError: (error => {
+            console.log(error);
+        }),
+        default: []
+    });
+
+    @computed
+    get oncokbCancerGeneEntrezGeneIds() {
```
adamabeshouse: As @alisman mentioned, we've run into issues with this a few times, and the conclusion we've come to is that,  in any case where a @computed can possibly not yet be ready (e.g. oncokbAnnotatedGeneEntrezGeneIds, since below you're waiting for oncokbCancerGenes before using oncokbAnnotatedGeneEntrezGeneIds), that it should just be made into a mobxpromise so it can be properly awaited and so that it models the true nature of that data as being unready then being ready.

src/pages/studyView/StudyViewPageStore.ts
alisman: filter their results.   so, instead of always wrapping mobxpromise with mobxpromise that only filters/manipulates result, the original promise could somehow have methods that filter the result property.     like,  
`  if (mpromise.isComplete) {
      console.log(mpromise.filteredResult)  

}`
where filteredResult is a function defined in mobxpromise declaration that operates on result).

src/shared/components/annotation/images/oncogenic-black.svg
alisman: I think you can do this with css.  Support is not perfect, but good enough I think.   I tried it out and it works.  The following makes image black.

filter: brightness(0.1)

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: I dont like that promise X needs to anticipate how it would be used.

Maybe another solution would be to add a nice factory method for mobxpromises like this:

```
readonly Y = filterMobxPromise(X, result=>{ // return filtered version of result; });

function filterMobxPromise<R,F>(promise:MobxPromise<R>, filter:(result:R)=>F) {
    return remoteData({
        await:()=>[promise],
        invoke:()=>Promise.resolve(filter(promise.result!))
    });
}

```

zhx828:
@khzhu  we also need to discuss how enable the cancer gene filtering for fusion table depends on which pr will be merged first.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -281,6 +279,17 @@ export type StatusMessage = {
     message: string
 };
 
+type oncokbCancerGene = {
+    oncokbAnnotated: boolean;
+    oncokbOcg: boolean;
```
adamabeshouse: what do these two abbrevs mean?

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -55,36 +58,34 @@ enum ColumnKey {
...
-        [ColumnKey.CNA]: 0,
-        [ColumnKey.NUMBER]: 0,
-        [ColumnKey.FREQ]: 0,
-    };
-
-    private disposers:IReactionDisposer[] = [];
+    @observable private isFilteredByCancerGeneList = true;
 
     constructor(props: ICNAGenesTablePros) {
         super(props);
```
adamabeshouse: constructor no longer necessary

src/pages/studyView/StudyViewPageStore.ts
khzhu: @zhx828 , I see, thanks for loop me in.

zhx828: @adamabeshouse this is oncogene. I will spell the full word to solve the confusion.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -212,6 +214,17 @@ export type StatusMessage = {
     message: string
 };
 
+type oncokbCancerGene = {
```
alisman: Initial cap?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2023,6 +2036,72 @@ export class StudyViewPageStore {
         }
     }, []);
 
+    readonly oncokbCancerGenes = remoteData<CancerGene[]>({
+        await: () => [],
+        invoke: async () => {
+            return oncoKBClient.utilsCancerGeneListGetUsingGET({});
+        },
+        onError: (error => {
+            console.log(error);
```
alisman: console.log.    If we specify onError, we disable the global handling of this error (error screen) so we should make sure that we handle it properly in code.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2023,6 +2036,72 @@ export class StudyViewPageStore {
...
+        }),
+        default: []
+    });
+
+    readonly oncokbCancerGeneEntrezGeneIds = remoteData<number[]>({
+        await: () => [this.oncokbCancerGenes],
+        invoke: async () => {
+            return this.oncokbCancerGenes.result.map(gene => Number(gene.entrezGeneId));
```
alisman: is there a reason why we have to convert them to number?  seems like they are typed elsewhere as strings

src/pages/studyView/TableUtils.tsx
```diff
@@ -0,0 +1,42 @@
+import * as React from 'react';
+import {getOncoKBCancerGeneListLinkout, getOncoKBReferenceInfo} from "./oncokb/OncoKBUtils";
+import styles from "./table/tables.module.scss";
+import classnames from 'classnames';
+import DefaultTooltip from "../../shared/components/defaultTooltip/DefaultTooltip";
+import {If} from 'react-if';
+
+export function getGeneColumnHeaderReder(cellMargin: number, headerName: string, isFilteredByCancerGeneList: boolean, cancerGeneIconToggle: (event: any) => void) {
```
alisman: type: Render?

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -164,88 +139,49 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
...
-                                            src={require("./images/gistic.png")}
-                                            className={styles.gistic}
-                                        />
-                                    </span>
-                                </DefaultTooltip>
-                            </If>
-                        </div>
-                    );
+                headerRender: () => {
+                    return getGeneColumnHeaderReder(this.cellMargin[ColumnKey.GENE], ColumnKey.GENE, this.isFilteredByCancerGeneList, this.toggleCancerGeneFilter);
```
alisman: typo.  guess it's the same as above

src/pages/studyView/table/MutatedGenesTable.tsx
```diff
@@ -464,18 +392,20 @@ export class MutatedGenesTable extends React.Component<IMutatedGenesTablePros, {
     public render() {
         return (
             <div>
```
alisman: unless you need a wrapped <div>, you can now use a fragment <>  </>

src/pages/studyView/table/MutatedGenesTable.tsx
```diff
@@ -464,18 +392,20 @@ export class MutatedGenesTable extends React.Component<IMutatedGenesTablePros, {
...
-                    data={this.props.promise.result || []}
-                    columns={this.tableColumns}
-                    showSelectSamples={true && this.preSelectedRows.length > 0}
-                    isSelectedRow={this.isSelectedRow}
-                    afterSelectingRows={this.afterSelectingRows}
-                    sortBy={this.sortBy}
-                    sortDirection={this.sortDirection}
-                    afterSorting={this.afterSorting}
-                />
+                {this.props.promise.isComplete && (
```
alisman: maybe put the modal inside if statement too?

zhx828: We don't need to rerender the model when promise changes.

zhx828:
[Two failing tests](https://circleci.com/gh/cBioPortal/cbioportal-frontend/40159?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link) are not related to this pr.

PR #2110, opened 2019-03-12
fix ellipsis tooltip on add chart button

dippindots:
fix ellipsis tooltip on add chart button
Fix https://github.com/cBioPortal/cbioportal/issues/5505.


alisman:
@dippindots make sure to test this in Firefox and Safari

dippindots:
@alisman Thanks for the reminder, just verified this fix works for Firefox, Safari, and Chrome.

dippindots:
@zhx828 I haven't found material to support this, but seems like the content of <label> tag doesn't support the scrollWidth property, it will always be 0.

PR #2111, opened 2019-03-12
Make the lollipop plot legend use same font

agarwalrounak:
Fix [cBioPortal/cbioportal#3767](https://github.com/cBioPortal/cbioportal/issues/3767)


agarwalrounak:
Sorry, I missed that. Now done!

alisman:
@agarwalrounak the failing test is unrelated to your PR.  We have some problems with e2e test stability right now so as long as I be sure that the changes in a PR have not caused a failure, I can merge it.

agarwalrounak:
Thanks for the review @alisman! Removing `fontWeight` would make them the same as other text in the legend. Do you want that?
Also, I didn't understand what is meant by `convert the span tags to tags instead`.

onursumer:
@agarwalrounak `<b>` tags :)

alisman:
Actually, I meant `<strong>`  tag.  Pretty much the same as `<b>`     
(I think the tag got stripped out by the editor.)
Using strong tag will automatically apply the same bolding.   When possible we want to avoid inline css and use semtantic html.

PR #2113, opened 2019-03-12
Disable gene validation when passing through inputGeneQuery

zhx828:
- Disable button when validation error or content is empty
- Disable gene validation when passing through inputGeneQuery
- Make GeneSymbolValidatorMessage function component using React.memo
- Upgrade enzyme-adapter-react-16 to 1.12.1. Only in version 1.10.0 and above enzyme-adapter-react-16 supports React.memo

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5834


alisman:
@zhx828 there still some strange architecture here.   for example, the onChange handler sets both the geneQuery and the queryToBeValidated property, which begs the question, why would we set geneQuery with something we do not yet know to be valid?    Would seem the the rest of the app then has to keep track of validation state, whereas if you defer the setting of queryGene to when you know the the query is valid, you're better off.

src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
```diff
@@ -1,46 +1,102 @@
...
 export default class GeneSymbolValidator extends React.Component<IGeneSymbolValidatorProps, {}> {
+	@memoize
+	private async getGeneSuggestions(alias: string): Promise<GeneReplacement> {
+		return {
+			alias,
+			genes: await client.getAllGenesUsingGET({ alias })
+		};
+	}
+
+	readonly genes = remoteData({
```
zhx828: Line 39-91 are from original code.

src/shared/components/GeneSelectionBox/GeneSelectionBox.tsx
```diff
@@ -28,78 +25,82 @@ export interface IGeneSelectionBoxProps {
...
-            () => this.props.inputGeneQuery,
-            inputGeneQuery => {
-                if ((inputGeneQuery || '').toUpperCase() !== this.geneQuery.toUpperCase())
-                    this.geneQuery = (inputGeneQuery || '').trim();
-            },
-            { fireImmediately: true }
-        );
+        this.textAreaRef = React.createRef<HTMLTextAreaElement>();
+
+        this.reactions = [
```
alisman: i would call it disposers

src/shared/components/GeneSelectionBox/GeneSelectionBox.tsx
```diff
@@ -28,78 +25,82 @@ export interface IGeneSelectionBoxProps {
...
+                if (newOql.error !== undefined) {
+                    this.textAreaRef.current!.setSelectionRange(newOql.error.start, newOql.error.end);
+                }
+            }, {fireImmediately: true})
+        ];
     }
 
     componentWillUnmount(): void {
-        this.geneStatusReaction();
+        for (const reaction of this.reactions) {
```
alisman: disposer of this.disposers

src/shared/components/GeneSelectionBox/GeneSelectionBox.tsx
```diff
@@ -28,78 +25,82 @@ export interface IGeneSelectionBoxProps {
...
-    componentDidMount() {
-        reaction(
-            () => this.props.inputGeneQuery,
-            inputGeneQuery => {
-                if ((inputGeneQuery || '').toUpperCase() !== this.geneQuery.toUpperCase())
-                    this.geneQuery = (inputGeneQuery || '').trim();
-            },
-            { fireImmediately: true }
-        );
+        this.textAreaRef = React.createRef<HTMLTextAreaElement>();
```
alisman: cool!  createRef.  where did you learn about this?

src/shared/components/GeneSelectionBox/GeneSelectionBox.tsx
```diff
@@ -28,78 +25,82 @@ export interface IGeneSelectionBoxProps {
...
+    
     @observable private geneQuery = '';
-    @observable private geneQueryErrorDisplayStatus: DisplayStatus = DisplayStatus.UNFOCUSED;
+    @observable private geneQueryIsValid = true;
+    @observable private queryToBeValidate = '';
     @observable private isFocused = false;
 
+    private readonly textAreaRef:React.RefObject<HTMLTextAreaElement>;
+    private updateQueryToBeValidateDebounce = _.debounce(()=>{
+        this.queryToBeValidate = this.currentTextareaValue;
```
alisman: should be "queryToBeValidated" (missing d at end)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -568,7 +568,8 @@ export class StudyViewPageStore {
     @autobind
     @action onCheckGene(hugoGeneSymbol: string) {
         //only update geneQueryStr whenever a table gene is clicked.
-        this.geneQueryStr = updateGeneQuery(this.geneQueries, hugoGeneSymbol);
+        this.geneQueries = updateGeneQuery(this.geneQueries, hugoGeneSymbol);
+        this.geneQueryStr = this.geneQueries.map(query=>unparseOQLQueryLine(query)).join(' ');
```
alisman: seems like geneQueryStr is just a computation of geneQueries.  if true, then i think it could just be a @computed?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -577,9 +578,9 @@ export class StudyViewPageStore {
     }
 
     @autobind
-    @action updateSelectedGenes(query: SingleGeneQuery[], genesInQuery: Gene[]) {
+    @action updateSelectedGenes(query: SingleGeneQuery[]) {
         this.geneQueries = query;
-        this.queriedGeneSet = new ObservableMap(stringListToSet(genesInQuery.map(gene => gene.hugoGeneSymbol)))
+        this.queriedGeneSet = new ObservableMap(stringListToSet(query.map(gene => gene.gene)))
```
alisman: again this looks like computation of geneQueries.   by keeping them separate, we have now three sources of truth.

src/pages/studyView/studyPageHeader/rightPanel/RightPanel.tsx
```diff
@@ -69,12 +68,9 @@ export default class RightPanel extends React.Component<IRightPanelProps, {}> {
...
                                     suggestions: GeneReplacement[];
                                 },
-                                queryStr: string,
-                                status: "pending" | "error" | "complete") {
-        this._isQueryButtonDisabled = queryStr==='' || (status === 'pending') || !_.isUndefined(oql.error) || genes.suggestions.length !== 0;
-        if (status === "complete") {
-            this.props.store.updateSelectedGenes(oql.query, genes.found);
-        }
+                                queryStr: string) {
+        this.geneValidationHasIssue = queryStr==='' || !_.isUndefined(oql.error) || genes.suggestions.length !== 0;
```
alisman: possibly not, but could this (geneValidationHasIssue) be computed?  the reason it catches my eye is that it's a state that has to be cleared, and yet depends on other observables.

zhx828: It actually only depends on the params, not observables.

src/pages/studyView/StudyViewPageStore.ts
zhx828: I would just leave geneQueryStr as an observer. It's an intermediate value to deal with the geneQuery string from the GeneSelctionBox  and gene selection from the MtuatedGenes/CNA Genes table.

src/pages/studyView/StudyViewPageStore.ts
zhx828: Thanks for pointing it out. I actually able to get rid of `queriedGeneSet`.

zhx828:
Tested two failed e2e. Not related to PR and works good locally,

zhx828:
@alisman six e2e failing tests are related to oncoprint but seems like actual failing due to data change?  But I don't think they are related to my pr. Merging.

PR #2116, opened 2019-03-13
Show bar/pie charts when there is only one sample in the study

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5845

### DIFF
--- a/src/pages/studyView/StudyViewUtils.spec.ts
+++ b/src/pages/studyView/StudyViewUtils.spec.ts
@@ -2,8 +2,9 @@ import {assert} from 'chai';
 import {
     calcIntervalBinValues,
     calculateLayout,
-    clinicalDataCountComparator,
+    calculateNewLayoutForFocusedChart,
     chartMetaComparator,
+    clinicalDataCountComparator,
     filterCategoryBins,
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -546,6 +546,10 @@ export function isLogScaleByValues(values: number[]) {
     );
 }
 
+export function theChartShouldBeShown(filer: Partial<StudyViewFilterWithSampleIdentifierFilters>, uniqueDataSize: number, sizeOfAllSamples: number) {
```
alisman: i would rename this shouldShowChart()

src/pages/studyView/StudyViewUtils.spec.ts
```diff
@@ -171,6 +179,31 @@ describe('StudyViewUtils', () => {
         });
     });
 
+    describe('theChartShouldBeShown', () => {
```
alisman: nice!

zhx828:
Tested the two failed e2e, they are showing correctly locally.

PR #2119, opened 2019-03-14
distinguish singular and plural

dippindots:
distinguish singular and plural
Fix https://github.com/cBioPortal/cbioportal/issues/5825.

### DIFF
--- a/src/shared/components/query/quickSearch/QuickSearch.tsx
+++ b/src/shared/components/query/quickSearch/QuickSearch.tsx
@@ -11,6 +11,7 @@ import { Label } from 'react-bootstrap';
 import * as moduleStyles from "./styles.module.scss";
 import {action, computed, observable, runInAction} from "mobx";
 import {remoteData} from "../../../api/remoteData";
+import Pluralize from 'pluralize';
 
 export const SHOW_MORE_SIZE: number = 20;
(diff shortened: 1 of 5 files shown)


dippindots:
@alisman I proposed to introduce a library to our project since we have lots of cases to distinguish singular and plural. Do we need a library like this?

src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -257,7 +258,8 @@ const Group = (props:any) => {
 
     const groupData:any = props.data.groupData;
 
-    const label = groupData.value + " more " + groupData.type + " (click to load " +
+    const label = groupData.value + " more " + 
+        (groupData.value !== 1 ? groupData.type : Pluralize.singular(groupData.type)) + " (click to load " +
```
alisman: i think the Pluralize library would actually take care of conditionality too.   So, you could do Pluralize(groupData.type, groupData.value) and it would output the correct thing.

alisman: see usage documentation https://github.com/blakeembrey/pluralize

dippindots: Oh yes! Thanks for reminding! I have updated this pr.

PR #2120, opened 2019-03-14
Add quicksearch config + fix public session id + move ln? to frontend

inodb:
1. Add quicksearch config:

See backend changes and docs:

https://github.com/cBioPortal/cbioportal/pull/5863

Changes quick search to pull virtual study iff a session id is provided in the serverConfig. This allows us to display the study title in the gene section of the search dropdown. If that is not defined and people don't use session service, they can use the cancer study list config instead, a title can be set as well in that case, but it won't show up in the results view. So it's a crappier version, but at least it would allow people w/o session service to enable quick search

2. Fix public session id on results view

This allows using a session id from another user that only contains public studies on the results view. Before this would give an error because we determined whether or not a virtual study existed using the list of virtual studies of the logged in user

3. Move ln? links to frontend

This used to be handled by backend. Now handle this on frontend instead. The ln? links are described in the "Linking to Us" section of the Web API docs:

https://www.cbioportal.org/webAPI


src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -83,76 +86,122 @@ export default class QuickSearch extends React.Component {
...
+                    query: AppConfig.serverConfig.skin_quick_search_gene_query_cancer_study_list,
+                    name: AppConfig.serverConfig.skin_quick_search_gene_query_cancer_study_list_name,
+                });
+            }
+        }
+    });
 
-                let options = [];
+    // tslint:disable-next-line:member-ordering
+    private options = remoteData<any[]>({
```
inodb: this looks like a huge change, but all it does is add an await for geneStudyQuery, rest is indentation

src/shared/components/query/quickSearch/QuickSearchOption.tsx
```diff
@@ -1,86 +0,0 @@
-import * as React from "react";
```
inodb: doesn't seem like this file is used anymore. Aaron prolly replaced this

src/config/serverConfigDefaults.ts
```diff
@@ -52,6 +52,18 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_show_web_api_tab:true,
     skin_show_tweet_button:false,
     skin_show_tissue_image_tab:true,
+    quick_search_show: false,
+    quick_search_gene_query_cancer_study_list:
+        "laml_tcga_pan_can_atlas_2018,acc_tcga_pan_can_atlas_2018,blca_tcga_pan_can_atlas_2018," +
```
alisman: fyi you can now use `  ` template string do multiline strings in javascript instead of annoying concats

src/pages/home/HomePage.tsx
```diff
@@ -80,7 +80,7 @@ export default class HomePage extends React.Component<IResultsViewPageProps, {}>
             <PageLayout className="homePageLayout" noMargin={true} rightBar={<RightBar queryStore={this.queryStore} />}>
 
                 <div className={"skinBlurb"} dangerouslySetInnerHTML={{__html:AppConfig.serverConfig.skin_blurb!}}></div>
-                <QueryAndDownloadTabs getQueryStore={this.getQueryStore}/>
+                <QueryAndDownloadTabs getQueryStore={this.getQueryStore} showQuickSearchTab={AppConfig.serverConfig.quick_search_show} />
```
alisman: quick_search_enabled?

src/shared/lib/redirectHelpers.ts
```diff
@@ -66,6 +67,27 @@ export function handleCaseDO(){
...
+        gene_list: currentQuery.q.replace(":MUT","").replace("+",","),
+        cancer_study_list:
+            currentQuery.cancer_study_id ||
+            // use same set of studies as quick search gene query if no
+            // specific study is supplied
+            AppConfig.serverConfig.skin_quick_search_gene_query_session_id ||
+            AppConfig.serverConfig.skin_quick_search_gene_query_cancer_study_list
+    };
+
+    (getBrowserWindow().routingStore as ExtendedRouterStore).updateRoute(data, "/results/mutations");
```
alisman: we should probably use clear=true and replace=true as last two params in this.   this will prevent other params from getting persisted and also prevent the original ln from getting added to history.

src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -84,76 +85,93 @@ export default class QuickSearch extends React.Component {
...
+        await: () => [
+            this.geneStudyQuery
+        ],
+        invoke: () => {
+
+            const input = this.inputValue;
+            const geneStudyQuery = this.geneStudyQuery.result;
+
+            if (input.length > 0) {
```
alisman: @dippindots i know you didn't write this code but we should clean this up a little.   very hard to read flow.  we should put all the get calls that in a separately declared array and then use await on Promise.All and then divide up the rest of this routine into utility functions. lets discuss.  we can do this in a follow up PR though.

alisman:
@inodb @dippindots my comments can/should be fixed in a follow up PR since I know we are trying to get this out soon.

src/shared/components/query/quickSearch/QuickSearch.tsx
dippindots: @alisman Sure!

inodb:
@alisman Thanks for reviewing! I've renamed the props and updated the documentation on the backend PR. Also added the replace+clear routing parameters

inodb:
Several E2E errors, but don't seem to related to this PR

PR #2124, opened 2019-03-15
remove highlight when selecting an option

dippindots:
remove highlight when selecting an option
Fix https://github.com/cBioPortal/cbioportal/issues/5826

### DIFF
--- a/src/shared/components/query/quickSearch/QuickSearch.tsx
+++ b/src/shared/components/query/quickSearch/QuickSearch.tsx
@@ -232,7 +232,22 @@ export default class QuickSearch extends React.Component {
                     maxMenuHeight={550}
                     styles={{
                         dropdownIndicator: ()=>{ return { display:'none' } },
-                        control: (provided:any)=>{ return { ...provided, cursor:'text' } }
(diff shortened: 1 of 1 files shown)


src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -232,7 +232,22 @@ export default class QuickSearch extends React.Component {
...
-                        control: (provided:any)=>{ return { ...provided, cursor:'text' } }
+                        control: (provided:any)=>{ return { ...provided, cursor:'text' } },
+                        option: (base:any, state:any) => {
+                            if (state.isSelected) {
+                                return {
+                                    ...base,
+                                    backgroundColor: 'inherit',
+                                    color: 'inherit',
+                                    ':hover': {
+                                        backgroundColor: '#DEEBFF'
```
alisman: is there a reason we need to specify a hover color?  could we just use inherit?

dippindots: If we use inherit, the background color will be 'none' since its parent doesn't have the background color. So I set the color to which defined in the select2 library.

PR #2126, opened 2019-03-15
move the instruction to the left

dippindots:
move the instruction text to the left
Fix https://github.com/cBioPortal/cbioportal/issues/5811.


dippindots:
We decided to use another approach.

PR #2130, opened 2019-03-18
Keep search text

dippindots:
For the quick search, allow the user to click the input field again and modified the existing text.
Fix https://github.com/cBioPortal/cbioportal/issues/5827.


alisman:
@inodb @dippindots https://cbioportal-frontend-pr-2130.herokuapp.com/
do you think the menu should open up again when you refocus on input (and it still has text in it) ?  feels awkward to me that i have to change the text to get results to open again.

PR #2131, opened 2019-03-18
Update rc with recent API changes

adamabeshouse:


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -81,10 +81,10 @@ export function getAlterationRowData(alterationEnrichments: AlterationEnrichment
...
             entrezGeneId: alterationEnrichment.entrezGeneId,
             cytoband: alterationEnrichment.cytoband, 
-            alteredCount: alterationEnrichment.alteredCount,
-            alteredPercentage: alterationEnrichment.alteredCount / totalAltered * 100,
-            unalteredCount: alterationEnrichment.unalteredCount,
-            unalteredPercentage: alterationEnrichment.unalteredCount / totalUnaltered * 100,
+            alteredCount: alterationEnrichment.set1CountSummary.alteredCount,
+            alteredPercentage: alterationEnrichment.set1CountSummary.alteredCount / totalAltered * 100,
+            unalteredCount: alterationEnrichment.set2CountSummary.alteredCount,
+            unalteredPercentage: alterationEnrichment.set2CountSummary.alteredCount / totalUnaltered * 100,
```
adamabeshouse: @onursumer @zhx828 @averyniceday should the denominator be ..CountSummary.profiledCount? or total count as it is now?

zhx828: profiledCount sounds right to me. This is similar to what we have in study view about sequenced vs all?

jjgao: I think it should be profiledCount. Note that different genes may have different profiledCount. We should really take the count of samples/patients that are profiled for both genes and it should probably be returned in the alterationEnrichment result.

How does the backend do the counting?

adamabeshouse: @averyniceday can you answer this?

averyniceday: @ao508 can you comment on this?

PR #2132, opened 2019-03-19
restore download tab on home page

dippindots:
restore download tab on the home page (sequence: query, quick-search, download).
Fix https://github.com/cBioPortal/cbioportal/issues/5810 and https://github.com/cBioPortal/cbioportal/issues/5872.


src/shared/components/query/QueryAndDownloadTabs.tsx
```diff
@@ -64,6 +65,8 @@ export default class QueryAndDownloadTabs extends React.Component<IQueryAndDownl
 	@autobind
 	onSelectTab(tabId:string)
 	{
+		this.store.forDownloadTab = tabId === DOWNLOAD
```
alisman: this method should be @action because it sets multiple observable values

dippindots: Thanks! I should remember this, I updated my pr.

PR #2133, opened 2019-03-19
Introduce configurable "quick select" buttons

alisman:


src/shared/components/query/StudyListLogic.ts
```diff
@@ -7,347 +7,314 @@ import {computed, action} from "mobx";
...
-	// this is temporary until we can better configure quick selection
-	// if there are pan can studies and there is no filtering, we want to show quick select button
-	@computed get showQuickSelect(){
-		return !getBrowserWindow().navigator.webdriver &&
-			( _.some(this.store.cancerStudies.result, (study)=>study.studyId.includes(PAN_CAN_SIGNATURE)) &&
-			!this.isFiltered);
-	}
+    // this is temporary until we can better configure quick selection
+    // if there are pan can studies and there is no filtering, we want to show quick select button
+    @memoize quickSelectButtons(quick_select_buttons:string|null) {
```
alisman: this function changed

src/shared/components/query/StudyListLogic.ts
```diff
@@ -7,347 +7,314 @@ import {computed, action} from "mobx";
...
-	{
-		let meta = this.getMetadata(node);
-		return meta.ancestors.length;
-	}
-
-	isHighlighted(node:CancerTreeNode):boolean
-	{
-		return !!this.store.searchText && !!this.map_node_filterBySearchText.get(node);
-	}
+export default class StudyListLogic {
```
alisman: @adamabeshouse this file has very few actual changes.  i just finally changed it's line endings to unix style (from Andy's windows).   I marked where I changed

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -1,391 +1,412 @@
...
+        }
+    }
+
+
+    render() {
+
+        const {selectableSelectedStudyIds, selectableSelectedStudies, shownStudies, shownAndSelectedStudies} =
+            this.logic.mainView.getSelectionReport();
+
+        const quickSetButtons = this.logic.mainView.quickSelectButtons(AppConfig.serverConfig.skin_quick_select_buttons);
```
alisman: this is new.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -1,391 +1,412 @@
...
+                        </Then>
+                    </If>
+                    <div className={styles.cancerStudyListContainer} data-test='cancerTypeListContainer'>
+
+                        <div className="checkbox" style={{marginLeft: 19}}>
+                            <If condition={shownStudies.length > 0}>
+                                <If condition={!this.logic.mainView.isFiltered && !_.isEmpty(quickSetButtons)}>
+                                    <Then>
+                                        <div className={styles.quickSelect}>
+                                            <QuickSelectButtons onSelect={(matches)=>this.selectMatchingStudies(matches)}
```
alisman: this is new

adamabeshouse: doesnt matter very much but you could autobind this.selectedMatchingStudies and just pass that in

src/config/config.spec.ts
```diff
@@ -0,0 +1,24 @@
+import {ServerConfigHelpers} from "./config";
+import { assert } from 'chai';
+
```
alisman: this whole file is new

src/shared/components/query/QuickSelectButtons.tsx
```diff
@@ -0,0 +1,36 @@
+import * as React from "react";
+import * as _ from "lodash";
+import {observer} from "mobx-react";
+import {CategorizedConfigItems} from "../../../config/IAppConfig";
+import DefaultTooltip from "../defaultTooltip/DefaultTooltip";
+
+
```
alisman: this whole file is new

src/config/config.ts
```diff
@@ -107,12 +107,14 @@ export class ServerConfigHelpers {
         return (matches) ? matches.map((s:string)=>s.trim()) : [];
     }
 
-    @memoize static priority_studies(str:string|null): PriorityStudies{
+    @memoize static parseConfigFormat(str:string|null): CategorizedConfigItems {
```
adamabeshouse: this seems a very general name for a specific config item format?

alisman: it's actually is now somewhat generalized, which is why i changed the name.  it used to be just for priority_studies, but now i'm using also for these quick select buttons.   really we need to move this kind of structured configuration to json, which i think we will do soon.  so, hopefully this stuff won't be around much longer.

adamabeshouse: okay, I guess that's fine

PR #2136, opened 2019-03-20
Fix bug with OQL eg BRCA1: MUT=DRIVER

adamabeshouse:


src/shared/lib/oql/oqlfilter.js
```diff
@@ -174,20 +174,30 @@ export function parsedOQLAlterationToSourceOQL(alteration) {
             } else {
                 ret = ["CNA",alteration.constr_rel,alteration.constr_val].join("");
             }
-            ret += alteration.modifiers.map(function(modifier) { return "_"+modifier; }).join("");
+            if (alteration.modifiers.length > 0) {
+                ret += "_";
+            }
+            ret += alteration.modifiers.join("_");
```
alisman: maybe doesn't matter but you could put line 180 inside the if statment above, no?

src/shared/lib/oql/oqlfilter.spec.ts
```diff
@@ -148,8 +148,8 @@ describe("unparseOQLQueryLine", ()=>{
         assert.equal(unparseOQLQueryLine(parsedLine), "TP53: MUT_GERMLINE MUT=MISSENSE_SOMATIC MUT=proteinchange_GERMLINE;");
     });
     it("unparses queries with driver modifier", ()=>{
-        const parsedLine = parseOQLQuery("TP53: DRIVER CNA_DRIVER DRIVER_CNA FUSION_DRIVER DRIVER_FUSION TRUNC_DRIVER DRIVER_TRUNC AMP_DRIVER DRIVER_HOMDEL GERMLINE SOMATIC_MISSENSE proteinchange_GERMLINE;")[0];
-        assert.equal(unparseOQLQueryLine(parsedLine), "TP53: DRIVER CNA_DRIVER CNA_DRIVER FUSION_DRIVER FUSION_DRIVER MUT=TRUNC_DRIVER MUT=TRUNC_DRIVER AMP_DRIVER HOMDEL_DRIVER MUT_GERMLINE MUT=MISSENSE_SOMATIC MUT=proteinchange_GERMLINE;");
+        const parsedLine = parseOQLQuery("TP53: DRIVER MUT=DRIVER CNA_DRIVER DRIVER_CNA FUSION_DRIVER DRIVER_FUSION TRUNC_DRIVER DRIVER_TRUNC AMP_DRIVER DRIVER_HOMDEL GERMLINE SOMATIC_MISSENSE proteinchange_GERMLINE;")[0];
```
alisman: maybe add an assertion on the parsedLine showing that the MUT was interpreted correctly?  i guess it might be covered by the unparsing below, but would be nice to see

src/shared/lib/oql/oqlfilter.js
adamabeshouse: yeah, that's true

src/shared/lib/oql/oqlfilter.spec.ts
adamabeshouse: Sure. I also handled a test that it filters correctly, further down in this file.

PR #2137, opened 2019-03-20
Making plot tabs more compact vertically 

coderrsid:
fixing https://github.com/cbioportal/cbioportal/issues/5633

Making plot tabs slightly more compact vertically so that it fits a 13 inch screen easily.
Made it compact, decrease font-size, decrease white spaces and moved the colour-samples-options inside the plot.

- Made it compact
- Decreased font-size
- Removed extra white space
- Moved colour sample options inside plot.

# Screenshots
Screenshot 2019-03-21 at 12 11 09 AM
Screenshot 2019-03-21 at 12 11 18 AM

# Notify reviewers
@schultzn @jjgao @alisman @adamabeshouse


alisman:
there's still a problem here.  you have 2 commits.  one with your changes.  one merge master to your branch.    you should never merge master to hotfix branch.  you should rebase your hotfix branch on master.   the way i would fix this start a new local branch called "whatever" from latest master.  then use git cherry-pick to take JUST the single commit that has your changes from your current branch.    then make a new PR from this branch.   then you should have just a single commit containing the changes you made.  let me know if you need help!  keep going!

src/pages/resultsView/plots/styles.scss
```diff
@@ -34,10 +34,32 @@
     .axisBlock {
       background:#eee;
       border-radius:4px;
+      padding-bottom: 0;
     }
+  } 
+}
+
+.tab-title {
```
alisman: just noticed something else.  these new style classes will actually be global to project and could affect things inadvertantly since they have generic names (e.g. label-text).   The sylesheets are not modular in this way even though it seems they should be.  So you should copy these classes inside the .plotsTab class above it.  then they will ONLY be applied to elements inside .plotsTab div

coderrsid:
@alisman @adamabeshouse check my new pr : https://github.com/cBioPortal/cbioportal-frontend/pull/2148

PR #2151, opened 2019-03-22
Fix BoxPlot styling issue

kalletlak:
1. Reduce height of zero count bars
2. Fix Axis offset
3. Reduce white space between groups
4. Reduce bar width if there are more than 10 categories

Screen Shot 2019-03-25 at 9 48 54 AM


src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -309,9 +309,14 @@ export default class MultipleCategoryBarPlot extends React.Component<IMultipleCa
             if (this.props.stacked) {
                 numBars = this.data[0].counts.length;
             } else {
-                numBars = _.sumBy(this.data, categoryPlotData => categoryPlotData.counts.length);
+                //majorCategories*minorCategories
+                numBars = this.data[0].counts.length * this.data.length;
+                //remove additional padding added by barSeparation
```
adamabeshouse: wait if this is a problem then we should just remove barSeparation. I'm confused why would we add and then remove space?

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -330,7 +335,9 @@ export default class MultipleCategoryBarPlot extends React.Component<IMultipleCa
     }
 
     @computed get barWidth() {
-        return this.props.barWidth || 10;
+        let barWidth = this.props.barWidth || 10;
+        //if number of minorCategories greater than 10 then reduce the width by half
+        return this.props.stacked ? barWidth : this.data.length > 10 ? barWidth/2 : barWidth;
```
adamabeshouse: can you write it out? I think nested ternarys are harder to read

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -355,8 +362,18 @@ export default class MultipleCategoryBarPlot extends React.Component<IMultipleCa
...
+                if (c.count === 0) {
+                    addOffset = true;
+                    break;
+                }
+            }
+            if (addOffset) break;
+        }
         //add a small offset when its not a stacked bar to show empty bar
-        return this.props.stacked ? 0 : (0.03 * (this.maxMajorCount / NUM_AXIS_TICKS));
+        return this.props.stacked ? 0 : addOffset ? (0.01 * (this.maxMajorCount / NUM_AXIS_TICKS)) : 0;
```
adamabeshouse: expand nested ternary please

adamabeshouse:
there should be zero bars here, right?

image

kalletlak:
there is, I reduced its height

adamabeshouse:
I dont see any zero bars?

kalletlak:
ok, I'll revert that back

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -326,11 +330,17 @@ export default class MultipleCategoryBarPlot extends React.Component<IMultipleCa
     }
 
     @computed get barSeparation() {
-        return 0.2*this.barWidth;
+        return this.props.stacked ? 0.2 * this.barWidth : 0;
```
adamabeshouse: :+1:

PR #2163, opened 2019-03-26
Fixes in response to Tali's comments

kalletlak:


src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -327,7 +320,7 @@ export default class ClinicalData extends React.Component<IClinicalDataProps, {}
                         vertData={(this.vertAxisDataPromise.result! as IStringAxisData).data}
                         horzCategoryOrder={(this.horzAxisDataPromise.result! as IStringAxisData).categoryOrder}
                         vertCategoryOrder={(this.vertAxisDataPromise.result! as IStringAxisData).categoryOrder}
-                        categoryToColor={this.categoryToColor}
+                        categoryToColor={STUDY_VIEW_CONFIG.colors.reservedValue}
```
adamabeshouse: I think the right way to handle this is to go into GroupComparisonStore and assign these colors to a group with the corresponding name. We would want those colors to be in every tab not just this one, right?

kalletlak: yeah, true. I'll update

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -267,14 +268,14 @@ export default class ClinicalData extends React.Component<IClinicalDataProps, {}
     }
 
     @computed get categoryToColor() {
-        return _.reduce(this.props.store.uidToGroup.result!,(acc, next, key)=>{
-            acc[key] = next.color
+        //add group colors and reserved category colors
+        return _.reduce(this.props.store.uidToGroup.result!, (acc, next) => {
+            acc[next.name ? next.name : next.uid] = next.color;
```
adamabeshouse: as discussed, use only name and not uid here

PR #2167, opened 2019-03-27
fix popup disappear

dippindots:
fix popup disappear
Fix https://github.com/cBioPortal/cbioportal/issues/5896.

### DIFF
--- a/src/shared/components/query/quickSearch/QuickSearch.tsx
+++ b/src/shared/components/query/quickSearch/QuickSearch.tsx
@@ -46,10 +46,13 @@ export default class QuickSearch extends React.Component {
     @observable private studyPageMultiplier: number = 0;
     @observable private genePageMultiplier: number = 0;
     @observable private patientPageMultiplier: number = 0;
-
     @observable private inputValue:string = "";
-    @observable private menuIsOpen:boolean = false;
+    @observable private isFocusing = false;
     @observable private isLoading = false;
+
+    @computed get menuIsOpen() {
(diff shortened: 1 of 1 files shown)


src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -278,10 +289,10 @@ export default class QuickSearch extends React.Component {
     render() {
 
         return (
-            <div>
+            <div onFocusCapture={this.onFocus}>
                 <Select
                     options={this.options.result || []}
-                    autoFocus={true}
+                    autoFocus={false}
```
inodb: @dippindots why does auto focus need to be disabled? I like that one :) - allows you to query immediately from the homepage without having to click on the searchbox

dippindots: I can turn it on, I am using the focus event to trigger the menu now. I turned off this because we will have an empty menu showing up when we landing to the quick-search tab. Should we allow this?

inodb: @dippindots i see - maybe we can hide the popup menu when the search bar is empty?

inodb:
@dippindots for some reason i don't see any tooltip when testing with:

```
localStorage.heroku = 'cbioportal-frontend-pr-2167'
```
only after clicking away from the searchbox and clicking on it again it starts giving some results

A minor note: it would also be nice if instead of it showing `No options` when there are no search results, it would be nice to show `No results`

PR #2168, opened 2019-03-27
group comparison manager - dont allow two groups of same name

adamabeshouse:
image

image

### DIFF
--- a/src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
+++ b/src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import {observer} from "mobx-react";
 import {
-    caseCounts,
+    caseCounts, DUPLICATE_GROUP_NAME_MSG,
     getNumPatients,
     getNumSamples,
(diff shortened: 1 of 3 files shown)


src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
```diff
@@ -74,6 +76,18 @@ export default class GroupCheckbox extends React.Component<IGroupCheckboxProps,
         return `${this.props.group.name} (${caseCounts(getNumSamples(this.props.group), getNumPatients(this.props.group))})`;
     }
 
+    @computed get editSubmitError() {
+        if (this.nameInput.length === 0) {
+            return { message: "Please enter a name."};
+        } else if (this.nameInput === this.props.group.name) {
+            return {};
```
kalletlak: what does this empty object means here?

adamabeshouse: It's a little confusing... it's to signify that there's an error, so that submit is disabled, but there's no message, so that there's no tooltip. I'll add a comment to explain that

PR #2171, opened 2019-03-27
don't show the quick-search and download tabs when modify the query

dippindots:
Don't show the quick-search and download tabs when modifying the query
Fix https://github.com/cBioPortal/cbioportal/issues/5897.

### DIFF
--- a/src/pages/resultsView/querySummary/QuerySummary.tsx
+++ b/src/pages/resultsView/querySummary/QuerySummary.tsx
@@ -144,7 +144,8 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
     @computed get queryForm(){
         return <div style={{marginTop:10}}>
             <QueryAndDownloadTabs onSubmit={this.onSubmit}
-                                  showQuickSearchTab={AppConfig.serverConfig.quick_search_enabled}
+                                  showQuickSearchTab={false}
+                                  showDownloadTab={false}
                                   showAlerts={true}
                                   getQueryStore={()=>createQueryStore(getBrowserWindow().routingStore.query)}
             />
--- a/src/shared/components/query/QueryAndDownloadTabs.tsx
+++ b/src/shared/components/query/QueryAndDownloadTabs.tsx
@@ -25,6 +25,7 @@ interface IQueryAndDownloadTabsProps
 {
 	onSubmit?:()=>void;
 	showQuickSearchTab:boolean;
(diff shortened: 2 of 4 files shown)


src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -145,6 +145,7 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
         return <div style={{marginTop:10}}>
             <QueryAndDownloadTabs onSubmit={this.onSubmit}
                                   showQuickSearchTab={AppConfig.serverConfig.quick_search_enabled}
+                                  onlyShowQueryTab={true}
```
inodb: @dippindots 	i think since we have one property `showQuickSearchTab`, it's slightly unclear what the combination `onlyShowQueryTab` & `showQuickSearchTab` does. Maybe better to add a prop `showDownloadTab` and set both `showQuickSearchTab` and `showDownloadTab` to false on the resultsviewpage

dippindots: Sure!

src/shared/components/query/QueryAndDownloadTabs.tsx
```diff
@@ -37,7 +38,7 @@ export default class QueryAndDownloadTabs extends React.Component<IQueryAndDownl
 	constructor(props:IQueryAndDownloadTabsProps){
 		super();
 
-		if (getBrowserWindow().localStorage.getItem(QUICK_SEARCH_LS_KEY) === QUICK_SEARCH_TAB_ID) {
+		if (getBrowserWindow().localStorage.getItem(QUICK_SEARCH_LS_KEY) === QUICK_SEARCH_TAB_ID && props.showQuickSearchTab) {
```
inodb: @dippindots minor thing - i prefer putting the props condition first. When i look at the file i like to mentally be able to blank out certain parts of the code based on the configuration of the component, so i know what i can ignore when im for instance not working on quick search. In this case it might also be slightly faster (although prolly negligible), because you don't have to do a localStorage lookup when `showQuickSearchTab ` is false. With `&&` the second condition does not get executed if the first is false

dippindots: Thanks Ino, I just saw this. Updated.

inodb:
Nice work @dippindots 🚀 !

PR #2174, opened 2019-03-28
SV: save user personalized view

kalletlak:
# What? Why?
Fixes part of https://github.com/cBioPortal/cbioportal/issues/5892

Model used to save user settings
```
{
    id: string,
    name?: string,
    chartType?: ChartType,
    groups?: CustomGroup[]
    layout?: {
        x: number,
        y: number,
        w: number;
        h: number;
    }
}
```
There are couple of the issues causing the client cannot be populated through swagger and clients
1. no model for comparison and main sessions
2. no proper support for JsonSubTypes(child classes of Session class) in swagger

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -123,6 +123,27 @@ import jStat from 'jStat'
 import {CancerGene, Gene as OncokbGene} from "../../shared/api/generated/OncoKbAPI";
 import MobxPromiseCache from "shared/lib/MobxPromiseCache";
 
+import Timer = NodeJS.Timer;
+import { AppStore } from 'AppStore';
+
+export type ChartUserSetting = {
+    id: string,
+    name?: string,
+    chartType?: ChartType,
+    groups?: CustomGroup[], //used when it is custom chart
+    layout?: {
+        x: number,
+        y: number,
+        w: number;
+        h: number;
+    },
+    patientAttribute: boolean
+}
+
+export type StudyPageSettings = {
+    chartSettings:ChartUserSetting[],
+    origin:string[]
+}
 
 export enum StudyViewPageTabKeyEnum {
     SUMMARY = 'summary',
@@ -177,11 +198,12 @@ export type StudyViewURLQuery = {
 
 export type CustomGroup = {
     name: string,
-    cases: CustomChartIdentifier[]
+    sampleIdentifiers: CustomChartIdentifier[]
 }
 
-export type NewChart = {
-    name: string,
+export type CustomChart = {
+    name?: string,
+    patientAttribute: boolean,
     groups: CustomGroup[]
 }
 
@@ -192,7 +214,6 @@ export const DataBinMethodConstants: {[key: string]: 'DYNAMIC' | 'STATIC'}= {
 
 export type CustomChartIdentifier = {
     studyId: string,
-    patientAttribute: boolean,
     sampleId: string,
     patientId: string
 }
@@ -229,17 +250,49 @@ export type CopyNumberCountByGeneWithCancerGene = CopyNumberCountByGene & Oncokb
 export class StudyViewPageStore {
     private reactionDisposers:IReactionDisposer[] = [];
 
-    constructor() {
+    constructor(private appStore:AppStore) {
         this.reactionDisposers.push(reaction(() => this.loadingInitialDataForSummaryTab, () => {
             if (!this.loadingInitialDataForSummaryTab) {
                 this.updateChartStats();
+                this.loadUserSettings();
             }
         }));
+
+        this.reactionDisposers.push(reaction(() => [this.visibleAttributes, this.columns, this.chartsDimension.toJS(), this.chartsType.toJS()], () => {
+            this.updateLayout();
+        }));
+
+        this.reactionDisposers.push(reaction(() => toJS(this.currentChartSettingsMap), () => {
+            if(this.isSavingUserSettingsPossible){
+                if (!this.hideRestoreSettingsMsg) {
+                    // hide restore message if its already shown
+                    // this is because the user setting session is going to be updated for every operation once the user is logged in
+                    this.hideRestoreSettingsMsg = true;
+                }
+                this.updateUserSettingsDebounce();
+            }
+        }));
+
+        this.reactionDisposers.push(reaction(() => this.fetchUserSettings.isComplete, (isComplete) => {
+            //execute if user log in from study page
+            if (isComplete && this.isLoggedIn && !this._loadUserSettingsInitially) {
+                this.previousSettings = this.currentChartSettingsMap;
+                this.loadUserSettings();
+            }
+        }));
+
+
+
+
+        
 
         // Include special charts into custom charts list
        SPECIAL_CHARTS.forEach((chartMeta:ChartMetaWithDimensionAndChartType) => {
            const uniqueKey = chartMeta.uniqueKey;
-           const chartType = chartMeta.chartType;
+           if (!this.chartsType.has(uniqueKey)) {
+               this.chartsType.set(uniqueKey, chartMeta.chartType);
+           }
+           const chartType = this.chartsType.get(uniqueKey);
            if (chartType !== undefined) {
                this._customCharts.set(uniqueKey, {
                    displayName: chartMeta.displayName,
@@ -259,6 +312,15 @@ export class StudyViewPageStore {
            }
        });
     }
+
+    @computed get isLoggedIn() {
+        return this.appStore.isLoggedIn;
+    }
+
+    @observable hideRestoreSettingsMsg = this.isLoggedIn;
+
+    //this is set on initial load
+    private _loadUserSettingsInitially = this.isLoggedIn;
 
     // make sure the reactions are disposed when the component which initialized store will unmount
     destroy() {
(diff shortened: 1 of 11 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -295,9 +296,22 @@ export class StudyViewPageStore {
         this.reactionDisposers.push(reaction(() => this.loadingInitialDataForSummaryTab, () => {
             if (!this.loadingInitialDataForSummaryTab) {
                 this.updateChartStats();
+                this.loadUserSettings();
```
zhx828: loadUserSettings should be done before updateChartStatus

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1885,8 +1902,7 @@ export class StudyViewPageStore {
     }
 
     @autobind
-    @action addCustomChart(newChart:NewChart) {
-        const uniqueKey = this.newCustomChartUniqueKey();
+    @action addCustomChart(newChart:NewChart, uniqueKey:string = this.newCustomChartUniqueKey(), newlyAdded:boolean=true) {
```
zhx828: After adding the parameter `newlyAdded`, this function will need to be called `updateCustomChart`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2046,6 +2067,7 @@ export class StudyViewPageStore {
 
     @computed
     get visibleAttributes(): ChartMeta[] {
+        //
```
zhx828: ?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2067,14 +2089,130 @@ export class StudyViewPageStore {
...
         } else {
             return false;
         }
     }
 
+    private updateSettingsTimeout: Timer;
+    private studySettings: any = {};
+
+    @autobind
+    updateUserSettings(chartVisibilityUpdated: boolean, chartTypeUpdated: boolean, customChartAdded: boolean) {
```
zhx828: I would rather separate this function to three su-functions. They are not inter connected and you have do all if check.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2067,14 +2089,130 @@ export class StudyViewPageStore {
...
+            this.fetchUserSettings.isPending
         ) {
             return true;
         } else {
             return false;
         }
     }
 
+    private updateSettingsTimeout: Timer;
+    private studySettings: any = {};
```
zhx828: Let's model studySettings.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2067,14 +2089,130 @@ export class StudyViewPageStore {
...
+                let userSettings = {} as any;
+                userSettings[studyIdString] = updatedUserSetting;
+                //TODO: replace with api call once the model is finalized and session service is ready
+                localStorage.setItem("userSettings", JSON.stringify(userSettings));
+                this.studySettings = updatedUserSetting;
+            }
+        }, 1000)
+    }
+
+    readonly fetchUserSettings = remoteData({
```
zhx828: We need to handle onError

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -974,16 +974,28 @@ export function calculateLayout(visibleAttributes: ChartMeta[], cols: number, cu
     // sort the visibleAttributes by priority
     visibleAttributes.sort(chartMetaComparator);
 
+    const visibleAttributeSet = _.keyBy(visibleAttributes,attribute=>attribute.uniqueKey)
```
zhx828: What's the purpose of this?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: it depends on how we model `studySettings`, whether these attributes goes this way or a normalized for each chart

src/pages/studyView/StudyViewPageStore.ts
kalletlak: will clean it up

src/pages/studyView/StudyViewPageStore.ts
kalletlak: instead it's better to rename `newlyAdded` to something else, may be `loadedByUserClick`?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: no, since `updateChartStats` updates chart visibility and type which might be different from user saved ones

src/pages/studyView/StudyViewUtils.tsx
kalletlak: it's to check whether the charts in `currentGridLayout` are visible. if chart is not visible then discard it's layout. I'll add a comment

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2261,17 +2312,129 @@ export class StudyViewPageStore {
...
             this.mutationProfiles.isPending ||
-            this.cnaProfiles.isPending
-        ) {
-            return true;
-        } else {
-            return false;
+            this.cnaProfiles.isPending;
+
+        if (this.loadInitialUserSettings) {
+            pending = pending || this.fetchUserSettings.isPending
```
zhx828: When a user logged in while in study view, I think this will trigger a whole page refresh?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2261,17 +2312,129 @@ export class StudyViewPageStore {
...
+
+    @autobind
+    @action
+    updateUserSettings(forceSave?:boolean) {
+        clearTimeout(this.updateSettingsTimeout);
+        //instantly save settings for any of these conditions
+        const timeout = forceSave || _.isEmpty(this.currentChartSettings) ? 0 : 3000;
+
+        this.updateSettingsTimeout = setTimeout(() => {
+            if(forceSave || _.isEmpty(this.previousSettings) || !_.isEqual(this.previousSettings, this.currentChartSettings)){
```
zhx828: The `_.isEmpty(this.previousSettings)` check is unnecessary

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1078,22 +1078,40 @@ export function getDefaultChartDimension(): ChartDimension {
...
+    useCurrentGridLayout: boolean,
+    currentFocusedChartByUser?: ChartMeta,
+    currentFocusedChartByUserDimension?: ChartDimension): Layout[] {
+
     let layout: Layout[] = [];
     let matrix = [new Array(cols).fill('')] as string[][];
     // sort the visibleAttributes by priority
     visibleAttributes.sort(chartMetaComparator);
-
+    const visibleAttributeMap = _.map(visibleAttributes, attribute => attribute.uniqueKey)
```
zhx828: Let's move this under `if (useCurrentGridLayout)`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2262,17 +2317,133 @@ export class StudyViewPageStore {
...
+        }
+    });
+
+    @autobind
+    @action
+    loadUserSettings() {
+        if (!_.isEmpty(this.fetchUserSettings.result)) {
+            this.previousSettings = _.keyBy(this.fetchUserSettings.result, chartSetting => chartSetting.id)
+            this._chartVisibility.clear();
+            this.currentGridLayout = [];
```
zhx828: We should have a function to clear up the layout settings. I think it's necessary `currentFocusedChartByUser` and `currentFocusedChartByUserDimension` should be set back to original state as well.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -724,6 +759,17 @@ export class StudyViewPageStore {
         }
     }
 
+    @action
+    async updateUserState(isLoggedIn:boolean) {
```
alisman: why not just make these observables (loadUserSettingsInitially) into @computed which reference the appStore directly?

alisman: one could argue that couples the two stores too tightly, but i think it's ok.

kalletlak: updated

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2262,17 +2310,140 @@ export class StudyViewPageStore {
...
+
+    private updateSettingsTimeout: Timer;
+    private previousSettings: {[id:string]:ChartUserSetting} = {};
+
+    @autobind
+    @action
+    updateUserSettings(forceSave?:boolean) {
+        // forceSave => instantaneously save user settings independent of whether there are any changes 
+        // between current and previously saved settings
+        clearTimeout(this.updateSettingsTimeout);
```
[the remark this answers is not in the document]
adamabeshouse: I never thought about it that way before but I think you're right. this would be debounce, I believe.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2262,17 +2310,140 @@ export class StudyViewPageStore {
...
+        },
+        default: [],
+        onError:()=>{
+            // fail silently when an error occurs
+        }
+    });
+
+    @autobind
+    @action private clearPageSettings() {
+        this._chartVisibility.clear();
```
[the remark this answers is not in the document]
kalletlak: no

alisman:
image
Now that I understand this behavior, I think we need to update this message.  What about "Saved layout preferences have been applied. Undo?"  @jjgao can you comment?

jjgao:
> Saved layout preferences have been applied. Undo?

I think that's better! Maybe "You previously saved layout preferences have been applied. Undo?"

kalletlak:
> > Saved layout preferences have been applied. Undo?
> 
> I think that's better! Maybe "You previously saved layout preferences have been applied. Undo?"

@alisman @jjgao previously saved preferences is not yet applied here. this message is to whether use previously saved ones or continue with what's present on the screen

jjgao:
@kalletlak got it. But the user may not remember what was saved...

jjgao:
Maybe we should just switch to the user's saved view and allow them to undo?

kalletlak:
@jjgao in this case user may loose applied filters, if any. is that acceptable?

jjgao:
@kalletlak good point. Maybe the ones that are used for filtering should be made hidden even if they are hidden in the previous state?

kalletlak:
@jjgao could you test it on https://beta.cbioportal.org/ with `localStorage.heroku = "cbioportal-frontend-pr-2174"`

PR #2177, opened 2019-03-29
Rewrite Overlap Venn diagram so that we can highlight set complements

adamabeshouse:
Venn.js doesn't support interaction with set complements (i.e. only group A, only group B)

It's hard to support set complements in SVG. I'm doing it by
(1) building regions for each intersection (good support in SVG)
(2) drawing the regions in order so that each lower intersection corresponds to the complement of it with what comes next

image

image

### DIFF
--- a/src/pages/groupComparison/Overlap.tsx
+++ b/src/pages/groupComparison/Overlap.tsx
@@ -11,7 +11,12 @@ import DownloadControls from 'shared/components/downloadControls/DownloadControl
 import {MakeMobxView} from "../../shared/components/MobxView";
 import Loader from "../../shared/components/loadingIndicator/LoadingIndicator";
 import ErrorMessage from "../../shared/components/ErrorMessage";
-import {getCombinations, getSampleIdentifiers} from "./GroupComparisonUtils";
+import {
+    ENRICHMENTS_NOT_2_GROUPS_MSG,
+    getCombinations,
+    getSampleIdentifiers,
+    OVERLAP_NOT_ENOUGH_GROUPS_MSG
+} from "./GroupComparisonUtils";
 import {remoteData} from "../../shared/api/remoteData";
 
 export interface IOverlapProps {
@@ -40,6 +45,27 @@ export default class Overlap extends React.Component<IOverlapProps, {}> {
     componentDidUpdate() {
         this.plotExists = !!this.getSvg();
     }
+
+    readonly tabUI = MakeMobxView({
+        await:()=>{
+            if (this.props.store.activeGroups.isComplete &&
+                this.props.store.activeGroups.result.length < 2) {
+                // dont bother loading data for and computing overlap if not enough groups for it
+                return [this.props.store.activeGroups];
+            } else {
+                return [this.props.store.activeGroups, this.overlapUI];
+            }
+        },
+        render:()=>{
+            if (this.props.store.activeGroups.result!.length < 2) {
+                return <span>{OVERLAP_NOT_ENOUGH_GROUPS_MSG}</span>;
+            } else {
+                return this.overlapUI.component;
+            }
+        },
+        renderPending:()=><Loader isLoading={true} centerRelativeToContainer={true} size={"big"}/>,
+        renderError:()=><ErrorMessage/>
+    });
 
     @autobind
     private getSvg() {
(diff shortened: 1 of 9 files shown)


src/pages/groupComparison/VennSimple.tsx
```diff
@@ -0,0 +1,206 @@
...
+            map[group.uid] = {
+                x: svgCenter.x + distanceFromCenter*Math.cos(this.angleIncrement*index),
+                y: svgCenter.y + distanceFromCenter*Math.sin(this.angleIncrement*index)
+            };
+            return map;
+        }, {} as {[uid:string]:{x:number, y:number}});
+    }
+
+    @computed get displayElements() {
+        const clipPaths = this.props.groups.map((group, index)=>(
```
adamabeshouse: this whole section probably needs a lot more commenting

adamabeshouse:
just realized this weird UI moment - complete overlap means that the base groups are empty, so the colors from the legend aren't actually in the plot

image


how should we handle this?

adamabeshouse:
@jjgao @alisman is it a problem that the intersection areas are not proportional?

src/pages/groupComparison/Overlap.tsx
```diff
@@ -41,6 +46,27 @@ export default class Overlap extends React.Component<IOverlapProps, {}> {
...
+            }
+        },
+        render:()=>{
+            if (this.props.store.activeGroups.result!.length < 2) {
+                return <span>{OVERLAP_NOT_ENOUGH_GROUPS_MSG}</span>;
+            } else {
+                return this.overlapUI.component;
+            }
+        },
+        renderPending:()=><Loader isLoading={true} size={"big"}/>,
```
alisman: I think this should be centered yes? i mean even if we switch to postiion:absolute.

adamabeshouse: yeah, makes sense

src/pages/groupComparison/VennSimple.tsx
```diff
@@ -0,0 +1,230 @@
...
+        default:
+            return d3.interpolateLab(
+                d3.interpolateLab(colors[0], colors[1])(0.5),
+                colors[2]
+            )(0.5);
+    }
+}
+
+function regionMouseOver(e:any) {
+    e.target.setAttribute("fill", e.target.getAttribute("data-hover-fill"));
```
alisman: how come we do this with dom manip instead of react?

adamabeshouse: It's more efficient. Maybe unnecessary optimization though, that has the downside of taking us out of our usual idioms.

src/pages/groupComparison/VennSimple.tsx
```diff
@@ -0,0 +1,230 @@
...
+}
+
+function regionMouseOut(e:any) {
+    e.target.setAttribute("fill", e.target.getAttribute("data-default-fill"));
+}
+
+function getExcludedIndexes(combination:number[], numGroupsTotal:number) {
+    const excl = [];
+    for (let i=0; i<numGroupsTotal; i++) {
```
alisman: filter?

adamabeshouse: would have to write a utility function to get the range [0,1,..., numGroupsTotal-1]. Nbd but just went this way instead.

src/pages/groupComparison/VennSimple.tsx
```diff
@@ -0,0 +1,230 @@
...
+        return Math.min(this.props.width, this.props.height) / 4;
+    }
+
+    @computed get minCircleRadius() {
+        return this.maxCircleRadius - 30;
+    }
+
+    @computed get circleRadii() {
+        // Make a linear scale where max group is max circle radius, and max group / 10 is min circle radius
+        const radius = (numCases:number)=>{
```
alisman: feels like it should have a test?

adamabeshouse: I don't know how I would test this besides plugging values into the same formula. I think visual test is good enough for this. What do you think?

src/pages/groupComparison/VennSimple.tsx
```diff
@@ -0,0 +1,263 @@
...
+function regionMouseOut(e:any) {
+    e.target.setAttribute("fill", e.target.getAttribute("data-default-fill"));
+}
+
+function getExcludedIndexes(combination:number[], numGroupsTotal:number) {
+    // get all indexes not in the given combination
+
+    const excl = [];
+    for (let i=0; i<numGroupsTotal; i++) {
+        if (combination.indexOf(i) === -1) {
```
alisman: nbd but trying to get peopke to use array.includes instad of indexof === -1

PR #2180, opened 2019-04-01
resolve issue 5652 to support multiple reference genomes

khzhu:
## What? Why?
Resolve issue [https://github.com/cBioPortal/cbioportal/issues/5652](url) to support multiple reference genomes
- [x] remove **_cytoband_** , **chromosome**, and **_length_** from **_gene_** entity
- [x] fetch reference genome genes of studies of interest using APIs
- [x] use **_cytoband_** from **_reference genome genes_** rather than from **_genes_**
- [x] load chromosome sizes of different reference genomes from JSON file for segments/mutations plots on the patient view
- [x] use **_chr_** from mutation entity instead of calculating from **_cytoband_**
- [x] disable **_View Summary_** when studies selected come from different reference genomes

Fix #

- a
- b

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:
#


khzhu:
Hi @adamabeshouse @onursumer , can I please have your review on the PR? I mades some changes in the codebase you have been working on. It will be great to have some feedback. Please let me if things are unclear or you need more information. Thanks!

adamabeshouse:
@khzhu I'm not familiar with this, I'll leave it to @onursumer

khzhu:
@adamabeshouse Okay. Would you be able to point one, since I need two reviewers. Thanks!

khzhu:
@adamabeshouse , just found I only need one approving review. Great!

khzhu:
@alisman @onursumer , the code is ready for your review. Thanks!

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -57,6 +58,12 @@ export default class CopyNumberTableWrapper extends React.Component<ICopyNumberT
         enableCivic: false
     };
 
+    getCytoband(hugoGeneSymbol:string) {
```
alisman: i would call this getCytobandForGene

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -57,6 +58,12 @@ export default class CopyNumberTableWrapper extends React.Component<ICopyNumberT
...
+        const refGene =  _.find(this.props.refGenes,(gene:ReferenceGenomeGene)=>
+            gene.hugoGeneSymbol===hugoGeneSymbol);
+        return refGene? refGene.cytoband: "-";
```
alisman: returing "-" seems like a presentation choice.  I would return null or undefined and then have view code specify what to show in that case.

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -31,7 +31,8 @@ export default class Tracks extends React.Component<TracksPropTypes, {}> {
         // --- end of raphael config ---
 
         // --- chromosome chart ---
-        var chmInfo = tracksHelper.getChmInfo();
+        const chmInfo = tracksHelper.getChmInfo( this.props.mutations?
+            this.props.mutations[0]!.ncbiBuild: "GRCh37");
```
alisman: if that is a default chromosome, you should make it a constant.  e.g. DEFAULT_CHROMOSOME at top of file

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -51,9 +51,15 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
         return generateSegmentFeatures(segments);
     }
 
+    @computed get genome() {
+        const study = this.props.store.studies.result?
+            this.props.store.studies.result[0]: undefined;
+        return study? study.referenceGenome: "hg19";
```
onursumer: maybe create a constant like `DEFAULT_GENOME`

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,38 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
 /**
  * storing chromesome length info
  */
-export function getChmInfo() {
+export function getChmInfo(genomeBuild:string) {
     let sel: any = {};
-    sel.hg19 = [0,249250621,243199373,198022430,191154276,180915260,171115067,159138663,146364022,141213431,135534747,135006516,133851895,115169878,107349540,102531392,90354753,81195210,78077248,59128983,63025520,48129895,51304566,155270560,59373566];
-    sel.total = 3095677412;
-    sel.perc = getChmEndsPerc(sel.hg19,sel.total);
+    sel.genomeRef = {};
```
alisman: you could put this property inside object above

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -286,7 +286,7 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
             this.props.store.activeTranscript &&
             this.props.store.transcriptsByTranscriptId[this.props.store.activeTranscript] &&
             this.props.store.transcriptsByTranscriptId[this.props.store.activeTranscript].proteinLength) ||
-            Math.round(this.props.store.gene.length / 3);
+            0;
```
onursumer: can we set it to some non-zero value like 10?

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,38 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+            break;
+        case "mm10":
+            genomeBuild = "GRCm38"
+            break;
+        default:
+            genomeBuild = "GRCh37";
+
+    }
+    const chromSizes = require('./chromSizes.json');
+    for (let i = 0; i < chromSizes.length; i++) {
```
alisman: instead of using loop break, use _.find to get first matching.  also, what happens if it doesn't find anything?

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,38 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+        if (chromSizes[i]["build"] === genomeBuild) {
+            sel.genomeRef = chromSizes[i]['size'];
+            break;
+        }
+    }
+    sel.total = 0
```
alisman: put this up in the definition of sel at top of this routine

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,38 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    for (let i=0; i < sel.genomeRef.length; i++ ) {
```
alisman: use lodash _.sum for this

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
khzhu: will change.

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
khzhu: agree.

src/pages/patientView/genomicOverview/Tracks.tsx
khzhu: agree. Will make **GRCh37** DEFAULT_GENOME_BUILD for now.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -643,6 +669,23 @@ export class PatientViewPageStore {
         }
     }, undefined);
 
+
+    readonly refGenes = remoteData<ReferenceGenomeGene[]>({
```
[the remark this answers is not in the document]
khzhu: sure, will change.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2079,6 +2081,11 @@ export class ResultsViewPageStore {
         return parsedQString.query.session_id;
     }
 
+    private getCytoBand(hugoGeneSymbol:string):string {
```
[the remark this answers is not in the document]
khzhu: will use getCytoband to be consistent.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2156,6 +2163,28 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly refGenes = remoteData<ReferenceGenomeGene[]>({
```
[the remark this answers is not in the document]
khzhu: agree.

src/pages/resultsView/cnSegments/CNSegments.tsx
khzhu: good idea.

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
khzhu: okay, use 10.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: will do.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: good point. thanks!

src/pages/patientView/genomicOverview/chromSizes.json
```diff
@@ -0,0 +1,14 @@
+[
```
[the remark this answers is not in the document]
khzhu: will change the name.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: good point! Will use _.find. Add if nothing found then default hg19 size to be returned.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: cool! will use _.sum. thanks!

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -134,6 +134,20 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
         );
     });
 
+    @computed
+    get checkReferenceGenome() {
```
alisman: if i understand correctly, this checks if all selected studies have same referencegenome.  It true, lets call it checkForSingleReferenceGenome

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -134,6 +134,20 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
...
+        let genomeValid = true;
+        if (this.store.selectableSelectedStudies.length > 0) {
+            const referenceGenome = this.store.selectableSelectedStudies[0].referenceGenome;
```
alisman: i think you could map to an array of all referneceGenomes, then call _.uniq on that array, then measure length > 1 to determine validity of query

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -225,9 +239,12 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
                             <Observer>
                                 {() => {
 
-                                    const studyLimitReached = (this.store.selectableSelectedStudyIds.length > 50);
+                                    const studyLimitReached = (!this.checkReferenceGenome) ||
```
alisman: seems odd to combine these two invalid "states" no?   I think we should have a separate message for studies from different reference genome

src/shared/components/query/CancerStudySelector.tsx
alisman: also, do we allow user to submit query when checkReferenceGenome is false (mixed genomes) ? Or do we block them from doing so, as we do if they select too many studies?

src/shared/components/query/CancerStudySelector.tsx
khzhu: yes, you are right. Will change.

src/shared/components/query/CancerStudySelector.tsx
khzhu: we do not allow merge studies from different genomes.

src/shared/components/query/CancerStudySelector.tsx
khzhu: agree. Let's have  a new message.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -707,8 +709,8 @@ export class ResultsViewPageStore {
             this.samples
         ],
         invoke: () => {
-            if (this.genes.result) {
-                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
+            if (this.refGenes.result) {
+                return Promise.resolve(_.reduce(_.uniq(this.refGenes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
```
alisman: this code is kind of hard to read.  i would first make a variable uniqueChromosommes and assign map to it.  Then call reduce on variable.

src/shared/lib/StoreUtils.ts
```diff
@@ -85,6 +99,31 @@ export async function fetchGenes(hugoGeneSymbols?: string[],
     }
 }
 
+export async function fetchReferenceGenomeGenes(genomeName:string, hugoGeneSymbols?: string[],
+                                 client: CBioPortalAPI = defaultClient)
+{
+    if (hugoGeneSymbols && hugoGeneSymbols.length) {
+        const order = stringListToIndexSet(hugoGeneSymbols);
+        return _.sortBy(await client.fetchGenesUsingPOST_1({
```
alisman: @onursumer noticed this but it seem slike fetchGenesUsingPost is overloaded (on backend) which produces these indexed methods on client.   We need to better distinguish referenceGenomeGenes from genes.

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -7,7 +7,7 @@ import {Nav, NavItem} from "react-bootstrap";
```
onursumer: We have another `CNSegments` tab for study view: `src/pages/studyView/tabs/CNSegments.tsx`. We may want to update that one as well with the genome information.

src/pages/resultsView/ResultsViewPageStore.ts
khzhu: good idea! will make changes.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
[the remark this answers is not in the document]
alisman: @jj @khzhu  I think our naming scheme needs to carefully distinguish this from the genes collection.  If I understand correctly, there can be multiple ReferenceGenomeGene's per Gene.  Do they represent a particular position on a particular reference genome?  Maybe the name could reflect that, like, ReferenceGenomeGeneDefinition  or ReferenceGenomeGeneVariant or ReferenceGenomeGenePosition?  Something that signals the difference with Gene concept.

src/shared/lib/StoreUtils.ts
onursumer: @alisman @khzhu I have added the relevant review for this one here: https://github.com/cBioPortal/cbioportal/pull/5891

src/pages/resultsView/cnSegments/CNSegments.tsx
khzhu: **_CNSegments.tsx_** page is not affected, since **_IntegrativeGenomicsViewer_** is the one load genome reference fa file from s3.

src/shared/lib/StoreUtils.ts
khzhu: thanks, @onursumer! Those notes I put on schema diagram might help with the review.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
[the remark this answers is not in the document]
khzhu: @alisman , you are right, a gene can have multiple reference genome genes. The reference genome gene are still genes but with different cytogenetic locations. I do not know if we need to distinguish genes from reference genes or just pick up right coordinates/cytobands when plot variants.  @JJ what do you think?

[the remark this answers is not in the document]
jjgao: @khzhu Is it true that you are using `ReferenceGenomeGene` to reflect the database name / api variable? If so, maybe we should keep it.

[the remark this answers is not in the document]
khzhu: @JJ, yes, that is right. I see Aaron's points. ReferenceGenomeGene object essentially encapsulates a set of properties of a gene from multiple genomes. ReferenceGenomeGeneDefinition or ReferenceGenomeGeneProperties might less confuse people.

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
khzhu: changed to getCytobandForGene

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
khzhu: replaced "-" with undefined.

src/pages/patientView/genomicOverview/Tracks.tsx
khzhu: made GRCh37 a const.

src/pages/patientView/genomicOverview/chromSizes.json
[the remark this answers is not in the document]
khzhu: renamed chromosome size file.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: change made. replaced old style javascript code with find and reduce.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: done.

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: used .reduce method.

src/pages/resultsView/ResultsViewPageStore.ts
khzhu: made a variable **_uniqueReferenceGenes_** and call reduce on the variable.

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
khzhu: used getCytoband.

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
khzhu: used referenceGenes.

src/pages/resultsView/cnSegments/CNSegments.tsx
khzhu: defined DEFAULT_GENOME in IGVUtils file.

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
khzhu: replaced 0 with 10.

src/shared/components/query/CancerStudySelector.tsx
khzhu: renamed to checkForSingleReferenceGenome

src/shared/components/query/CancerStudySelector.tsx
khzhu: added a new message for multiple reference genomes.

khzhu:
Hi @alisman @onursumer , I've addressed all comments from your reviews.Please take a final review on this PR and let me know if anything else need to be done. Thanks again for your time!

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1,805 +1,822 @@
...
+
+                return getPathologyReport(this.patientId, 0);
+            } else {
+                return Promise.resolve([]);
+            }
+        },
+        onError: (err: Error) => {
+            // fail silently
+        }
+
```
alisman: looks like you have reformatted this file, causing every line to be updated?

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,36 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    build: string,
+    size: number[]
+};
+
+export function getChmInfo(genomeBuild:string) {
+    const sel: any = {};
+    sel.genomeRef = {};
+    sel.total = 0
+
+    const chromSizes:chromosomeSizeType[] = require('./chromosomeSizes.json');
```
alisman: you should put this require at top of file

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +56,36 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    let referenceGenome = genomeBuilds.get(genomeBuild);
+    if (!referenceGenome || referenceGenome === "") {
+        referenceGenome = DEFAULT_CHROMOSOME;
+    }
+    const genomeSize = chromSizes.find(c => c.build === referenceGenome);
+    if (genomeSize) {
+        sel.genomeRef = genomeSize.size;
+        sel.total = sel.genomeRef.reduce((a: number, b: number) => a + b);
```
alisman: use _.sum

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
khzhu: the file had carriage return in windows format ^M, got red ! warning in WebStorm, so changed to linux CRLF. Should I roll back the change?

src/pages/patientView/genomicOverview/tracksHelper.ts
khzhu: changed.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
alisman: were you able to roll this back?   I just checked PatientVIewPageStore in master and webstorm says it's set to LF, like all rest of files, so perhaps try rebasing?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -721,14 +723,14 @@ export class ResultsViewPageStore {
             this.samples
         ],
         invoke: () => {
-            if (this.genes.result) {
-                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
+            if (this.referenceGenes.result) {
+                const uniqueReferenceGenes = _.uniq(this.referenceGenes.result.map(g => g.chromosome));
```
alisman: is it right that we are assigning "chromosome"s to a variable called referenceGenes?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -721,14 +723,14 @@ export class ResultsViewPageStore {
...
+                return Promise.resolve(uniqueReferenceGenes.reduce(
```
alisman: lets assign result of reduce to local variable first and then resolve it.  hard to read

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
khzhu: will do a rebase. thanks!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2095,6 +2097,15 @@ export class ResultsViewPageStore {
         return parsedQString.query.session_id;
     }
 
+    private getCytoband(hugoGeneSymbol:string):string {
+        const referenceGenes = this.referenceGenes!.result;
+        if (referenceGenes) {
+            const gene = _.find(referenceGenes, (g: ReferenceGenomeGene) => g.hugoGeneSymbol === hugoGeneSymbol);
+            return gene ? gene.cytoband : "";
```
alisman: should we really use empty string here?  seems it would be better to use null until actual use

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2172,6 +2183,28 @@ export class ResultsViewPageStore {
...
 
+    readonly referenceGenes = remoteData<ReferenceGenomeGene[]>({
+        await: ()=>[
+            this.studies,
+            this.mutationEnrichmentData,
+            this.copyNumberAmpEnrichmentData,
+            this.genes
+        ],
+        invoke: async () => {
+            const queryGenes = this.genes.result!.map((g:Gene)=>g.hugoGeneSymbol.toUpperCase());
```
alisman: is there a reason why we have to uppercase them?  seems like not a good practice.

src/pages/resultsView/ResultsViewPageStore.ts
khzhu: we are retrieving copy number segments by chromosomes in reference genome genes.

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -397,6 +399,26 @@ export default class GroupComparisonStore {
...
+        ],
+        invoke: async () => {
+            const mutGenes = this.mutationEnrichmentData.result!.map(
+                (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            const cnvGenes = this.copyNumberData.result!.map(
+                (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            return fetchReferenceGenomeGenes(this.studies.result[0].referenceGenome,
+                mutGenes.concat(cnvGenes));
+        },
+        onError:(err)=>{
```
alisman: we should actually get rid of onError unless the error is specifically dealt with in your code.  Sorry, we are updating our practice around errors.

alisman: getting rid of onError will allow global error handling to detect and show error message

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -70,7 +73,19 @@ export default class AlterationEnrichmentContainer extends React.Component<IAlte
     }
 
     @computed get data(): AlterationEnrichmentRow[] {
+        if (this.props.data && this.props.data.length >0) {
+            this.props.data.forEach((a:AlterationEnrichment)=>{
+            const refGenes = this.props.referenceGenes;
+            if (refGenes && refGenes.length > 0) {
+                const gene = refGenes.find(g => g.entrezGeneId === a.entrezGeneId);
+                if (gene) {
+                    a.cytoband = gene.cytoband;
```
alisman: @adamabeshouse wanted to get your opinion here.  Seems like we are extending api data here.  One question I have is why AlterationEnrichment type has a cytoband property if it is not populated?  or are overriding it here?  It concerns me that we mutating global data source based on referenceGenes property of this component.  If this is accetable to do a general level (application wide) then we should do this in the store to ensure that all consumers of the data see the same thing under all circumstances.

src/pages/studyView/tabs/CNSegments.tsx
```diff
@@ -69,6 +69,12 @@ export default class CNSegments extends React.Component<{ store: StudyViewPageSt
         return this.activePromise.isComplete && (!this.activePromise.result || this.activePromise.result.length === 0);
     }
 
+    @computed get genome() {
+        const study = this.props.store.allPhysicalStudies.result?
+            this.props.store.allPhysicalStudies.result[0]: undefined;
+        return study? study.referenceGenome: DEFAULT_GENOME;
```
alisman: i think there should be a space before ? and another before the colon in this ternary.

alisman: if the reference genome is "global," meaning there is nothing specific to CNSegments, then it probably belongs in the store rather than here?

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -134,6 +134,12 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
         );
     });
 
+    @computed
```
alisman: seems like this belongs in the store

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -225,21 +231,25 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
                             {!!(!this.store.cancerTypes.isPending && !this.store.cancerStudies.isPending) && (
                                 <Observer>
                                     {() => {
+                                    const studyLimitReached = this.store.selectableSelectedStudyIds.length > 50;
+                                    const multipleReferenceGenomes = (!this.checkForSingleReferenceGenome);
+                                    let tooltipMessage = <span>Open summary of selected studies in a new window.</span>;
+                                    if (studyLimitReached) {
```
alisman: i think these two ifs should be if/else and probably studyLimitReached should show first.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -225,21 +231,25 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
...
+                                    return (
                                             <DefaultTooltip
                                                 placement="top"
                                                 overlay={tooltipMessage}
                                                 disabled={!this.store.summaryEnabled}
                                                 mouseEnterDelay={0}
                                             >
 
-                                                <Button bsSize="xs" disabled={studyLimitReached} bsStyle="primary"
+                                                <Button bsSize="xs" disabled={studyLimitReached||multipleReferenceGenomes} bsStyle="primary"
```
alisman: lets take this condition out and assign it to a local variable of the method just to make it super clear.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -225,21 +231,25 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
                             {!!(!this.store.cancerTypes.isPending && !this.store.cancerStudies.isPending) && (
                                 <Observer>
                                     {() => {
+                                    const studyLimitReached = this.store.selectableSelectedStudyIds.length > 50;
```
alisman: @khzhu i realized that this button has been eliminated (replaced by "forked mode" button).  You should rebase and then reapply this logic to the forked button.  Let me know if you need help. SHould be pretty clear when you rebase (or if you look at latest public portal)

src/pages/resultsView/cnSegments/CNSegments.tsx
inodb: should IGV be given some parameter to pick the correct reference genome?

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
adamabeshouse: agreed. we should do it at the top.

src/pages/studyView/tabs/CNSegments.tsx
adamabeshouse: agreed, `genome()` should be put in resultsviewpagestore

src/shared/components/query/CancerStudySelector.tsx
adamabeshouse: agreed. also please rename to `onlyOneReferenceGenome` or something like that, to agree with the getter syntax.

src/shared/components/query/CancerStudySelector.tsx
adamabeshouse: agreed

src/shared/lib/IGVUtils.ts
```diff
@@ -3,6 +3,7 @@ import {CopyNumberSeg} from "shared/api/generated/CBioPortalAPI";
 import {TrackProps} from "shared/components/igv/IntegrativeGenomicsViewer";
 
 export const WHOLE_GENOME = "all";
+export const DEFAULT_GENOME = "hg19";
```
adamabeshouse: if the reference genome is a page wide setting, I don't think this should go in IGVUtils. It should go in ResultsViewPageStoreUtils

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -57,6 +58,12 @@ export default class CopyNumberTableWrapper extends React.Component<ICopyNumberT
         enableCivic: false
     };
 
+    getCytobandForGene(hugoGeneSymbol:string) {
+        const refGene =  _.find(this.props.referenceGenes,(gene:ReferenceGenomeGene)=>
```
adamabeshouse: this is a linear search through props.referenceGenes every time you want to call this method, which is inefficient and could become a bit slow for a query with a lot of rows and a lot of genes.

Instead,

```
@computed get hugoGeneSymbolToCytoband() {
    if (this.props.referenceGenes) {
        return this.props.referenceGenes.reduce((map, next)=>{
            map[next.hugoGeneSymbol] = next.cytoband;
        });
    } else {
        return {};
    }
}
```

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -14,6 +14,8 @@ interface TracksPropTypes {
     width:number;
 }
 
+export const DEFAULT_CHROMOSOME="GRCh37";
```
adamabeshouse: just want to confirm this belongs in `Tracks`, so it is a `Tracks`-specific constant that's not used elsewhere. otherwise, it should go in a utils file that's more generally used.

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +58,33 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
-    sel.perc = getChmEndsPerc(sel.hg19,sel.total);
+export const genomeBuilds: Map<string, string> = new Map([
+    [ "hg19", "GRCh37" ],
+    [ "37", "GRCh37" ],
+    [ "hg38", "GRCh38" ],
+    [ "38", "GRCh38" ],
+    [ "mm10", "GRCm38" ]
+]);
+
+export type chromosomeSizeType = {
```
adamabeshouse: generally it's not good to name a type using the word "type". also, types should be capitalized. would `ChromosomeSizes` work?

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +58,33 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    build: string,
+    size: number[]
+};
+
+export function getChmInfo(genomeBuild:string) {
```
adamabeshouse: as in a comment above, it would be more efficient just to build a map genomeBuild->chromosome info, then index into that map, rather than calling a function.

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +58,33 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    const sel: any = {genomeRef:{}, total:0};
+    const chromSizes:chromosomeSizeType[] = chromosomeSizes;
+    let referenceGenome = genomeBuilds.get(genomeBuild);
+    if (!referenceGenome || referenceGenome === "") {
+        referenceGenome = DEFAULT_CHROMOSOME;
+    }
+    const genomeSize = chromSizes.find(c => c.build === referenceGenome);
```
adamabeshouse: better to build a map: genomeBuild->genomeSize. then index into that. otherwise, every time you make this function call, you're doing a linear time scan of chromosomeSizes.json.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -305,7 +305,7 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                             (store.studies.isComplete && store.sampleLists.isComplete && store.samples.isComplete) &&
                             (<Network genes={store.genes.result!}
                                       profileIds={store.rvQuery.selectedMolecularProfileIds}
-                                      cancerStudyId={store.studies.result[0].studyId}
+                                      cancerStudyId={(store.studies.result.length > 0) ?store.studies.result![0].studyId:""}
```
adamabeshouse: here and everywhere else, ternary spaces should be consistent. just looks more readable.
either `a?b:c` or `a ? b : c`, I think the second one is better. but not `a ?b:c`

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: aaron is right, uniqueReferenceGenes contains chromosomes. it should be called uniqueReferenceGeneChromosomes or something like that

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -795,14 +797,14 @@ export class ResultsViewPageStore {
...
-                    });
-
-                    return map;
-                }, {}));
+            if (this.referenceGenes.result) {
+                const uniqueReferenceGenes = _.uniq(this.referenceGenes.result.map(g => g.chromosome));
+                return Promise.resolve(uniqueReferenceGenes.reduce(
+                    (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
+                        map[chromosome] = remoteData<CopyNumberSeg[]> ({
+                            invoke: () => fetchCopyNumberSegmentsForSamples(this.samples.result, chromosome)});
```
adamabeshouse: please put the` });` on the next line, it looks cleaner

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2190,6 +2191,15 @@ export class ResultsViewPageStore {
         return parsedQString.query.session_id;
     }
 
+    private getCytoband(hugoGeneSymbol:string):string {
```
adamabeshouse: should be `readonly referenceGeneHugoSymbolToCytoband =  remoteData`. need to avoid these linear time scans in function calls when they can be precomputed once into a map and then constant time lookup every time afterwards.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: we do this everywhere. it is unpleasant but... it is

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -51,9 +51,15 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
         return generateSegmentFeatures(segments);
     }
 
+    @computed get genome() {
+        const study = this.props.store.studies.result?
+            this.props.store.studies.result[0]: undefined;
```
adamabeshouse: pull out and put in the store, as discussed below in CNSegmentsTab

khzhu:
I have done my corrections. All comments have been addressed. For the very few ones not corrected I explained why I decided to keep them the way they are. 
Please let me know anything else I need to do to get your approval. Thanks again for those informative and detailed suggestions!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2267,6 +2270,50 @@ export class ResultsViewPageStore {
...
+            const mutGenes = this.mutationEnrichmentData.result!.map(
+                                    (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            const cnvGenes = this.copyNumberAmpEnrichmentData.result!.map(
+                                        (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            return fetchReferenceGenomeGenes(this.studies.result[0].referenceGenome,
+                                             this.hugoGeneSymbols.concat(queryGenes, mutGenes, cnvGenes));
+        }
+    });
+
+    @computed get hugoGeneSymbolToCytoband() {
```
adamabeshouse: (1) After discussion over time, @alisman and I have determined that in cases like this (taking this as an example): it should be a remoteData which awaits this.referenceGenes, rather than a computed which checks the referenceGenes status.
(2) Isn't the name unclear as written, because it does not distinguish from the cytoband of a gene not on the selected reference genome? I think it should have the word "reference gene" in there somewhere
(3) `hugoGeneSymbolToCytoband` and `hugoGeneSymbolToChromosome` are redundant - should just be `hugoGeneSymbolToReferenceGene`, then users can do `hugo...Gene[gene].cytoband` or `.chromosome`

`referenceGeneHugoGeneSymbolToCytoband` is better, to distinguish from the query genes that are not from the selected reference genome. Also, please make this a `remoteData` which awaits `this.referenceGenes`. Please also make `hugoGeneSymbolToChromosome`

khzhu:
@alisman , I've pushed all changes (done so far) back to the repo for you to take a look. Please let me know if anything else you need from me. Thanks again for the help!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -795,14 +798,15 @@ export class ResultsViewPageStore {
...
         invoke: () => {
-            if (this.genes.result) {
-                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
-                    map[chromosome] = remoteData<CopyNumberSeg[]> ({
-                        invoke: () => fetchCopyNumberSegmentsForSamples(this.samples.result, chromosome)
-                    });
-
-                    return map;
-                }, {}));
+            if (this.referenceGenes.result) {
```
adamabeshouse: you need to await referenceGenes if you use it

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2267,6 +2270,93 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly referenceGenes = remoteData<ReferenceGenomeGene[]>({
+        await: ()=>[
+            this.studies,
+            this._mutationEnrichmentDataWithoutCytoband,
+            this._copyNumberHomdelEnrichmentData,
+            this._copyNumberAmpEnrichmentData,
```
kalletlak: this would be fetching both mutation and cna enrichments data together. this might cause delay in loading the page that is viewed. @adamabeshouse any suggestions?

adamabeshouse: good point. one way to solve it is fetch and deal with mutation enrichment reference genes separately from cna enrichment reference genes.

khzhu:
new comments from @adamabeshouse @kalletlak have been addressed. Please let me know anything else I might have missed. Thank you all again for those cool tips/suggestions!

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +58,40 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+]);
+
+export type ChromosomeSizes = {
+    genomeBuild: string,
+    chromosomeSize: number[]
+};
+
+const chromSizes:ChromosomeSizes[] = chromosomeSizes;
+
+const referenceGenomeSizes:{[genomeBuild:string]:number[]} =
```
adamabeshouse: (1) whats the point of having chromSizes when all it does is assign a reference to chromosomeSizes?
(2) this is easier to read as _.keyBy(chromosomeSizes, s=>s.genomeBuild)

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -56,11 +58,40 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
+    [ "38", "GRCh38" ],
+    [ "mm10", "GRCm38" ]
+]);
+
+export type ChromosomeSizes = {
+    genomeBuild: string,
+    chromosomeSize: number[]
+};
+
+const chromSizes:ChromosomeSizes[] = chromosomeSizes;
```
adamabeshouse: (1) whats the point of having chromSizes when all it does is assign a reference to chromosomeSizes?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2267,6 +2271,81 @@ export class ResultsViewPageStore {
...
+        }
+    });
+
+    readonly hugoGeneSymbolToReferenceGene = remoteData({
+        await: ()=>[
+            this.referenceGenes
+        ],
+        invoke: ()=>{
+            // build reference gene map
+            const result:{[hugosymbol:string]:ReferenceGenomeGene} =
```
adamabeshouse: theres no need to save `result`, just return `Promise.resolve(_.keyBy(...))`

if there's typing issues, you can annotate the type above:
`remoteData<{[hugoSymbol:string]:ReferenceGenomeGene}>({ ...`

src/pages/resultsView/cnSegments/CNSegments.tsx
```diff
@@ -52,10 +52,9 @@ export default class CNSegments extends React.Component<{ store: ResultsViewPage
     }
 
     @computed get chromosome() {
-        const gene = this.props.store.genes.result ?
-            this.props.store.genes.result.find(g => g.hugoGeneSymbol === this.activeLocus) : undefined;
-
-        return gene ? gene.chromosome : undefined;
+        if (this.props.store.hugoGeneSymbolToReferenceGene.result) {
```
adamabeshouse: generally in this case we would want to use a remoteData - await hugoGeneSymbolToReferenceGene and then use it. unless it genuinely doesnt matter and `undefined` is an equally valid value. In the latter case, you would still need to `return undefined` explicitly

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -48,15 +50,27 @@ type ICopyNumberTableWrapperProps = {
...
 export default class CopyNumberTableWrapper extends React.Component<ICopyNumberTableWrapperProps, {}> {
-
     public static defaultProps = {
         enableOncoKb: true,
         enableCivic: false
     };
 
+    @computed get hugoGeneSymbolToCytoband() {
+        // build reference gene map
+        const result:{[hugosymbol:string]:string} =
```
adamabeshouse: you should do a check for referenceGenes to be defined here, otherwise could end up with a bug. Unless referenceGenes is mandatory, in which case remove the `?` from the prop.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -792,17 +795,19 @@ export class ResultsViewPageStore {
...
         invoke: () => {
-            if (this.genes.result) {
-                return Promise.resolve(_.reduce(_.uniq(this.genes.result.map(g => g.chromosome)), (map: { [chromosome: string]: MobxPromise<CopyNumberSeg[]> }, chromosome: string) => {
-                    map[chromosome] = remoteData<CopyNumberSeg[]> ({
-                        invoke: () => fetchCopyNumberSegmentsForSamples(this.samples.result, chromosome)
-                    });
-
-                    return map;
-                }, {}));
+            if (this.referenceGenes.result) {
```
adamabeshouse: if you're awaiting referenceGenes, then `result` will always be there. no need to check

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2891,7 +2972,7 @@ export class ResultsViewPageStore {
         invoke: async () => pickCopyNumberEnrichmentProfiles(this.molecularProfilesInStudies.result!)
     });
 
-    readonly copyNumberHomdelEnrichmentData = makeEnrichmentDataPromise({
+    readonly _copyNumberHomdelEnrichmentData = makeEnrichmentDataPromise({
```
adamabeshouse: WithoutCytoband, for clarity and coherence

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2902,7 +2983,7 @@ export class ResultsViewPageStore {
         }
     );
 
-    readonly copyNumberAmpEnrichmentData = makeEnrichmentDataPromise({
+    readonly _copyNumberAmpEnrichmentData = makeEnrichmentDataPromise({
```
adamabeshouse: WithoutCytoband

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -397,6 +399,22 @@ export default class GroupComparisonStore {
...
+            this.mutationEnrichmentData,
+            this.copyNumberData
+        ],
+        invoke: async () => {
+            const mutGenes = this.mutationEnrichmentData.result!.map(
+                (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            const cnvGenes = this.copyNumberData.result!.map(
+                (a:AlterationEnrichment)=>a.hugoGeneSymbol.toUpperCase());
+            return fetchReferenceGenomeGenes(this.studies.result[0].referenceGenome,
+                mutGenes.concat(cnvGenes));
```
adamabeshouse: seems like you have to do the same thing here, right? or else the cytoband column will be empty in group comparison

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2267,6 +2271,81 @@ export class ResultsViewPageStore {
...
+            // build reference gene map
+            const result:{[hugosymbol:string]:ReferenceGenomeGene} =
+                _.keyBy(this.referenceGenes.result!, g=>g.hugoGeneSymbol);
+            return Promise.resolve(result);
+        }
+    });
+
+    readonly mutationEnrichmentData  = remoteData({
+        await: ()=>[
+            this._mutationEnrichmentDataWithoutCytoband,
```
adamabeshouse: since you're now just doing `fetchAllReferenceGenomeGenes`, theres no need to do this thing anymore where we calculate the cytoband afterwards. see what I mean? so you can just put it back to the way you had it before, where you add the cytoband inside makeEnrichmentsDataPromise

src/pages/groupComparison/GroupComparisonStore.ts
adamabeshouse: nvm: see my comment https://github.com/cBioPortal/cbioportal-frontend/pull/2180/files#r310723393

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: nvm: see my comment here https://github.com/cBioPortal/cbioportal-frontend/pull/2180/files#r310723393

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: nvm: see my comment here https://github.com/cBioPortal/cbioportal-frontend/pull/2180/files#r310723393

PR #2181, opened 2019-04-01
oncoprint loader - make it not fixed position

adamabeshouse:


src/shared/components/loadingIndicator/styles.module.scss
```diff
@@ -12,6 +12,15 @@
 
 }
 
+.centered-relative {
```
alisman: i'm not wild about a class with relative in the name having a position:absolute.  
I think the solution to this problem might be to target the loader at the oncoprint DIV so that it gets positioned relative to it instead of the tab.

PR #2183, opened 2019-04-01
plots tab - add back data source info icon

adamabeshouse:


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1041,7 +1054,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                     </div>
                     <div className="form-group">
                         <label>{dataSourceLabel}</label>
-                        <div style={{display:"flex", flexDirection:"row"}}>
+                        <div style={{display:"flex", flexDirection:"row", alignItems:"center"}}>
```
alisman: default flex-direction is already "row"

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1051,6 +1064,13 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                                 clearable={false}
                                 searchable={true}
                             />
+                            { dataSourceDescription && (
```
alisman: just want to confirm that it in fact makes sense to render the menu even before dataSourceDescription has loaded and then re-render it when it loads.

adamabeshouse: fixed

PR #2188, opened 2019-04-03
Update karma and webpack config to support IDE debugging

pvannierop:
# What? Why?
For development of unit tests it is effective to be able to attach debugging software and step through the code. Preferably this is done in the IDE used to create the code. This PR will fix exports of source maps by webpack and introduce a yarn script that runs the tests in a browser that allows attachment of remote debugging software.

# Tutorial
## 1. Start karma in debug mode
To debug a single test file (for instance `DeltaUtils.spec.ts`)
- Start the yarn debug browser
```
yarn test:debug -- --grep=DeltaUtils
```
This opens the tests on the console and leaves the test in _watch mode_. _Watch mode_ makes the tests run automatically at the moment any of the files under test are updated/saved.
Screenshot from 2019-04-03 14-35-37_1

And opens a browser window that runs the karma tests:

- Press the `DEBUG` button at the top right corner of the karma browser window. A second tab will open:

## 2. Attach debugging software
Debugging can be performed with the integrated DevTools of the Chrome browser or by attaching external debugging software. Both approaches are explained below.

### Google Chrome DevTools
- Open DevTools by pressing `ctr-shift-I` key combination. This opens a new window in the debug tab.
- Navigate to the source code file of choice. Source code files can be found at the `webpack://` entry on the `Page` tab. For instance, `DeltaUtils.spec.ts` can be found at path `webpack://./src/shared/components/oncoprint/DeltaUtils.spec.ts`:

- Place a break point by clicking the line number in front of the line of code. A blue arrow will appear:

Screenshot from 2019-04-03 15-05-40

- Restart the test by refreshing the browser page (hit `F5` key or press refresh icon). When the break point is reached execution stops. Variables can be inspected in the `Scope` tab.

Screenshot from 2019-04-03 15-07-31

Note: when the code is updated (for instance in a text editor) and the karma test is restarted automatically, the karma browser window must be refreshed manually to see the code changes in the webpack://` section.

### External debugging software.
The new configuration allows to attach debugging software on port 9333. Different IDE's require different ways this can be achieved.

#### VSCode
- Install the `Debugger for Chrome` [VSCode extension](https://code.visualstudio.com/blogs/2016/02/23/introducing-chrome-debugger-for-vs-code).
- Add the following configuration entry to `launch.json`: 
```
        {
            "type": "chrome",
            "request": "attach",
            "name": "Attach to Karma Chrome",
            "address": "localhost",
            "port": 9333,
            "pathMapping": {
                "/": "${workspaceRoot}/",
                "/base/": "${workspaceRoot}/"
            },
            "webRoot": "${workspaceFolder}",
            "sourceMapPathOverrides": {
                "webpack:///*": "${webRoot}/*"
            }
        }
```
- Start the debugger by pressing `F5` and attach to `Karma DEBUG RUNNER` tab of the karma browser.


webpack.config.js
```diff
@@ -366,7 +366,7 @@ if (isDev) {
 
 if (isDev || isTest) {
 
-    config.devtool = 'source-map';
+    config.devtool = 'inline-source-map';
```
alisman: can you check to see how performance of build is in DEV mode?  my worry is that this (without grepping) will produce gargantuan bundle which browser debugger may have trouble with.     Obviously we aren't grepping when we run all tests on CI also, so  it could impact that too.  is there a reason to do inline?

alisman:
@pvannierop my concern about inline-sourcemap comes from: 
https://webpack.js.org/configuration/devtool/
you can see it's listed as slowest of all sourcemap settings.

webpack.config.js
pvannierop: Ok, it all seems to work and there are no performance issues on my computer. I do get an error message in the karma browser that I did not observe before. These errors also appear when running the `karma test:watch` command. Tests and debugging seem to work though. 

Screenshot from 2019-04-03 16-26-04

pvannierop: I will investigate other source map options and report back.

karma.conf.js
```diff
@@ -25,7 +25,7 @@ webpackConfig.module.rules.push({
 
 //this will be used by in test context to load corresponding spec files if there is a grep passed (or all if not)
 webpackConfig.plugins.push(new webpack.DefinePlugin({
-    'SPEC_REGEXP': ('grep' in argv) ? `/(global|(${argv.grep}[a-z]*))\.spec\./i` : '/\.spec\.(jsx*|tsx*)$/'
+    'SPEC_REGEXP': ('grep' in argv) ? `/(global|(${argv.grep}[a-z]*))\.spec\./i` : '/\.spec\.(jsx|tsx|ts)$/'
```
adamabeshouse: you're missing `js` now

webpack.config.js
```diff
@@ -147,7 +147,7 @@ var config = {
     module: {
         rules: [
             {
-                test: /\.tsx?$/,
+                test: /\.(tsx|ts)?$/,
```
adamabeshouse: the question mark doesnt make sense anymore I think..

webpack.config.js
pvannierop: A new way of using inline-source-map has been implemented. The inline-source-map is now activated via the `--debug` flag passed to karma. This way the inline-source-map is not used when testing on circleCI.

karma.conf.js
pvannierop: js was added.

webpack.config.js
pvannierop: Change was reverted.

webpack.config.js
```diff
@@ -487,13 +487,6 @@ if (isDev || isTest) {
             'process.env': {
                 'NODE_ENV': `"${process.env.NODE_ENV || 'production'}"`
             }
-        }),
-        new webpack.optimize.UglifyJsPlugin({
```
alisman: this is removing uglification in production build, is it not?  it's possible btw that e2e doesn't build in "test" mode.  but in that case we need to fix that.

pvannierop: Ah, I accidentally committed this change to the branch. It solves the 1hr build time for the frontend on my machine and jitpack. I have reverted these lines back to rc state.  This problems still needs to be solved though...

PR #2195, opened 2019-04-04
Exposed certain components as cbioportal-frontend-commons module

onursumer:
- Moved certain shared libraries under `src/public-lib` to be able to create a separate `cbioportal-frontend-commons` bundle.
- Added separate webpack bundler
- Added separate tsconfig file to target only the `src/commons` folder, and generate types.
- Moved `LollipopMutationPlot` and `TrackPanel` components (together with related sub-components and utilities) to a separate repository: https://github.com/cbioportal/react-mutation-mapper
- Added dependency to `react-mutation-mapper` package
- Added `buildPublicLib` script into `package.json`

# Checks
  the `-s` option)


package.json
```diff
@@ -41,7 +43,7 @@
   "author": "",
   "license": "",
   "dependencies": {
-    "3dmol": "1.1.4",
+    "3dmol": "^1.3.7",
```
adamabeshouse: is this on purpose? because this caused some issues when I tried to update it, which is why I removed the "^"

src/pages/patientView/vafPlot/VAFPlot.tsx
```diff
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import $ from 'jquery';
+// @ts-ignore
```
adamabeshouse: can you add a comment explaining why

adamabeshouse:
whats the purpose of this?

src/pages/patientView/vafPlot/VAFPlot.tsx
onursumer: Good catch! I was experimenting with tsconfig options when I was trying to expose the whole `shared` directory as is. We don't need to modify this file at this point, I'll revert it.

package.json
```diff
@@ -273,12 +286,25 @@
     "mocha": "^5.2.0",
     "phantomjs-polyfill": "^0.0.2",
     "phantomjs-prebuilt": "^2.1.16",
+    "postcss-simple-vars": "^5.0.2",
     "request": "^2.88.0",
+    "rollup": "^1.6.0",
```
alisman: we should kill this right?

package.json
onursumer: The latest version of `3dmol` allows using `import` instead of `require`. But I can skip this update since this is not required at this point.

adamabeshouse: hey if it works it works, I just found that it was breaking the install step on e2e

package.json
onursumer: yes, I'll cleanup rollup config & dependencies. let's stick to webpack for now.

.gitignore
```diff
@@ -11,7 +12,9 @@ stats.json
 tsDist/
 dist/
 common-dist/
+commons-dist/
```
inodb: slightly confusing to have both commons-dist and common-dist which are completely different things. Maybe we should rename common-dist to something else? Like vendor-bundles-dist?

commons.webpack.config.js
```diff
@@ -0,0 +1,392 @@
+var ExtractTextPlugin = require('extract-text-webpack-plugin');
```
inodb: this looks like it might have quite some overlap with `webpack.config.js`. Maybe certain things can be imported from there? Or `webpack.config.js` can be adjusted with a few environment variables to output the common-dist artifacts?

onursumer: for now, I just copy & paste from `webpack.config.js`, but ideally we should be able to share the overlapping content.

onursumer:
Rebased this PR onto `release-3.0.0` branch to get the latest webpack configuration changes. Do not merge this PR yet. Rebase again on master after the `release-3.0.0` is merged into master.

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -33,24 +30,22 @@ import {uniqueGenomicLocations, genomicLocationString, groupMutationsByProteinSt
 import {defaultHotspotFilter, groupHotspotsByMutations} from "shared/lib/CancerHotspotsUtils";
 import {defaultOncoKbIndicatorFilter, groupOncoKbIndicatorDataByMutations} from "shared/lib/OncoKbUtils";
 import { getMutationsToTranscriptId } from "shared/lib/MutationAnnotator";
-import Mutations from "pages/resultsView/mutation/Mutations";
```
onursumer: This should've never imported here, this was actually a bug.

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -382,10 +295,10 @@ export default class MutationMapperStore
     }
 
     @computed get transcriptsByTranscriptId(): {[transcriptId:string]: EnsemblTranscript} {
-        return _.keyBy(this.allTranscripts.result, ((transcript: EnsemblTranscript) => transcript.transcriptId));
+        return _.keyBy(this.allTranscripts.result as EnsemblTranscript[], (transcript  => transcript.transcriptId));
     }
 
-    @computed get mutationsByTranscriptId(): {[transcriptId:string]: Mutations[]} {
+    @computed get mutationsByTranscriptId(): {[transcriptId:string]: Mutation[]} {
```
onursumer: The correct return type for this function is `Mutation[]`, not `Mutations[]`, `Mutations` is a UI component...

src/shared/components/structureViewer/StructureVisualizer3D.ts
```diff
@@ -9,8 +9,7 @@ import {
     IResidueHelper, IResidueSelector, generateResiduePosToSelectorMap, findUpdatedResidues
 } from "./PdbResidueUtils";
 
-// 3Dmol expects "this" to be the global context
-const $3Dmol = require('imports-loader?this=>window!3dmol/build/3Dmol-nojquery.js');
+import $3Dmol from '3dmol';
```
onursumer: updated to the latest version, so that we don't need to use `require`

src/shared/components/annotation/CancerHotspots.tsx
```diff
@@ -38,89 +37,6 @@ export default class CancerHotspots extends React.Component<ICancerHotspotsProps
         return score;
     }
 
-    public static hotspotInfo(isHotspot:boolean, is3dHotspot:boolean, count?: number, customInfo?: JSX.Element)
```
onursumer: contents moved into `HotspotInfo` component under `react-mutation-mapper`

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -130,141 +154,6 @@ export default class MutationMapperStore
         }
     }, []);
 
-    readonly ptmData = remoteData({
```
onursumer: Some of these read-only remote data moved into `DefaultMutationMapperStore`

src/shared/components/mutationMapper/MutationMapperUserSelectionStore.ts
```diff
@@ -1,14 +1,5 @@
 import {observable} from "mobx";
-import {TrackNames, TrackVisibility} from "../tracks/TrackSelector";
-
-export function initDefaultTrackVisibility(): TrackVisibility {
```
onursumer: function moved into `react-mutation-mapper`

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -1,97 +1,19 @@
...
 export default class ProteinChangeColumnFormatter
 {
     public static getSortValue(d:Mutation[]):number|null {
-        return ProteinChangeColumnFormatter.extractSortValue(
+        return calcProteinChangeSortValue(
             ProteinChangeColumnFormatter.getTextValue(d)
         );
     }
 
-    // this is to sort alphabetically
```
onursumer: functions moved into `react-mutation-mapper`

src/shared/lib/StoreUtils.ts
```diff
@@ -98,65 +98,6 @@ export async function fetchPdbAlignmentData(ensemblId: string,
     }
 }
 
-export function fetchPtmData(ensemblId: string,
```
onursumer: functions moved into `react-mutation-mapper`

PR #2197, opened 2019-04-05
add query button

dippindots:
Add query button and add the unsampled download on the download tab.
Fix https://github.com/cBioPortal/cbioportal/issues/5665.

- [x] change "Sample affected" to "Altered samples"
- [x] Add "Query" after "Download" to query the altered samples with the same genes.
- [x] Add a new line "Unaltered samples: ..." with the same options.
- [x] Update download tab texts(Altered samples, Unaltered samples, and Sample matrix)
- [x] When the user clicked the "Query" button. Open "Modify Query" interface and change the querying sample list
- [x] Show a highlight alert "altered/ unaltered samples has been added" to tell the user it's changed.
- [x] Add a button to open the virtual study popup.

image


src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -204,6 +206,18 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
         invoke:()=>Promise.resolve(this.alteredSamples.result!.join("\n"))
     });
 
+    readonly unalteredSamples = remoteData<string[]>({
```
alisman: Give this a name that suggests what it returns.  I would guess Samples, but it actually returns sampleIds.   What I would actually do here is return the sample objects and then do the mapping to composite ID in business logic.

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -204,6 +206,18 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+        await:()=>[this.caseAlterationData],
+        invoke:()=>Promise.resolve(this.caseAlterationData.result!
+            .filter(caseAlteration => !caseAlteration.altered)
+            .map(caseAlteration => `${caseAlteration.studyId}:${caseAlteration.sampleId}`))
+    });
+
+    readonly unalteredSamplesText = remoteData<string>({
+        await: ()=>[this.unalteredSamples],
+        invoke:()=>Promise.resolve(this.unalteredSamples.result!.join("\n"))
```
alisman: again, just do this in logic.  i don't think there's a strong reason to encapsulate this simple operation inside a promise

src/pages/resultsView/download/DownloadTab.tsx
dippindots: Sure, will do!

dippindots:
@jjgao @tmazor Could you give this pr a review from the production side? The Heroku link is here: https://cbioportal-frontend-pr-2197.herokuapp.com/. Thanks!

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -373,14 +378,57 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+        );
+    }
+
     readonly alteredSamplesDownloadControls = remoteData({
-        await:()=>[this.alteredSamplesText],
+        await:()=>[this.alteredCaseAlterationData],
+        invoke:()=>{
+            const alteredSampleIds = _.map(this.alteredCaseAlterationData.result!, caseAlteration => `${caseAlteration.studyId}:${caseAlteration.sampleId}`);
+            const handleDownload = () => alteredSampleIds.join("\n");
+            const handleQuery = () => this.handleQueryButtonClick(alteredSampleIds.join("\n"));
```
alisman: would be nice if did not serialize these to strong until we actually need to.   like, the store property that holds these should be an array, not a string

alisman: same with alteredSampleIds.   Someone reading code would think this is an array

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -373,14 +378,57 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+            const handleVisualStudy = () => this.handleVisualStudyButtonClick(this.alteredCaseAlterationData.result!);
+
+            return Promise.resolve(this.copyDownloadQueryControlsRow("Altered samples: List of samples with alterations",
```
alisman: We could probably use adam's MobxPromiseComponent for this type of thing.

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -434,4 +482,32 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+        let studies = _.reduce(_.groupBy(alterations, "studyId"), (acc: { id: string; samples: string[] }[], alteration, studyId) => {
+            acc.push({
+                id: studyId,
+                samples: alterations.map(alteration => alteration.sampleId)
+            })
+            return acc;
+        }, []);
+
+        let parameters = {
+            name: "visual study",
```
alisman: @jjgao if we're going to create a virtual study ("visual study" is an error), what should it be named?  Maybe we can grab a name from the lable we are using for this row in table @dippindots

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -434,4 +482,32 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+            description: "visual study",
+            origin: studies.map(study => study.id),
+            studies: studies
+        }
+        const saveStudyResult = await sessionServiceClient.saveVirtualStudy(parameters, false);
+        if (saveStudyResult && saveStudyResult.id) {
+            window.open(buildCBioPortalPageUrl({pathname:'study', query: {id: saveStudyResult.id}}), "_blank");
```
dippindots: @alisman Just talked with @jjgao in person. Rather than create a virtual study, we can just open a dropdown as same as the one in study view page.
image
So we can let the user decide to save/share the virtual study or just open it.

src/shared/components/copyDownloadControls/CopyDownloadQueryLinks.tsx
```diff
@@ -0,0 +1,75 @@
...
+            <span className={this.props.className}>
+                {
+                    this.props.showCopy &&
+                    <a onClick={this.props.handleCopy} ref={this.props.copyLinkRef}>
+                        <i className='fa fa-clipboard' style={{marginRight: 5}}/>{this.props.copyLabel}
+                    </a>
+                }
+                {
+                    this.props.showCopy && this.props.showDownload &&
+                    <span style={{margin: '0px 10px'}}>|</span>
```
alisman: not a big deal but would be nicer to do this with css.   you could put a right border on all but the last <a> element in the list

src/pages/resultsView/download/DownloadTab.tsx
alisman: cool!

src/shared/components/copyDownloadControls/CopyDownloadQueryLinks.tsx
dippindots: @alisman Thanks for your suggestion! I will try to address them all.

alisman:
@dippindots we need to improve how this alert looks.  really it should go inside the query component.  that might be hard.  maybe we can give query container a new property "alertMessages" and pass in components we want it to display in an alert area.
image

dippindots:
@alisman That's a good idea! I will definitely try to do this.

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -150,6 +154,18 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
         return this.props.store.genesInvalid || this.props.store.isQueryInvalid;
     }
 
+    @computed get getQueryStore() {
+        let queryStore = createQueryStore(getBrowserWindow().routingStore.query);
+        if (this.props.store.selectedSampleIdsForQueryButton) {
+            queryStore.caseIds = this.props.store.selectedSampleIdsForQueryButton;
```
alisman: i don't think we should store the info for the query in the resultsViewStore.  It's a kind of transient state.  Like a message.

alisman:
@jjgao after reviewing this a little more closely, I'm not sure the "query" and "virtual study" buttons belong in the download tab of results view.  These seem more like "filter states" of the results view.     I don't think it's intuitive that a user would look in download tab to find these.  Perhaps they could be part of the new interface I am working on that is meant to be home to such "global" query modification UI.

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -26,15 +27,18 @@ import {buildCBioPortalPageUrl} from "../../../shared/api/urls";
 export default class QuerySummary extends React.Component<{ routingStore:ExtendedRouterStore, store: ResultsViewPageStore, onToggleQueryFormVisiblity:(visible:boolean)=>void }, {}> {
 
     @autobind
+    @action
     private toggleQueryFormVisibility() {
-        this.props.onToggleQueryFormVisiblity(this._queryFormVisible);
-        this._queryFormVisible = !this._queryFormVisible;
+        this.props.onToggleQueryFormVisiblity(this.props.store.queryFormVisible);
+        // if clicked the query button in the download tab and want to close the query form, clear the selected sample ids
+        if (this.props.store.selectedSampleIdsForQueryButton && this.props.store.queryFormVisible === true) {
```
alisman: i think what bothers me about this is that this data belongs to a very specific part of the results view: the modify query view.  and really only to once instance of it.   How about we keep the data not on the resultsPageStore but on this view itself (QuerySummary).   Furthermore we can nest it inside of a modifyQueryParams object to make sure it's super clear what it is.   Then, you can just clear the whole thing by setting that property to undefined.  Make sense?

alisman:
@dippindots maybe we should go ahead and continue to work on this given that we do want the functionality wherever we end up deciding the button should go.  it may go in many places.

jjgao:
@alisman It seems like an intuitive place to put the two buttons if you think about it as a new session instead of filtering the current session. I think we should move forward with this design. If we can find another place to place the button later, we can do both.

src/pages/resultsView/querySummary/QuerySummary.tsx
dippindots: Yes!

alisman: ok, i understand your confusion.  you are right that the data needs to live at a global level (resultsPageStore).  So lets make a new property of resultsPageStore called modifyQueryParams and nest all necessary params there.   Then we just need to figure out a way to delete that object.   Lets discuss by voice.

PR #2200, opened 2019-04-05
Update patient view screenshot

alisman:
Data update


alisman:
@jjgao this fixes patient view screenshot which reflected old data.  there are three failing screenshots on this PR but they are obviously unrelated to this

PR #2201, opened 2019-04-08
Updating Tables to show number of genes profiled in tooltip

lemccarthy:
# What? Why?
Fix # .
Fix cBioPortal/cbioportal#5851

    Adds a tooltip to the MutatedGenesTable and CNAGenesTable that shows the total number of genes profiled when the user hovers over them.
    Change the API to fetch the number of genes profiled instead of the frequency.

### DIFF
--- a/src/pages/studyView/table/CNAGenesTable.tsx
+++ b/src/pages/studyView/table/CNAGenesTable.tsx
@@ -1,12 +1,12 @@
 import * as React from "react";
 import * as _ from "lodash";
-import {CNAGenesData, CopyNumberAlterationIdentifier} from "pages/studyView/StudyViewPageStore";
-import {action, computed, IReactionDisposer, observable, reaction} from "mobx";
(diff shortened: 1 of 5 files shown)


src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -65,20 +65,20 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
         [ColumnKey.FREQ]: 0,
     };
 
-    private reactions:IReactionDisposer[] = [];
+    private reactions: IReactionDisposer[] = [];
```
alisman: looks like maybe you have a code formatter that adds spaces?

lemccarthy: Yes, I typically do use a code formatter, do we have existing linting/formatting settings that are project wise? I can use those instead if we have them.

PR #2205, opened 2019-04-09
reduce number of mutation in example mutation mapper dataset

alisman:
Large genome nexus queries from mutation mapper page are leading to test timeouts.  Reduced example data set should lead to less latency


alisman:
@onursumer wow good catch

PR #2207, opened 2019-04-09
organize/optimize e2e tests

alisman:
Break end to end specs into more sensible files to optimize for parallelization


end-to-end-tests/wdio/webdriver-manager.conf.js
```diff
@@ -272,3 +284,24 @@ exports.config = {
     // onComplete: function(exitCode) {
     // }
 };
+
+const doBrowserstack = false;
```
alisman: this is for running against browserstack when we test locally

end-to-end-tests/specs/oncoprinter.screenshot.spec.js
```diff
@@ -0,0 +1,169 @@
...
+
+        pasteToElement(".oncoprinterSamplesInput", sampleList);
```
alisman: we used to use setText.  it took >30secs to set a large block text because it does character by character.  so we should use this method whenever we have to input substantial amounts of text to a field

end-to-end-tests/package.json
```diff
@@ -6,13 +6,14 @@
   "scripts": {
     "test-travis": "rm -rf screenshots/diff/; wdio wdio/travis.conf.js || bash image-compare/upload_failed_screenshots.sh",
     "test-webdriver-manager": "rm -rf screenshots/diff/; cd .. && ./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cd end-to-end-tests && (wdio wdio/webdriver-manager.conf.js || bash image-compare/upload_failed_screenshots.sh)",
-    "e2e-local": "rm -rf screenshots/screen/; rm -rf screenshots/reference/; rm -rf screenshots/diff/; cd .. && ./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cd end-to-end-tests && (wdio wdio/webdriver-manager.conf.js || bash image-compare/upload_failed_screenshots.sh)"
-
+    "e2e": "yarn run clear && yarn run test-webdriver-manager",
+    "clear": "rm -rf screenshots/screen/; rm -rf screenshots/reference/; rm -rf errorShots/; rm -rf screenshots/diff/;"
```
adamabeshouse: whats the purpose of this?

end-to-end-tests/specs/specUtils.js
```diff
@@ -141,4 +149,7 @@ module.exports = {
     setOncoprintMutationsMenuOpen: setOncoprintMutationsMenuOpen,
     getNthOncoprintTrackOptionsElements: getNthOncoprintTrackOptionsElements,
     setInputText: setInputText,
+    pasteToElement: pasteToElement
```
adamabeshouse: whats the purpose of this?

end-to-end-tests/package.json
alisman: when you run tests locally, the screenshots are going to diff anyway, so it's sort of pointless to ever do a comparison against the real reference.

adamabeshouse: ah okay

adamabeshouse: while we're here do you want to delete `test-travis`

alisman: sure

PR #2208, opened 2019-04-09
mutex tab - remove significant badge

adamabeshouse:


src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.spec.tsx
```diff
@@ -395,8 +395,8 @@ describe("MutualExclusivityUtil", () => {
             assert.equal(cells.at(6).html(), "<td><span>&gt;3</span></td>");
             assert.equal(cells.at(7).html(), "<td><span>0.004</span></td>");
             assert.equal(cells.at(8).html(), "<td><b><span>0.024</span></b></td>");
-            assert.equal(cells.at(9).html().replace(/<!--[^>]*-->/g, ""), "<td><div class=\"styles-module__Tendency__2Y46X\">" + 
-                "Co-occurrence<span class=\"badge\" style=\"background-color: rgb(88, 172, 250);\">Significant</span></div></td>");
+            assert.equal(cells.at(9).html().replace(/<!--[^>]*-->/g, ""), "<td><div class=\"styles-module__Tendency__2Y46X styles-module__Significant__34YnP\">" +
```
alisman: i'm surprised that class names are actually stable.  Probably would be about the same work to use /fdsaf/.test() but not that important.

adamabeshouse: I had the same thought but went ahead with it because it's apparently worked until now

PR #2212, opened 2019-04-10
Sort mode of oncoprint set via URL parameter

pvannierop:
# What? Why?
Oncoprint samples can be sorted based on different properties such as case identifier, mutation types, etc. The property that controls sort order (a.k.a. _sort mode_) is controlled by the user via UI events in the `Sort` menu of the OncoprintControls component. In the current implementation oncoprint view cannot be opened with a predetermined sort mode that requires no UI actions by the user.

# Changes by this PR
For integration with tools external to cBioPortal we have created a new feature that allows to open oncoprint view with a specified sort mode via the URL. This PR will implement a new URL parameter `oncoprint_sortby` for oncoprint. This parameter can have 6 different values that control the different sort modes:

| param value  | sort mode |
|  :--- |  :--- |
| `case_id` | sort by sample/patient identifier |
| `case_list` | sort by sample/patient case list |
| `data` | sort by data |
| `data_mut` | sort by data and mutation type |
|`data_dri` | sort by data and driver/passenger mutation |
| `data_mut_dri` | sort by data, mutation type and driver/passenger mutation |

When the sort mode is updated by the user via an UI event, the URL is updated (w/o page refresh) to reflect this change so that the URL can used to open the same view.

The PR includes unit tests for the URL param processing by the ResultsViewOncoprint component.

# Notes
- Sort order by heatmap is not not covered by this PR.

### DIFF
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.spec.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.spec.tsx
@@ -1,0 +1,85 @@
+import {assert} from "chai";
+import ResultsViewOncoprint from "./ResultsViewOncoprint";
+import { ResultsViewPageStore } from "pages/resultsView/ResultsViewPageStore";
+import { Sample, Patient } from "shared/api/generated/CBioPortalAPI";
+import getBrowserWindow from "shared/lib/getBrowserWindow";
+import ExtendedRouterStore from "shared/lib/ExtendedRouterStore";
+import sinon from "sinon";
+import { isOccupied } from "pages/studyView/StudyViewUtils";
+import { SortByUrlParamValue } from 'shared/components/oncoprint/ResultsViewOncoprint';
+
+describe('Oncoprint sortBy URL parameter', () => {
+
+    const samples = [
+        {sampleId: "Sample1", uniqueSampleKey: "Sample1_key"},
+        {sampleId: "Sample2", uniqueSampleKey: "Sample2_key"},
+        {sampleId: "Sample3", uniqueSampleKey: "Sample3_key"}
+    ] as Sample[];
+
+    const patients = [
+        {patientId: "Patient1", uniquePatientKey: "Patient1_key"},
+        {patientId: "Patient2", uniquePatientKey: "Patient2_key"},
+        {patientId: "Patient3", uniquePatientKey: "Patient3_key"}
+    ] as Patient[];
+
+    const caseList = [
+        {sampleId: "Sample2", uniqueSampleKey: "Sample2_key", uniquePatientKey: "Patient2_key"},
+        {sampleId: "Sample3", uniqueSampleKey: "Sample3_key", uniquePatientKey: "Patient3_key"},
+        {sampleId: "Sample1", uniqueSampleKey: "Sample1_key", uniquePatientKey: "Patient1_key"}
+    ] as Sample[];
+
+    const storeMock = {
+        samples: {isComplete: true, result: samples},
+        patients: {isComplete: true, result: patients},
+        givenSampleOrder: {isComplete: true, result: caseList}
+    } as any as ResultsViewPageStore;
+
+    it('`case_id` provides sorted sample config to oncoprint', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam:SortByUrlParamValue.CASE_ID, columnMode: 'sample'});
+        assert.deepEqual( oncoprintView.sortConfig.order, ["Sample1_key","Sample2_key","Sample3_key"]);
+    });
+
+    it('`case_id` provides sorted patient config to oncoprint', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam:SortByUrlParamValue.CASE_ID, columnMode:'patient'});
+        assert.deepEqual( oncoprintView.sortConfig.order, ["Patient1_key","Patient2_key","Patient3_key"]);
+    });
+
+    it('`case_list` provides sorted sample config to oncoprint when case list is available', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam:SortByUrlParamValue.CASE_LIST, columnMode: 'sample', caselistEnabled: true});
+        assert.deepEqual(oncoprintView.sortConfig.order,  ["Sample2_key","Sample3_key","Sample1_key"]);
+    });
+
+    it('`case_list` provides sorted patient config to oncoprint when case list is available', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam:SortByUrlParamValue.CASE_LIST, columnMode: 'patient', caselistEnabled: true});
+        assert.deepEqual(oncoprintView.sortConfig.order,  ["Patient2_key","Patient3_key","Patient1_key"]);
+    });
+
+    it('`case_list` provides no sort config to oncoprint when case list is unavailable', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam:SortByUrlParamValue.CASE_LIST, columnMode: 'patient', caselistEnabled: false});
+        assert.isUndefined(oncoprintView.sortConfig.order);
+    });
+
+    interface IHelperFunction {
+        sortByParam:SortByUrlParamValue;
+        columnMode?:"sample"|"patient" ;
+        caselistEnabled?:boolean;
+    }
+
+    const initResultsViewWithSortByParam = (params:IHelperFunction) => {
+        // mock the url params by mocking the ExtendedRouterStore class
+        const routingStub = sinon.createStubInstance(ExtendedRouterStore);
+        routingStub.location = { query: {oncoprint_sortby: params.sortByParam}};
+        getBrowserWindow().globalStores = {routing: routingStub };
+        if (params.caselistEnabled !== undefined) {
+            storeMock.givenSampleOrder.isComplete = params.caselistEnabled;
+        }
+        const oncoprintView = new ResultsViewOncoprint(
+            {divId: "", store: storeMock, routing: ""}
+        );
+        if (params.columnMode !== undefined) {
+            oncoprintView.columnMode = params.columnMode;
+        }
+        return oncoprintView;
+    };
+
+});
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -35,6 +35,8 @@ import ProgressIndicator, {IProgressIndicatorItem} from "../progressIndicator/Pr
 import autobind from "autobind-decorator";
 import getBrowserWindow from "../../lib/getBrowserWindow";
 import MobxPromise from "mobxpromise";
+import {browserHistory} from "react-router";
+import * as JQuery from 'jquery';
 
 interface IResultsViewOncoprintProps {
     divId: string;
@@ -43,10 +45,17 @@ interface IResultsViewOncoprintProps {
     addOnBecomeVisibleListener?:(callback:()=>void)=>void;
 }
 
+export enum SortByUrlParamValue {
+    CASE_ID = "case_id",
+    CASE_LIST = "case_list",
+    NONE = ""
+}
+
 export type SortMode = (
     {type:"data"|"alphabetical"|"caseList", clusteredHeatmapProfile?:undefined} |
     {type:"heatmap", clusteredHeatmapProfile:string}
 );
+
 
 export interface IGenesetExpansionRecord {
     entrezGeneId: number;
@@ -58,6 +67,7 @@ export interface IGenesetExpansionRecord {
 export const SAMPLE_MODE_URL_PARAM = "show_samples";
 export const CLINICAL_TRACKS_URL_PARAM = "clinicallist";
 export const HEATMAP_TRACKS_URL_PARAM = "heatmap_track_groups";
+export const ONCOPRINT_SORTBY_URL_PARAM = "oncoprint_sortby";
 
 const CLINICAL_TRACK_KEY_PREFIX = "CLINICALTRACK_";
 
@@ -142,7 +152,6 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                 this.selectedClinicalAttributeIds.set(attr.clinicalAttributeId, true);
             }
         });
-        
         const self = this;
 
         this.onChangeSelectedClinicalTracks = this.onChangeSelectedClinicalTracks.bind(this);
@@ -168,6 +177,8 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         this.onMouseEnter = this.onMouseEnter.bind(this);
         this.onMouseLeave = this.onMouseLeave.bind(this);
 
+        // update URL parameters according to UI events
+        // and trigger a page refresh
         this.urlParamsReaction = reaction(
             ()=>[
                 this.columnMode,
@@ -225,7 +236,7 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                 return self.sortByMutationType;
             },
             get sortByCaseListDisabled() {
-                return !self.props.store.givenSampleOrder.isComplete || !self.props.store.givenSampleOrder.result.length;
+                return ! self.caseListSortPossible;
             },
             get distinguishMutationType() {
                 return self.distinguishMutationType;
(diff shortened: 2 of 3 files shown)


Sjoerd-van-Hagen:
@adamabeshouse hi Adam,

I assigned you for this one since you do most of the work on the oncoprint. When will you have time to review it? I don't expect it will take much time.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -191,6 +196,53 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
             }
         );
 
+        // update URL parameters according to sort direction UI events
+        // and DO NOT trigger a page refresh
+        this.urlParamsReaction = reaction(
```
adamabeshouse: please merge this one into the other one on line 176

src/shared/components/oncoprint/ResultsViewOncoprint.spec.tsx
```diff
@@ -0,0 +1,112 @@
...
+        assert.isTrue(oncoprintView.sortByDrivers);
+    });
+
+    it('`data_mut_dri` provides data+mutation+driver-based sort config to oncoprint', () => {
+        const oncoprintView = initResultsViewWithSortByParam({sortByParam: 'data_mut_dri'});
+        assert.isUndefined(oncoprintView.sortConfig.order);
+        assert.isTrue(oncoprintView.sortByMutationType);
+        assert.isTrue(oncoprintView.sortByDrivers);
+    });
+
```
adamabeshouse: tests look great

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -191,6 +196,53 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
+                this.sortByDrivers
+            ],
+            ()=>{
+                const newParams = Object.assign({}, getBrowserWindow().globalStores.routing.location.query);
+                if (this.sortMode) {
+                    const mode = this.sortMode.type;
+                    let sortby_param = "";
+                    switch (mode) {
+                        case "alphabetical":               // sort by sample or patient id (a.k.a. alphabetical)
+                            sortby_param = "case_id";
```
adamabeshouse: let's use an enum for the possible values of sortby_param instead of strings

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -191,6 +196,53 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
+                            sortby_param = "";
+                            break;
+                    }
+                    newParams[ONCOPRINT_SORTBY_URL_PARAM] = sortby_param;
+                } else {
+                    delete newParams[ONCOPRINT_SORTBY_URL_PARAM];
+                }
+
+                // update URL without page refresh
+                const {basename, pathname} = getBrowserWindow().globalStores.routing.location;
```
adamabeshouse: this block will be redundant once you move your changes into the other reaction, but just FYI you can just do it like this:

`getBrowserWindow().globalStores.routing.updateRoute(newParams, undefined, true, true);` as we do on line 195

Sjoerd-van-Hagen:
Thanks @adamabeshouse !

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: The reason for housing the updates in a different reaction is that changes of the sortBy settings do update the URL, but do not trigger the page to reload. If I would house these in the same reaction I would have to do overly complicated evaluation to see which value was updated and trigger a page refresh or url update depending on this.

There was a obvious mistake that I discovered by your comment and that was that the `urlParamsReaction` reaction was overwritten with the reaction for the sortBy params. I have created a new variable ` urlParamsReactionNoRefresh` to remove this collision.

src/shared/components/oncoprint/ResultsViewOncoprint.spec.tsx
pvannierop: Thnx, I am getting the hang of JS test :)

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: Updated. A string valued enum has been added replacing hard coded strings in the code:
```
export enum SortByUrlParamValue {
    CASE_ID = "case_id",
    CASE_LIST = "case_list",
    DATA = "data",
    DATA_MUTATION = "data_mut",
    DATA_DRIVER = "data_dri",
    DATA_MUTATION_DRIVER = "data_mut_dri",
    NONE = ""
}
```

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: Again, I put these in separate reaction because I do **not** want to trigger a page update and AFAIKS `getBrowserWindow().globalStores.routing.updateRoute(newParams, undefined, true, true);` does trigger a refresh.

If I am mistaken I woul dbe more than happy to move, but first plz confirm that you understand that the URL action for the sortBy param differs from that of  columnMode, heatmapTrackGroupsUrlParam and clinicalTracksUrlParam.

alisman: @pvannierop can you clarify what you mean by "refresh"?   if you mean reload of actual HTML page, it should not do that.   it may trigger re-render of whole oncoprint.  the reason is that it changes what we think of as the "query."   there is a complication in all this b/c for large queries we cannot store all params in url.  we have to save it to session.    should these params be considered part of that session?  would a user want these settings to be reproduced when sharing the query with another user?   ultimately, we may need to decouple what we think of as the session from some settings which are also stored in the url.  complicated, i know!  perhaps we should have a meeting to go over this.

adamabeshouse: yes the existing reaction definitely does not cause a refresh, please clarify

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -168,6 +183,8 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
         this.onMouseEnter = this.onMouseEnter.bind(this);
         this.onMouseLeave = this.onMouseLeave.bind(this);
 
+        // update URL parameters according to UI events
```
alisman: we want to try to maintain the url as the main source of truth, so instead of setting these on the store, we would change the url and then let the app respond to that.

adamabeshouse: @alisman I think you are saying we should no longer have a JS representation for columnMode, heatmapTrackGroupsUrlParam, and clinicalTracksUrlParam - any code that tries to update those should be rewritten to update straight into the URL. That sounds like a good idea, but unrelated to this PR.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: @pvannierop still confused what you mean - neither reaction causes a refresh

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
Sjoerd-van-Hagen: Does that mean that we can get this merged now? @adamabeshouse @alisman

adamabeshouse: @Sjoerd-van-Hagen we still don't understand the issue of refreshing, these reactions should be combined. @pvannierop

Sjoerd-van-Hagen: Sure, no problem. How high is this on your list of priorities? We would like to see this in the next release if possible.

adamabeshouse: @Sjoerd-van-Hagen sorry, to be clear, we need @pvannierop 's clarification to our confusion please.

I'm not sure where it falls on our priorities but I think it wouldn't be too problematic to merge @alisman

Sjoerd-van-Hagen: Right, you need a reply to the other comment you made.

Pim is OOO until Wednesday though :-(

adamabeshouse: Can we wait until then?

Sjoerd-van-Hagen: Yes, let's do that then. I was actually under the impression that Pim has had a call with @alisman about this.

pvannierop: Yes, I had a call with @alisman. I was under the impression that @adamabeshouse and @alisman were internally discussing a solution to the refresh problem.

The refresh of oncoprint refers to the loading icon appearing when triggering `updateRoute()` command after setting the `oncoprint_sortby` url param. This behavior is different from toggling the sort order via the menu (no refresh of oncoprint) and is, therefore, undesired.

I propose this include this topic in our telco of this afternoon.

jjgao:
I am not sure about `data_mut`, `data_dri`, and `data_mut_dri`. It may make more sense to be consistent with the options from the menu, i.e. to create two more parameters, e.g. `by_driver` and `by_mutation_type`.

We would like to add the state (starting with oncoprint) into the session https://github.com/cBioPortal/cbioportal/issues/5525. Maybe we should just do that instead of adding more into the url?

Sjoerd-van-Hagen:
@jjgao having this in the URL is a requirement for us because the client would like to be able to link from their system to cBioPortal, and see the patients/samples in the same order.

jjgao:
@Sjoerd-van-Hagen I see -- that makes sense. I think we should document the url.

What do you think about my first point?

Sjoerd-van-Hagen:
@jjgao what you say makes sense but when I go to the oncoprint there is also the option to enable 'sorting on data' while unchecking both checkboxes, and this does change the sorting of the oncoprint. What it actually does, is sort on (has_any_mutation, sample_id). Not sure how that is useful, but it does mean that Pim's implementation matches current oncoprint behavior.

Our client is waiting to update their instances to have this feature (and the gene panels fix) so it is important that this will make it into the next release. If that means removing the data_mut options, I think that is fine, as I do not see the point of this way of sorting, but perhaps someone else does?

jjgao:
I think it is worth a discussion about how to translate the state of oncoprint (and other UI) into url parameters. It is not ideal that we have to write special code every time we would like to add certain state into the url. For example, we may also want to code "sorting by clinical attributes".

One dumb idea is to define a state json (probably already exist) and add it to the url, similar to what GDC is dong: https://portal.gdc.cancer.gov/exploration?filters=%7B%22op%22%3A%22and%22%2C%22content%22%3A%5B%7B%22op%22%3A%22in%22%2C%22content%22%3A%7B%22field%22%3A%22cases.primary_site%22%2C%22value%22%3A%5B%22Brain%22%5D%7D%7D%5D%7D

Sjoerd-van-Hagen:
@jjgao sure, but that is a separate discussion imo.

jjgao:
@Sjoerd-van-Hagen I worry that once we add something in the url, we don't have the freedom to remove it in the future. If the json solution is what we would like to do, I think it's better not to have another set of parameters to maintain.

Maybe we can support json now and start with the oncoprint_sort parameter in the json?

jjgao:
Discussed with @Sjoerd-van-Hagen on Slack. We couldn't agree on what's the best way to do it, but agree to disagree. For this PR, we agreed to:

- [ ] Leave the URL change undocumented so that we will have the freedom to change it later.
- [ ] Drop data_mut, data_dri, and data_mut_dri.

Sjoerd-van-Hagen:
Thanks @jjgao as for the best way, that will require more discussion with a broader group.

@pvannierop can you update the PR accordingly?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -592,6 +694,37 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
+                    this.configureSortMode('data', false, true);
+                    break;
+                case SortByUrlParamValue.DATA_MUTATION_DRIVER:                                    // sort by data + mutation type + driver/passenger
+                    this.configureSortMode('data', true, true);
+                    break;
+            }
+        }
+    }
+
+    private configureSortMode(type:'data'|'caseList'|'alphabetical', byMutation:boolean, byDriver:boolean){
```
adamabeshouse: should be @action

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -187,10 +207,88 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
+                    newParams[ONCOPRINT_SORTBY_URL_PARAM] = sortby_param;
+                }
                 getBrowserWindow().globalStores.routing.updateRoute(newParams, undefined, true, true);
             }
         );
 
+        // update URL parameters according to sort direction UI events
+        // and DO NOT trigger a page refresh
+        // this.urlParamsReactionNoRefresh = reaction(
+        //     ()=>[
```
adamabeshouse: delete comment

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -43,11 +45,22 @@ interface IResultsViewOncoprintProps {
     addOnBecomeVisibleListener?:(callback:()=>void)=>void;
 }
 
+export enum SortByUrlParamValue {
+    CASE_ID = "case_id",
+    CASE_LIST = "case_list",
+    DATA = "data",
+    DATA_MUTATION = "data_mut",
```
adamabeshouse: rename as @jjgao recommended?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: Ah really? It is just a method that is called with real arguments in the code block above. It is not and should not be autotriggered by observable changes. Or am I wrong in this?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: Deleted

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: All references to `data` and related sort subtypes were removed. Renaming is not relevant for this code anymore.

pvannierop:
@adamabeshouse  @jjgao I have updated the PR. Please review and comment.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: action is a decorator you put on functions that change observables. you're thinking of autorun. action wraps the function in a transaction so that observable changes are not triggered until the function completes.

jjgao:
thanks, @pvannierop!

Since we are not going to document it (and commit to the parameters yet), would it be possible NOT to show the `oncoprint_sortby` parameter in the url after changing sort_by in OncoPrint?

pvannierop:
>Since we are not going to document it (and commit to the parameters yet), would it be possible NOT to show the oncoprint_sortby parameter in the url after changing sort_by in OncoPrint?

Ok :) Setting of the URL parameter by the sort mode dialog was removed.

Sjoerd-van-Hagen:
@adamabeshouse @jjgao can this be merged today?

jjgao:
no further comments from me for the product part.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
adamabeshouse: please do this

adamabeshouse:
@Sjoerd-van-Hagen @pvannierop there is an outstanding comment, besides that it's ready https://github.com/cBioPortal/cbioportal-frontend/pull/2212#discussion_r283465875

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
pvannierop: @adamabeshouse Ok, annotation was added. Sorry for the confusion.

Sjoerd-van-Hagen:
@adamabeshouse should be done now :-)

Sjoerd-van-Hagen:
@adamabeshouse can this be merged?
@inodb and be part of the 3.0.0 release?

alisman:
@Sjoerd-van-Hagen yes, just spoke with pim.  we rebased on 3.0.0 and i will merge this today after we test it out one last time

PR #2213, opened 2019-04-10
Global settings menu for results page

adamabeshouse:
@alisman styling help is invited

image

image

### DIFF
--- a/end-to-end-tests/specs/mutationsTab.spec.js
+++ b/end-to-end-tests/specs/mutationsTab.spec.js
@@ -16,4 +16,17 @@ describe("mutations tab", function() {
         browser.waitForVisible('[data-test="LazyMobXTable_CountHeader"]', 10000);
         assert(browser.getHTML('[data-test="LazyMobXTable_CountHeader"]', false).indexOf("0 Mutations") > -1);
     });
+    it("uses germline filtering", function() {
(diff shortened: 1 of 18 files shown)


alisman:
@adamabeshouse @jjgao  which other tabs are influenced by these global settings?  just curious.

I think the UI of the "Case type" select is not as intuitive as it is in oncoprint itself.  "Events by sample vs Events by patient" is super easy to grok.  Can we have same in global settings?

alisman:
Is "VUS" a universally understood acronym amount our userbase?

alisman:
I know the "cBioPortal > 10" labeling has existed forever in oncoprint, but i it's not totally clear what it means (at least to me!)

alisman:
@adamabeshouse i think we could get rid of the "Settings" header in dropdown and just have  a header that better explains the the Patient/Sample control and is same size as rest of headers.  

Also, I think putting a horizontal rule under each section would be nice.

adamabeshouse:
@alisman it affects the tabs listed here https://github.com/cbioportal/cbioportal/issues/5499

I agree with all your comments.. @jjgao what do you think about making it "cbioportal count" instead of just "cbioportal". or is that implied?

adamabeshouse:
@alisman made changes, see updated screenshot above

adamabeshouse:
@jjgao what design/placement are we thinking now for this?

jjgao:
@adamabeshouse I think the position looks good. @alisman thoughts?

Does the global setting persist between sessions?

adamabeshouse:
@jjgao 
> Does the global setting persist between sessions?
currently, no

jjgao:
I think they should be included at least in the session / url.

I am not sure about persisting between sessions because opening a shared url may mess up with my setting. Maybe we can make it an option - save to my setting? If so, ideally we should save it to local storage.

adamabeshouse:
https://github.com/cBioPortal/cbioportal-frontend/pull/2838/files

PR #2216, opened 2019-04-11
change the error message on query page to the same style as the result page(query limit)

dippindots:
change the error message on query page to the same style as the result page(query limit)
Fix # https://github.com/cbioportal/cbioportal/issues/5957.


alisman:
@dippindots i think what @jjgao meant was, we should show the query component, populated with the query and in the state of >100 genes.

PR #2217, opened 2019-04-11
show query form when URL contains invalid query

dippindots:
show query form when URL contains invalid query
Fix # https://github.com/cbioportal/cbioportal/issues/5957.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -412,24 +412,19 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
         }
     }
 
-    @computed get isQueryValid() {
-        return this.resultsViewPageStore.hugoGeneSymbols.length <= AppConfig.serverConfig.query_gene_limit;
-    }
-
     @computed get pageContent(){
-        // if qeury invalid, return error page
-        if (!this.isQueryValid) {
-            return (
-                <div className="alert alert-warning queryInvalid" role="alert">
-                    <p>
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -412,24 +412,19 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
-                    </p>
-                </div>
-            );
-        }
-
         // if studies are complete but we don't have a tab id in route, we need to derive default
         return (<div>
+            {
+                // if qeury invalid, return error page
+                (this.resultsViewPageStore.isQueryInvalid) && (
```
alisman: we should probably have a mechanism different types of invalidity (of which gene length is only one) and messaging for each one.  but since right now we only do gene length, this is ok.  maybe put a comment explaining that.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -444,7 +439,7 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                         </div>
 
                         {
-                            (!this.resultsViewPageStore.genesInvalid) && (
+                            (!this.resultsViewPageStore.genesInvalid && !this.resultsViewPageStore.isQueryInvalid) && (
```
alisman: like for example, genesInvalid should probably be geneNamesInvalid and should be part of isQueryValid.  but we should follow up with that.  we don't need to handle this now.

dippindots:
@alisman Added some comments and rebased to the latest master branch.

PR #2218, opened 2019-04-11
Fix bug so that mutations tab will respond to VUS filtering

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1579,7 +1579,7 @@ export class ResultsViewPageStore {
             // use oql filtering in mutations tab only if query contains mutation oql
             mutations = (this.oqlFilteredAlterations.result || []).filter(a=>isMutation(a));
         } else {
-            mutations = this.mutations.result || [];
+            mutations = this.putativeDriverAnnotatedMutations.result || [];
```
alisman: is it weird that you had to know to use putativeDriver annotated mutations here?  i mean, maybe this.mutations should have the putativeDriverAnnotations already there and we should have some rawMutations or unAnnotated mutations collection?  just thinkin

adamabeshouse: it would be unannotated unfiltered mutations. so its just hard to name these the right way. raw mutations is close but I find it a bit confusing too.

PR #2224, opened 2019-04-15
Break the word when no space in the description

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5947


alisman:
@jjgao i'm merging this in spite of a single failure which, based on screenshot, is clearly not related to this fix.  it's server latency

PR #2231, opened 2019-04-17
improvements to group comparison overlap tab

adamabeshouse:
(1) Fix layout issue in venn diagrams (https://github.com/cBioPortal/cbioportal/issues/5968)
(2) remove unused VennDiagram legacy class
(3) add minimum size for venn diagram regions
(4) adjust tooltip to follow mouse in UpSet like it does in venn diagram

BEFORE:
image

AFTER:
image


src/pages/groupComparison/VennUtils.ts
```diff
@@ -0,0 +1,44 @@
+import _ from "lodash";
+
+export function computeDisplaySizes<R extends {
```
kalletlak: this might need some testing

adamabeshouse: I'm not sure how

PR #2236, opened 2019-04-19
Always show virtual study category in query page

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4792

Before

---
After

With virtual studies
Screen Shot 2019-06-26 at 2 07 37 PM

Without virtual studies
Screen Shot 2019-06-26 at 2 21 50 PM

### DIFF
--- a/src/shared/components/query/CancerStudyTreeData.ts
+++ b/src/shared/components/query/CancerStudyTreeData.ts
@@ -5,17 +5,21 @@ import {VirtualStudy} from "shared/model/VirtualStudy";
 
 export const CANCER_TYPE_ROOT = 'tissue';
 export const VIRTUAL_STUDY_NAME= 'My Virtual Studies';
+export const PHYSICAL_STUDY_NAME= 'Studies';
 
-export type CancerTreeNode = CancerType|CancerStudy;
+export type CancerTypeWithVisibility = CancerType & { alwaysVisible?:boolean };
+
+export type CancerTreeNode = CancerTypeWithVisibility|CancerStudy;
+
 export type NodeMetadata = {
 	isCancerType:boolean,
 	isPriorityCategory:boolean,
-	priorityCategories:CancerType[],
-	childCancerTypes:CancerType[],
+	priorityCategories:CancerTypeWithVisibility[],
+	childCancerTypes:CancerTypeWithVisibility[],
 	childStudies:CancerStudy[],
 	descendantStudies:CancerStudy[],
-	descendantCancerTypes:CancerType[],
-	ancestors:CancerType[], // in order of ascending distance from node, ending in root node
+	descendantCancerTypes:CancerTypeWithVisibility[],
+	ancestors:CancerTypeWithVisibility[], // in order of ascending distance from node, ending in root node
 	siblings:CancerStudy[], // descendants of highest level non-root ancestor
 	searchTerms:string, // all related text in a single string for easy searching
 };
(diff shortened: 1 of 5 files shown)


src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -84,6 +88,13 @@ export default class CancerStudyTreeData
 		}
 		// add virtual study category, and studies
 		cancerTypes = [this.virtualStudyCategory, ...this.priorityCategories, this.rootCancerType, ...cancerTypes];
+		cancerTypes = _.map(cancerTypes,cancerType=>{
+
+			if(cancerType.cancerTypeId === "acc"){
```
zhx828: ?

kalletlak: oh, forgot to clean it

alisman:
@tmazor @jjgao seems like virtual studies are now listed below the physical studies.  seems like they should be above, right?  also @kalletlak they are laid out a little bit different than the physical studies.  indentation and such
image

tmazor:
Agreed - virtual studies should always appear at the top of the list, and the indentation should be the same as physical studies

jjgao:
Maybe we need a header for the physical studies when we just list them all?

kalletlak:
@jjgao we do require a header for it. This is how it looks without header

Screen Shot 2019-06-14 at 4 03 06 PM

jjgao:
@tmazor @alisman any suggestions on header? Or any other way?

alisman:
image
Unfortunately, I think we do need a header for the list of physical studies in case where max depth is zero.  The above is confusing b/c it's hard to say when virtuals end and physicals begin.

tmazor:
I agree, it looks like we need a header. How about "Primary Studies" ?

jjgao:
I like "Primary Studies".

schultzn:
I associate "primary" with "primary tumor" - maybe this could be misleading? "Original Studies", "Curated Studies"?

alisman:
Just throwing this out there but what if instead of titles, we put a "virtual" marker on the studies themselves?
image

jjgao:
hmm, now sure. Maybe we just need a head of "Studies"?

Sjoerd-van-Hagen:
How about 'groups' and 'studies'? It matches the way it is used in the group comparison.

alisman:
@Sjoerd-van-Hagen i think that would conflate groups and virtual studies.   Maybe the _should_ be the same thing but at the moment they are not.

Sjoerd-van-Hagen:
True, but finding another synoniem for group is probably also not advisable. I would like to keep study was study, simply because people know that it is. With 'primary study' I am not sure if that is the case.

Perhaps 'cross-study group' or 'cross-study cohort'? Cannot make it more explicit than that.

alisman:
I understand what you mean.  I like Studies and Virtual Studies.

kalletlak:
Updated

With virtual studies
Screen Shot 2019-06-26 at 2 07 37 PM


Without virtual studies
Screen Shot 2019-06-26 at 2 13 03 PM

tmazor:
Looks good @kalletlak !

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -5,17 +5,21 @@ import {VirtualStudy} from "shared/model/VirtualStudy";
 
 export const CANCER_TYPE_ROOT = 'tissue';
 export const VIRTUAL_STUDY_NAME= 'My Virtual Studies';
```
zhx828: Should we use the Pluralize to deal with the study/studies?

alisman: No, I think it would not be right to have a cateogory "My Virtual Study" which would seem to convey "you only get one!".    Just like the github Issues tab is not Issue if there's only one issue.

PR #2237, opened 2019-04-19
Create tooltips with list of gene panels that are clickable and summo…

lemccarthy:
…n a modal. Contains API changes and prettier-reformatting of files. Also removed unused phantomjs dependencies that were causing problems on a slow internet connection. 

Fix #5937


src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -88,14 +102,27 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
         }
     }
 
+    toggleModal = (panelName: string, genes: GenePanelToGene[]) => {
```
alisman: This should be a bound method.  You can do this with @autobind decorator.  Also, it should be have mobx's @action decorator.  Any time  we mutate mobx managed state we want to, want to do so inside of an action.  First and foremost, this batches the updates so you don't get multiple reactions (reactions being sycnronous).  It can also be useful for debugging because a debugger can log actions.  It's a little bit like redux's dispatcher.  Look or @action on other methods.

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -88,14 +102,27 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
...
+        this.modalOpen = !this.modalOpen;
+        if (!this.modalOpen) {
```
alisman: I would just go ahead and set them every time.  The cost is totally negligible.

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -88,14 +102,27 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
...
+            return;
+        }
+        this.modalPanelName = panelName;
+        this.modalPanelGenes = genes;
+    }
+
+    closeModal = () => {
```
alisman: same here.  Make this an @action decorated method.  @autobind will handle bindind to instance.

src/pages/studyView/table/GenePanelModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+                            toggleModal(panel.genePanelId, panel.genes);
+                        }}
+                    >
+                        {panel.genePanelId}
+                    </a>
+                ])}
+            </span>
+        );
+    } else {
+        return <span />;
```
alisman: null?

src/pages/studyView/table/GenePanelModal.tsx
```diff
@@ -0,0 +1,65 @@
...
+}
+
+export const GenePanelList: React.FunctionComponent<IGenePanelTooltipProps> = ({
+    genePanels,
+    toggleModal
+}) => {
+    if (genePanels.length > 0) {
+        return (
+            <span>
+                <br />
```
alisman: what's reason for this BR?

src/pages/studyView/table/CNAGenesTable.tsx
lemccarthy: Using the arrow syntax creates binding, my linter shows it as bound. I will add the @action one though!

src/pages/studyView/table/GenePanelModal.tsx
lemccarthy: for spacing, its use here is equivalent to a newline

src/pages/studyView/table/GenePanelModal.tsx
lemccarthy: My thinking here was to avoid a null to simplify the type definition and error checking. Visually this does nothing. I can return null and update the type definitions if you want.

src/pages/studyView/table/CNAGenesTable.tsx
lemccarthy: Ok I've updated it to this syntax

src/shared/api/generated/CBioPortalAPIInternal.ts
```diff
@@ -555,7 +573,7 @@ export type VariantCountIdentifier = {
  */
 export default class CBioPortalAPIInternal {
 
-    private domain: string = "";
+    private domain: string = "http://localhost:8080/cbioportal/api";
```
alisman: hardcode heads up

PR #2238, opened 2019-04-22
redirect cgds_r.jsp

inodb:


inodb:
Merging this b/c the only e2e test that is failing is an unrelated test that fails frequently according to insights tab

PR #2240, opened 2019-04-22
E2E: Adjust global waitFor timeout

alisman:


end-to-end-tests/wdio/webdriver-manager.conf.js
```diff
@@ -294,8 +294,8 @@ if (doBrowserstack) {
 
     config.browserstackLocal = true;
 
-    config.user = "aaronlisman2";
```
alisman: gonna change pword since i realized this is git history now

PR #2245, opened 2019-04-23
tracking study download

dippindots:
front-end part tracking for study download.
Partial Fix #https://github.com/cBioPortal/cbioportal/issues/6003.


nabeelmerchant1:
1.For the backed a new database tab could be created for all studies which would track each time the download button is clicked.  When button is clicked you can increment the counter and serve the file.


2. Either way we could also use google analytics
<a href="http://www.example.com/files/map.pdf" onClick="javascript: _gaq.push(['_trackPageview', '/downloads/map']);"> 

Another suggestion could be to also track the number of time a study or genes have been viewed.

PR #2250, opened 2019-04-24
Compress layout of plots tab

alisman:


src/routes.jsx
```diff
@@ -98,7 +98,7 @@ export const makeRoutes = (routing) => {
                 <Route path="/loading/comparison" onEnter={()=>{$(document).scrollTop(0)}} component={GroupComparisonLoading} />
                 <Route path="/results/legacy_submission" onEnter={handleLegacySubmission} component={getBlankPage()} />
 
-                <Route path="/results(/:tab)" onEnter={()=>{$(document).scrollTop(0)}} getComponent={lazyLoadComponent(ResultsViewPage)} />
+                <Route path="/results(/:tab)" onEnter={()=>{}} getComponent={lazyLoadComponent(ResultsViewPage)} />
```
adamabeshouse: why?

vendor-bundles.webpack.config.js
```diff
@@ -28,13 +28,11 @@ let config = {
...
     },
 
-    devtool : 'source-map',
-
     plugins: [
 
     ],
 
-    devtool : 'source-map'
+    devtool : ''
```
adamabeshouse: why?

alisman: attempt to reduce built time

adamabeshouse: wont it make it harder to debug?

src/routes.jsx
alisman: because when user switches tabs on results view, Niki wants browser to stay where it is, not go to top.

adamabeshouse: ah okay

PR #2254, opened 2019-04-25
exclude overlapping samples and patients by default for survival and clinical tab

dippindots:
This pr includes these changes:
1. Change option text from  `Exclude overlapping samples` to `Exclude overlapping samples and patients`
2. For survival and clinical tab, exclude overlapping samples and patients by default and show alert `We exclude overlapping samples and patients by default to display this tab`
3. Don't grey out the clinical tab and always show clinical table

Fix # https://github.com/cBioPortal/cbioportal/issues/6010.

### DIFF
--- a/src/pages/groupComparison/GroupComparisonStore.ts
+++ b/src/pages/groupComparison/GroupComparisonStore.ts
@@ -56,8 +56,8 @@ import ComplexKeyGroupsMap from "../../shared/lib/complexKeyDataStructures/Compl
 import {GroupComparisonURLQuery} from "./GroupComparisonPage";
 
 export enum OverlapStrategy {
-    INCLUDE = "Include overlapping samples",
-    EXCLUDE = "Exclude overlapping samples",
+    INCLUDE = "Include overlapping samples and patients",
+    EXCLUDE = "Exclude overlapping samples and patients",
 }
 
 export default class GroupComparisonStore {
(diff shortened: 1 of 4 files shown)


kalletlak:
@dippindots I think even the underlying logic should be updated. I mean excluding overlapping patients

dippindots:
@kalletlak Thanks for your prompt review, I should ask you before do this. I will make the change and tell you when it is ready, thanks!

adamabeshouse:
@kalletlak patients are already excluded independently of samples

kalletlak:
@adamabeshouse yeah true

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -211,6 +211,21 @@ export default class GroupComparisonStore {
         invoke:()=>Promise.resolve(this.availableGroups.result!.filter(group=>this.isGroupActive(group)))
     });
 
+    readonly filteredAvailableGroups = remoteData<OverlapFilteredComparisonGroup[]>({
```
adamabeshouse: lets change `filtered` to `nonOverlapping`, it's a bit more specific

adamabeshouse: also lets use this in the body of `availableGroups` in the case `OverlapStrategy.EXCLUDE`

PR #2258, opened 2019-04-26
Ignore the invalid cases in the Custom Selection

zhx828:
Fix https://github.com/cBioPortal/cbioportal/issues/5986

Screen Shot 2019-05-14 at 11 58 17 AM
Screen Shot 2019-05-14 at 11 58 33 AM

### DIFF
--- a/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelectionUtils.tsx
+++ b/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelectionUtils.tsx
@@ -7,6 +7,7 @@ import {
 } from "../../StudyViewPageStore";
 import * as _ from 'lodash';
 import {Sample} from "../../../../shared/api/generated/CBioPortalAPI";
+import Pluralize from "pluralize";
 
 type Code =
     'MULTI_NAME'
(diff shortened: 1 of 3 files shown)


src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelectionUtils.tsx
```diff
@@ -143,25 +146,25 @@ export function validateLines(lines: InputLine[], caseType: ClinicalDataType, al
...
             } else {
-                _case = `${line.studyId}:${line.caseId}`;
-                if (validPair[_case] === undefined) {
-                    invalidCases.push(`${line.studyId}:${line.caseId}`);
-                    if (invalidCases.length >= TOO_MANY) {
-                        return false;
-                    }
-                } else {
+                _case = getUniqueCaseId(line.studyId, line.caseId);
+                if (validPair[_case] !== undefined) {
```
kalletlak: combine two if conditions

zhx828: They are covering different cases though

alisman:
Is there a reason we can't just expect the user to clean up the list?  

The message should be "The following case is invalid and will be ignored"   instead of "which"

alisman:
While we're at this, could we apply bootstrap styling to the TEXTAREA?  it's always bothered me.  And is it possible to make the warning message an alert and use the bootstrap semantic styling?

zhx828:
@alisman thanks, I will update the styles. The use case is that users of bigger studies may have a list with 1k+ samples and by the time, there may be 200 or more not available. But in this case, they don't care about these invalid cases, but only want to take a look the rest. Manually removing all the invalid cases is doable but created an extra burdens especially for users using this feature very frequently. Niki asked to change the error message to warning to better suit the use case.

zhx828:
@inodb @alisman I took a look the e2e, they are not related to this pr. But these tests seem need to be updated due to data change, but need to do additional verification. Could one of you restart the netlify ? If they are passing, then this pr can be merged.

PR #2260, opened 2019-04-29
Added google analytics badge to study view

lemccarthy:
https://github.com/cBioPortal/cbioportal/issues/5953

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -90,6 +90,8 @@ import {getMDAndersonHeatmapStudyMetaUrl, getStudyDownloadListUrl} from "../../s
 import onMobxPromise from "../../shared/lib/onMobxPromise";
 import request from 'superagent';
 import {trackStudyViewFilterEvent} from "../../shared/lib/tracking";
+import queryStudyAnalytics from '../../shared/api/queryStudyAnalytics';
+import AppConfig from "appConfig";
 
 export enum ClinicalDataTypeEnum {
     SAMPLE = 'SAMPLE',
@@ -3318,4 +3320,7 @@ export class StudyViewPageStore {
         default: []
     });
 
+    readonly studyGoogleAnalytics = remoteData({
+        invoke: () => queryStudyAnalytics(this.studyIds, AppConfig.serverConfig.analytics_report_url || ""),
+    });
 }
--- a/src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
+++ b/src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
@@ -28,6 +28,7 @@ export default class StudyPageHeader extends React.Component<IStudyPageHeaderPro
                     studies={this.props.store.displayedStudies.result}
                     originStudies={this.props.store.originStudies}
                     showOriginStudiesInSummaryDescription={this.props.store.showOriginStudiesInSummaryDescription}
+                    studyAnalytics={this.props.store.studyGoogleAnalytics.result}
                 />
                 <RightPanel
                     store={this.props.store}/>
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -5,6 +5,7 @@ export interface IAppConfig {
     frontendUrl?: string;
     serverConfig:IServerConfig;
     hide_login?:boolean;
+
 }
 
 export type CategorizedConfigItems = {
@@ -25,6 +26,7 @@ export type StudyView = {
 };
 
 export interface IServerConfig {
+    "analytics_report_url": string | null;
     "app_name": string | null;
     "app_version": string|null;   // default: "1.0"
     "authenticationMethod": string | undefined;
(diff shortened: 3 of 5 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2930,6 +2931,10 @@ export class StudyViewPageStore {
         }
     });
 
+    readonly studyGoogleAnalytics = remoteData({
+        invoke: async () => await queryStudyAnalytics(this.studyIds, process.env.API_URL || ""),
```
alisman: you actually can 1. take away async and 2. get rid of await since invoke just wants a promise returned.  we only really need async/await when we want to take advantage of the synrconous assignment sugar:
```var data = await getData();```

File: src/pages/studyView/queryStudyAnalytics.ts [hunk 2]
```diff
@@ -0,0 +1,24 @@
+import * as request from "superagent";
```
alisman: you should put this in shared/api folder where we keep the other api clients

src/pages/studyView/queryStudyAnalytics.ts
```diff
@@ -0,0 +1,24 @@
...
+
+export interface IAnalyticsResponse {
```
alisman: Lets call it PortalAnalyticsResponse

src/pages/studyView/queryStudyAnalytics.ts
```diff
@@ -0,0 +1,24 @@
...
+    totalUses: number;
+    totalUniqueUses: number;
+    timesVisited: number;
+    timesUniqueVisited: number;
+}
+
+export interface IQueryResponse extends request.Response {
```
alisman: Too general I think.    Probably you could get away with just typing this inline since it's unlikely to be used elsewhere.

src/pages/studyView/queryStudyAnalytics.ts
alisman: Actually, lets just call it PortalAnalyticsData since the fact that was delivered via api is immaterial

src/pages/studyView/queryStudyAnalytics.ts
```diff
@@ -0,0 +1,24 @@
...
+    body: IAnalyticsResponse;
+}
+
+export default async function queryStudyAnalytics(
+    studyIds: string[],
+    apiURL: string
+): Promise<IAnalyticsResponse> {
+    try {
+        const res: IQueryResponse = await request.get(`${apiURL}?studyName=${studyIds[0]}`);
```
alisman: I see, we're making up for fact that we don't have SuperAgent typings.   I'm sure they exist.  Anyways, this is totally not important.

src/pages/studyView/queryStudyAnalytics.ts
```diff
@@ -0,0 +1,24 @@
...
+        return res.body;
+    } catch (error) {
+        throw new Error("400");
```
alisman: We don't really want to catch this error and throw a new one because any error listeners (sentry.js) will then lose the browser provided error context.   Mobxpromise will reflect the error (isError) and that can then be handled in code (maybe just hide badge)

src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
```diff
@@ -1,25 +1,31 @@
...
-import {StudyViewPageStore} from 'pages/studyView/StudyViewPageStore';
+import * as React from "react";
+import { observer } from "mobx-react";
+import { StudyViewPageStore } from "pages/studyView/StudyViewPageStore";
 import RightPanel from "./rightPanel/RightPanel";
 import StudySummary from "./studySummary/StudySummary";
 import UserSelections from "../UserSelections";
-import * as _ from 'lodash';
-
+import { keys, values } from "lodash";
```
alisman: I'm not sure about this.  I understand that if we do this, we could potentially save weight of unused lodash methods (if tree shake), but the whole library is only 24k gzipped and I think we probably use more than half of the methods.  The problem above is that it's harder to know the functions are from lodash.  You have to look up to imports if you want to make sure.  Would rather stick with what we have.

alisman:
@lemccarthy hey the prettier js reformatting is making it pretty hard to for me to see the real changes.  i'm 100% committed to prettier but i want to do it in a controlled manner.   still trying to figure out why github doesn't show the diff properly

src/pages/studyView/queryStudyAnalytics.ts
```diff
@@ -0,0 +1,24 @@
...
+    totalUniqueUses: number;
+    timesVisited: number;
+    timesUniqueVisited: number;
+}
+
+export interface IQueryResponse extends request.Response {
+    body: IAnalyticsResponse;
+}
+
+export default async function queryStudyAnalytics(
```
alisman: lets call this "getStudyAnalytics", inline with other api clients

alisman:
@lemccarthy looks like you pushed this branch to main repository.  you should fork cbioportal-frontend and then push branches to your personal origin.  then make PRs to the main repository

src/pages/studyView/StudyViewPageStore.ts
lemccarthy: Changes have since been made to implement this.

src/pages/studyView/queryStudyAnalytics.ts
lemccarthy: The file is now moved

src/pages/studyView/queryStudyAnalytics.ts
lemccarthy: So renamed

src/pages/studyView/queryStudyAnalytics.ts
lemccarthy: Ok, that catch is removed.

src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
lemccarthy: Reverted. I can see your point, its something I do out of habbit, and as a best practice when writing something from scratch.

PR #2262, opened 2019-04-29
Percent stacked bar end axis no extra space

dippindots:
remove extra space for percent stacked bar chart
Fix # https://github.com/cBioPortal/cbioportal/issues/6020.


kalletlak:
@dippindots you need to update screenshots

dippindots:
@kalletlak Thanks! Will do!

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -293,7 +293,7 @@ export default class MultipleCategoryBarPlot extends React.Component<IMultipleCa
 
     @computed get chartDomainPadding() {
         return {
-            [this.countAxis]:this.countAxisDomainPadding,
+            [this.countAxis]: this.props.percentage ? 0 : this.countAxisDomainPadding,
```
alisman: would it make sense to do this even if the axis is absolute (not percentage?)

dippindots: @alisman I found an example. Seems like we cannot do this to not percentage chart, right?
image

alisman: @schultzn @jjgao can you comment on the above?  With absolute values, should we always try to show a tick at the top?

jjgao: @alisman I think it's ok not to have a tick for absolute values.

PR #2265, opened 2019-04-29
change protein and mrna up/down to high/low in oncoprint tooltip

dippindots:
change protein and mrna up/down to high/low in oncoprint tooltip
Fix # https://github.com/cbioportal/cbioportal/issues/6017.


src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -21,12 +21,12 @@ export interface IDownloadFileRow {
 }
 
 const mrnaMap:any = {
```
adamabeshouse: these maps are probably not necessary anymore since they map a value to itself

dippindots: Yes, I will replace them.

alisman:
We should be using constants or string enums for these very repetetive strings.

PR #2272, opened 2019-04-30
Overlap changes

adamabeshouse:
Several changes to how excluding overlap is handled in group comparison

https://github.com/cBioPortal/cbioportal/issues/6032

### DIFF
--- a/src/pages/groupComparison/GroupComparisonUtils.tsx
+++ b/src/pages/groupComparison/GroupComparisonUtils.tsx
@@ -117,20 +117,18 @@ export function getVennPlotData(combinationSets: { groups: string[], cases: stri
 
 export function caseCountsInParens(
     samples:MobxPromise<any[]>|any[],
-    patients:MobxPromise<any[]>|any[],
-    asteriskForSamples:boolean = false,
-    asteriskForPatients:boolean = false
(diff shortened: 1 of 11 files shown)


src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -335,8 +333,55 @@ export function getOverlapFilteredGroups(
...
-    return `We can only show enrichments when two groups are selected. Please ${tooMany ? "deselect" : "select"} groups in the 'Active Groups' section so that only two are selected.`;
+export function MakeEnrichmentsTabUI(
+    getStore:()=>GroupComparisonStore,
+    getEnrichmentsUI:()=>MobxViewAlwaysComponent
+) {
+    return MakeMobxView({
+        await:()=>{
+            const store = getStore();
+            const ret:any[] = [store._selectedGroupsNotOverlapRemoved, store.activeGroups, store.activeStudyIds];
+            if ((store.activeGroups.isComplete &&
```
kalletlak: may be logical not instead of empty if block

adamabeshouse: yeah I wasn't sure about it but I think when its written this way it's easier to understand the logic. no?

PR #2274, opened 2019-05-01
upgrade to webpack 4

alisman:
Update to webpack 4
- do not output sourcemaps on cicleci
- cache babel output


vendor-bundles.webpack.config.js
```diff
@@ -1,76 +1,63 @@
-"use strict";
-/* eslint-disable */
+const webpack = require("webpack");
+const path = require("path");
 
-let webpack = require('webpack');
-const path = require('path');
+// we don't need sourcemaps on circleci
+const sourceMap = process.env.IS_CIRCLECI ? "" : "source-map";
```
inodb: @alisman minor stylistic thing but I tend to prefer `DISABLE_SOURCEMAP=true` and then set `DISABLE_SOURCEMAP=true` on CircleCI. You might want to reuse this configuration elsewhere that's not CircleCI.

inodb: personal preference thing though, i can see advantages of doing it the other way around

webpack.config.js
```diff
@@ -62,6 +64,11 @@ const css = join(src, 'styles');
 const fontPath = 'reactapp/[hash].[ext]';
 const imgPath = 'reactapp/images/[hash].[ext]';
 
+const cacheFolder = process.env.CACHE_FOLDER || false;
```
inodb: could be good to be explicit here `BABEL_CACHE_FOLDER`

.circleci/config.yml
```diff
@@ -40,6 +40,7 @@ run_e2e_and_save_artifacts: &run_e2e_and_save_artifacts
         environment:
           JUNIT_REPORT_PATH: ./junit/
           FRONTEND_TEST_USE_LOCAL_DIST: true
+          NO_PARALLEL: true
```
alisman: this is really weird but it's the only way i could get circleCI to build.  something to do with webpack4's new minification plugin.  just doesn't work by default on CircleCI.  when i shut off parrelization, it works.

inodb: @alisman odd that it only is required for `serveDist` and `yarn run test`. Or maybe it should be set for the run build command as well, but wasn't caught b/c of CircleCI caching?

PR #2278, opened 2019-05-01
Max width, and truncation with ellipsis, for group names in enrichments tab

adamabeshouse:
Fixes one entry of https://github.com/cBioPortal/cbioportal/issues/6031

### DIFF
--- a/src/shared/components/Checkbox.tsx
+++ b/src/shared/components/Checkbox.tsx
@@ -1,0 +1,27 @@
+import * as React from "react";
+import {observer} from "mobx-react";
+import DefaultTooltip from "./defaultTooltip/DefaultTooltip";
+
+export interface ICheckboxProps {
+    value?:string;
+    checked:boolean;
+    onClick:()=>void;
+    label:any;
+}
+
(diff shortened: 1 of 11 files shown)


src/shared/components/Checkbox.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface ICheckboxProps {
+    value?:string;
+    checked:boolean;
+    onClick:()=>void;
+    label:any;
+}
+
+@observer
+export default class Checkbox extends React.Component<ICheckboxProps, {}> {
```
kalletlak: there is already `react-bootstrap` checkbox. how different is this from that?

alisman: agreed.  if this is different, name should describe what it is.

adamabeshouse: the difference is the flex alignment, should I change the name to FlexAlignedCheckbox

kalletlak: I think even that does it right

PR #2215, opened 2019-04-10
Modify query show loading icon instead of 0 studies

dippindots:
improve loading behavior of query selector when default selection present.
Fix # https://github.com/cbioportal/cbioportal/issues/5911.

PR #2282, opened 2019-05-02
fix query page error when there is no study selected

dippindots:
fix query page error when there is no study selected
Fix issue introduced in https://github.com/cBioPortal/cbioportal-frontend/pull/2215.


alisman:
merging with unrelated e2e failures

PR #2288, opened 2019-05-04
update mutex tab(exclude unprofiled samples)

dippindots:
Update mutex tab(exclude unprofiled samples)
Fix # https://github.com/cbioportal/cbioportal/issues/6033.

Proposed changes:
a. exclude unprofiled oql track add a warning when oql is not profiled in all samples
image

b. change alert when there are less than 2 available oql
image

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1018,9 +1018,9 @@ export class ResultsViewPageStore {
 
 
     readonly isSampleAlteredMap = remoteData({
-        await: () => [this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine, this.samples],
+        await: () => [this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine, this.samples, this.coverageInformation, this.selectedMolecularProfiles],
         invoke: async() => {
-            return getSampleAlteredMap(this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine.result!, this.samples.result, this.rvQuery.oqlQuery);
+            return getSampleAlteredMap(this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine.result!, this.samples.result, this.rvQuery.oqlQuery, this.coverageInformation.result, this.selectedMolecularProfiles.result!.map((profile) => profile.molecularProfileId));
         }
     });
 
--- a/src/pages/resultsView/ResultsViewPageStoreUtils.spec.ts
+++ b/src/pages/resultsView/ResultsViewPageStoreUtils.spec.ts
@@ -26,6 +26,7 @@ import sessionServiceClient from "shared/api//sessionServiceInstance";
 import { VirtualStudy, VirtualStudyData } from "shared/model/VirtualStudy";
 import client from "shared/api/cbioportalClientInstance";
 import AccessorsForOqlFilter, {getSimplifiedMutationType} from "../../shared/lib/oql/AccessorsForOqlFilter";
+import { AlteredStatus } from "./mutualExclusivity/MutualExclusivityUtil";
 
 describe("ResultsViewPageStoreUtils", ()=>{
     describe("computeCustomDriverAnnotationReport", ()=>{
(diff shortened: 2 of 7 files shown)


src/pages/resultsView/ResultsViewPageStoreUtils.spec.ts
```diff
@@ -1707,6 +2316,74 @@ describe('getSampleAlteredMap', () => {
         assert.deepEqual(ret["RAS"], expectedResult["RAS"], "merged gene track(with group name)");
 
     });
+
+    it('should even handle not profiled samples correctly', () => {
```
adamabeshouse: can you rewrite this to explain what it means for it to handle it correctly? i.e. "not profiled samples are excluded" or is it "not profiled samples are considered unaltered" ?

dippindots: @adamabeshouse How about "should set undefined for the not profiled samples"?

adamabeshouse: perfect

src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.tsx
```diff
@@ -10,15 +10,18 @@ export function calculateAssociation(logOddsRatio: number): string {
...
+export function countOccurences(valuesA: (boolean | undefined)[], valuesB: (boolean | undefined)[]): [number, number, number, number] {
 
     let neither = 0;
     let bNotA = 0;
     let aNotB = 0;
     let both = 0;
 
     valuesA.forEach((valueA, index) => {
-
+        // jump over this comparison if there is undefined value(not profiled value)
```
alisman: @dippindots i don't see a test that exercises this new logic.  would seem like it would have to assert against the counts returned by this function.

dippindots: @alisman Yes, I should add the tests for this function, will update soon.

src/pages/resultsView/mutualExclusivity/MutualExclusivityUtil.spec.tsx
```diff
@@ -123,6 +123,19 @@ describe("MutualExclusivityUtil", () => {
...
+            assert.deepEqual(countOccurences([undefined, undefined, true, false], [true, true, false, false]), [1, 0, 1, 0]);
+        });
+
+        it("returns [0, 0, 0, 0] for [undefined, undefined, true, false] and [true, true, undefined, undefined]", () => {
+            assert.deepEqual(countOccurences([undefined, undefined, true, false], [true, true, undefined, undefined]), [0, 0, 0, 0]);
+        });
+        
+        it("returns [0, 0, 0, 0] for [undefined, undefined, undefined, undefined] and [true, true, false, false]", () => {
+            assert.deepEqual(countOccurences([undefined, undefined, undefined, undefined], [true, true, false, false]), [0, 0, 0, 0]);
+        });
```
dippindots: @alisman The new logic tests have been updated here.

alisman: great

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -988,9 +988,9 @@ export class ResultsViewPageStore {
 
 
     readonly isSampleAlteredMap = remoteData({
-        await: () => [this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine, this.samples],
+        await: () => [this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine, this.samples, this.coverageInformation, this.selectedMolecularProfiles],
```
alisman: @adamabeshouse does it make sense to exclude non profile samples here?  would that be assumed by a consumer of isSampleAlteredMap?  It seems like it makes sense.     If a sample is not profiled according to currently selected molecular profile, we should not consider it not-altered, right?  Or should value type be updated to include different states, i.e. true, false, unprofiled?   Just making sure we're doing the right thing here.

adamabeshouse: I think you have a point and it would be future-proof and most complete to do it like that, with an enum like Altered, Unaltered, Unprofiled

src/pages/staticPages/tools/oncoprinter/OncoprinterStore.ts
```diff
@@ -290,9 +291,11 @@ export default class OncoprinterStore {
...
                     const datum = sampleToDatum[sampleId];
                     if (!datum) {
-                        return false;
+                        return CoverageStatus.UNPROFILED;
+                    } else if (datum.data.length > 0) {
+                        return CoverageStatus.ALTERED;
                     } else {
-                        return datum.data.length > 0;
+                        return CoverageStatus.UNALTERED;
                     }
```
dippindots: Hi @adamabeshouse , could I get your review on this function in the Oncoprinter? Thanks!

adamabeshouse: looks great! except I think `CoverageStatus` should really be `AlteredStatus`. what do you think?

dippindots: That's also one thing I am gonna ask, yes I think that's great. I can change it to `AlteredStatus`

dippindots: Updated!

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -410,23 +412,51 @@ export function doesQueryHaveCNSegmentData(
...
+                return _.some(_.map(selectedMolecularProfileIds, (selectedMolecularProfileId) => {
+                    // if not profiled in every genes, then the sample is not profiled, or we think it is profiled
+                    return _.every(_.map(groupGenes, (gene) => {
+                        return isSampleProfiled(sampleKey, selectedMolecularProfileId, gene, coverageInformation);
+                    }), (profiled) => profiled === false);
+                }) , (notProfiled) => notProfiled === true);
+            });
             result[getMultipleGeneResultKey(groupedOql)] = samples.map((sample: Sample) => {
-                return sampleKeys.includes(sample.uniqueSampleKey);
+                if (unProfiledSampleKeys.includes(sample.uniqueSampleKey)) {
```
alisman: i just realized we should optimize this.    you should first make a map from the array . {  uniqueSampleId: true }
then in loop you can do simple lookup instead of using expensive array.includes.  this will probably be 20 times faster.   since we can tens of thousands of samples, this is actually important.

alisman: i would add that you should only do this when you know N is going to be big and operation itself is expensive.  this is an obvious case.   you do the work of indexing the samples once up front and that saves you from using .includes  over and over again. (includes is just doing a full loop)

dippindots: Thanks! I think I can make `sampleKeys` and `unProfiledSampleKeys` to `sampleKeysMap` and `unProfiledSampleKeysMap`

adamabeshouse: a handy method for this is `_.keyBy(listOfStrings)` returns `{ string1:"string1", string2:"string2", ...}` then you can check `sample in myMap`

PR #2290, opened 2019-05-06
use aws beta env

inodb:
we want to retire public-tomcat on dashi/dashi2

### DIFF
--- a/src/shared/api/generated/CBioPortalAPIInternal.ts
+++ b/src/shared/api/generated/CBioPortalAPIInternal.ts
@@ -183,6 +183,16 @@ export type CountSummary = {
         'profiledCount': number
 
 };
+export type DataAccessToken = {
+    'creation': string
+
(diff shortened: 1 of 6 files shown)


src/shared/api/generated/GenomeNexusAPI.ts
```diff
@@ -2409,7 +2409,7 @@ export default class GenomeNexusAPI {
                 return response.body;
```
inodb: genome nexus also got a recent update

src/shared/api/generated/CBioPortalAPIInternal.ts
```diff
@@ -182,6 +182,16 @@ export type CountSummary = {
 
         'profiledCount': number
 
+};
+export type DataAccessToken = {
```
inodb: beta instance has api tokens

PR #2293, opened 2019-05-06
gc survival - (togglable) curves for each overlap region

adamabeshouse:
image

image

### DIFF
--- a/src/pages/groupComparison/GroupComparisonStore.ts
+++ b/src/pages/groupComparison/GroupComparisonStore.ts
@@ -38,7 +38,7 @@ import {PatientSurvival} from "shared/model/PatientSurvival";
 import request from "superagent";
 import {getPatientSurvivals} from "pages/resultsView/SurvivalStoreHelper";
 import {SURVIVAL_CHART_ATTRIBUTES} from "pages/resultsView/survival/SurvivalChart";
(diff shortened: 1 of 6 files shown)


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -706,39 +691,41 @@ export default class GroupComparisonStore {
...
-                });
-                return acc;
-            }, {} as { [patientKey: string]: string[] });
-            return Promise.resolve(patientToAnalysisGroups);
+        invoke:()=>{
+            const partitionMap = new ComplexKeyGroupsMap<string>();
+            const patientToSamplesSet = this.patientToSamplesSet.result!;
+            const groupToPatientKeys = this._originalGroups.result!.reduce((map, group)=>{
+                map[group.uid] = _.keyBy(getPatientIdentifiers([group]).map(id=>{
+                    return patientToSamplesSet.get({ studyId: id.studyId, patientId: id.patientId })![0].uniquePatientKey;
```
kalletlak: may be use `patientToSamplesSet.get(id)` since id has both `studyId` and `patientId`.

adamabeshouse: Good point

PR #2298, opened 2019-05-08
Changed how the font.json file was loaded to reduce the final webpack…

lemccarthy:
… bundle size


src/shared/lib/svgToPdfDownload.ts
```diff
@@ -1,5 +1,6 @@
 import svg2pdf from "svg2pdf.js";
 import {jsPDF} from "jspdf-yworks";
+import font from "font.json";
```
alisman: lets change the name of this file to something more descriptive.  like pdf-font.   or perhaps the name of the actual font.

vendor-bundles.webpack.config.js
```diff
@@ -37,6 +40,10 @@ let config = {
     devtool : 'source-map'
 };
 
+config.resolve = {
+    modules: [path.join(__dirname, "src", "common"), "node_modules"]
```
alisman: i doubt you need __dirname here.   if we needed it, wouldn't it be necessary for node_modules as well?

src/shared/lib/svgToPdfDownload.ts
lemccarthy: Got it, changed the name.

vendor-bundles.webpack.config.js
lemccarthy: Ok, removed

src/shared/lib/svgToPdfDownload.ts
```diff
@@ -1,7 +1,10 @@
 import svg2pdf from "svg2pdf.js";
 import {jsPDF} from "jspdf-yworks";
+import font from "FreeSans.json";
 import _ from "lodash";
 
+console.log(font);
```
alisman: kill console

alisman:
@jjgao merging this with 5 unrelated e2e tests failing

PR #2300, opened 2019-05-09
use AWS rc env

inodb:


inodb:
oncokb api sync is failing, but that's already been addressed in master branch

PR #2303, opened 2019-05-09
Fix CircleCI caching e2e+unit tests

inodb:
the e2e and unit tests were also cached based on the source code changes instead of just the compiled artifacts


inodb:
the caching was fine - the issue was something else related to using `this.retries` in an arrow function. See e.g.: https://github.com/mochajs/mocha/issues/2018

PR #2305, opened 2019-05-09
Adjust loading animation and page layout

alisman:


src/appBootstrapper.jsx
```diff
@@ -58,8 +60,6 @@ if (browser) {
     });
 }
 
-const moo = "aaron";
```
adamabeshouse: why delete this?

alisman: it really is elegant.

PR #2307, opened 2019-05-10
Update column headers in enrichments table

alisman:


alisman:
@kalletlak will this change clash with changes you are making to enrichments table to support >2 groups?  
Note: this needs to be made to mrna as well

alisman:
This will be outdated by @kalletlak's changes to support >2 groups comparisons

PR #2311, opened 2019-05-13
Genie fusion issue 5613

khzhu:
# What? Why?
Fix #

Decouple fusions from mutations should fix the [issue.](https://github.com/cBioPortal/cbioportal/issues/5613) 

- [x] add a new Fusion table without Frequency column in the study view to show all fusions currently recorded in the mutation database table.

# Checks
  the `-s` option)

# Any screenshots or GIFs?

# Notify reviewers

@cBioPortal/frontend

and notify them here:


src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -271,8 +271,8 @@ export default class LollipopMutationPlot extends React.Component<ILollipopMutat
 
     @computed private get domainMap(): {[pfamAccession:string]: PfamDomain}
     {
-        if (!this.props.store.pfamDomainData.isPending && 
-            this.props.store.pfamDomainData.result && 
+        if (!this.props.store.pfamDomainData.isPending &&
```
alisman: lets revert this

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1054,6 +1056,9 @@ export class StudyViewPageStore {
                 case ChartTypeEnum.MUTATED_GENES_TABLE:
                     this.resetGeneFilter();
                     break;
+                case ChartTypeEnum.FUSION_GENES_TABLE:
+                    this.resetGeneFilter();
```
alisman: just confirming that we don't have separate FusionGene filter?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2479,6 +2506,22 @@ export class StudyViewPageStore {
         default: []
     });
 
+    readonly fusionGeneData = remoteData<FusionGenesData>({
+        await: () => [this.mutationProfiles],
+        invoke: async () => {
+            if (!_.isEmpty(this.mutationProfiles.result!)) {
```
alisman: not too important, but you might not have to use ! after result here.  one would think that isEmpty typing would accept undefined since it's meant to test for undefined (among other things).

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2479,6 +2506,22 @@ export class StudyViewPageStore {
...
+                // TODO: get data for all profiles
```
alisman: not sure if i understand the TODO

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
khzhu: will do.

src/pages/studyView/table/FusionGenesTable.tsx
```diff
@@ -0,0 +1,310 @@
...
+enum ColumnKey {
+    GENE = 'Gene',
+    NUMBER_MUTATIONS = '# Mut',
+    NUMBER = '#'
+}
+
+class FusionGenesTableComponent extends FixedHeaderTable<MutationCountByGene> {
+}
+
+@observer
```
alisman: @khzhu @zhx828 Just for discusion but do you think there's opportunity for reusing more code (super class?) between these different table types?  i would say we should do it only if it's obvious that we're copying and pasting a lot.  Or if there are going to be a lot more table types.

src/pages/studyView/table/FusionGenesTable.tsx
```diff
@@ -0,0 +1,310 @@
...
+        return {
+            [ColumnKey.GENE]: correctColumnWidth(this.props.width * 0.35),
+            [ColumnKey.NUMBER_MUTATIONS]: correctColumnWidth(this.props.width * 0.25),
+            [ColumnKey.NUMBER]: correctColumnWidth(this.props.width * 0.25)
+        };
+    }
+
+    @autobind
+    @action
+    updateCellMargin() {
```
alisman: this seems like it could be shared b/c there's nothing implementation specific.  or is there?

src/pages/studyView/StudyViewPageStore.ts
khzhu: we do not. Fusions are not properly tracked in the database at the moment. This is really a temporary solution to address GENIE study mutations over counting problem.

src/pages/studyView/StudyViewPageStore.ts
khzhu: I see. thanks for pointing this out. Will remove !.

src/pages/studyView/StudyViewPageStore.ts
khzhu: sorry, I forgot too. It has been really a while, will back to you later.

src/pages/studyView/StudyViewPageStore.ts
khzhu: update: discussed with @zhx828, added a new gene filter for fusions.

src/pages/studyView/StudyViewPageStore.ts
khzhu: ! removed.

src/pages/studyView/table/FusionGenesTable.tsx
khzhu: @JJ, @alisman, @zhx828, 
how will fusion table be displayed once structure variant PR merged in, will still be as same as mutation table? 
I am re-thinking how to address Aaron's above comment to factor gene tables. CNAGenesTable and MutationGenesTable are a bit different since we have cytobands and alterations in the CNAGenesTable, while have Frequencies in the MutationGenesTable, makes sense to give them their own tables. But new fusionGenesTable is merely a copy of Mutation table, since it does not have its own data model at the moment. If FusionGenesTable is always as same as MutationGenesTable, it will make more sense to have a super class for them. 
Screen Shot 2019-08-19 at 5 42 39 PM

jjgao: @khzhu we may keep it the same way after using the SV table for the first iteration. In the long run, we may want to to find a way to display the partner genes, but let's not worry about it for now.

khzhu: thanks, @JJ. 
I vote for having FusionGenesTable by its own for the future so we could easily add partner genes and possible remove Freq coloumn(do not think is useful). What do you think @alisman?

khzhu: here are three tables side by side
Screen Shot 2019-08-20 at 11 54 06 AM

src/pages/studyView/StudyViewPageStore.ts
zhx828: I will remove these TODOs. I placed there as a reminder for a feature which should be transferred to a proper issue on github.

src/pages/studyView/table/FusionGenesTable.tsx
zhx828: @alisman these tables are using the same component FixedHeaderTable. There is no much of space to create a super class for these study view tables since they are highly customized. But I do think there are more space to move some of the functions to utils so they can be shared.

src/shared/components/lollipopMutationPlot/LollipopMutationPlot.tsx
khzhu: reverted.

src/pages/studyView/table/FusionGenesTable.tsx
khzhu: moved some common interface/type to TabUtil file for share.

src/pages/studyView/table/FusionGenesTable.tsx
khzhu: no longer use updateCellMargin().

src/pages/studyView/StudyViewConfig.ts
```diff
@@ -77,6 +78,7 @@ export enum ChartTypeNameEnum {
     TABLE = 'table',
     SCATTER = 'density plot',
     MUTATED_GENES_TABLE = 'table',
+    FUSION_GENES_TABLE = 'table',
```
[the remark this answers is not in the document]
zhx828: Updated

src/pages/studyView/table/FusionGenesTable.tsx
zhx828: @alisman I merged three tables to one component. It's not a super class, since we already using FixedHeaderTable. Essentially, it's just a util component including all the column definitions.

zhx828:
@kalletlak @alisman there are more space to merge the settings for these three tables. But I want to leave it out from this pr since it's already getting too big. I will send a separate pr for that.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1216,12 +1246,49 @@ export class StudyViewPageStore {
...
     }
 
+    @autobind
+    @action
+    addFusionGeneFilters(uniqueKeys: string[]) {
+        const genes = uniqueKeys.map(uniqueKey => parseMutationUniqueKey(uniqueKey));
+
+        trackStudyViewFilterEvent("geneFilter", this);
+
+        genes.forEach(gene => this.geneMapCache[gene.entrezGeneId] = gene.hugoGeneSymbol);
```
alisman: i think "genemapCache" name could be improved.  Looks like it's entrezGeneId to hugoGeneSymbol?  why is that a cache?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1216,12 +1246,49 @@ export class StudyViewPageStore {
...
+        this._fusionGeneFilter = [...this._fusionGeneFilter, {entrezGeneIds: genes.map(gene => gene.entrezGeneId)}];
+    }
+
+    @autobind
+    @action
+    removeFusionGeneFilter(toBeRemoved: number) {
+        this._fusionGeneFilter = _.reduce(this._fusionGeneFilter, (acc, next) => {
+            const newGroup = _.reduce(next.entrezGeneIds, (list, entrezGeneId) => {
```
alisman: isn't this just a filter operation?

src/pages/studyView/StudyViewPageStore.ts
khzhu: agree. it is just a normal map.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3309,18 +3439,42 @@ export class StudyViewPageStore {
             return '';
     }
 
+    public async getFusionGenesDownloadData() {
```
alisman: I don't think this needs to be async as it doesn't seem to do any fetching/awaiting.  This function I think is really worthy of a test.

src/pages/studyView/TableUtils.tsx
```diff
@@ -47,27 +86,49 @@ export function getCancerGeneFilterToggleIcon(isFilteredByCancerGeneList:boolean
...
-                    (numberOfAlteredCases / numberOfSamplesProfiled) * 100
-                )}
-            </span>
+            {getCellContent()}
         </DefaultTooltip>
     );
 }
+
+export function rowIsChecked(entrezGeneId: number, preSelectedRows: AlteredGenesTableUserSelectionWithIndex[], selectedRows: AlteredGenesTableUserSelectionWithIndex[]) {
+    return !!_.find(
```
alisman: super minor but i thinke _.some would be more semantic.  it returns boolean so wouldn't require the !!

src/pages/studyView/table/GenePanelModal.tsx
```diff
@@ -49,26 +49,30 @@ export const GenePanelList: React.FunctionComponent<IGenePanelTooltipProps> = ({
...
+                            ))}
+                        </div>}
+                    </Modal.Body>
+                    <Modal.Footer>
+                        <Button onClick={this.props.hide}>Close</Button>
+                    </Modal.Footer>
+                </Modal>
+            );
+        } else {
+            return <span />;
```
alisman: null?

src/pages/studyView/StudyViewPageStore.ts
khzhu: agree. no need for async here.

src/pages/studyView/StudyViewPageStore.ts
zhx828: @alisman yeah, we were trying to emphasize the purpose of the variable

kalletlak: this won't be needed anymore once hugoGeneSymbols are supported. removed in my pr https://github.com/cBioPortal/cbioportal-frontend/pull/2714/files#diff-a3ffe2439b6c26eb3cbf8050e37410a8L703

zhx828: @kalletlak @alisman  just in case I don't introduce more conflicts. I will leave the name as it.

src/pages/studyView/TableUtils.tsx
zhx828: Good point.

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -0,0 +1,521 @@
...
+        }, {} as { [key in GeneTableColumnKey]: number });
+    }
+
+    @computed get tableData() {
+        return this.isFilteredByCancerGeneList ? _.filter(this.props.promise.result, data => data.isCancerGene) : (this.props.promise.result || []);
+    }
+
+    @computed
+    get tableColumns() {
+        return _.reduce(this.props.columns, (acc, column) => {
```
kalletlak: why not use `map`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -768,6 +769,18 @@ export class StudyViewPageStore {
             }, [] as MutationGeneFilter[]);
         }
 
+        if (_.isArray(filters.fusionGenes) && filters.fusionGenes.length > 0) {
+            this._fusionGeneFilter = _.reduce(filters.fusionGenes, (acc, next) => {
```
alisman: i can see that this logic follows pattern of existing blocks in use of _.reduce, but worth noting that these reduces are really mapping operations.  would be more semantic to use map.   reduce should only be used if there's a conditional inclusion (filter would also work) or if we're producing a map from an array.     also, we can now use native reduce, map, filter methods of array object.  lodash unnecessary in these cases.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1216,12 +1246,44 @@ export class StudyViewPageStore {
...
+
+        trackStudyViewFilterEvent("geneFilter", this);
+
+        genes.forEach(gene => this.geneMapCache[gene.entrezGeneId] = gene.hugoGeneSymbol);
+        this._fusionGeneFilter = [...this._fusionGeneFilter, {entrezGeneIds: genes.map(gene => gene.entrezGeneId)}];
+    }
+
+    @autobind
+    @action
+    removeFusionGeneFilter(toBeRemoved: number) {
```
alisman: toBeRemoved = entrezIdToBeRemoved

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2397,6 +2470,18 @@ export class StudyViewPageStore {
             };
         }
 
+        if (!_.isEmpty(this.mutationProfiles.result)) {
```
alisman: just want to double check that this is the right condition for including fusion table?  (mutationProfiles not empty which is used by MUTATED_GENES_TABLE as well)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2602,6 +2689,8 @@ export class StudyViewPageStore {
 
             if (chartUserSettings.chartType === UniqueKey.MUTATED_GENES_TABLE) {
                 this._filterMutatedGenesTableByCancerGenes = chartUserSettings.filterByCancerGenes === undefined ? true : chartUserSettings.filterByCancerGenes;
+            } else if (chartUserSettings.chartType === UniqueKey.FUSION_GENES_TABLE) {
```
alisman: seems like this if else if else structure should really be a switch statement.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2995,19 +3092,47 @@ export class StudyViewPageStore {
...
+            if (!_.isEmpty(this.mutationProfiles.result)) {
                 let mutatedGenes = await internalClient.fetchMutatedGenesUsingPOST({
                     studyViewFilter: this.filters
                 });
                 return mutatedGenes.map(item => {
                     return {
                         ...item,
+                        uniqueKey: getMutationUniqueKey(item.entrezGeneId, item.hugoGeneSymbol),
+                        oncokbAnnotated: this.oncokbCancerGeneFilterEnabled ? this.oncokbAnnotatedGeneEntrezGeneIds.result.includes(item.entrezGeneId) : false,
+                        oncokbOncogene: this.oncokbCancerGeneFilterEnabled ? this.oncokbOncogeneEntrezGeneIds.result.includes(item.entrezGeneId) : false,
```
alisman: if these values are boolean, i think the names should reflect it.    e.g. isOncoKBOncogene

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3023,19 +3148,19 @@ export class StudyViewPageStore {
         default: []
     });
 
-    readonly cnaGeneData = remoteData<CopyNumberCountByGeneWithCancerGene[]>({
+    readonly cnaGeneData = remoteData<GeneTableRow[]>({
```
alisman: Is this a TableRow in the sense of a database row or a <html> table row?    If the latter, i think name of the promise (cnaGeneData) should reflect that.  "cnaGeneTableRowData"

src/pages/studyView/TableUtils.tsx
```diff
@@ -47,27 +87,49 @@ export function getCancerGeneFilterToggleIcon(isFilteredByCancerGeneList:boolean
...
+            <span>{`# of samples profiled for ${detailedTypeInfo} in this gene: ${numberOfSamplesProfiled.toLocaleString()}`}</span>
             <GenePanelList
                 genePanelIds={matchingGenePanelIds}
-                toggleModal={toggleModal}
+                toggleModal={toggleModal!}
             />
         </span>
     );
+
+    function getCellContent() {
```
alisman: kinda strange pattern here.   uses style variable from closure.  i would rather just see this inside of the markup as it was

src/pages/studyView/chartHeader/ChartHeader.tsx
```diff
@@ -40,7 +40,7 @@ export interface IChartHeaderProps {
     chartControls?   : ChartControls;
     changeChartType  : (chartType: ChartType) => void;
     getSVG?          : ()=>Promise<SVGElement | null>;
-    getData?         : (dataType?:DataType)=>Promise<string | null>;
+    getData?         : ((dataType?:DataType)=>Promise<string | null>) | ((dataType?:DataType)=>string);
```
alisman: i'm curious why?   seems like code will now have to check whether it's a promise or not, no?

src/pages/studyView/table/GenePanelModal.tsx
```diff
@@ -42,33 +42,37 @@ export const GenePanelList: React.FunctionComponent<IGenePanelTooltipProps> = ({
...
-                            <p key={gene.entrezGeneId}>{gene.hugoGeneSymbol}</p>
-                        ))}
-                    </div>}
-                </Modal.Body>
-                <Modal.Footer>
-                    <Button onClick={this.props.hide}>Close</Button>
-                </Modal.Footer>
-            </Modal>
-        );
+        if (this.props.genePanelCache) {
```
alisman: seems more natural to do this outside of the modal.

src/pages/studyView/StudyViewPageStore.ts
zhx828: @alisman Won't the map alter the original array?

src/pages/studyView/chartHeader/ChartHeader.tsx
zhx828: Not sure what's the best way here. There are some getData methods don't require a promise to produce the data. If we don't provide the second type, then these methods have to attach an async tag.

src/pages/studyView/table/GenePanelModal.tsx
zhx828: I will update. I guess the original implementer was trying to avoid additional checks which have been used in three components. But now we merge these three together, so we should do the check at the place where initialize the component.

src/pages/studyView/StudyViewPageStore.ts
zhx828: Sorry, it was for another scenario. In this case, it will not. Will update. Thanks.

zhx828:
The two coexpression tests on master env just impossible to finish. Need to double check before merging back to master

PR #2313, opened 2019-05-13
Support opening comparison session for the cancer studies chart

adamabeshouse:
Also default group names

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -454,17 +454,16 @@ export class StudyViewPageStore {
         statusCallback(LoadingPhase.DOWNLOADING_GROUPS);
         // for now, the only one possible is cancer studies
         return new Promise<string>(resolve=>{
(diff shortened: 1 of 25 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -602,15 +551,15 @@ export class StudyViewPageStore {
 
     @autobind
     public async openComparisonPage(params:{
-        type:ChartTypeEnum.PIE_CHART|ChartTypeEnum.TABLE|ChartTypeEnum.BAR_CHART,
-        clinicalAttribute: ClinicalAttribute,
+        chartMeta: ChartMeta,
+        clinicalAttribute?: ClinicalAttribute,
```
kalletlak: I think there is no need to pass `clinicalAttribute` as `chartMeta` would be having it.

alisman:
@adamabeshouse what is a "special" chart?   Could we improve that name or explain it better in comments?

adamabeshouse:
@alisman it was named that way when I got here. It means a chart which does not come from a server-fetched clinical attribute. we can rename it to something like "local clinical attribute" if that's clearer?

PR #2316, opened 2019-05-14
Optimize group combinations calculation for Upset

adamabeshouse:
Improves the exponential factor of running time from 2^(number of groups) to 2^(maximum number of groups that share a sample), which hopefully most of the time will not be very big, and in particular in queries that come out of a large pie chart it is always =1 for the mutually exclusive groups.

### DIFF
--- a/src/pages/groupComparison/OverlapUtils.spec.tsx
+++ b/src/pages/groupComparison/OverlapUtils.spec.tsx
@@ -4,6 +4,8 @@ import expectJSX from 'expect-jsx';
 import * as React from "react";
 expect.extend(expectJSX);
 import {
+    getAllCombinationsOfKey,
+    getCombinations,
     getExcludedIndexes,
     getStudiesAttrForPatientOverlapGroup,
     getStudiesAttrForSampleOverlapGroup,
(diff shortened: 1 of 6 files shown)


src/pages/groupComparison/OverlapUtils.spec.tsx
```diff
@@ -16,6 +18,95 @@ import {Sample} from "../../shared/api/generated/CBioPortalAPI";
...
+        });
+        it("three", ()=>{
+            assertDeepEqualInAnyOrder(getAllCombinationsOfKey({ a:true, b:true, c:true}), [
+                {a:true}, {b:true}, {a:true, b:true}, {c:true},
+                {a:true, c:true}, {b:true, c:true}, {a:true, b:true, c:true}
+            ]);
+        });
+    });
+    describe('getCombinations', () => {
+        it('when empty groups', () => {
```
adamabeshouse: I just copied these tests over

src/pages/groupComparison/OverlapUtils.tsx
```diff
@@ -123,4 +124,52 @@ export function getStudiesAttrForPatientOverlapGroup(
         studiesAttr = unionPatients(studiesAttr, regionStudiesAttr);
     }
     return convertPatientsStudiesAttrToSamples(studiesAttr, patientToSamplesSet);
+}
+
```
adamabeshouse: this file is where all the action is

PR #2319, opened 2019-05-14
fix two screenshot tests

alisman:
Fix two screenshot tests


alisman:
@adamabeshouse this seems to work in fixing the screenshots under question

PR #2320, opened 2019-05-14
Updates to enrichment tab plots and table

kalletlak:

### DIFF
--- a/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
+++ b/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
@@ -162,6 +162,13 @@ export default class AlterationEnrichmentContainer extends React.Component<IAlte
         return [];
     }
 
+    @computed get scatterPlotLabels() {
+        if(this.props.groups.length === 2) {
+            return [this.props.groups[0].name, this.props.groups[1].name];
+        }
+        return [];
+    }
+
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -162,6 +162,13 @@ export default class AlterationEnrichmentContainer extends React.Component<IAlte
         return [];
     }
 
+    @computed get scatterPlotLabels() {
+        if(this.props.groups.length === 2) {
+            return [this.props.groups[0].name, this.props.groups[1].name];
+        }
+        return [];
```
alisman: should this throw an error because it should not be called if there are more than two groups?

kalletlak: those plots are not show if its more than 2

alisman: right so what i'm saying is rather than return empty, we could throw an error in this situation.  not super important, but just a thought.  i mean, empty array seems wrong. though i guess typescript would complain if it didn't return anything.

alisman: this is not important, just thinking about best practice

PR #2323, opened 2019-05-15
Update GN client

zhx828:


zhx828:
@inodb @alisman I saw these six tests failing in my another pr as well. So I figure they are not related to this pr. Merging.

PR #2325, opened 2019-05-15
Only show cancer study column in clinical data tab if the chart/column is visible

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6098


zhx828:
The failing e2e are not related to this pr.

PR #2327, opened 2019-05-15
Enable custom selection box for CNA Segments tab

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6101


zhx828:
Failing e2e are note related to this pr.

PR #2328, opened 2019-05-15
group comparison adjustments

adamabeshouse:
(1) survival tab should only use selected groups for multiple-group curves
(2) tab overlap warnings specify precisely whats filtered out - samples, patients, or both

### DIFF
--- a/src/pages/groupComparison/OverlapExclusionIndicator.tsx
+++ b/src/pages/groupComparison/OverlapExclusionIndicator.tsx
@@ -5,6 +5,7 @@ import {caseCounts} from "./GroupComparisonUtils";
 
 export interface IOverlapExclusionIndicatorProps {
     store:GroupComparisonStore;
+    only?:"sample"|"patient"
 }
 
 @observer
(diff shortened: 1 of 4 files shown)


src/pages/groupComparison/OverlapExclusionIndicator.tsx
```diff
@@ -5,6 +5,7 @@ import {caseCounts} from "./GroupComparisonUtils";
 
 export interface IOverlapExclusionIndicatorProps {
     store:GroupComparisonStore;
+    only?:"sample"|"patient"
```
alisman: how about onlyCaseType?

src/pages/groupComparison/OverlapExclusionIndicator.tsx
```diff
@@ -14,14 +15,32 @@ export default class OverlapExclusionIndicator extends React.Component<IOverlapE
...
+                    let count = 0;
+                    if (this.props.only === "sample") {
+                        count = selectionInfo.overlappingSamples.length;
+                    } else {
+                        count = selectionInfo.overlappingPatients.length;
+                    }
+                    const plural = (count !== 1);
+                    caseCountsSummary = `${count} ${this.props.only}${plural ? "s" : ""}`;
+                    break;
+                default:
```
alisman: should there just be a default value for props.only?

adamabeshouse: no, because it's optional and if undefined means its both samples and patients

src/pages/groupComparison/OverlapExclusionIndicator.tsx
adamabeshouse: personally I don't find that clearer

PR #2330, opened 2019-05-16
Add Gene Box Plot in enrichment analysis

kalletlak:
# What? Why?
Add remaining part of https://github.com/cBioPortal/cbioportal/issues/5990

In result page
Screen Shot 2019-05-16 at 4 46 17 PM

In comparison page
Screen Shot 2019-05-16 at 4 46 31 PM


src/globalStyles/global.scss
```diff
@@ -508,6 +508,7 @@ h3 {
   border:1px dashed $borderColor;
   position:relative;
   max-width:100%;
+  height:100%;
```
alisman: are you sure this does something?

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -400,4 +406,109 @@ export function getGroupColumns(groups: { name: string, description: string }[],
         });
     });
     return columns;
+}
+
+export function getEnrichmentBarPlotData(data: AlterationEnrichmentRow[], genes: string[]): IMultipleCategoryBarPlotData[] {
+    const usedMajorCategories: {[gene:string]:boolean} = {};
```
alisman: if it's category = gene, should we just called it usedGenes?

jjgao:
Thanks, @kalletlak. Looks great.

Could you change the order of bars to reflect the order of groups from left to right?

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -400,4 +406,109 @@ export function getGroupColumns(groups: { name: string, description: string }[],
...
+                }
+                acc[groupName][datum.hugoGeneSymbol] = group.alteredCount;
+                usedMajorCategories[datum.hugoGeneSymbol] = true;
+            }
+        });
+        return acc;
+    }, {} as { [minorCategory: string]: { [majorCategory: string]: number } });
+
+    const majorCategoryTotalCounts: { [id: string]: number } = {}
+    // ensure entries for all used minor categories - we need 0 entries for those major/minor combos we didnt see
```
alisman: i don't get what minor/major categories means from the code.   i think it needs to be more domain specific naming

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
adamabeshouse: +1 on domain specific naming. I always thought major/minor category language was pretty confusing but I couldnt think of a better way to describe it in the abstract setting

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -94,15 +94,16 @@ export function makeBarSpecs(
...
     percentage:boolean
 ):{
     fill: string,
     data:{ x:number, y: number, majorCategory: string, minorCategory:string, count:number, percentage:number}[] // one data per major category, in correct order - either specified, or alphabetical
 }[] // one bar spec per minor category, in correct order - either specified, or alphabetical 
 {
     data = sortDataByCategory(data, d=>d.minorCategory, minorCategoryOrder);
-    // for vertical bars, reverse order, to put first on top
-    if (!horizontalBars) {
+    // reverse the order of stacked or horizontal bars
```
adamabeshouse: its really stacked xor horizontal

PR #2331, opened 2019-05-16
Fix e2e

adamabeshouse:


adamabeshouse:
@alisman seems like 3 is the highest number of threads that it can tolerate consistently

PR #2335, opened 2019-05-17
Update to gene plot in enrichment analysis

kalletlak:
# What? Why?
Fix
1. Indicate significant genes
2. number of genes in dropdown
3. correctly include order of genes show in box plot
4. show percentages instead of samples count for box plot
5. remove plot type. only show grouped bar plot
6. update Bar tooltip
Screen Shot 2019-05-20 at 10 34 39 AM


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -408,22 +408,25 @@ export function getGroupColumns(groups: { name: string, description: string }[],
...
+            _.each(datum.groupsSet, group => {
+                const groupName = group.name;
                 if (!acc[groupName]) {
                     acc[groupName] = {};
                 }
-                acc[groupName][datum.hugoGeneSymbol] = group.alteredCount;
-                usedGenes[datum.hugoGeneSymbol] = true;
-            }
-        });
+                const displayedGene = gene + (datum.qValue < 0.05 ? '*' : '')
```
alisman: displayedGeneName?   also, missing semicolon

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -408,22 +408,25 @@ export function getGroupColumns(groups: { name: string, description: string }[],
...
+                acc[groupName][displayedGene] = group.alteredPercentage;
```
alisman: is there any way you could add the * at render time?  or make a property of this object called displayedGeneName, rather than use the asterisked name as a key?

src/pages/resultsView/enrichments/GeneBarPlot.tsx
```diff
@@ -76,57 +80,50 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}>
...
 
     @autobind
-    @action
-    private onPlotTypeSelect(option: any) {
-        this.plotType = option.value
-    }
+    private getTooltip(datum: any) {
+        let geneSymbol = datum.majorCategory as string;
+        //remove * from end of gene symbol if any
+        if (geneSymbol.endsWith("*")) {
```
alisman: unfortunately IE11 doesn't support endsWith and it's not polyfilled as far as I can see.  Probably simplest just to use /*?/.test() for now

alisman: actually, instead of testing for this you could just do a one liner.  geneSymbol = genSymbo.replace(/*$/,"")

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
kalletlak: this would involve lot of changes `MultipleCategoryBarPlot` component.

PR #2338, opened 2019-05-20
Improve scrolling/sizing behavior of gene bar plott

alisman:
The gene bar chart in Enrichments tab needs to scroll when there is not enough horizontal space

### DIFF
--- a/src/pages/resultsView/enrichments/GeneBarPlot.tsx
+++ b/src/pages/resultsView/enrichments/GeneBarPlot.tsx
@@ -177,10 +177,9 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}> 
             width = 600;
         }
         return (
-            <div data-test="ClinicalTabPlotDiv" className="borderedChart" style={{ width, overflow: "hidden" }}>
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/enrichments/GeneBarPlot.tsx
```diff
@@ -177,10 +177,9 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}>
             width = 600;
         }
         return (
-            <div data-test="ClinicalTabPlotDiv" className="borderedChart" style={{ width, overflow: "hidden" }}>
-                <ScrollBar style={{ top: -5 }} getScrollEl={this.getScrollPane} />
+            <div data-test="ClinicalTabPlotDiv" className="borderedChart" style={{ overflowX:'auto', overflowY:'hidden', flexGrow:1  }}>
                 {this.toolbar}
-                <div ref={this.assignScrollPaneRef} style={{ position: "relative", display: "inline-block" }}>
+                <div style={{ position: "absolute" }}>
```
alisman: this is necessary to get the container to "collapse" around the chart inside of it

PR #2342, opened 2019-05-21
GC: add screenshot test to overlap upset, survival and clinical tab

kalletlak:


kalletlak:
@adamabeshouse added

kalletlak:
@dippindots done, removed tooltip

PR #2343, opened 2019-05-21
enrichments - improve axis padding for volcano and frequency plots

adamabeshouse:
image


alisman:
@schultzn @adamabeshouse do you think we need a tick and dashed line to show actual zero? right now, someone might easily conclude that the data points are > 0.

schultzn:
Can we do something like this?

image

adamabeshouse:
@schultzn yes easily in fact thats the victory default behavior that i was working to avoid :p

schultzn:
Maybe this is slightly better?

image

adamabeshouse:
@schultzn isnt that the same as the suggestion you just made? yes ill see about doing that

schultzn:
I modified it slightly - the dashed lines now don’t go all the way to the axes.

adamabeshouse:
ohh I see

PR #2344, opened 2019-05-22
Correct checkbox position

dippindots:
move the checkbox to the top.
Fix #https://github.com/cBioPortal/cbioportal/issues/6127.

- a move to the top:
image


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1514,6 +1514,36 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                     <div>
                         <div data-test="PlotsTabPlotDiv" className="borderedChart posRelative">
                             <ScrollBar style={{position:'relative', top:-5}} getScrollEl={this.getScrollPane} />
+                            <div style={{textAlign:"center", position:"relative", top:-5, marginBottom:"-10px", minWidth: this.mutationDataCanBeShown && this.cnaDataCanBeShown ? 600 : 0}}>
```
dippindots: Changed style at here, and added a min-width just for the case when we show both checkboxes,

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -774,16 +774,11 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
     }
 
     readonly cnaPromise = remoteData({
-        await:()=>{
-            const queries = getCnaQueries(this.horzSelection, this.vertSelection, this.cnaDataShown);
```
dippindots: @adamabeshouse Could I get your review on the cna related lines? I think we are not changing the data when we select the colors, so I removed `cnaDatashown`. Could you help me to confirm this? Thanks!

adamabeshouse: cnaDataShown is there so that we don't actually request any data if cnaDataShown is false. check out getCnaQueries to see how that works. so we do need to revert this change

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1298,7 +1293,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                             molecularProfileIds: _.values(this.props.store.studyToMutationMolecularProfile.result!).map(p=>p.molecularProfileId),
                             data: this.mutationPromise.result!
                         } : undefined,
-                        this.cnaDataShown ? {
+                        this.cnaDataExists.result ? {
```
adamabeshouse: why?

src/pages/resultsView/plots/PlotsTab.tsx
dippindots: @adamabeshouse The problem is when we select `Color Sample By Copy Number Alteration`, the plots will fetch the data again when we uncheck or check the checkbox.
You can see this here: http://www.cbioportal.org/results/plots?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=nsclc_tcga_broad_2016&case_set_id=nsclc_tcga_broad_2016_cnaseq&data_priority=0&gene_list=TP53&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=nsclc_tcga_broad_2016_cna&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=nsclc_tcga_broad_2016_mutations&tab_index=tab_visualize

PR #2348, opened 2019-05-22
Update gene selection in enrichment pages

kalletlak:
# What? Why?
Partial fix https://github.com/cBioPortal/cbioportal/issues/6129


jjgao:
- [ ] update the gene list onblur of the box (or enter)
- [ ] change the text "Genes with..." to "Showing genes with..."
- [ ] remove "Group" in the y-axis label
- [ ] hide the dialog after submit
- [ ] limit to 100 (?) genes
- [ ] should we left align the text and Select Genes and download buttons?

PR #2359, opened 2019-05-24
New default cancer summary

dippindots:
change the default value of `min # total cases` to 10.
Fix # https://github.com/cBioPortal/cbioportal/issues/6132.

- a: change the default value to 10 if there are more than 10 cases.
image

- b: still set 0 if cases less than 10
image

- c: if input value overflow, change to the default value(min / max)
image
image


src/pages/resultsView/cancerSummary/CancerSummaryContent.tsx
```diff
@@ -139,6 +140,11 @@ export class CancerSummaryContent extends React.Component<ICancerSummaryContentP
         this.setPngAnchor = this.setPngAnchor.bind(this);
     }
 
+    componentDidMount() {
+        // update the slider value after min and max computed
+        this.clearSliderValue();
```
alisman: seems odd that we have to clearSliderValue on mount of component

PR #2360, opened 2019-05-24
change checkbox class to flexbox layout

dippindots:
change checkbox class to flexbox layout
Fix # https://github.com/cBioPortal/cbioportal/issues/6151.

- a
image


alisman:
Since this was an existing component, can you check where else we use it and make it continues to looks good.

src/shared/components/FlexAlignedCheckbox.tsx
```diff
@@ -14,14 +14,14 @@ export interface ICheckboxProps {
...
-            <div className="checkbox" style={Object.assign({display:"flex", alignItems:"center"}, this.props.style)}>
+            <div style={Object.assign({display:"flex", alignItems:"center"}, this.props.style)}>
                 <label>
                     <input
                         type="checkbox"
                         value={this.props.value}
                         checked={this.props.checked}
                         onClick={this.props.onClick}
-                        style={{marginTop:1}}
+                        style={{marginTop:1, marginRight:5}}
```
alisman: lets apply all these styles (above too) in css classes.  We shouldn't use inline styles in library code.  If it appears many times on a page, inline css is unperformant.

alisman:
There's a screenshot that needs updating.

Also, lets see if we can get the checkboxes inline with text.  
image

PR #2365, opened 2019-05-28
Gc updates

kalletlak:
# What? Why?
Fix
https://github.com/cBioPortal/cbioportal/issues/6162
https://github.com/cBioPortal/cbioportal/issues/6155
https://github.com/cBioPortal/cbioportal/issues/6152


src/pages/groupComparison/Overlap.tsx
```diff
@@ -19,6 +19,8 @@ export interface IOverlapProps {
 }
 
 const SVG_ID = "comparison-tab-overlap-svg";
+const SAMPLE_SVG_ID = "comparison-tab-overlap-sample-svg";
+const PATIENT_SVG_ID = "comparison-tab-overlap-patinet-svg";
```
adamabeshouse: typo patinet

src/pages/groupComparison/Overlap.tsx
```diff
@@ -57,7 +59,29 @@ export default class Overlap extends React.Component<IOverlapProps, {}> {
 
     @autobind
     private getSvg() {
-        return document.getElementById(SVG_ID) as SVGElement | null;
+        if (this.plotType.result! === PlotType.Upset) {
```
alisman: not sure what all this is about.  why have we left React paradigm?  there some stuff i would improve about the below but i want to first understand the bigger picture.

adamabeshouse: @alisman do you mean instead of using a ref?

alisman: ok, i see.  ok, lets switch to use ref paradigm.  i know it's an "academic" change but i think it's important to avoid anti-patterns that might screw things up when other devs go at it.

alisman: you could use react-dom library to find a components top level svg element

alisman: ```node = ReactDOM.findDOMNode(this);``` where this = instance of component

adamabeshouse: thats a great replacement

PR #2367, opened 2019-05-28
GC - add curve name to survival tooltip

adamabeshouse:


src/pages/groupComparison/Survival.tsx
```diff
@@ -9,6 +9,7 @@ import {SURVIVAL_TOO_MANY_GROUPS_MSG} from "./GroupComparisonUtils";
 import ErrorMessage from "../../shared/components/ErrorMessage";
 import {blendColors} from "./OverlapUtils";
 import OverlapExclusionIndicator from "./OverlapExclusionIndicator";
+import {COLORS} from "../studyView/StudyViewUtils";
```
zhx828: Included but never used?

PR #2368, opened 2019-05-28
group comparison loading page - make study view closing behavior more intuitive

adamabeshouse:
image

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -555,7 +555,6 @@ export class StudyViewPageStore {
     }) {
         // open window before the first `await` call - this makes it a synchronous window.open,
         //  which doesnt trigger pop-up blockers. We'll send it to the correct url once we get the result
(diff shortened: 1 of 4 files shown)


alisman:
@adamabeshouse i think the message will only confuse.  "Sorry" suggests something went wrong.  We should just keep it very simple and show the loader, perhaps with a message saying "Preparing comparison session."  It really doesn't need to update.   If we're worried about people closing the original window, we can put an alert/cancel message in the unload event of that original window which tells them that we are in process of loading session and they should keep window open.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -555,21 +555,34 @@ export class StudyViewPageStore {
     }) {
         // open window before the first `await` call - this makes it a synchronous window.open,
         //  which doesnt trigger pop-up blockers. We'll send it to the correct url once we get the result
+        console.log(this.studyIds);
```
alisman: console.log

PR #2370, opened 2019-05-28
Upset changes

adamabeshouse:
image


adamabeshouse:
@jjgao now that we are doing partition I think it's confusing that this is different tooltip than the venn diagram. My opinion is we should change the one on the venn diagram as it's unnecessarily verbose. thoughts?

image

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -623,4 +623,28 @@ export function convertPatientsStudiesAttrToSamples(
             })
         )
     }));
+}
+
+export function partitionCases(
```
alisman: can we have a more descriptive name or a comment on this function that indicates the nature of the partition?

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -800,30 +800,41 @@ export default class GroupComparisonStore {
...
+            this.sampleSet,
+            this.samples
+        ],
+        invoke:()=>{
+            const sampleSet = this.sampleSet.result!;
+            return Promise.resolve(
+                partitionCases(
+                    this._activeGroupsNotOverlapRemoved.result!,
+                    (group)=>getSampleIdentifiers([group]),
+                    (id)=>sampleSet.get({ studyId: id.studyId, sampleId: id.sampleId })!.uniqueSampleKey,
```
alisman: strange that something called "id" is actually an object with other props.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -623,4 +623,28 @@ export function convertPatientsStudiesAttrToSamples(
...
+    const partitionMap = new ComplexKeyGroupsMap<string>();
+    const groupToCaseKeys = groupsNotOverlapRemoved.reduce((map, group)=>{
+        map[group.uid] = _.keyBy(getCaseIdentifiers(group).map(id=>{
+            return getUniqueCaseKey(id);
+        }));
+        return map;
+    }, {} as {[uid:string]:{[uniqueCaseKey:string]:any}});
+
+    for (const caseKey of caseKeys) {
+        const key:any = {};
```
alisman: expect key to be a string, no?

alisman: i guess a key can be an object nowadays?  is that fully supported?

src/pages/groupComparison/GroupComparisonStore.ts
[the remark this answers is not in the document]
adamabeshouse: this is actually important - for ComplexKey data structures you have to make a specific object because it uses the entire object as a key. typescript doesnt ensure that -it only ensures the object meets an interface

adamabeshouse: its an identifier, containing study and sample

jjgao:
> @jjgao now that we are doing partition I think it's confusing that this is different tooltip than the venn diagram since it now describes the same situation (intersection of included, minus non-included groups). My opinion is we should change the one on the venn diagram as it's unnecessarily verbose. thoughts?

Agreed. We should use the same text.

jjgao:
For the title, should we still call it "Samples overlap"? And "Overlap count" for y-axis?

jjgao:
Is it possible to add te group color (e.g. a square or dot) after the group name in this plot?

PR #2375, opened 2019-05-30
Optimize logging output in webpack build for netlify

alisman:
Netlify told us that verbose logging causes builds to be sluggish and timeout on their platform.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -78,6 +78,10 @@ loader:'sass-resources-loader',
 
 var config = {
 
+    stats: {
+        colors:true
+    },
+
     entry: [
         `babel-polyfill`,
         `${path.join(src, 'appBootstrapper.jsx')}`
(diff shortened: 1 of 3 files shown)


webpack.config.js
```diff
@@ -85,6 +85,10 @@ loader:'sass-resources-loader',
 
 var config = {
 
+    stats: {
```
alisman: this is turned off for netlify down below

webpack.config.js
```diff
@@ -512,6 +516,12 @@ if (isDev || isTest) {
 
 }
 
+// reduce logging to optize netlify build
+if (process.env.IS_NETLIFY === "true") {
+    stats: 'errors-only'
```
inodb: @alisman shouldn't this be `config['stats'] = 'errors-only';`

Also i tend to prefer env variables like `BUILD_REPORT_ERRORS_ONLY=true`, instead of referring to the CI platform. That way it's more clear what the variable does and we could use it on cirlceci or locally as well for instance. But not a big deal

alisman: ooh. yeah

webpack.config.js
```diff
@@ -512,6 +516,12 @@ if (isDev || isTest) {
 
 }
 
+// reduce logging to optize netlify build
+if (process.env.BUILD_REPORT_ERRORS_ONLY === "true") {
+    config.stats = 'errors-only'
```
alisman: yes, it can be just a string

PR #2376, opened 2019-05-30
GC: update legends based on active groups order

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6172

### DIFF
--- a/src/pages/groupComparison/OverlapVenn.tsx
+++ b/src/pages/groupComparison/OverlapVenn.tsx
@@ -85,17 +85,12 @@ export default class Venn extends React.Component<IVennProps, {}> {
     }
 
     @computed get legendData() {
-        const usedGroups = _.keyBy(this.props.sampleGroups.map(group=>group.uid));
-        const legendData:any[] = [];
-        _.forEach(this.props.uidToGroup, (group) => {
-            if (group.uid in usedGroups) {
-                legendData.push({
-                    name: truncateWithEllipsis(group.nameWithOrdinal, 100, "Arial", "13px"),
-                    symbol: { fill: group.color, strokeOpacity:0, type:"square", size: 6 }
-                });
+        return this.props.sampleGroups.map(sampleGroup=>{
+            return {
+                name: truncateWithEllipsis(this.props.uidToGroup[sampleGroup.uid].nameWithOrdinal, 100, "Arial", "13px"),
+                symbol: { fill: this.props.uidToGroup[sampleGroup.uid].color, strokeOpacity:0, type:"square", size: 6 }
             }
         });
-        return legendData;
     }
 
     @computed get vennPlotAreaWidth() {
--- a/src/pages/groupComparison/Survival.tsx
+++ b/src/pages/groupComparison/Survival.tsx
@@ -9,6 +9,7 @@ import {SURVIVAL_TOO_MANY_GROUPS_MSG} from "./GroupComparisonUtils";
 import ErrorMessage from "../../shared/components/ErrorMessage";
 import {blendColors} from "./OverlapUtils";
 import OverlapExclusionIndicator from "./OverlapExclusionIndicator";
+import _ from "lodash";
 
 export interface ISurvivalProps {
     store: GroupComparisonStore
@@ -26,12 +27,31 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
             this.props.store.uidToGroup
         ],
         invoke: () => {
+            const orderedActiveGroupUidSet = _.reduce(this.props.store._activeGroupsNotOverlapRemoved.result!, (acc, next, index) => {
+                acc[next.uid] = index;
+                return acc;
+            }, {} as { [id: string]: number });
             const partition = this.props.store.patientsVennPartition.result!;
+
+            // ascending sort partition bases on number of groups in each parition.
+            // if they are equal then sort based on the give order of groups
+            partition.sort((a, b) => {
+                const aUids = Object.keys(a.key).filter(uid=>a.key[uid]);
+                const bUids = Object.keys(b.key).filter(uid=>b.key[uid]);
+                if (aUids.length !== bUids.length) {
+                    return aUids.length - bUids.length;
+                }
+                const aCount = _.sumBy(aUids, uid=>orderedActiveGroupUidSet[uid])
+                const bCount = _.sumBy(bUids, uid=>orderedActiveGroupUidSet[uid])
+                return aCount - bCount;
+            });
             const uidToGroup = this.props.store.uidToGroup.result!;
             const analysisGroups = [];
             const patientToAnalysisGroups:{[patientKey:string]:string[]} = {};
             for (const entry of partition) {
                 const partitionGroupUids = Object.keys(entry.key).filter(uid=>entry.key[uid]);
+                // sort by give order of groups
+                partitionGroupUids.sort((a, b) => orderedActiveGroupUidSet[a] - orderedActiveGroupUidSet[b]);
                 if (this.props.store.overlapStrategy === OverlapStrategy.EXCLUDE && partitionGroupUids.length > 1) {
                     // dont show the overlap curves if we're excluding overlapping cases
                 } else {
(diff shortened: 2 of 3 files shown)


src/pages/groupComparison/OverlapVenn.tsx
```diff
@@ -17,12 +17,14 @@ export interface IVennProps {
...
         cases: string[];
+        color: string;
+        nameWithOrdinal: string;
     }[];
     patientGroups: {
         uid: string;
         cases: string[];
     }[];
-    uidToGroup: { [uid: string]: ComparisonGroup };
+    //uidToGroup: { [uid: string]: ComparisonGroup };
```
adamabeshouse: why?

src/pages/groupComparison/OverlapVenn.tsx
```diff
@@ -85,17 +87,23 @@ export default class Venn extends React.Component<IVennProps, {}> {
...
+        return this.props.sampleGroups.map(group=>{
+            return {
+                name: truncateWithEllipsis(group.nameWithOrdinal, 100, "Arial", "13px"),
+                symbol: { fill: group.color, strokeOpacity:0, type:"square", size: 6 }
             }
         });
-        return legendData;
+        // const usedGroups = _.keyBy(this.props.sampleGroups.map(group=>group.uid));
+        // const legendData:any[] = [];
+        // _.forEach(this.props.uidToGroup, (group) => {
```
adamabeshouse: delete comment

src/pages/groupComparison/Survival.tsx
```diff
@@ -26,12 +27,52 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+                    }
+                    return acc;
+                }, [] as string[]);
+
+                if (a1Keys.length !== a2Keys.length) {
+                    return a1Keys.length - a2Keys.length;
+                }
+
+                const x1Count = _.reduce(a1Keys, (acc, next) => {
+                    acc += orderedActiveGroupUids.indexOf(next);
```
adamabeshouse: we really should be avoiding using indexOf for lookups. It gives O(n^2) time complexity. a hashmap is the preferred choice. I have a helper method for this called stringListToIndexSet. Let me know if you need further explanation.

src/pages/groupComparison/OverlapVenn.tsx
```diff
@@ -17,12 +17,15 @@ export interface IVennProps {
...
+        color: string;
+        nameWithOrdinal: string;
     }[];
     patientGroups: {
         uid: string;
         cases: string[];
+        color: string;
+        nameWithOrdinal: string;
     }[];
-    uidToGroup: { [uid: string]: ComparisonGroup };
```
adamabeshouse: sorry I should have been clearer - I think it will be a lot more easily maintainable to just keep doing it the way we've been doing - the input is uid, and we just use uidToGroup to get any info that we need. is there a reason you want to change it this way?

src/pages/groupComparison/Survival.tsx
```diff
@@ -26,12 +27,51 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+            const orderedActiveGroupUidSet = _.reduce(this.props.store._activeGroupsNotOverlapRemoved.result!, (acc, next, index) => {
+                acc[next.uid] = index;
+                return acc;
+            }, {} as { [id: string]: number });
             const partition = this.props.store.patientsVennPartition.result!;
+
+            // ascending sort partition bases on number of groups in each parition.
+            // if they are equal then sort based on the give order of groups
+            partition.sort((a, b) => {
+                let a1Keys = _.reduce(a.key, (acc, next, key) => {
```
adamabeshouse: I think this is a little hard to read - Object.keys(a.key).filter(uid=>a.key[uid]) is one line and clearer

adamabeshouse: also `a1Keys` is not a good variable name - should be `aUids` or something

src/pages/groupComparison/Survival.tsx
```diff
@@ -26,12 +27,51 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+                    if (next === true) {
+                        acc.push(key);
+                    }
+                    return acc;
+                }, [] as string[]);
+                let a2Keys = _.reduce(b.key, (acc, next, key) => {
```
adamabeshouse: `bUIds`

src/pages/groupComparison/Survival.tsx
```diff
@@ -26,12 +27,51 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+                        acc.push(key);
+                    }
+                    return acc;
+                }, [] as string[]);
+
+                if (a1Keys.length !== a2Keys.length) {
+                    return a1Keys.length - a2Keys.length;
+                }
+
+                const a1Count = _.reduce(a1Keys, (acc, next) => {
```
adamabeshouse: _.sumBy(a1Keys, uid=>orderedActiveGroupUidSet[uid])

src/pages/groupComparison/UpSet.tsx
```diff
@@ -208,11 +199,15 @@ export default class UpSet extends React.Component<IUpSetrProps, {}> {
         return index * (this.barWidth() + this.barSeparation()); // half box + separation + half box
     }
 
+    @computed get uidToGroup() {
+        return _.keyBy(this.props.groups, group=>group.uid)
```
adamabeshouse: why do you want to remake uidToGroup here instead of reusing the one thats passed in?

src/pages/groupComparison/VennSimple.tsx
```diff
@@ -22,12 +22,13 @@ export interface IVennSimpleProps {
...
         uid: string;
         cases: string[];
+        color: string;
+        nameWithOrdinal: string;
     }[]; // either 2 or 3
     x:number;
     y:number;
     width:number;
     height:number;
-    uidToGroup: { [uid: string]: ComparisonGroup };
```
adamabeshouse: same comment as before. I don't get why this is preferred - this way we have to keep changing the prop whenever we add something to the group. I think it's best to just leave it as uidToGroup and keep the prop simple - pass in uid and cases.

src/pages/groupComparison/OverlapVenn.tsx
kalletlak: had to do it this way to maintain group order. and looks like only these two parameters are used from `ComparisonGroup`. do you have any other idea?

adamabeshouse: I dont understand what you can do now re: group order that you couldnt do before?

adamabeshouse: in legendData just loop through sampleGroups as you're doing. that gives the order. then to get color and nameWithOrdinal go through uidToGroup

kalletlak: hmm ok

adamabeshouse:
@kalletlak is it okay if we wait to merge this for https://github.com/cBioPortal/cbioportal-frontend/pull/2370/files because I think it will be easier to rebase your changes onto mine than vice versa since this is only 25 lines

kalletlak:
sure!

src/pages/groupComparison/Survival.tsx
```diff
@@ -26,12 +27,31 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+            const orderedActiveGroupUidSet = _.reduce(this.props.store._activeGroupsNotOverlapRemoved.result!, (acc, next, index) => {
+                acc[next.uid] = index;
+                return acc;
+            }, {} as { [id: string]: number });
             const partition = this.props.store.patientsVennPartition.result!;
+
+            // ascending sort partition bases on number of groups in each parition.
+            // if they are equal then sort based on the give order of groups
+            partition.sort((a, b) => {
+                const aUids = Object.keys(a.key).filter(uid=>a.key[uid]);
```
alisman: i don't quite understand this.  you're taking keys of aj object (a.key) then filtering out those where the corresponding values is false?    seems like it would be a lot clearer with reduce

src/pages/groupComparison/Survival.tsx
alisman: ha.  i wrote the exact opposite.  ok.  not important enough to fight about

src/pages/groupComparison/Survival.tsx
adamabeshouse: @alisman I asked Karthik to change it from reduce. I don't see what is unclear about this, it's a one liner, it does what it says. why would reduce make it more clear?

PR #2378, opened 2019-05-30
udpate API clients

alisman:


inodb:
ordering of API endpoints is intended change: https://github.com/cBioPortal/cbioportal/pull/5973

PR #2382, opened 2019-05-30
More upset changes - add colored squares to axis, and change title/axis label

adamabeshouse:
image

### DIFF
--- a/src/pages/groupComparison/GroupComparisonStore.ts
+++ b/src/pages/groupComparison/GroupComparisonStore.ts
@@ -812,7 +812,7 @@ export default class GroupComparisonStore {
         await:()=>[
             this._activeGroupsNotOverlapRemoved,
             this.sampleSet,
-            this.samples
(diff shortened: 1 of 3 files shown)


adamabeshouse:
looks like there is some bug here (note to self: https://beta.cbioportal.org/comparison?sessionId=5ce6fc2ce4b0ab41378740bb)

please ignore the bug and look at teh design

alisman:
@adamabeshouse @jjgao @schultzn.  i'm not sure this helps.  the name of the group is right there.  the little blue squares makes me think they are a part of the data visualization that I need to figure out how to interpret.   i think color is only important when user needs to so associate data points with a legend.

jjgao:
@alisman I think the colors can be helpful to set up a mental map for the user across tabs. Let's try to make it less like data, e.g. put it in front of the group name?

schultzn:
I agree, if we use them consistently, they could be very useful. Maybe explore that for v2?

PR #2388, opened 2019-05-31
remove survival charts from study view

dippindots:
Remove survival charts from study view
Fix # https://github.com/cBioPortal/cbioportal/issues/6128.

- a removed the survival charts

previously:
image

now:
image


alisman:
@zhx828 @jjgao does it make sense to completely remove all this code or just to configure the study view not to show survival plot?  just want to make sure

zhx828:
@alisman if the survival is not desired due to the group comparison I think removing all related code makes sense. The survival plot is really designed for particular data type. I would not expect to reuse the plot for other data at the moment.

jjgao:
@dippindots I think we should complete remove the code at certain point. I am just not sure if we have missed anything and if people will ask us to bring it back. For this PR, please make sure they won't show up and also no data for the survival charts will be requested. Please create another issue to do clearn up of the code.

dippindots:
Thanks @zhx828 @jjgao. I will just hide the survival plots and remove the data fetching for this pr, and create another pr to clean up the code. Do you think that's ok? @alisman

PR #2389, opened 2019-05-31
Bar chart title alignment

dippindots:
Bar chart title left alignment
Fix # https://github.com/cBioPortal/cbioportal/issues/6197.

image

### DIFF
--- a/src/pages/resultsView/enrichments/GeneBarPlot.tsx
+++ b/src/pages/resultsView/enrichments/GeneBarPlot.tsx
@@ -154,49 +154,49 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}> 
 
     @computed get toolbar() {
         return (
-            <div>
-                <div style={{ zIndex: 10, position: "absolute", top: "10px", left: "10px" }}>
-                    <div className={styles.ChartControls}>
+            <React.Fragment>
+                <div style={{ zIndex: 10, position: "absolute", top: "10px", left: "90px" }}>
                         <strong>{ this._label || this.defaultOption.label}</strong>
-                        <DefaultTooltip
-                            trigger={['click']}
-                            destroyTooltipOnHide={false}
-                            visible={this.isGeneSelectionPopupVisible}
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/enrichments/GeneBarPlot.tsx
```diff
@@ -154,37 +154,41 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}>
...
+                        >
+                            <div>
+                                <button className="btn btn-default btn-xs">
+                                    Select genes
+                                </button>
+                            </div>
+                        </DefaultTooltip>
+                    </div>
+                </div>
+                <div style={{ zIndex: 10, position: "absolute", top: "10px", right: "10px" }}>
```
dippindots: @alisman Only changed L158 and L191 to align them to left and right respectively.

src/pages/resultsView/enrichments/GeneBarPlot.tsx
```diff
@@ -154,37 +154,41 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}>
...
-                        }
-                        placement="bottomLeft"
-                    >
-                        <div>
-                            <button className="btn btn-default btn-xs">
-                                Select genes
-                            </button>
-                        </div>
-                    </DefaultTooltip>
+            <div>
```
alisman: @adamabeshouse @dippindots this might be a good example where we could use react fragment!  <> instead of <div> since div is superfluous.  @dippindots try it out!

jjgao:
@dippindots I am wondering it's possible to assign it to the start of y-axis so that it's more like a plot title?

dippindots:
@jjgao like this?
image

Or you think this way is better?
image

alisman:
@dippindots need to update screenshots

https://38500-66571349-gh.circle-artifacts.com/0/imageCompare.html

dippindots:
@alisman Yes, will do when we finalized the design!

PR #2390, opened 2019-05-31
Add most enriched group for 3 or more groups enrichment analysis

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6203

Screen Shot 2019-05-31 at 4 23 23 PM


src/pages/resultsView/enrichments/EnrichmentsUtil.spec.tsx
```diff
@@ -457,13 +447,19 @@ describe("EnrichmentsUtil", () => {
         });
     });
 
+    /* 
+    export function getAlterationRowData(
```
adamabeshouse: delete comment

PR #2392, opened 2019-05-31
Allow maximum 5 groups in a session workspace

adamabeshouse:
Implements UI to select which ones

image

clicking on "354 more groups..." brings up a modal

image

image

image

image

image

### DIFF
--- a/src/pages/groupComparison/GroupComparisonStore.ts
+++ b/src/pages/groupComparison/GroupComparisonStore.ts
@@ -8,7 +8,8 @@ import {
     isGroupEmpty,
     ClinicalDataEnrichmentWithQ,
     OverlapFilteredComparisonGroup, getSampleIdentifiers,
-    GroupComparisonTab, getOrdinals, partitionCasesByGroupMembership, defaultGroupOrder
+    GroupComparisonTab, getOrdinals, partitionCasesByGroupMembership,
+    defaultGroupOrder, MAX_GROUPS_IN_WORKSPACE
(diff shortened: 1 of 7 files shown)


schultzn:
Wow. This looks really nice. 

I don’t like the term “in workspace” - can you maybe replace that with something less technical, such as “in group comparisons”?

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -88,6 +90,12 @@ export default class GroupComparisonStore {
         return this.appStore.isLoggedIn;
     }
 
+    @autobind
+    @action
+    public setWorkspaceGroups(map:ObservableMap<boolean>) {
+        this.isGroupInWorkspace = map.toJS();
```
alisman: this sounds like a boolean but is actually a map.  why not a method that takes group as param and looks in map? also, why toJS?

adamabeshouse: sure, I can split it into a map and a clearer getter method

toJS just makes a copy of it, I think it's a little cleaner that way than using the map thats passed in which the caller still has access to. and it's not necessary to be an observable map on this end, so toJS is good

adamabeshouse:
image

adamabeshouse:
lets force open the modal if there are more than 5 groups in the workspace

adamabeshouse:
image

image

adamabeshouse:
image

PR #2394, opened 2019-06-03
Update screenshots

inodb:
- This is fixed after re-import: https://github.com/cBioPortal/datahub/issues/742
- gene symbol change GET1 (alias) -> WRB
- text alignment in plot change


inodb:
merging cause it fixed all end_to_end_tests

PR #2395, opened 2019-06-03
Fix e2e2 - all samples text in query selector

inodb:
This fixed the timeout issues in the e2e tests after changing to the new aws db


inodb:
This fixed the timeout issues in the e2e tests

PR #2399, opened 2019-06-03
Max 20 groups in a comparison session

adamabeshouse:
(1) Prevent more than 20 groups creating a session from the groups dropdown
(2) When creating a session from a large chart, make session only out of the top 20 biggest groups, and confirm with user first

image

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -261,7 +261,6 @@ export class StudyViewPageStore {
     @action
     public setComparisonConfirmationModal(getModal:((hideModal:()=>void)=>JSX.Element)) {
         this._comparisonConfirmationModal = getModal(()=>{this._comparisonConfirmationModal = null; });
-        console.log(this.comparisonConfirmationModal);
     }
 
     // <comparison groups code>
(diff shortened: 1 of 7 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -524,18 +524,29 @@ export class StudyViewPageStore {
...
-                            origin: this.studyIds,
-                            color: lcValueToColor[lcValue].color,
-                        };
-                    });
+
+                    // create groups using data - up to MAX_GROUPS_IN_SESSION - ordered by size
+                    const groups:SessionGroupData[] = [];
+                    const sortedAttrVals = _.sortBy(clinicalAttributeValues, attrVal=>-attrVal.count);
+                    for (const attrVal of sortedAttrVals) {
+                        if (groups.length >= MAX_GROUPS_IN_SESSION) {
```
kalletlak: may be move this break statement to end of loop

adamabeshouse: imo it's good here. in principle moving it to the end could lead to overshooting it, if anything about the previous code changes

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -253,6 +253,16 @@ export class StudyViewPageStore {
         }
     }
 
+    @observable.ref private _comparisonConfirmationModal:JSX.Element|null = null;
```
alisman: is there a reason to bother with get/set when we can't use set (i guess because of the decorator problem)?

alisman: hm. this is weird.  not sure what this does

adamabeshouse: I can see why it might be weird. i just thought it was syntactically sensible  since the getter doesn't do any computation. the actual reason we cant use a setter is because the setters argument structure is very constrained

PR #2405, opened 2019-06-04
MiniOncoprint: color group bar (top) according to group colors

alisman:
image

### DIFF
--- a/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
+++ b/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
@@ -27,7 +27,7 @@ import './styles.scss';
 
 export interface IAlterationEnrichmentContainerProps {
     data: AlterationEnrichmentWithQ[];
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -125,8 +125,9 @@ export default class AlterationEnrichmentContainer extends React.Component<IAlte
 
                               const groups = _.map(data.groupsSet);
 
-                              const group1 = groups[0];
-                              const group2 = groups[1];
+                              // we want to order groups according to order in prop.groups
+                              const group1 = groups.find((group)=>group.name===this.props.groups[0].name);
```
alisman: this should probably be improved.  groupSet itself (data passed to column renderer) should have ordering and color info, but it doesn't and need to get this done

adamabeshouse: it's NBD since you're doing it a small constant number of times (2)

PR #2420, opened 2019-06-05
use normal quartiles in comparisons from bar charts

adamabeshouse:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -353,97 +353,57 @@ export class StudyViewPageStore {
...
-                                    origin: this.studyIds,
-                                };
-                            });
-                            break;
-                    }
+                    const groups = quartiles.map(quartile=>{
+                        let studies;
+                        if (clinicalAttribute.patientAttribute) {
+                            studies = getStudiesAttr(
+                                _.flattenDeep<any>(
```
kalletlak: why not flatMap?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -353,97 +353,57 @@ export class StudyViewPageStore {
...
                     });
 
-                    // sort values
-                    // group into halves, thirds, or fourths, depending on how many distinct values there are
-                    const numericalData = data.filter(d=>!isNaN(d.value as any));
-                    const groupedByValue = _.groupBy(numericalData, d=>parseFloat(d.value));
-                    const distinctValues = _.sortBy(Object.keys(groupedByValue).map(v=>parseFloat(v)));
-
-                    statusCallback(LoadingPhase.CREATING_SESSION);
+                    // get quartiles of data
```
alisman: we should really have a test around this stuff, no?   we could follow up with that

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -353,97 +353,57 @@ export class StudyViewPageStore {
...
+                    data = _.chain(data)
+                        .filter(d=>!isNaN(d.value as any))
+                        .sortBy(d=>parseFloat(d.value))
+                        .value();
+
+                    const quarterLength = data.length / 4;
```
alisman: i think we have a library that generates quartiles

PR #2425, opened 2019-06-06
GC: fix box plot category order in clinical tab

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6240


src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -245,7 +245,8 @@ export default class ClinicalData extends React.Component<IClinicalDataProps, {}
     private readonly groupSampleDataPromise = remoteData({
         await: () => [this.props.store.sampleKeyToActiveGroups, this.props.store.uidToGroup],
         invoke: async () => {
-            const axisData: IAxisData = { data: [], datatype: "string" };
+            const categoryOrder = _.map(this.props.store.uidToGroup.result!,group=>group.nameWithOrdinal).sort()
```
adamabeshouse: This is hacky to me - ordinals being alphabetical order isn't a guarantee of the ordinals even though it sort of makes sense. It's just logical wire-crossing IMO. Better to do it official way which is based on the order in _originalGroups. `stringListToIndexSet` is useful for this

PR #2429, opened 2019-06-06
Unit test overlap computations

adamabeshouse:


alisman:
@adamabeshouse when you pulled functions out to test then, did you do any re-writing of the code?

adamabeshouse:
@alisman no. only a bit of typing work. but the logic, no.

PR #2436, opened 2019-06-07
Delete group functionality in group comparison page

adamabeshouse:


adamabeshouse:
broken tests are unrelated

PR #2458, opened 2019-06-14
Support multiple social_auth services

kalletlak:
# What? Why?
Front-end changes for https://github.com/cBioPortal/cbioportal/issues/6217


alisman:
We need to think about how this will affect existing instances/configurations.  Is that property new? What does it default to?

PR #2462, opened 2019-06-17
Add tests to gene bar plot

kalletlak:
# What? Why?
Add test to gene bar plot in enrichment analysis https://github.com/cBioPortal/cbioportal/issues/6025


kalletlak:
@dippindots should be fixed now

alisman:
There is one relevant screenshot that's failing.  I don't see the difference.  This test is not part of the PR, right?  

https://40566-66571349-gh.circle-artifacts.com/0/imageCompare.html

kalletlak:
looks like there is a difference, but that's not related to this pr

PR #2466, opened 2019-06-18
Custom group tracks in oncoprint

adamabeshouse:


jjgao:
Thanks, @adamabeshouse!

There is issue: if I share a session including "in group" tracks with another person, that person will be able to render the "in group" tracks.

https://www.cbioportal.org/results/oncoprint?clinicallist=NUM_SAMPLES_PER_PATIENT%2CIN_COMPARISON_GROUP_Non-Small%20Cell%20Lung%20Cancer&session_id=5d0a6174e4b0ab4137876774&show_samples=false

image

I think it was due to authentication, ie. only the user who created the groups is allowed to access it. I am wondering how we can get around that? There are potentially two options:

1. allow read access to any group for anyone
2. duplicate the group data in query session

@kalletlak @adamabeshouse what do you think?

adamabeshouse:
@jjgao I'm not sure I understand what (2) means. But we could accomplish this by allowing read access to the group for anyone.

adamabeshouse:
@jjgao we can merge it now and loop back on that. I'll change the code to make the clinical attribute id in the URL use the group uid instead of the name, so that we can come back and do that if we want

PR #2469, opened 2019-06-18
Use https for cancerhotspots links

inodb:
Some of these links don't seem to work when using `http`. At least in my browser


inodb:
e2e failures related to genome nexus and group comparison, not the link update

PR #2470, opened 2019-06-18
Fix mutationTable tests

alisman:


end-to-end-tests/specs/core/mutationTable.spec.js
```diff
@@ -51,11 +51,12 @@ describe('Mutation Table', function() {
         it('should show more exon number after clicking "Show more"', ()=>{
             // click "show more" to add more data
             browser.click("button*=Show more");
+
+            //browser.debug();
```
adamabeshouse: comment

end-to-end-tests/specs/core/mutationTable.spec.js
```diff
@@ -51,11 +51,12 @@ describe('Mutation Table', function() {
...
             // check if 6 exact matches for 6 appear
+
             browser.waitUntil(() => {
-                textValuesWith6 = browser.getText('//*[text()="6"]');
-                // 4 values from Exon column + 1 value from #Mut in Sample column + 1 value from COSMIC column
-                return textValuesWith6.length === 6;
+                return testJqueryExpression( ()=>$("[class*=exon-module__exon-table]").length) == 38;
```
alisman: @adamabeshouse this is helper which i think you would like.   allows you to use full power of jquery selectors instead of crappy webdriverio selectors

end-to-end-tests/specs/specUtils.js
```diff
@@ -148,6 +148,10 @@ function checkOncoprintElement(selector) {
     return browser.checkElement(selector || "#oncoprintDiv", { hide:[".qtip", '.dropdown-menu', ".oncoprintjs__track_options__dropdown", ".oncoprintjs__cell_overlay_div"] });
 }
 
+function testJqueryExpression(callback){
```
adamabeshouse: badly named function - it does something much more general

end-to-end-tests/wdio/webdriver-manager.conf.js
```diff
@@ -324,9 +324,9 @@ if (doBrowserstack) {
     config.key = process.env.BROWSERSTACK_KEY;
 }
 
-// config.specs = [
-//     './specs/**/oncoprint.spec.js'
-// ];
+config.specs = [
```
adamabeshouse: this should be commented

end-to-end-tests/specs/specUtils.js
alisman: hurt feelings.  but you're right.  in fact, this doesn't really add much of anything to browser.execute except call value (which is not even documented!!!!).   maybe it's worth it though just for value

adamabeshouse: sorry!! I have a very terse github style.. I should be more sensitive.

end-to-end-tests/specs/core/mutationTable.spec.js
adamabeshouse: oooh cool. how about

```
 testJqueryExpression(expression:string) {
           return browser.execute(()=>$("[class*=exon-module__exon-table]").length).value;
}
```

end-to-end-tests/specs/core/mutationTable.spec.js
```diff
@@ -51,11 +51,9 @@ describe('Mutation Table', function() {
...
             // click "show more" to add more data
             browser.click("button*=Show more");
-            // check if 6 exact matches for 6 appear
+
             browser.waitUntil(() => {
-                textValuesWith6 = browser.getText('//*[text()="6"]');
-                // 4 values from Exon column + 1 value from #Mut in Sample column + 1 value from COSMIC column
-                return textValuesWith6.length === 6;
+                return testJqueryExpression( ()=>$("[class*=exon-module__exon-table]").length) == 38;
             }, 60000, "Exon data not in Exon column after clikcing show more button");
```
leexgh: `clikcing` should be `clicking`, I wrote this...

end-to-end-tests/specs/specUtils.js
adamabeshouse: how about 
```
testJqueryExpression(expression:string) {
    return browser.execute(()=>$(expression).length).value;
}
```

PR #2473, opened 2019-06-19
updates to group comparison saving

adamabeshouse:
(1) Immediately save new session when adding/deleting a group
(2) Save selected groups and drag order only to URL


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -70,56 +70,100 @@ export default class GroupComparisonStore {
...
+    @action public updateGroupOrder(oldIndex:number, newIndex:number) {
+        let groupOrder = this.groupOrder;
+        if (!groupOrder) {
+            groupOrder = this._originalGroups.result!.map(g=>g.name);
+        }
+        groupOrder = groupOrder.slice();
+        const poppedUid = groupOrder.splice(oldIndex, 1)[0];
+        groupOrder.splice(newIndex, 0, poppedUid);
+
+        (window as any).routingStore.updateRoute({ groupOrder: JSON.stringify(groupOrder) } as Partial<GroupComparisonURLQuery>);
```
kalletlak: may be join by `,`

adamabeshouse: a group name could have a comma in it

alisman: lets pass in routingStore from view instead of using window global.

adamabeshouse: done

PR #2474, opened 2019-06-19
Add gene panel warning on cancer types, survival, enrichments tabs

adamabeshouse:


tmazor:
Looks good! But now that I see the warning text in context, I'd like to change it slightly. Can we do two different messages, depending on how many genes are queried?

One gene: The results below do not reflect gene panel data. Samples which were not profiled for the queried gene are included in the analysis below.

Multiple genes: The results below do not reflect gene panel data. Samples which were not profiled for some or all of the queried genes are included in the analysis below.

adamabeshouse:
@tmazor updated

PR #2475, opened 2019-06-19
Coloring clinical value alive and not progressed; Support clinical value with multiple words

zhx828:
Also include the combination that all words are camel cased

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6292#event-2425315229

### DIFF
--- a/src/shared/lib/Colors.ts
+++ b/src/shared/lib/Colors.ts
@@ -52,9 +52,11 @@ export let RESERVED_CLINICAL_VALUE_COLORS: { [value: string]: string } = {
     true: CLI_YES_COLOR,
     yes: CLI_YES_COLOR,
     positive: CLI_YES_COLOR,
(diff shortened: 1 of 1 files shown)


zhx828:
@adamabeshouse I'm not so sure the proper way to deal with the clinical value format. Ideally, we should remove all spaces and lowercase the value, don't bother doing camel case/upper case etc. But seems oncoprint at the moment only accepts the exact match. Any thoughts?

adamabeshouse:
@zhx828 yeah. we could make an update to the oncoprint library to take a function for `category_to_color` so we don't have to do this. would that be better or is this okay for now?

zhx828:
@adamabeshouse it's ok for now, just when we see more extreme cases, then we should update oncoprint.

zhx828:
The failing test is not related to this PR.

PR #2477, opened 2019-06-20
update cbioportal api docs

inodb:
Add new treatment/response related endpoints


pvannierop:
All good!

inodb:
2 errors unrelated to this PR:

1. order of mutations on patient view is different b/c it's a different db
2. some issue with gnomad column test

PR #2478, opened 2019-06-20
GC -E2e and screenshot tests for upset create group

adamabeshouse:


end-to-end-tests/specs/core/groupComparison.spec.js
```diff
@@ -51,22 +48,22 @@ describe("group comparison venn diagram tests", function() {
     
         it("select from patient venn diagram", function () {
             // unselect sample venn diagram first
-            browser.leftClick('rect[data-test="#d2644esampleComparisonGroupRegionRect"]');
-            browser.leftClick('rect[data-test="#d2644epatientComparisonGroupRegionRect"]');
```
adamabeshouse: @dippindots I changed this because we don't want the test to be tied to the particular colors we use for rendering. It's good to minimize unnecessary dependencies across the codebase - when linking two things together, try to think about whether one of them could easily change without the other changing.

dippindots: @adamabeshouse Got it! Thanks!

PR #2480, opened 2019-06-20
Use similar component for oncoprint clinical attribute selection as study view

adamabeshouse:


jjgao:
Thanks, @adamabeshouse. It looks great!

- Should we move the whole menu to left now?
- When I click on Groups tab, the dialog jumped to top
- For groups, would it be better to display the % of samples intead of always 100%?
- If I share an oncoprint including a group track with another person (e.g. open w/o login), would they see a group tab with that track?

adamabeshouse:
@jjgao 
- Do you mean the whole toolbar?
- Can you give a link where this happened?
- Yes that's a good idea.
- Yes, the group referenced in the URL is loaded into the comparison groups available for the query, which includes showing it in the groups tab.

alisman:
If I make my window just a little less than 800px high, the whole UI component jumps to the top of the page, presumably because there's not enough room for it.   See screenshot.  I'm sure this is some feature of the library we're using. 

image

alisman:
@adamabeshouse maybe a simple solution would be to hide the tooltip on browser resize or body scroll.

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
```diff
@@ -0,0 +1,228 @@
...
+                    count=>100*count/numSamples
+                )
+            );
+        }
+    });
+
+    readonly addClinicalTracksMenu = MakeMobxView({
+        await:()=>[this.options],
+        render:()=>(
+            <AddChartByType
```
alisman: should we generalize the name of this thing?

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
```diff
@@ -0,0 +1,228 @@
...
+                options={this.options.result!.groups}
+                freqPromise={this.clinicalAttributeIdToAvailableFrequency}
+                onAddAll={this.addAll}
+                onClearAll={this.clear}
+                onToggleOption={this.toggleClinicalTrack}
+                optionsGivenInSortedOrder={true}
+            />
+        ),
+        renderPending:()=><LoadingIndicator isLoading={true} small={true}/>,
```
alisman: i think most everywhere else we're using size attribute.  We should probably get rid of small attribute.  There's only <5 mentions

alisman: could be another PR

jjgao:
@adamabeshouse for the 1st point, yes the whole toolbar. I think we need a better position for it. Right now, it floats in the middle.

adamabeshouse:
@alisman doesnt seem like that will completely cover it because just changing between clinical and group menus sometimes does it. I will try to dynamically set the tooltip height

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
adamabeshouse: yes probably though I don't know what it should be. maybe we should just leave it until we find a use that is different enough that it needs more generalizing

adamabeshouse:
@alisman @jjgao I fixed the issues, please test again when you can

alisman:
@adamabeshouse looks good except for one thing: the arrow.  When you reduce browser height to exactly the height at which the dropdown fits entirely in portal (or smaller), the arrow shifts to the right and points at another button in toolbar.  I noticed that none of other dropdown in Oncoprint have arrow thingies so maybe we just lose it?  

image

adamabeshouse:
good idea @alisman

alisman:
of course, we should be trying to normalize these dropdowns across the site, but i think we can follow up later with that.

PR #2486, opened 2019-06-21
Germline and VUS filtering warning, toggleable in mutations tab

adamabeshouse:


alisman:
@adamabeshouse, do we show the message on a query where there are no germline mutations?  it would see we do not

adamabeshouse:
@alisman that's correct

jjgao:
- "48 germline mutations are not shown." is not accurate in Enrichments and Survival tab. Maybe we can something similar to "The results below reflect the OQL specification from your query."
- when showing germline mutations, I don't think we need to show a message in the Mutations tab.
- this should apply to Mutex tab too.

adamabeshouse:
@jjgao 

(1) how about 48 germline mutations are excluded - is that general enough to capture that theyre not considered for the altered group?

(2) so that means that it wont be toggleable the way oql is, and they'll have to toggle the global setting to get that option back. is that okay?

(3) which would apply to mutex tab?

tmazor:
@adamabeshouse the VUS count doesn't seem to be working. I toggled on "hide VUS" for this query, and it says 632 VUS are not shown, but it should be something like 5.

https://cbioportal-frontend-pr-2486.herokuapp.com/results/oncoprint?session_id=5c339a01e4b05228701fa8ae

tmazor:
@adamabeshouse for that same query, it also seems that show germline is trumping hide VUS. When I hide VUS, the VUS that are also germline are still visible in OncoPrint. If I hide VUS, I think that should hide all VUS, regardless of somatic vs germline.

tmazor:
@adamabeshouse MutEx seems to be the only tab without warnings when VUS/germline are hidden. Is there a reason for that? The numbers do appear to reflect the filtering.

tmazor:
I'm finding the Mutations tab somewhat confusing for two reasons.
1) It has a control to filter for germline mutations (in the absence of any filter applied in OncoPrint) but not for VUS, which would be, I think, a more common usecase given that germline data is relatively rare.
2) The control over filtering germline mutations appears to be independent of the OncoPrint setting

Are there reasons I'm not appreciating for why these were implemented this way?

adamabeshouse:
@tmazor looping in @jjgao to answer your last comments questions

tmazor:
Looks good @adamabeshouse !

I have one more question about the VUS filter:

Hide VUS hides copy number alterations as well as mutations. However, I generally think of VUS as a term that applies to mutations. And, the "hide" option is in a menu called "Mutations". And, the copy number driver annotation is invisible to the user on OncoPrint.

For all those reasons, I am finding it confusing when "hide VUS" also hides copy number changes.

I'm not necessarily against this behavior overall, I just want to raise the issue that this might be confusing for users.

@jjgao @schultzn do you also find this confusing or is it just me?

schultzn:
Good point @tmazor - maybe we can simply change the description of the action from "Hide VUS (variants of unknown significance)" to "Hide variants and copy-number changes of unknown significance". It's not perfect, and technically copy-number changes should not be included in a menu called "Mutations", but maybe it is ok...

adamabeshouse:
@jjgao we need your responses please

jjgao:
@adamabeshouse could you summarize the discussion into a to-do list and anything you need clarification?

adamabeshouse:
@jjgao 
------
> "48 germline mutations are not shown." is not accurate in Enrichments and Survival tab. Maybe we can something similar to "The results below reflect the OQL specification from your query."

how about "48 germline mutations are excluded from analysis"?

> when showing germline mutations, I don't think we need to show a message in the Mutations tab.

so that means that it wont be toggleable the way oql is, and they'll have to toggle the global setting to get that option back. is that okay?

> this should apply to Mutex tab too.

sorry, which?

-------
@tmazor raised a question about how we are handling the mutations tab:

"I'm finding the Mutations tab somewhat confusing for two reasons.

    It has a control to filter for germline mutations (in the absence of any filter applied in OncoPrint) but not for VUS, which would be, I think, a more common usecase given that germline data is relatively rare.
    The control over filtering germline mutations appears to be independent of the OncoPrint setting

Are there reasons I'm not appreciating for why these were implemented this way?"
---------
finally, @tmazor raised a question about whether the language VUS is confusing when it also applies to CNA.

jjgao:
Thanks, @adamabeshouse. Here is a to-do/discuss list.

1. Change the warning message to "48 germline mutations are excluded from analysis"
2. I think for mutations tab, when we are not filtering out germline mutations, we should not show a message. (In OQL or VUS, we only show a message when there is filtering) This will address @tmazor's point too.
3. Change "Hide VUS (variants of unknown significance)" to "Hide mutations and copy number alterations of unknown significance"
4. All filtering should be applied to MutEx tab (ie, when hiding germline mutaitons, a sample with only germline mutations will not be considered in the altered group)
5. All filtering should be applied to CancerTypeSummary tab (please coordinate with @kalletlak #2501)
6. Any other tabs?

adamabeshouse:
@jjgao @tmazor just updated with changes we discussed.

Please let me know if anythings missing/should be changed!

jjgao:
@adamabeshouse it looks great!

tmazor:
@adamabeshouse this looks great! I found just 2 things:

1) The co-expression tab has a warning banner about filtering but as best I can tell, the plot isn't colored according to the filters. 

2) If I click "include" on the Mutations tab, that doesn't remove the filter for any other tab, correct? It just changes what is shown on the Mutations tab? If that's the case, I think we should change the warning text on the Mutations tab to make that more clear.

I propose: "... are excluded from analysis." --> "... are hidden below."
The button can say "Show" rather than "Include".
After clicking the button, the text can say "... are shown below."

I think that will make the functionality more clear.

jjgao:
@tmazor good points!

For co-expression, I think we should just say that the filter was not applied.

For mutations, maybe we can also change the button to "Show" and "Hide"?

jjgao:
Found another small issue, when I only select mutations for analysis (http://bit.ly/2xKWgd8), the message is still referring to both mutations and copy number.

jjgao:
For Plots tab, I think the filtering is only applied to mutations but not CNAs, for example, the amplification in [this query](is a vus).

image

I think we should not apply these filters in Plots tab. We can apply the filter for coloring the samples, but how are we going to deal with the data on x or y-axis? e.g if we excluded AMP, how should we place these samples?

adamabeshouse:
Thanks @tmazor ! Just pushed requested updates.

tmazor:
@adamabeshouse The mutations tab only has a germline filter warning - the VUS one seems to have disappeared?

And I second @jjgao 's point about changing the filter warning text to be specific to mutations or copy number if only one of those data types is queried.

In terms of the plots tab, I do think it's worth figuring out how to apply the filters here. To me, one of the best use cases for this filter is in the plots tab -- if I put mutated vs wildtype on one axis, it would be awesome to have that reflect the VUS filter. I'm ok with making that a separate task though, once we put some more thought into how to do that properly, and for now removing the banner and having the colors ignore those filters.

adamabeshouse:
@tmazor in the mutations tab, it depends on which gene you're viewing. if there are no VUS for that gene, you wont see the alert. for example with ovarian tcga nature 2011 and brca1 brca2, only brca2 has vus

adamabeshouse:
@jjgao @tmazor sorry, confused about what to do for plots tab? and for coexpression, is it enough to hide the banner? if not, what should the text say?

jjgao:
@adamabeshouse for both Plot and Co-expression, is it possible not to apply these two filters? We should add warnings that the filters are not applied.

adamabeshouse:
@jjgao so then for plots tab we are also not showing driver annotation?

adamabeshouse:
updated @jjgao @tmazor

tmazor:
Awesome @adamabeshouse !

PR #2490, opened 2019-06-24
Show query interface and error message when results page url contains invalid studyId

alisman:
image


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -413,31 +415,39 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+        if (this.resultsViewPageStore.invalidStudyIds.result.length > 0) {
+            return (<ErrorScreen
+                title={'Invalid Query'}
+                body={<p>{`You query contains the following invalid study IDs: ${this.resultsViewPageStore.invalidStudyIds.result.join(", ")}`}</p>}
+            />);
+        } else {
+            return (<div>
+                {
+                    // if query invalid(we only check gene length for now), return error page
+                    (this.resultsViewPageStore.isQueryInvalid) && (
```
alisman: probably we should combine the handling of "invalid" queries here with the above

adamabeshouse: should this be an else if branch?

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -413,31 +414,39 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+                    )
+                }
+                {
+                    (this.resultsViewPageStore.studies.isComplete) && (
+                        <Helmet>
+                            <title>{buildResultsViewPageTitle(this.resultsViewPageStore.hugoGeneSymbols, this.resultsViewPageStore.studies.result)}</title>
+                        </Helmet>
+                    )
+                }
+                {(this.resultsViewPageStore.studies.isComplete) && (
```
adamabeshouse: merge these two in a fragment, since they have the same condition?

PR #2491, opened 2019-06-24
Do not show expression tab when there's no expression data to show

alisman:
Hide expression tab when queried studies do not have an expression data


alisman:
@adamabeshouse gonna write an e2e test for this

PR #2493, opened 2019-06-24
New SVG component for truncated text with tooltip, and application to survival chart legend

adamabeshouse:
https://www.cbioportal.org/comparison/survival?sessionId=5cf13d70e4b0ab4137874293
image

downloaded SVG:

image


src/shared/components/TruncatedTextSVG.tsx
```diff
@@ -0,0 +1,118 @@
...
+            <>
+                <text
+                    {...rest}
+                    onMouseOver={this.onMouseOver}
+                    onMouseOut={this.onMouseOut}
+                    onMouseMove={this.onMouseMove}
+                >
+                    {this.textReport.text}
+                </text>
+                {this.textReport.truncated && this.tooltipOpen && (ReactDOM as any).createPortal(
```
alisman: we can use react-portal library to avoid imperative call to creatPortal.  e.g. <Portal>

adamabeshouse: okay cool

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -300,7 +307,8 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
 
     @autobind
     private getSvg() {
-        return this.svgContainer.firstChild;
+        return this.toSVGDOMNode();
+        //return this.svgContainer.firstChild;
```
zhx828: We can get rid of this now?

src/shared/components/TruncatedTextSVG.tsx
```diff
@@ -0,0 +1,118 @@
+import * as React from "react";
```
alisman: should we add tooltip to the name of this component?  also, would it make sense to somehow combine this with victory label so we could use it elsewhere?

alisman:
The ellipsis are right up against the right edge of chart making them still a little hard to make out.   Lets tweak to get it right.

src/shared/components/TruncatedTextSVG.tsx
adamabeshouse: yeah, this is victory compatible. it's a plug in as `labelComponent`

adamabeshouse:
@alisman @zhx828 made requested changes

PR #2494, opened 2019-06-25
add clinvar column to mutation table

leexgh:
Add clinvar column to mutation table.

1. shows clinvar id in the column
image

2. tootip for each clinVar id (the clinVar id links to https://www.ncbi.nlm.nih.gov/clinvar/variation/43594/
)
image

3. tooltip when no available data

image

### DIFF
--- a/end-to-end-tests/specs/core/mutationTable.spec.js
+++ b/end-to-end-tests/specs/core/mutationTable.spec.js
@@ -114,4 +114,29 @@ describe('Mutation Table', function() {
 
     });
 
+    describe('try getting ClinVar id from genome nexus', ()=>{
+        before(()=>{
+            var url = `${CBIOPORTAL_URL}/results/mutations?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=brca_broad&case_set_id=brca_broad_sequenced&data_priority=0&gene_list=TP53&geneset_list=%20&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=brca_broad_mutations&tab_index=tab_visualize`;
+
+            goToUrlAndSetLocalStorage(url);
+            // mutations table should be visiable after oncokb icon shows up,
+            // also need to wait for mutations to be sorted properly
+            browser.waitForVisible("tr:nth-child(1) [data-test=oncogenic-icon-image]",30000);
+        });
+
+        it('should show the ClinVar id after adding the ClinVar column', ()=>{
+            // click on column button
+            browser.click("button*=Columns");
+            // scroll down to activated "ClinVar" selection
+            browser.scroll(1000, 1000);
+            // click "clinvar"
+            browser.click('//*[text()="ClinVar ID"]');
+            let res;
+            browser.waitUntil(() => {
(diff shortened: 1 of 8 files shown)


inodb:
Great work @leexgh!

Maybe we can have some mouseover for the column header as well. Something like:

[ClinVar](https://www.ncbi.nlm.nih.gov/clinvar/) aggregates information about genomic variation and its relationship to human health.

Not sure exactly what to put in the mouseover when there is data. Maybe:

`Click to see variant on ClinVar website` (the currently listed link is the homepage for NCBI, homepage for ClinVar is https://www.ncbi.nlm.nih.gov/clinvar/)

inodb:
@jjgao 	@tmazor @schultzn any product comments on this other than what I mentioned in:

https://github.com/cBioPortal/cbioportal-frontend/pull/2494#issuecomment-505934204

Link here:

https://cbioportal-frontend-pr-2494.herokuapp.com/results/mutations?session_id=5d13b1bde4b0ab4137877d0c

Then click on Columns and enable the ClinVar column

Thanks!

tmazor:
Looks good!

I think we should just make clear that we are showing the ClinVar Variation ID, as opposed to a count or anything else. For some mutations, the number in the COSMIC column and the ClinVar variation ID can be similar, which could be confusing for anyone who isn't super familiar with ClinVar (e.g. run a query for KRAS). The fact that the ClinVar column is clearly a link helps, but maybe add that to the header tooltip? Or make the column name "ClinVar ID"?

leexgh:
all update, thanks!

inodb:
@jjgao 	any other comments?

Link here:

cbioportal-frontend-pr-2494.herokuapp.com/results/mutations?session_id=5d13b1bde4b0ab4137877d0c

Then click on Columns and enable the ClinVar column

jjgao:
- Change the head from "ClinVar ID" to "ClinVar"
- Instead of displaying the ID, is it possible to display "Interpretation"? Maybe we should add it to the Annotation column?
https://www.ncbi.nlm.nih.gov/clinvar/variation/163380/
image

jjgao:
Here is the list of interpretation: https://www.ncbi.nlm.nih.gov/clinvar/docs/clinsig/. I am wondering if we can design an icon and add different colors for different interpretations.

leexgh:
@jjgao good idea. what if one mutation has multiple interpretations? For example https://www.ncbi.nlm.nih.gov/clinvar/variation/163380/ has three, are we going to add three icons into the annotation column?

jjgao:
I think we can define priorities for each interpretation, for "Pathogenic/Likely pathogenic, drug response​", maybe we can pick Pathogenic.

Ideally we should get this into Annotation column, but defining an icon is not necessary trivial. Thoughts, @inodb ?

inodb:
Let's maybe investigate the multi column annotation sorting before adding new columns:

https://github.com/cBioPortal/cbioportal/issues/4625

Then we can do the things JJ proposed:

https://github.com/cBioPortal/cbioportal/issues/6351

Tali btw specifically requested to have the column header by "ClinVar ID" to make it clear it is not a count in the Clinvar database but an ID

jjgao:
@inodb before we achieve #6351, maybe we can show interpretation instead of clinvar IDs in the clinvar column?

jjgao:
https://github.com/cBioPortal/cbioportal/issues/6353

PR #2495, opened 2019-06-25
fix id not found error

dippindots:
Find the virtual study only when **virtualStudiesMap** not undefined.
Fix #https://github.com/cBioPortal/cbioportal/issues/6316 .


dippindots:
Sorry, no, should be the opposite. I updated the solution.

PR #2497, opened 2019-06-25
override Arial with FreeSans when download pdf

dippindots:
Change the label style back to the original style and override Arial with FreeSans when downloading pdf.
Fix # https://github.com/cbioportal/cbioportal/issues/3225.

### DIFF
--- a/src/shared/lib/svgToPdfDownload.ts
+++ b/src/shared/lib/svgToPdfDownload.ts
@@ -22,11 +22,11 @@ export default function svgToPdfDownload(fileName: string, svg: any) {
     if (height > width) {
         direction = 'p';
     }
-
     const pdf = new jsPDF(direction, 'pt', [width, height]);
 
+    // override Arial with FreeSans to display special charactors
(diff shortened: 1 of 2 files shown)


src/shared/lib/svgToPdfDownload.ts
```diff
@@ -26,7 +26,7 @@ export default function svgToPdfDownload(fileName: string, svg: any) {
     const pdf = new jsPDF(direction, 'pt', [width, height]);
 
     pdf.addFileToVFS("FreeSans-normal.ttf", font.FreeSans);
-    pdf.addFont("FreeSans-normal.ttf", "FreeSans", "normal");
+    pdf.addFont("FreeSans-normal.ttf", "Arial", "normal");
```
alisman: Lets put a comment above this explaining why were are doing this slightly strange thing

dippindots: @alisman Sure. Also removed the unused functions (we don't use promise anymore, they are not referenced).

PR #2498, opened 2019-06-25
Fork query experience

alisman:
Break homepage experience into two steps: 1. choose study, 2. if query by gene, show query criteria UI 
(hidden by default)

This will help users see the study page as a first-class citizen, sibling of results view.  Right now, most users submit queries and don't visit study view, suggesting that perhaps they don't know how to navigate to it.

Also, make study chooser height responsive.

Fixes: https://github.com/cBioPortal/cbioportal/issues/6363


src/globalStyles/global.scss
```diff
@@ -44,11 +44,16 @@ button:active, button:focus, div:focus, div:active {
 
 a {
   cursor:pointer;
-  text-decoration:none !important;
+  text-decoration:none;
+
+  &.disabled {
```
alisman: annoying.  we can't use <button> because a disabled button has no mouseover events and thus can't have tooltip to explain why disabled. classic!

src/pages/home/HomePage.tsx
```diff
@@ -78,12 +78,22 @@ export default class HomePage extends React.Component<IResultsViewPageProps, {}>
 
         return (
             <PageLayout className="homePageLayout" noMargin={true} rightBar={<RightBar queryStore={this.queryStore} />}>
+                {/*<div className={"headBlock"} dangerouslySetInnerHTML={{__html:AppConfig.serverConfig.skin_blurb!}}></div>*/}
```
alisman: i actually have to restore this to support other portals

src/pages/home/HomePage.tsx
```diff
@@ -78,12 +78,22 @@ export default class HomePage extends React.Component<IResultsViewPageProps, {}>
...
+
+                <div className={"headBlock"} style={{paddingBottom:0}}>
+
+                    <div className={"citationRule"}>
+                        Please cite: <a href="http://cancerdiscovery.aacrjournals.org/content/2/5/401.abstract" target="_blank">Cerami et al., 2012</a> &amp; <a href="http://www.ncbi.nlm.nih.gov/pubmed/23550210" target="_blank">Gao et al., 2013</a>
```
alisman: this will become configurable

src/shared/components/query/QueryContainer.tsx
```diff
@@ -1,152 +1,277 @@
...
-						<Else>
-							<LoadingIndicator isLoading={true} center={true} size={"big"}/>
-						</Else>
-					</If>
-				}
-			</FlexCol>
-        );
-    }
-}
+import * as React from "react";
```
alisman: finally changing newline style from days of Andy windows format

end-to-end-tests/specs/core/home.spec.js
```diff
@@ -18,7 +22,7 @@ var searchInputSelector = ".autosuggest input[type=text]";
 
 describe('homepage', function() {
 
-    this.retries(2);
+    this.retries(0);
```
adamabeshouse: delete line?

end-to-end-tests/specs/core/home.spec.js
```diff
@@ -1,12 +1,16 @@
...
+var {
+    goToUrlAndSetLocalStorage,
+    clickQueryByGeneButton,
+    useExternalFrontend,
+    setInputText,
+    clickQueryByGeneButton,
+    waitForNumberOfStudyCheckboxes,
+    clickModifyStudySelectionButton
+} = require('./../specUtils');
+
```
adamabeshouse: much better

adamabeshouse:
I'm seeing excess space on the bottom in firefox 
image

adamabeshouse:
the modify query screen should just fill the screen I think, right now it overshoots it

image

src/globalStyles/global.scss
adamabeshouse: since standard behavior is disabled doesnt receive mouse events, and we're potentially programming around that elsewhere, can you make this selector more specific?

src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -138,6 +138,7 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
     @computed get queryForm(){
         return <div style={{margin:"10px -20px 0 -20px"}}>
             <QueryAndDownloadTabs onSubmit={this.onSubmit}
+                                  forkedExperience={false}
```
adamabeshouse: I find it vague, can you rename this prop, maybe "bigStudyViewButton" or something?

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -170,93 +171,35 @@ export default class CancerStudySelector extends React.Component<ICancerStudySel
...
-                                </Observer>
-                            )}
 
+                            <StudySelectorStats store={this.store}/>
+
+                            {/*{(!!(!this.store.forDownloadTab) && !!(!this.store.cancerTypes.isPending && !this.store.cancerStudies.isPending)) && (*/}
+                            {/*    <Observer>*/}
+                            {/*        {() => {*/}
+                            {/*            let hasSelection = this.store.selectableSelectedStudyIds.length > 0;*/}
+
```
adamabeshouse: still need this code?

src/shared/components/query/QueryContainer.tsx
```diff
@@ -1,152 +1,277 @@
...
+                                            {!!(this.store.submitError) && (
+                                                <span className={styles.errorMessage} data-test="oqlErrorMessage">
+										            {this.store.submitError}
+									            </span>
+                                            )}
+
+                                            {this.store.oqlMessages.map(msg => {
+                                                return (
+                                                    <span className={styles.oqlMessage}>
+												        <i className='fa fa-info-circle' style={{marginRight: 5}}/>
```
adamabeshouse: some formatting issues here? wouldnt necessarily mention it besides that you're doing a reformat

src/shared/components/query/QueryContainer.tsx
```diff
@@ -1,152 +1,277 @@
...
+                                    </DefaultTooltip>
+                                    OR
+                                    <DefaultTooltip placement={"top"}
+                                                    trigger={["hover"]}
+                                                    destroyTooltipOnHide={true}
+                                                    mouseEnterDelay={0}
+                                                    mouseLeaveDelay={0}
+                                                    disabled={!this.exploreCohortsButtonDisabled}
+                                                    overlay={this.exploreCohortsButtonTooltipMessage}>
+                                            <a onClick={()=>!this.exploreCohortsButtonDisabled && this.store.openSummary()}
```
adamabeshouse: another option (not 100% sure if it'll work) would be to wrap this <a> in a div which will thus receive the tooltip mouse events

src/shared/components/query/StudySelectorStats.tsx
```diff
@@ -0,0 +1,40 @@
...
+    if (!props.store.cancerTypes.isPending && !props.store.cancerStudies.isPending && !props.store.profiledSamplesCount.isPending) {
+        return (<Observer>
+            {() => {
+                let numSelectedStudies = expr(() => props.store.selectableSelectedStudyIds.length);
+                return (
+                    <>
+                        <a
+                            onClick={() => {
+                                if (numSelectedStudies)
+                                    props.store.showSelectedStudiesOnly = !props.store.showSelectedStudiesOnly;
```
adamabeshouse: a bit hard on the eye, could you put it in brackets even though its just one line?

src/shared/components/query/QueryContainer.tsx
alisman: tried that but even with 10px padding around button it still doesn't always fire.

src/pages/resultsView/querySummary/QuerySummary.tsx
alisman: What about forkedMode?  I actually think "forkedXXX" is more descriptive.  It at least hints at the concept behind this mode.   If we can come up with a better name than "forked" I'm open to it.   I will change to mode because it's better than "experience" for sure.

adamabeshouse: sure, that works for me

src/globalStyles/global.scss
alisman: We get the benefit of the bootstrap "disabled" styles.  I think this i a better default behavior.  Not sure what the negative repercussions could be.

adamabeshouse: it means that disabled buttons can still be clicked and potentially cause bugs whereas previously onClick could assume it wasnt disabled.

just make it more specific like `&.disabled { &.yes-pointer-events {}}`

src/shared/components/query/styles/styles.module.scss
```diff
@@ -1,12 +1,43 @@
...
+
+    .forkedButtons {
+      margin-top:10px;
+      justify-content: center;
+      align-items: center;
+      > div {   // we have to use span to fix issue with mouseover events and disabled button controls
+
+      }
+      a {
+        pointer-events:all !important;
```
adamabeshouse: :+1:

PR #2500, opened 2019-06-25
GC - fix venn to always show circle borders

adamabeshouse:
BEFORE:
image

AFTER:
image

https://www.cbioportal.org/comparison/overlap?sessionId=5d127ba7e4b0ab41378778e1


alisman:
I think the white regions really throw me off.  Seems central to Venn paradigm that the you have 2 or 3 solid color circles.  The white space requires me to think too much and I'm not sure I'm able to come to any conclusion.

adamabeshouse:
@alisman just implemented this - does it make it any clearer?

image

PR #2501, opened 2019-06-26
Cancer Types Summary tab updates

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6301

1. Show whether profiled data is present or not
2. Include gene panel data


alisman:
@jjgao do you think we need to add definition for plus/minus:  - = no data available?  or is that obvious?

jjgao:
It would be a good idea to add a tooltip, - mean "Not profiled"

kalletlak:
@jjgao then what should be the symbol for samples profiled but not alterations?

kalletlak:
@jjgao should we hide bars with no data. Example, hiding `Glioma` and `Gastrointestinal Stromal Tumor` here
Screen Shot 2019-07-01 at 11 24 11 AM

jjgao:
> @jjgao then what should be the symbol for samples profiled but not alterations?

It should be +.

@jjgao should we hide bars with no data. Example, hiding Glioma and Gastrointestinal Stromal Tumor here

I think that's a good idea, but only hide bars when none of the samples are profiled in any of the profiles.

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
```diff
@@ -60,61 +76,145 @@ export function mergeAlterationDataAcrossAlterationTypes(alterationData:ICancerS
...
-        return this.bodyWidth + this.rightPadding + this.leftPadding;
+    @computed get barPlotTooltipComponent() {
+        if (!this.barPlotTooltipModel) {
+            return null;
+        } else {
+            const tooltipModel = this.barPlotTooltipModel
+            const maxWidth = 400;
+            let tooltipPlacement = (this.mousePosition.x > WindowStore.size.width - maxWidth ? "left" : "right");
+
+            return (ReactDOM as any).createPortal(
```
alisman: Not important but you can use <Portal> component for this

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
```diff
@@ -162,14 +257,13 @@ export class CancerSummaryChart extends React.Component<CancerSummaryChartProps,
...
-                                        x:props.x + 20 - this.scrollPane.scrollLeft,
-                                        y:props.y - 18,
+                                    self.barPlotTooltipModel = {
+                                        ...props,
                                         groupName:props.datum.x,
                                         alterationData:self.props.countsByGroup[props.datum.xKey]
                                     };
                                 } else {
-                                    self.hideTooltip();
+                                    self.barPlotTooltipModel = null;
```
alisman: why not just change this inside hideTooltip method?

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
```diff
@@ -245,65 +325,210 @@ export class CancerSummaryChart extends React.Component<CancerSummaryChartProps,
...
+    private barChartHeight() {
+        return 400;
+    }
 
-        // if we're not showing result broken down the alterations then we need to merge
+    @computed get scatterChartHeight() {
+        return this.scatterChartExtent;
+    }
+
+    @computed get scatterChartExtent() {
```
alisman: seems poorly named.  If it's only being used by the height getter above, why not just put the calculation there?

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
```diff
@@ -245,65 +325,210 @@ export class CancerSummaryChart extends React.Component<CancerSummaryChartProps,
...
-                    <div ref={(el:HTMLDivElement)=>this.scrollPane=el} style={{overflowX:'auto', overflowY:'hidden'}}>
-                    {
-                        (this.tooltipModel) && (this.buildTooltip(this.tooltipModel))
-                    }
-                    <VictoryChart width={this.width} height={this.height} theme={CBIOPORTAL_VICTORY_THEME}
-                                  padding={{bottom:this.bottomPadding, top:this.topPadding, left:this.leftPadding, right:this.rightPadding }} domainPadding={{x: 20, y: 20}}
-                                  containerComponent={<VictoryContainer containerRef={(ref: any) => this.svgContainer = ref} responsive={false}/>}
+            <div style={this.overflowStyle} className="borderedChart">
+                <div style={{ overflowX: 'auto', overflowY: 'hidden' }}>
+                    <svg
```
alisman: what's reason for all this extra wrapping?

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
kalletlak: it's the same statement even in that method

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
kalletlak: this is to have both bar plot and scatter plot together

src/pages/resultsView/cancerSummary/CancerSummaryChart.tsx
alisman: then why not just use it?   not important, but just trying to understand

PR #2503, opened 2019-06-26
Remove dimension/chartType from ChartMeta

zhx828:
Move dimension and chartType out of ChartMeta to prevent unnecessary computing.
dimension and chartType are frequently updated and should not be associated with any promise if not related.

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5549

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -81,7 +81,7 @@ import {
     showOriginStudiesInSummaryDescription,
     SPECIAL_CHARTS,
     StudyWithSamples,
-    submitToPage,
+    submitToPage, ChartMetaWithDimensionAndChartType,
     UniqueKey
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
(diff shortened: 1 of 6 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -638,9 +638,9 @@ export class StudyViewPageStore {
 
     private geneMapCache:{[entrezGeneId:number]:string} = {};
 
-    @observable private chartsDimension: { [uniqueKey: string]: ChartDimension } = {};
+    @observable public chartsDimension = observable.map<ChartDimension>();
```
alisman: i know this was existing code but I think we should change this to chartDimensions (plural at end).  It hard to see s in the middle.  Is an array of ChartDimension so ...

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -638,9 +638,9 @@ export class StudyViewPageStore {
...
 
-    @observable private chartsType = observable.map<ChartType>();
+    @observable public chartsType = observable.map<ChartType>();
```
alisman: same here.  ChartTypes

PR #2513, opened 2019-06-28
GC - only use venn diagram for 2 groups

adamabeshouse:
image

### DIFF
--- a/src/globalStyles/global.scss
+++ b/src/globalStyles/global.scss
@@ -555,6 +555,10 @@ h3.hr {
     margin-bottom:0 !important;
   }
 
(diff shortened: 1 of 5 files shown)


alisman:
Not a big fan of the contained icon.  Also, we should try to make the icons for alerts roughly equal size.   I think this could be done pretty easily with existing fontawesome icons nested.

src/globalStyles/global.scss
```diff
@@ -555,6 +555,10 @@ h3.hr {
     margin-bottom:0 !important;
   }
 
+  .alert {
```
alisman: maybe prevent this from last-child?

PR #2514, opened 2019-06-28
GC - add overlap exclusion strategy to URL

adamabeshouse:
(default exclude)

example URL is https://www.cbioportal.org/comparison/overlap?overlapStrategy=Include&sessionId=5d165f9be4b0ab4137878513

### DIFF
--- a/end-to-end-tests/specs/core/groupComparison.screenshot.spec.js
+++ b/end-to-end-tests/specs/core/groupComparison.screenshot.spec.js
@@ -7,6 +7,7 @@ var checkElementWithTemporaryClass = require('./../specUtils').checkElementWithT
 var checkElementWithMouseDisabled = require('./../specUtils').checkElementWithMouseDisabled;
 
 const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+const OverlapStrategy = require("./../../../src/pages/groupComparison/GroupComparisonStore").OverlapStrategy;
 
 
 describe("group comparison page screenshot tests", function () {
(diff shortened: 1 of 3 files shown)


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -59,21 +59,29 @@ import ifndef from "../../shared/lib/ifndef";
...
     @observable private _currentTabId:GroupComparisonTab|undefined = undefined;
-    @observable private _overlapStrategy:OverlapStrategy = OverlapStrategy.EXCLUDE;
     @observable private sessionId:string;
     @observable public newSessionPending = false;
 
     constructor(sessionId:string, private appStore:AppStore, private routing:any) {
         this.sessionId = sessionId;
     }
 
+    @action public updateOverlapStrategy(strategy:OverlapStrategy) {
```
alisman: is there a reason to do this in store?  if possible it seems to me the Page component is the right thing be updating router.

adamabeshouse: but then the value is get and set in different places?

adamabeshouse: Ill merge it for now because we're already doing a bunch of URL manipulation in the store anyway. So if we want to change that we can deal with it all at once.

end-to-end-tests/specs/core/groupComparison.screenshot.spec.js
```diff
@@ -7,6 +7,7 @@ var checkElementWithTemporaryClass = require('./../specUtils').checkElementWithT
 var checkElementWithMouseDisabled = require('./../specUtils').checkElementWithMouseDisabled;
 
 const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+const OverlapStrategy = require("./../../../src/pages/groupComparison/GroupComparisonStore").OverlapStrategy;
```
alisman: @adamabeshouse this actually brok the whole spec file.   We can't import ts files into js because none of this is compiled for typescript.

PR #2515, opened 2019-06-28
Only fetch gene panel info when needed

zhx828:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6337

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -35,7 +35,7 @@ import {
     ClinicalData,
     ClinicalDataMultiStudyFilter,
     CopyNumberSeg,
-    Gene,
+    Gene, GenePanel, GenePanelData,
     MolecularProfile,
     MolecularProfileFilter,
     Patient
@@ -121,6 +121,7 @@ import {sleepUntil} from "../../shared/lib/TimeUtils";
 import ComplexKeyMap from "../../shared/lib/complexKeyDataStructures/ComplexKeyMap";
 import jStat from 'jStat'
 import {CancerGene, Gene as OncokbGene} from "../../shared/api/generated/OncoKbAPI";
+import MobxPromiseCache from "shared/lib/MobxPromiseCache";
 
 
 export enum StudyViewPageTabKeyEnum {
@@ -2661,6 +2662,12 @@ export class StudyViewPageStore {
         onError: (error => {}),
         default: []
     });
+
+    public genePanelCache = new MobxPromiseCache<{ genePanelId: string }, GenePanel>(q => ({
+        invoke: () => {
+            return defaultClient.getGenePanelUsingGET(q);
+        }
+    }));
 
     readonly mutatedGeneData = remoteData<MutationCountByGeneWithCancerGene[]>({
         await: () => this.oncokbCancerGeneFilterEnabled ?
--- a/src/pages/studyView/table/CNAGenesTable.tsx
+++ b/src/pages/studyView/table/CNAGenesTable.tsx
@@ -6,7 +6,6 @@ import { observer } from "mobx-react";
 import styles from "./tables.module.scss";
 import {
     CopyNumberGeneFilterElement,
-    GenePanelToGene
 } from "shared/api/generated/CBioPortalAPIInternal";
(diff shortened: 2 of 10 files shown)


zhx828:
The pr can only merged when the backend pr is deployed. https://github.com/cBioPortal/cbioportal/pull/6341

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2662,6 +2662,19 @@ export class StudyViewPageStore {
         default: []
     });
 
+    private genePanelsCache: { [genePanelId: string]: GenePanel } = {};
```
kalletlak: I think caching is not required as browser already caches it.

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -68,11 +69,11 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
     @observable private modalSettings: {
         modalOpen: boolean;
         modalPanelName: string;
-        modalPanelGenes: GenePanelToGene[];
+        modalGetPanelData?: (genePanelId: string) => Promise<GenePanel>;
     } = {
         modalOpen: false,
         modalPanelName: "",
-        modalPanelGenes: []
+        modalGetPanelData: undefined
```
kalletlak: by default this would be undefined

src/pages/studyView/StudyViewPageStore.ts
zhx828: Ah, right. The design was a bit different when I introduced the cache. Good catch, let me remove.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2662,6 +2662,11 @@ export class StudyViewPageStore {
         default: []
     });
 
+    @autobind
+    public async getGenePanelData(genePanelId: string): Promise<GenePanel> {
+        return Promise.resolve(await defaultClient.getGenePanelUsingGET({genePanelId: genePanelId}));
```
adamabeshouse: this is equivalent to just returning `defaultClient....` , doesnt need to be `async`

src/pages/studyView/table/GenePanelModal.tsx
```diff
@@ -40,25 +45,41 @@ export const GenePanelList: React.FunctionComponent<IGenePanelTooltipProps> = ({
...
-            </Modal.Footer>
-        </Modal>
-    );
+@observer
+export class GenePanelModal extends React.Component<IGeneModalProps, {}> {
+    @observable dataLoading = false;
+    @observable genes: GenePanelToGene[] = [];
+
+    @autobind
+    componentWillReceiveProps(nextProps: Readonly<IGeneModalProps>): void {
```
adamabeshouse: why do you want to do it like this instead of using a mobxpromise? seems more complicated this way

adamabeshouse: one construct that might help which you may not be aware of is MobxPromiseCache, which handles the case here - you would pass in the panel name, and it would create a promise based on that panel name. Let me know if you want to talk about it

src/pages/studyView/table/CNAGenesTable.tsx
```diff
@@ -68,11 +70,10 @@ export class CNAGenesTable extends React.Component<ICNAGenesTablePros, {}> {
     @observable private modalSettings: {
         modalOpen: boolean;
         modalPanelName: string;
-        modalPanelGenes: GenePanelToGene[];
+        modalGetPanelData?: (genePanelId: string) => Promise<GenePanel>;
```
adamabeshouse: we can get rid of this now right?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2662,6 +2663,12 @@ export class StudyViewPageStore {
         default: []
     });
 
+    public genePanelCache = new MobxPromiseCache<{ genePanelId: string }, GenePanel>(q => ({
+        invoke: async () => {
```
kalletlak: async not needed

zhx828:
Two failing tests are not related to the pr.

PR #2516, opened 2019-06-28
fix case set not working for virtual study

dippindots:
do not concatenate all sample specs
Fix # https://github.com/cBioPortal/cbioportal/issues/5634.

### DIFF
--- a/src/pages/resultsView/ResultsViewPageHelpers.ts
+++ b/src/pages/resultsView/ResultsViewPageHelpers.ts
@@ -82,15 +82,17 @@ export function substitutePhysicalStudiesForVirtualStudies(cancerStudyIds:string
 
 }
 
-export function populateSampleSpecificationsFromVirtualStudies(samplesSpecification:SamplesSpecificationElement[], virtualStudies:VirtualStudy[]){
+export function populateSampleSpecificationsFromVirtualStudies(samplesSpecifications:SamplesSpecificationElement[], virtualStudies:VirtualStudy[]){
 
     const virtualStudiesKeyedById = _.keyBy(virtualStudies,(virtualStudy)=>virtualStudy.id);
-
+    const samplesSpecificationsKeyedByStudyId = _.keyBy(samplesSpecifications,(spec)=>spec.studyId);
     // remove specs for virtual studies (since they mean nothing to api)
     // and then populate with ids
-    samplesSpecification = _.filter(samplesSpecification,(spec)=>!virtualStudiesKeyedById[spec.studyId]);
+    samplesSpecifications = _.filter(samplesSpecifications,(spec)=>!virtualStudiesKeyedById[spec.studyId]);
 
-    const allVirtualStudySampleSpecs = _.flatMapDeep(virtualStudies.map((virtualStudy)=>{
+    // only add sample specs when this virtual study is in the samplesSpecifications
+    const selectedVirtualStudies = _.filter(virtualStudies, (virtualStudy) => samplesSpecificationsKeyedByStudyId[virtualStudy.id]);
+    const allSelectedVirtualStudySampleSpecs = _.flatMapDeep(selectedVirtualStudies.map((virtualStudy)=>{
         return virtualStudy.data.studies.map((study)=>{
             return study.samples.map((sampleId)=>{
                 return {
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -102,8 +102,11 @@ export function populateSampleSpecificationsFromVirtualStudies(samplesSpecificat
         }) as SamplesSpecificationElement[][];
     }));
 
-    // ts not resolving type above and not sure why, so cast it
-    samplesSpecification = samplesSpecification.concat(allVirtualStudySampleSpecs as SamplesSpecificationElement[]);
+    // concatenate allVirtualStudySampleSpecs only when we don't have any samplesSpecifications
+    if (samplesSpecification.length === 0) {
```
alisman: if we have a mixture of virtual studies and physical studies, there might be a problem b/c there will be sampleSpecifications for the physical studies.

alisman: please try that possibility and make sure we handle it properly

src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -85,12 +85,14 @@ export function substitutePhysicalStudiesForVirtualStudies(cancerStudyIds:string
 export function populateSampleSpecificationsFromVirtualStudies(samplesSpecification:SamplesSpecificationElement[], virtualStudies:VirtualStudy[]){
 
     const virtualStudiesKeyedById = _.keyBy(virtualStudies,(virtualStudy)=>virtualStudy.id);
-
+    const samplesSpecificationKeyedByStudyId = _.keyBy(samplesSpecification,(spec)=>spec.studyId);
     // remove specs for virtual studies (since they mean nothing to api)
     // and then populate with ids
     samplesSpecification = _.filter(samplesSpecification,(spec)=>!virtualStudiesKeyedById[spec.studyId]);
```
alisman: probably should be sampleSpecifications.  Plural should be at end because it's an array of specifications.

src/pages/resultsView/ResultsViewPageHelpers.ts
```diff
@@ -82,15 +82,17 @@ export function substitutePhysicalStudiesForVirtualStudies(cancerStudyIds:string
...
-export function populateSampleSpecificationsFromVirtualStudies(samplesSpecification:SamplesSpecificationElement[], virtualStudies:VirtualStudy[]){
+export function populateSampleSpecificationsFromVirtualStudies(samplesSpecifications:SamplesSpecificationElement[], virtualStudies:VirtualStudy[]){
 
     const virtualStudiesKeyedById = _.keyBy(virtualStudies,(virtualStudy)=>virtualStudy.id);
-
+    const samplesSpecificationsKeyedByStudyId = _.keyBy(samplesSpecifications,(spec)=>spec.studyId);
     // remove specs for virtual studies (since they mean nothing to api)
     // and then populate with ids
-    samplesSpecification = _.filter(samplesSpecification,(spec)=>!virtualStudiesKeyedById[spec.studyId]);
+    samplesSpecifications = _.filter(samplesSpecifications,(spec)=>!virtualStudiesKeyedById[spec.studyId]);
```
alisman: @dippindots i actually want "sampleSpecifications".    the plural "s" is on specifications, not on samples.   "sample" is a descriptive word.  I know you didn't write the code from the beginning but I want to make sure I communicate proper pluralization rules for you for the future!

alisman: Once we make that change, we can release this.

dippindots: Thanks, @alisman! That helps a lot, I will remember this in the future. The only reason I change it to `samplesSpecifications` is that it is an array of `SamplesSpecificationElement`: https://github.com/cBioPortal/cbioportal-frontend/pull/2516/files#diff-98feaa0c1f30ed49f15d832754badb45R85. And the `SamplesSpecificationElement` is the specification of the `sampleList`(list of samples), I don't know if we should just use `sampleSpecifications` in this circumstances.

PR #2518, opened 2019-07-02
Allow user defines the bar chart bins

zhx828:
### What & Why
This fixes https://github.com/cbioportal/cbioportal/issues/5819

### Update bins
Aug-14-2019 10-22-04

### When bin definition is invalid
Aug-14-2019 10-22-59

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -82,7 +82,7 @@ import {
     SPECIAL_CHARTS,
     StudyWithSamples,
     submitToPage, ChartMetaWithDimensionAndChartType,
-    UniqueKey
+    UniqueKey, ClinicalDataType
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
 import {SingleGeneQuery} from 'shared/lib/oql/oql-parser';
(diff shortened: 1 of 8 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1299,7 +1299,27 @@ export class StudyViewPageStore {
             !this._clinicalDataBinFilterSet.get(uniqueKey)!.disableLogScale;
     }
 
-    @action addCharts(visibleChartIds:string[]) {
+    @autobind
+    @action
+    public updateCustomBins(uniqueKey: string, bins: number[]) {
+        let newFilter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey))!;
```
zhx828: Here is the observable map we discussed.

src/pages/studyView/chartHeader/ChartHeader.tsx
```diff
@@ -139,20 +143,45 @@ export class ChartHeader extends React.Component<IChartHeaderProps, {}> {
             items.push(
                 <li>
                     <a className="dropdown-item"
+                       style={{display: 'flex'}}
```
alisman: is it necessary?

alisman:
@zhx828 think maybe a single line of instructional text would help above the text area.   also, @jjgao, i experienced some funky behavior with the flyout from the flyout behavior.  hard to reproduce.  the bins flyout switched alignment.   a simpler, most robust UI would be to open it in a modal dialog.   the drawback would be that the user couldn't see the underlying chart for reference.  but that's true even now when the dialog opens to the left.   
image

src/pages/studyView/StudyViewPageStore.ts
alisman: @zhx828 we should examine why this causes entire studyview to reload.  i think this the the only action which takes place, right?  i agree with @jjgao that this should really not result in a reload.   i would be interested in getting to the bottom of it with you

jjgao:
I think a modal dialog may be a good idea in this case. Some description would also be good.

zhx828:
@jjgao @alisman I updated the PR with modal implementation. The gif above has been updated. You can also check the feather here https://cbioportal-frontend-pr-2518.herokuapp.com/study/summary?id=acc_tcga_pan_can_atlas_2018
Let me know what you think.

src/pages/studyView/StudyViewPageStore.ts
zhx828: This issue has been resolved.

zhx828:
The failing e2e is not related to this pr.

PR #2524, opened 2019-07-08
update e2e test

dippindots:
update the e2e mutational table test.

### DIFF
--- a/end-to-end-tests/specs/core/mutationTable.spec.js
+++ b/end-to-end-tests/specs/core/mutationTable.spec.js
@@ -84,7 +84,7 @@ describe('Mutation Table', function() {
 
         it('should show the gnomad table after mouse over the frequency in gnomad column', ()=>{
 
-            browser.waitForText('//*[text()="TCGA-78-7540-01"]',60000);
+            browser.waitForText('//*[text()="LUAD-B00416-Tumor"]',60000);
             // show the gnomad column
             browser.scroll(1000, 0);
             // click on column button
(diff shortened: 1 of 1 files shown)


end-to-end-tests/specs/core/mutationTable.spec.js
```diff
@@ -94,21 +94,21 @@ describe('Mutation Table', function() {
             // click "GNOMAD"
             browser.click('//*[text()="gnomAD"]');
             // find frequency
-            const frequency = '[data-test2="TCGA-78-7540-01"][data-test="gnomad-column"]';
+            const frequency = '[data-test2="LUAD-B00416-Tumor"][data-test="gnomad-column"]';
```
alisman: @inodb @jjgao can you confirm it makes sense that the sampleIds would change in this way?

dippindots: @alisman Maybe it is because we changed the default sorting of this table? This sample `TCGA-78-7540-01` still exists, just not on the top of the list. In this [link](https://www.cbioportal.org/results/mutations?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=nsclc_tcga_broad_2016&case_set_id=nsclc_tcga_broad_2016_cnaseq&data_priority=0&gene_list=TP53&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=nsclc_tcga_broad_2016_cna&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=nsclc_tcga_broad_2016_mutations&tab_index=tab_visualize) If we sort by `# Mut in Sample` then this sample will come to the top of the list.

alisman: did we change the default sorting?

dippindots: @alisman That's the weird part. I haven't found the related code change in history. Only knows this happened two weeks ago on `RC` branch: https://github.com/cBioPortal/cbioportal-frontend/pull/2487

PR #2525, opened 2019-07-08
disable mouse event for oncoprint tests

dippindots:
Disable mouse event for oncoprint tests
Fix # https://github.com/cBioPortal/cbioportal/issues/6325.


dippindots:
@alisman it works for SVG, same as the Venn diagram.

adamabeshouse:
@dippindots the mouse events that are problematic with the oncoprint actually occur on `canvas` objects, not svg. so hopefully this will still work

dippindots:
Thanks for pointing this out @adamabeshouse. I tested it again, and it works fine!

alisman:
The problem is quite intermittent so only time will tell if it's really fixed.

PR #2526, opened 2019-07-08
GC - use new group label components for overlap tab and survival tab

adamabeshouse:
Addresses https://github.com/cBioPortal/cbioportal/issues/6218

image

image

image

image


jjgao:
@adamabeshouse looks good overall. For UpSet plot, I am wondering if it's possible (and looks good) to just add a colored square before the group name?

adamabeshouse:
@jjgao how does this look?
image

jjgao:
@adamabeshouse I like the square version. Maybe don't worry about the ticks in clinical tab and I think other tabs look good.

src/pages/groupComparison/Survival.tsx
```diff
@@ -75,7 +79,7 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
                             name,
                             color: blendColors(partitionGroupUids.map(uid => uidToGroup[uid].color)),
                             value,
-                            legendText: name
+                            legendText: JSON.stringify(partitionGroupUids),
```
alisman: just want to confirm this is intended

adamabeshouse: yes its a bit weird but that's how I'm passing it through the system I have rigged up

PR #2532, opened 2019-07-09
MatchMiner Trial Match tab in patient view

victoria34:
# What? Why?
Create Trial Match Tab in Patient View to display matched trials.
image

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -47,10 +47,12 @@ import {QueryParams} from "url";
 import {AppStore} from "../../AppStore";
 import request from 'superagent';
 import {remoteData} from "../../public-lib/api/remoteData";
+import TrialMatchTable from "./trialMatch/TrialMatchTable";
 
 import 'cbioportal-frontend-commons/styles.css';
 import 'react-mutation-mapper/dist/styles.css';
 import 'react-table/react-table.css';
+import getBrowserWindow from "../../public-lib/lib/getBrowserWindow";
 
 const patientViewPageStore = new PatientViewPageStore();
 
@@ -192,6 +194,11 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
     hideTissueImageTab(){
         return patientViewPageStore.hasTissueImageIFrameUrl.isPending || patientViewPageStore.hasTissueImageIFrameUrl.isError
             || (patientViewPageStore.hasTissueImageIFrameUrl.isComplete && !patientViewPageStore.hasTissueImageIFrameUrl.result);
+    }
+
+    private shouldShowTrialMatch(patientViewPageStore: PatientViewPageStore): boolean {
+        return getBrowserWindow().localStorage.trialmatch === 'true' &&
+            patientViewPageStore.detailedTrialMatches.isComplete && patientViewPageStore.detailedTrialMatches.result.length > 0;
     }
 
 
@@ -537,6 +544,13 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                         </div>
                     </MSKTab>
                     )}
+
+                    <MSKTab key={7} id="trialMatchTab" linkText="Matched Trials"
+                            hide={!this.shouldShowTrialMatch(patientViewPageStore)}>
+                        <TrialMatchTable sampleManager={sampleManager}
+                                         detailedTrialMatches={patientViewPageStore.detailedTrialMatches.result}
+                                         containerWidth={WindowStore.size.width-20}/>
+                    </MSKTab>
 
                     {/*<MSKTab key={5} id="mutationalSignatures" linkText="Mutational Signature Data" hide={true}>*/}
                         {/*<div className="clearfix">*/}
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -78,6 +78,10 @@ import { ClinicalAttribute } from 'shared/api/generated/CBioPortalAPI';
 import getBrowserWindow from "../../../public-lib/lib/getBrowserWindow";
 import {getNavCaseIdsCache} from "../../../shared/lib/handleLongUrls";
 import {CancerGene} from "public-lib/api/generated/OncoKbAPI";
+import { fetchTrialsById, fetchTrialMatchesUsingPOST } from "../../../shared/api/MatchMinerAPI";
+import { IDetailedTrialMatch, ITrial, ITrialMatch, ITrialQuery } from "../../../shared/model/MatchMiner";
+import { groupTrialMatchesById } from "../trialMatch/TrialMatchTableUtils";
+
 
 type PageMode = 'patient' | 'sample';
 
@@ -806,5 +810,60 @@ export class PatientViewPageStore {
     @action clearErrors() {
         this.ajaxErrors = [];
     }
+
+    readonly trialMatches = remoteData<ITrialMatch[]>({
+        invoke: () => {
+            return fetchTrialMatchesUsingPOST({mrn: this.patientId});
+        }
+    }, []);
+
+    readonly trialIds = remoteData<ITrialQuery>({
+        await: () => [
+            this.trialMatches
+        ],
+        invoke: async() => {
+            let nctIds = new Set<string>(); // Trial unique id from clinicaltrials.gov
+            let protocolNos = new Set<string>(); // Trials's MSK ID same as protocol_number or protocol_id
+            _.forEach(this.trialMatches.result, (trialMatch: ITrialMatch) => {
+                if (_.isEmpty(trialMatch.protocolNo)) {
+                    nctIds.add(trialMatch.nctId);
+                } else {
+                    protocolNos.add(trialMatch.protocolNo);
+                }
+            });
+            return {
+                nct_id: [...nctIds],
+                protocol_no: [...protocolNos]
+            };
+        }
+    }, {
+        nct_id: [],
+        protocol_no: []
+    });
+
+    readonly trials = remoteData<ITrial[]>({
+        await: () => [
+            this.trialIds
+        ],
+        invoke: async () => {
+            if (this.trialIds.result.protocol_no.length > 0 || this.trialIds.result.nct_id.length > 0) {
+                return fetchTrialsById(this.trialIds.result);
+            }
+            return [];
+        }
+    }, []);
+
+    readonly detailedTrialMatches = remoteData<IDetailedTrialMatch[]>({
+        await: () => [
+            this.trials,
+            this.trialMatches
+        ],
+        invoke: async () => {
+            if (this.trials.result && this.trialMatches.result ) {
+                return groupTrialMatchesById(this.trials.result, this.trialMatches.result);
+            }
+            return [];
+        }
+    }, []);
 
 }
--- a/src/pages/patientView/trialMatch/TrialMatchTable.tsx
+++ b/src/pages/patientView/trialMatch/TrialMatchTable.tsx
@@ -1,0 +1,266 @@
+import * as React from 'react';
+import { If, Then, Else } from 'react-if';
+import {observer} from "mobx-react";
+import {
+    IClinicalGroupMatch, IGenomicGroupMatch, IGenomicMatch, IDetailedTrialMatch, IArmMatch
+} from "../../../shared/model/MatchMiner";
+import styles from './style/trialMatch.module.scss';
+import { computed } from "mobx";
+import LazyMobXTable from "../../../shared/components/lazyMobXTable/LazyMobXTable";
+import SampleManager from "../sampleManager";
+import DefaultTooltip, { placeArrowBottomLeft } from "../../../public-lib/components/defaultTooltip/DefaultTooltip";
+
+export type ITrialMatchProps = {
+    sampleManager: SampleManager | null;
+    detailedTrialMatches: IDetailedTrialMatch[];
+    containerWidth: number;
+}
+
+enum ColumnKey {
+    ID = 'ID',
+    TITLE = 'Title',
+    MATCHING_CRITERIA = 'Matching Criteria'
+}
+
+enum ColumnWidth {
+    ID = 140
+}
+
+class TrialMatchTableComponent extends LazyMobXTable<IDetailedTrialMatch> {
+
+}
+
+@observer
+export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
+
+    @computed
+    get columnWidths() {
+        return {
+            [ColumnKey.ID]: ColumnWidth.ID,
+            [ColumnKey.TITLE]: 0.35 * (this.props.containerWidth - ColumnWidth.ID),
+            [ColumnKey.MATCHING_CRITERIA]: 0.65 * (this.props.containerWidth - ColumnWidth.ID)
+        };
+    }
+
+    private _columns = [{
+        name: ColumnKey.ID,
+        render: (trial: IDetailedTrialMatch) => (
+            <div>
+                <If condition={trial.protocolNo.length > 0}>
+                    <div><a target="_blank" href={"https://www.mskcc.org/cancer-care/clinical-trials/" + trial.protocolNo}>{trial.protocolNo}</a></div>
+                </If>
+                <If condition={trial.nctId.length > 0}>
+                    <div><a target="_blank" href={"https://clinicaltrials.gov/ct2/show/" + trial.nctId}>{trial.nctId}</a></div>
+                </If>
+                <div>{trial.status}</div>
+            </div>
+        ),
+        sortBy: (trial: IDetailedTrialMatch) => trial.protocolNo,
+        width: this.columnWidths[ColumnKey.ID]
+    }, {
+        name: ColumnKey.TITLE,
+        render: (trial: IDetailedTrialMatch) => (<span>{trial.shortTitle}</span>),
+        sortBy: (trial: IDetailedTrialMatch) => trial.shortTitle,
+        width: this.columnWidths[ColumnKey.TITLE]
+    }, {
+        name: ColumnKey.MATCHING_CRITERIA,
+        render: (trial: IDetailedTrialMatch) => (
+            <div>
+                {trial.matches.map((armMatch: IArmMatch, index: number) => (
+                    <div>
+                        <div>
+                            {armMatch.matches.map((clinicalGroupMatch: IClinicalGroupMatch, cgIndex:number) => (
+                                <div className={styles.criteriaContainer}>
+                                    {this.getGenomicMatch(clinicalGroupMatch)}
+                                    {this.getClinicalMatch(clinicalGroupMatch)}
+                                    <If condition={cgIndex < armMatch.matches.length - 1}><hr className={styles.criteriaHr}/></If>
+                                </div>
+                            ))}
+                        </div>
+                         <If condition={armMatch.armDescription !== ''}>
+                            <div className={styles.armDiv}>
+                                <span>Arm: {armMatch.armDescription}</span>
+                            </div>
+                        </If>
+                        <If condition={armMatch.drugs.length > 0}>
+                            <div className={styles.armDiv}>
+                                <span>Intervention: {armMatch.drugs.join(', ')}</span>
+                            </div>
+                        </If>
+                        <If condition={index < trial.matches.length - 1}><hr className={styles.criteriaHr}/></If>
+                    </div>
+                ))}
+            </div>
+        ),
+        width: this.columnWidths[ColumnKey.MATCHING_CRITERIA]
+    }];
+
+    public getClinicalMatch(clinicalGroupMatch: IClinicalGroupMatch) {
+        return (
+            <div className={styles.firstRight}>
+                <span className={styles.secondLeft}>{clinicalGroupMatch.trialAgeNumerical + ' yrs old'}</span>
+                <span className={styles.secondRight}>
+                    {clinicalGroupMatch.trialOncotreePrimaryDiagnosis.positive.join(', ')}
+                    {clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.length > 0 &&
+                        <span>
+                            <b> except in </b>
+                            <If condition={clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.length < 4}>
+                                <Then>
+                                    <span>{clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.join(', ').replace(/,(?!.*,)/gmi, ' and')}</span>
+                                </Then>
+                                <Else>
+                                     <DefaultTooltip
+                                         placement='bottomLeft'
+                                         trigger={['hover', 'focus']}
+                                         overlay={this.tooltipClinicalContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative)}
+                                         destroyTooltipOnHide={true}
+                                         onPopupAlign={placeArrowBottomLeft}>
+                                        {this.mainContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.length + ` cancer types`)}
+                                    </DefaultTooltip>
+                                </Else>
+                            </If>
+                        </span>
+                    }
+                </span>
+            </div>
+        );
+    }
+
+    public getGenomicMatch(clinicalGroupMatch: IClinicalGroupMatch) {
+        return (
+            <div className={styles.firstLeft}>
+                <div>
+                    {clinicalGroupMatch.matches.map((genomicGroupMatch: IGenomicGroupMatch) => (
+                        <div>
+                            <If condition={genomicGroupMatch.matches.length === 1 &&
+                                genomicGroupMatch.genomicAlteration === `${genomicGroupMatch.matches[0].trueHugoSymbol} ${genomicGroupMatch.matches[0].trueProteinChange}`}>
+                                <Then>
+                                    {this.getGenomicExactMatch(genomicGroupMatch)}
+                                </Then>
+                                <Else>
+                                    <span className={styles.firstLeft}>{`${genomicGroupMatch.genomicAlteration}: `}
+                                        {this.getGenomicVariantCategoryMatch(genomicGroupMatch)}
+                                    </span>
+                                </Else>
+                            </If>
+                        </div>
+                    ))}
+                </div>
+                <div>
+                    { clinicalGroupMatch.notMatches.length > 0 &&
+                        <DefaultTooltip
+                            placement='bottomLeft'
+                            trigger={['hover', 'focus']}
+                            overlay={this.tooltipGenomicContent(clinicalGroupMatch.notMatches)}
+                            destroyTooltipOnHide={false}
+                            onPopupAlign={placeArrowBottomLeft}>
+                            {this.mainContent(`No alterations in ${this.getHugoSymbolName(clinicalGroupMatch.notMatches, 3)} defined by the trial`)}
+                        </DefaultTooltip>
+                    }
+                </div>
+            </div>
+        );
+    }
+
+    // Patient Genomic(i.e., BRAF V600E) info are exactly matched to trial genomic alteration(i.e., BRAF V600E)
+    public getGenomicExactMatch(genomicGroupMatch: IGenomicGroupMatch) {
+        return (
+            <div>{`${genomicGroupMatch.genomicAlteration} `}
+                {genomicGroupMatch.matches[0].sampleIds.map((sampleId: string) => (
+                    <span className={styles.genomicSpan}>
+                        {this.props.sampleManager!.getComponentForSample(sampleId, 1, '')}
+                    </span>
+                ))}
+            </div>
+        );
+    }
+
+    // Patient Genomic(i.e., BRAF V600E) info are matched to trial genomic alteration(i.e., BRAF Oncogenic Mutation).
+    // V600E belongs to Oncogenic Mutation.
+    public getGenomicVariantCategoryMatch(genomicGroupMatch: IGenomicGroupMatch) {
+        return (
+            <If condition={genomicGroupMatch.matches.length > 1}>
+                <Then>
+                    <ul className={styles.alterationUl}>
+                        {genomicGroupMatch.matches.map((genomicMatch: IGenomicMatch) => (
+                            <li>{`${genomicMatch.trueHugoSymbol} ${genomicMatch.trueProteinChange} `}
+                                {genomicMatch.sampleIds.map((sampleId: string) => (
+                                    <span className={styles.genomicSpan}>
+                                        {this.props.sampleManager!.getComponentForSample(sampleId, 1, '')}
+                                    </span>
+                                ))}
+                            </li>
+                        ))}
+                    </ul>
+                </Then>
+                <Else>
+                    <span>{`${genomicGroupMatch.matches[0].trueHugoSymbol} ${genomicGroupMatch.matches[0].trueProteinChange} `}
+                        {genomicGroupMatch.matches[0].sampleIds.map((sampleId: string) => (
+                            <span className={styles.genomicSpan}>
+                                {this.props.sampleManager!.getComponentForSample(sampleId, 1, '')}
+                            </span>
+                        ))}
+                    </span>
+                </Else>
+            </If>
+        );
+    }
+
+    public tooltipGenomicContent(data: IGenomicGroupMatch[]) {
+        return (
+            <div className={styles.tooltip}>
+                {data.map((genomicGroupMatch: IGenomicGroupMatch) => (
+                    <div className={styles.genomicSpan}><b>Not </b>{`${genomicGroupMatch.genomicAlteration.replace(/!/g, '')} `}
+                        {genomicGroupMatch.matches[0].sampleIds.map((sampleId: string) => (
+                            <span className={styles.genomicSpan}>
+                                {this.props.sampleManager!.getComponentForSample(sampleId, 1, '')}
+                            </span>
+                        ))}
+                    </div>
+                ))}
+            </div>
+        );
+    }
+    public tooltipClinicalContent(data: string[]) {
+        return (
+            <div className={styles.tooltip}>
+                <ul className={styles.alterationUl}>
+                    {data.map((cancerType: string) => (
+                        <li>{cancerType}</li>
+                    ))}
+                </ul>
+            </div>
+        );
+    }
+
+    public mainContent(content: string) {
+        return (
+            <span>
+                <a>{content}</a>
+            </span>
+        );
+    }
+
+    public getHugoSymbolName(matches: IGenomicGroupMatch[], threshold: number) {
+        const hugoSymbolSet = new Set([...matches].map(x => x.genomicAlteration.split(' ')[0].slice(1)));
+        if (hugoSymbolSet.size <= threshold) {
+            return [...hugoSymbolSet].join(', ');
+        }
+        return `${hugoSymbolSet.size} genes`;
+    }
+
+    render() {
+        return (
+            <div>
+                <TrialMatchTableComponent
+                    data={this.props.detailedTrialMatches}
+                    columns={this._columns}
+                    showCopyDownload={false}
+                />
+                <div className={styles.powered}>
+                    Powered by <a href="https://matchminer.org/" target="_blank">MatchMiner</a>
+                </div>
+            </div>
+        )
+    }
+}
--- a/src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
+++ b/src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
@@ -1,0 +1,153 @@
+import {
+    IArmMatch, IClinicalGroupMatch, IDetailedTrialMatch, IGenomicGroupMatch, IGenomicMatch, ITrial,
+    ITrialMatch, IArm, IDrug
+} from "../../../shared/model/MatchMiner";
+import * as _ from 'lodash';
+
+export function groupTrialMatchesById(trials: ITrial[], trialMatches:ITrialMatch[]): IDetailedTrialMatch[] {
+    trialMatches = excludeControlArms(trialMatches);
+    const matchesGroupedById = _.groupBy(trialMatches, (trial: ITrialMatch) => trial.id);
+    const matchedTrials: IDetailedTrialMatch[] = _.map(matchesGroupedById, (trialGroup, trialId) => {
+        const originalMatchedTrial: ITrial = _.find( trials, (trial) => trial.id === trialId )!;
+        const matchedTrial: IDetailedTrialMatch = {
+            id: originalMatchedTrial.id,
+            nctId: originalMatchedTrial.nctId,
+            protocolNo: originalMatchedTrial.protocolNo,
+            phase: originalMatchedTrial.phase,
+            shortTitle: originalMatchedTrial.shortTitle,
+            status: originalMatchedTrial.status,
+            matches: [],
+            priority: 0 // highest priority
+        };
+        matchedTrial.matches = groupTrialMatchesByArmDescription(trialGroup, originalMatchedTrial);
+        matchedTrial.priority = calculateTrialPriority(matchedTrial.matches);
+        return matchedTrial;
+    });
+    return _.sortBy(matchedTrials, (trial: IDetailedTrialMatch) => trial.priority);
+}
+
+export function groupTrialMatchesByArmDescription(trialGroup: ITrialMatch[], originalMatchedTrial: ITrial): IArmMatch[] {
+    const matchesGroupedByArm = _.groupBy(trialGroup, (trial: ITrialMatch) => trial.armDescription);
+    const matches = _.map(matchesGroupedByArm, (armGroup, armDescription) => {
+        const armMatch: IArmMatch = {
+            armDescription: armDescription,
+            drugs: [],
+            matches: []
+        };
+        if (!_.isUndefined(originalMatchedTrial.treatmentList.step[0].arm)) {
+            armMatch.drugs = getDrugsFromArm(armDescription, originalMatchedTrial.treatmentList.step[0].arm);
+        }
+        armMatch.matches = groupTrialMatchesByAgeNumerical(armGroup);
+        return armMatch;
+    });
+    return matches;
+}
+
+export function groupTrialMatchesByAgeNumerical(armGroup: ITrialMatch[]): IClinicalGroupMatch[] {
+    const matchesGroupedByAge = _.groupBy(armGroup, (trial: ITrialMatch) => trial.trialAgeNumerical);
+    const matches = _.map(matchesGroupedByAge, (ageGroup, age) => {
+        const cancerTypes =  _.uniq(_.map(ageGroup, (trial: ITrialMatch) => trial.trialOncotreePrimaryDiagnosis));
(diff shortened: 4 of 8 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -805,4 +809,56 @@ export class PatientViewPageStore {
...
+    }, []);
+
+    readonly trialIds = remoteData<ITrialQuery>({
+        await: () => [
+            this.trialMatches
+        ],
+        invoke: async() => {
+            let nctIds = new Set<string>();
+            let protocolNos = new Set<string>();
+            _.forEach(this.trialMatches.result, (trialMatch: ITrialMatch) => {
```
alisman: maybe have a comment that explains this.  Might be clearer if I knew what "nct" was

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -805,4 +809,56 @@ export class PatientViewPageStore {
...
+        protocol_no: []
+    });
+
+    readonly matchMinerTrials = remoteData<ITrial[]>({
+        await: () => [
+            this.trialIds
+        ],
+        invoke: async () => {
+            if (this.trialIds.result.protocol_no.length > 0 || this.trialIds.result.nct_id.length > 0) {
+                return postMatchMinerTrialsById(this.trialIds.result);
```
alisman: this isn't a good name for what I think is a fetch function (gets trials).  Follow naming scheme of existing clients.   fetchSomethingUsingPost

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -805,4 +809,56 @@ export class PatientViewPageStore {
...
+                nct_id: [...nctIds],
+                protocol_no: [...protocolNos]
+            };
+        },
+        onError: (err: Error) => {
+            // fail silently
+        }
+    }, {
+        nct_id: [],
+        protocol_no: []
```
alisman: does "protocol_no" mean protocolor_number?   if so, lets spell it out: protocol_number or protocol_id

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+    sampleManager?: SampleManager | null;
+    trials: Array<ITrial>;
+    trialMatches:Array<ITrialMatch>;
+    containerWidth: number;
+}
+
+enum ColumnKey {
+    ID = 'ID',
+    TITLE = 'Title',
+    MATCHINGCRITERIA = 'Matching Criteria'
```
alisman: MATCHING_CRITERIA

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                    ))}
+                </ul>
+            </div>
+        );
+    }
+
+    public mainContent(content: string) {
+        return (
+            <span>
+                <a>{content}</a>
```
alisman: how come wrapped in <a> tag if not clickeable?

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+            </span>
+        );
+    }
+
+    public getHugoSymbolName(matches: Array<IGenomicGroupMatch>, threshold: number) {
+        const hugoSymbolSet = new Set([...matches].map(x => x.genomicAlteration.split(' ')[0].slice(1)));
```
alisman: i think the only reason to use Set here is if this is meant to be a unique collection.  Set will automatically enforce uniqueness.  Is that required, if so, we should use Set.  If not, use Array.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+        if (hugoSymbolSet.size <= threshold) {
+            return [...hugoSymbolSet].join(', ');
+        }
+        return `${hugoSymbolSet.size} genes`;
+    }
+
+    public buildDetailedTrialMatches() {
+        const groupByTrialId = _.groupBy(this.props.trialMatches, 'nctId');
```
alisman: can we rename this trialsGroupedById?

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+        let matchedTrials: Array<IDetailedTrialMatch> = [];
+        _.forEach(groupByTrialId, (trialGroup, trialId) => {
+            let matchedTrial:any = _.find( this.props.trials, { 'nctId': trialId } );
+            matchedTrial['priority'] = 0; // highest priority
```
alisman: the matchedTrial is, i think, api data from store.  here we are mutating it, which is something we need to avoid doing unless absolutely necessary.  If possible we should have a wrapper object that has a reference to the matchedTrial data and then stores the meta data we are introducing here  (e.g. priority).   Also, since this data is a computation of the property data, the method (buildDetailedTrialMatches) should be @computed getter so that it only recomputes if underlying data has changed

alisman: you might consider putting this whole thing in the store.  and i think there's too much logic in this method.  too many nested forEach.  need to break this down into testable pure functions that are called in forEach

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+            matchedTrial['matches'] = [];
+            const groupByArm = _.groupBy(trialGroup, 'armDescription');
+            _.forEach(groupByArm, (armGroup, armDescription) => {
+                let armMatch: IArmMatch = {
+                    armDescription: armDescription,
+                    drugs: [],
+                    matches: []
+                };
+                const isControlArm = getDrugsAndExcludeControlArms(armDescription, matchedTrial['treatmentList']['step'][0]['arm'], armMatch);
```
alisman: instead of matchedTrial['treatmentList'] (which i don't think is typesafe), use dot notation matchedTrial.treatmentList so typescript will check it

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                if (isControlArm) return; // skip control arm
+
+                const groupByAge = _.groupBy(armGroup, 'trialAgeNumerical');
```
alisman: we never use the lodash string shortcuts b/c they are not typesafe.  this should be _.groupBy(armGroup,(trial)=>trial.trialAgeNumerical). 
Otherwise, if that property were to change in the future, typescript would not catch it.   I see a lot of this elsewhere in PR so please review and find em.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                _.forEach(groupByAge, (ageGroup, age) => {
+                    const cancerTypes =  _.uniq(_.map(ageGroup, 'trialOncotreePrimaryDiagnosis'));
```
alisman: Here again (underscore property name shortcut)

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                    let generalCancerTypes: Array<string> = [];
+                    let notCancerTypes: Array<string> = [];
+                    _.map(cancerTypes, (item) => {
+                       if (item.indexOf('!') > -1) {
```
alisman: use item.includes instead of indexOf.  it's boolean.

alisman: this "!" thing definately needs a comment.  @zhx828 seems a little hacky?

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                    let clinicalGroupMatch: IClinicalGroupMatch = {
+                        trialAgeNumerical: age,
+                        trialOncotreePrimaryDiagnosis: {
+                            general: generalCancerTypes,
+                            not: notCancerTypes
+                        },
+                        matches: [],
+                        notMatches: []
+                    };
+                    const groupByGenomicAlteration = _.groupBy(groupByAge[age], 'genomicAlteration');
```
alisman: shortcut

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                            genomicAlteration: genomicAlteration,
+                            matches: []
+                        };
+                        _.forEach(groupByPatientGenomic, (patientGenomicGroup) => {
+                            let genomicMatch: IGenomicMatch = {
+                                trueHugoSymbol: patientGenomicGroup[0].trueHugoSymbol,
+                                trueProteinChange: patientGenomicGroup[0].trueProteinChange,
+                                sampleIds: []
+                            };
+                            const sampleIds = _.uniq(_.map(patientGenomicGroup, 'sampleId'));
```
alisman: shortcut

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                            if (sampleIds.length > 1 && this.props.sampleManager) {
```
alisman: per @zhx828 comment, is there a reason we wouldn't have sampleManager?

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                    });
+                    matchedTrial['priority'] += setTrialMatchPriority(clinicalGroupMatch);
+                    armMatch.matches.push(clinicalGroupMatch);
+                });
+                matchedTrial['matches'].push(armMatch);
+            });
+            matchedTrials.push(matchedTrial);
+        });
+
+        return _.sortBy(matchedTrials, 'priority');
```
alisman: shortcut

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    render() {
+        const detailedTrialMatches: Array<IDetailedTrialMatch> = this.buildDetailedTrialMatches();
```
alisman: use IDetailedTrialMatch[] instead of Array<...>
You probably don't need to put type here as it should be inferred by typescript from return type of function..

src/pages/patientView/trialMatch/TrialMatchTableUtils.spec.tsx
```diff
@@ -0,0 +1,294 @@
...
+                                "sampleIds": [ "P-0000628-T02-IM5" ]
+                            }
+                        ]
+                    }
+                ],
+                "notMatches": []
+            },
+            expected: 0
+        } ];
+        _.forEach(data, (item) => {
```
alisman: probably better to avoid include the expected value in data structure above.  if there were a lot of assertions, maybe, but for only three, not important.  you could also then put a failure message as third param of assert, which gives better indication of the failure.

src/pages/patientView/trialMatch/TrialMatchTableUtils.spec.tsx
```diff
@@ -0,0 +1,294 @@
+import React from 'react';
```
alisman: spec files should be spec.ts (instead of tsx) if they do not use any JSX markup (e.g. <div>whatever</div>).

src/pages/patientView/trialMatch/TrialMatchTableUtils.tsx
```diff
@@ -0,0 +1,27 @@
+import { IArmMatch, IClinicalGroupMatch } from "../../../shared/model/MatchMiner";
```
alisman: this file can be .ts file because it doesn't involve any jsx.   There is a small build performance hit for .tsx files because the tsx compiler needs to search if for markup.  By giving it .ts extension, we save it from having to do that.

src/shared/api/MatchMinerAPI.ts
```diff
@@ -0,0 +1,86 @@
+import * as request from 'superagent';
+import { ITrial, ITrialMatch } from "shared/model/MatchMiner.ts";
+import { buildCBioPortalAPIUrl } from "./urls";
+
+/**
+ * Retrieves the trial matches for the query given, if they are in the MatchMiner API.
+ */
+
+const cbioportalUrl = buildCBioPortalAPIUrl('api-legacy/proxy/matchminer/api');
+export async function postMatchMinerTrialMatches(query: object): Promise<Array<ITrialMatch>> {
```
alisman: per comment above, lets rename this following pattern (fetchTrialMatchesUsingPOST)

src/shared/api/MatchMinerAPI.ts
```diff
@@ -0,0 +1,86 @@
...
+    return request.post(cbioportalUrl + '/post_trial_match')
+    .set('Content-Type', 'application/json')
+    .send(query)
+    .then((res) => {
+        const response = JSON.parse(JSON.parse(res.text));
```
alisman: why do we parse it twice?

alisman: @zhx828 the fact that we have to parse it all could be some indication of funkiness further up stream.   think superagent should parse it automatically if it detects the right content type.   not super important, but worth checking

alisman: maybe content type gets screwed upsince we're proxying

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+                                <span>{clinicalGroupMatch.trialOncotreePrimaryDiagnosis.not.join(', ').replace(/,(?!.*,)/gmi, ' and')}</span>
+                            ) : (
+                                <DefaultTooltip
+                                    placement='bottomLeft'
+                                    trigger={['hover', 'focus']}
+                                    overlay={this.tooltipClinicalContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.not)}
+                                    destroyTooltipOnHide={false}
+                                    onPopupAlign={placeArrowBottomLeft}
+                                >
+                                    {this.mainContent(`${clinicalGroupMatch.trialOncotreePrimaryDiagnosis.not.length} cancer types`)}
```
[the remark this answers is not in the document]
victoria34: changed to {this.mainContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.not.length + ` cancer types`)}

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+            return [...hugoSymbolSet].join(', ');
+        }
+        return `${hugoSymbolSet.size} genes`;
+    }
+
+    public buildDetailedTrialMatches() {
+        const groupByTrialId = _.groupBy(this.props.trialMatches, 'nctId');
+        let matchedTrials: Array<IDetailedTrialMatch> = [];
+        _.forEach(groupByTrialId, (trialGroup, trialId) => {
+            let matchedTrial:any = _.find( this.props.trials, { 'nctId': trialId } );
```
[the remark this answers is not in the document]
victoria34: You know the `match` structure of the trial is very flexible and nested, so I use `any` at here instead of modeling it.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,321 @@
...
+} from "../../../shared/model/MatchMiner";
+import styles from './style/trialMatch.module.scss';
+import { computed } from "mobx";
+import LazyMobXTable from "../../../shared/components/lazyMobXTable/LazyMobXTable";
+import SampleManager from "../sampleManager";
+import DefaultTooltip, { placeArrowBottomLeft } from "../../../public-lib/components/defaultTooltip/DefaultTooltip";
+import { getDrugsAndExcludeControlArms, setTrialMatchPriority } from "./TrialMatchTableUtils";
+
+export type ITrialMatchProps = {
+    sampleManager?: SampleManager | null;
```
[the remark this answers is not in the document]
victoria34: Removed `?`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
victoria34:
```
let nctIds = new Set<string>(); // Trial unique id from clinicaltrials.gov
let protocolNos = new Set<string>(); // trials's MSK ID
```

src/pages/patientView/trialMatch/TrialMatchTable.tsx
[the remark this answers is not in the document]
victoria34: I remodeled the ITrial for the whole trial structure including match clauses.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: changed

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
victoria34: changed

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Yes, I want to have an unique collection of hugo_symbols so I use Set at here.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: It's not clickable but it can be hovered over. I want to use the <a> tag's style to reminder user that this is a link please move cursor on it.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
victoria34: Sorry, I cannot change the protocol_no in here. This object will be sent to api for querying trials. Besides, protocol_no is defined in Clinical Trial Markup Language so I use the same name in frontend.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to 
```
const isControlArm = getDrugsAndExcludeControlArms(armDescription, originalMatchedTrial.treatmentList.step[0].arm, armMatch);
```

src/pages/patientView/trialMatch/TrialMatchTableUtils.tsx
victoria34: Thanks for pointing it out. Will change to ts.

src/pages/patientView/trialMatch/TrialMatchTableUtils.spec.tsx
victoria34: Thanks for pointing it out. Will change to ts.

src/shared/api/MatchMinerAPI.ts
victoria34: Renamed all apt methods.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed variables' name following this pattern.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to (trial)=>trial.key

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to `(trial: ITrialMatch) => trial.trialOncotreePrimaryDiagnosis`.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to includes and left comment to explain the meaning of "!".

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to `const matchesGroupedByGenomicAlteration = _.groupBy(matchesGroupedByAge[age], (trial: ITrialMatch) => trial.genomicAlteration);`.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to `const sampleIds = _.uniq(_.map(patientGenomicGroup, (trial: ITrialMatch) => trial.sampleId));`.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Removed the check on sampleManager.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to `return _.sortBy(matchedTrials, (trial: IDetailedTrialMatch) => trial.priority);`.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to []

src/pages/patientView/trialMatch/TrialMatchTableUtils.spec.tsx
victoria34: Stored expect value in another array.

src/shared/api/MatchMinerAPI.ts
victoria34: This is proxy issue and I have fixed it. Now it only need to be parse once.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -538,6 +544,15 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                     </MSKTab>
                     )}
 
+                    {(getBrowserWindow().localStorage.trialmatch === 'true') && (
```
alisman: you could put the check for localStorage up in shouldShowTrialMatch.  i realize it's there for testing purposes.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -807,4 +811,71 @@ export class PatientViewPageStore {
         this.ajaxErrors = [];
     }
 
+    readonly trialMatches = remoteData<ITrialMatch[]>({
+        invoke: () => {
+            return fetchTrialMatchesUsingPOST({mrn: this.patientId});
+        },
+        onError: (err: Error) => {
```
alisman: lets remote these onError handlers.  We probably don't want it to fail silently

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
alisman: ok

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,236 @@
...
+
+class TrialMatchTableComponent extends LazyMobXTable<IDetailedTrialMatch> {
+
+}
+
+@observer
+export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
+
+    @computed
+    get columnsWidth() {
```
alisman: this is very minor, but it should be "columnWidths", the plural being the widths.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,236 @@
...
+                        <span>
+                            <b> except in </b>
+                            { clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.length < 4 ? (
+                                <span>{clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.join(', ').replace(/,(?!.*,)/gmi, ' and')}</span>
+                            ) : (
+                                <DefaultTooltip
+                                    placement='bottomLeft'
+                                    trigger={['hover', 'focus']}
+                                    overlay={this.tooltipClinicalContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative)}
+                                    destroyTooltipOnHide={false}
```
alisman: should probably be destroyTooltipOnHide=trye.  we don't want to keep lots of tooltip HTML in dom.  better just to regenerate it.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
alisman: but what happens when someone hovers over it?  i mean, i don't see any behavior attached to it

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,236 @@
...
+                        </span>
+                    }
+                </span>
+            </span>
+        );
+    }
+
+    public getGenomicMatch(clinicalGroupMatch: IClinicalGroupMatch) {
+        return (
+            <span className={styles.firstLeft}>
```
alisman: insetad of attaching classes, you could probably use css selectors first-child, last-child

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,236 @@
...
+            <div>
+                {trial.matches.map((armMatch: IArmMatch, index: number) => (
+                    <div>
+                        <div>
+                            <div>
+                                {armMatch.matches.map((clinicalGroupMatch: IClinicalGroupMatch, cgIndex:number) => (
+                                    <span className={styles.criteriaContainer}>
+                                        {this.getGenomicMatch(clinicalGroupMatch)}
+                                        {this.getClinicalMatch(clinicalGroupMatch)}
+                                        <If condition={cgIndex < armMatch.matches.length - 1}><hr className={styles.criteriaHr}/></If>
```
alisman: you could hide the last <hr> with css by using hr:last-child css selector

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+    IArmMatch, IClinicalGroupMatch, IDetailedTrialMatch, IGenomicGroupMatch, IGenomicMatch, ITrial,
+    ITrialMatch, IArm, IDrug
+} from "../../../shared/model/MatchMiner";
+import * as _ from 'lodash';
+
+export function groupTrialMatchesById(trials: ITrial[], trialMatches:ITrialMatch[]): IDetailedTrialMatch[] {
+    const matchesGroupedById = _.groupBy(trialMatches, (trial: ITrialMatch) => trial.id);
+    let matchedTrials: IDetailedTrialMatch[] = [];
+    _.forEach(matchesGroupedById, (trialGroup, trialId) => {
+        const originalMatchedTrial: ITrial = _.find( trials, { 'id': trialId } )!;
```
alisman: should be `_.find(trials, (trial)=>trial.id===trialId)`
It's typesafe that way.   We've decided to never use the lodash shortcuts.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
+import {
+    IArmMatch, IClinicalGroupMatch, IDetailedTrialMatch, IGenomicGroupMatch, IGenomicMatch, ITrial,
+    ITrialMatch, IArm, IDrug
+} from "../../../shared/model/MatchMiner";
+import * as _ from 'lodash';
+
+export function groupTrialMatchesById(trials: ITrial[], trialMatches:ITrialMatch[]): IDetailedTrialMatch[] {
+    const matchesGroupedById = _.groupBy(trialMatches, (trial: ITrialMatch) => trial.id);
+    let matchedTrials: IDetailedTrialMatch[] = [];
+    _.forEach(matchesGroupedById, (trialGroup, trialId) => {
```
alisman: you could probably do matchedTrials = _.map(matchesGroupedById).  it's better style to use direct assignment.  no big deal

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+            matches: [],
+            priority: 0 // highest priority
+        };
+        groupTrialMatchesByArmDescription(trialGroup, matchedTrial, originalMatchedTrial);
+        matchedTrials.push(matchedTrial);
+    });
+    return _.sortBy(matchedTrials, (trial: IDetailedTrialMatch) => trial.priority);
+}
+
+export function groupTrialMatchesByArmDescription(trialGroup: ITrialMatch[], matchedTrial: IDetailedTrialMatch, originalMatchedTrial: ITrial) {
```
alisman: functions should always return something, even if it's just mutating data passed in.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+    });
+}
+
+export function groupTrialMatchesByPatientGenomic(genomicAlterationGroup: ITrialMatch[], genomicAlteration: string): IGenomicGroupMatch {
+    const matchesGroupedByPatientGenomic = _.groupBy(genomicAlterationGroup, (trial: ITrialMatch) => trial.patientGenomic);
+    const genomicGroupMatch: IGenomicGroupMatch = {
+        genomicAlteration: genomicAlteration,
+        matches: []
+    };
+    _.forEach(matchesGroupedByPatientGenomic , (patientGenomicGroup: ITrialMatch[]) => {
```
alisman: this  could be genomicGroupMatch.matches = _.map()

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+            trueProteinChange: patientGenomicGroup[0].trueProteinChange,
+            sampleIds: _.uniq(_.map(patientGenomicGroup, (trial: ITrialMatch) => trial.sampleId))
+        };
+        genomicGroupMatch.matches.push(genomicMatch);
+    });
+    return genomicGroupMatch;
+}
+
+export function setTrialMatchPriority(clinicalGroupMatch: IClinicalGroupMatch): number {
+    if (clinicalGroupMatch.notMatches.length > 0) {
```
alisman: would be nice to have a comment explaining this business logic here

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+        }
+        return 1;
+    }
+    return 0;
+}
+
+export function getDrugsAndExcludeControlArms(armDescription: string, arms: IArm[], armMatch: IArmMatch): boolean {
+    const hiddenArmTypes = ['Control Arm', 'Placebo Arm'];
+    if ( armDescription !== '' ) { // match for specific arm
+        const matchedArm: IArm = _.find( arms, { 'arm_description': armDescription } )!;
```
alisman: shouldn't use lodash shortcut (not typesafe)

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,129 @@
...
+    if (clinicalGroupMatch.notMatches.length > 0) {
+        if (clinicalGroupMatch.matches.length === 0) {
+            return 2;
+        }
+        return 1;
+    }
+    return 0;
+}
+
+export function getDrugsAndExcludeControlArms(armDescription: string, arms: IArm[], armMatch: IArmMatch): boolean {
```
alisman: this function is confusing because it returns a boolean value and also mutates data passed into it.  in general it is not a good idea for a function to directly mutate the data passed into it.  it should return the data which is necessary for the mutation and controlling code should actually mutate.  the boolean nature can be accomplished by returning a zero length array

src/shared/model/MatchMiner.ts
```diff
@@ -0,0 +1,117 @@
...
+}
+
+export interface IGenomic {
+    hugo_symbol?: string,
+    annotated_variant?: string,
+}
+
+export interface ITrialMatch {
+    id: string;
+    nctId: string | '';
```
alisman: do you need the `| ''` ?

src/pages/patientView/PatientViewPage.tsx
victoria34: Moved to shouldShowTrialMatch.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
victoria34: Remove all onError handlers.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Updated

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to true

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Updated

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Changed to (arm) => arm.description === armDescription

src/shared/model/MatchMiner.ts
victoria34: This field will be showed on the page. If it's null, I need to do checks which is not convenient so I set the default value to `''`.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: This is not for first child. This is for flex-box tricks. Maybe the name I assigned is not so clear. https://css-tricks.com/snippets/css/a-guide-to-flexbox/

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: I tried but it doesn't work at here. The `:last-child` selector matches every element that is the last child of its parent. Each `<hr>` is wrapped by `<div>`, so it's always the last child of its parent `<div>`.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34:
```
genomicGroupMatch.matches = _.map(matchesGroupedByPatientGenomic , (patientGenomicGroup: ITrialMatch[]) => {
        const genomicMatch: IGenomicMatch = {
            trueHugoSymbol: patientGenomicGroup[0].trueHugoSymbol,
            trueProteinChange: patientGenomicGroup[0].trueProteinChange,
            sampleIds: _.uniq(_.map(patientGenomicGroup, (trial: ITrialMatch) => trial.sampleId))
        };
        return genomicMatch;
    });
```

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: updated

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Add comments.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Added `return` statement for all functions.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Separate it to excludeControlArms() and getDrugsFromArm().

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+                                <span>{clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.join(', ').replace(/,(?!.*,)/gmi, ' and')}</span>
+                            ) : (
+                                <DefaultTooltip
+                                    placement='bottomLeft'
+                                    trigger={['hover', 'focus']}
+                                    overlay={this.tooltipClinicalContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative)}
+                                    destroyTooltipOnHide={true}
+                                    onPopupAlign={placeArrowBottomLeft}
+                                >
+                                    {this.mainContent(clinicalGroupMatch.trialOncotreePrimaryDiagnosis.negative.length + ` cancer types`)}
```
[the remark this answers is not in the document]
victoria34: Updated

src/shared/model/MatchMiner.ts
```diff
@@ -0,0 +1,96 @@
...
+    };
+}
+
+export interface IStep {
+    match?: object[];
+    arm?: IArm[];
+}
+
+export interface IArm {
+    arm_description: string | ''; // Arm full name.
```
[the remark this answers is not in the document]
victoria34: Yes, arm_description can be null sometimes.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -0,0 +1,234 @@
...
+                    <div>
+                        {(genomicGroupMatch.matches.length === 1 && genomicGroupMatch.genomicAlteration === `${genomicGroupMatch.matches[0].trueHugoSymbol} ${genomicGroupMatch.matches[0].trueProteinChange}`) ? (
+                        <span>{`${genomicGroupMatch.genomicAlteration} `}
+                            {genomicGroupMatch.matches[0].sampleIds.map((sampleId: string) => (
+                                <span className={styles.genomicSpan}>
+                                    {this.props.sampleManager!.getComponentForSample(sampleId, 1, '')}
+                                </span>
+                            ))}
+                        </span>
+                        ) : (
```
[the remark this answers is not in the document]
victoria34: Switched to <If> and created getGenomicExactMatch() and getGenomicVariantCategoryMatch() to wrap them.

jjgao:
@victoria34 could you change the tab title to "Matched Trials" and add a note "Powered by [MatchMiner](https://matchminer.org)" similar to the PathwayMapper implementation: http://bit.ly/2LLeHon

image

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+            trueHugoSymbol: patientGenomicGroup[0].trueHugoSymbol!,
+            trueProteinChange: patientGenomicGroup[0].trueProteinChange!,
+            sampleIds: _.uniq(_.map(patientGenomicGroup, (trial: ITrialMatch) => trial.sampleId))
+        };
+        return genomicMatch;
+    });
+    return genomicGroupMatch;
+}
+
+export function setTrialMatchPriority(clinicalGroupMatch: IClinicalGroupMatch): number {
```
alisman: calculateTrialMatchPriority

victoria34:
@jjgao I added the `Powered by` at the right bottom corner. Is this fine?

image

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+    return clinicalGroupMatch;
+}
+
+export function groupTrialMatchesByPatientGenomic(genomicAlterationGroup: ITrialMatch[], genomicAlteration: string): IGenomicGroupMatch {
+    const matchesGroupedByPatientGenomic = _.groupBy(genomicAlterationGroup, (trial: ITrialMatch) => trial.patientGenomic);
+    const genomicGroupMatch: IGenomicGroupMatch = {
+        genomicAlteration: genomicAlteration,
+        matches: []
+    };
+    genomicGroupMatch.matches = _.map(matchesGroupedByPatientGenomic , (patientGenomicGroup: ITrialMatch[]) => {
```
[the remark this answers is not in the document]
victoria34: Updated to `const matches = `

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+}
+
+export function groupTrialMatchesByGenomicAlteration(ageGroup: ITrialMatch[], clinicalGroupMatch: IClinicalGroupMatch) {
+    const matchesGroupedByGenomicAlteration = _.groupBy(ageGroup, (trial: ITrialMatch) => trial.genomicAlteration);
+    _.forEach(matchesGroupedByGenomicAlteration, (genomicAlterationGroup, genomicAlteration) => {
+        const genomicGroupMatch = groupTrialMatchesByPatientGenomic(genomicAlterationGroup, genomicAlteration);
+        // If a genomic alteration(i.e. BRAF V600E) contains a "!",
+        // it means this trial cannot be used for the genomic alteration, which is a "NOT" match.
+        if(genomicAlteration.includes('!')) {
+            clinicalGroupMatch.notMatches.push(genomicGroupMatch);
```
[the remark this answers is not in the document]
victoria34: Updated to `return { matches: ..., notMatches: ... };`

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+            trialAgeNumerical: age,
+            trialOncotreePrimaryDiagnosis: {
+                positive: positiveCancerTypes,
+                negative: negativeCancerTypes
+            },
+            matches: [],
+            notMatches: []
+        };
+        clinicalGroupMatch = groupTrialMatchesByGenomicAlteration(matchesGroupedByAge[age], clinicalGroupMatch);
+        matchedTrial.priority += setTrialMatchPriority(clinicalGroupMatch);
```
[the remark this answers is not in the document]
victoria34: Created calculateTrialPriority()

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -0,0 +1,137 @@
...
+            armDescription: armDescription,
+            drugs: [],
+            matches: []
+        };
+        if (!_.isUndefined(originalMatchedTrial.treatmentList.step[0].arm)) {
+            armMatch.drugs = getDrugsFromArm(armDescription, originalMatchedTrial.treatmentList.step[0].arm);
+        }
+        return groupTrialMatchesByAgeNumerical(armGroup, armMatch, matchedTrial);
+    });
+    return matchedTrial;
```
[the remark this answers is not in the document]
victoria34: Updated as you suggested

PR #2535, opened 2019-07-10
Fix oncoprint bug with sort by case list order

adamabeshouse:
Bug happens when more than one sample per patient


jjgao:
Thanks! I am wondering if we should handle this on the page level, ie. remove redundant samples from the query?

adamabeshouse:
no, that's not the problem - it's just a bug with how the oncoprint calculates the patient order from the sample order

PR #2536, opened 2019-07-11
Removed external dependencies from public lib

onursumer:
# What? Why?
Fix cBioPortal/cbioportal/issues/6376.

# Checks
  the `-s` option)

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -127,7 +127,6 @@ var config = {
             styles: join(src, 'styles'),
             pages: join(src, 'pages'),
             shared: join(src, 'shared'),
-            commons: join(src, 'commons'),
             appConfig: path.join(__dirname + '/src', 'config', ((process.env.NODE_ENV === 'test')? 'test.' : '') + 'config')
         }
     },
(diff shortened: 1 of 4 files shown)


onursumer:
@inodb  @alisman  I think we should make `yarn buildPublicLib` part of CI so that we won't merge breaking changes in the future.

onursumer:
We decided to hardcode the font URL to avoid this confusing workaround

inodb:
@onursumer 	could you maybe at it to CI as part of this PR? It can be added as a new job here: https://github.com/cBioPortal/cbioportal-frontend/blob/master/.circleci/config.yml#L65. Maybe call it something like `build_public_lib_test` or something?

onursumer:
@alisman I've simplified the PR by making the font url hardcoded.

@ino I've added the new `public_lib` test into circle CI config. It is not marked as `Required` though, I think we need to change that on GitHub settings after merging this PR, right?

webpack.config.js
```diff
@@ -127,7 +127,6 @@ var config = {
             styles: join(src, 'styles'),
             pages: join(src, 'pages'),
             shared: join(src, 'shared'),
-            commons: join(src, 'commons'),
```
inodb: what's this?

PR #2537, opened 2019-07-11
Skip group comparison tests that improperly import ts modules

alisman:
Cannot import ts modules into e2e specs


alisman:
@adamabeshouse at some point we should set up ts compilation for e2e tests.  maybe when finally go about upgrading webdriver.io (i think gonna be painful).  for now, lets just hardcode those values

alisman:
@inodb this was the only way i could get tests to run locally.  13 tries to install chromedriver that is only  compatible with chrome 76, which is still in beta.  test then won't run at all.

inodb:
@alisman On CircleCI we run update and start with specific version of the chrome driver: https://github.com/cBioPortal/cbioportal-frontend/blob/master/.circleci/config.yml#L25-L26

PR #2542, opened 2019-07-12
Upgrade to react-mutation-mapper 0.2.0

onursumer:
# What? Why?
Upgraded to the `react-mutation-mapper` release 0.2.0, and cleaned up components moved to `react-mutation-mapper`

- `MutationMapper` is now extending default mutation mapper from `react-mutation-mapper`
- Partially replaced annotation column components (`OncoKB` and `CancerHotspots`) with the components from `react-mutation-mapper` (Cannot completely replace them because table implementations are different)
- Moved more stuff into `public-lib` and released version `0.0.13`

# Checks
  the `-s` option)


src/public-lib/components/WindowWrapper.ts
```diff
@@ -0,0 +1,33 @@
+import autobind from "autobind-decorator";
```
onursumer: Moved from `shared/components/window/WindowStore.ts` (to support responsive lollipop plot in `react-mutation-mapper`)

src/public-lib/components/SVGAxis.tsx
```diff
@@ -14,34 +14,38 @@ type SVGAxisProps = {
     rangeLower:number;
     rangeUpper:number;
     vertical?:boolean;
+    reverse?:boolean;
```
onursumer: this is to support revert axis values from (0..max) to (max..0)

src/public-lib/components/SVGAxis.tsx
```diff
@@ -14,34 +14,38 @@ type SVGAxisProps = {
...
+    invertTicks?:boolean;
```
onursumer: This means a visual inversion of the ticks (i.e. mirror image)

src/shared/components/annotation/Civic.spec.tsx
```diff
@@ -22,16 +21,10 @@ describe('Civic with no data', () => {
     });
 
     it('displays a load spinner when there is no civic data', () => {
-        const spinner = component.find(Circle);
+        const spinner = component.find(".fa-spinner");
```
onursumer: `react-mutation-mapper` is now using a simple `fa-spinner` instead of a standalone `Circle` component

src/shared/components/annotation/oncokb/OncoKB.spec.tsx
```diff
@@ -1,13 +1,15 @@
+// TODO move this test into react-mutation-mapper repo!
```
onursumer: Leaving this for the next PR

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -52,8 +44,6 @@ export interface IMutationMapperStoreConfig {
 
 export default class MutationMapperStore extends DefaultMutationMapperStore
 {
-    @observable public activeTranscript: string | undefined = undefined;
```
onursumer: Most of the logic implemented here is now move to the default mutation mapper store in the `react-mutation-mapper`

src/shared/components/window/WindowStore.ts
```diff
@@ -1,34 +1,3 @@
...
-
-    private handleWindowResize = _.debounce(this.setWindowSize, 200);
-    private windowObj: any;
-
-    @observable size: WindowSize
-}
-
-export default new WindowStore();
\ No newline at end of file
+export default new WindowWrapper();
```
onursumer: separated out the `WindowWrapper` component, but we still use a `singleton` instance for portal frontend, so still keeping the `WindowStore` component here

src/shared/lib/OncoKbUtils.ts
```diff
@@ -1,60 +1,15 @@
+// TODO remove this file and completely rely on react-mutation-mapper implementation?
```
onursumer: Leaving this for another PR, since the portal cache implementation still needs these functions

PR #2543, opened 2019-07-12
Merge master into release-3.1.0

adamabeshouse:


adamabeshouse:
@inodb no nothing. there was a conflict in the README but I think it was just two separately added sections so I added them both.

inodb:
seems like the localdb test is already failing on release-3.1.0 so prolly unrelated

adamabeshouse:
broken e2e tests are already broken

PR #2544, opened 2019-07-12
Include clinical ID in the clinical attribute chart/column tooltip

zhx828:
Also do not show the description when the description is the same as dispaly name

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6354

Screen Shot 2019-07-12 at 11 17 54 AM
Screen Shot 2019-07-31 at 5 51 46 PM

### DIFF
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -26,6 +26,7 @@ import MobxPromise from 'mobxpromise';
 import {getTextWidth} from "../../public-lib/lib/TextTruncationUtils";
 import {CNA_COLOR_AMP, CNA_COLOR_HOMDEL, DEFAULT_NA_COLOR, getClinicalValueColor} from "shared/lib/Colors";
 import {StudyViewComparisonGroup} from "../groupComparison/GroupComparisonUtils";
+import styles from './styles.module.scss';
 
 
 // Cannot use ClinicalDataTypeEnum here for the strong type. The model in the type is not strongly typed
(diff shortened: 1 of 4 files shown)


alisman:
@zhx828 the style of the tooltips in screenshots above are not totally consistent.  Is that because of scaling of the screenshots?

zhx828:
@alisman I believe so. There is no special treatment there.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -208,10 +209,16 @@ const OPERATOR_MAP: {[op:string]: string} = {
...
 };
 
-export function getClinicalAttributeOverlay(displayName: string, description: string): JSX.Element {
-    return <div style={{maxWidth: '500px'}}>
-        <b>{displayName}</b><br/>
-        {description}
+export function getClinicalAttributeOverlay(displayName: string, description: string, clinicalAttributeId?: string): JSX.Element {
+    const comparisonDisplayName = displayName.toLowerCase().trim();
+    let comparisonDescription = description.toLowerCase().trim();
+    if (comparisonDescription.substring(comparisonDescription.length - 1) === '.') {
```
alisman: you could just do comparisonDescription.replace(/\.$/,"")   and you wouldn't have to bother checking for presence of dot at end.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -208,10 +209,16 @@ const OPERATOR_MAP: {[op:string]: string} = {
...
+        comparisonDescription = comparisonDescription.substring(0, comparisonDescription.length - 1);
+    }
+    return <div style={{maxWidth: '500px', display: 'flex', flexDirection: 'column'}}>
```
alisman: i'm not sure you need this flex stuff.  where you want it to break line, just use div and instead of span

alisman:
in the second screenshot, is the tooltip meant to point to info icon?  is it misaligned?

alisman:
also, seems like there should be a condition where we don't show the info button at all, no?

zhx828:
> also, seems like there should be a condition where we don't show the info button at all, no?

@alisman we actually have a condition in the ChartHeader. But I need to update that logic now.

zhx828:
> in the second screenshot, is the tooltip meant to point to info icon? is it misaligned?

@alisman I tried to adjust the tooltip placement and updated the screenshots. But it seems still not perfectly aligned.

PR #2548, opened 2019-07-16
Do not retrieve tags for virtual studies

oplantalech:
# What? Why?
Fix for cbioportal/cbioportal#6285. There is a bug when using virtual studies and hovering the info icon in the query page: the endpoint `/studies/{studyId}/tags` returns an error after attempting to retrieve the tags given the virtual study ID (instead of an expected non-virtual study ID). Since virtual studies can be made of multiple studies, I think the best way to deal with the error is not to retrieve tags for virtual studies. Feel free to comment if you think there is a better solution to deal with the issue.

### DIFF
--- a/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
+++ b/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
@@ -41,19 +41,16 @@ export type StudyTagsTooltipProps = {
     children: any;
 };
 
-export type BuildOverlayTooltipProps = {
+export type StudyInfoOverlayTooltipProps = {
     studyDescription: string;
     studyId: string;
     isVirtualStudy: boolean;
 };
 
 @observer
-default class BuildOverlay extends React.Component<BuildOverlayTooltipProps, {}> {
+default class StudyInfoOverlay extends React.Component<StudyInfoOverlayTooltipProps, {}> {
     @observable readonly studyMetadata = remoteData({
         invoke: async () => {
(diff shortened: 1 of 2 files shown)


src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -43,12 +44,16 @@ export type StudyTagsTooltipProps = {
...
     studyDescription: string;
     studyId: string;
+    isVirtualStudy: boolean;
 };
 
 @observer
 default class BuildOverlay extends React.Component<BuildOverlayTooltipProps, {}> {
     @observable readonly studyMetadata = remoteData({
         invoke: async () => {
+            if (this.props.isVirtualStudy) {
```
alisman: you should just deal with this condition down in the render method instead of muddying the promise with it.

src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -43,12 +44,16 @@ export type StudyTagsTooltipProps = {
 export type BuildOverlayTooltipProps = {
     studyDescription: string;
     studyId: string;
+    isVirtualStudy: boolean;
 };
 
 @observer
 default class BuildOverlay extends React.Component<BuildOverlayTooltipProps, {}> {
```
alisman: While we're at it, might as well change this to StudyInfoOverlay

PR #2552, opened 2019-07-16
Support summary/full data download for pie chart

zhx828:
There is one concern in this PR. The submenu of the download is not perfectly aligned on the left side. See the animation below. 
Since we do absolute positioning for submenu, then we need to calculate the left position based on the max width of the submenu item. Personally, I'm not a fan of doing that because you need to update the function when the style of the dropdown updated, for instance, padding/margin/border width. And also when I hover over the download option, my attention has been brought to the submenu, then the key here is actually how to quickly go to the submenu. we also need to consider if we calculate the width, then if the submenu is really wide, due to the items in the submenu are left aligned, then user has to move the mouse quite far away in order to download the data. But please let me know if you think I should try to align the submenu. @jjgao @schultzn @alisman 

Jul-16-2019 15-25-39

# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6357


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2782,6 +2783,25 @@ export class StudyViewPageStore {
         return survivalTypes;
     }
 
+    public async getPieChartDataDownload(chartMeta: ChartMeta, dataType?: DataType) {
+        const infoForSummaryDataDownload = this.isCustomChart(chartMeta.uniqueKey) ?
+            this.getCustomChartDataCount(chartMeta) : this.getClinicalDataCount(chartMeta);
+
+        return dataType && dataType === 'summary' ?
```
alisman: i think we should avoid nested ternary.  just too hard to read

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1238,15 +1238,19 @@ export function pickClinicalAttrFixedColors(data: ClinicalDataCount[]): {[attrib
...
 
-export type ClinicalDataCountWithColor = ClinicalDataCount & { color: string }
+export type ClinicalDataCountSummary = ClinicalDataCount & { color: string , percentage: number, freq: string}
 
-export function getClinicalDataCountWithColorByClinicalDataCount(counts:ClinicalDataCount[]):ClinicalDataCountWithColor[] {
+export function getClinicalDataCountWithColorByClinicalDataCount(counts:ClinicalDataCount[]):ClinicalDataCountSummary[] {
     counts.sort(clinicalDataCountComparator);
     const colors = pickClinicalDataColors(counts);
-    return counts.map(slice =>{
+    const sum = _.sum(counts.map(count => count.count));
```
alisman: _.sumBy

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -149,8 +150,8 @@ export class StudySummaryTab extends React.Component<IStudySummaryTabProps, {}>
                     props.onResetSelection = this.handlers.onValueSelection;
                 }
                 props.onChangeChartType = this.handlers.onChangeChartType;
-                props.getData = () => this.store.isCustomChart(chartMeta.uniqueKey) ? this.store.getCustomChartDownloadData(chartMeta) : this.store.getClinicalData(chartMeta);
-                props.downloadTypes = ["Data", "SVG", "PDF"];
+                props.getData = (dataType?: DataType) => this.store.getPieChartDataDownload(chartMeta, dataType);
+                props.downloadTypes = this.store.isCustomChart(chartMeta.uniqueKey) ? ["Summary Data", "SVG", "PDF"] : ["Summary Data", "Full Data", "SVG", "PDF"];
```
alisman: maybe a commend explaining why we don't do full data for custom?

alisman: this comment might belong somewhere else.  but would be nice to have somewhere

alisman:
@zhx828 does this need particular backend?

jjgao:
The arrow pointing left seems a little strange to me, but maybe it's just me... I also don't have a good solution for this. @alisman what do you think?

zhx828:
@alisman it does not.

src/pages/studyView/tabs/SummaryTab.tsx
zhx828: I agree. Actually after second thought, the custom chart should support  both downloads. Just updated the pr.

jjgao:
@zhx828 could you make the arror alway point to the right. I think that's more intuitive.

image

zhx828:
@jjgao updated

PR #2556, opened 2019-07-18
Fix show all samples button on patient view

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/6398


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -260,6 +260,20 @@ export class PatientViewPageStore {
         []
```
alisman: seems that this error was caused by change inside fetchSamplesForPatient.  code was forked so that if sampleId is present, it only fetches single sample.   We should inquire with @onursumer why that change was made.   Seems we might as well get all samples and for the patient and just snatch the one we need when we're in sample mode.

inodb: @alisman the order is slightly different compared to patient mode:

```
sampleId -> fetchSample -> derive patient id -> fetch all samples
vs
patientId -> fetch all samples
```

But yeah let's check what @onursumer says

onursumer: @alisman @inodb apparently I only moved the exact logic from `PatientViewStore` to `StoreUtils` with this commit: https://github.com/cBioPortal/cbioportal-frontend/commit/e5990026a64802816e2f072b3b9e1754c6d3b05c#diff-f69e4be5407424f0d3f365ca9b817cc5L193. So not sure if the change inside `fetchSamplesForPatient` is the culprit :)

alisman: @onursumer not sure i understand your comment.

PR #2560, opened 2019-07-19
Patient view mutations VAF oncoprint prototype

adamabeshouse:
https://cbioportal-frontend-pr-2560.herokuapp.com/patient?studyId=prad_fhcrc&caseId=05-144

there are issues with the oncoprint tooltip placement, I've logged it in another issue

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -47,6 +47,7 @@ import {QueryParams} from "url";
 import {AppStore} from "../../AppStore";
 import request from 'superagent';
 import {remoteData} from "../../public-lib/api/remoteData";
+import MutationOncoprint from "./oncoprint/MutationOncoprint";
 
 const patientViewPageStore = new PatientViewPageStore();
 
@@ -409,6 +410,8 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                             </div>
                                         )
                                     }
+                                    
+                                    <MutationOncoprint store={patientViewPageStore}/>
 
                                     <LoadingIndicator isLoading={patientViewPageStore.mutationData.isPending || patientViewPageStore.uncalledMutationData.isPending || patientViewPageStore.oncoKbAnnotatedGenes.isPending || patientViewPageStore.studyIdToStudy.isPending} />
 
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -47,7 +47,7 @@ import {
     fetchMutationData,
     fetchDiscreteCNAData,
     generateUniqueSampleKeyToTumorTypeMap,
-    findMutationMolecularProfileId,
+    findMutationMolecularProfile,
     findUncalledMutationMolecularProfileId,
     mergeMutationsIncludingUncalled,
     fetchGisticData,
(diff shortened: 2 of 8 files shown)


jjgao:
@adamabeshouse cool, that was quick! I am wondering if we can do a heatmap clustering?

adamabeshouse:
@jjgao :+1: updated

tmazor:
Cool!!

I think this is great! It's a really nice complementary visualization to the Allele Freq column in the table and the genomic overview plot.

For example, the evolution between samples for this patient is immediately super clear: 
image
https://cbioportal-frontend-pr-2560.herokuapp.com/patient?caseId=P17&studyId=lgg_ucsf_2014

I also have a bunch of very specific thoughts/requests, but since the PR is called "prototype", I presume you're just looking for general feedback at this point?

adamabeshouse:
@tmazor I don't know what exactly I'm looking for 😄 I'm not sure what the timeline is/where we're at right now as far as specifying the feature. that's up to @jjgao

jjgao:
@tmazor if the feedback is positive, we are pushing this for production. (based on your comments, it is like to be the case :) Please add all your thoughts/requests.

... loop in @cBioPortal/product

jjgao:
I also have a few requests, but I'll hold them off and let you (@tmazor) send yours first :)

adamabeshouse:
let 'er rip! :)

tmazor:
Ok! Here, in no particular order, is my first round of feedback:

1) When I hover over the plot, the tooltip should include the gene name & protein change and the VAF should be truncated to just a few digits

2) It would be great to be able to toggle between clustering the samples and having the samples in the same order as all the other plots

3) Every other figure on this page uses a number-in-a-circle to refer to the samples, but this one uses the full sample name. I like having the sample name, but it is a bit confusing to essentially use 2 different ways of referring to each sample

4) When the plot is wider than the page, there are no visual indications that the plot is scrollable (see https://cbioportal-frontend-pr-2560.herokuapp.com/patient?caseId=P01&studyId=lgg_ucsf_2014)

5) Overall, this plot just feels a bit out of place in the page. There should be a horizontal break line below the plot/above the mutations table. The text doesn't all need to be bold. There are probably other things to do that I can't think of at the moment...

6) This might be too much, but what about actually labeling the columns with gene name + protein change? Maybe as an option? Maybe just for download?

7) This plot should have a download option

8) This issue might be specific to lgg_ucsf_2014, but plots end up looking a little funny because indels in this study don't have an allele frequency - so when an indel is called, the samples with the indel have the "-" no data symbol, and samples without the indel show grey for AF=0 (see: https://cbioportal-frontend-pr-2560.herokuapp.com/patient?caseId=P17&studyId=lgg_ucsf_2014) Can we come up with a better way to deal with this? Maybe an alternate color for "variant is present in this sample but has no allele frequency"? The "no data" symbol should probably be reserved for "gene not on gene panel" like in oncoprint.

9) There's a huge amount of white space in between the sample name and the left-most edge of the heatmap

10) The 3-circle-button next to the sample name should probably be removed if it can't be used for anything (although if we do end up providing a non-clustered option, then let's keep it so users can re-order samples)

11) The heatmap color scheme shows 0=white, but the plot uses 0=grey. 

How's that to start @adamabeshouse ?  :)

alisman:
@tmazor @jjgao, a stab in the dark here:  It strikes me that chronology might be an important vector with cfdna.  Adam tells me that in this prototype, the samples are ordered according to clustering logic, which would seem to obscure the chronological relationship.   Could we/should we design the visualization to bring that out more?

jjgao:
wow! @tmazor 

@alisman about ordering, I agree with @tmazor that we should give the user option to switch between clustering and the same order as in header.

@adamabeshouse another request is an option to only show driver mutations (based on oncokb and hotspots).

adamabeshouse:
@jjgao @tmazor almost done. a question:
for (3), do we need the numbers in circle to be draggable? this is a key question in how i will implement it.

jjgao:
@adamabeshouse Could you explain a little more? Do you mean: do we need both drag & drop feature and circle feature?

adamabeshouse:
@jjgao we'll discuss today but what i meant is do you need to be able to drag and drop the order of the samples

adamabeshouse:
@tmazor @jjgao regarding the column names:

- do we want to implement it interactive, or just for download?
- do we want it on the top, or bottom?

jjgao:
@adamabeshouse let's try bottom. It'd be good to make an option that you can add them to the heatmap (off by default).

adamabeshouse:
@tmazor @jjgao do we need (3) and (6) for the demo?

tmazor:
@adamabeshouse I'm not sure what demo you're talking about... ?

adamabeshouse:
@jjgao mentioned wanting to make a demo of the new mutations tab

adamabeshouse:
merged into https://github.com/cBioPortal/cbioportal-frontend/pull/2573

PR #2561, opened 2019-07-22
Fix alteration enrichemnts when case of all groups are altered

kalletlak:
# What? Why?
Frontend changes for https://github.com/cBioPortal/cbioportal/issues/6395

Backend enrichments api wont be sending p-value when cases of all groups(>2) are altered. In this case p-value and q-value is are shown as `NA` in the table.

### DIFF
--- a/src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
+++ b/src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
@@ -22,6 +22,7 @@ export interface IAlterationEnrichmentTableProps {
     onCheckGene?: (hugoGeneSymbol: string) => void;
     onGeneNameClick?: (hugoGeneSymbol: string) => void;
     checkedGenes?:string[];
+    groupsCount:number
 }
 
 export enum AlterationEnrichmentTableColumnType {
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -120,17 +121,25 @@ export default class AlterationEnrichmentTable extends React.Component<IAlterati
...
-            render: (d: AlterationEnrichmentRow) => <span style={{whiteSpace: 'nowrap'}}>{toConditionalPrecision(d.pValue, 3, 0.01)}</span>,
-            tooltip: <span>Derived from Fisher's exact test</span>,
-            sortBy: (d: AlterationEnrichmentRow) => d.pValue,
+            render: (d: AlterationEnrichmentRow) => {
+                const pValue = d.pValue == undefined ? 'NA' : toConditionalPrecision(d.pValue, 3, 0.01);
+                return <span style={{ whiteSpace: 'nowrap' }}>{pValue}</span>;
+            },
+            tooltip: <span>Derived from {this.props.groupsCount > 2 ? "Chi-Square" : "Fisher's exact"} test</span>,
+            // set to 0 when p-value is undefined. this happens when all the cases in all groups are altered
+            sortBy: (d: AlterationEnrichmentRow) => d.pValue || 0,
```
adamabeshouse: just confirming it's intended that now these rows will appear at the top when sorting in ascending order of p-value?

kalletlak: yes, this is to indicate all the cases are altered in all groups

kalletlak:
@jjgao looks like p-value( through Fisher exact test) is 1 when there are 2 groups in comparison. which is different when there are >2 groups. is this OK?

PR #2573, opened 2019-07-30
patient view mutations tab with VAF line chart and oncoprint

adamabeshouse:
- [x] when hovering a line, highlight the whole line
- [x] use the same sample order as in the header
- [x] add the circled numbers
- [x] cross highlight between table and chart.
- [x] turn on new mutations tab using URL parameter - enableMutations=true
- [x] special search terms: oncogenic, hotspot
- [x] downloadable VAF line chart
- [x] tabs for heatmap and VAF chart
- [x] never say wild type, just say not detected, in tooltip
- [x] tab name Genomic Evolution
- [x] dont show mutations in heatmap with no VAF data in any sample
- [x] don't keep hiding it behind enableMutations=true
- [x] hide mutations tab if theres no VAF data for any mutation
- [x] transpose heatmap - sample axis on bottom. have transpose button. start the way it is now.
- [x] curves shouldnt be highlighted when its "show only highlighted". rows shouldnt be highlighted when its "show only highlighted"

**next release:**
- [ ] allow custom ordering of samples https://github.com/cBioPortal/cbioportal-frontend/pull/2727
- [ ] handle case of mutated but no VAF data in heatmap: show some kind of glyph, and make it clusterable
- [ ] allow custom selection of samples (? what does this mean)

https://cbioportal-f-pv-mutatio-c012sw.herokuapp.com/patient/genomicEvolution?caseId=P04&studyId=lgg_ucsf_2014&enableMutations=true
image


jjgao:
@adamabeshouse this is awesome!

@tmazor unfortunately, you cannot access the msk sample. The lgg one gives a taste.

https://www.cbioportal.org/patient/mutations?caseId=P04&studyId=lgg_ucsf_2014

adding others @cBioPortal/product @alisman for comments.

jjgao:
@adamabeshouse it would also be cool to allow custom selection and ordering of samples. I am wondering if we can utilize the header line (similar to comparison page)?

jjgao:
just noticed the sample orders are different from the one in header. It would be good to use the same order.

Is it possible to add the circled number here?

jjgao:
- [ ] when hovering a line, highlight the whole line
- [ ] use the same sample order as in the header
- [ ] add the circled numbers
- [ ] allow custom selection and ordering of samples. I am wondering if we can utilize the header line (similar to comparison page)

adamabeshouse:
lets do custom selection/ordering in a future work/PR?

adamabeshouse:
@jjgao updated

tmazor:
Very cool @adamabeshouse !

Some thoughts...
1) I like the idea of being able to reorder samples in the header!
2) I see you have "cross highlight between table and chart" in your list, and I'm not sure exactly what you have in mind for that, but there are a few ways I can think of that might be nice to link the chart and the table:
    - when I search/filter the table, highlight those mutations in the line chart
    - allow the user to select/click-and-drag mutations in the chart, and then update the table accordingly -- for example, there are several mutations at ~50% in sample 2 and it would be nice to highlight those lines and see what those mutations are in the table 
image
https://www.cbioportal.org/patient/mutations?caseId=P17&studyId=lgg_ucsf_2014
3) The y-axis should probably have a label
4) Since red is used for metastatic samples, maybe we should use a different color for these lines? Maybe something more neutral?

adamabeshouse:
@tmazor @jjgao lets try to be precise with our language because the concepts can get tricky here

so currently we have two concepts: highlighting and selecting. when we mouse over a row in the table, or mouse over a line in the plot, that line and the mutation row get *highlighted*. when we click the checkbox to select a row, only the lines corresponding to the checked rows stay in the VAF plot. 

so what do you have in mind when you're talking about highlighting in response to filtering the table, and to selecting multiple mutations in the plot?

adamabeshouse:
@tmazor @jjgao do we need reordering for the first pass?

schultzn:
This looks really nice. 

Just want to flag a few things for the graph:
1 - the y-axis needs a label
2 - it would be nice to separate the axes from the data, i.e. leave a bit of a gap between the y-axis and the first sample, as well as a gap between 0 and the sample labels
3 - the y-axis needs a label for 0 (zero)

adamabeshouse:
@schultzn thanks niki. Not sure I see the difference between "a gap between y-axis and first sample" and "a gap between 0 and the sample labels". Do you mean just padding on both axes?

schultzn:
Yes, padding on both axes… sorry.

adamabeshouse:
@schultzn updated

tmazor:
@adamabeshouse here's a hopefully more clear description of what I was thinking in (2), using this patient as an example:
https://www.cbioportal.org/patient/mutations?caseId=P17&studyId=lgg_ucsf_2014

- If I filter the table to TP53, I would like the lines in the chart corresponding to those 2 mutations to be highlighted. All other lines would still be visible, but those two mutations would be highlighted in blue.

- If I click on a line in the line chart, that line will be "permanently" highlighted in blue, all other mutations will remain visible in the chart, and the table below will filter to show just the selected mutation.

- Ideally, this could be extended to allow clicking and dragging over multiple lines to select multiple mutations, and the table below would filter to show all selected mutations.

These are slightly different than the checkbox, which essentially remove all other mutations from the line chart. I think it would be useful to have mechanisms to highlight a subset of mutations relative to all other mutations.

adamabeshouse:
> If I click on a line in the line chart, that line will be "permanently" highlighted in blue, all other mutations will remain visible in the chart, and the table below will filter to show just the selected mutation.

Maybe instead of filtering the table in this case, we can sort the selected mutations to the top?


I think it's a bit confusing that theres two kinds of filtering
Filtering the table causes highlighting the curves, and vice-versa
But selecting the mutations causes only showing certain curves

I'm finding that a bit confusing, like there are two similar but not identical mechanisms going on.

I think the reason I find it confusing is that theres no way to *select* a curve from the VAF plot - you have to select it (and thus hide others) from the table. 

I don't think it's inherently confusing to have these two separate mechanisms, but I think the UI should be better. What do you think?

adamabeshouse:
bump @tmazor @jjgao

tmazor:
@adamabeshouse great points!

I don't have a great answer, and will keep thinking, but here's one thought I had:

What if we make the decision to filter vs highlight a user setting?

We could allow users to select mutations in the VAF plot by clicking on lines/dragging over lines and in the mutations table by clicking the check boxes. By default each of those actions highlights the mutations in the other.

Then there could be a button or something that a user could click to filter based on their selections rather than just highlight.

adamabeshouse:
@tmazor  I just talked with @jjgao and we were thinking that maybe we could skip the feature of highlighting curves when filtering the table. That gives a pretty clean situation. We still filter the table on click and on click-drag. And users can still filter the table, then run their mouse over the rows to see the highlighted curves - we couldn't think of a filter use case where this wouldn't be sufficient. What do you think?

tmazor:
sure @adamabeshouse  - that seems reasonable to me

adamabeshouse:
@tmazor @jjgao please see updated heroku build

inodb:
@adamabeshouse this is so cool!	🎉 

It might be useful to have optional log scale on the y axis. Hard to see absence vs presence currently. With CCF this would be less of a problem

tmazor:
@adamabeshouse something isn't working... i can load the patient view mutations tab, but as soon as my mouse hovers over the VAF plot, the screen goes blank and this error appears in the console
image

adamabeshouse:
@tmazor link?

tmazor:
@adamabeshouse https://www.cbioportal.org/patient/mutations?caseId=P01&studyId=lgg_ucsf_2014#navCaseIds=lgg_ucsf_2014:P01,lgg_ucsf_2014:P02,lgg_ucsf_2014:P04,lgg_ucsf_2014:P05,lgg_ucsf_2014:P06,lgg_ucsf_2014:P07,lgg_ucsf_2014:P08,lgg_ucsf_2014:P09,lgg_ucsf_2014:P10,lgg_ucsf_2014:P11,lgg_ucsf_2014:P12,lgg_ucsf_2014:P13,lgg_ucsf_2014:P15,lgg_ucsf_2014:P16,lgg_ucsf_2014:P17,lgg_ucsf_2014:P18,lgg_ucsf_2014:P21,lgg_ucsf_2014:P24,lgg_ucsf_2014:P25,lgg_ucsf_2014:P26,lgg_ucsf_2014:P27,lgg_ucsf_2014:P28,lgg_ucsf_2014:P29

adamabeshouse:
@tmazor updated

inodb:
@adamabeshouse 	if there's only one sample it might be good not to show the tab? https://cbioportal-frontend-pr-2573.herokuapp.com/patient/mutations?caseId=TCGA-BK-A0CC&studyId=ucec_tcga_pub

inodb:
@adamabeshouse 	@tmazor 	noticed another small thing. Looks like a data issue. For some mutations `Ref Reads` is missing and it results in some strange behavior in the plot (the line is not drawn between all samples in the plot):

Screen Shot 2019-08-15 at 2 27 09 PM

tmazor:
ooo good catch @inodb  - the indel caller we used for this study didn't provide allele counts, so none of the indels have allele frequencies - so they show up as 0 in samples that they weren't called in, and N/A in the samples where they were called. I think this came up in the heatmap PR also. Maybe just exclude mutations that don't have allele freqs? I can't really think of a good way to visualize a mutation without an allele freq in this plot.

@adamabeshouse this is looking so awesome! 

One thought - I love how selecting lines in the chart makes the highlight stick! What about doing the same thing from the table? So if I click on a row in the table, that makes the line stay highlighted? When there are many mutations, it can be hard to click on the line of the specific mutation that you might want, so doing it from the table could be easier.

adamabeshouse:
love that idea @tmazor . 

@inodb @tmazor regarding the partial lines, @jjgao and I discussed this. the solution we came up with is:

- if a sample either isnt profiled for the gene, isnt mutated in the gene, or doesnt have complete VAF data for the mutation, then its a "no data" point. 

- a "no data" point that has some "yes data" point to its left *and* right is plotted as a gray dot, with info about the reason for no data in the tooltip, and its y-axis coordinate is given by linear interpolation with the "yes data" points on either side.

- a "no data" point that does not have "yes data" points to its left and right are omitted and the line does not go all the way there.

should we handle it differently?

tmazor:
@adamabeshouse @jjgao I like the plan for "no data" if a gene isn't profiled or a mutation has no VAF data. But if I'm understanding correctly, a sample which is profiled for a gene but doesn't have a particular mutation detected will be classified as "no data"? I think a mutation that is profiled but undetected should be plotted at VAF=0. Imagine a patient with multiple cfDNA samples -- a mutation might be detected at timepoints 1 & 3, but not be detected at timepoint 2. Timepoint 2 might represent a response to therapy or drop in disease burden or an alternate clone emerging or some other biological phenomena -- and that would be lost if we interpolate a VAF for it.

adamabeshouse:
@tmazor I misspoke - yes that is exactly how we're handling it as per @jjgao's specification

adamabeshouse:
@jjgao @tmazor now that we can click to highlight a row, I think the checkbox will become more confusing - we should explicitly add an eye icon to the header of that row, or something, which indicates visibility

adamabeshouse:
@tmazor we're running into a confusing UI double standard:

so when you click on a curve, it highlights it and filters the table. 

now we're suggesting when you click on a row, it highlights the curve. but should it filter the table? if not, then it is a UI incongruence. if so, then it seems to partially defeat the purpose of allowing clicking on the rows, because as soon as you click on the row you cant click on other rows.

adamabeshouse:
@jjgao @inodb @tmazor @schultzn 

Merged the oncoprint into this PR, see updated heroku!

jjgao:
@adamabeshouse @tmazor  I find it confusing to have both selection and highlighting in the table. And the eye icon does not necessarily make it more clear. Maybe we can do the followings:

- Remove the select column
- keep the click and highlight functionality
- provide an option in the line chart to only show the highlighted mutations

tmazor:
woah... there are two mutations tabs....

image

adamabeshouse:
@tmazor oops I have no idea how that happened! should be fixed now

tmazor:
The two mutations tabs was kinda fun, but this is definitely clearer :)

Some thoughts...

1) The Summary tab has thin horizontal lines separating the sections - can we try that here? 
2) I remember a discussion of how to deal with mutations without AFs... has that just not been implemented yet? Because I still see indels (which have no AF in lgg_ucsf_2014) showing up at 0 in the samples where that indel wasn't called.
3) I know we're still finalizing how the table & line chart will interact in terms of highlighting/filtering/etc, but now that the heatmap is on the same page, I think we also need to consider having those interactions include the heatmap. It's weird that right now I can hover over a mutation in the table, and it highlights in the line chart but not in the heatmap. This will obviously require some thought in terms of how to implement it, but I think we need to do something.
4) There are a lot of different styles of tooltips on this page - the line chart, heatmap and table all seem to be slightly different - can that be standardized?
5) Can mutation labels on the heatmap be optional? Or can we add a zoom? It would be nice to have a way to see all mutations at once, which is currently impossible for patients with many mutations.
6) For the heatmap tooltips, when a mutation isn't called, can the tooltip say "Mutation not detected" rather than "wildtype"?

adamabeshouse:
@tmazor on it. Specifically which mutation is problematic for your point (2)?

tmazor:
@adamabeshouse you can use the third mutation in the table (SPRTN) as an example https://cbioportal-frontend-pr-2573.herokuapp.com/patient/mutations?caseId=P04&studyId=lgg_ucsf_2014

tmazor:
Also - another issue I just realized. I had been testing on a large monitor where I could see the two charts & the table all at once. On a 13" laptop screen, it's hard to get the line chart & table to be visible at the same time (because of the heatmap in between). We might need to think about some alternate layouts for this tab.

adamabeshouse:
@tmazor for (6), you mean only when the mutation has reads but is uncalled, right?

adamabeshouse:
@tmazor @jjgao bump ^^

tmazor:
@adamabeshouse 
(6) Yes - I mean when the gene is included on the gene panel, and has reads covering it, and the mutation isn't called.

(3) Yes? I'm just not sure exactly what that looks like, but in general the more similar we can make interactions between the two charts, the better.

adamabeshouse:
@tmazor for (3) I mean would it be sufficient to highlight a column the same way the column is highlighted on mouseover *of the heatmap*

i.e. something like this
image

tmazor:
@adamabeshouse gotcha - yes, that seems reasonable. Maybe also highlight the column label with it (similar to how rows in the table get highlighted)?

adamabeshouse:
@tmazor @jjgao @schultzn bump, could you check it out at http://cbioportal-frontend-pr-2573.herokuapp.com/patient/mutations?caseId=P04&studyId=lgg_ucsf_2014&enableMutations=true

schultzn:
Looks really nice!

A couple of small comments: 
- left-justify log scale / download icon with the line chart, in the example it is not aligned with the chart at all
- is there a need for the show all mutations / only show highlighted mutations? What else would the selection of all vs individual variants be for? At the moment, it looks like one can select "show all" but then select only a couple of mutations (without an apparent effect)
- this seems related, but I can see that I can select specific mutations to be highlighted in the heatmap. However, it is not clear in the table which ones I have selected for that feature. See the screenshot below.
- will this feature be enable by default? What will happen to the current copy-number and mutation fingerprint view? And timeline data? Can users hide and un-hide this new view in any way?


image

jjgao:
@adamabeshouse this looks awesome!

@schultzn This is in the new Mutations tab - we should only show it when there are mutliple samples. Maybe we can release this one first and then integrate it more with timeline and igv view later?

@adamabeshouse I find two ways of selection (checkbox and row click) confusing. I would propose to remove the checkbox column and add an option in the line chart to show only selected mutations.

adamabeshouse:
looping in @tmazor regarding the hiding/selecting issue. what do you think?

tmazor:
I think @jjgao's show all/show selected option for the line chart makes sense. Let's do that.

A few other thoughts:
- The heatmap right now uses the same color for "Mutated but we don't have VAF data" and "No mutation". I think "Mutated but we don't have VAF data" should be a different color. Or maybe mutations without a VAF in any sample just shouldn't be included? That's what we're doing with the line chart, right?
- Can heatmap tooltips say "No mutation detected" rather than "Wildtype (VAF:0)"?
- I can't think of an example to test this, but will there be a different representation for "gene not on panel" vs "gene on panel but VAF=0" for the line chart and heatmap? (for patients with samples sequenced on different panels)

adamabeshouse:
@tmazor It does say no mutation detected when there are supporting reads, but for 0 supporting reads it will just say wild type. And yes it does say "not profiled" vs "wild type"

Just let me know what to do about rendering the various no data cases. I agree that mutated w/o VAF is qualitatively different than wild type.. but I'm not sure what would be useful to show for the purposes of the heatmap.

adamabeshouse:
@jjgao @tmazor @schultzn just pushed an update with all the requested changes. let me know if I missed anything

tmazor:
Ok, let's try this for the heatmap:
First, a question - when the heatmap is clustered, it uses VAF, right? Is it possible to include the mutations without a VAF in the clustering?
- If yes, then when a mutation is detected, but no VAF is available, color the box black and add a legend item for black = "Mutation detected but no VAF reported"
- If no, then mutations that don't have a VAF in at least one sample should not be included in the heatmap

I'm a bit confused about the tooltip texts and what you mean by supporting reads... don't we only know read counts if the mutation was called? This is my proposal for the tooltips:
- If a mutation is called and has a VAF, the VAF is displayed
- If a mutation is called and doesn't have a VAF - well, that's the discussion above
- If a mutation is not called in a sample, but the gene is included in the gene panel used for that sample, then the tooltip should say "Mutation not detected"
- If a mutation is not detected in a sample, and the gene is not part of the gene panel, then the tooltip should say "Not profiled"

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -269,22 +265,36 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+                return Promise.resolve(new SampleManager(patientData.samples!));
+            }
+        },
+        default: null
+    });
 
     public render() {
-
-        let sampleManager: SampleManager | null = null;
+        const sampleManager = this.sampleManager.result!;
```
alisman: i know we've debated this before but i just find this so weird.  it's not a sampleManager, it's a sampleManagerPromise.   Data I can deal with being packaged with the business of its loading, but I a whole object like this is stranger.  I would rather just see a list of awaited promises out in the open.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -619,6 +638,47 @@ export class PatientViewPageStore {
...
     }, []);
 
+    readonly coverageInformation = remoteData<CoverageInformation>({
+        await:()=>[
+            this.mutatedGenes,
+            this.samples,
+            this.molecularProfilesInStudy
+        ],
+        invoke:async()=>{
+            const sampleMolecularIdentifiers:SampleMolecularIdentifier[] = [];
```
alisman: this whole invoke could use comments

src/pages/patientView/mutation/PatientViewMutationsDataStore.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    public get onlyShowSelectedInTable() {
+        return this._onlyShowSelectedInTable;
+    }
+
+    public get onlyShowSelectedInVAFChart() {
+        return this._onlyShowSelectedInVAFChart;
+    }
+
+    public setMouseOverMutation(m:Readonly<Mutation>|null) {
```
alisman: action?   i know these only set one thing, but i think it's good for less experienced devs to keep ethic "always make mutating methods actions".

src/pages/patientView/mutation/PatientViewMutationsDataStore.ts
```diff
@@ -0,0 +1,104 @@
...
+
+    public setOnlyShowSelectedInVAFChart(o:boolean) {
+        this._onlyShowSelectedInVAFChart = o;
+    }
+
+    @action
+    public toggleSelectedMutation(m:Readonly<Mutation>) {
+        const key = mutationIdKey(m);
+        if (this.selectedMutationsMap.has(key)) {
+            this.selectedMutationsMap.delete(key);
```
alisman: would be interesting to discus whether we should be using maps like this or more immutable approach where we use lodash methods to replace entire reference.  i'm not totally sure.

src/pages/patientView/mutation/VAFLineChart.tsx
```diff
@@ -0,0 +1,773 @@
...
+                            fill:"white",
+                            strokeWidth:2,
+                            opacity:1,
+                            pointerEvents:"none"
+                        }}
+                    />
+                ));
+
+                return (
+                    <Portal isOpened={true} node={this.highlightContainer}>
```
alisman: curious why this is in a portal

src/pages/patientView/mutation/VAFLineChart.tsx
```diff
@@ -0,0 +1,773 @@
...
+        } else {
+            return (
+                <div style={{
+                    display:"flex",
+                    justifyContent:"center",
+                    alignItems:"center",
+                    height:this.svgHeight,
+                    width:"100%"
+                }}>
+                    No VAF data to show. Select mutations below or un-check the "{SHOW_ONLY_SELECTED_LABEL}" filter above.
```
alisman: should this be `${SHOW_ONLY_SELEcTED_LABEL}` . ??

src/pages/patientView/mutation/oncoprint/MutationOncoprint.tsx
```diff
@@ -0,0 +1,571 @@
...
+            const mutations = [];
+            for (const d of this.mutations) {
+                mutations.push({
+                    mutation: d,
+                    id: generateMutationIdByGeneAndProteinChangeAndEvent(d)
+                });
+            }
+            return Promise.resolve(
+                _.chain(mutations)
+                    .uniqBy((m:any)=>m.id)
```
alisman: how come any?

src/shared/lib/invertIncreasingFunction.ts
```diff
@@ -1,9 +1,41 @@
+function findXRange(
+    func:(x:number)=>number,
+    y:number
+) {
+    // find lower
+    let lower = 0;
+    let step = 1;
+    while (func(lower) > y) {
+        if (lower < Number.MIN_SAFE_INTEGER) {
+            debugger;
```
alisman: debugger

src/shared/lib/invertIncreasingFunction.ts
```diff
@@ -1,9 +1,41 @@
+function findXRange(
```
alisman: seems like this could use tests

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -882,6 +954,53 @@ export class PatientViewPageStore {
         return this.mergedDiscreteCNAData;
     }
 
+    @computed get existsSomeMutationWithVAFData() {
+        return _.some(this.mergedMutationDataIncludingUncalled, mutationList=>{
+            return _.some(mutationList, m=>{
+                const vaf = getVariantAlleleFrequency(m);
+                return vaf != null && vaf > 0;
```
alisman: we should discuss whether we think this should have a test or not.

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: but it shouldnt be recomputed on every render. so then it would be in a @computed. but then we need to duplicate the list of dependencies in render() as in @computed which IMO is not good at all

src/pages/patientView/mutation/VAFLineChart.tsx
adamabeshouse: ah its legacy from when I had to (1) put those in an observer (2) put those somewhere inside the victory component due to layering. but now since it just goes on the top it's not necessary. I'll simplify.

src/pages/patientView/mutation/oncoprint/MutationOncoprintUtils.ts
```diff
@@ -0,0 +1,150 @@
...
+    return _.groupBy<IMutationOncoprintTrackDatum>(allData, d=>d.mutationId);
+}
+
+export function makeMutationHeatmapData(
+    samples:Sample[],
+    mutations:Mutation[],
+    coverageInformation:CoverageInformation,
+    mode:MutationOncoprintMode
+) {
+    const mutationsByKey = _.keyBy(mutations, generateMutationIdByGeneAndProteinChangeAndEvent);
```
adamabeshouse: tests

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -481,6 +493,41 @@ export default class MutationTable<P extends IMutationTableProps> extends React.
                 userEmailAddress: this.props.userEmailAddress,
                 studyIdToStudy: this.props.studyIdToStudy
             })),
+            filter: (d:Mutation[], filterString:string, filterStringUpper:string)=>{
+                (window as any).data = this.props.oncoKbData;
```
adamabeshouse: delete

src/pages/patientView/mutation/VAFLineChart.tsx
```diff
@@ -0,0 +1,773 @@
...
+        return order;
+    }
+
+    @computed get sampleIdIndex() {
+        return stringListToIndexSet(this.sampleIdOrder);
+    }
+
+    @computed get renderData() {
+
+        const grayPoints:IPoint[] = []; // points that are purely interpolated for rendering, dont have data of their own
```
adamabeshouse: tests

src/pages/patientView/SampleManager.tsx
```diff
@@ -58,7 +58,8 @@ export function sortSamples(samples: Array<ClinicalDataBySampleId>,
         let eventOrdering = collectionDayMap[id];
 
         // 2. if cases have derived normalized case types, put primary first
-        let sampleTypeIndex = sampleTypeOrdering.indexOf(clinicalDataLegacyCleanAndDerived[id].DERIVED_NORMALIZED_CASE_TYPE);
+        const caseType = clinicalDataLegacyCleanAndDerived[id].DERIVED_NORMALIZED_CASE_TYPE;
+        let sampleTypeIndex = caseType ? sampleTypeOrdering.indexOf(caseType.toLowerCase()) : -1;
```
alisman: instead of first assigning -1 and then correcting it on ln 64, could we just assing sampleTypeOrdering.lenth in the ternary?

src/pages/patientView/mutation/PatientViewMutationsDataStore.ts
```diff
@@ -0,0 +1,104 @@
...
+            const selectedFilter = !this._onlyShowSelectedInTable || this.selectedMutations.length === 0 || _.some(this.selectedMutations, m=>mutationMatch(d, m));
+
+            return stringFilter && selectedFilter;
+        });
+    }
+
+    constructor(getData:()=>Mutation[][]) {
+        super(getData);
+
+        this.dataHighlighter = (d:Mutation[])=>{
```
alisman: i don't feel like you used to use single letter variables like this.  i find it kind of hard to follow.

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -25,6 +25,7 @@ import {
 } from "./TooltipUtils";
 import {MolecularProfile} from "../../api/generated/CBioPortalAPI";
 import {AlterationTypeConstants} from "pages/resultsView/ResultsViewPageStore";
+import ifndef from "../../lib/ifndef";
```
alisman: if not defined?  hard to know what this is

adamabeshouse: oh sorry, ifndef is a preprocessor command from C that does the same thing, so I thought it would work as a name.

It stands for if not defined

src/pages/patientView/SampleManager.tsx
adamabeshouse: good call

src/pages/patientView/mutation/PatientViewMutationsDataStore.ts
adamabeshouse: oh  sorry, let me change that

src/pages/patientView/SampleManager.tsx
adamabeshouse: oh I just realized, that doesn't capture the case when `sampleTypeOrdering.indexOf(caseType.toLowerCase())` is -1

alisman:
@adamabeshouse Adding a little padding here

image

alisman:
@adamabeshouse noticed a bug, if you click a single sample (header of page), we get an exception and page breaks

PR #2577, opened 2019-07-31
Support urls in clinical data

kalletlak:
# What? Why?
Partial fix https://github.com/cBioPortal/cbioportal/issues/6383
Supports only links in clinical data in study view and patient view pages

Example

Screen Shot 2019-07-31 at 12 30 54 PM
Screen Shot 2019-07-31 at 12 31 20 PM

### DIFF
--- a/src/public-lib/lib/StringUtils.spec.ts
+++ b/src/public-lib/lib/StringUtils.spec.ts
@@ -1,6 +1,6 @@
 import { assert } from 'chai';
 import {
-    unescapeTabDelimited, longestCommonStartingSubstring, stringListToIndexSet, stringListToSet, unquote
+    unescapeTabDelimited, longestCommonStartingSubstring, stringListToIndexSet, stringListToSet, unquote, isUrl
 } from "./StringUtils";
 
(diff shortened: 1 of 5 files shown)


src/public-lib/lib/StringUtils.spec.ts
```diff
@@ -38,4 +38,16 @@ describe("unquote and decodeTabDelimited", () => {
         assert.equal(unquote(unescapeTabDelimited("\"cBioPortal\\n'Redefining'\\t\"Data Analysis\"\"")),
             "cBioPortal\n\'Redefining\'\t\"Data Analysis\"");
     });
+});
+
+
+describe("isUrl", () => {
+    it("gives correct result on various inputs", ()=>{
+        assert.isFalse(isUrl("example.com"))
```
zhx828: We don't want to support any link without protocol?

kalletlak: that isn't supported. only http and https are supported
btw its `assert.isFalse`

zhx828:
These failing tests are related to the base branch.

PR #2583, opened 2019-08-05
Make the EllipsisTextTooltip a block component

zhx828:
- Make the EllipsisTextTooltip a block component
- Remove property shownWidth (it was introduced in https://github.com/cBioPortal/cbioportal-frontend/pull/2110 to solve the scrollWidth is 0 issue, which has explained in the following link)
- Also let TrucatedText component supports `never` addTooltip

Based on the following article, the scrollWidth is not alwasy working especially in chrome. We are changing the component to a block component.
https://stackoverflow.com/questions/13226296/is-scrollwidth-property-of-a-span-not-working-on-chrome


zhx828:
@adamabeshouse I agree. Any place wants to use the EllipsisTextTooltip should update the parent style considering this is a block element. Or use the TruncatedText for a inline component.

alisman:
So, what's the process for merging this?  I mean, sounds like there need to be updates?

zhx828:
@alisman all components where use the EllipsisTextTooltip have been either updated to new spec or changed to use TruncatedText. The pr should be good to go if there is no other code related issues.

zhx828:
@alisman oh I'm gonna update the screenshots

zhx828:
@alisman updated all screenshots.

I do think all 10 screenshots should be updated, @adamabeshouse just in case, could you take a look?

PR #2588, opened 2019-08-07
Fix query link after saving virtual study

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5179


alisman:
@kalletlak There was some github bug.  This was merged successfully but github didn't record it as merged.

PR #2603, opened 2019-08-15
Fix issue while fething user preference

kalletlak:


zhx828:
The failing tests are in the base branch, not related to this pr.

PR #2606, opened 2019-08-16
Change treatmentCache from @cached getter to readonly variable

pvannierop:
# Problem
When treatment response profiles are the only type of molecular profile in a study, heatmap tracks do not show for these profiles in oncoprint.

# Fix
The @cached annotation of the treatmentCache caused this behaviour. After consultation with @adamabeshouse  the @cached annotation was replaced with a readonly variable. This solves the problem, but leaves open why the @cached annotation caused problems.


pvannierop:
Of course. GeneCache and genesetCache were changed into readonly vars.

inodb:
Gonna merge this - let's fix further in post review if new errors show up

PR #2609, opened 2019-08-19
Clean up and modify overlap exclusion indicator

adamabeshouse:


src/pages/groupComparison/OverlapExclusionIndicator.tsx
```diff
@@ -12,174 +13,168 @@ export interface IOverlapExclusionIndicatorProps {
...
-                            </span>
-                        );
-                    }
-                    break;
+        return (
+            <div>
+                <i
+                    className={`fa fa-md ${this.classNames.icon}`}
+                    style={{
+                        color: "#000000",
```
alisman: isn't this default color?

adamabeshouse: true

tmazor:
Looks good! Just a few super minor comments:

- I think the first line here (overlap tab, overlapping samples included) can be removed since it's mostly redundant with the new text.
image

- For consistency with the updated text on the other tabs, can we remove "n=" from this text (survival with overlapping samples included)?
image

- This is really me getting picky but... on the clinical tab, we show both the sample & patient text, and I'm finding it slightly confusing to say "this sample-level analysis" and "this patient-level analysis". How about changing it to say "...[included or excluded] from [sample or patient]-level analysis below." If we do this, let's do it on all tabs, not just clinical.

- I don't think I can test this, but the 3.1 release has the patient-level enrichment option. Just want to confirm that when this code is merged with that code, that the Mutations and Copy-Number tabs will switch between the sample & patient messages according to which is selected?

tmazor:
Actually I just came across a weird bug...

Go to the link below. When I move my mouse over the KM plot, the page goes blank and the following appears in the console. This happens only when local.storage is pointing to this PR.

image


https://www.cbioportal.org/comparison/survival?overlapStrategy=Include&sessionId=5cf6aa24e4b0ab4137874305&unselectedGroups=%5B%22Thyroid%20Cancer%22%2C%22Soft%20Tissue%20Sarcoma%22%2C%22Skin%20Cancer%2C%20Non-Melanoma%22%2C%22Esophagogastric%20Cancer%22%2C%22Endometrial%20Cancer%22%2C%22Head%20and%20Neck%20Cancer%22%2C%22Glioma%22%2C%22Gastrointestinal%20Stromal%20Tumor%22%2C%22Melanoma%22%2C%22Hepatobiliary%20Cancer%22%5D

adamabeshouse:
@tmazor I made your first three changes. Re: patient enrichments, there's no way to make that change before we merge it unfortunately, so I'll just have to consider this issue open until we merge master to release, and I'll make the necessary changes.

Thats a weird error, I'm looking into it. Note to self: looks like an issue with using a react functional component in heroku

adamabeshouse:
@tmazor actually looks now like the error has disappeared. can you confirm?

tmazor:
Awesome! Looks great to me!

I can no longer replicate that error anymore either, so yes, I guess it is fixed.

src/pages/groupComparison/OverlapExclusionIndicator.tsx
adamabeshouse: I was just about to take it out but then I thought, theres some chance that errant CSS could target this, so maybe it makes sense to make it explicit here?

alisman: I mean, if it's part of the class of things "text" then it should be default.  If not then ....

PR #2612, opened 2019-08-19
Fix layout in localdb e2e tests.  Add attribute brackets.

alisman:


end-to-end-test/local/runtime-config/run_container_screenshot_test.sh
```diff
@@ -10,6 +10,7 @@ docker run -it --rm \
     --network="$DOCKER_NETWORK_NAME" \
     -e CBIOPORTAL_URL="$CBIOPORTAL_URL" \
     -e SCREENSHOT_DIRECTORY="$SCREENSHOT_DIRECTORY"\
+    -e FRONTEND_TEST_USE_LOCAL_DIST=true \
```
alisman: in absence of this setting, tests will be run with localdev=true.   the app will still work, but the global stylesheet will not be included.   (in development mode, the local stylesheet it is injected by dev server as a <style> tag so that it can be hot-reloaded.)

PR #2619, opened 2019-08-20
Save cancer genes filter to user settings

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6274

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -137,7 +137,8 @@ export type ChartUserSetting = {
         w: number;
         h: number;
     },
-    patientAttribute: boolean
+    patientAttribute: boolean,
+    filterByCancerGenes?:boolean
 }
 
 export type StudyPageSettings = {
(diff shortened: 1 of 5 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2443,12 +2449,25 @@ export class StudyViewPageStore {
         }
     });
 
+    @observable _filterMutatedGenesTableByCancerGenes?: boolean;
```
alisman: is there a reason not to just initialize these here? why leave them undefined?

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -196,6 +196,10 @@ export class StudySummaryTab extends React.Component<IStudySummaryTabProps, {}>
                 props.getData = () => this.store.getMutatedGenesDownloadData();
                 props.genePanelCache = this.store.genePanelCache;
                 props.downloadTypes = ["Data"];
+                props.filterByCancerGenes= this.store.filterMutatedGenesTableByCancerGenes;
+                props.onToggleCancerGeneFilter = (filtered: boolean) => {
+                    this.store._filterMutatedGenesTableByCancerGenes = filtered;
```
alisman: if the property has an underscore, i think that is supposed to indicate it's private.  but here we mutate it from outside.    i realize we have consistent convention here.     also, this should be runInAction.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2398,12 +2399,17 @@ export class StudyViewPageStore {
...
                 const id = attribute.uniqueKey
+                const chartType = this.chartsType.get(id)
                 chartSettingsMap[attribute.uniqueKey] = {
                     id,
-                    chartType: this.chartsType.get(id), // add chart type
+                    chartType,
                     patientAttribute: this.chartMetaSet[id].patientAttribute // add chart attribute type
                 }
-
+                if(chartType === "MUTATED_GENES_TABLE") {
```
alisman: seems like these constants are used elsewhere so maybe an enum?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2398,12 +2399,17 @@ export class StudyViewPageStore {
         if(this.isSavingUserPreferencePossible) {
             this.visibleAttributes.forEach(attribute => {
                 const id = attribute.uniqueKey
+                const chartType = this.chartsType.get(id)
```
alisman: lets get some semi colons here

PR #2627, opened 2019-08-22
Enrichments: Sync Gene Bar plot to table

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6143

### DIFF
--- a/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
+++ b/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
@@ -27,16 +27,12 @@ export const CNA_TO_ALTERATION:{[cna:number]:string} = {
     "-2": "HOMDEL"
 };
 
-export enum GeneOptions {
+export enum GeneOptionLabel {
     USER_DEFINED_OPTION='User-defined genes',
     HIGHEST_FREQUENCY='Genes with highest frequency in any group',
     AVERAGE_FREQUENCY='Genes with highest average frequency',
-    SIGNIFICANT_P_VALUE='Genes with most significant p-value'
-}
-
-export const USER_DEFINED_OPTION:{label:string, genes:string[]} = {
-    label: GeneOptions.USER_DEFINED_OPTION,
-    genes: []
+    SIGNIFICANT_P_VALUE='Genes with most significant p-value',
+    SYNC_WITH_TABLE="Sync with table (up to 100 genes)"
 }
 
 export enum AlterationContainerType {
@@ -586,9 +582,12 @@ export function getEnrichmentBarPlotData(data: { [gene: string]: AlterationEnric
     });
 }
 
-export function getGeneListOptions(data: AlterationEnrichmentRow[], includeAlteration?:boolean): { label: string, genes: string[] }[] {
+export function getGeneListOptions(data: AlterationEnrichmentRow[], includeAlteration?:boolean): { label: GeneOptionLabel, genes: string[] }[] {
     if (_.isEmpty(data)) {
-        return [USER_DEFINED_OPTION];
+        return [{
+            label: GeneOptionLabel.USER_DEFINED_OPTION,
+            genes: []
+        }];
     }
 
     let dataWithOptionName: (AlterationEnrichmentRow & {optionName?:string})[] = data;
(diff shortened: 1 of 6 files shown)


src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -137,6 +137,20 @@ export default class AlterationEnrichmentTable extends React.Component<IAlterati
         return columns;
     }
 
+    @observable.ref alterationEnrichmentTableComponent:AlterationEnrichmentTableComponent|null = null;
+
+    @autobind
+    private alterationEnrichmentTableComponentRef(ref:AlterationEnrichmentTableComponent|null) {
```
alisman: will this param actually ever be null?

src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -137,6 +137,20 @@ export default class AlterationEnrichmentTable extends React.Component<IAlterati
...
+        this.alterationEnrichmentTableComponent = ref;
+    }
+
+    @computed get visibleGeneSymbols() {
+        if (this.alterationEnrichmentTableComponent) {
```
alisman: seems slighty weird.  you are consulting the view component for it's state.   But the view component itself seems to update store when clicked.  So shouldn't the state (visible genes) be derivable from the store?

src/pages/resultsView/enrichments/GeneBarPlot.tsx
```diff
@@ -155,11 +154,25 @@ export default class GeneBarPlot extends React.Component<IGeneBarPlotProps, {}>
         </div>)
     }
 
+    @computed private get selectedOption() {
+        if (this._label && this._geneQuery !== undefined) {
```
dippindots: Just want to confirm that `this._geneQuery` can be an empty string like `""`? Should we return the default result for that case?

kalletlak: if its empty then it would be in sync mode

src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -145,7 +159,8 @@ export default class AlterationEnrichmentTable extends React.Component<IAlterati
         return (
             <AlterationEnrichmentTableComponent initialItemsPerPage={20} paginationProps={{ itemsPerPageOptions: [20] }}
                 columns={orderedColumns} data={this.props.data} initialSortColumn={this.props.initialSortColumn} 
-                onRowClick={this.props.onGeneNameClick ? this.onRowClick : undefined} dataStore={this.props.dataStore}/>
+                onRowClick={this.props.onGeneNameClick ? this.onRowClick : undefined} dataStore={this.props.dataStore}
```
alisman: here is where the table component updates store state on click.  Where I've commented above, I think we should be reading the state from the store rather than from the visibleGenes property of the component.   It is a big antipattern for react components to read state from each other.  If state needs to be shared, it should be managed by a common store.  It would appear that the data already exists in the store.

src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
alisman: see comment below.

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -27,16 +27,12 @@ export const CNA_TO_ALTERATION:{[cna:number]:string} = {
...
     HIGHEST_FREQUENCY='Genes with highest frequency in any group',
     AVERAGE_FREQUENCY='Genes with highest average frequency',
-    SIGNIFICANT_P_VALUE='Genes with most significant p-value'
-}
-
-export const USER_DEFINED_OPTION:{label:string, genes:string[]} = {
-    label: GeneOptions.USER_DEFINED_OPTION,
-    genes: []
+    SIGNIFICANT_P_VALUE='Genes with most significant p-value',
+    SYNC_WITH_TABLE="Sync with table (upto 100 genes)"
```
adamabeshouse: up to

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -27,16 +27,12 @@ export const CNA_TO_ALTERATION:{[cna:number]:string} = {
     "-2": "HOMDEL"
 };
 
-export enum GeneOptions {
+export enum GeneOptionLabels {
```
adamabeshouse: I think for an enum this should be singular (`GeneOptionLabel`)

src/shared/lib/ILazyMobXTableApplicationDataStore.ts
```diff
@@ -29,6 +33,8 @@ export class SimpleGetterLazyMobXTableApplicationDataStore<T> implements ILazyMo
     @observable public filterString:string;
     @observable public sortMetric:SortMetric<T>|undefined;
     @observable public sortAscending:boolean|undefined;
+    @observable public page:number;
+    @observable public itemsPerPage:number;
```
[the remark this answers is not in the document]
adamabeshouse: disagree with @onursumer  - the goal of Simple... is to provide the a simple implementation of ILazyMobXTableApplicationDataStore. since the interface changed, Simple.. must change as well

jjgao:
@kalletlak is this code on beta?

kalletlak:
@jjgao yes

PR #2628, opened 2019-08-22
Support remaining StudyViewFilter params in url hash

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/4983


zhx828:
@kalletlak never mind. This is the filter coming from the URL. I thought this is for the virtual study.

zhx828:
@kalletlak looks good to me, but could you figure out the e2e?

PR #2630, opened 2019-08-23
Fix querypage e2e test

pvannierop:
# What? Why?
In release-3.1.o branch there were e2e-tests failing.

# Fix
- A screenshot test that tested presence of a warning message on query page triggered by a invalid study id was replaced by a DOM based localdb test. I chose a DOM based test since this is sufficient to confirm functionality. 
- A reference screenshot for a patient view test was updated.


alisman:
this is just migrating test to local?

pvannierop:
No, not only migrating to local, but also transforming the screenshot test into a DOM based test.

PR #2646, opened 2019-08-29
Wrap contents of oncoprint tooltip to contents

pvannierop:
Note: this is a duplicate of [PR #2633](https://github.com/cBioPortal/cbioportal-frontend/pull/2633) from other remote (problems with triggering CircleCI jobs).

# What? Why?
Long sample names would overflow the boundaries of the oncoprint tooltip (see screenshots).

# Fix
The css properties of oncoprint tooltip were overridden so that the tooltip scales with the length of the sample names.

# Before
Screenshot from 2019-08-26 16-54-49
Screenshot from 2019-08-26 16-55-04

# After
image
image

### DIFF
--- a/src/shared/components/oncoprint/styles.scss
+++ b/src/shared/components/oncoprint/styles.scss
@@ -1,9 +1,14 @@
 .oncoprintjs__tooltip {
   padding: 5px;
   margin-top: -10px; // to compensate position for padding, since tooltip is positioned above (if positioned at center, wouldnt be problem)
-  width:250px;
+  max-width: none;
+}
+
+.oncoprint__tooltip {
+  min-width: 160px;
+  max-width: none;
 }
 
 .oncoprint-hidden {
   opacity: 0;
-}
+} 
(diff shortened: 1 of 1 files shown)


src/shared/components/oncoprint/styles.scss
```diff
@@ -1,9 +1,14 @@
 .oncoprintjs__tooltip {
   padding: 5px;
   margin-top: -10px; // to compensate position for padding, since tooltip is positioned above (if positioned at center, wouldnt be problem)
-  width:250px;
+  max-width: none;
+}
+
+.oncoprint__tooltip {
```
alisman: ?

pvannierop: I did not create these classes, but just updated the css to make it work. Style changes to both classes are required for the tooltip.

https://github.com/cBioPortal/cbioportal-frontend/pull/2633#discussion_r317689299
> @alisman i think max-width existed for reason. probably in some scenario there's a large body of text that we want to have wrap (@adamabeshouse ?). to fix this situation more surgically, we could apply word-break: break-all; to the the a tag that is at issue.
> @adamabeshouse thats correct

Ok, I will update the code to use word wrap.

pvannierop: See updated Images in the PR description for the tooltips with word-wrap.

PR #2647, opened 2019-08-29
Fix oncoprint sort for treatment profiles

pvannierop:
Note: this is a duplicate of [PR #2637](https://github.com/cBioPortal/cbioportal-frontend/pull/2637) from other remote (problems with triggering CircleCI jobs).

# What? Why?
Oncoprint does not handle sorting of treatment profile heatmap tracks correctly (see screenshot below)
63562317-2cf48a00-c52b-11e9-9cd9-d34e46a36c61

# Fix
The cause of this is that the sorting comparator function does not handle profile values that are `NaN` (value allowed for profile data). Problem was removed by filtering out NaN values in oncoprint.
fixed_oncoprint_sort

### DIFF
--- a/src/shared/components/oncoprint/DataUtils.ts
+++ b/src/shared/components/oncoprint/DataUtils.ts
@@ -283,16 +283,17 @@ export function fillHeatmapTrackDatum<T extends IBaseHeatmapTrackDatum, K extend
 ) {
     trackDatum[featureKey] = featureId;
     trackDatum.study_id = case_.studyId;
-    if (data) {
-        _.remove(data, (d) => isNaN(d.value)); // remove data with non-available values
-    }
-    if (!data || !data.length) {
+
+    // remove data points of which `value` is NaN
+    const dataWithValue =  _.filter(data, (d) => isNaN(d.value));
+
+    if (!dataWithValue || !dataWithValue.length) {
         trackDatum.profile_data = null;
         trackDatum.na = true;
-    } else if (data.length === 1) {
-        trackDatum.profile_data = data[0].value;
-        if (data[0].thresholdType) {
-            trackDatum.thresholdType = data[0].thresholdType;
+    } else if (dataWithValue.length === 1) {
+        trackDatum.profile_data = dataWithValue[0].value;
+        if (dataWithValue[0].thresholdType) {
+            trackDatum.thresholdType = dataWithValue[0].thresholdType;
             trackDatum.category = trackDatum.profile_data && trackDatum.thresholdType? `${trackDatum.thresholdType}${trackDatum.profile_data.toFixed(2)}` : undefined;
         }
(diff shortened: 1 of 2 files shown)


src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -283,6 +283,9 @@ export function fillHeatmapTrackDatum<T extends IBaseHeatmapTrackDatum, K extend
 ) {
     trackDatum[featureKey] = featureId;
     trackDatum.study_id = case_.studyId;
+    if (data) {
```
alisman: this mutates the array of items passed in as argument which is a bit of a no-no.   you might want to use filter instead to produce a new array (elements will not be cloned)

pvannierop: Ok, code updated to use _.filter and capture of filtered array in a a new variable.

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -283,13 +283,17 @@ export function fillHeatmapTrackDatum<T extends IBaseHeatmapTrackDatum, K extend
 ) {
     trackDatum[featureKey] = featureId;
     trackDatum.study_id = case_.studyId;
-    if (!data || !data.length) {
+
+    // remove data points of which `value` is NaN
+    const dataWithValue =  _.filter(data, (d) => isNaN(d.value));
```
adamabeshouse: I'm confused, wouldn't you want to filter for !isNaN?

alisman:
@pvannierop did you figure out why test passed inspite of wrong sign on isNaN checked?

src/shared/components/oncoprint/DataUtils.ts
pvannierop: Reading your comment again, I am sure that this is also the way this method works in the current master; the `trackDatum` that is passed in is returned after modification. Maybe we have to consult with @adamabeshouse to find out whether this is a thing that we need to fix?

src/shared/components/oncoprint/DataUtils.ts
pvannierop: Eh yes. My bad. I have updated this line to state `!isNaN`.

pvannierop:
Yes, there was no screenshot test for this. Sort of treatment response tracks was not different from ther tracks at the time. Now its is since the data is not 100 the same for treatment profiles. I am currently adding a screenshot test for treament response sort.

pvannierop:
Yes, for the sort to work well the NaN data points needed to be removed. Now this is the case and the unit test for `fillHeatmapTrackDatum` were updated accordingly. I have been unable to include a screenshot test of said functionality since I am unable to fire the sort func for a heatmap track with WebdriverIO.

PR #2658, opened 2019-09-03
Added a checklist select component

onursumer:
- Added a new component called `Checklist`
- An alternative to `CheckedSelect` for sets with a few select options.

Sample usage:
mutation_mapper_right_panel_take2

# Checks
  the `-s` option)

### DIFF
--- a/src/public-lib/components/checkedSelect/Checklist.tsx
+++ b/src/public-lib/components/checkedSelect/Checklist.tsx
@@ -1,0 +1,91 @@
+import autobind from "autobind-decorator";
+import * as _ from "lodash";
+import {action, computed} from "mobx";
+import {observer} from "mobx-react";
+import * as React from "react";
+
+import {CheckBoxType, getOptionLabel, getSelectedValuesMap, Option} from "./CheckedSelectUtils";
+
+type ChecklistProps = {
+    onChange: (values: {value: string}[]) => void;
+    value: {value: string}[];
+    options: Option[];
+    checkBoxType?: CheckBoxType;
+    isDisabled?: boolean;
+    numberOfColumnsPerRow?: number;
+};
+
+function generateTableRows(selectComponents: JSX.Element[], numberOfColumnsPerRow: number = 1)
+{
+    return _.chunk(selectComponents, numberOfColumnsPerRow)
+        .map(components => <tr>{components}</tr>);
+}
+
+@observer
+export default class Checklist extends React.Component<ChecklistProps, {}>
+{
+    public static defaultProps: Partial<ChecklistProps> = {
+        isDisabled: false,
+        numberOfColumnsPerRow: 1,
+        checkBoxType: CheckBoxType.HTML
+    };
+
+    @computed
+    get selectedValues() {
+        return getSelectedValuesMap(this.props.value);
+    }
+
+    public render() {
+        return (
+            <table>
+                {
+                    generateTableRows(this.props.options.map(this.optionSelectComponent),
+                        this.props.numberOfColumnsPerRow)
+                }
+            </table>
+        );
+    }
+
+    @autobind
+    private getOptionLabel(option: Option): JSX.Element
+    {
+        return getOptionLabel(option, this.selectedValues, this.props.checkBoxType);
+    }
+
+    @autobind
+    protected optionSelectComponent(option: Option)
+    {
+        const onClick = () => this.handleSelect(option.value);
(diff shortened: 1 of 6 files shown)


src/public-lib/components/SelectHelper.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from "react";
```
adamabeshouse: this isn't really a component right? should probably put this into /checkedSelect/  and call it utils or something

src/public-lib/components/SelectHelper.tsx
```diff
@@ -0,0 +1,21 @@
...
+
+export type Option = {
+    value: string;
+    label: string | JSX.Element;
+    disabled?: boolean;
+};
+
+export function getCheckBox(option: Option,
+                            valueMap: {[key: string]: {value: string}}): string
```
adamabeshouse: I think the typing of `valueMap` is confusing - the key is named `key` but you're checking if `option.value` is a key, which makes me feel like the key should be named `value`. Since you're only checking presence, it's confusing to add the type of the values as well. 

In summary, I would call it
`selectedValues: {[ value:string ]: any}`

src/public-lib/components/customSelect/ChecklistSelect.tsx
```diff
@@ -0,0 +1,89 @@
...
+@observer
+export default class ChecklistSelect extends React.Component<CustomSelectProps, {}>
+{
+    public static defaultProps: Partial<CustomSelectProps> = {
+        isDisabled: false,
+        numberOfColumnsPerRow: 1
+    };
+
+    @computed
+    get valueMap() {
```
adamabeshouse: `selectedValues` is more descriptive

src/public-lib/components/SelectHelper.tsx
onursumer: This is a shared utility class, for both `CheckedSelect` and `ChecklistSelect`, so maybe move this under `lib` instead of components?. `SelectUtils` might be a better name than `SelectHelper`.

src/public-lib/components/SelectHelper.tsx
onursumer: hmm, how about `selectedValues: {[ optionValue:string ]: Partial<Option>}`?

src/public-lib/components/SelectHelper.tsx
adamabeshouse: maybe both can be under `checklists` or something?

src/public-lib/components/SelectHelper.tsx
adamabeshouse: yeah but you never use the value, right? so why include it in the typing. you're using it as a set to only test membership

adamabeshouse:
@onursumer ohhhhh, I see. I think just `Checklist` would be clearer

src/public-lib/components/SelectHelper.tsx
onursumer: ideally this should be a hash set, but I'll go with `any` :)

src/public-lib/components/SelectHelper.tsx
onursumer: It makes sense to put both these components and the utility file under the same folder. Maybe just keep the directory name `checkedSelect` for now? Checklist is a special kind of `CheckedSelect`. `CheckedSelect` should probably also be renamed to `DropdownCheckedSelect`, but I'll leave it as is for now.

src/public-lib/components/checkedSelect/Checklist.tsx
```diff
@@ -0,0 +1,89 @@
...
+    };
+
+    @computed
+    get selectedValues() {
+        return getSelectedValuesMap(this.props.value);
+    }
+
+    public render() {
+        return (
+            <table>
```
alisman: is there a reason to use table to lay this out instead of bootstrap form layout?

onursumer: We've been using the same structure for the original component in mutations tab, I just followed that example: https://github.com/cBioPortal/cbioportal-frontend/blob/c1388e15d02f7b4d46e5fa338100cccbb83d4250/src/shared/components/mutationTypePanel/ProteinImpactTypePanel.tsx#L44

Is it easy with bootstrap to force a `2x2` layout without line wrapping?

PR #2662, opened 2019-09-05
Update PlotsTab styling (overflow, utilities menu, header bar)

pvannierop:
# What? Why?
1. When a plot exceeded the plot in the horizontal dimension various elements were misplaced.
- Utilities menu was centered above plot and would move off-screen to the right.
- The download button would move off-screen to the right.
- The top-bar element `Queried genes are altered in 764 (72%) of queried patients/samples
` would move off-screen to the right. 

2. The CNA entry in the utilities menu was called `CNA Type`. A better name `Copy Number Alteration` was proposed in [an Issue](https://github.com/cBioPortal/cbioportal/issues/6545).

# Fix
- The utilities menu was aligned to the left.
- The `.inlineBlock` element was restyled to adhere to horz. screen width.
- The Plots section was restyled to to adhere to horz. screen width. This fixed the already implemented hortizontal scroll functionality.
- `CNA Type` wase renamed to `Copy Number Alteration` in the utilities men.

# Before
image

# After
image


pvannierop:
@alisman The overscroll-x modification should be handled by the <ScrollBar> component in PlotsTab. This component does not seem to work (also not before my changes). Can we have a look at why it is not working? Possibly we can then remove the overscroll-x styling and have a nicer solution.

src/globalStyles/global.scss
```diff
@@ -552,7 +553,8 @@ h3.hr {
 }
 
 .inlineBlock {
-  display:inline-block;
+  float: right;
+  overflow: hidden;
```
alisman: probably not a good idea to do this in a generic class like this

src/globalStyles/global.scss
```diff
@@ -359,6 +359,7 @@ th.reactable-header-sort-asc:after {
 
 #mainColumn {
   padding-top: $gutterWidth;
+  width: 100%;
```
alisman: what does this solve?  need to be careful changing template.

src/shared/components/Scrollbar/styles.module.scss
```diff
@@ -1,9 +1,9 @@
 .scrollbar {
-  width:90%;
+  width:100%;
```
alisman: can you put a comment here about why this was necessary (so i can check it in browser)

src/globalStyles/global.scss
pvannierop: This solves the mainColumn to exceed the header that contains text `ueried genes are altered in 764...` extending off-screen (see screenshot above). I have updated the scope of this change somewhat and used max-width to be more careful:
```
.contentWidth {
   #mainColumn {
    ...
    max-width: 100%;
  }
```

src/shared/components/Scrollbar/styles.module.scss
pvannierop: It just looked better.

src/globalStyles/global.scss
pvannierop: I have implemented a new class .chartWrapper used in PlotsTab that uses the overflow:hidden style property.

alisman:
@pvannierop @jjgao i confirmed the single failure is due to small clinical data different between production and develelopment db. gonna merge this   @jjgao do you think if we tested a study other than Impact, it would be more stable?

PR #2667, opened 2019-09-06
Correct tooltip position in horizontal waterfall plot

pvannierop:
# What? Why?
- The tooltips for search labels in the waterfall plot were incorrect when the plot is in horizontal orientation.
- Axis placement at bottom of the waterfall plot when in horizontal orientation was inconvenient since at bottom of the screen. This was problematic for very long plots.

# Fix
- Tooltip position was fixed.
- Horizontal axis was moved to the top of horizontal waterfall plot.

# Before
Screenshot from 2019-09-06 16-50-32

# After
Screenshot from 2019-09-06 16-49-58


src/shared/components/plots/WaterfallPlot.tsx
```diff
@@ -420,6 +420,8 @@ export default class WaterfallPlot<D extends IBaseWaterfallPlotData> extends Rea
         const dataPoints = _.filter(this.waterfallPlotData, (d:any) => this.props.highlight!(d) );
 
         const searchLabels = _.cloneDeep(dataPoints);
+        // add marker field to search label
+        _.each(searchLabels, (o) => o.search_label = true);
```
adamabeshouse: we've been using camelCase, so let's call it `searchLabel`?

pvannierop: Ok, updated to camelCase.

PR #2668, opened 2019-09-06
update screenshots after data update

alisman:


alisman:
@pvannierop @inodb i think Pim changed these tests in release-3.1.0 so they don't have data dependency.  we'll have to remember to resolve the conflict in the right direction

PR #2679, opened 2019-09-09
molecular files download

dippindots:
move data download from Query to ResultView.
Fix # https://github.com/cBioPortal/cbioportal/issues/5189.

- add one row to download data from other profiles (select the first option by default)
multiple studies:
image
single study(with tooltip):
image

- add a warning on the download tab (query page)
image

- sort the profiles by type, unit tests included (DownloadUtils.ts)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -170,6 +170,7 @@ import {DEFAULT_GENOME} from "pages/resultsView/ResultsViewPageStoreUtils";
 import { StudyWithSamples, getFilteredStudiesWithSamples, ChartMeta, getClinicalAttributeUniqueKey, getChartMetaDataType, getPriorityByClinicalAttribute, UniqueKey, getDefaultPriorityByUniqueKey } from "pages/studyView/StudyViewUtils";
 import { MUTATION_COUNT, FRACTION_GENOME_ALTERED } from "pages/studyView/StudyViewPageStore";
 import { IVirtualStudyProps } from "pages/studyView/virtualStudy/VirtualStudy";
+import { decideMolecularProfileSortingOrder } from "./download/DownloadUtils";
 
 type Optional<T> = (
     {isApplicable: true, value: T}
@@ -473,6 +474,12 @@ export class ResultsViewPageStore {
     @observable public sessionIdURL = '';
 
     @observable queryFormVisible: boolean = false;
+
+    @observable public selectedMolecularProfileNameForDownload: string | undefined = undefined;
+
+    @computed get doNonSelectedMolecularProfilesExist() {
+        return this.nonSelectedMolecularProfilesGroupByName.result && _.keys(this.nonSelectedMolecularProfilesGroupByName.result).length > 0;
+    }
 
     @observable public modifyQueryParams: ModifyQueryParams | undefined = undefined;
 
@@ -922,6 +929,46 @@ export class ResultsViewPageStore {
                         molecularDataMultipleStudyFilter:({
                             entrezGeneIds: _.map(this.genes.result,(gene:Gene)=>gene.entrezGeneId),
                             sampleMolecularIdentifiers:identifiers
+                        } as MolecularDataMultipleStudyFilter)
+                    });
+                }
+            }
+
+            return Promise.resolve([]);
+
+        }
+    });
+
+    // other molecular profiles data download needs the data from non queried molecular profiles
+    readonly nonQueriedMolecularData = remoteData<NumericGeneMolecularData[]>({
+        await: () => [
+            this.studyToDataQueryFilter,
+            this.genes,
+            this.nonSelectedMolecularProfiles,
+            this.samples
+        ],
+        invoke: () => {
+            // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
+            const profilesWithoutMutationProfile = _.filter(this.nonSelectedMolecularProfiles.result, (profile: MolecularProfile) => profile.molecularAlterationType !== AlterationTypeConstants.MUTATION_EXTENDED);
+            const genes = this.genes.result;
+
+            if (profilesWithoutMutationProfile.length && genes != undefined && genes.length) {
+
+                const profilesWithoutMutationProfileGroupByStudyId = _.groupBy(profilesWithoutMutationProfile, (profile) => profile.studyId);
+                // find samples which share studyId with profile and add identifier
+                const sampleIdentifiers : SampleMolecularIdentifier[] = this.samples.result.filter((sample)=> sample.studyId in profilesWithoutMutationProfileGroupByStudyId).reduce((acc: SampleMolecularIdentifier[], sample) => {
+                    acc = acc.concat(profilesWithoutMutationProfileGroupByStudyId[sample.studyId].map((profile) => {
+                        return { 'molecularProfileId':profile.molecularProfileId, 'sampleId':sample.sampleId } as SampleMolecularIdentifier;
+                    }));
+                    return acc;
+                },[]);
+
+                if (sampleIdentifiers.length) {
+                    return client.fetchMolecularDataInMultipleMolecularProfilesUsingPOST({
+                        projection:'DETAILED',
+                        molecularDataMultipleStudyFilter:({
+                            entrezGeneIds: _.map(this.genes.result,(gene:Gene)=>gene.entrezGeneId),
+                            sampleMolecularIdentifiers:sampleIdentifiers
                         } as MolecularDataMultipleStudyFilter)
                     });
                 }
@@ -2406,6 +2453,22 @@ export class ResultsViewPageStore {
             })
         }
     }, []);
+
+    readonly nonSelectedMolecularProfiles = remoteData<MolecularProfile[]>({
+        await:()=>[this.molecularProfilesInStudies, this.selectedMolecularProfiles],
+        invoke:() => {   
+            return Promise.resolve(_.difference(this.molecularProfilesInStudies.result!, this.selectedMolecularProfiles.result!));
+        }
+    }, []);
+
+    // If we have same profile accros multiple studies, they should have the same name, so we can group them by name to get all related molecular profiles in multiple studies.
+    readonly nonSelectedMolecularProfilesGroupByName = remoteData<{ [profileName: string]: MolecularProfile[] }>({
+        await:()=>[this.nonSelectedMolecularProfiles],
+        invoke:() => {
+            const sortedProfiles = _.sortBy(this.nonSelectedMolecularProfiles.result, (profile) => decideMolecularProfileSortingOrder(profile.molecularAlterationType));
+            return Promise.resolve(_.groupBy(sortedProfiles, (profile) => profile.name));
+        }
+    }, {});
 
     readonly molecularProfileIdToMolecularProfile = remoteData<{ [molecularProfileId: string]: MolecularProfile }>({
         await: () => [this.molecularProfilesInStudies],
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -475,8 +475,6 @@ export class ResultsViewPageStore {
 
     @observable queryFormVisible: boolean = false;
 
-    @observable public selectedMolecularProfileNameForDownload: string | undefined = undefined;
-
     @computed get doNonSelectedMolecularProfilesExist() {
         return this.nonSelectedMolecularProfilesGroupByName.result && _.keys(this.nonSelectedMolecularProfilesGroupByName.result).length > 0;
     }
--- a/src/pages/resultsView/download/DownloadTab.tsx
+++ b/src/pages/resultsView/download/DownloadTab.tsx
@@ -19,7 +19,6 @@ import {
 import styles from "./styles.module.scss";
 import classNames from 'classnames';
 import OqlStatusBanner from "../../../shared/components/banners/OqlStatusBanner";
-import WindowStore from "../../../shared/components/window/WindowStore";
 import {WindowWidthBox} from "../../../shared/components/WindowWidthBox/WindowWidthBox";
 import {remoteData} from "../../../public-lib/api/remoteData";
 import LoadingIndicator from "shared/components/loadingIndicator/LoadingIndicator";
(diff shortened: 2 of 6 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -469,6 +469,12 @@ export class ResultsViewPageStore {
 
     @observable queryFormVisible: boolean = false;
 
+    @observable public selectedMolecularProfileNameForDownload: string | undefined = undefined;
+
+    @computed get isOtherMolecularProfileExist() {
```
adamabeshouse: could you please explain what this is? I'm confused by the name

dippindots: @adamabeshouse this is an indicator for the `nonSelectedMolecularProfiles`. This means if there is a molecular profile in the study and the user doesn't select this/those profiles. So maybe should be called `isNonSelectedMolecularProfilesExist`?

adamabeshouse: ah okay, yeah I like your suggestion much better. the right grammar is `doNonSelectedMolecularProfilesExist`

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -446,6 +494,55 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+                                    autosize={false}
+                            />
+                        </div>
+                    </div>
+                </td>
+                <td>
+                    {
+                        (this.props.store.isOtherMolecularProfileExist) && (
+                            <div>
+                                <a onClick={(event) => this.handleOtherMolecularProfileDownload(event, selectedProfileName)}>
```
adamabeshouse: I'm curious why you are doing it by profile name, and not by profile id?

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: @adamabeshouse Thanks!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -886,6 +892,47 @@ export class ResultsViewPageStore {
         }
     });
 
+    readonly nonQueriedMolecularData = remoteData<NumericGeneMolecularData[]>({
```
alisman: we need a comment to explain why we need "nonQueried" data.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -886,6 +892,47 @@ export class ResultsViewPageStore {
...
+        await: () => [
+            this.studyToDataQueryFilter,
+            this.genes,
+            this.nonSelectedMolecularProfiles,
+            this.samples
+        ],
+        invoke: () => {
+            // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
+            const profilesWithoutMutationProfile = _.filter(this.nonSelectedMolecularProfiles.result, (profile: MolecularProfile) => profile.molecularAlterationType !== 'MUTATION_EXTENDED');
```
alisman: i believe there's an enum for somewhere for MUTATION_EXTENDED.  see if you can find it and use it.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -886,6 +892,47 @@ export class ResultsViewPageStore {
...
+            const genes = this.genes.result;
+
+            if (profilesWithoutMutationProfile.length && genes != undefined && genes.length) {
+
+                const identifiers : SampleMolecularIdentifier[] = [];
+
+                profilesWithoutMutationProfile.forEach((profile:MolecularProfile)=>{
```
alisman: you can do this operation more efficiently by putting all studyIds in a map and then samples.result.filter( (sample)=>sample.studyId in myMap).

alisman: remember try to avoid declaring an array and then using forEach to populate it.  usually it's more clear and efficient to use filter or reduce or map.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -886,6 +892,47 @@ export class ResultsViewPageStore {
...
+            this.samples
+        ],
+        invoke: () => {
+            // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
+            const profilesWithoutMutationProfile = _.filter(this.nonSelectedMolecularProfiles.result, (profile: MolecularProfile) => profile.molecularAlterationType !== 'MUTATION_EXTENDED');
+            const genes = this.genes.result;
+
+            if (profilesWithoutMutationProfile.length && genes != undefined && genes.length) {
+
+                const identifiers : SampleMolecularIdentifier[] = [];
```
alisman: sampleIdentifiers

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: Thanks for the advice!

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: Sure!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2337,6 +2384,25 @@ export class ResultsViewPageStore {
         }
     }, []);
 
+    readonly nonSelectedMolecularProfiles = remoteData<MolecularProfile[]>({
+        await:()=>[this.molecularProfilesInStudies, this.selectedMolecularProfiles, this.molecularProfileIdToMolecularProfile],
```
alisman: since you are trying to obtain the profiles, not the Ids, there is more efficient way to do this.  you can probably use _.difference directly on the arrays of profiles.   since the profile objects themselves are actually the same, they will have referential equality.    result = _.difference( allProfiles, selectedProfiles)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2337,6 +2384,25 @@ export class ResultsViewPageStore {
...
+            const selectedMolecularProfileIds = this.selectedMolecularProfiles.result!.map((profile)=>profile.molecularProfileId);         
+            const allMolecularProfileIds = this.molecularProfilesInStudies.result!.map((profile)=>profile.molecularProfileId);           
+            const nonSelectedMolecularProfileIds = _.difference(allMolecularProfileIds, selectedMolecularProfileIds);
+            const nonSelectedMolecularProfiles = _.map(nonSelectedMolecularProfileIds, (id) => {return this.molecularProfileIdToMolecularProfile.result[id]})
+            return Promise.resolve(nonSelectedMolecularProfiles);
+        }
+    }, []);
+
+    // If we have same profile accros multiple studies, they should have the same name, so we can group them by name to get all related molecular profiles in multiple studies.
+    readonly nonSelectedMolecularProfilesGroupByName = remoteData<_.Dictionary<MolecularProfile[]>>({
```
alisman: I'm not sure what _.Dictionary is.   Probably better just to return { [profileName:string]: whatever } which we use elsewhere.  It communicates more information about key.

dippindots: Ok!

tmazor:
Looks good! Just one major comment: the list of other profiles should be sorted in the same order as on the current download tab. For example, for lgg_tcga:
image
image

Also, the current download interface has the (?) icon to provide information about each genomic profile, and we've lost that in this new implementation. It would be nice to keep that, but not sure how much it matters.... @jjgao ?

Finally, a related thought - when we disable the Download tab on the homepage (not sure if that plan is to do that immediately? @jjgao?), rather than just removing the download tab, we should probably keep the tab and provide a message to users telling them to run a query instead.

jjgao:
> Looks good! Just one major comment: the list of other profiles should be sorted in the same order as on the current download tab. For example, for lgg_tcga:

+1

> Also, the current download interface has the (?) icon to provide information about each genomic profile, and we've lost that in this new implementation. It would be nice to keep that, but not sure how much it matters.... @jjgao ?

@dippindots is it possible to add the (?) back for single study? For multi-study, it's ok to skip. If not easy, let's log it as an issue and release the current one. The profile names are usually very descriptive already.

> Finally, a related thought - when we disable the Download tab on the homepage (not sure if that plan is to do that immediately? @jjgao?), rather than just removing the download tab, we should probably keep the tab and provide a message to users telling them to run a query instead.

Good idea. @dippindots could you put a message in the Download interface and tell people to go to the Download tab after submitting query. @tmazor I think we should remove the form. Or do you think we should keep the form for a while?

tmazor:
I agree we can remove the form - no reason to keep it once we have the same functionality elsewhere.

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -359,11 +403,12 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
                                 <tbody>
                                     {hasValidData(this.cnaData.result!) && this.cnaDownloadControls()}
                                     {hasValidMutationData(this.mutationData.result!) && this.mutationDownloadControls()}
-                                    {hasValidData(this.mrnaData.result!) && this.mrnaExprDownloadControls()}
-                                    {hasValidData(this.proteinData.result!) && this.proteinExprDownloadControls()}
+                                    {hasValidData(this.mrnaData.result!) && this.mrnaExprDownloadControls(this.props.store.selectedMolecularProfiles.result!.find((profile) => profile.molecularAlterationType === AlterationTypeConstants.MRNA_EXPRESSION)!.name)}
+                                    {hasValidData(this.proteinData.result!) && this.proteinExprDownloadControls(this.props.store.selectedMolecularProfiles.result!.find((profile) => profile.molecularAlterationType === AlterationTypeConstants.PROTEIN_LEVEL)!.name)}
```
dippindots: Once we have valid mRNA/protein data, we will only get one molecular profile in the selectedMolecularProfiles that matches the specific molecularAlterationType. We pass the profile name to the download controls.

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +457,128 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+
+
+                </td>
+            </tr>
+        );
+    }
+
+    @autobind
+    @action
+    private formatOptionLabel(props: any) {
```
dippindots: format option label (just for single study query)

src/pages/resultsView/download/DownloadUtils.ts
```diff
@@ -458,4 +476,25 @@ export function hasValidMutationData(sampleAlterationDataByGene: {[key: string]:
...
+            return 4;
+        case "METHYLATION":
+            return 5;
+        case "METHYLATION_BINARY":
+            return 6;
+        case "PROTEIN_LEVEL":
+            return 7;
+        default:
+            return Number.MAX_VALUE;
+    }
```
dippindots: sort the molecular profiles as we did in the [MolecularProfileSelector](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/shared/components/query/MolecularProfileSelector.tsx#L40-L46)

src/shared/components/query/QueryContainer.tsx
```diff
@@ -167,6 +167,11 @@ export default class QueryContainer extends React.Component<QueryContainerProps,
                 }
 
                 {
+                    this.store.forDownloadTab && 
+                    <div className={"alert alert-danger"} role="alert">The Download interface has been removed. Enhanced download functionality is now available through the Query interface. Run a query with your gene(s) of interest and use the Download tab to download all available data types.</div>
+                }
```
dippindots: show a warning temporarily.

alisman:
@jjgao @tmazor sorry to chime in at late date on this one but i think the dropdown for "other profiles" is very confusing.   i would not guess that it is simply a way to access more items that are siblings of those in the list above.     i think a more standard approach would be to have a button below the table reading "show more", which would just show the rest of the table rows.  would that work?

jjgao:
`show more` sounds like a great idea 👍

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +477,113 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+                            <a onClick={(event) => this.handleTransposedOtherMolecularProfileDownload(event, option.label)}>
+                                <i className='fa fa-cloud-download' style={{marginRight: 5}}/>Transposed Matrix
+                            </a>
+                        </div>
+                    </td>
+                </tr>
+            )
+        )
+        
+        components.push(
```
alisman: this should be conditional

alisman: and this should be outside of table

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +477,113 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+                </div>
+            )
+        )
+        return components;
+    }
+
+    @autobind
+    @action
+    private toggleShowMore() {
+        this.isShowMoreClicked = !this.isShowMoreClicked;
```
alisman: showAllRows

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +477,113 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+                return {
+                    "value": name,
+                    "label": name,
+                    "description": singleStudyProfile.description
+                }
+            }
+            return {"value": name, "label": name}
+        });
+
+        const filteredProfileOptions = allProfileOptions.slice(0, Math.min(7 - this.props.store.selectedMolecularProfiles.result!.length, allProfileOptions.length));
```
alisman: make the limit a constant

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +477,113 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
+                        <div style={{display: "flex", alignItems:"center"}}>
+                            {option.label}
+                            {
+                                (option.description) && (
+                                        <DefaultTooltip
+                                            mouseEnterDelay={0}
+                                            placement="right"
+                                            overlay={<div className={styles.tooltip}>{option.description}</div>}
+                                        >
+                                            <FontAwesome className={styles.infoIcon} name='exclamation-circle'/>
```
alisman: info not exclamation

src/pages/resultsView/download/DownloadTab.tsx
```diff
@@ -412,40 +477,113 @@ export default class DownloadTab extends React.Component<IDownloadTabProps, {}>
...
                         <i className='fa fa-cloud-download' style={{marginRight: 5}}/>Transposed Matrix
                     </a>
                 </td>
             </tr>
         );
     }
 
+    private nonSelectedProfileDownloadRow(nonSelectedMolecularProfilesGroupByName:_.Dictionary<MolecularProfile[]>)
+    {
+        const allProfileOptions = _.keys(nonSelectedMolecularProfilesGroupByName).map((name) => {
```
alisman: use forEach and it will give you both key and value in callback

alisman: actually, use _.map

dippindots:
@alisman File validation:
**1. single study:**
search `lgg_tcga` on download tab in the query page, select `Brain Lower Grade Glioma (TCGA, Provisional)
`
image

download the `transposed file` and compare with the file from the new [download tab](https://cbioportal-frontend-pr-2679.herokuapp.com/results/download?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=lgg_tcga&case_set_id=lgg_tcga_cnaseq&data_priority=0&gene_list=TP53&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=lgg_tcga_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=lgg_tcga_mutations&tab_index=tab_visualize).

**2. multiple studies**
search `cellline_nci60`  on download tab in the query page, select `NCI-60 Cell Lines (NCI, Cancer Res 2012)

image

download the `transposed file` of `Relative linear copy-number values` and compare with the file from the new [download tab](https://cbioportal-frontend-pr-2679.herokuapp.com/results/download?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=cellline_nci60%2Clgg_tcga&case_set_id=all&data_priority=0&gene_list=TP53&geneset_list=%20&tab_index=tab_visualize).

PR #2681, opened 2019-09-10
Treatment selection menu init from `treatment_list` url param

pvannierop:
# What? Why?
Oncoprint component allows selection of treatments via the heatmap menu. This sets a URL param `treatment_list`. The init of the heatmap menu selection component from the URL param (so, the other way around) is not implemented yet.

# Fix
The current PR implements pre-selection of treatments in the treatment selected menu from the `treatment_list` URL param.


src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -208,6 +211,18 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
         this.onHorzZoomSliderSet = this.onHorzZoomSliderSet.bind(this);
         this.onSetHorzZoomTextInput = this.onSetHorzZoomTextInput.bind(this);
 
+        // add pre-selected treatments (passed in via URL in ResultsViewOncoprint) to selected treatments
+        this.initSelectedTreatments = autorun(() => {
```
alisman: i think we should be able to do this without autorun.    Isn't _selectedTreatmentOptionsObsArray just a computation of  treatmentOptionsByValueMap and this.props.state?

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -208,6 +211,18 @@ export default class OncoprintControls extends React.Component<IOncoprintControl
...
+            const selectionMap = this.treatmentOptionsByValueMap;
+            if (_.keys(selectionMap).length > 0) {
+                const initTreatmentOpts = this.props.state.selectedTreatmentsInit ?
+                    this.props.state.selectedTreatmentsInit.map((treatmentId) => selectionMap[treatmentId])
```
alisman: minor but you might want to reference the props which passed to contructor here instead of this.props, which i think wouldn't exist were this callback to fire immediately (syncronously).   props is available here via closure

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: I tried this at first, but any solution that used @computed does not allow the _de_selection of treatments. The autorun solution correctly handles init of the selected treatments.

pvannierop: Autorun has been removed. Treatment selections and options are passed in as props.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
pvannierop: Updated

PR #2685, opened 2019-09-11
Implement gene panel API calls in PatientViewStore

pvannierop:
# What? Why?
As defined in RFC45 information on gene panels will be integrated in the PatientView component. This component does not yet retrieve information on gene panels for the patient samples from the backend.

#  Fix
This PR adds API calls to:
1. get the gene panels associated with each sample.
2. get genes associated with the gene panels retrieved in 1.

Note: at present the backend returns no gene panel id for a sample when a samples was analyzed with a whole genome approach. The implementation already accommodates an alternative strategy proposed by @n1zea144 involving `WES` and `WGS` tags for whole exome sequencing and whole genome sequencing profiles.

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -2,7 +2,7 @@ import * as _ from 'lodash';
 import {ClinicalDataBySampleId} from "../../../shared/api/api-types-extended";
 import {
     ClinicalData, MolecularProfile, Sample, Mutation, DiscreteCopyNumberFilter, DiscreteCopyNumberData, MutationFilter,
-    CopyNumberCount, ClinicalDataMultiStudyFilter, ReferenceGenomeGene
+    CopyNumberCount, ClinicalDataMultiStudyFilter, ReferenceGenomeGene, GenePanelData, GenePanel
 } from "../../../shared/api/generated/CBioPortalAPI";
 import client from "../../../shared/api/cbioportalClientInstance";
 import internalClient from "../../../shared/api/cbioportalInternalClientInstance";
@@ -69,7 +69,10 @@ import {
     concatMutationData,
     fetchOncoKbCancerGenes,
     fetchVariantAnnotationsIndexedByGenomicLocation,
-    fetchReferenceGenomeGenes
+    fetchReferenceGenomeGenes,
+    fetchGenePanelData,
+    fetchGenePanel,
+    noGenePanelUsed
 } from "shared/lib/StoreUtils";
 import {fetchHotspotsData} from "shared/lib/CancerHotspotsUtils";
 import {stringListToSet} from "../../../public-lib/lib/StringUtils";
@@ -145,6 +148,7 @@ function transformClinicalInformationToStoreShape(patientId: string, studyId: st
 }
 
 export class PatientViewPageStore {
+
     constructor() {
         labelMobxPromises(this);
         this.internalClient = internalClient;
@@ -762,6 +766,74 @@ export class PatientViewPageStore {
         ],
         invoke: ()=>Promise.resolve(indexHotspotsData(this.hotspotData))
     });
+
+    readonly sampleToMutationGenePanelData = remoteData<{[sampleId: string]: GenePanelData}>({
+        await:()=>[
+            this.mutationMolecularProfileId
+        ],
+        invoke: async() => {
+            if (this.mutationMolecularProfileId.result) {
+                return fetchGenePanelData(this.mutationMolecularProfileId.result, this.sampleIds);
+            }
+            return {};
+        }
+    }, {});
+
+    readonly sampleToMutationGenePanelId = remoteData<{[sampleId: string]: string}>({
+        await:()=>[
+            this.sampleToMutationGenePanelData
+        ],
+        invoke: async() => {
+            return _.mapValues(this.sampleToMutationGenePanelData.result, (genePanelData) => genePanelData.genePanelId);
+        }
+    }, {});
+
+    readonly sampleToDiscreteGenePanelData = remoteData<{[sampleId: string]: GenePanelData}>({
+        await:()=>[
+            this.molecularProfileIdDiscrete
+        ],
+        invoke: async() => {
+            if (this.molecularProfileIdDiscrete.result) {
+                return fetchGenePanelData(this.molecularProfileIdDiscrete.result, this.sampleIds);
+            }
+            return {};
+        }
+    }, {});
+
+    readonly sampleToDiscreteGenePanelId = remoteData<{[sampleId: string]: string}>({
+        await:()=>[
+            this.sampleToDiscreteGenePanelData
+        ],
+        invoke: async() => {
+            return _.mapValues(this.sampleToDiscreteGenePanelData.result, (genePanelData) => genePanelData.genePanelId);
+        }
+    }, {});
+
+    readonly genePanelIdToPanel = remoteData<{[genePanelId: string]: GenePanel}>({
+        await:()=>[
+            this.sampleToMutationGenePanelData,
+            this.sampleToDiscreteGenePanelData
+        ],
+        invoke: async() => {
+            const sampleGenePanelInfo = _.concat(_.values(this.sampleToMutationGenePanelData.result), _.values(this.sampleToDiscreteGenePanelData.result));
+            const panelIds = _(sampleGenePanelInfo)
+                .map((genePanelData) => genePanelData.genePanelId)
+                .filter((genePanelId) => !noGenePanelUsed(genePanelId))
+                .value();
+            return fetchGenePanel(panelIds);
+        }
+    }, {});
+
+    readonly genePanelIdToEntrezGeneIds = remoteData<{[genePanelId: string]: number[]}>({
+        await:()=>[
+            this.genePanelIdToPanel
+        ],
+        invoke: async() => {
+            return _(this.genePanelIdToPanel.result)
+            .mapValues((genePanel) => _.map(genePanel.genes, (genePanelToGene) => genePanelToGene.entrezGeneId))
+            .value();
+        }
+    }, {});
 
     @computed get mergedMutationData(): Mutation[][] {
         return mergeMutations(this.mutationData);
--- a/src/shared/lib/StoreUtils.ts
+++ b/src/shared/lib/StoreUtils.ts
@@ -3,7 +3,7 @@ import $ from 'jquery';
 import {
     fetchVariantAnnotationsByMutation as fetchDefaultVariantAnnotationsByMutation,
     fetchVariantAnnotationsIndexedByGenomicLocation as fetchDefaultVariantAnnotationsIndexedByGenomicLocation,
-} from "react-mutation-mapper"
+} from "react-mutation-mapper";
 import {
     default as CBioPortalAPI,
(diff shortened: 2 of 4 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -55,12 +57,21 @@ import {AlterationTypeConstants} from "../../pages/resultsView/ResultsViewPageSt
...
 import {GeneticTrackDatum_Data} from "../components/oncoprint/Oncoprint";
 import {normalizeMutations} from "../components/mutationMapper/MutationMapperUtils";
+import client from 'shared/api/cbioportalClientInstance';
 
 export const ONCOKB_DEFAULT: IOncoKbData = {
     uniqueSampleKeyToTumorType : {},
     indicatorMap : {}
 };
 
+export const CompleteProfileTypeSignature = {
```
adamabeshouse: can you explain more what this is/how it relates to the data?

adamabeshouse: are these now special values that will come back from the API?

src/shared/lib/StoreUtils.ts
```diff
@@ -55,12 +57,21 @@ import {AlterationTypeConstants} from "../../pages/resultsView/ResultsViewPageSt
...
+    UNKNOWN: undefined,
+    WHOLE_EXOME_SEQ: 'WES',
+    WHOLE_GENOME_SEQ: 'WGS'
+};
+
+export const noGenePanelUsed:(o:string|undefined) => boolean = (o) => o === CompleteProfileTypeSignature.UNKNOWN || o === CompleteProfileTypeSignature.WHOLE_EXOME_SEQ || o === CompleteProfileTypeSignature.WHOLE_GENOME_SEQ;
```
adamabeshouse: A few things about this

- I think just defining a function is easier to read than defining a constant with a function type. So just `export function noGenePanelUsed(...):boolean {` (the return type is probably unnecessary IMO since it's easy to see by inspection, but up to you)
- It would help a lot in understanding what this is to name the parameter something more descriptive than `o` - `genePanelId` I think?

src/shared/lib/StoreUtils.ts
```diff
@@ -528,6 +539,27 @@ export async function fetchCopyNumberData(discreteCNAData:MobxPromise<DiscreteCo
...
+export async function fetchSampleGenePanelData(molecularProfileId:string|undefined, sampleIds:string[]):Promise<{[sampleId: string]: string|undefined}> {
+    const genePanelDataFilter = {sampleIds} as GenePanelDataFilter;
+    if (molecularProfileId) {
+        const remoteData = await client.getGenePanelDataUsingPOST({molecularProfileId, genePanelDataFilter});
+        return _(remoteData).keyBy('sampleId').mapValues('genePanelId').value();
+    }
+    return {};
+}
+
+export async function fetchGenePanelGeneData(genePanelIds:(string|undefined)[]):Promise<{[genePanelId:string]: number[]}> {
```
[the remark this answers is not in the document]
pvannierop: Removed.

src/shared/lib/StoreUtils.ts
```diff
@@ -528,6 +539,27 @@ export async function fetchCopyNumberData(discreteCNAData:MobxPromise<DiscreteCo
...
+    let genePanelId2Genes = {} as {[genePanelId:string]: number[]};
+    for (const genePanelId of genePanelIds) {
+        if (genePanelId) {
+            const genePanel = await client.getGenePanelUsingGET({genePanelId});
```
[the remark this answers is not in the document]
pvannierop: Ah I did not know about that. Code was updated as per your suggestion.

src/shared/lib/StoreUtils.ts
adamabeshouse: got it. is it possible to encode that in swagger?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -745,6 +749,33 @@ export class PatientViewPageStore {
         invoke: ()=>Promise.resolve(indexHotspotsData(this.hotspotData))
     });
 
+    readonly mutationSampleToGenePanelId = remoteData<{[sampleId: string]: string|undefined}>({
```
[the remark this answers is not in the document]
adamabeshouse: verbally I think it makes more sense to do `sampleToMutationGenePanelId` and `sampleToDiscreteGenePanelId`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -745,6 +749,33 @@ export class PatientViewPageStore {
...
+    }, {});
+
+    readonly discreteSampleToGenePanelId = remoteData<{[sampleId: string]: string|undefined}>({
+        await:()=>[
+            this.molecularProfileIdDiscrete
+        ],
+        invoke: async() => fetchSampleGenePanelData(this.molecularProfileIdDiscrete.result, this.sampleIds)
+    });
+
+    readonly genePanelIdToGene = remoteData<{[genePanelId: string]: number[]}>({
```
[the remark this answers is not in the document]
adamabeshouse: `ToEntrezGeneIds`?

src/shared/lib/StoreUtils.ts
pvannierop: > also, could this be a ts Enum?

No, since the types that signify a whole genome approach can be string or _undefined_ a TS enum is not possible. I tried this :)

> got it. is it possible to encode that enum in swagger?

I do not understand this question. You mean whether the allowed variables could be made explicit in the swagger documentation of the backend?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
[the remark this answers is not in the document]
pvannierop: I agree. Names updated.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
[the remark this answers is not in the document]
pvannierop: Yes, name updated.

src/shared/lib/StoreUtils.ts
adamabeshouse: > You mean whether the allowed variables could be made explicit in the swagger documentation of the backend?

Yes, exactly

pvannierop: I guess we can, but I do not know how. This is autogenerated from the Java code of the REST controllers isn't it? This feature is not yet implemented there. So, should I better remove this evaluation using unimplemented backend functionality? I am confused.

adamabeshouse: oh I see. yeah, lets just do it in the frontend, and I think a string-valued enum as @alisman suggested would be perfect. And I think a different name would make it clearer.

```
export enum GenePanelIdSpecialValue {
    WHOLE_EXOME_SEQ="WES",
    ...
}
```

thoughts?

pvannierop: @adamabeshouse 
> I think a string-valued enum as @alisman suggested would be perfect

No, since the types that signify a whole genome approach can be **string** or **undefined** a TS enum is not possible. I tried this :) The values must be of the same type...

adamabeshouse: ahhh okay. then your approach sounds good. I still think the name should be changed, though

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -745,6 +749,59 @@ export class PatientViewPageStore {
...
+            const sampleGenePanelInfo = _.concat(_.values(this.sampleToMutationnGenePanelData.result), _.values(this.sampleToDiscreteGenePanelData.result));
+            const panelIds = _(sampleGenePanelInfo)
+                .map((genePanelData) => genePanelData.genePanelId)
+                .filter((genePanelId) => !noGenePanelUsed(genePanelId))
+                .value();
+            return fetchGenePanel(panelIds);
+        }
+    }, {});
+
+    @computed get genePanelIdToEntrezGeneIds() {
```
adamabeshouse: I was about to comment that I think this should be a remoteData, but then it seems like this isn't used to maybe can delete it?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -745,6 +749,59 @@ export class PatientViewPageStore {
         invoke: ()=>Promise.resolve(indexHotspotsData(this.hotspotData))
     });
 
+    readonly sampleToMutationnGenePanelData = remoteData<{[sampleId: string]: GenePanelData}>({
```
adamabeshouse: `Mutation`(one n)

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
pvannierop: @adamabeshouse 
Well, it will be used by the PRs of RFC45 that are based on this PR. It would have been better to include this line in those PRs, but it would for now be more efficient to leave it in. Is that ok with you?

Also it is not a `remoteData` since @computed from `this.genePanelIdToPanel` which is a remoteData. Or am I mistaken here?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
pvannierop: I corrected the name. Well spotted.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: since genePanelIdToPanel is remotedata, this won't be ready until that's ready. so users might want to await this as well, and only show when it's ready, for example. what do you think?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: it's looking the same right now, I guess you have it changed on local?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
pvannierop: Ah I agree. I have updated the code to the remoteData version of it.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
pvannierop: Indeed, change was pushed and now it shows.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: the difference is that on a remoteData you can check the status, so your views can decide what to show in case the data is still loading. you can distinguish the cases of (1) empty data (2) still loading, whereas with @computed you cant

pvannierop: Ah I see. Thnx for clearing that up.

src/shared/lib/StoreUtils.ts
pvannierop: Good suggestion. Variable name was updated.

Sjoerd-van-Hagen:
@alisman can you re-review this so we can get it merged soon?

PR #2694, opened 2019-09-13
Fix reseting table page to defaults

kalletlak:


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -536,9 +536,12 @@ export class LazyMobXTableStore<T> {
         } else {
             this.dataStore = new SimpleLazyMobXTableApplicationDataStore<T>(props.data || []);
         }
-
-        this.dataStore.page = 0;
-        this.itemsPerPage = props.initialItemsPerPage || 50;
+        if(this.dataStore.page === undefined) {
```
alisman: @onursumer @kalletlak @adamabeshouse should we do this in contructor of SimpleLazyMobXTableApplicationDataStore?

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -536,9 +536,12 @@ export class LazyMobXTableStore<T> {
...
+            this.dataStore.page = 0;
+        }
+        if(this.itemsPerPage === undefined) {
+            this.itemsPerPage = props.initialItemsPerPage || 50;
```
alisman: defaultProps?

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
kalletlak: `initialItemsPerPage` is a parameter of `LazyMobXTable` component. I would prefer to leave as is for now.

PR #2696, opened 2019-09-14
remove unused ucsc.build prop

inodb:
See: https://github.com/cBioPortal/cbioportal/pull/6602#issuecomment-531466494

### DIFF
--- a/end-to-end-test/local/runtime-config/portal.properties
+++ b/end-to-end-test/local/runtime-config/portal.properties
@@ -3,4 +3,3 @@ db.password=cbio_pass
 db.portal_db_name=endtoend_local_cbiodb
 db.connection_string=jdbc:mysql://cbiodb-endtoend:3306/
 db.host=cbiodb-endtoend
-ucsc.build=hg19


pvannierop:
@inodb code change is ok, but e2e-localdb tests appear to fail a the stage of loading the studies into the database. There are errors related to a CacheManager:
>org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'ehCacheStatistics' defined in URL [jar:file:/cbioportal/scripts/target/scripts-3.0.7-152-gc05ea9930-dirty-SNAPSHOT.jar!/org/cbioportal/persistence/util/EhCacheStatistics.class]: Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'javax.cache.CacheManager' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}

Is this perhaps related to the new EHcache used by the backend? The e2e-localdb tests pull an image from DuckerHub. Do you see posibilities for investigating the cause of this problem?

PR #2699, opened 2019-09-16
Do not render TrialMatch tab on patient view (hiding not enough)

alisman:
Do not output TrialMatch tab on patient view unless localStorage flag is true.  We were calling endpoint even when tab was hidden.

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -545,12 +545,17 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                     </MSKTab>
                     )}
 
-                    <MSKTab key={7} id="trialMatchTab" linkText="Matched Trials"
-                            hide={!this.shouldShowTrialMatch(patientViewPageStore)}>
-                        <TrialMatchTable sampleManager={sampleManager}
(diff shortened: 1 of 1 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -545,12 +545,17 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
                     )}
 
-                    <MSKTab key={7} id="trialMatchTab" linkText="Matched Trials"
-                            hide={!this.shouldShowTrialMatch(patientViewPageStore)}>
-                        <TrialMatchTable sampleManager={sampleManager}
-                                         detailedTrialMatches={patientViewPageStore.detailedTrialMatches.result}
-                                         containerWidth={WindowStore.size.width-20}/>
-                    </MSKTab>
+                    {
+                        this.shouldShowTrialMatch(patientViewPageStore) && (
```
alisman: @victoria34 i should have caught this earlier but the way mobx promise works, the promise gets called even if component is hidden.  This change ensures that it is not invoked unless shouldShow is true

victoria34: @alisman Sorry, I didn't notice that. Will be careful next time. Thanks for letting me know.

PR #2702, opened 2019-09-16
remove common-dist

inodb:
somehow common-dist ended up checked into git


.gitignore
```diff
@@ -11,7 +11,7 @@ stats.json
 .DS_Store
 tsDist/
 dist/
-commons-dist/
```
onursumer: `commons-dist` is the build directory for `cbioportal-frontend-commons`, we should be still ignoring it

inodb: fixed

PR #2707, opened 2019-09-17
URLWrapper class approach to isolating changes in URL query

adamabeshouse:


src/shared/lib/URLWrapper.ts
```diff
@@ -0,0 +1,51 @@
...
+        protected routing:ExtendedRouterStore,
+        protected properties:Property<QueryParamsType>[]
+    ) {
+        const initValues:Partial<QueryParamsType> = {};
+        for (const property of properties) {
+            initValues[property.name] = undefined;
+        }
+        this.query = observable<Partial<QueryParamsType>>(initValues);
+
+        intercept(this.query, change=>{
```
alisman: interesting.  didn't realize you could set one for all properties.  and the callback fires for each individual property/

adamabeshouse: yeah!

src/shared/lib/URLWrapper.ts
```diff
@@ -0,0 +1,51 @@
...
+        });
+        autorun(()=>{
+            const query = routing.query as QueryParamsType;
+            for (const property of properties) {
+                this.query[property.name] = query[property.name];
+            }
+        });
+    }
+
+    public updateQuery(query:Partial<QueryParamsType>) {
```
alisman: what's thinking of putting this here?  i mean as opposed to just letting code use router directly.    updateRoute has optional params to clear state btw.  if we decide to keep this maybe we need that here too?

src/pages/groupComparison/GroupComparisonURLWrapper.ts
```diff
@@ -0,0 +1,33 @@
...
+    sessionId: string;
+    groupOrder?:string; // json stringified array of names
+    unselectedGroups?:string; // json stringified array of names
+    overlapStrategy?:OverlapStrategy;
+};
+
+export default class GroupComparisonURLWrapper extends URLWrapper<GroupComparisonURLQuery> {
+    constructor(routing:ExtendedRouterStore) {
+        super(routing, [
+            { name:"sessionId", persisted: false },
```
alisman: I wonder if "sessionProp" would be more descriptive?   persisted is kind of misleading because we DO want to persist all params, just not in a remote session.

src/shared/lib/URLWrapper.ts
adamabeshouse: my thinking was to put getting and setting in the same place - I thought it might be a bit confusing for a user to get using wrapper[param] and then set using routing.updateRoute. Wouldnt be opposed to adding those optional params here.

src/pages/groupComparison/GroupComparisonURLWrapper.ts
```diff
@@ -0,0 +1,33 @@
...
+            { name:"groupOrder", persisted: false },
+            { name:"unselectedGroups", persisted: false },
+            { name:"overlapStrategy", persisted: false }
```
alisman: are these typesafe?  I mean, will ts complain if these are not represented in GroupComparisonURLQuery?

adamabeshouse: yeah - setting the generic parameter with `URLWrapper<GroupComparisonURLQuery>` means that `name` must be `keyof GroupComparisonURLQuery` (see `URLWrapper.ts::Property`

src/pages/groupComparison/GroupComparisonURLWrapper.ts
adamabeshouse: `sessionProp` sounds pretty good to me

src/pages/groupComparison/GroupComparisonURLWrapper.ts
```diff
@@ -0,0 +1,35 @@
...
+        ]);
+    }
+
+    @computed public get tabId() {
+        return getTabId(this.pathName) || GroupComparisonTab.OVERLAP;
+    }
+
+    @autobind
+    public setTabId(tabId:GroupComparisonTab, replace?:boolean) {
+        this.routing.updateRoute({}, `comparison/${tabId}`, false, replace);
```
alisman: should this use it's own updateRoute method?  not too big a deal.

adamabeshouse: I think we can leave it like this for now, I don't think we need to fold anything into it thats not exposed to callers

adamabeshouse: I merged it, if you want me to make that change I can do it NP. I just think its best to watch the scope creep of URLWrapper - I'm envisioning it foremost as a URL interface thats more friendly than routing

PR #2715, opened 2019-09-18
Make first action button and related text render early in study view

Luke-Sikina:
# What? Why?
Fix # 6607.

- Make parts of the action button that don't need to wait render early

# Checks

# Any screenshots or GIFs?
new-improved

### DIFF
--- a/src/pages/studyView/studyPageHeader/ActionButtons.tsx
+++ b/src/pages/studyView/studyPageHeader/ActionButtons.tsx
@@ -15,6 +15,7 @@ import {AppStore} from "../../../AppStore";
 import {serializeEvent} from "../../../shared/lib/tracking";
 
 export interface ActionButtonsProps {
+    loadingComplete: boolean;
     store: StudyViewPageStore;
     appStore: AppStore;
 }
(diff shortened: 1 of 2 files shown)


alisman:
@Luke-Sikina instead of hiding the 2 (of three) buttons until isLoading is false, I think we should disable them with disabled attribute.   That way, there's no jumping in no possibility of inadvertent clicking.  Unfortunately, disabled attribute to the button doesn't functionally disable the mouseover and click events, so we have to do that in the handlers (check for the disabled status).

src/pages/studyView/studyPageHeader/ActionButtons.tsx
```diff
@@ -64,8 +65,22 @@ export default class ActionButtons extends React.Component<ActionButtonsProps, {
         return 'Download clinical data for the selected cases';
     }
 
+    get virtualStudy(): JSX.Element | null {
```
alisman: you should make this @computed so that it doesn't recompute unless it's observable dependecies have actually changed

src/pages/studyView/studyPageHeader/ActionButtons.tsx
```diff
@@ -15,6 +15,7 @@ import {AppStore} from "../../../AppStore";
 import {serializeEvent} from "../../../shared/lib/tracking";
 
 export interface ActionButtonsProps {
+    loaded: boolean;
```
alisman: lets make this loadingComplete

Luke-Sikina: Sure. Do you have any particular naming pattern you're following here? It makes no difference to me, so if there is one, I'm happy to follow it if you point me in the right direction.

alisman: Well, the mobx promise status flags (isComplete, isPending, isError) would be a good place to start.  So, it's not strict, but I think including "complete" is good.  So, I say "loadingComplete"

PR #2720, opened 2019-09-19
Use GeneSelectionBox on Query View

Luke-Sikina:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/5378

- Replaced raw text area and logic with GeneSelectionBox
- Kept the GeneSetSelector because it has other functionality
- Added an optional text field to GeneSelectionBox to change the prompt
- Deleted commented out code

# Checks

# Any screenshots or GIFs?
No. Visually, there are no changes.

### DIFF
--- a/src/shared/components/query/GeneSetSelector.tsx
+++ b/src/shared/components/query/GeneSetSelector.tsx
@@ -3,18 +3,22 @@ import * as styles_any from './styles/styles.module.scss';
 import {Modal} from 'react-bootstrap';
 import ReactSelect from 'react-select1';
 import {observer} from "mobx-react";
-import {computed} from 'mobx';
+import {computed, action} from 'mobx';
 import {FlexRow, FlexCol} from "../flexbox/FlexBox";
 import gene_lists from './gene_lists';
-import GeneSymbolValidator from "./GeneSymbolValidator";
 import classNames from 'classnames';
 import {getOncoQueryDocUrl} from "../../api/urls";
-import {QueryStoreComponent} from "./QueryStore";
+import {QueryStoreComponent, Focus, GeneReplacement} from "./QueryStore";
 import MutSigGeneSelector from "./MutSigGeneSelector";
 import GisticGeneSelector from "./GisticGeneSelector";
 import SectionHeader from "../sectionHeader/SectionHeader";
 import AppConfig from "appConfig";
 import {ServerConfigHelpers} from "../../../config/config";
+import OQLTextArea, { GeneBoxType } from '../GeneSelectionBox/OQLTextArea';
+import { SingleGeneQuery } from 'shared/lib/oql/oql-parser';
+import { Gene } from 'shared/api/generated/CBioPortalAPI';
+import {bind} from "bind-decorator";
+import GenesetsValidator from './GenesetsValidator';
 
 const styles = styles_any as {
 	GeneSetSelector: string,
@@ -60,18 +64,17 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
 	    ];
 	}
 
-	@computed get textAreaRef()
-	{
-		if (this.store.geneQueryErrorDisplayStatus === 'shouldFocus')
-			return (textArea:HTMLTextAreaElement) => {
-				let {error} = this.store.oql;
-				if (textArea && error)
-				{
-					textArea.focus();
-					textArea.setSelectionRange(error.start, error.end);
-					this.store.geneQueryErrorDisplayStatus = 'focused';
-				}
-			};
+	@bind
+	@action
+	handleOQLUpdate(
+		oql: {query: SingleGeneQuery[], error?: {start: number, end: number, message: string}},
+		genes: {found: Gene[], suggestions: GeneReplacement[]},
+		queryStr: string,
+	): void {
+		if (queryStr !== this.store.geneQuery) {
+			this.store.geneQuery = queryStr;
+			this.store.oql.error = oql.error;
+		}
 	}
 
 	render()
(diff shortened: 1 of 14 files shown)


src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -94,36 +94,14 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
...
-					ref={this.textAreaRef}
-					className={classNames(styles.geneSet, this.store.geneQuery ? styles.notEmpty : styles.empty)}
-					rows={5}
-					cols={80}
-					placeholder="Enter HUGO Gene Symbols, Gene Aliases, or OQL"
-					title="Enter HUGO Gene Symbols, Gene Aliases, or OQL"
-					value={this.store.geneQuery}
-					onChange={event => this.store.geneQuery = event.currentTarget.value}
-					data-test='geneSet'
+				<GeneSelectionBox // misleading: this is where you type genes / oql. Select is the ReactSelect
```
alisman: Good point.  Lets change it!  OQLInput? OQLTextArea?

Luke-Sikina: OQLTextArea works for me

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -60,18 +64,14 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
...
-					this.store.geneQueryErrorDisplayStatus = 'focused';
-				}
-			};
+	@autobind
+	geneInputUpdate(
+		oql: {query: SingleGeneQuery[], error?: {start: number, end: number, message: string}},
+		genes: {found: Gene[], suggestions: GeneReplacement[]},
+		queryStr: string,
+	): void {
+		this.store.geneQuery = queryStr;
```
alisman: Whenever we mutate mobx observables, we need to do so inside an "action."   Doing so allows for batch updates.  Setting multiple values (as you are doing here) will only trigger one update "signal" to observers when all changes are made.  Even when we're only making a single change, we should do it inside an action because it can help for debugging. 
You can do this here by adding @action decorator on this method.

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -60,18 +64,14 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
...
-				let {error} = this.store.oql;
-				if (textArea && error)
-				{
-					textArea.focus();
-					textArea.setSelectionRange(error.start, error.end);
-					this.store.geneQueryErrorDisplayStatus = 'focused';
-				}
-			};
+	@autobind
+	geneInputUpdate(
```
alisman: lets call this "handleOQLChange"

src/shared/components/query/GeneSetSelector.tsx
alisman: FYI, this rule was not obeyed in the previous code here

jjgao:
@Luke-Sikina fyi - I edit the description of this pr to link the issue in the main repo.

Luke-Sikina:
Block from merging pending feedback from product team.

alisman:
I think we should rebase instead of merge changes from master to branch.

alisman:
Looks like it isn't quite fully hooked up with query store because when you have validated genes, you still can't submit.   (also, when genes are invalid, doesn't say "edit genes" at bottom.

src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -60,18 +62,14 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}>
...
-			return (textArea:HTMLTextAreaElement) => {
-				let {error} = this.store.oql;
-				if (textArea && error)
-				{
-					textArea.focus();
-					textArea.setSelectionRange(error.start, error.end);
-					this.store.geneQueryErrorDisplayStatus = 'focused';
-				}
-			};
+	@action
```
alisman: this method also needs to be @bind .  this could explain failure to update store.   yes, i see it doesn't have context (this).

Luke-Sikina:
> I think we should rebase instead of merge changes from master to branch.

Yeah, I had no clue that the `Update branch` button didn't rebase. I feel betrayed.

alisman:
Just found one more problem with this one.  There is some behavior that was changed in the old feature but not in GeneSelectionBox: when you type a color after gene, EGFR: 
It will show a validation error AND highlight the colon.   A colon is actually legal if followed by a driver modifier (e.g. EGFR: MUT).    Look how it's handled on live homepage:   Invalid message will show but no highlighting.  User is now able to complete their typing.

src/shared/components/query/QueryStore.ts
```diff
@@ -1,1734 +1,1740 @@
-/* tslint:disable: indent linebreak-style */
-import * as _ from 'lodash';
```
Luke-Sikina: Woah. I'm going to go check why this diff is so insane.

Luke-Sikina: Ah, line endings

Luke-Sikina: Ok, I don't know what to do about this. I changed my `core.autocrlf` settings to `false` both globally and locally and this file is still getting nuked, presumably by line endings.

src/shared/components/query/QueryStore.ts
```diff
@@ -1,1734 +1,1740 @@
...
-            params.genetic_profile_ids_PROFILE_MUTATION_EXTENDED,
-            params.genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION,
-            params.genetic_profile_ids_PROFILE_MRNA_EXPRESSION,
-            params.genetic_profile_ids_PROFILE_METHYLATION,
-            params.genetic_profile_ids_PROFILE_PROTEIN_EXPRESSION,
-            params.genetic_profile_ids_PROFILE_GENESET_SCORE,
-            params.genetic_profile_ids_PROFILE_GENERIC_ASSAY
-        ];
-
-        let queriedStudies = params.cancer_study_list ? params.cancer_study_list.split(",") : (params.
```
Luke-Sikina: Here's 1 change

Luke-Sikina: These were both changed to enums

Luke-Sikina: Our style guide says no underscore prefixes to indicate intended scope, so I renamed this from `_genesetQuery`

PR #2721, opened 2019-09-19
Add button to reset to default view

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6543

Screen Shot 2019-09-19 at 3 46 34 PM


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -329,8 +330,31 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
...
+                                        <Modal bsSize={'small'} show={this.showModal} onHide={() => { this.showModal = false; }} keyboard>
+                                            <Modal.Header closeButton>
+                                                <Modal.Title>Reset to Default view</Modal.Title>
+                                            </Modal.Header>
+                                            <Modal.Body>
+                                                <div>Are you sure you want to override the list and order of the selected charts
+                                                     with the default setting? This is irreversible.</div>
+                                            </Modal.Body>
+                                            <Modal.Footer>
+                                                <button className='btn btn-primary btn-sm'
```
zhx828: User could exit the modal when decided not to restore by clicking the x icon on the top right. I think it also make sense to have a `No` button after `Yes` to naturally close the dialog.

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -268,6 +269,20 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
                     />
                 </MSKTab>
             </MSKTabs>
+
+            <div style={{
```
zhx828: It could just be a span element, so you don't need to specify display flex

jjgao:
- when the user is not logged in, maybe we don't need to "Default button"?
- Change the title of warning message to "Reset charts"
- Let's change the warning message to: "Please confirm that you would like to replace the current charts with the default list." and add "Confirm" and "Cancel" buttons.

kalletlak:
@zhx828 @jjgao updated

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -73,6 +73,7 @@ export default class StudyViewPage extends React.Component<IStudyViewPageProps,
     private queryReaction:IReactionDisposer;
     @observable showCustomSelectTooltip = false;
     @observable showGroupsTooltip = false;
+    @observable private showModal:boolean = false;
```
alisman: seems like this is a particular modal.  showReturnToDefaultChartList modal?

kalletlak:
@alisman updated

jjgao:
@kalletlak @alisman I don't think `reset charts` on the page is clear and necessary. Maybe we can rename `+ Add Chart` to `Charts` and have the `reset charts` button in the dialog?

PR #2724, opened 2019-09-20
Add tool tip to all bar charts in summary view

Luke-Sikina:
# What? Why?
Fix #[5425](https://github.com/cBioPortal/cbioportal/issues/5425) .

- Created BarChartToolTip component
    - Logic to deal with special text
    - Logic to deal with range edge cases
    - Logic to deal with positioning, avoiding going off screen
- Added BarChartToolTip to BarChart
   - Change handlers for mouse enter/exit on the bars of the bar chart
   - Logic for tracking mouse position while in the div

# Checks

# Any screenshots or GIFs?
tool-tip

### DIFF
--- a/src/pages/studyView/charts/barChart/BarChart.tsx
+++ b/src/pages/studyView/charts/barChart/BarChart.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import {observer} from "mobx-react";
 import {VictoryAxis, VictoryBar, VictoryChart, VictorySelectionContainer} from 'victory';
-import {computed} from "mobx";
+import {computed, observable} from "mobx";
 import _ from "lodash";
 import CBIOPORTAL_VICTORY_THEME from "shared/theme/cBioPoralTheme";
 import {ClinicalDataIntervalFilterValue, DataBin} from "shared/api/generated/CBioPortalAPIInternal";
@@ -19,6 +19,10 @@ import {
 import {STUDY_VIEW_CONFIG} from "../../StudyViewConfig";
 import {getTextDiagonal, getTextHeight, getTextWidth} from "public-lib/lib/TextTruncationUtils";
 import {DEFAULT_NA_COLOR} from "shared/lib/Colors";
+import BarChartToolTip, { ToolTipModel } from "./BarChartToolTip";
+import { IAlterationData } from "pages/resultsView/cancerSummary/CancerSummaryContent";
+import WindowStore from "shared/components/window/WindowStore";
+import ReactDOM from "react-dom";
 
 export interface IBarChartProps {
     data: DataBin[];
@@ -48,6 +52,15 @@ const TILT_ANGLE = 50;
 export default class BarChart extends React.Component<IBarChartProps, {}> implements AbstractChart {
 
     private svgContainer: any;
+    
+    @observable.ref
+    private mousePosition = {x: 0, y: 0};
+
+    @observable
+    private currentBarIndex = -1;
+
+    @observable
+    private toolTipModel: ToolTipModel | null = null;
 
     constructor(props: IBarChartProps) {
         super(props);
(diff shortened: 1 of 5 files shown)


src/pages/studyView/charts/barChart/BarChartToolTip.tsx
```diff
@@ -0,0 +1,110 @@
...
+        if (text === "<=") {
+            return "≤";
+        }
+
+        return text;
+    }
+
+
+    @computed
+    get rangeText(): string {
```
zhx828: The start and end need  to be converted using `toLocaleString`. Also the format of the range should keep the consistency as in the density plot which we use `a-b` without the spaces. We should probably consolidate the code a bit. The related function is `getBinStatsForTooltip` in StudyViewUtils.

src/pages/studyView/charts/barChart/BarChartToolTip.tsx
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    
+    render() {
+        if (!this.props.range) {
+            return null;
+        }
+
+        return <Popover
+            arrowOffsetTop={OFFSET}
```
zhx828: I'm not sure what the consideration is to come up with the offset. We also some offset settings for the ScatterPlot in the component `ScatterPlotTooltip`. May worth a discussion to consolidate them.

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -157,10 +173,52 @@ export default class BarChart extends React.Component<IBarChartProps, {}> implem
...
+                            self.sampleCount = event.datum.y;                            
+                        }
+                    }];
+                },
+                onMouseLeave: () => {
+                    return [{
+                        target: "data",
+                        mutation: () => {
+                            self.toolTipRange = null;
+                            self.sampleCount = -1;
```
zhx828: Why do we need to update sampleCount?

zhx828:
@Luke-Sikina the pr looks good. Just few minor comments. Thanks.

src/pages/studyView/charts/barChart/BarChartToolTip.tsx
Luke-Sikina: Hmm. I took this value from `CancerSummaryChart`. I looked at `ScatterPlotTooltip` and I don't think those values would be appropriate for a bar chart - each bar is only `20px` wide, so a `25px` would point to the wrong bar.

That said, you are right that I borrowed a value and didn't consolidate. I will convert the offsets in `CancerSummaryChart` to constants and reuse them.

src/pages/studyView/charts/barChart/BarChart.tsx
Luke-Sikina: We don't. I left it in as a way to preserve data honesty for anyone trying to debug this. When the mouse leaves the bar, we don't know what `sampleCount` is, because there is no bar. I thought  setting `sampleCount` to a clearly invalid value was the best way to indicate this.

Ultimately, this isn't a big deal for me. If you want me to remove it, that's not a problem. I could also switch `-1` to a static constant to cut down on the magic numbers a bit. I'll defer to you on this one.

src/pages/studyView/charts/barChart/BarChartToolTip.tsx
Luke-Sikina: This is a good suggestion - I will put my logic in there and reuse what's already there if I can. Thank you.

src/pages/studyView/charts/barChart/BarChart.tsx
alisman: I think it's better to keep all this stuff on tooltipModel to make it super clear what the context of the data is.  When mouse leaves, you just set the model to null.

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -48,6 +52,18 @@ const TILT_ANGLE = 50;
 export default class BarChart extends React.Component<IBarChartProps, {}> implements AbstractChart {
 
     private svgContainer: any;
+    
+    @observable
```
alisman: whenever making an object observable, consider using @observable.ref, which makes the object reference observable, but not individual properties.  it's basically an immutable at that point.  there is a cost to making things deeply observable and we shouldn't pay it unless necessary.  this case is totally negligible, but good for illustration.  at 178, instead of setting x and separately, you can just set mousePosition to a new object with x,y properties.

src/pages/studyView/charts/barChart/BarChartToolTip.tsx
```diff
@@ -0,0 +1,77 @@
...
+    sampleCount: number,
+    windowWidth: number,
+};
+
+export const VERTICAL_OFFSET = 17;
+export const HORIZONTAL_OFFSET = 8;
+
+const WIDTH = 150;
+
+export default class BarChartToolTip extends React.Component<BarChartToolTipProps, {}> {
```
alisman: This might need to be @observer depending on whether its props can change during lifecycle (mouse position)?

src/pages/studyView/charts/barChart/BarChart.tsx
Luke-Sikina: Ok, yes. That is a good idea that does both things.

alisman:
@zhx828, was reviewing a recent PR by @Luke-Sikina and told him he should not use toLocalString because our app is not localized.  The numbers should appear the same for everyone regardless of locale.  Is there something I'm missing?  Thanks

PR #2727, opened 2019-09-20
Make patient view sample order draggable

adamabeshouse:


src/pages/patientView/timeline/Timeline.tsx
```diff
@@ -14,22 +14,21 @@ import {PatientViewPageStore} from "../clinicalInformation/PatientViewPageStore"
...
 export default class Timeline extends React.Component<ITimelineProps, {}> {
 
     private currentWidth:number;
 
     shouldComponentUpdate(nextProps:ITimelineProps){
-        if (nextProps.width !== this.currentWidth) {
-            // only rerender to resize
-            this.drawTimeline(nextProps.width);
-        }
+        this.drawTimeline(nextProps.width);
```
adamabeshouse: @inodb is this okay?

src/pages/patientView/sampleManager.tsx
```diff
@@ -122,21 +123,33 @@ class SampleManager {
...
-            this.sampleIndex[sample.id] = i;
-            this.sampleLabels[sample.id] = String(i+1);
-        });
-        // order as array of sample ids (used further downstream)
-        this.sampleOrder = _.sortBy(Object.keys(this.sampleIndex), (k) => this.sampleIndex[k]);
+        this._samples = sortSamples(_samples, this.clinicalDataLegacyCleanAndDerived, events);
+    }
+
+    @computed public get samples() {
+        const urlSampleOrder = JSON.parse(this.urlWrapper.query.sampleIdOrder || "[]");
```
alisman: i have that we do the parsing on as needed basis.  i think it should take place in one place.  i don't think SampleManager should be aware of url

src/pages/patientView/timeline/Timeline.tsx
alisman: how come you took check off?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -73,24 +75,19 @@ export interface IPatientViewPageProps {
...
     @observable private mutationTableColumnVisibility: {[columnId: string]: boolean}|undefined;
     @observable private cnaTableColumnVisibility: {[columnId: string]: boolean}|undefined;
+    private urlWrapper:PatientViewURLWrapper;
 
     constructor(props: IPatientViewPageProps) {
 
         super(props);
 
-
+        this.urlWrapper = new PatientViewURLWrapper(props.routing);
```
alisman: if we're going to put in urlWrapper, maybe we should go ahead and refactor the rest of page?

src/pages/patientView/timeline/Timeline.tsx
adamabeshouse: because I made it into an observer so its going to rerender based on changes to observables inside the sampleManager

PR #2728, opened 2019-09-23
validate sample and patient before render the patient view

dippindots:
validate sample and patient before render the patient view
Fix # https://github.com/cbioportal/cbioportal/issues/5357.

- a show error when sample or patient not exists
image


dippindots:
test links:
1. [sample](https://cbioportal-frontend-pr-2728.herokuapp.com/patient?sampleId=P-0035148-T01-IM6&studyId=msk_impact_2017)
2. [patient](https://cbioportal-frontend-pr-2728.herokuapp.com/patient?studyId=ucec_tcga_pub&caseId=not-a-sample)

alisman:
@dippindots 
For this case, we should not say "URL invalid."  The URL form is valid, it's just that the ID doesn't exist.   So, error message should be that Patient doesn't exist.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -109,12 +109,20 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
-                    }
-                    if ('caseId' in query) {
-                        patientViewPageStore.setPatientId(query.caseId as string);
-                    } else if ('sampleId' in query)
-                    {
-                        patientViewPageStore.setSampleId(query.sampleId as string);
+
+                        if ('caseId' in query) {
+                            patientViewPageStore.setPatientId(query.caseId as string);
+                            const validationPatient = await validatePatient(query.caseId, query.studyId);
```
alisman: I think this check belongs in the patientViewStore, not here.

alisman:
When a user submits a request with invalid sample/patient, we already get 404s.  So we can accomplish this ticket by properly handling those errors states for the existing mobx promises.  In other words, we should just show a message when we detect isError=true

here is example 
https://www.cbioportal.org/patient?studyId=ucec_tcga_pub&caseId=TCG-BK-A0CC

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -267,6 +269,9 @@ export class PatientViewPageStore {
     readonly samples = remoteData(
         {
             invoke: async () => fetchSamplesForPatient(this.studyId, this._patientId, this.sampleId),
+            onError: (err: Error) => {
```
alisman: per discussion, test for specific error situations and then add via appStore

src/shared/components/errorContainer/ErrorContainer.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
```
alisman: this will then be unnecessary

dippindots: Yes, I will use ErrorScreen rather than this.

PR #2736, opened 2019-09-26
Study view gene level tab

dippindots:
# What? Why?
frontend changes for https://github.com/cBioPortal/cbioportal/issues/6278
CC: @kalletlak 

- a add a gene level tab in the `add chart` button
image


kalletlak:
Feedback from Tali
- [ ] After I add a chart, the chart appears in the Genomic tab as a greyed out option with 0% frequency. This is not an accurate reflection of the data
- [ ] The added charts don’t have a “Compare Groups” option
    > this should be resolved with https://github.com/cBioPortal/cbioportal/issues/6783
- [x] The info button on the added charts should show from the profile description
- [ ] It would be helpful to show the number of samples with data for each profile in the molecular profile dropdown (like on the coexpression tab)
- [x] In the “Gene Level” view, the text says “Click gene symbols below or enter here” but there are no gene symbols below.
- [x] The box to enter gene symbols doesn’t need to be that big
- [x] The box allows me to enter genes with OQL, but the OQL is ignored. If we’re not going to listen to the OQL, better to not allow it
- [ ] When I add a chart, it doesn’t add to the end of the page, it adds somewhere in the middle. This makes it hard to find. When I add a clinical/genomic chart it adds to the end with a yellow outline… can we replicate that behavior?
    > Chart position depends on the priority. Not always charts are added at the end. Yes, yellow outline is shown for new added charts.

- [x] Download data option doesn’t work in the added charts

alisman:
@kalletlak should this be closed?

kalletlak:
Opened a new one, https://github.com/cBioPortal/cbioportal-frontend/pull/2890. closing this

PR #2737, opened 2019-09-26
Fix protein enrichments tabs profile selection

kalletlak:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal/issues/6645


jjgao:
@kalletlak @alisman would this be worth a screenshot test?

PR #2742, opened 2019-09-30
Quick search for samples

Luke-Sikina:
Fix #6399
- Quick Search Sample Results
    - Searching for a keyword that matches a sample returns that sample
    in the select
    - Expanded OptionType enum to include values for sample
    - Added logic to add sample results to select
    - Added logic to call new sample endpoint, map results to option
    - Reran the swagger gen to build the API
- Quick Search Patient Results
    - Searching for a keyword that matches a sample also returns patients
    that have that sample
    - No front end changes needed
- e2e tests
    - Created quick search test. It runs a query and verifies that all
    four categories lead to their respective categories.

# Checks

# Any screenshots or GIFs?


src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -165,6 +178,11 @@ export default class QuickSearch extends React.Component {
                         value: parseInt(response[5].headers["total-count"]) - patientOptions.length,
                         type: OptionType.PATIENT_COUNT
                     }
+                    const sampleOptions: any = response[6].body.map(this.sampleToOption);
```
alisman: i can see this is an issue the predates this PR but i'm curious why this stuff isn't typed.    like, what is the actual return type of the client methods?

src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -232,12 +259,21 @@ export default class QuickSearch extends React.Component {
             parameters = {studyId: newOption.studyId, caseId: newOption.patientId};
             route = "patient";
             this.trackClick("patient", this.inputValue);
+        } else if (newOption.type === OptionType.SAMPLE){
+            parameters = {
+                studyId: newOption.studyId,
+                sampleId: newOption.value,
+            };
+            route = "patient";
```
alisman: lets see if there's an enum for this somewhere

src/shared/components/query/quickSearch/QuickSearch.tsx
Luke-Sikina: There are two separate problems here: on the line you selected, the type of the response is not the reason `sampleOptions` is typed as any. Rather, the option type used by `studyToOption, geneToOption, patientToOption, sampleToOption` is super loosely typed. It's loosely typed because each of those use cases is unique. I could use union types to try and type it, but I'm not sure what the end result would be. If you want me to, I will. I'll probably try anyways because type systems are fun to fool around with.

The other problem is with the response type itself. You can type the response in the `then` as `[request.Response, request.Response, request.Response, request.Response, request.Response, request.Response, request.Response, request.Response, request.Response]` and that part compiles. Unfortunately, the type declaration seems a bit incorrect, because it thinks it should be `response[5].header` and not `response[5].headers`. Headers clearly works, so I'm blaming `node_modules/@types/superagent/index.d.ts#Response`. I can fix this by declaring an interface that extends `request.Response` that adds the `headers` field. However, `headers` would have to be of type `any` or have an index signature. I don't think there is a lot to gain here. Furthermore, upon looking at the actual `Object.keys(reponse[0])`, the interface I'd be extending is pretty far off, and it would be misleading to reference it. The console says that the Response is just `["body", "text", "headers", "statusCode", "status", "ok"]`, but the type declaration thinks it should have all of these fields:
```
        text: string;
        body: any;
        files: any;
        header: any;
        type: string;
        charset: string;
        status: number;
        statusType: number;
        info: boolean;
        ok: boolean;
        redirect: boolean;
        clientError: boolean;
        serverError: boolean;
        error: Error;
        accepted: boolean;
        noContent: boolean;
        badRequest: boolean;
        unauthorized: boolean;
        notAcceptable: boolean;
        notFound: boolean;
        forbidden: boolean;
        xhr: XMLHttpRequest;
        get(header: string): string;
        redirects: string[];
``` 
I think the existing `any`s are the best solution here.

Luke-Sikina: Update: I typed the Options so you can see what it looks like. The changes are in their own commit so you can read through them easily. I like static types a lot, so I am not an impartial observer.

If you like the changes I will squash them and keep them. If you don't like them, that's fine too- I can throw them out just as easily.

alisman: I took a look at generated client and think i see what's happening.  All methods have accompanying WithHTTPInfo siblings.  These sibling method allow you to access the http headers, but for reason they do NOT type the response data (as the plain methods do).  You can see this looking at getAllGenesUsingGET and getAllGenesUsingGETWithHTTPInfo.    I think this is a crappy detail of the ts client generator.   Anyhow, lets not worry about any of this.  You've improved the situation with the Option type.  Thanks!

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3297,7 +3297,7 @@ export class StudyViewPageStore {
                     record.qValue === undefined ? '' : getQValue(record.qValue),
                     record.totalCount,
                     record.numberOfAlteredCases,
-                    getFrequencyStr(record.numberOfAlteredCases / record.numberOfSamplesProfiled * 100)
```
Luke-Sikina: @alisman I an worried I did the wrong thing here. When I ran `generateAPI`, the changes broke everything you see me fix in the commit. I looked at the type decl and it looked like it was just a rename, but I have no real clue.

alisman:
@inodb @jjgao  any reason we shouldn't deploy this?  i tested it out, seems to work great.   
https://cbioportal-frontend-pr-2742.herokuapp.com
Should we indicate in placeholder text of search field that we can also search by sample name now?

PR #2744, opened 2019-10-01
Adjustments for mouse cBioPortal

oplantalech:
- Fix chromosome tracks in Patient View: when having mouse studies, do not show chromosomes 20, 21 and 22 in the Patient View (they do not exist in mouse).
- In the study view, do not filter "Mutated Genes" by OncoKB genes ~if species is `mouse`~ if `show_oncokb` is `false`.

~Requirement before merging this PR: [the "species" property needs to be passed from the backend to the frontend (backend PR)](https://github.com/cBioPortal/cbioportal/pull/6659).~


src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -132,16 +132,21 @@ export function getChmInfo(genomeBuild:string) {
...
 }
 
-export function plotChromosomes(p: any, config: any,chmInfo: any) {
+export function plotChromosomes(p: any, config: any,chmInfo: any, genomeBuild: any) {
     var yRuler = config.rowMargin+config.ticHeight;
     drawLine(config.wideLeftText,yRuler,config.wideLeftText+config.GenomeWidth,yRuler,p,'#000',1);
     // ticks & texts
     for (var i=1; i<chmInfo.genomeRef.length; i++) {
+        if (chmInfo.genomeRef[i] === 0 && genomeBuild === "GRCm38") {
+            // Do not display chromosomes 20, 21 and 22 in mouse, they do not exist
```
alisman: nice comment!

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2259,6 +2260,9 @@ export class StudyViewPageStore {
     });
 
     @computed get oncokbCancerGeneFilterEnabled() {
+        if (AppConfig.serverConfig.species === "mouse") {
```
alisman: @jjgao @inodb is it correct that a whole portal instance could be species specific and should it be treated as a configuration property like this?

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -15,6 +15,7 @@ interface TracksPropTypes {
 }
 
 export const DEFAULT_GENOME_BUILD="GRCh37";
+// export const DEFAULT_GENOME_BUILD="mm10";
```
SRodenburg: This could be deleted. Elsewhere `GRCm38` is used instead?

src/pages/studyView/StudyViewPageStore.ts
inodb: @alisman the species is in the backend. Currently the database schema would allow for different species in one instance:

```
mysql> select * from reference_genome;
+---------------------+---------+------+------------+-------------+---------------------------------------------------------+---------------------+
| REFERENCE_GENOME_ID | SPECIES | NAME | BUILD_NAME | GENOME_SIZE | URL                                                     | RELEASE_DATE        |
+---------------------+---------+------+------------+-------------+---------------------------------------------------------+---------------------+
|                   1 | human   | hg19 | GRCh37     |  2897310462 | http://hgdownload.cse.ucsc.edu/goldenPath/hg19/bigZips  | 2009-02-01 00:00:00 |
|                   2 | human   | hg38 | GRCh38     |  3049315783 | http://hgdownload.cse.ucsc.edu/goldenPath/hg38/bigZips  | 2013-12-01 00:00:00 |
|                   3 | mouse   | mm10 | GRCm38     |  2652783500 | http://hgdownload.cse.ucsc.edu//goldenPath/mm10/bigZips | 2012-01-01 00:00:00 |
+---------------------+---------+------+------------+-------------+---------------------------------------------------------+---------------------+
```
I don't think anyone is using it like that at the moment, so a single portal instance per species (or genome build for that matter) probably makes sense, but you should be able to retrieve for a particular study whether it is mouse or human using the API given this db schema. So maybe better to use an endpoint instead of using `serverConfig`

Maybe @n1zea144 or @jjgao can elaborate

alisman: @JiaoJiao123 @n1zea144 please make call on whether this important

jjgao: I don't think our code frontend/backend code can support multiple species, e.g. how to deal with the same gene symbol in different species?

@oplantalech can we not add the `species` property and use oncokb-related properties to turn off oncokb for mouse instance? Ideally we should remove the mouse reference genome from the human seed database as well.

oplantalech:
I agree with @jjgao's comment. I have changed the `species` for the `show_oncokb` property. I think it does make sense, even in human, that if we should only filter genes by OncoKB if `show_oncokb` is `true`.

src/pages/patientView/genomicOverview/tracksHelper.ts
alisman: on further thought this seems like a cloodge.     DOes the 0 value indicate more generally that chromosome doesn't exist?   Shouldn't the collection of chromosome data reflect the species chromosome further upstream?   At very least, lets make a constant for the genomeBuild value . MOUSE_GENOME or something like that.

oplantalech: @alisman Because this was created for human, chromosomes 23 and 24 are turned to X and Y. I thought the best way to adapt it for mouse without touching the code too much was to set the length of chromosomes 20, 21 and 22, which do not exist in mouse, to 0, to respect the structure of `chromosomeSizes.json`.

Regarding the genome build names, they are unique, even between species. For example, build 38 in human is `GRCh38`, while the mouse one is `GRCm38`.

Sjoerd-van-Hagen:
@alisman @jjgao can we move this forward?

PR #2745, opened 2019-10-01
Filter mutations in genome tracks based on gene coverage filter in Patient View

pvannierop:
Note: this branch builds on top of branch `patientview_gene_coverage_filter` and the PR is to this branch for the reason of reviewing the code. This PR is therefore in _draft_ mode.

# What? Why?

At present, the genomic tracks of the mutation table is not responding to the gene/mutation filter setting. Also, there is no indication in the genomic tracks to the user with what gene panel was used to analyze a particular sample.

# Fix
- We have implemented gene panel icons behind mutation and CNA genomic tracks.
- Mutations in the mutation genomic tracks are filtered when genes in the mutation table are filtered

Note: continuous CNA's in the copy number genomic tracks are not filtered bu the gene filter option in the discrete CNA table.  

# Preview
Peek 2019-10-01 12-28


src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -5,21 +5,32 @@ import * as tracksHelper from './tracksHelper';
...
 
+
 export default class Tracks extends React.Component<TracksPropTypes, {}> {
 
     componentDidMount() {
+        this.drawTracks();
+    }
+
+    componentDidUpdate() {
+        this.drawTracks();
```
alisman: this will result in unnecessary re-render of the visualization because update will fire every time component parents render cycle kicks in whether or not props have changed.   where rendering is non-negligible (here), we should use an optimization like should componentUpdate or pure renderer to prevent this.   We can merge this but lets revisit.

pvannierop: @alisman yes, let's discuss this and provide an update.

PR #2746, opened 2019-10-01
Integrate gene panel information and gene filtering in VAF plot in Patient View

pvannierop:
Note: this branch builds on top of branch `patientview_filter_genome_overview` and the PR is to this branch for the reason of reviewing the code. This PR is therefore in _draft_ mode.

# What? Why?

At present, the legend of VAF plot there is no indication what gene panel was used to analyze a particular sample. In addition, when filtering mutations in the mutation table the plot did not reflect the new mutation frequencies.

# Fix
- We have implemented gene panel icons in front of the sample icons in the VAF plot legend.
- Mutation frequencies in the VAF plot reflect the mutation currently shown in the mutation table (updated after filtering).

# Preview
Peek 2019-10-01 15-13


jjgao:
Thanks, @pvannierop.

It looks like we have the `W` icon even if all the samples were WXS/WGS. It's confusing the tooltip for `W` is "Gene panel: N/A". When we don't have a gene panel, maybe we should not display the `W` icon?

image

The filtering on `Gene` column is also not necessary when there is only one sample or when the gene panels for all samples are the same. When there are multiple gene panels, it may be a good idea to add the gene panels to this tooltip, or at least explain "Different gene panels were used for the samples."

image

@cBioPortal/product

pvannierop:
@jjgao Nice feedback!

Changes that I will apply based on you feedback:
1. In genomic tracks I will show gene panel icons only when:
   - a mix of whole genome analysis and gene panel analysis were used
   - different gene panels were used
The gene panel icons will be hidden when:
   - all samples were analyzed with whole genome approach
   - the same gene panel was used
2. The tooltip for whole genome analysis icons will be renamed from `Gene panel : N/A` to `Whole genome`.
3. In the mutation and CNA tables the filter icon only appears when:
   - there are samples that have not been profiled for a listed gene/mutation
   - there are less than two samples.
4. The popup menu for the filter icon will get text:
```
Different gene panels were used for the samples. Filter mutations for:
0 Genes profiled in any sample
0 Genes profiled in all samples
```
Do I translate your proposals correctly?

jjgao:
@pvannierop thanks for the changes and summary!

> 1. In genomic tracks I will show gene panel icons only when: a mix of whole genome analysis and gene panel analysis were used or different gene panels were used

Maybe we can always show the icons when gene panel info exists?

> 2. The tooltip for whole genome analysis icons will be renamed from Gene panel : N/A to Whole genome.

Maybe change to "Gene panel information not found, assuming the sample is whole exome/genome sequennced"

> 3. In the mutation and CNA tables the filter icon only appears when:
>  there are samples that have not been profiled for a listed gene/mutation'

+1

>  there are less than two samples.

Do you mean 2 or more samples?

4. The popup menu for the filter icon will get text:

+1

Sjoerd-van-Hagen:
> Maybe we can always show the icons when gene panel info exists?

Definitely. Even if there is only one gene panel we need to show that. Otherwise the user will assume WGS.

> Maybe change to "Gene panel information not found, assuming the sample is whole exome/genome sequennced"

+1

> > 1. In the mutation and CNA tables the filter icon only appears when:
> >    there are samples that have not been profiled for a listed gene/mutation'
> 
> +1
> 

If the same gene panel is used for all the samples, there is no need to show the filter as 'any' and 'all' will be the same. This is also the case if there are multiple panels and the genes shown are on both of them. But the user does not know that! If you have a multipage table with a lot of mutations and you see different gene panels have been used, you would likely still want to be able to select on any or all, even though the result would be the same. Just to be sure of what you are looking at.
Perhaps it would be better just to gray out the icon and show a tooltip saying that it does not have an effect. Simply because the user will be expecting the icon there and might think it is a bug if it is not.

jjgao:
> If the same gene panel is used for all the samples, there is no need to show the filter as 'any' and 'all' will be the same. This is also the case if there are multiple panels and the genes shown are on both of them. But the user does not know that! If you have a multipage table with a lot of mutations and you see different gene panels have been used, you would likely still want to be able to select on any or all, even though the result would be the same. Just to be sure of what you are looking at.
Perhaps it would be better just to gray out the icon and show a tooltip saying that it does not have an effect. Simply because the user will be expecting the icon there and might think it is a bug if it is not.

That's a good point. Maybe we should simplify the logic and show the filtering icon when there are different panels (including wxs)?

Sjoerd-van-Hagen:
@jjgao yes, I think that would be best.

pvannierop:
> Do you mean 2 or more samples?
Yes, I had this reversed.

>If the same gene panel is used for all the samples, there is no need to show the filter as 'any' and 'all' will be the same.

This is all covered by the condition `- there are samples that have not been profiled for a listed gene/mutation`.

end-to-end-test/local/junit/customReport.json
```diff
@@ -0,0 +1 @@
+{"meta":{},"tests":[{"test":{"type":"afterTest","err":{"message":"element (\"#patientViewPageTabs\") still not existing after 20000ms","stack":"Error: element (\"#patientViewPageTabs\") still not existing after 20000ms\n    at elements(\"#patientViewPageTabs\") - isExisting.js:46:17\n    at isExisting(\"#patientViewPageTabs\") - waitForExist.js:67:22","type":"WaitUntilTimeoutError"},"title":"shows gene panel icons behind mutation and CNA genomic tracks","parent":"gene panel icons","fullTitle":"patient view page gene panel icons shows gene panel icons behind mutation and CNA genomic tracks","pending":false,"file":"/home/pnp300/git/cbioportal-frontend/end-to-end-test/local/specs/core/patientview.screenshot.spec.js","currentTest":"shows gene panel icons behind mutation and CNA genomic tracks","passed":false},"network":{}}]}
```
alisman: i think these were committed by mistake, yes?

alisman:
i think we should squash these commits.

pvannierop:
I squashed the commits.

end-to-end-test/local/junit/customReport.json
pvannierop: Yes, files removed....

PR #2747, opened 2019-10-01
Change default setting of gene filters in Patient View from portal.properties

pvannierop:
Note: this branch builds on top of branch `patientview_gene_coverage_filter` and the PR is to this branch for the reason of reviewing the code. This PR is therefore in _draft_ mode.

# What? Why?

At present, the mutation and CNA tables default to filtering genes based on the ANY SAMPLE rule. We have implemented a setting in the cbioportal backend that can change the default behavior for gene filtering to ALL SAMPLES in [PR #6658](https://github.com/cBioPortal/cbioportal/pull/6658). The Patient View does not yet respond to this variable passed by the backend. 

# Fix
- Patient View now starts up depending on the  `skin.patientview.filter_genes_profiled_all_samples` property.

# Preview
Peek 2019-10-01 16-09

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -1,5 +1,4 @@
 import * as _ from 'lodash';
-import {ClinicalDataBySampleId} from "../../../shared/api/api-types-extended";
 import {
     ClinicalData, MolecularProfile, Sample, Mutation, DiscreteCopyNumberFilter, DiscreteCopyNumberData, MutationFilter,
     CopyNumberCount, ClinicalDataMultiStudyFilter, ReferenceGenomeGene, GenePanelData, GenePanel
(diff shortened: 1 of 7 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -753,6 +754,14 @@ export class PatientViewPageStore {
         }
     }, undefined);
 
+    getGeneFilterDefault(config:IServerConfig):GeneFilterOption {
+        const propName = 'skin_patientview_filter_genes_profiled_all_samples';
```
adamabeshouse: my gut says this should be a util since it doesn't use `this`

pvannierop: Yes I agree with your gut feeling. Method was moved to new utils file. Also moved unit tests to new file.

inodb:
@pvannierop nice work! 

I noticed a minor bug for this one:

https://cbioportal-f-pv-mutatio-c012sw.herokuapp.com/patient/summary?caseId=P-0001830&studyId=lung_msk_2017

- [ ] The filter disappears on the CNA table:

gene_filter_bug

- [ ] Another minor stylistic thing is that the filter always looks enabled. Might be good to color it light gray if the `Genes profiled in any sample` is selected to indicate no filtering is done

PR #2749, opened 2019-10-01
Update GN clients

zhx828:


zhx828:
@alisman : @leexgh is working on the issues. I might need to close this pr if she makes any progress on that.

PR #2751, opened 2019-10-02
Flexible Gene Query Restrictions

Luke-Sikina:
Fix cBioPortal/cbioportal#6384

- Behavior
    - Previous: users cannot query more than 100 genes
    - New: users cannot make a query where `genes * samples > 1,000,000`
- Code changes
    - Changed the logic in QueryStore and ResultsViewPageStore for determining
if the query is valid
    - Changed error message to explain new restrictions
    - Changed the related config value from 100 to 1,000,000
# Checks

# Any screenshots or GIFs?
large-sample-set
small-sample-set
results-view

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -37,6 +37,7 @@ import QueryAndDownloadTabs from "shared/components/query/QueryAndDownloadTabs";
 import {createQueryStore} from "pages/home/HomePage";
 import ExtendedRouterStore from "shared/lib/ExtendedRouterStore";
 import {CancerStudyQueryUrlParams} from "../../shared/components/query/QueryStore";
+import GeneSymbolValidationError from "shared/components/query/GeneSymbolValidationError";
 
 function initStore(appStore:AppStore) {
 
@@ -431,16 +432,16 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
                 />
             </div>);
         } else {
-            return (<>
-                {
-                    // if query invalid(we only check gene length for now), return error page
-                    (this.resultsViewPageStore.isQueryInvalid) && (
-                        <div className="alert alert-danger queryInvalid" style={{marginBottom: "15px"}} role="alert">
-                            <p>
-                                Queries are limited to 100 genes. Please <a
-                                href={`mailto:${AppConfig.serverConfig.skin_email_contact}`}>let us know</a> your use
-                                case(s) if you need to query more than 100 genes.
-                            </p>
+            return (
+                <>
+                {// if query invalid(we only check gene count * sample count < 1,000,000 for now), return error page
+                    this.resultsViewPageStore.isQueryInvalid && (
+                        <div className="alert alert-danger queryInvalid" style={{ marginBottom: "40px" }} role="alert">
+                            <GeneSymbolValidationError
+                                sampleCount={this.resultsViewPageStore.samples.result.length}
+                                queryProductLimit={AppConfig.serverConfig.query_product_limit}
+                                email={AppConfig.serverConfig.skin_email_contact}
+                            />
                         </div>
                     )
                 }
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -1,59 +1,73 @@
-import * as React from "react";
-import * as _ from "lodash";
-import $ from "jquery";
+import * as React from 'react';
+import * as _ from 'lodash';
+import $ from 'jquery';
 import URL from 'url';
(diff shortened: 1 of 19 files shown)


jjgao:
Thanks, @Luke-Sikina! Could you change the logic a bit:

1. if genes <= 100, pass.
2. otherwise (genes >100),
   1. if genes x samples <= 1m, pass.
   1. otherwise, fail.

ie. only fail when there are more than 100 genes AND 1M genes x samples.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -433,13 +433,17 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
-                    // if query invalid(we only check gene length for now), return error page
+                    // if query invalid(we only check gene count * sample count < 1,000,000 for now), return error page
                     (this.resultsViewPageStore.isQueryInvalid) && (
-                        <div className="alert alert-danger queryInvalid" style={{marginBottom: "15px"}} role="alert">
+                        <div className="alert alert-danger queryInvalid" style={{marginBottom: "40px"}} role="alert">
                             <p>
-                                Queries are limited to 100 genes. Please <a
-                                href={`mailto:${AppConfig.serverConfig.skin_email_contact}`}>let us know</a> your use
-                                case(s) if you need to query more than 100 genes.
+                                Queries are limited by gene and sample count. A query must either have up to {AppConfig.serverConfig.query_gene_limit.toLocaleString()} genes or
```
alisman: Lets reverse the language here:   "A query is INVALID if gene count x sample count exceeds ${limit}"

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2634,7 +2634,14 @@ export class ResultsViewPageStore {
     }
 
     @computed get isQueryInvalid() {
-        return this.hugoGeneSymbols.length > AppConfig.serverConfig.query_gene_limit;
+        return (
+            this.hugoGeneSymbols.length > AppConfig.serverConfig.query_gene_limit &&
```
alisman: i think this is actually a really weird rule.  it's hard to describe.   if you go over 100 genes then all of the sudden, we start to care about the number of samples.  But if you simply have less than 100 genes, then we don't care!   Lets bring this up at standup.

src/shared/components/query/GeneSymbolValidator.tsx
```diff
@@ -43,9 +43,19 @@ export default class GeneSymbolValidator extends QueryStoreComponent<{}, {}>
...
 			return (
 				<div className={styles.GeneSymbolValidator}>
-					<div className={styles.invalidBubble} title="Please limit your queries to 100 genes or fewer.">
+					<div
+						className={styles.invalidBubble}
+						title= {`Please limit your queries to ${Math.max(this.store.geneLimit, AppConfig.serverConfig.query_gene_limit)} genes or fewer.`}>
 						<FontAwesome className={styles.icon} name='exclamation-circle'/>
-						<span>Queries are limited to 100 genes. Please <a style={{color: '#ccc'}} href={`mailto:${AppConfig.serverConfig.skin_email_contact}`}>let us know</a> your use case(s) if you need to query more than 100 genes.</span>
+						<span>
+							Queries are limited by gene and sample count. A query must either have up to {AppConfig.serverConfig.query_gene_limit.toLocaleString()} genes or
```
alisman: lets figure out a way not to duplicate this text.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -433,13 +433,17 @@ export default class ResultsViewPage extends React.Component<IResultsViewPagePro
...
+                                the product of gene and sample count must be less than {AppConfig.serverConfig.query_product_limit.toLocaleString()}.
+                                You have selected {this.resultsViewPageStore.samples.result.length.toLocaleString()} samples, so you can have no more
```
alisman: i'm actually not sure we want to use toLocalString.  Our app is not localized.  I believe our user base would actually be surprised to find anything in a language other than English, as scientific discourse these days is apparently done entirely in English as a rule.

alisman: you may want to use numeral.js library to format the number properly, (e.g. add commas)

alisman:
@Luke-Sikina ok, really sorry about this one but after long discussion and testing session with JJ, we decided the right solution here is a simple limit to samples x genes.   For public portal that will be 1 million.

jjgao:
@Luke-Sikina sorry that we need to revert back to your original implementation.

src/pages/resultsView/ResultsViewPageStore.ts
Luke-Sikina: You and JJ agree, so I'm down too.

PR #2756, opened 2019-10-04
Prettier JS

Luke-Sikina:
- config
    - made to match our current coding style
    - also made to add trailing commas because I like them
- README
    - deleted outdated linting info
    - added new linting info
- other
    - installed prettier for the people

# Checks


.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
+    "trailingComma": "es5",
+    "tabWidth": 4,
+    "printWidth": 120
```
Luke-Sikina: The software that will enforce trailing commas can't parse it's config file if it has them.

alisman: hah

.prettierrc.json
```diff
@@ -0,0 +1,5 @@
+{
```
alisman: Prettier plugin for Idea seems to require that this file be named .prettierrc to respect it

README.md
```diff
@@ -66,16 +66,17 @@ To run unit/integration tests in watch mode (where specName is a fragment of the
 yarn run test:watch -- --grep=#specName#
 ```
 
-To run linting
+## Linting
+We are starting to lint commits using Prettier JS. Right now this is optional, but you may run into formatting commits.
```
alisman: lets say "format code using Prettier JS"

.prettierrc.js
```diff
@@ -0,0 +1,6 @@
+// prettier.config.js or .prettierrc.js
```
alisman: seems like it has to .prettierrc   (no .js) for Idea to like it.    This is also what they have in prettier repo itself.

.prettierrc.json
Luke-Sikina: Ah ok. I was about to quote the docs saying that it can have a `js` or `json` extension, but you're saying the plugin specifically. I will change this to appease the plugin.

package.json
```diff
@@ -258,11 +256,6 @@
     "deep-equal-in-any-order": "^1.0.10",
     "enzyme": "^3.8.0",
     "enzyme-adapter-react-16": "^1.12.1",
-    "eslint": "^3.14.1",
```
alisman: i think these deletions need to be repped in the lock file in order for it to pass lock file test remotely

PR #2765, opened 2019-10-09
Using Data Access Tokens mouseover

ao508:
Data Access Tokens dropdown menu now includes mouseover on "Download token" that links out to Web API page where how to create DAT is described and there is a button for users to click to generate and save the Data Access Token file to their file system.

Screen Shot 2019-10-09 at 9 22 07 AM

Added new property to set the default doc page describing how to use data access tokens. (This property needs to be added to `cbioportal/src/main/resources/portal.properties.EXAMPLE` as well. 

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

# Checks

# Notify reviewers

### DIFF
--- a/src/appShell/App/PortalHeader.tsx
+++ b/src/appShell/App/PortalHeader.tsx
@@ -81,8 +81,7 @@ export default class PortalHeader extends React.Component<{ appStore:AppStore },
                 address:"/about",
                 internal:true,
                 hide:()=>AppConfig.serverConfig.skin_show_about_tab === false
-            },
-
+            }
         ];
 
 
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -57,6 +57,7 @@ export interface IServerConfig {
     show_genomenexus: boolean;
     skin_documentation_about: string | null;
     skin_documentation_baseurl: string | null;
+    "skin_documentation_dat": string | null;
     skin_blurb: string | null;
     skin_custom_header_tabs: string | null;
     skin_data_sets_footer: string | null;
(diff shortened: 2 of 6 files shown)


alisman:
@ao508 I don't think having the link to the page in the tooltip makes sense.  I would recommend that the dropdown button open a new tab where the user can obtain a token.  I think the link text should be "API Access Tokens"

ao508:
@n1zea144 Can you comment on this? Do you think we should keep the token download functionality here in the dropdown menu or make a tab that shows the DAT documentation and also has a button (or link) that download the token for the user. 
https://github.com/cBioPortal/cbioportal-frontend/pull/2765#issuecomment-540804778

alisman:
@alisman just to clarify my point:  I don't think "Download token" is specific enough language whether we have a tooltip or not.  It could be "Download data access token"

jjgao:
I like @alisman's idea.

schultzn:
I agree, a tooltip won't work. It could just be a third item in the list:
1 - Sign out
2 - Download data access token
3 - Help about data access tokens

Or, here is what ICGC does in their portal: 
Once you login, the token manager icon shows up.. and when you click on that, you get more information. Probably not what we should do for cBioPortal.
image

n1zea144:
Hi @ao508, I spoke with @alisman about this a short while ago.  We came up with a proposed solution.  What we would like do is change the current "Download Token" action so that instead of initiating a request for a token, it takes the user to the web-api tab.  We can add a Data Access Token Section to this page (maybe after the introduction) that describes how a token is required to access a password protected portal.  At this point, we can have the action to download the token.

The fact that we are directing a user's attention to this page makes me wonder if we should update the docs, remove the old webservice.do text and, at least, link to the swagger page, despite not having officially released the web service.

Thoughts?
@jjgao @schultzn

jjgao:
That's a good idea. It would also be good to add links to the token doc section in other places such as R/Matlab and swagger page.

PR #2767, opened 2019-10-10
Fix 6130

kalletlak:
Fix Frontend changes for https://github.com/cBioPortal/cbioportal/issues/6130


src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -566,9 +566,14 @@ export function makeEnrichmentDataPromise<T extends {cytoband?:string, hugoGeneS
                 }
 
                 const sortedByPvalue = _.sortBy(data, c=>c.pValue);
-                const qValues = calculateQValues(sortedByPvalue.map(c=>c.pValue));
-                qValues.forEach((qValue, index)=>{
-                    sortedByPvalue[index].qValue = qValue;
+                const qValues = calculateQValues(sortedByPvalue.filter(c => c.pValue !== undefined).map(c => c.pValue));
+
+                let index = 0;
+                sortedByPvalue.forEach(datum => {
```
alisman: this might be a little more straightforward if you just saved the result of the filter above and then did map on that, assigning qValue.  not that important.

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -566,9 +566,14 @@ export function makeEnrichmentDataPromise<T extends {cytoband?:string, hugoGeneS
...
+                    if (datum.pValue !== undefined) {
+                        datum.qValue = qValues[index];
+                        index++;
```
adamabeshouse: kind of small but you can do `forEach((datum, index)=>`

src/pages/resultsView/ResultsViewPageStoreUtils.ts
adamabeshouse: I actually think this is better because that way you dont duplicate the logic of pValue !== undefined. it's not a big deal in this case because it's easy to recompute, but I would say in general that makes the code slightly harder to read and more brittle. I think it's definitely worth doing

kalletlak: I could, but the results contain everything (even when pValue is undefined)

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -565,12 +565,20 @@ export function makeEnrichmentDataPromise<T extends {cytoband?:string, hugoGeneS
...
-                qValues.forEach((qValue, index)=>{
-                    sortedByPvalue[index].qValue = qValue;
+                const dataWithpValue: T[] = [];
+                const dataWithoutpValue: T[] = [];
+                data.forEach(datum => {
+                    datum.pValue === undefined ? dataWithoutpValue.push(datum) : dataWithpValue.push(datum);
                 });
-                return sortEnrichmentData(sortedByPvalue);
+
+                const sortedByPvalue = _.sortBy(dataWithpValue, c => c.pValue);
```
alisman: capitalize pValue

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -565,12 +565,20 @@ export function makeEnrichmentDataPromise<T extends {cytoband?:string, hugoGeneS
...
+                const qValues = calculateQValues(sortedByPvalue.map(c => c.pValue));
```
alisman: minor but i think we should always camel case pvalue in same way.  Eg. PValue (see cases above)

PR #2776, opened 2019-10-11
Fix issues due to backend changes for 3.1.1

zhx828:


zhx828:
We still have five failing tests, but they are related to the recent backend changes. Need to discuss with the team.Merge for now.

PR #2782, opened 2019-10-14
Gene Panel Modal in Patient View

ngocnn1104:
# Gene Panel Modal in Patient View
Resolves cBioPortal/cbioportal#6670

- Created PatientViewGenePanelModal that makes use of GenePanelModal component;
- Created GenesList component that displays a list of genes belonging to a gene panel with filtering, copy and download features - this component is passed as a child component to GenePanelModal at the moment and could be used in a separate page in the future;
- Triggered the Gene Panel modal from several places in Patient View: Patient header, Genomic tracks, Mutations table and Copy Number Alterations table;
- Implemented unit tests for the new component;
- Implemented local end-to-end tests.

# GIF
ezgif com-gif-maker


src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+}
+
+@observer
+export default class PatientViewGenePanelModal extends React.Component<
+    IGenesListProps,
+    IGenesListState
+> {
+    constructor(props: IGenesListProps) {
+        super(props);
+        this.state = {
```
alisman: i think we should stick to just using mobx observables in this project (instead of traditional React component state). 
Using both will cause confusion.  So, no setState please.

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+            filter: '',
+        };
+    }
+
+    handleChangeInput = (value: string) => {
+        this.setState({ filter: value });
+    };
+
+    filterGenes = () => {
```
alisman: this should be computed property.   That will avoid unecessary recomputation

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+        const { genes } = this.props.genePanelCache;
+        const { filter } = this.state;
+        if (filter) {
+            const lowerCaseFilter = filter.toLowerCase();
```
alisman: instead of toLowerCase(), lets RegExp(str,"i").test().  Case insensitive.

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+            return genes.filter(
+                each =>
```
alisman: gene instead of "each"?

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+            genesByRows.push(genesPerRow);
+        }
+        return genesByRows;
+    };
+
+    columnsCount = () => {
+        return this.props.columns || 1;
+    };
+
+    genesCountPerColumn = (totalLength: number) => {
```
alisman: geneCountPerColumn        gene should not have s following.  minor but better english.

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+    };
+
+    renderColumns = () => {
+        const filtered = this.filterGenes();
+        if (filtered.length === 0) {
+            return <td>No matches.</td>;
+        }
+        return this.genesDividedToColumns(filtered).map(col => (
+            <td>
+                {col.map(each => (
```
alisman: i don't like generic "each".  should be gene.

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+    getDownloadData = () => {
+        const downloadData = [['Genes'], ...this.genesDividedToRows()];
+        return serializeData(downloadData);
+    };
+
+    genesDividedToRows = () => {
+        const { genes } = this.props.genePanelCache;
+        const genesByColumns = this.genesDividedToColumns(genes);
+        const genesByRows = [];
+        for (let i = 0; i < this.genesCountPerColumn(genes.length); i++) {
```
alisman: when you call a function in the limit of a for loop, it gets executed EVERY iteration.   so you should always put it in a local variable above loop.

src/pages/patientView/PatientViewGenePanelModal/styles.scss
```diff
@@ -0,0 +1,31 @@
+#patient-view-gene-panel {
```
alisman: lets convert this to a css module.    you can look at other styles.module.scss files for examples.  happy to help.

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+                    />
+                </div>
+                <SimpleCopyDownloadControls
+                    className="pull-right genes-list-controls"
+                    downloadData={this.getDownloadData}
+                    downloadFilename={`gene_panel_${this.props.panelName}.tsv`}
+                    controlsStyle="BUTTON"
+                    containerId={this.props.id}
+                />
+                <table>
```
alisman: you should use bootstrap styles for this table.

src/pages/patientView/PatientViewGenePanelModal/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+  table {
```
alisman: use bootstrap style classes and override when necessary

src/pages/patientView/PatientViewGenePanelModal/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+  h4.panel-name {
+    margin-bottom: 15px;
+  }
+}
+
+.patientview-modal {
+  z-index: 1080 !important;
+}
+
+.genes-list-controls {
```
alisman: isn't gene-list-conrols inide of .patienview-modal?

src/pages/patientView/PatientViewGenePanelModal/styles.scss
```diff
@@ -0,0 +1,31 @@
...
+  .input-group-sm {
+    margin-left: 5px;
+  }
+  
+  h4.panel-name {
+    margin-bottom: 15px;
+  }
+}
+
+.patientview-modal {
```
alisman: there could be many modals on patient view.   if you use css modules, you can just name this .modal

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -448,6 +457,8 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                                     containerWidth={WindowStore.size.width-20}
                                                     sampleIdToMutationGenePanelId={patientViewPageStore.sampleToMutationGenePanelId.result}
                                                     sampleIdToCopyNumberGenePanelId={patientViewPageStore.sampleToDiscreteGenePanelId.result}
+                                                    onSelectGenePanel={patientViewPageStore.toggleModal}
```
alisman: the modal and its open state are really properties of the view more than the store.  Lets put them on the PagView instead.  Also, there could be more than one modal, so we need to be more specific with the names.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -36,6 +37,36 @@ export default class ClinicalInformationPatientTable extends React.Component<ICl
         }
         return ret;
     }
+    
+    handleClick = (name:string) => {
```
alisman: this should be methods, not functions assigned to properties.    so, handleClick() { ... }
You can then apply decorators like @bind or @action to them

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -36,6 +37,36 @@ export default class ClinicalInformationPatientTable extends React.Component<ICl
...
+        if (this.props.onSelectGenePanel) {
+            this.props.onSelectGenePanel(name);
+        }
+    }
+    
+    renderGenePanelLinks = (genePanels:string) => {
+        const links: (string|JSX.Element)[] = [];
+        const genePanelsArray = genePanels.split(",");
```
alisman: i think these are genePanelIds, no?   SHould be genePanelIds or genePanelNames.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -883,6 +886,10 @@ export class PatientViewPageStore {
             this.studiesForSamplesWithoutCancerTypeClinicalData,
             this.samplesWithoutCancerTypeClinicalData);
     }
+    
+    @computed get genePanelData() {
```
alisman: this isn't a good name because it hides the fact that it's data for a single selected genePanel.   better just to get rid of this computed and de-reference the data where you need it.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -947,6 +954,14 @@ export class PatientViewPageStore {
     @action clearErrors() {
         this.ajaxErrors = [];
     }
+    
+    @autobind
+    @action toggleModal(name?:string|undefined) {
```
alisman: "toggle" strongly implies that you are just turning something on or off.  I would

alisman: see comment above.  should be on view and needs more specific name

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -15,19 +15,32 @@ interface TracksPropTypes {
...
 }
 
-export const DEFAULT_GENOME_BUILD="GRCh37";
-
+interface TracksStateTypes {
+    genePanelInTooltip:string;
+}
 
-export default class Tracks extends React.Component<TracksPropTypes, {}> {
+export const DEFAULT_GENOME_BUILD="GRCh37";
```
alisman: not sure but i think this constant already exists elsequere.  we should have it one place unless it's only the default for this specific feature

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -15,19 +15,32 @@ interface TracksPropTypes {
...
+const noGenePanelMessage = 'Gene panel information not found. Sample is presumed to be whole exome/genome sequenced.';
 
+export default class Tracks extends React.Component<TracksPropTypes, TracksStateTypes> {
+    constructor(props:TracksPropTypes) {
+        super(props);
+        this.state = { genePanelInTooltip: '' };
```
alisman: no state

src/shared/components/mutationTable/column/PanelColumnFormatter.tsx
```diff
@@ -57,6 +56,27 @@ const getGenePanelIds = (props:PanelColumnFormatterProps) => {
   return [];
 };
 
+const getGenePanelLinks = (props:PanelColumnFormatterProps) => {
```
alisman: You could make this a React Functional Component and type it as such. Then use it in code via JSX:   <GenePanelLinks ... >

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import { GenePanel } from 'shared/api/generated/CBioPortalAPI';
+import { observer } from 'mobx-react';
+import GenePanelModal from 'shared/components/GenePanelModal/GenePanelModal';
+import GenesList from './GenesList';
+
+interface IGenePanelModalProps {
```
adamabeshouse: should be called `IPatientViewGenePanelModalProps`

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+    genePanelCache: GenePanel;
```
adamabeshouse: looks like its posible for genePanelCache not to exist, based on line 26. so type should be `genePanelCache?:...`. Also, why is it called `genePanelCache`, when its just a single `GenePanel` object?

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+    panelName: string;
```
adamabeshouse: is it necessary to pass this in separately? cant it be inferred from the `GenePanel` object?

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,40 @@
...
+    show: boolean;
+    hide: () => void;
```
adamabeshouse: is this a callback that fires when it hides? if so then it should be called `onHide`

src/pages/patientView/SampleManager.tsx
```diff
@@ -159,7 +159,9 @@ class SampleManager {
     getComponentForSample(sampleId: string,
                           fillOpacity: number = 1,
                           extraTooltipText: string = '',
-                          additionalContent: JSX.Element|null = null)
+                          additionalContent: JSX.Element|null = null,
+                          onSelectGenePanel?:(name:string)=>void,
+                          tooltipIsDisabled?:boolean)
```
adamabeshouse: this is subtle but I think it should be `disableTooltip`. `tooltipIsDisabled` sounds a bit off to me, almost like its describing the state of some other tooltip, whereas what its really doing is specifying whether the tooltip here should be disabled.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
adamabeshouse: just to clarify, I think @alisman means `handleClick(name:string) { ... }`

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -36,6 +37,36 @@ export default class ClinicalInformationPatientTable extends React.Component<ICl
...
+        genePanelsArray.forEach((panelName, index) => {
+            panelName = panelName.trim();
+            if (panelName.includes("N/A")) {
+                links.push(panelName);
+            } else {
+                // split gene panel name from "TESTPANEL2 (mut)" to ["TESTPANEL2", "(mut)"]
```
adamabeshouse: is it known 100% that all gene panel names are in this form? if not, then maybe using a regular expression would be more appropriate

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: I agree that it's not a good name but I think the computed makes sense for this. name should be `modalSelectedGenePanel` or something like that

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -310,22 +318,45 @@ export function plotMuts(p: any, config: any,chmInfo: any,row: any, mutations: A
...
     
 }
 
-function addToolTip(node: any, tip: any,showDelay: any, position: any) {
+function addToolTip(node: any, tip: any,showDelay: any, position: any, genePanel?:string, tooltipContentEl?:any, setGenePanelInTooltip?:(genePanelId:string)=>void) {
     var param = {
-        content: {text:tip},
+        content: {
+            text: () => {
+                return tooltipContentEl ? tooltipContentEl : tip
```
adamabeshouse: this is a little harder to read version of `tooltipContentEl || tip`

src/shared/components/GenePanelModal/GenePanelModal.tsx
```diff
@@ -0,0 +1,55 @@
+import * as React from 'react';
+import { Modal, Button } from 'react-bootstrap';
+import { observer } from 'mobx-react';
+import LoadingIndicator from 'shared/components/loadingIndicator/LoadingIndicator';
+
+interface IGeneModalProps {
+    panelName: string;
+    show: boolean;
+    hide: () => void;
```
adamabeshouse: `onHide`

alisman:
@ngocnn1104 we put a lot of comments on this!   let me know if you need any clarification or want to discuss.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
alisman: yes

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
```diff
@@ -0,0 +1,140 @@
...
+        return this.props.columns || 1;
+    };
+
+    genesCountPerColumn = (totalLength: number) => {
+        return Math.ceil(totalLength / this.columnsCount());
+    };
+
+    render() {
+        return (
+            <div id={this.props.id}>
```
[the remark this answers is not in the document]
ngocnn1104: The reason for using id here is that the SimpleCopyDownloadControls component (which uses clipboard.js lib) is inside Bootstrap Modal, and clipboard won't work in this case if it's not provided with the exact element in focus.

Please refer to Clipboard's [documentation](https://clipboardjs.com/):
Screenshot from 2019-10-28 14-17-12

Please let me know if you have a more subtle way to make it work.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
ngocnn1104: yeah they are indeed gene panel ids, but they are wrapped inside the anchor element and trigger gene panel modal on click so I gave them the name genePanelLinks. If you think it's confusing, I will switch to genePanelIds.

src/pages/patientView/genomicOverview/Tracks.tsx
ngocnn1104: I searched through the source code and only found this constant used in this feature, so I think it's alright to keep it here.

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
ngocnn1104: I kept the prop name from the previous version of the GenePanelModal component, will change it to genePanel.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
ngocnn1104: The Gene Panel names display in ClinicalInformationPatientTable are either `TESTPANEL1 (mut)` or simply a gene panel name like `TESTPANEL1`, but I might be wrong. I'm not really familiar with using regular expression, so how would you do it in this case?

src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
ngocnn1104: I have switched to using the shared `SimpleTable`.

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,37 @@
...
+    IPatientViewGenePanelModalProps,
+    {}
+> {
+    render() {
+        return (
+            <GenePanelModal
+                panelName="Gene Panel"
+                show={this.props.show}
+                onHide={this.props.onHide}
+                isLoading={!this.props.genePanel}
```
adamabeshouse: the type of genePanel is `genePanel:GenePanel`, not `genePanel?:GenePanel` (i.e. `GenePanel|undefined`). so `!this.props.genePanel` is always `false`. If that's what you're going for, then you should just pass `isLoading={false}`

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
```diff
@@ -0,0 +1,37 @@
...
+
+@observer
+export default class PatientViewGenePanelModal extends React.Component<
+    IPatientViewGenePanelModalProps,
+    {}
+> {
+    render() {
+        return (
+            <GenePanelModal
+                panelName="Gene Panel"
```
adamabeshouse: ah okay, now I see why you passed in the gene panel name in your previous version. because the `GenePanel` object doesn't have a `name` attribute. sorry about that. but just want to make sure it's okay to use this generic "Gene Panel" name instead of passing in the exact name

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -254,6 +256,18 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
                                             patientViewPageStore.sampleToDiscreteGenePanelId.result,
                                             patientViewPageStore.genePanelIdToEntrezGeneIds.result);
     }
+    
+    @autobind
+    @action toggleGenePanelModal(genePanelId?:string|undefined) {
+        this.genePanelModal = {
+            isOpen: !this.genePanelModal.isOpen,
+            genePanelId: typeof genePanelId === 'string' ? genePanelId : ''
```
adamabeshouse: `genePanelId: genePanelId || ''` is easier to read

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
adamabeshouse: if you know for sure that will be the format, and not e.g. <pre>"   TESTPANEL1     (mut)"</pre> or <pre>"TESTPANEL1(mut)"</pre>, then the way you're doing it is fine. otherwise, let me know and i'll help with regex because its a bit more involved

src/pages/patientView/genomicOverview/Tracks.tsx
```diff
@@ -15,21 +18,29 @@ interface TracksPropTypes {
...
-
+    @observable genePanelInTooltip:string = '';
+    
     componentDidMount() {
         this.drawTracks();
     }
 
     componentDidUpdate() {
-        this.drawTracks();
+        // this.drawTracks(); // causes unneccesary re-drawing of tracks when component local state changes
```
adamabeshouse: just delete `componentDidUpdate`

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
ngocnn1104: That's true, my bad. I have removed the `isLoading` prop because it's optional.

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
ngocnn1104: You were right before, gene panel name could be retrieved from `GenePanel.genePanelId`. Based on the mockup, the generic "Gene Panel" is just the header of the modal, while the exact gene panel name is placed in the modal body.
This is how the gene panel modal looks:
Screenshot from 2019-10-30 11-14-02

src/pages/patientView/PatientViewPage.tsx
ngocnn1104: I agree, and have updated according to your suggestion. However, this function was passed directly to handle the `onHide` event of Bootstrap Modal and `onClick` event of Bootstrap Button, and whenever the function was called, an event object would get passed to the parent component and be seen as `genePanelId`. In order to avoid this, I modified the shared `GenePanelModal` component like below:

```
    @autobind handleClose() {
        this.props.onHide();
    }
    
    render() {
        return (
            <Modal
                show={this.props.show}
                onHide={this.handleClose} // instead of onHide={this.props.onHide}
                ...
```

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
ngocnn1104: I found that the gene panel names here display exactly like what is returned from an API call, so I believe there are back-end logic which keeps the format consistent.

src/pages/patientView/PatientViewGenePanelModal/PatientViewGenePanelModal.tsx
adamabeshouse: ah okay, sounds good thanks for clarifying.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
adamabeshouse: sounds good

ngocnn1104:
> please add a comment as discussed here [#2782 (comment)](https://github.com/cBioPortal/cbioportal-frontend/pull/2782#discussion_r340714426), and other than that looks good thanks!

done. Thank you!

PR #2783, opened 2019-10-14
Temporarily skip two time consuming coexpression tests

zhx828:
On the master environment, the instance is not powerful enough to run these tests. They should be enabled once we move back to the production.


zhx828:
The only failing test has been tested working on the master portal.

PR #2784, opened 2019-10-14
Fix flaky e2e-localdb test related to utilities menu

pvannierop:
Fix this flaky test:
image


zhx828:
Thanks @pvannierop  I ran into this issue as well. Thought that just related to the release candidate.

alisman:
@pvannierop how come this is different from whatever's in waitForPlotsTab function?

pvannierop:
The waitForPlotsTab function waits for the axis menu is showing. In this case it was needed to actually have the plot showing.

PR #2786, opened 2019-10-14
Use latest oncoprint, which is with typescript, and update code for new typings

adamabeshouse:
Includes updating jquery typings, and updating code to work with that


src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -224,7 +224,7 @@ export default class Oncoprinter extends React.Component<IOncoprinterProps, {}>
     }
 
     @autobind
-    private oncoprintRef(oncoprint:OncoprintJS<any>) {
```
adamabeshouse: oncoprint is no longer generic typed because its useless in practice

src/public-lib/components/WindowWrapper.ts
```diff
@@ -10,15 +10,19 @@ export interface WindowSize {
 export default class WindowWrapper
 {
     @observable public size: WindowSize;
+    @observable public size500Ms: WindowSize;
 
     private handleWindowResize = _.debounce(this.setWindowSize, 200);
+    private handleWindowResize500Ms = _.debounce(this.setWindowSize500Ms, 500);
```
adamabeshouse: attention here. is this okay? I added this because for something really expensive to rerender, like oncoprint, its better to debounce it even more

package.json
```diff
@@ -56,7 +56,7 @@
     "@types/detect-browser": "^1.6.0",
     "@types/enzyme": "2.8.0",
     "@types/google.analytics": "0.0.39",
-    "@types/jquery": "2.0.46",
+    "@types/jquery": "^3.3.31",
```
alisman: we should make sure this works with jQuery migrate.  i think it's still necessary from some legacy stuff

adamabeshouse: im only updating the types, not the jquery version though

src/pages/studyView/charts/scatterPlot/StudyViewScatterPlot.tsx
```diff
@@ -91,8 +91,8 @@ export default class StudyViewScatterPlot extends React.Component<IStudyViewScat
     }
 
     componentWillUnmount() {
-        $(document).off("keydown",this.onKeyDown);
-        $(document).off("keyup", this.onKeyUp);
+        $(document).off("keydown" as any,this.onKeyDown as any);
```
alisman: bizarre that types don't include this?

adamabeshouse: agreed

src/public-lib/components/WindowWrapper.ts
```diff
@@ -10,24 +10,60 @@ export interface WindowSize {
 export default class WindowWrapper
 {
     @observable public size: WindowSize;
+    @observable public sizeByMs:{
```
alisman: not sure what sizeByMs means.  can you spell out Ms?

package.json
alisman: nevermind.  e2e tests should catch any failures

src/public-lib/components/WindowWrapper.ts
alisman: I think ms is milliseconds.   but if the value is the same, then why?

adamabeshouse: ms is milliseconds. the value is not the same, its different debouncing amounts. definitely open to different naming structure for this

PR #2787, opened 2019-10-14
Fix warning message for invalid genes

dippindots:
Fix cBioPortal/cbioportal#6706

### DIFF
--- a/src/pages/resultsView/querySummary/QuerySummary.tsx
+++ b/src/pages/resultsView/querySummary/QuerySummary.tsx
@@ -81,8 +81,11 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
     @autobind
     @action
     closeQueryForm(){
-        this.toggleQueryFormVisibility();
-        $(document).scrollTop(0);
+        // toggle QueryForm visibility only when queryFormVisible is true
+        if (this.props.store.queryFormVisible === true) {
+            this.toggleQueryFormVisibility();
+            $(document).scrollTop(0);
+        }
     }
 
     readonly multipleStudyUI = MakeMobxView({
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/querySummary/QuerySummary.tsx
```diff
@@ -81,8 +81,11 @@ export default class QuerySummary extends React.Component<{ routingStore:Extende
     @autobind
     @action
     closeQueryForm(){
-        this.toggleQueryFormVisibility();
-        $(document).scrollTop(0);
+        // toggle QueryForm visibility only when queryFormVisible is true
+        if (this.props.store.queryFormVisible === true) {
+            this.toggleQueryFormVisibility();
```
pvannierop: Why depend on a toggle rather setting the visibility to true or false?

dippindots: Hi @pvannierop. The reason I use `toggleQueryFormVisibility` is because this [button](https://github.com/cBioPortal/cbioportal-frontend/blob/47e140b29171c06c6e1b74f6b235865405798e84/src/pages/resultsView/querySummary/QuerySummary.tsx#L187) is actually toggling the query visibility, and we also have some logic in the [toggleQueryFormVisibility](https://github.com/cBioPortal/cbioportal-frontend/blob/47e140b29171c06c6e1b74f6b235865405798e84/src/pages/resultsView/querySummary/QuerySummary.tsx#L31-L38) function. So I decide not to duplicate those logic, just to see if I should close the form by perform `toggle` action here.

pvannierop: Ok, I see that now. Thnx.

PR #2790, opened 2019-10-15
Automated master-rc merge (mMRc5)

pvannierop:
This pull request has been automatically created for the merge of master to rc branch. Please resolve any merge errors and errors in the CI tests and merge.


inodb:
@pvannierop i updated the screenshots. Rc should be using the same db as master. Not sure why the order of the mutations changed for `rc`, but it does not seem to change the results drastically so fine to update i think

PR #2791, opened 2019-10-15
Fix check for .only in spec files

inodb:
The current check missed this one:

https://github.com/cBioPortal/cbioportal-frontend/pull/2779

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,7 +149,8 @@ jobs:
       - attach_workspace:
           at: /tmp/
       # check that no `.only` statements are still in the tests
-      - run: "! git grep '\\.only' end-to-end-tests**/*spec* src**/*spec*"
+      # Ignore doTest.only in oql parser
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -149,7 +149,7 @@ jobs:
       - attach_workspace:
           at: /tmp/
       # check that no `.only` statements are still in the tests
-      - run: "! git grep '\\.only' end-to-end-tests**/*spec* src**/*spec*"
+      - run: "! git grep '\\.only' $(find . -name '*spec*')"
```
alisman: test failing.    i think we have to check for paren .  .only\(

inodb: i just did an explicit ignore for those lines in oql parser with a `grep -v`

inodb:
Double checked that it worked here's an old run: https://circleci.com/gh/cBioPortal/cbioportal-frontend/51883?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link

PR #2794, opened 2019-10-17
Persist url params in ResultsView after update of query parameters

pvannierop:
# Why
1. The default configuration of certain components in ResultsView (e.g. sort order of oncoprint) can be controlled by setting a URL param. When modifying the query these params are lost from the url.
2. The oncoprint_sortby url param controls the default sort order of the oncoprint when it is loaded. At present this param is not updated when the sort order is updated by the user via the UI (Sort menu in OncoprintControls).

# Fix
This PR will:
1. persist all url params that were set by any component in ResultsView. URL params are no longer erased when the user modifies the query.
2. update the url param _oncoprint_sortby_ when the user changes the sort order via the UI. 

Important: the update of the query params  will not trigger a page reload, but will re-render the oncoprint component. A way should be investigated that prevents this from happening.

### DIFF
--- a/src/pages/home/HomePage.tsx
+++ b/src/pages/home/HomePage.tsx
@@ -29,7 +29,7 @@ export interface IResultsViewPageProps {
     routing: any;
 }
 
-export function createQueryStore(currentQuery?:any) {
+export function createQueryStore(currentQuery?:any, clear:boolean = true) {
 
     const win:any = window;
 
@@ -41,7 +41,7 @@ export function createQueryStore(currentQuery?:any) {
         query.cancer_study_list = query.cancer_study_list || query.cancer_study_id;
         delete query.cancer_study_id;
 
-        win.routingStore.updateRoute(query, "results", true);
+        win.routingStore.updateRoute(query, "results", clear);
 
     };
 
--- a/src/pages/home/HomePage.tsx
+++ b/src/pages/home/HomePage.tsx
@@ -29,7 +29,7 @@ export interface IResultsViewPageProps {
     routing: any;
 }
 
-export function createQueryStore(currentQuery?:any, clear:boolean = true) {
+export function createQueryStore(currentQuery?:any, clearQueryOnSubmit:boolean = true) {
 
     const win:any = window;
 
@@ -41,7 +41,7 @@ export function createQueryStore(currentQuery?:any, clear:boolean = true) {
         query.cancer_study_list = query.cancer_study_list || query.cancer_study_id;
         delete query.cancer_study_id;
 
-        win.routingStore.updateRoute(query, "results", clear);
+        win.routingStore.updateRoute(query, "results", clearQueryOnSubmit);
 
     };
 
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -3,7 +3,7 @@ import * as _ from 'lodash';
 import $ from 'jquery';
 import URL from 'url';
 import { inject, observer } from 'mobx-react';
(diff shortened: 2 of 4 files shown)


src/pages/home/HomePage.tsx
```diff
@@ -29,7 +29,7 @@ export interface IResultsViewPageProps {
     routing: any;
 }
 
-export function createQueryStore(currentQuery?:any) {
+export function createQueryStore(currentQuery?:any, clear:boolean = true) {
```
alisman: clearUrlOnSubmit?

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -178,13 +178,24 @@ export default class ResultsViewPage extends React.Component<
...
+
         getBrowserWindow().resultsViewPageStore = this.resultsViewPageStore;
     }
 
     private handleTabChange(id: string, replace?: boolean) {
         this.props.routing.updateRoute({}, `results/${id}`, false, replace);
     }
 
+    @autobind
+    private handleSilentUrlUpdate(url:string) {
```
alisman: this should become much easier when i finish my URLWrapper PR, which should allow features to observe separate parameters of the URL instead of having to observe the entire thing.

adamabeshouse: @pvannierop is it possible to wait until @alisman 's PR is merged? @alisman how much longer will it be?

pvannierop: >  Just to confirm, the point is to update sortMode without throwing the oncoprint into loading state, right?

Yes, that is correct.

> @pvannierop is it possible to wait until @alisman 's PR is merged?

For sure. No problem. I am happy to see there may be a solution.

src/pages/home/HomePage.tsx
pvannierop: I think `clearQueryOnSubmit` is more appropriate. Do you agree?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -221,6 +225,32 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
...
             }
         );
+        
+        if (this.props.replaceRoutingHistoryCallback) {
+            this.silentUrlParamsReaction = reaction(
+                ()=>[
+                    this.sortMode
+                ],
+                ()=>{
+                    const regex = new RegExp('&' + ONCOPRINT_SORTBY_URL_PARAM + '=[^&]*');
```
alisman: not important here but just want to make sure you know about es6 template strings which allow for string composition without annoying concatenation.  again, you don't need to update this.

alisman: just to be clear, you do NOT need to update this

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -221,6 +225,32 @@ export default class ResultsViewOncoprint extends React.Component<IResultsViewOn
                 getBrowserWindow().globalStores.routing.updateRoute(newParams, undefined, true, true);
             }
         );
+        
+        if (this.props.replaceRoutingHistoryCallback) {
+            this.silentUrlParamsReaction = reaction(
```
alisman: could this go in the reaction up above (line 200)

pvannierop: Well, I have added this new reaction because the final functionality is not to trigger refresh of components when the url is updated for certain url parameters (such as _oncoprint_sortby_).  If your changes in the future accomodate this, I agree.

PR #2797, opened 2019-10-17
range selection for survival chart

dippindots:
Fix cBioPortal/cbioportal#2176

- a add a chart on the top of original survival chart for range selection (show first 5 years by default)
8e4105f50a1c8d9ec4bf0b9edd767021


src/appBootstrapper.jsx
```diff
@@ -43,6 +43,49 @@ initializeConfiguration();
 // THIS TELLS WEBPACK BUNDLE LOADER WHERE TO LOAD SPLIT BUNDLES
 __webpack_public_path__ = AppConfig.frontendUrl;
 
+// TODO: temparary fix the wheel event described in: https://github.com/facebook/react/issues/14856
```
alisman: lets put this in a imported function from some util file.  handleTouchEventsForChromeBug or something like that.  just don't want this file getting too crowded

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -410,10 +414,50 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
         return this.victoryLegendData.length > 0;
     }
 
+    @observable.shallow zoomDomain: any = {x: [0, this.initialRangeMaxValue]};
```
alisman: you probably want @observable.ref if all you ever do is reset the reference

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -410,10 +414,50 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
+    }
+
+    @autobind
+    private updateDomain(domain: any) {
+        this.zoomDomain = {x: domain.x};
+    }
+
+    @computed get maximumDataMonthValue() {
+        let max = Number.MIN_VALUE;
+        _.forEach(this.sortedGroupedSurvivals, (survivals) => {
```
alisman: _.map().max()   would be easier to understand.  whenever you are using foreach to populate a local variable (e.g. max) you should think whether it can be done with assignment instead.   
const max = _().map().max().

alisman: code is more readable and less prone to error

src/appBootstrapper.jsx
dippindots: Sure!

src/pages/resultsView/survival/SurvivalChart.tsx
dippindots: Thanks!

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -410,10 +414,50 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
+
+    @computed get initialRangeMaxValue() {
+        if (this.maximumDataMonthValue < SURVIVAL_CHART_INITIAL_SHOWN_MONTHS) {
+            return this.maximumDataMonthValue;
+        }
+        return SURVIVAL_CHART_INITIAL_SHOWN_MONTHS;
+    }
+
+    @computed get additionalProps() {
+        if (this.props.showRangeSelector) {
```
alisman: getter should always return something.  undefined would work here.  i'm surprised that ts doesn't object to this

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -410,10 +414,50 @@ export default class SurvivalChart extends React.Component<ISurvivalChartProps,
...
+            return {
+                zoomDomain: this.zoomDomain
+            }
+        }
+    }
+
+    @autobind
+    private handleWheel(e: React.WheelEvent) {
```
alisman: comment would be nice here to explain why this is necessary

src/shared/components/rangeSelector/RangeSelector.tsx
```diff
@@ -0,0 +1,57 @@
...
+    @autobind
+    private onSelection(domain: any, props: any) {
+        if (this.props.updateDomain) {
+            this.props.updateDomain(domain);
+        }
+    }
+
+    render() {
+        return (
+            <div style={{marginBottom:"-50px"}}>
```
alisman: i kind of like using numerical values (-50) instead of stringts

src/pages/resultsView/survival/SurvivalChart.tsx
zhx828: @dippindots also I think it should be typed?

src/appBootstrapper.jsx
```diff
@@ -43,6 +44,9 @@ initializeConfiguration();
 // THIS TELLS WEBPACK BUNDLE LOADER WHERE TO LOAD SPLIT BUNDLES
 __webpack_public_path__ = AppConfig.frontendUrl;
 
+// this fixes a chrome bug related to touch and wheel events
```
zhx828: Could you add a reference to the issue? So later we can clean it up if it's resolved?

src/shared/components/rangeSelector/RangeSelector.tsx
```diff
@@ -0,0 +1,57 @@
...
+    styleOpts: any; // see victory styles, or styleOptsDefaultProps in SurvivalChart for examples
+    xAxisTickCount: number
+    selectedDomain: any;
+    minAxisValue?: number;
+    updateDomain?: (domain: any)=>void;
+}
+
+@observer
+export default class RangeSelector extends React.Component<IRangeSelectorProps> {
+    constructor(props:IRangeSelectorProps) {
```
zhx828: Seems not necessary

src/shared/components/rangeSelector/RangeSelector.tsx
```diff
@@ -0,0 +1,57 @@
...
+        super(props);
+    }
+
+    public static defaultProps = {
+        minAxisValue: 0
+    };
+
+    @autobind
+    private onSelection(domain: any, props: any) {
```
zhx828: @alisman just curious, should we not bind the event at all if the property is not available?

src/shared/lib/handleTouchAndWheelEventsForChromeBug.ts
```diff
@@ -0,0 +1,43 @@
+// This is a temporary workaround until https://github.com/facebook/react/issues/14856 is fixed
+export function handleTouchAndWheelEventsForChromeBug(document: any) {
```
zhx828: Did you write the following code? Or is it coming from an existing solution? Just curious whether we need to review the following file.

src/appBootstrapper.jsx
dippindots: Sure!

src/shared/components/rangeSelector/RangeSelector.tsx
dippindots: I will remove it when creating another pr latter.

src/shared/lib/handleTouchAndWheelEventsForChromeBug.ts
dippindots: Yes, it's coming from an existing solution: https://github.com/facebook/react/issues/14856#issuecomment-478144231

dippindots:
Close since we will change to use `react-rangeslider`.

PR #2802, opened 2019-10-18
Dont filter out profiles without data for selection in coexpression and plots

adamabeshouse:


adamabeshouse:
@jj @pieterlukasse this is what shows up in the coexpression tab now if you select a profile with no data for the query gene. is this okay?

image

adamabeshouse:
after talking with @jjgao it now looks like this

PR #2803, opened 2019-10-18
Error screen when session service is not enabled

kalletlak:
Show error screen when session service is not enabled

Screen Shot 2019-10-18 at 3 31 55 PM

### DIFF
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -15,6 +15,7 @@ import UserMessager from "shared/components/userMessager/UserMessage";
 import {formatError} from "shared/lib/errorFormatter";
 import {buildCBioPortalPageUrl} from "shared/api/urls";
 import ErrorScreen from "shared/components/errorScreen/ErrorScreen";
+import { ServerConfigHelpers } from 'config/config';
 
 interface IContainerProps {
     location: Location;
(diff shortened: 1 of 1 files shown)


src/appShell/App/Container.tsx
```diff
@@ -62,16 +63,29 @@ export default class Container extends React.Component<IContainerProps, {}> {
...
-                       <Then>
-                           <div className="contentWrapper">
-                               <ErrorScreen
-                                    title={"Oops. There was an error retrieving data."}
-                                    body={<a href={buildCBioPortalPageUrl("/")}>Return to homepage</a>}
-                                    errorLog={formatError(this.appStore.undismissedSiteErrors)}
-                               />
-                           </div>
-                       </Then>
+                    <If condition={this.appStore.isErrorCondition || !ServerConfigHelpers.sessionServiceIsEnabled()}>
```
alisman: You'll notice above that we check for isSessionLoaded.   Apparently that will evaluate as true if session service is missing.  Nevertheless, for clarity, I think we should put this new ErrorScreen outside that check.  The render method can just return <ErrorScreen> component and nothing else.

alisman: In other words, at line 49.   sessionServiceDisabled.

kalletlak: updated

PR #2804, opened 2019-10-21
Plots Tab Selection History

Luke-Sikina:
- functionality
    - any changes made to gene, geneset, treatment, or dataSource will
    be remembered if you switch to a different data type and switch back
    - bugfix: changing gene on axis A while having gene set to same on
    axis B no longer has strange behavior if you then switch datatype A
    - bugfix: same scenario as above, swapping axes also works as
    expected
- implementation
    - created PlotsTabSelectionHistory to manage the selection history
    - it maps from each dataType selected to the sub setting selected
    for that datatype
    - when the datatype is changed, it loads the previous settings for
    the new datatype, and calls the related onChange functions
    - the map model is populated as selections are made and is empty by
    default

Fix cBioPortal/cbioportal#6724
Fix cBioPortal/cbioportal#6723

# Checks
- [x] Has tests
- [x] The commit log is comprehensible. 

# Any screenshots or GIFs?
demo
demo-1

### DIFF
--- a/src/pages/resultsView/plots/LastPlotsTabSelectionForDatatype.spec.ts
+++ b/src/pages/resultsView/plots/LastPlotsTabSelectionForDatatype.spec.ts
@@ -1,0 +1,211 @@
+import {assert} from 'chai';
+import LastPlotsTabSelectionForDatatype from './LastPlotsTabSelectionForDatatype';
+import { AxisMenuSelection, MutationCountBy } from './PlotsTab';
+import sinon from 'sinon';
+
+// couldn't figure out how to do this with sinon
+function createFakeCallbacks() {
+    return {
+        gene: sinon.stub(),
+        geneSet: sinon.stub(),
+        source: sinon.stub(),
+        treatment: sinon.stub(),
+    }
+}
+
+// parts of AxisMenuSelection needed to make things compile that are not
+// used in PlotsTabSelectionHistory
+const untestedSelectionFields: AxisMenuSelection = {
+    mutationCountBy: MutationCountBy.MutationType,
+    logScale: false,
+}
+
+describe("PlotsTabSelectionHistory", () => {
+    describe("horizontal updates", () => {
+        it("should update nothing when no changes have been made", () =>{
+            const subject = new LastPlotsTabSelectionForDatatype();
+            const fakes = createFakeCallbacks();
+
+            subject.runHorizontalUpdaters(
+                "foo",
+                fakes.gene,
+                fakes.geneSet,
+                fakes.source,
+                fakes.treatment,
+            )
+
+            assert.equal(fakes.gene.args.length, 0);
+            assert.equal(fakes.geneSet.args.length, 0);
+            assert.equal(fakes.source.args.length, 0);
+            assert.equal(fakes.treatment.args.length, 0);
+        });
+
+        it("should update genes when changes have been made to genes", () => {
+            const subject = new LastPlotsTabSelectionForDatatype();
+            const fakes = createFakeCallbacks();
+            const newSelection: AxisMenuSelection = {
+                dataType: "COPY_NUMBER_ALTERATION",
+                selectedGeneOption: {value: 0, label: "BRAF"},
+                ...untestedSelectionFields,
+            }
+            
+            subject.updateHorizontalFromSelection(newSelection)
+            subject.runHorizontalUpdaters(
+                "COPY_NUMBER_ALTERATION",
+                fakes.gene,
+                fakes.geneSet,
+                fakes.source,
+                fakes.treatment,
+            )
+
+            assert.equal(fakes.gene.args.length, 1);
+            assert.equal(fakes.geneSet.args.length, 0);
+            assert.equal(fakes.source.args.length, 0);
+            assert.equal(fakes.treatment.args.length, 0);
+
+            assert.deepEqual(fakes.gene.args[0], [{value: 0, label: "BRAF"}]);
+        });
+
+        it("should update genesets when changes have been made to genesets", () => {
+            const subject = new LastPlotsTabSelectionForDatatype();
+            const fakes = createFakeCallbacks();
+            const newSelection: AxisMenuSelection = {
+                dataType: "COPY_NUMBER_ALTERATION",
+                selectedGenesetOption: {value: "BRAF", label: "BRAF"},
+                ...untestedSelectionFields,
+            }
+            
+            subject.updateHorizontalFromSelection(newSelection)
+            subject.runHorizontalUpdaters(
+                "COPY_NUMBER_ALTERATION",
+                fakes.gene,
+                fakes.geneSet,
+                fakes.source,
+                fakes.treatment,
+            )
+
+            assert.equal(fakes.gene.args.length, 0);
+            assert.equal(fakes.geneSet.args.length, 1);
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -141,10 +142,10 @@ export type AxisMenuSelection = {
     entrezGeneId?:number;
     genesetId?:string;
     treatmentId?:string;
-    selectedGeneOption?:{value:number, label:string}; // value is entrez id, label is hugo symbol
-    selectedDataSourceOption?:{value:string, label:string};
-    selectedGenesetOption?:{value:string, label:string};
-    selectedTreatmentOption?:{value:string, label:string};
+    selectedGeneOption?:PlotsTabGeneOption; // value is entrez id, label is hugo symbol
```
adamabeshouse: nitpick but lets move the explanatory comment down to the type definition now (i.e. on line 166)

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -160,6 +161,15 @@ export interface IPlotsTabProps {
     store:ResultsViewPageStore;
 };
 
+export type PlotsTabOption = { value: string; label: string };
+export type PlotsTabGeneOption = {
+    value: number,
+    label: string,
+}
+export type PlotsTabDataSource = {
```
adamabeshouse: name is a little off, I think. it's really `PlotsTabDataTypeToSources` (?)

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2366,6 +2409,9 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
     }
 
     public render() {
+        console.log("v: ", this.vertSelection.selectedGeneOption);
+        console.log("h: ", this.horzSelection.selectedGeneOption);
```
adamabeshouse: remove logs

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1863,21 +1921,6 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
         }
     });
 
-    /*readonly mutationProfileDuplicateSamplesReport = remoteData({
```
adamabeshouse: why commented out?

adamabeshouse: lets leave this in and add comment that "in case we want to handle duplicate samples differently"

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
```diff
@@ -0,0 +1,93 @@
...
+
+            if (selectionToUpdate === undefined) {
+                selectionToUpdate = {};
+            }
+
+            axis.set(newSelection.dataType, this.updateSelection(selectionToUpdate, newSelection))
+        }
+    }
+
+    private updateSelection(selectionToUpdate: Selection, newSelection: AxisMenuSelection): Selection {
```
adamabeshouse: super nitpick but since it doesnt use `this` it would be cool for this method to be static

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
```diff
@@ -0,0 +1,93 @@
...
+    gene?: PlotsTabGeneOption | undefined,
+    geneSet?: PlotsTabOption | undefined,
+    source?: PlotsTabOption | undefined,
+    treatment?: PlotsTabOption | undefined,
+
+}
+
+type FieldUpdater = (option: any) => void;
+
+export default class PlotsTabSelectionHistory {
```
adamabeshouse: so the selection history is only per datatype? if so, then I think naming should reflect that

adamabeshouse: also, I think `history` isnt a great name because it doesnt store indefinitely backwards, its only the most recent selection, right? so how about something like `PlotsTabLastSelectionForDatatype`. kind of a mouthful... but still might be the most clear

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
```diff
@@ -0,0 +1,93 @@
...
+        gene: FieldUpdater,
+        geneSet: FieldUpdater,
+        source: FieldUpdater,
+        treatment: FieldUpdater,
+    ) {
+        if (!axis.get(type)) {
+            return;
+        } 
+
+        const tuples = [
```
adamabeshouse: can you rename `tuples` to something that describes what the data is, rather than the data type?

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
```diff
@@ -0,0 +1,93 @@
...
+    private updateSelection(selectionToUpdate: Selection, newSelection: AxisMenuSelection): Selection {
+        selectionToUpdate.gene = newSelection.selectedGeneOption;
+        selectionToUpdate.geneSet = newSelection.selectedGenesetOption;
+        selectionToUpdate.source = newSelection.selectedDataSourceOption;
+        selectionToUpdate.treatment = newSelection.selectedTreatmentOption;
+        return selectionToUpdate;
+    }
+
+    @autobind
+    public updateHorizontalExternalSelection(
```
adamabeshouse: what is this method for? Im confused by the external selection and the updaters thing. maybe more descriptive naming would help, maybe comments?

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
Luke-Sikina: Sure. I went with `LastPlotsTabSelectionForDatatype`. I think that covers your concerns.

src/pages/resultsView/plots/PlotsTabSelectionHistory.ts
Luke-Sikina: I simplified the name a little bit and added some documentation. Hopefully that helps.

src/pages/resultsView/plots/LastPlotsTabSelectionForDatatype.spec.ts
```diff
@@ -0,0 +1,214 @@
...
+import { AxisMenuSelection, MutationCountBy } from './PlotsTab';
+
+// couldn't figure out how to do this with sinon
+function createFakeCallbacks() {
+    const geneCalls: any[] = [];
+    const genesetCalls: any[] = [];
+    const sourceCalls: any[] = [];
+    const treatmentCalls: any[] = [];
+
+    return {
```
alisman: you should indeed use sinon for this.   you use stubs.  
const someStub = new sinon.stub();  
assert.equal(someStub.args.length, 1) // args is an array of length corresponding to number of times stub was invoked.   each element should be the array of parameters passed to function when it was invoked

alisman: sinon also has something called "fakes" which might make this even easier, but we would have to upgrade to 5.0.  not necessary

Luke-Sikina: ah, thanks

PR #2807, opened 2019-10-23
Updated treatment selection in heatmap menu

ngocnn1104:
## Changes:
- Added border-radius to treatment textarea to make it consistent with gene selection textarea;
- Updated CheckedSelect component in OncoprintControls to fix issue [#6717](https://github.com/cBioPortal/cbioportal/issues/6717): "Select all" button should only add the filtered treatments;
- Prevented clearing the search value to keep the filtered treatments list open after selecting an option;
- Added local e2e tests

## GIF:
treatment-selection

### DIFF
--- a/src/public-lib/components/checkedSelect/CheckedSelect.tsx
+++ b/src/public-lib/components/checkedSelect/CheckedSelect.tsx
@@ -1,7 +1,7 @@
 import * as React from "react";
 import ReactSelect from "react-select";
 import autobind from "autobind-decorator";
-import {computed} from "mobx";
+import {computed, observable, action} from "mobx";
 import {observer} from "mobx-react";
 
 import {CheckBoxType, getOptionLabel, getSelectedValuesMap, Option} from "./CheckedSelectUtils";
@@ -23,6 +23,8 @@ type CheckedSelectProps = {
     clearAllLabel?: string | JSX.Element;
     showControls?: boolean;
     height?:number;
+    onInputChange?: (input:string) => void;
+    inputValue?:string;
 };
 
 @observer
@@ -35,7 +37,9 @@ export default class CheckedSelect extends React.Component<CheckedSelectProps, {
         showControls: true,
         checkBoxType: CheckBoxType.STRING
     };
-
+    
+    @observable defaultInputValue = '';
+    
     @computed
     get selectedValues() {
         return getSelectedValuesMap(this.props.value);
@@ -85,6 +89,21 @@ export default class CheckedSelect extends React.Component<CheckedSelectProps, {
     private getOptionLabel(option: Option): JSX.Element
     {
         return getOptionLabel(option, this.selectedValues, this.props.checkBoxType);
+    }
+    
+    @computed get inputValue() {
+        return this.props.inputValue || this.defaultInputValue;
+    }
+    
+    @autobind
+    @action defaultOnInputChange(input:string, options:{action:string}) {
+        // The input value (which is a blank string in the case of action === 'set-value')
+        // will be passed to `this.props.onInputChange` by default without the `if` condition.
+        // This leads to undesirable behaviour so adding the if condition will prevent that.
+        if (options.action !== "set-value") {
+            if (this.props.onInputChange) this.props.onInputChange(input);
+            this.defaultInputValue = input;
+        }
     }
 
     @autobind
@@ -179,6 +198,9 @@ export default class CheckedSelect extends React.Component<CheckedSelectProps, {
                     getOptionLabel={this.getOptionLabel}
                     value={this.props.value}
                     labelKey="label"
+                    onInputChange={this.defaultOnInputChange}
+                    inputValue={this.inputValue}
+                    backspaceRemovesValue={false}
                 />
             </div>
         );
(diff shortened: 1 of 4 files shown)


src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -568,9 +569,30 @@ export default class OncoprintControls extends React.Component<
...
+            });
+        } else {
+            this.filteredTreatmentOptions = this.props.treatmentSelectOptions;
+        }
+    }
+    
+    @autobind onAddAllTreatments() {
+        if (this.filteredTreatmentOptions) {
+            // merge the current selected options with all the filtered ones and remove duplicates
+            this.onSelectTreatments(_.uniqBy([...this.selectedTreatmentsJS, ...this.filteredTreatmentOptions], 'value'))
```
alisman: we try to never use the lodash property shortcuts ("value") because they aren't type checked.  Always use thing=>thing.someProp

ngocnn1104: I see. I have updated this line of code and added some logic to prevent clearing search input value after selecting an option to this PR. Please review it again. Thank you.

src/public-lib/components/checkedSelect/CheckedSelect.tsx
```diff
@@ -86,6 +90,19 @@ export default class CheckedSelect extends React.Component<CheckedSelectProps, {
...
     }
+    
+    @computed get inputValue() {
+        return this.props.inputValue || this.defaultInputValue;
+    }
+    
+    @autobind
+    @action defaultOnInputChange(input:string, options:{action:string}) {
+        // to prevent clearing input value when selecting an option
+        if (options.action !== "set-value") {
```
adamabeshouse: can you explain more whats going on/why this is necessary?

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -207,6 +207,7 @@ export default class OncoprintControls extends React.Component<
     @observable heatmapGenesReady = false;
     @observable private _selectedTreatmentIds: string[] = [];
     private textareaTreatmentText = '';
+    @observable filteredTreatmentOptions = this.props.treatmentSelectOptions;
```
adamabeshouse: its problematic to initialize a variable with the value of props, since the initial props might not be meaningful, it might be an empty array for example.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -567,10 +568,45 @@ export default class OncoprintControls extends React.Component<
             label: d.id,
         }));
     }
-    
+
+    @autobind
+    @action
+    onInputChange(input: string) {
```
adamabeshouse: this isnt really idiomatic to how we've been doing things. the way we would do things is keep an observable for the input, change the input in `onInputChange`, then make `@computed get filteredTreatmentOptions()` return what you compute here.

src/public-lib/components/checkedSelect/CheckedSelect.tsx
ngocnn1104: The change events of input in react-select include: `set-value, input-change, input-blur, menu-close`. On the `set-value` event, an option is selected from the list and a blank string will be updated to the search input by default. We want to prevent clearing the search input value after selecting an option in order to keep the filtered list open for the users to select more than one items from the list.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
ngocnn1104: Thanks for your review. I have updated my PR.

src/public-lib/components/checkedSelect/CheckedSelect.tsx
```diff
@@ -86,6 +90,19 @@ export default class CheckedSelect extends React.Component<CheckedSelectProps, {
...
+            if (this.props.onInputChange) this.props.onInputChange(input);
+            if (this.props.inputValue === undefined) this.defaultInputValue = input;
```
adamabeshouse: I think you should set `this.defaultInputValue` regardless, because choosing this.props.inputValue is done in `@computed get inputValue`

src/public-lib/components/checkedSelect/CheckedSelect.tsx
adamabeshouse: what confuses me is where the blank string stuff is done here. in the call to this.props.onInputChange? if so then this.props.onInputChange should have a different name that makes clear that it always will clear the value. otherwise, you are tailoring a general library function to a very specific use

ngocnn1104: Without the `if` condition, the input value (which will be a blank string in the case of `action === 'set-value'`) will always be passed to `this.props.onInputChange`. This leads to undesirable behaviour so adding the `if` condition will prevent that.

adamabeshouse: oh I see. that makes sense. can you put that in the comment?

Sjoerd-van-Hagen:
Thanks @alisman !

PR #2808, opened 2019-10-23
Fix #6697 colored samples by copy number alteration with treatment data

ngocnn1104:
### Changes:
- Updated the getCnaQueries function to cover the case where horizontal selection data type is "GENERIC_ASSAY" and vertical selection data type is "COPY_NUMBER_ALTERATION".

### Screenshot after the fix:
Screenshot from 2019-10-23 14-44-34

### DIFF
--- a/src/pages/resultsView/plots/PlotsTabUtils.tsx
+++ b/src/pages/resultsView/plots/PlotsTabUtils.tsx
@@ -479,7 +479,7 @@ function scatterPlotCnaLegendData(
         })
         .sortBy((v:number)=>-v) // sorted descending
         .value();
-
+        
     const legendData:any[] = uniqueDispCna.map(v=>{
         const appearance = cnaToAppearance[v as -2|-1|0|1|2];
         return {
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1796,6 +1796,11 @@ export function getCnaQueries(
         && vertSelection.entrezGeneId !== undefined) {
         queries.push({entrezGeneId: vertSelection.entrezGeneId});
     }
+    if (vertSelection.dataType === AlterationTypeConstants.COPY_NUMBER_ALTERATION
+        && horzSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY
```
alisman: @pvannierop @ngocnn1104 similar to other PR, i just want to make sure that we are not unnecessarily equating GENERIC_ASSAY with treatment type.   will this require refactoring down the line?

PR #2811, opened 2019-10-23
SV: remember custom bins and log scale in user preference

kalletlak:
Frontend changes for https://github.com/cBioPortal/cbioportal/issues/6721

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -149,7 +149,9 @@ export type ChartUserSetting = {
         h: number;
     },
     patientAttribute: boolean,
-    filterByCancerGenes?:boolean
+    filterByCancerGenes?: boolean,
+    customBins?: number[],
+    disableLogScale?: boolean
 }
 
 export type StudyPageSettings = {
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2655,6 +2664,7 @@ export class StudyViewPageStore {
     @observable private _defualtChartsDimension = observable.map<ChartDimension>();
     @observable private _defaultChartsType = observable.map<ChartType>();
     @observable private _defaultVisibleChartIds: string[] = [];
+    @observable private _defualtClinicalDataBinFilterSet = observable.map<ClinicalDataBinFilter>();
```
alisman: spelling (defualt)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2629,6 +2632,12 @@ export class StudyViewPageStore {
         this._filterMutatedGenesTableByCancerGenes = true;
         this._filterFusionGenesTableByCancerGenes = true;
         this._filterCNAGenesTableByCancerGenes = true;
+
+        this._clinicalDataBinFilterSet.keys().forEach(chartUniqueKey => {
```
alisman: i'm just curious why we wouldn't just clear the whole filter set here?

kalletlak: this is initialized when clinical attributes are fetched and is required to fetch clinical data. Other option is to reset it to default(_defualtClinicalDataBinFilterSet)

kalletlak:
@zhx828 removed prettier related changes.
Regarding your second point, reset settings on the chart level. This should be addressed in another pr.

PR #2816, opened 2019-10-28
Store trial matches by match type

victoria34:
- Group trial matches by match type
- Use comment icon to replace hover link

image

### DIFF
--- a/src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
+++ b/src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
@@ -1,6 +1,6 @@
 import {
     IArmMatch, IClinicalGroupMatch, IDetailedTrialMatch, IGenomicGroupMatch, IGenomicMatch, ITrial,
-    ITrialMatch, IArm, IDrug
+    ITrialMatch, IArm, IDrug, IGenomicMatchType
 } from "../../../shared/model/MatchMiner";
 import * as _ from 'lodash';
 
@@ -65,8 +65,18 @@ export function groupTrialMatchesByAgeNumerical(armGroup: ITrialMatch[]): IClini
                 positive: positiveCancerTypes,
                 negative: negativeCancerTypes
             },
-            matches: [],
-            notMatches: []
+            matches: {
+                MUTATION: [],
+                CNA: [],
+                MSI: [],
+                WILDTYPE: []
+            },
+            notMatches: {
+                MUTATION: [],
+                CNA: [],
+                MSI: [],
+                WILDTYPE: []
+            }
         };
         const positiveAndNegativeMatches = groupTrialMatchesByGenomicAlteration(matchesGroupedByAge[age]);
         clinicalGroupMatch.matches = positiveAndNegativeMatches.matches;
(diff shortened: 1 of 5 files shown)


src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -78,16 +88,42 @@ export function groupTrialMatchesByAgeNumerical(armGroup: ITrialMatch[]): IClini
...
+        cnaMatches: [],
+        wildTypeMatches: []
+    }; // negative matches
     _.forEach(matchesGroupedByGenomicAlteration, (genomicAlterationGroup, genomicAlteration) => {
         const genomicGroupMatch = formatTrialMatchesByMatchType(genomicAlterationGroup, genomicAlteration);
         // If a genomic alteration(i.e. BRAF V600E) contains a "!",
         // it means this trial cannot be used for the genomic alteration, which is a "NOT" match.
         if(genomicAlteration.includes('!')) {
-            notMatches.push(genomicGroupMatch);
+            if (genomicAlterationGroup[ 0 ][ 'matchType' ] === 'MUTATION') {
```
alisman: these long if/else constructs should be re-written using switch statements

alisman: or even better, make a table with match type to the corresponding array and then you can just look it up and push into it.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -140,15 +176,26 @@ export function calculateTrialPriority(armMatches: IArmMatch[]): number {
...
             return 2; // A trial only has negative matches.
         }
         return 1; // A trial has both positive matches and negative matches.
     }
     return 0; // A trial only has positive matches.
 }
 
+export function getMatchesLength(genomicMatchType: IGenomicMatchType): number {
+    let length = 0;
+    length += genomicMatchType.mutationMatches.length;
```
alisman: use lodash _.sum(...)

victoria34: Updated. Thanks for pointing it out.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
victoria34: Sorry, I'm confused. Do you mean to create tables for different type of trial matches? If yes, that's not what curators want. They want to see trial matches grouped by trial id instead of match type.

src/pages/patientView/trialMatch/TrialMatchTableUtils.ts
```diff
@@ -140,15 +158,22 @@ export function calculateTrialPriority(armMatches: IArmMatch[]): number {
...
+        if ( matchesLength === 0) {
             return 2; // A trial only has negative matches.
         }
         return 1; // A trial has both positive matches and negative matches.
     }
     return 0; // A trial only has positive matches.
 }
 
+export function getMatchesLength(genomicMatchType: IGenomicMatchType): number {
+    return _.sum([genomicMatchType.MUTATION.length, genomicMatchType.CNA.length,
```
alisman: you don't have to make this change, but why not just take sum of any array found in the collection?  you could do this with _.chain(genomicMatchType).map().sumBy(a=>a.length).
again, you won't have to worry in future about chaning this code.

PR #2817, opened 2019-10-28
Enable Patient GA Tracking

Luke-Sikina:
Fix cBioPortal/cbioportal#6751
changes proposed in this pull request:
- When the PatientViewPage is created, it logs the study id to Google
Analytics

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -53,6 +53,9 @@ import 'cbioportal-frontend-commons/styles.css';
 import 'react-mutation-mapper/dist/styles.css';
 import 'react-table/react-table.css';
 import getBrowserWindow from "../../public-lib/lib/getBrowserWindow";
+import { trackPatient, trackEvent } from "shared/lib/tracking";
+import PatientViewUrlWrapper from "./PatientViewUrlWrapper";
+import { PagePath } from "shared/enums/PagePaths";
 import { GeneFilterOption } from "./mutation/GeneFilterMenu";
 import { checkNonProfiledGenesExist } from "./PatientViewPageUtils";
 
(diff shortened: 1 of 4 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -135,28 +161,28 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
         this.onCnaTableColumnVisibilityToggled = this.onCnaTableColumnVisibilityToggled.bind(this);
     }
 
+    @computed
+    private get studyIdAndCaseId():string {
```
alisman: i would just put this stuff up in the reaction.

src/pages/patientView/PatientViewUrlWrapper.ts
```diff
@@ -0,0 +1,33 @@
...
+    constructor(routing: ExtendedRouterStore) {
+        super(routing, [
+            { name:"studyId", isSessionProp: false },
+            { name:"caseId", isSessionProp: false },
+            { name:"sampleId", isSessionProp: false },
+        ]);
+    }
+
+    @computed
+    public get asParams(): {[key: string]: string} {
```
alisman: what's this for?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -85,19 +88,44 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
     constructor(props: IPatientViewPageProps) {
 
         super(props);
+        this.urlWrapper = new PatientViewUrlWrapper(props.routing);
 
+        if (
+            this.props.routing.location.pathname.includes("/" + PagePath.Patient) &&
+            this.urlWrapper.query.studyId
+        ) {
+            trackPatient(this.urlWrapper.query.studyId);
```
alisman: i'm guessing you call trackPatient here because the reaction below only fires on subsequent changes?  If that's the case, look at option fireImmediately that reaction takes (https://mobx.js.org/refguide/reaction.html).

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -85,19 +88,44 @@ export default class PatientViewPage extends React.Component<IPatientViewPagePro
...
+        }
+        
+        reaction(
+            () => [this.studyIdAndCaseId],
+            ([_]) => {
+                if (
+                    this.props.routing.location.pathname.includes("/" + PagePath.Patient) &&
```
alisman: how come we do this test?

Luke-Sikina: I saw this test further down when the routing store is populated from the url. The logic there was that you didn't want to do that stuff if this isn't a patient view. Along the same lines, I felt that if this wasn't a patient view, we didn't want to log that we had visited a patient.

src/pages/patientView/PatientViewPage.tsx
Luke-Sikina: Hmm. I was having trouble with the reaction firing last week, but now it seems to work fine with a simpler implementation, so yes, I will do this.

PR #2818, opened 2019-10-29
Show hidden axis (BoxScatterPlot)

dippindots:
Fix cBioPortal/cbioportal#6727

- a add style in axis for BoxScatterPlot forcing to show all the lines (lines for first category will be hide automatically)
image


dippindots:
@alisman the line was hiding because the first data point was treated as `zero`, and victory chart will hide the line for `zero` by default.
Another thing is the e2e tests cannot catch the change like this (the light color change).

PR #2820, opened 2019-10-30
Adjusted buttons in Oncoprint controls

ngocnn1104:
## Changes:
- Removed custom padding and margin of "minimap" button to make its height the same as other buttons
- Removed the `span` between "Add clinical tracks" and "Sort" buttons when study does not contain molecular profiles
- Updated remote screenshot references

### Before:
Screenshot from 2019-10-30 14-08-41

### After:
Screenshot from 2019-10-30 14-06-03


src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -1551,12 +1551,11 @@ export default class OncoprintControls extends React.Component<
                     <Button
                         active={this.showMinimap}
                         onClick={this.toggleShowMinimap}
-                        className="oncoprint__controls__minimap_button"
```
adamabeshouse: why remove this?

pvannierop: Because the only reference to this class is a removed entry in `src/shared/components/oncoprint/controls/styles.scss`

adamabeshouse: fair enough

PR #2823, opened 2019-10-31
Added a new property to turn on/off GSVA features

ngocnn1104:
### Changes:
- Added `skin_show_gsva` to frontendConfig with default value = `false`;
- When there is a need to show the GSVA feature, go to `src/config/serverConfigDefaults.ts` and change `skin_show_gsva: false,` to `skin_show_gsva: true,`;
- Updated local e2e test with a function that manually updates `skin_show_gsva` in frontendConfig to `true`.

### DIFF
--- a/end-to-end-test/shared/specUtils.js
+++ b/end-to-end-test/shared/specUtils.js
@@ -73,6 +73,12 @@ function sessionServiceIsEnabled() {
     }).value;
 }
 
+function showGsva() {
+    browser.execute(function() {
+        window.frontendConfig.serverConfig.skin_show_gsva = true;
+    });
+}
+
 function waitForNumberOfStudyCheckboxes(expectedNumber, text) {
     browser.waitUntil(()=>{
         var ret = browser.elements('[data-test="cancerTypeListContainer"] > ul > ul').value.length === expectedNumber;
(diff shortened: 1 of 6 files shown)


end-to-end-test/shared/specUtils.js
```diff
@@ -67,6 +67,12 @@ function sessionServiceIsEnabled() {
     }).value;
 }
 
+function showGsva() {
```
alisman: does this actually work!?

src/shared/components/query/MolecularProfileSelector.tsx
```diff
@@ -26,11 +27,15 @@ const styles = styles_any as {
 @observer
 export default class MolecularProfileSelector extends QueryStoreComponent<{}, {}>
 {
+	private get showGSVA() {
```
alisman: We should only interface with configuration via AppConfig.  So this would be AppConfig.serverConfig.skin_show_gsva.  
This MIGHT make it harder to change configuration dynamically in test.  It could still work the way you have it in the spec up above.

end-to-end-test/shared/specUtils.js
alisman: ah, i see, down below you are directly referencing in app code.  see note below

src/shared/components/query/MolecularProfileSelector.tsx
ngocnn1104: @alisman Thanks. I have updated the PR. So when should we use `getBrowserWindow()`?

Sjoerd-van-Hagen:
@alisman what still needs to be done to get this one merged?

src/shared/components/query/MolecularProfileSelector.tsx
alisman: ideally never!  it's a global and should be avoided.   it's in the code base as a convenience to avoid having to pass, say, the routing store into the view layers.  but really it's not a good practice.   if you for some reason you find yourself needing to use it, let me know.

PR #2826, opened 2019-11-01
Replace entrezGeneId with hugoGeneSymbol in fusion gene api

kalletlak:
And fix errors

### DIFF
--- a/src/pages/studyView/StudyViewUtils.spec.ts
+++ b/src/pages/studyView/StudyViewUtils.spec.ts
@@ -151,7 +151,7 @@ describe('StudyViewUtils', () => {
                     }]
                 }],
                 mutatedGenes: [{ "hugoGeneSymbols": ["GENE1"] }],
-                fusionGenes: [{ "entrezGeneIds": [1] }],
+                fusionGenes: [{ "hugoGeneSymbols": ["GENE1"] }],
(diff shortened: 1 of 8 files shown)


src/pages/studyView/StudyViewUtils.spec.ts
```diff
@@ -151,7 +151,7 @@ describe('StudyViewUtils', () => {
                     }]
                 }],
                 mutatedGenes: [{ "hugoGeneSymbols": ["GENE1"] }],
-                fusionGenes: [{ "entrezGeneIds": [1] }],
+                fusionGenes: [{ "hugoGeneSymbols": ["GENE1"] }],
```
zhx828: Should we use a different gene for fusion?

src/shared/api/generated/CBioPortalAPIInternal-docs.json
```diff
@@ -1797,8 +1793,7 @@
       "required": [
         "counts",
         "entrezGeneId",
-        "hugoGeneSymbol",
-        "pValue"
```
zhx828: where is the pvalue?

kalletlak: not sure. I just updated the client.

src/pages/studyView/StudyViewUtils.spec.ts
kalletlak: I think it's fine.

PR #2830, opened 2019-11-02
Add drug icon and trial principal investigator

victoria34:
- Add drug icon and trial principal investigator
- Display sample ids individually

image


alisman:
@zhx828 @victoria34 is there a reason that the matchminer api doesn't have swagger/generated clients?  Do we control it?

victoria34:
@alisman I'm using a python project called matchminer-api to build the API. It didn't have swagger.json at that time. I'll check with MatchMiner team to see if they have swagger now.

PR #2837, opened 2019-11-04
Persistent API Cache

Luke-Sikina:
Fix #6635

- Persistently cache endpoints that
    - Are relatively static
    - Are either expensive, or have a large response
- Added new dependency that abstracts IndexedDB into a key value store
- The persistent cache works the same way the existing cache works, replacing the existing methods with ones that check the persistent cache before doing a request
- Invalidation

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -11562,6 +11562,11 @@ icss-utils@^4.1.0:
   dependencies:
     postcss "^7.0.14"
 
+idb-keyval@~3.2.0:
+  version "3.2.0"
+  resolved "https://registry.yarnpkg.com/idb-keyval/-/idb-keyval-3.2.0.tgz#cbbf354deb5684b6cdc84376294fc05932845bd6"
+  integrity sha512-slx8Q6oywCCSfKgPgL0sEsXtPVnSbTLWpyiDcu6msHOyKOLari1TD1qocXVCft80umnkk3/Qqh3lwoFt8T/BPQ==
+
 ieee754@^1.1.12, ieee754@^1.1.4:
   version "1.1.12"
   resolved "https://registry.yarnpkg.com/ieee754/-/ieee754-1.1.12.tgz#50bf24e5b9c8bb98af4964c941cdb0918da7b60b"
--- a/package.json
+++ b/package.json
@@ -153,6 +153,7 @@
     "git-revision-webpack-plugin": "^3.0.3",
     "html-webpack-plugin": "3.2.0",
     "igv": "2.3.5",
+    "idb-keyval": "~3.2.0",
     "imports-loader": "^0.8.0",
     "jStat": "^1.7.0",
     "javascript-natural-sort": "^0.7.1",
--- a/packages/cbioportal-frontend-commons/package.json
+++ b/packages/cbioportal-frontend-commons/package.json
@@ -37,6 +37,7 @@
   "dependencies": {
     "autobind-decorator": "^2.1.0",
     "classnames": "^2.2.5",
+    "idb-keyval": "~3.2.0",
     "jquery": "^3.2.1",
     "lodash": "^4.17.11",
     "measure-text": "0.0.4",
--- a/packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
@@ -14,7 +14,7 @@ function hash(str: string) {
     return hash;
 }
 
-function getHash(obj: any) {
+export function getHash(obj: any) {
     return hash(JSON.stringify(obj));
 }
 
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -52,6 +52,7 @@ export interface IServerConfig {
     pubmed_url: string | null;
     priority_studies: string | null;
     api_cache_limit: number;
+    enable_persistent_cache: boolean;
     show_hotspot: boolean | undefined;
     show_oncokb: boolean;
     show_civic: boolean;
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -36,6 +36,10 @@ import {
 import AppConfig from 'appConfig';
 import { sendSentryMessage } from '../shared/lib/tracking';
 import { log } from '../shared/lib/consoleLog';
+import {
+    EndpointUpdateInfo,
+    cacheMethods,
+} from 'shared/api/persistentApiClientCache';
 
 const config: any = (window as any).frontendConfig || { serverConfig: {} };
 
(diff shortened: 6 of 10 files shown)


src/public-lib/lib/persistantApiClientCache.ts
```diff
@@ -0,0 +1,52 @@
...
+                                included: string[],
+                                log?: (message: string) => void,
+                                sentryLog?: (message: string) => void
+): void {
+    //functionality check
+    if (!('caches' in self)) {
+        return;
+    }
+
+    Object.getOwnPropertyNames(api.prototype)
```
alisman: I find this to be a confusing sequence.  I prefer the style of the existing cachePostMethod in src/public-lib/lib/apiClientCache.ts
The function is a little long, but you see that oldMethod is stored in the closure and then invoked in the new method if there isn't anything in cache

src/public-lib/lib/persistantApiClientCache.ts
```diff
@@ -0,0 +1,52 @@
+type NameMethodPair = {
+    name: string,
+    method: any,
+}
+
+function createCacheCheckingVersionOfGet(pair: NameMethodPair, api: any): NameMethodPair {
+    const method = function(parameters: any) {
```
alisman: the interceptor is going to need to return a promise whether the data exists or not, since that's the original method's response type.    And it will need to intercept the response of the original method in order to put the response in cache.

alisman:
When you complete this PR, lets actually implement a simple cache expiration policy.  Like, it's cached as long as key is in local storage.  Then we can clear it easily.

src/public-lib/lib/persistentApiClientCache.spec.ts
```diff
@@ -0,0 +1,222 @@
+import { assert } from 'chai';
+import sinon from 'sinon';
+import { cacheGetMethods, clearPersistentCache } from './persistentApiClientCache';
+import { sleep } from './../../shared/lib/TimeUtils'; // normal file path was causing buildPublicLib to break?
```
Luke-Sikina: @alisman Any idea why I had to do this? Previously it was `shared/lib/TimeUtils`, and it built fine, but `yarn run buildPublicLib` failed.

yarn.lock
```diff
@@ -7882,6 +7882,11 @@ icss-utils@^4.1.0:
   dependencies:
     postcss "^7.0.14"
 
+idb-keyval@~3.2.0:
```
Luke-Sikina: All I needed was a persistent key value store, so I used [this](https://github.com/jakearchibald/idb-keyval)
It looks well maintained and popular enough. I pinned it to the minor version-  hopefully that's enough to keep things stable.

alisman: sounds good

src/public-lib/lib/persistentApiClientCache.ts
```diff
@@ -0,0 +1,130 @@
...
+    const method = async function(parameters: any): Promise<request.Response> {
+        // Wait for timestamps to finish loading to avoid race condition
+        // where the persistent cache would sometimes be ignored
+        if (endpointUpdateTimes instanceof Promise) {
+            await endpointUpdateTimes;
+        }
+
+        // for every GET in the api, there is a corresponding
+        // GETURL function that returns the url as a string
+        const urlMethod = constructUrlMethodName(name, log);
```
alisman: i wanted to float the idea of just hashing the parameters (as we do in existing memory cache).  You could potentially add the timestamp to that has (would this obviate the need to store the timestamps elsewhere?).  The other possible benefit is that it would work for both post and get requests.

Luke-Sikina: Ok, I'm interested. I'll give it a shot. I don't like hashes because it makes understanding exactly what is in the cache difficult, but the added utility of being able to cache posts is probably worth it.

src/public-lib/lib/persistentApiClientCache.spec.ts
alisman: sorry i missed this.  my guess is that it has to do with aliases.    in main project we have alias set up for "shared" that resolve to a to real path.    Probably there's no such alias set up in the public lib config and so it doesn't work in that context.  just a guess though! (@Onur?)

Luke-Sikina: Okay, I took some time and figured this out. I added the shared alias for the public lib, since that dependency should be acceptable. I also realized that I was breaking the public lib build with the AppConfig import for related reasons. That import doesn't belong in the public lib, so I pulled the check out into `config.ts` and now everything builds

commons.webpack.config.js
```diff
@@ -95,16 +95,9 @@ var config = {
...
-        //     css: join(src, 'styles'),
-        //     containers: join(src, 'containers'),
-        //     components: join(src, 'components'),
-        //     utils: join(src, 'utils'),
-        //     styles: join(src, 'styles'),
-        //     pages: join(src, 'pages'),
-        //     shared: join(src, 'shared'),
-        //     appConfig: path.join(__dirname + '/src', 'config', ((process.env.NODE_ENV === 'test')? 'test.' : '') + 'config')
-        // }
+        alias: {
```
alisman: how come we restored this?  just wondering.

Luke-Sikina: I import `import internalClient from 'shared/api/cbioportalInternalClientInstance';` in `persistentApiClientCache.ts`. I didn't see any harm in aliasing shared since it's shared and it makes the imports look more like they do in the main app.

alisman: @onursumer we want to import the clients from the common library now, right?  
I also don't love the alias (even though i set it up originally myself) because I don't think IDE's can deal with them.  At least mine doesn't.  I.e. it will always try to import using relative path.

onursumer: @alisman we haven't moved cbioportal API client under the commons library because we don't use them anywhere else yet. So, they are still under `shared`. But right, ideally, we should import API clients from the `cbioportal-frontend-commons` package.

src/public-lib/lib/persistentApiClientCache.ts
```diff
@@ -0,0 +1,151 @@
...
+            }
+            return await oldMethod.apply(this, [parameters]);
+        }
+        const url = name + '?' + getHash(parameters);
+        // check the cache for a matching url
+        var response: TimestampedReponse = await get(
+            PERSISTENT_CACHE_PREFIX + url
+        );
+
+        // if not in cache, fetch a response and cache it
```
alisman: we are most likely ok here, but there is a subtle difference between this and the other cache we use: the other cache returns the actual promise that is returned by the "oldMethod."  That makes me more confident that it's a silent "spy" on the original method being proxied.   here we are returning a new promise which will resolve with a response.  i would just want to make double sure that errors errors are still caught properly by calling code.

Luke-Sikina: Reminder to self: still haven't done this. Do tomorrow.

PR #2706, opened 2019-09-17
Upgraded react-mutation-mapper version to 0.3.0

onursumer:
Before:
mutations_tab_current

After:
mutation-mapper-multi-filter

Fix cBioPortal/cbioportal/issues/5338

# Checks
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Airbnb React/JSX Style guide](https://github.com/airbnb/javascript/tree/master/react).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)

PR #2840, opened 2019-11-05
fix gnomad link url

leexgh:
generate gnomad url by vcf

fix: https://github.com/cBioPortal/cbioportal/issues/6770


onursumer:
@leexgh @inodb We should apply these changes to `react-mutation-mapper` as well. Ideally, we should directly import the component from `react-mutation-mapper` to avoid duplicate code management. I will apply these changes to `react-mutation-mapper` and open a PR to address this issue after we merge #2706

PR #2842, opened 2019-11-06
Gene selection and radio buttons when displaying clinical attributes with treatment response

ngocnn1104:
Resolves cBioPortal/cbioportal#6668
Resolves cBioPortal/cbioportal#6547

## Changes:
- Showing gene selection box and radio buttons above all plots excluding bar charts & tables, keeping the styles of selection box and radio buttons consistent and adding the third button to show `Both` types (to style samples by both Mutation and Copy number);
- A 'None' option is added to gene drop-down, which removes samples styling;
- The selection of gene drop-down in utilities menu is controlled both by itself and by the two axes - but not the other way around, i.e. selecting genes from axes will change selected gene in utilities menu, while selecting genes in utilities menu won't affect gene selection in axes;
- Fixed alignments of radio buttons and their labels;
- Fixed plot data for download not reflecting the alterations selected for the plot;
- When there are different genes in axes, display mutation and CNA types options and keep stylings consistent in all box plots and scatter plots;
- Showing selected gene for styling as legend title.

### Before:
Screenshot from 2019-11-06 10-56-02

### After:
Peek 2020-01-14 12-35


pvannierop:
@ngocnn1104 Nice work! I thing that there should be an element in the plot that reflects the currently selected gene in the utilities menu when the user exports the image for publication. In the waterfall plot we added a title that (among other things) shows the gene used to style the data points.

image

I think we can do the same for the clinical vs treatment plot.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2004,6 +2024,14 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
             ));
         }
     });*/
+    
+    @computed get selectedGene() {
```
pvannierop: Change to `@computed get selectedGene():Gene` and update import.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2138,7 +2161,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
         invoke: ()=>{
             const horzAxisData = this.horzAxisDataPromise.result;
             const vertAxisData = this.vertAxisDataPromise.result;
-            if (!horzAxisData || !vertAxisData) {
+            if (!horzAxisData || !vertAxisData || (this.showClinicalAttrsWithTreatments && !this.horzGeneOptions.isComplete)) {
```
pvannierop: It is correct like it is, but to make it more readable it should state:
```
if (!horzAxisData || !vertAxisData || (this.showClinicalAttrsWithTreatments && !this.utilityMenuGeneOptions.isComplete)) {
```
This means that `this.utilityMenuGeneOptions` should be converted into a `readonly utilitiyMenuGeneOptions = remoteData...` construct.

ngocnn1104: @pvannierop Thanks for your input! I have updated the PR.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -211,6 +212,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
     @observable searchMutationInput:string;
     @observable viewMutationType:boolean = true;
     @observable viewCopyNumber:boolean = false;
+    @observable viewMutationAndCNA:boolean = false;
```
adamabeshouse: why dont you want to derive this as `@computed`: `viewMutationType && viewCopyNumber`?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -387,15 +389,33 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
         if (this.limitValuesCanBeShown) {
             return PotentialViewType.LimitVal;
         }
+        
 
         // neither axis gene or treatment
         return PotentialViewType.None;
     }
+    
+    @computed get showClinicalAttrsWithTreatments() {
```
adamabeshouse: I dont love the name. I think since its a computed and not an action, the verb should be the 'ing' tense, rather than imperative. Also neither clinical attr nor treatment should be plural.

overall I would suggest the name `isClinicalAttrVsTreatment`

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2004,6 +2031,14 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
             ));
         }
     });*/
+    
+    @computed get selectedGene():Gene {
```
adamabeshouse: since this is specifically for treatment profiles, the name should reflect that. e.g. `selectedGeneForTreatment`

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2206,6 +2237,30 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
         return ( (this.vertSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY || this.horzSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY)
                   && this.vertTreatmentOptions.length === 0);
     }
+    
+    @computed get showUtilitiesMenu() {
+        return (this.plotDataExistsForTwoAxes || this.waterfallPlotIsShown) && (this.mutationDataCanBeShown || this.cnaDataCanBeShown);
+    }
+    
+    @computed get sharedPlotTitle():string {
```
adamabeshouse: I would like for the name to be more descriptive. can you explain more what the purpose of this title is?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2408,7 +2465,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
                                                             value={this.utilitiesMenuSelection.selectedGeneOption ? this.utilitiesMenuSelection.selectedGeneOption.value : undefined}
                                                             onChange={this.onUtilitiesGeneSelect}
                                                             isLoading={this.horzGeneOptions.isPending}
-                                                            options={this.utilityMenuGeneOptions}
+                                                            options={this.utilityMenuGeneOptions.result}
```
adamabeshouse: looks like `this.utilityMenuGeneOptions` is now a mobxpromise, so it should be inserted into the `promises` array on line 2266 to be awaited

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1530,10 +1534,11 @@ export function makeScatterPlotData(
         const sampleCoverageInfo = coverageInformation[d.uniqueSampleKey];
         let profiledMutations = undefined;
         let profiledCna = undefined;
+        const noGenesInAxes = !horzData.hugoGeneSymbol && !vertData.hugoGeneSymbol;
         if (mutations || copyNumberAlterations) {
             const profiledReport = makeScatterPlotData_profiledReport(
-                horzData.hugoGeneSymbol,
-                vertData.hugoGeneSymbol,
+                (noGenesInAxes && selectedGene) ? selectedGene.hugoGeneSymbol : horzData.hugoGeneSymbol,
```
adamabeshouse: this will still return horzData.hugoGeneSymbol even if vertData.hugoGeneSymbol is set but horzData.hugoGeneSymbol is not. is that what you're going for?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1838,6 +1845,11 @@ export function showWaterfallPlot(horzSelection:AxisMenuSelection, vertSelection
     || (horzSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY && vertSelection.dataType === NONE_SELECTED_OPTION_STRING_VALUE);
 }
 
+export function showPlotWithClinicalAttrsAndTreatments(horzSelection:AxisMenuSelection, vertSelection:AxisMenuSelection):boolean {
```
adamabeshouse: you can call this function in the `@computed` in `PlotsTab` with the same logic, so as not to duplicate logic

jjgao:
@ngocnn1104 somehow I don't see the changes in the heroku instance... https://cbioportal-f-plotstab-r-bnr3o6.herokuapp.com/results/plots?session_id=5dc30143e4b04a9c23e2743e

ngocnn1104:
> @ngocnn1104 somehow I don't see the changes in the heroku instance... https://cbioportal-f-plotstab-r-bnr3o6.herokuapp.com/results/plots?session_id=5dc30143e4b04a9c23e2743e

@jjgao Select Clinical Attribute for one axis and Treatment Response for the other, then you will see the changes.

src/pages/resultsView/plots/PlotsTab.tsx
ngocnn1104: @adamabeshouse oh I just didn't think of it in the first place. Thanks for pointing it out.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2206,6 +2237,30 @@ export default class PlotsTab extends React.Component<IPlotsTabProps,{}> {
...
+    
+    @computed get showUtilitiesMenu() {
+        return (this.plotDataExistsForTwoAxes || this.waterfallPlotIsShown) && (this.mutationDataCanBeShown || this.cnaDataCanBeShown);
+    }
+    
+    @computed get sharedPlotTitle():string {
+        if (this.showClinicalAttrsWithTreatments) {
+            const selection = this.horzSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY ? this.horzSelection: this.vertSelection;
+            const treatmentName = selection.selectedTreatmentOption!.label;
+            const profileName = selection.selectedDataSourceOption!.label;
```
pvannierop: For a clinical vs treatment profile scatter plot the title should reference both axes.  I may get very long, but just see and discuss when we see the result.

src/pages/resultsView/plots/PlotsTab.tsx
ngocnn1104: As Pim suggested, there should be a title to describe the currently selected gene, and this title will go along with the svg in the exported file when users want to download the plot. I named it `sharedPlotTtile` because it appears in both the scatter plot and the box plot.

src/pages/resultsView/plots/PlotsTab.tsx
ngocnn1104: The result of `utilityMenuGeneOptions` is only expected in a specific case, and I don't think it should affect the display of the entire plot. When the promise is pending, the ReactSelect component that shows the gene options will be in a loading state and I think that might be enough.

src/pages/resultsView/plots/PlotsTabUtils.tsx
ngocnn1104: Yes it is already the case before. The type definition of the first 2 arguments passed to `makeScatterPlotData_profiledReport` is `string|undefined`, so it is okay for one axis to have gene data and the other does not, or both axes does not have any selected gene. I added the condition to account for when gene is selected from the utilities menu above the plot rather than from the axis menu on the side.

src/pages/resultsView/plots/PlotsTab.tsx
ngocnn1104: The plot title becomes long indeed, and there are cases when it is overflowing outside of the svg area. Unfortunately, there is no text-wrap property in svg, so I'm still trying to figure out how to dynamically breaks the title to multiple lines based on the svg width.

A screenshot for reference:
Screenshot from 2019-11-07 12-52-58

ngocnn1104: So here's what it looks like:
Screenshot from 2019-11-07 13-36-27

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: okay that makes sense to me.

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1494,7 +1497,8 @@ export function makeScatterPlotData(
     copyNumberAlterations?:{
         molecularProfileIds:string[],
         data:AnnotatedNumericGeneMolecularData[]
-    }
+    },
+    selectedGene?:Gene
```
adamabeshouse: I think this parameter name (`selectedGene`) needs to be clearer. is this specifically for treatments?

if so then the condition on line 1540 and 1541 should be specifically linked to treatments, not just if there are no genes selected.

jjgao:
@ngocnn1104 cool, now I see it. Thanks!

@Sjoerd-van-Hagen @ngocnn1104 I am wondering if you can also apply this to other plots as discussed in cBioPortal/cbioportal#6668, if it's not too much work? That way, it can be useful for all data types and studies.

jjgao:
@ngocnn1104 btw, could you please use [github closing keywords](https://help.github.com/en/github/managing-your-work-on-github/closing-issues-using-keywords) so that the corresponding issue will be closed once the pr is merged?

src/pages/resultsView/plots/PlotsTabUtils.tsx
ngocnn1104: @adamabeshouse At the moment, yes this parameter is specifically for treatment vs clinical attribute plot. When calling the function in `PlotsTab`, I already added the condition `this.isClinicalAttrVsTreatment ? this.selectedGeneForTreatment : undefined`, so I thought it wasn't necessary to add condition inside the function's logic.

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: I think the name is too vague, `shared` doesn't really tell a reader what it is. how about `scatterAndBoxTitle`?

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: ah okay. in that case, I believe the parameter name should definitely contain this information, named something like `selectedGeneForClinicalAttrVsTreatment`

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: as you mentioned that selectedGene is only passed through when its a clinical attr vs treatment plot, aka when there are no genes in axes, isnt the `noGenesInAxes` check redundant? cant you just check whether the `selectedGene` parameter is present?

ngocnn1104:
> @Sjoerd-van-Hagen @ngocnn1104 I am wondering if you can also apply this to other plots as discussed in [cBioPortal/cbioportal#6668](https://github.com/cBioPortal/cbioportal/issues/6668), if it's not too much work? That way, it can be useful for all data types and studies.

Yes I can implement it. However, currently for other non-waterfall plots, check-boxes are used and they allow for selecting none, one or all of the options. If we replace them with radio buttons now, please be aware that it is not normal behavior to deselect all of the options because they are mutually exclusive. Are you sure you want that?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1530,10 +1534,11 @@ export function makeScatterPlotData(
         const sampleCoverageInfo = coverageInformation[d.uniqueSampleKey];
         let profiledMutations = undefined;
         let profiledCna = undefined;
+        const noGenesInAxes = !horzData.hugoGeneSymbol && !vertData.hugoGeneSymbol;
         if (mutations || copyNumberAlterations) {
             const profiledReport = makeScatterPlotData_profiledReport(
-                horzData.hugoGeneSymbol,
-                vertData.hugoGeneSymbol,
+                (noGenesInAxes && selectedGeneForClinicalAttrVsTreatment) ? selectedGeneForClinicalAttrVsTreatment.hugoGeneSymbol : horzData.hugoGeneSymbol,
```
adamabeshouse: isn't it redundant to check if no genes in axis? isn't it true that theres definitely no genes in axis if you're passing in a gene for clinical attr vs treatment plot?

ngocnn1104: Last time I checked it's true that for clinical vs treatment plot, there are no genes in both axes. I agree that it's a bit redundant, perhaps checking if `selectedGeneForClinicalAttrVsTreatment` exists is good enough because this parameter is passed as an argument only in this specific case. I will remove the `noGenesInAxes` condition.

alisman:
@ngocnn1104 @jjgao @pvannierop @Sjoerd-van-Hagen 
1. Could we unify the following looks.  Also, i'm not sure I like the idea of UI appearing inside the chart container (dotted line) when it could as easily fit above the container.  What do you guys think?

image

image

pvannierop:
@alisman @Sjoerd-van-Hagen The grey background is a remnant of the _limit values_ checkbox that a time ago also was displayed in the utilities menu. So I agree that when you mean with 'unifying' you mean remove the background  color, I agree.

jjgao:
> @jjgao Yes I can implement it. However, currently for other non-waterfall plots, check-boxes are used and they allow for selecting none, one or all of the options. If we replace them with radio buttons now, please be aware that radio buttons don't allow deselecting all of the options because they are mutually exclusive. Are you sure you want that?

@ngocnn1104 that's a good point, I think we should give the open of selection none in the gene dropdown.

ngocnn1104:
@alisman @Sjoerd-van-Hagen @pvannierop I have removed the background of the utilities menu to unify the looks.

@jjgao Could you elaborate your idea a bit more? I wasn't talking about the gene selection but the styling of samples in the plots. With the radio buttons, samples will always be styled either by "Mutations", "CNA" or "Both", while currently with the checkboxes, samples could be styled by none of those things because checkbox options can be deselected. Does removing the styling of samples in the plots mean anything important here? If not then we can just forget about this concern.

jjgao:
@ngocnn1104 if we give the option to select no genes, that will be equivalent to uncheck both mutation and cna.

ngocnn1104:
> @ngocnn1104 if we give the option to select no genes in the genes dropdown menu, that will be equivalent to uncheck both mutation and cna.

I have added a `None` option to the genes dropdown. Please check it in the deployment when it's finished.

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1833,6 +1843,11 @@ export function showWaterfallPlot(horzSelection:AxisMenuSelection, vertSelection
     || (horzSelection.dataType === AlterationTypeConstants.GENERIC_ASSAY && vertSelection.dataType === NONE_SELECTED_OPTION_STRING_VALUE);
 }
 
+export function showClinicalVsTreatmentPlot(horzSelection:AxisMenuSelection, vertSelection:AxisMenuSelection):boolean {
```
alisman: i think this name is a bit confusing.   isClinicalVsTreatmentPlot?

alisman:
@pvannierop @ngocnn1104 lets swash commits here and then you can rebase and merge.  ino and I did some test and looks good.

jjgao:
Thanks, @ngocnn1104. I am wondering if you could also apply this to other plots as discussed in cBioPortal/cbioportal#6668 and above?

ngocnn1104:
@jjgao I have implemented it. Please review again, thanks.

jjgao:
Thanks, @ngocnn1104. Let's apply your new menu to all charts including the ones with molecular data. Or do you see any reason we should not?

jjgao:
@ngocnn1104 it looks great! Thank you. 

A couple of comments:

- can we change "Style samples by:" to "Color samples by:"
- Is it possible to increase the width of the drop-down? If necessary we can increase the width of plot section.
- When there are two different genes for the axes, let's just use the exactly same menu as when there is one gene for both axes - I don't think we need to plot the two genes simultaneously now that we can switch genes fro styling.

image

ngocnn1104:
> * When there are two different genes for the axes, let's just use the exactly same menu as when there is one gene for both axes - I don't think we need to plot the two genes simultaneously now that we can switch genes for coloring.

@jjgao Then I think we can just use the same menu for all situations, which includes a `None` option and all of the genes. Adding the `Same gene` option wouldn't add much value to the dropdown and also would cause confusion if the users see 2 `Same gene` options when there are different genes on the axes.

jjgao:
> Then I think we can just use the same menu for all situations, which includes a None option and all of the genes. Adding the Same gene option wouldn't add much value to the dropdown and also would cause confusion if the users see 2 Same gene options when there are different genes on the axes.

@ngocnn1104 For one gene, I think the "Same gene" option can still be useful so that when switching genes for the axes, the coloring gene is also switched automatically. for two genes, we could add two options "Same gene as H axis" and "Same gene as V axis". But I think that's very tedious.

Alternatively, we could just remove the "Same gene" option and do the following when a different gene is selected in either axis:

- for one gene, switch the new gene for coloring.
- for two genes, if the current gene for coloring is not selected in either axis, switch to the first x-axis gene.

Comparing to the "Same gene" option, the only thing we lose with the alternative approach is that the coloring gene can not stay the same when switching genes in axes, which I think is not a common use case.

Let's try to alternative approach? @ngocnn1104 

cc @tmazor @Sjoerd-van-Hagen and @schultzn

ngocnn1104:
> Alternatively, we could just remove the "Same gene" option and do the following when a different gene is selected in either axis:
> 
> * for one gene, switch the new gene for coloring.
> * for two genes, if the current gene for coloring is not selected in either axis, switch to the first x-axis gene.
> 
> Comparing to the "Same gene" option, the only thing we lose with the alternative approach is that the coloring gene can not stay the same when switching genes in axes, which I think is not a common use case.
> 
> Let's try to alternative approach? @ngocnn1104 

@jjgao Agreed. I will implement the alternative approach and inform you when it's ready.

jjgao:
@ngocnn1104 I think the review app is not working due to some issue that @onursumer is fixing.

onursumer:
@jjgao @ngocnn1104 We now use netlify to deploy review apps, and the deployment is actually working: https://deploy-preview-2842--cbioportalfrontend.netlify.com/. (You need to scroll down all the way to the bottom in the `checks` panel to be able to see the netlify deploy preview link)

jjgao:
Thanks, @onursumer. I guess it's time to remove the integration with heroku? @inodb

jjgao:
@ngocnn1104 when there are different genes for the axes, I think the coloring of dots are different from other plots. Also `CNA` and `both` options are not available. I think we should get rid of the old style and menu for two genes and use the same style and menu as when there is only one gene.

image

jjgao:
@tmazor it would be great if you can test this. I think the other parts except 2-gene plots are ready.

tmazor:
sure thing - will take a look later today @jjgao

tmazor:
Looks good overall - just a few small issues:

1) The alignment of the radio buttons and text is off - the radio buttons should be higher than they are:
image
2) We need to include the gene that is used for coloring samples in the plot somewhere. If I download the plot, that information is lost.
3) If I download the data, there is a column to display the mutations of whichever gene is being used to color the plot. However, this is the case even if I selected "copy number alterations" or "both". The data that is downloaded should reflect the alterations that have been selected to plot.
4) Agree with @jjgao about the coloring when there are 2 different genes selected.

pvannierop:
> We need to include the gene that is used for coloring samples in the plot somewhere. If I download the plot, that information is lost.

For the waterfall plot we have done this by stating the gene name in the plot title.
Screenshot from 2020-01-07 08-45-15

I suggest we do this for all plots where the gene is selected for styling.

ngocnn1104:
> 1. The alignment of the radio buttons and text is off - the radio buttons should be higher than they are
> 2. We need to include the gene that is used for coloring samples in the plot somewhere. If I download the plot, that information is lost.
> 3. If I download the data, there is a column to display the mutations of whichever gene is being used to color the plot. However, this is the case even if I selected "copy number alterations" or "both". The data that is downloaded should reflect the alterations that have been selected to plot.
> 4. Agree with @jjgao about the coloring when there are 2 different genes selected.

I have made the needed changes to fix the above issues. Regarding the gene used for coloring samples, @pvannierop and I chose to include it as the title of the plot's legend because it is a logical place to put that information, keeping things concise and it's simple to implement. @tmazor @jjgao Please let us know what you think.

Screenshot from 2020-01-14 12-36-33

ngocnn1104:
Thanks @jjgao and @tmazor. I have updated failing tests resulted from my changes, please review the codes again if needed and merge this PR if you think it's ready. @adamabeshouse @alisman

alisman:
@dippindots can you take a look at this and determine to what extent it will conflict with generic assay refactor of plots tab?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -2393,46 +2352,72 @@ export function showWaterfallPlot(
...
+        GenericAssayTypeConstants.TREATMENT_RESPONSE,
+        CLIN_ATTR_DATA_TYPE,
+        NONE_SELECTED_OPTION_STRING_VALUE
+    ];
+    const horzDataType = horzSelection.dataType || '';
+    const vertDataType = vertSelection.dataType || '';
+    return (
+        noMolecularProfileDataTypes.includes(horzDataType) &&
+        noMolecularProfileDataTypes.includes(vertDataType));
+}
```
dippindots: I think we should give this function a better name to clarify both axes should be (TREATMENT_RESPONSE or CLIN_ATTR_DATA_TYPE or NONE_SELECTED_OPTION_STRING_VALUE), `bothAxesNoMolecularProfile` sounds like we just need both axes to have any dataType.

dippindots:
@alisman This pr basically changed the `Gene selection` in the plots tab, just has a very minor conflict with the plots tab refactor, which should be easy to fix.

src/pages/resultsView/plots/PlotsTabUtils.tsx
ngocnn1104: @dippindots Do you have any specific suggestion to rename this function?

dippindots: @ngocnn1104 I just noticed we also have other similar names like `oneAxisMolecularProfile ` and `bothAxesMolecularProfile`. I think it's fine to keep this name. I also want to make sure `NoMolecularProfile` should include all `LIMIT_VALUE` data, right? (I need to change apply the logic to all `LIMIT_VALUE` profiles)

ngocnn1104: @dippindots Thanks for clarifying this point for me via Slack. I'm resolving this conversation.

ngocnn1104:
> @alisman This pr basically changed the `Gene selection` in the plots tab, just has a very minor conflict with the plots tab refactor, which should be easy to fix.

@alisman Do you think this PR is ready to be merged?

dippindots:
Thanks for @ngocnn1104 's great work on this feature. @alisman and I had decision to merge the duplication of this pr which passed the tests, so we merged: https://github.com/cBioPortal/cbioportal-frontend/pull/3091 and close this.

PR #2845, opened 2019-11-07
Generate Heatmap menu button's text based on available heatmap profiles

ngocnn1104:
Resolves cBioPortal/cbioportal#6570

## Changes:
- Dynamically generates the text content of Heatmap menu button in Oncoprint tab based on available heatmap profile options:
  + If there is one available heatmap profile, add it to the button text (e.g. `mRNA Heatmap`);
  + If there are more than one available heatmap profile, separate them with `/` (e.g. `mRNA / Protein Heatmaps`);

### Before:
Screenshot from 2019-11-06 16-35-41

### After:
Screenshot from 2019-11-06 16-31-22
Screenshot from 2019-11-06 16-33-35


inodb:
Thanks a lot @ngocnn1104 ! This is so cool!

I noticed a minor styling difference of the borders: https://github.com/cBioPortal/cbioportal-frontend/pull/2845/files?short_path=576ffc5#diff-576ffc593d7d0b21f0040a1a58cf9c5e. One seems slightly darker gray than the other? I don't have a any preference of one over the other, but was wondering if it was on purpose

ngocnn1104:
> I noticed a minor styling difference of the borders: https://github.com/cBioPortal/cbioportal-frontend/pull/2845/files?short_path=576ffc5#diff-576ffc593d7d0b21f0040a1a58cf9c5e. One seems slightly darker gray than the other? I don't have a any preference of one over the other, but was wondering if it was on purpose

@inodb You have good eyes! Actually that's not a styling change. Previously the darker border represented an empty `span` when the Heatmap button did not appear, I think it is unnecessary so I removed it.

alisman:
@inodb @jjgao @Sjoerd-van-Hagen I think the button is getting way too long.  Are we going to add genesets too?  And what if there are more heatmap track types in the future.  Perhaps we can just come up with a better name for the button that is more suggestive of the types within?   Heatmap Tracks?

Sjoerd-van-Hagen:
@alisman It was @schultzn who came up with the idea. Maybe he has some suggestions?

Personally, I agree that it is too long, but I also like the idea of seeing what data is actually available. Perhaps if we have some great shorthands or icons it would work.

schultzn:
I like that idea. We would need icons for:
- RNA
- methylation
- protein
- Tx
what else?

Just some ideas below.

image

image

image

image


image

schultzn:
But would have to be small and distinct from each other

adamabeshouse:
I agree the button is too long.

I don't think icons are useful if they're not already universally recognizable. The subtleties of those icons won't show through at the size we're talking about as well. Good luck reading CH3 :) !

I think for one type, e.g. mrna as shown above, it works well. For more, I don't think it's possible to do it in the button.

Sjoerd-van-Hagen:
How about RNA, Prot, Meth, Tx? With a tooltip that lists them all?

Sjoerd-van-Hagen:
I agree it does not look very tidy but being able to see what data is there would be nice.

A completely different approach could be that the heatmap button shows a menu with:
- mRNA
- Protein
- Methylation
- Treatment
- Gene sets
first, and when you click one of them you get the dialog to select the genes, compounds or gene sets. This way you only have one click to get an overview of the data types, instead of going through the extra step of using the drop down menu in the dialog.

adamabeshouse:
@Sjoerd-van-Hagen what about if the heatmap menu had tabs and to add e.g. a protein track youd navigate to the protein tab. a glance at the tabs shows the data types

Sjoerd-van-Hagen:
Or, instead of having the heat map menu at the top, we put a component at the bottom where we have way more space to put extra information. Since this is also the spot where the heat maps end up when they are added I think it actually makes sense.

Forget about my other suggestions, I think I would actually prefer this approach. cBioPortal/product

Sjoerd-van-Hagen:
> @Sjoerd-van-Hagen what about if the heatmap menu had tabs and to add e.g. a protein track youd navigate to the protein tab. a glance at the tabs shows the data types. is that what you're suggesting?

@adamabeshouse having tabs also has the advantage of showing all the data types in one glance, but I would like to consider doing something at the bottom of the oncoprint now. Kinda like a 'load more...' button you sometimes see on website. See my previous comment. What are your thoughts on that?

adamabeshouse:
@Sjoerd-van-Hagen it's an interesting idea, I am interested in understanding more what your vision is for this component

tmazor:
I'm also curious to see more, but my initial reaction is that it will be confusing to have controls split between the top and bottom of the OncoPrint. I like the idea of keeping the menu where it is but using tabs to clearly display all available datatypes.

alisman:
@cBioPortal/product what if collapsed clinical data and heatmaps into one button "Add data".  We could put icons in for data types into that top level button to catch people's eye.  Then when they click that, we give them tabs or even a dialog.  

Alternately, if the goal is really just to bring attention to to the data types available, just add a message above the oncoprint.
image

schultzn:
I vote against another row of text. 

How about “Add heatmap” or “Add heatmaps” if data are available, and a small circle with a number inside of it that shows how many different platforms those are available for?

adamabeshouse:
@schultzn I like the idea of changing it to "Add Heatmap", especially now that in the new pull request we're removing the "Remove" and "Cluster" buttons from that menu, so it really is only adding.

I like the idea of the circle with the number. On mouseover it could give a list of data types in a tooltip as well?

jjgao:
I am late on this. 

I agree with @tmazor that splitting the menu may be confusing and using tabs to display the datatypes might be a good idea. Of course, this will be different from Plots tab (dropdown to select data types). Maybe we can come up with a consistent interface?

I am not sure if I like the idea of circled number - sorry, @schultzn, I have a compulsive behavior to click on any circled number :(

Sjoerd-van-Hagen:
My proposal is something like this for adding heatmaps.
image
This way:
- you can see the data types that are available, and there is plenty of space for 20 data types or more
- heat maps are added where the buttons are, so it is intuitive and easy to find, and consistent with many other web UIs where you need to add things at the bottom
- we have more space in the top bar for the other buttons which are more focused on the oncoprint

With the new heatmap headers and their small menu that Adam implemented, we can add an 'add tracks' option there. This way for each heat map that is displayed you have the option to add tracks close at hand. You could actually remove the data type from the bottom list after adding a heat map because it would not be necessary any more. It would reappear when the heatmap is deleted.

I think this adequately addresses all concerns that were raised earlier.

To Tali's concern that it splits the menu, I think it actually makes things more clear. I think the menu is a bit overloaded with too many different things. I would actually suggest we consider taking out the clinical tracks as well. We could put a similar component at the top to add clinical tracks, where the buttons would correspond to the most popular clinical tracks, with one button that says 'other...' or something. This way all the 'adding' workflows are taken out of the menu, if consistency in that respect is a concern.

Perhaps we should also give some thought on what to do the the extra space in the top menu because I think we can put that to use to make it a lot more user friendly.

For instance, we now have a button 'mutations', which to me is really unclear in terms of what I can expect there. I would prefer to have separate menus 'Color by' and 'Filter'. Perhaps we can also take the events per sample/patient selector out of the view menu, and have a 'Show' menu for the other options in 'View'. I am sure you would agree that that would be a big improvement in usability.

@cBioPortal/product

adamabeshouse:
whats the plan for this?

jjgao:
I think separating "Add Clinical Tracks" and "Add Heatmaps" from the rest of the menu is a good idea. But moving "Add heatmaps" to the bottom of the OncoPrint may not be the best option, for example, the legend can be really long and therefor push "Add heatmaps" down. Adding it between the tracks and legend may not be ideal either since that would break the oncoprint and also not be good for screenshots.

Maybe a simple option is to keep "Add Clinical Tracks" and "Add Heatmaps" to the left and move the rest to the right? (I think @tmazor proposed that.) For "Add Heatmaps", we can have subtabs for mRNA, protein, methylation, and drug responses.

tmazor:
I second @jjgao's proposal - it makes the most sense to me and seems like a relatively simple set of changes to clean up the interface.

schultzn:
+1

Sjoerd-van-Hagen:
Except that it does not solve the issue!

The legend sometimes being too long is not a good reason imo. If you have 20 items in your legend because you added a clinical track for cancer type detailed on a pan cancer study, that is not helpful at all. In those cases users should make a smaller selection of samples first. If they do not, they will have to deal with the menu button being all the way down, but I think that is fair in this case.

Perhaps we can do some things to make the legend smaller, but that is a separate issue.

schultzn:
I really don’t think we should add buttons that are potentially “below the fold”. Even without the legend, oncoprints with a large number of genes could fill the screen even without the legend.

Any other ideas?

Sjoerd-van-Hagen:
When I add 15+ genes I get a scroll bar, so there is not really a problem there I think?

schultzn:
On a 13" laptop screen, the fold is right in the legend. So the button would have to be above the legend? See below.

https://www.cbioportal.org/results/oncoprint?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=msk_impact_2017&case_set_id=msk_impact_2017_cnaseq&data_priority=0&gene_list=EGFR%2520ERBB2%2520MET%2520PDGFRA%2520RET%2520ROS1%2520ALK%2520KRAS%2520NRAS%2520HRAS%2520NF1%2520ARAF%2520BRAF%2520RAF1%2520MAP2K1&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=msk_impact_2017_cna&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=msk_impact_2017_mutations&tab_index=tab_visualize&clinicallist=NUM_SAMPLES_PER_PATIENT

image

Sjoerd-van-Hagen:
In terms of width, the menu at the top is not that wide. You could add a button per data type there. If I remove the heatmap menu, the space left is about as wide as four 'tab titles', so four buttons should easily fit, without requiring a wider screen. There are measures that can be taken for smaller screens, such as a hamburger button when it overflows. Actually, the buttons will already wrap around when the screen is not wide enough.

How many people use 13" screens when visiting cBioPortal? We could track screen sizes.

schultzn:
Yes, I think we are ok with screen width, and multiple buttons would work. 

I use cBioPortal on my laptop all the time - don't even have a screen that I plug into at home. There are probably many others.

alisman:
Here are screen resolution stats from GA:
image

alisman:
@schultzn  mentioned the idea of having a "hint" mode (default on) in which we provide more friendly descriptive instructions to users.   I think this is a good idea.   When the mode is off, we have an icon somewhere in the header that turns it back on and also shows some count of the hints that are available on the page.  

image

schultzn:
The hints could be useful, also for other features of the site. Let's explore that idea some more.

But maybe something as simple as the below could work well?

image

Sjoerd-van-Hagen:
@schultzn I think most users, when they see a number there, would automatically assume it would be the common data types like expression, protein, methylation, and still not discover if there would be something else like treatments.

alisman:
I propose we start with Niki's idea about the bubbles.  Also put in some GA tracking on this behavior.  Then follow up with the hints (which can themselves be dynamic and call out the particular types available).   Sound good?

schultzn:
> @schultzn I think most users, when they see a number there, would automatically assume it would be the common data types like expression, protein, methylation, and still not discover if there would be something else like treatments.

@Sjoerd-van-Hagen So should we have one button for "Add Genomic Heatmaps" and one "Add Tx Heatmaps"?

alisman:
@Sjoerd-van-Hagen I'm going to close this one for the time being.  We will be deploying the info bubbles in the buttons today, which will at least partly address this issue.  Lets bring it up again in steering call if we want to take things further (hinting).

Sjoerd-van-Hagen:
@alisman Sure, np.

PR #2859, opened 2019-11-11
Treatment with generic assay api

dippindots:
The fourth step of https://github.com/cBioPortal/cbioportal/issues/6769. Change the treatment_response implementation to use `generic_assay` apis.

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -25,7 +25,7 @@ import {
     SampleFilter,
     SampleIdentifier,
     SampleList,
-    SampleMolecularIdentifier,
+    SampleMolecularIdentifier
 } from 'shared/api/generated/CBioPortalAPI';
 import client from 'shared/api/cbioportalClientInstance';
 import {action, computed, observable, ObservableMap} from 'mobx';
@@ -94,8 +94,6 @@ import {
     GenesetDataFilterCriteria,
     GenesetMolecularData,
     MolecularProfileCasesGroupFilter,
-    Treatment,
-    TreatmentFilter,
 } from '../../shared/api/generated/CBioPortalAPIInternal';
 import internalClient from '../../shared/api/cbioportalInternalClientInstance';
 import {CancerGene, IndicatorQueryResp,} from '../../public-lib/api/generated/OncoKbAPI';
@@ -183,6 +181,7 @@ import {
 import {FRACTION_GENOME_ALTERED, MUTATION_COUNT,} from 'pages/studyView/StudyViewPageStore';
 import {IVirtualStudyProps} from 'pages/studyView/virtualStudy/VirtualStudy';
 import {decideMolecularProfileSortingOrder} from './download/DownloadUtils';
+import { fetchTreatmentByMolecularProfileIds, Treatment } from 'shared/lib/GenericAssayUtils';
 
 type Optional<T> =
     | { isApplicable: true; value: T }
@@ -197,6 +196,10 @@ export const AlterationTypeConstants = {
     GENESET_SCORE: 'GENESET_SCORE',
     METHYLATION: 'METHYLATION',
     TREATMENT_RESPONSE: 'TREATMENT',
+};
+
+export enum GenericAssayTypeConstants {
+    'TREATMENT_RESPONSE' = 'TREATMENT_RESPONSE'
 };
 
 export const AlterationTypeDisplayConstants = {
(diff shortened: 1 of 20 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -221,6 +223,10 @@ export const AlterationTypeConstants = {
     GENERIC_ASSAY: 'GENERIC_ASSAY',
 };
 
+export const GenericAssayTypeConstants = {
```
alisman: i think this should be an Enum

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3694,13 +3700,28 @@ export class ResultsViewPageStore {
...
-            return internalClient.fetchTreatmentsUsingPOST({
-                treatmentFilter: {
-                    studyIds: this.studyIds.result!,
-                } as TreatmentFilter,
-            });
+            const treatmentResponseProfiles = this.molecularProfilesInStudies.result.filter((profile) => {
+                return profile.genericAssayType === GenericAssayTypeConstants.TREATMENT_RESPONSE;
+            })
+            const treatmentResponseProfileIds = treatmentResponseProfiles.map(profile => profile.molecularProfileId);
+            const treatmentsMeta: GenericAssayMeta[] = await client.fetchGenericAssayMetaDataUsingPOST(
```
alisman: ts should be infer type from client method return type.  if it's not, something is wrong.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3694,13 +3700,28 @@ export class ResultsViewPageStore {
...
+                {
+                    genericAssayMetaFilter: {
+                        molecularProfileIds: treatmentResponseProfileIds
+                    } as GenericAssayMetaFilter
+                } as any
```
alisman: how come this needs to be cast as any?  are we leaving stuff out?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3694,13 +3700,28 @@ export class ResultsViewPageStore {
...
+            )
+            const treatments = _.map(treatmentsMeta, (meta) => {
+                return {
+                    treatmentId: meta.stableId,
+                    name: "NAME" in meta.genericEntityMetaProperties ? meta.genericEntityMetaProperties["NAME"] : "NA",
+                    description: "DESCRIPTION" in meta.genericEntityMetaProperties ? meta.genericEntityMetaProperties["DESCRIPTION"] : "NA",
+                    refLink: "URL" in meta.genericEntityMetaProperties ? meta.genericEntityMetaProperties["URL"] : "NA",
+                } as Treatment
```
alisman: again how come it needs to be cast?

src/shared/cache/TreatmentMolecularDataCache.ts
```diff
@@ -83,20 +85,35 @@ async function fetch(
...
-            // use 'as' to tell TypeScript that this object really does fit.
-            // tslint:disable-next-line: no-object-literal-type-assertion
-            treatmentDataFilterCriteria: {
-                treatmentIds: treatmentIdsByProfile[profileId],
+            molecularProfileId: profileId,
+            genericAssayDataFilter: {
+                genericAssayStableIds: treatmentIdsByProfile[profileId],
                 ...sampleFilterByProfile[profileId]
-            } as TreatmentDataFilterCriteria
+                // the Swagger-generated type expected by the client method below
```
alisman: nice comment!   maybe we should try to correct the type in code.

alisman:
There's a lot of casting in this PR.  I now realize it's probably because we are going from generic endpoint to a particular type.   I believe the use the "as" keyword is weak.  If instead you set type of variable and then assign to it, checking will be stronger.  However, this may be harder.  Lets discuss.

src/shared/lib/GenericAssayUtils.ts
```diff
@@ -0,0 +1,106 @@
...
+    'patientId': string;
+    'sampleId': string;
+    'stableId': string;
+    'studyId': string;
+    'treatmentId': string;
+    'uniquePatientKey': string;
+    'uniqueSampleKey': string;
+    'value': string;
+}
+
```
dippindots: @alisman Since we removed the `treatment` implementation from the backend, the frontend should response to having types `Treatment` and `TreatmentMolecularData`.

dippindots:
Close since we need to make a draft pr for testing.

PR #2864, opened 2019-11-11
Always proxy oncokb request

zhx828:
The coming up new OncoKB website is no longer allowing cross domain requests. All requests need to be proxied and attached the auth header

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -133,17 +133,20 @@ export function getOncoQueryDocUrl() {
     return buildCBioPortalPageUrl("/oql");
 }
 
+export function trimProtocol(url:string) {
+    // we need to support legacy configuration values
+    url = url.replace(/^http[s]?:\/\//,''); // get rid of protocol
+    url = url.replace(/\/$/,""); // get rid of trailing slashes
+    url = url.replace(/^\/+/,""); // get rid of leading slashes
+    return url;
+}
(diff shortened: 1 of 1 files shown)


src/shared/api/urls.ts
```diff
@@ -153,8 +156,13 @@ export function getProxyUrlIfNecessary(url:any) {
 
 export function getOncoKbApiUrl() {
     let url = AppConfig.serverConfig.oncokb_public_api_url;
-    return getProxyUrlIfNecessary(url);
+    if (typeof url === 'string') {
+        return buildCBioPortalAPIUrl(`proxy/${trimProtocol(url)}`);
+    } else {
+        return undefined;
```
alisman: does calling code handle undefined?  I guess typescript would complain if it didn't, right?

zhx828: As far as I can tell, we inject these url to the client directly bypassing the constructer. It will have problem if url is undefined. But it seems like we will never run into this scenario since we have the serverConfigDefaults defined. @alisman

zhx828:
@inodb I agree. This is a fix to make sure any instances that do not wish to upgrade to 3.2(which oncokb auth will be included) can run smoothly. This will eliminate a cross-domain issue if the oncokb instance does not allow that.
In the 3.2, I will remove oncokb_public_url from the frontend.

zhx828:
I took a look places where oncokb is used. They all work properly.

PR #2866, opened 2019-11-12
Automated master-rc merge (mMRc8)

pvannierop:
This pull request has been automatically created for the merge of master to rc branch. Please resolve any merge errors and errors in the CI tests and merge.


pvannierop:
@alisman @adamabeshouse @inodb
E2e-tests are failing because of the changes to the backend (at least related to the generic_asssay endpoint feature). @dippindots Are you coordinating with the frontend team?

inodb:
Thanks for letting us know @pvannierop - @dippindots is looking into it. I've extended the issue a bit for running e2e tests on the backend: https://github.com/cBioPortal/cbioportal/issues/3850. Feel free to add more if you have any suggestions

dippindots:
@pvannierop @inodb. I saw this message when I tried to query a study on `rc.cbioportal.org`. 
image
This `genericAssayType` is a new column introduced in the `generic_assay` implementation. I guess the database needs to run the latest `migration.sql` to update the database schema.

inodb:
@dippindots I've updated the rc database. Looks like a lot less errors now. The e2e tests only have a single failure related to mutations ordering on patient view so that's a false positive. The localdb tests fails for various treatment data tests. Any clue? Do we need to apply your new migration for treatment data? Thanks!

dippindots:
@inodb I see, I think the reason why `end_to_end_tests_localdb` fail is that the `localdb_test` import the data by using the importer in `rc` branch (which will set `treatment` data as `TREATMENT` type), but the data existing in the `remote rc database` is the data imported by previous importer (which will set `treatment` data as `GENERIC_ASSAY` type).
Maybe we can talk about this in person.

PR #2869, opened 2019-11-12
Create feedback form for trials

victoria34:
Fix https://github.com/oncokb/oncokb/issues/1816

image

image

### DIFF
--- a/src/pages/patientView/trialMatch/TrialMatchTable.tsx
+++ b/src/pages/patientView/trialMatch/TrialMatchTable.tsx
@@ -6,11 +6,14 @@ import {
     IClinicalGroupMatch, IGenomicGroupMatch, IGenomicMatch, IDetailedTrialMatch, IArmMatch, IGenomicMatchType
 } from "../../../shared/model/MatchMiner";
 import styles from './style/trialMatch.module.scss';
-import { computed } from "mobx";
+import { computed, observable } from "mobx";
 import LazyMobXTable from "../../../shared/components/lazyMobXTable/LazyMobXTable";
 import SampleManager from "../SampleManager";
 import DefaultTooltip, { placeArrowBottomLeft } from "../../../public-lib/components/defaultTooltip/DefaultTooltip";
 import { getAgeRangeDisplay } from "./TrialMatchTableUtils";
+import TrialMatchFeedback from "./TrialMatchFeedback";
+import AppConfig from 'appConfig';
+import { Button } from "react-bootstrap";
 
 export type ITrialMatchProps = {
     sampleManager: SampleManager | null;
@@ -43,6 +46,9 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
             [ColumnKey.MATCHING_CRITERIA]: 0.65 * (this.props.containerWidth - ColumnWidth.STATUS)
         };
     }
+
+    @observable showTrialFeedback = false;
+    @observable showGeneralFeedback = false;
 
     private _columns = [{
         name: ColumnKey.TITLE,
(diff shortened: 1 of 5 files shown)


src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -118,8 +125,22 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
...
         name: ColumnKey.STATUS,
         render: (trial: IDetailedTrialMatch) => (
-            <div className={styles.statusButton}>
-                <a target="_blank" href={"https://www.mskcc.org/cancer-care/clinical-trials/" + trial.protocolNo}><span className={styles.statusBackground}>{trial.status}</span></a>
+            <div className={styles.statusContainer}>
+                <a target="_blank" href={"https://www.mskcc.org/cancer-care/clinical-trials/" + trial.protocolNo}>
+                    <span className={styles.statusBackground}>{trial.status}</span>
+                </a>
+                <span className={styles.feedback}>
+                    <button className={"btn btn-default btn-sm btn-xs " + styles.feedbackButton} onClick={() => this.showTrialFeedback=true}>Feedback</button>
```
zhx828: You could use a Button component from react-bootstrap

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -118,8 +125,22 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
     }, {
         name: ColumnKey.STATUS,
         render: (trial: IDetailedTrialMatch) => (
-            <div className={styles.statusButton}>
-                <a target="_blank" href={"https://www.mskcc.org/cancer-care/clinical-trials/" + trial.protocolNo}><span className={styles.statusBackground}>{trial.status}</span></a>
+            <div className={styles.statusContainer}>
+                <a target="_blank" href={"https://www.mskcc.org/cancer-care/clinical-trials/" + trial.protocolNo}>
```
zhx828: Why not just use Button from react-bootstrap, and `variant="link"`

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -306,6 +327,13 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
     render() {
         return (
             <div>
+                <p style={{marginBottom: '0'}}>Curated genomic and clinical criteria from open clinical trials at Memorial Sloan Kettering. Please <a href="mailto:team@oncokb.org">contact us</a> or submit <a onClick={() => this.showGeneralFeedback=true}>feedback form</a> if you have any questions.</p>
```
zhx828: I'm surprised that changing `showGeneralFeedback` on the click event would actually work in mobx observable. @alisman what's the situation that a `@autobind` has to be attached?

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Changed to Button component

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Some css styles cannot apply for the Button. Besides, this is a link not a button. I don't want to mix them together.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
alisman: this will work.  if you pass it a function directly (not an anonymous callback) then it has to be bound with autobind.  Using an anonymous callback is discouraged b/c it causes virtual dom to be differed each time. But in this case, it's totally negligible and not worth changing.   So, rule of thumb is, if something is going to be rendered a lot and occurs many times in DOM, then use a bound handler.  If not, do the above.

PR #2874, opened 2019-11-13
SV: Fix rendering bar plot that are not loaded initially

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6797

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -1870,7 +1870,6 @@ export class StudyViewPageStore {
                         this.initialVisibleAttributesClinicalDataBinCountData);
                 },
                 invoke: async () => {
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1885,7 +1884,7 @@ export class StudyViewPageStore {
                         !attributeChanged) {
                         result = this.initialVisibleAttributesClinicalDataBinCountData.result;
                     } else {
-                        if (this._clinicalDataIntervalFilterSet.has(uniqueKey)) {
+                        if (this._clinicalDataIntervalFilterSet.has(uniqueKey) || this.isInitialFilterState) {
```
zhx828: The await might need to be updated as well.

kalletlak: Looks like its already present

PR #2878, opened 2019-11-14
SV: select more than one category on custom charts

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6793


zhx828:
Thanks for the fix!

PR #2882, opened 2019-11-19
Track every route change in loggly

alisman:
We want to track every route change in loggly in order to better correlate performance issues with usage patters


alisman:
@Luke-Sikina correct.  for logging purposes, i care about e2e because it's causing load on server.  i actually have suspicion it causes problems because of very rapid fire use of certain api endpoints.  try/catch is just to be safe, make sure that non critical code can't cause problems in app

PR #2883, opened 2019-11-19
Avoid creating multiple forms

victoria34:
Avoid creating multiple popup modals

### DIFF
--- a/src/pages/patientView/trialMatch/TrialMatchTable.tsx
+++ b/src/pages/patientView/trialMatch/TrialMatchTable.tsx
@@ -6,7 +6,7 @@ import {
     IClinicalGroupMatch, IGenomicGroupMatch, IGenomicMatch, IDetailedTrialMatch, IArmMatch, IGenomicMatchType
 } from "../../../shared/model/MatchMiner";
 import styles from './style/trialMatch.module.scss';
-import { computed, observable } from "mobx";
+import { action, computed, observable } from "mobx";
 import LazyMobXTable from "../../../shared/components/lazyMobXTable/LazyMobXTable";
 import SampleManager from "../SampleManager";
 import DefaultTooltip, { placeArrowBottomLeft } from "../../../public-lib/components/defaultTooltip/DefaultTooltip";
@@ -19,6 +19,11 @@ export type ITrialMatchProps = {
     sampleManager: SampleManager | null;
     detailedTrialMatches: IDetailedTrialMatch[];
     containerWidth: number;
+}
+
+export type ISelectedTrialFeedbackFormData = {
+    nctId: string;
+    protocolNo: string;
 }
 
 enum ColumnKey {
@@ -47,7 +52,7 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
         };
     }
 
-    @observable showTrialFeedback = false;
+    @observable selectedTrialFeedbackFormData: ISelectedTrialFeedbackFormData | undefined;
     @observable showGeneralFeedback = false;
 
     private _columns = [{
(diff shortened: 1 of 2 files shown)


src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -48,6 +53,10 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
     }
 
     @observable showTrialFeedback = false;
+    @observable selectedTrial: ISelectedTrial = {
```
alisman: by default this should be undefined

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -196,6 +196,12 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
         );
     }
 
+    public openFeedbackForm(nctId: string, protocolNo: string) {
+        this.showTrialFeedback = true;
```
alisman: instead of having a flag, you can just can just have selectedTrialFeedbackFormData which is either undefined (hide) or shown (populated).  In general, it's best not to update multiple properties like this.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -196,6 +196,12 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
         );
     }
 
+    public openFeedbackForm(nctId: string, protocolNo: string) {
```
alisman: this modifies mobx state and so should have @action decorator

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Removed the flag

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Updated

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Added @action

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -335,6 +339,17 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
                     title="OncoKB Matched Trials General Feedback"
                     userEmailAddress={AppConfig.serverConfig.user_email_address}
                 />
+                {   this.selectedTrialFeedbackFormData &&
+                    <TrialMatchFeedback
+                        show={!!this.selectedTrialFeedbackFormData}
```
alisman: don't need show prop and internal logic anymore.

src/pages/patientView/trialMatch/TrialMatchTable.tsx
```diff
@@ -335,6 +339,17 @@ export default class TrialMatchTable extends React.Component<ITrialMatchProps> {
...
+                        onHide={() => !!(this.selectedTrialFeedbackFormData=undefined)}
```
alisman: onHide also modifies state so should be action.  What I would do is change openFeedbackForm to openCloseFeedbackForm and have it accept undefined if it's close

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: I added the logic here because the window always popup the Exit Confirmation modal.
Screen Shot 2019-11-19 at 6 04 58 PM

src/pages/patientView/trialMatch/TrialMatchTable.tsx
victoria34: Updated

PR #2886, opened 2019-11-19
Using components from react-mutation-mapper for ClinVar and Gnomad columns

onursumer:
Fix cBioPortal/cbioportal/issues/6785

Related PRs: 

- cBioPortal/react-mutation-mapper/pull/80
- cBioPortal/react-mutation-mapper/pull/68
- cBioPortal/react-mutation-mapper/pull/118

# Checks


onursumer:
@leexgh deployment is manual for frontend PRs. I have just initiated one, should be ready in a few mins.

leexgh:
@onursumer thanks! All looks great 🙂 One minor thing is that we have a tooltip showing `Variant has no data in gnomAD.` when there is no gnomad data. Looks like mutation mapper doesn't have it?

onursumer:
@leexgh good catch! Actually, the logic was copied over to mutation mapper, but seems like I have introduced a bug when doing that. I will open a PR for that in `react-mutation-mapper` and then update this PR.

onursumer:
@leexgh I've updated the PR. Could you take a final look and confirm that everything looks good?

PR #2888, opened 2019-11-20
Provide a oncokb model for types are not available through swagger ge…

zhx828:
…nerator

The swagger generator would not parse the enum as a typescript enum.
This file needs to be updated once the relavent generated clients is modified


zhx828:
The failing test is irrelevant.

PR #2889, opened 2019-11-21
Oncokb auth

zhx828:


src/shared/api/urls.ts
```diff
@@ -153,12 +153,19 @@ export function getProxyUrlIfNecessary(url:any) {
...
-    } else {
-        return undefined;
+export function getOncoKbProxyApiUrl() {
+    if (AppConfig.serverConfig.oncoKbTokenEnabled) {
+        return buildCBioPortalAPIUrl(`proxy/oncokb`);
+    }
+    // TODO workaround for temporary backward compatibility
+    //  remove this after documenting OncoKB token for all portal instances
+    else {
+        let url = (AppConfig.serverConfig as any).oncokb_public_api_url;
```
onursumer: need to cast as any since `oncokb_public_api_url` is no longer part of the `serverConfig` definition

PR #2891, opened 2019-11-21
Routing Improvements

Luke-Sikina:
Fix cBioPortal/cbioportal#6768

- invalid subpaths (tabs) result in a 404 page
- 404 page is more consistent with regular error page

### DIFF
--- a/src/pages/groupComparison/GroupComparisonTabs.ts
+++ b/src/pages/groupComparison/GroupComparisonTabs.ts
@@ -1,0 +1,9 @@
+export enum GroupComparisonTab {
+    OVERLAP = "overlap",
+    MUTATIONS = "mutations",
+    CNA = "cna",
+    MRNA = "mrna",
+    PROTEIN = "protein",
+    SURVIVAL = "survival",
+    CLINICAL = "clinical"
+}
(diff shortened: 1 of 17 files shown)


src/pages/groupComparison/GroupComparisonTabs.ts
```diff
@@ -0,0 +1,9 @@
+export enum GroupComparisonTab {
```
Luke-Sikina: I could use some help here. I had to pull these enums into their own tabs in order to get the imports to work correctly in `routes.jsx`. When I just imported them in their respective files, the app compiled, but failed at runtime with the following message
```
utilities.js:38 Uncaught TypeError: Cannot read property 'appendChild' of null
    at Object../node_modules/units-css/lib/utilities.js.utilities.getCreatedElementDimensions (utilities.js:38)
    at Object../node_modules/units-css/lib/utilities.js.utilities.getCreatedElementWidth (utilities.js:51)
    at utilities.js:114
    at Object../node_modules/units-css/lib/utilities.js (utilities.js:117)
    at __webpack_require__ (bootstrap:766)
    at fn (bootstrap:129)
    at Object../node_modules/units-css/lib/conversions/length.js (length.js:6)
    at __webpack_require__ (bootstrap:766)
    at fn (bootstrap:129)
    at Object../node_modules/units-css/lib/conversions/index.js (index.js:8)
```

alisman:
@Luke-Sikina, so, looking at this i'm thinking that we need to take a different route (no pun intended).  It's my fault.  The tabId path segment /page/tabId should really be handled entirely by the tab component.   If it gets fed a tabid that doesn't corrrespond to anything, it should have an error (no correpoding tab).  This should be built into MSKTabs component.    Lets discuss further

PR #2892, opened 2019-11-22
Monorepo with yarn workspaces and lerna

onursumer:
- Enabled `yarn workspaces` for package management -- needed to update the yarn to the latest version for this.
- Using `lerna` for multi package tasks (test, build, publish, etc.)
- Moved `src/public-lib` contents under `packages/cbioportal-frontend-commons` with a proper `package.json` and updated references accordingly -- this increases the number of files changed dramatically, but majority of these changes are just reference updates.
- Using shared `rollup` and `ts` config options to bundle modules under packages.
- Added github workflow to auto publish public packages to npm.

# Checks


inodb:
@onursumer Nice work! Just wanted to share I found something describing lerna and yarn workspaces. Maybe of interest: https://doppelmutzi.github.io/monorepo-lerna-yarn-workspaces/. Lerna seems more powerful for publishing individual packages but yarn workspaces allows skipping the bootstrap step. You can use both as well

When we introduce lerna, we prolly need some docs on how to install packages. Like how to install a dependency for one of the projects vs a dependency for all of them.  As well as how to run the individual projects

lerna.json
```diff
@@ -0,0 +1,9 @@
+{
+  "npmClient": "yarn",
```
onursumer: use yarn instead of lerna

lerna.json
```diff
@@ -0,0 +1,9 @@
...
+  "packages": [
```
onursumer: we need to add root to make `cbioportal-frontend` part of the monorepo, all other modules should be added under packages from now on.

lerna.json
```diff
@@ -0,0 +1,9 @@
...
+    ".",
+    "packages/*"
+  ],
+  "version": "independent",
```
onursumer: this is required for each package to have a different package version

lerna.json
```diff
@@ -0,0 +1,9 @@
...
+  "useWorkspaces": true
```
onursumer: enable `yarn workspaces`

package.json
```diff
@@ -1,29 +1,37 @@
 {
   "name": "cbioportal-frontend",
+  "private": true,
+  "version": "0.0.0",
+  "workspaces": {
+    "packages": [".", "packages/*"]
```
onursumer: this is identical to the `packages` property of `lerna.json`

package.json
```diff
@@ -1,29 +1,37 @@
...
+  },
   "description": "",
   "scripts": {
     "start": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server",
     "startSSL": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server --https",
+    "startAll": "lerna run start --parallel",
```
onursumer: runs `yarn start` for all workspaces in parallel. great feature for development.

package.json
```diff
@@ -1,29 +1,37 @@
...
     "clean": "rimraf dist tsDist common-dist",
     "build": "yarn run clean && yarn run compileOqlParser && yarn run buildDLL:prod && cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=2048 webpack",
+    "buildAll": "lerna run build --stream",
+    "buildModules": "lerna run build --ignore=cbioportal-frontend --stream",
```
onursumer: build all workspaces except the main app

package.json
```diff
@@ -1,29 +1,37 @@
...
+    "publishModules": "lerna run publish from-package --yes",
```
onursumer: publish packages that are not already published to npm, by checking the `version` of each `package.json` (see https://github.com/lerna/lerna/tree/master/commands/publish#bump-from-package)

package.json
```diff
@@ -32,12 +40,13 @@
     "test": "export NODE_OPTIONS=--max-old-space-size=2048; karma start karma.conf.js",
     "test:watch": "yarn run test -- --watch",
     "test:debug": "yarn run test -- --debug --browsers=Chrome_with_debugging --single-run=false",
+    "testAll": "lerna run test --stream",
```
onursumer: run `yarn test` for all workspaces

package.json
```diff
@@ -32,12 +40,13 @@
...
     "prettierLastCommit": "yarn run prettier --write $(git diff-tree --no-commit-id --name-only -r $(git rev-parse HEAD))",
     "syncmock": "node src/test/fetchMockData.js --diff"
   },
   "engines": {
     "node": "8.12.0",
-    "yarn": "1.13.0"
+    "yarn": "1.21.1"
```
onursumer: we need to update the yarn version for workspaces to work properly

packages/config/rollup.config.js
```diff
@@ -0,0 +1,58 @@
+import typescript from 'rollup-plugin-typescript2';
```
onursumer: Shared rollup config file. Currently there is only one module (`cbioportal-frontend-commons`) using this config. But the plan is to move `react-mutation-mapper` under this monorepo as well.

src/shared/components/query/MolecularProfileSelector.tsx
```diff
@@ -1,174 +1,174 @@
-import * as _ from 'lodash';
-import * as React from 'react';
-import {MolecularProfile} from "../../api/generated/CBioPortalAPI";
-import FontAwesome from "react-fontawesome";
-import * as styles_any from './styles/styles.module.scss';
-import {observer} from "mobx-react";
-import classNames from 'classnames';
-import { FlexRow } from "../flexbox/FlexBox";
-import {QueryStoreComponent} from "./QueryStore";
-import DefaultTooltip from "../../../public-lib/components/defaultTooltip/DefaultTooltip";
```
onursumer: This is actually the only change in this file. The whole file seems to be updated because of a `new line` difference between Windows and other systems.

package.json
```diff
@@ -1,29 +1,38 @@
...
+  },
   "description": "",
   "scripts": {
     "start": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server",
     "startSSL": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server --https",
+    "startAll": "lerna run start --parallel",
     "clean": "rimraf dist tsDist common-dist",
     "build": "yarn run clean && yarn run compileOqlParser && yarn run buildDLL:prod && cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=2048 webpack",
+    "buildAll": "lerna run build --stream",
+    "buildNetlify": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && yarn run buildAll",
```
inodb: @onursumer Maybe `buildAll` can be the same as `buildNetlify`, seems like you would always want to use the env variables right?

inodb:
@onursumer just thought of something: we should also check whether jitpack will still work. When you include the frontend repo in the backend repo it gets build on jitpack using this maven pom file: https://github.com/cBioPortal/cbioportal-frontend/blob/master/pom.xml#L28-L31. You might need to add lerna there as well. You can test it locally by running `mvn clean install` in the root of the frontend repo. To test on jitpack you can try to compile the backend repo pointing this pom to your fork and commit: https://github.com/cBioPortal/cbioportal/blob/master/pom.xml#L273-L274

package.json
```diff
@@ -1,43 +1,55 @@
...
-    "start": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server",
-    "startSSL": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server --https",
+    "start": "lerna run watch --parallel",
+    "startSSL": "lerna run watchSSL --parallel",
+    "watch": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server",
+    "watchSSL": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && cross-env NODE_ENV=development webpack-dev-server --https",
     "clean": "rimraf dist tsDist common-dist",
-    "build": "yarn run clean && yarn run compileOqlParser && yarn run buildDLL:prod && cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=2048 webpack",
+    "build": "./scripts/env_vars.sh && eval \"$(./scripts/env_vars.sh)\" && yarn run buildAll",
+    "buildAll": "yarn run buildModules && yarn run buildMain",
```
onursumer: currently, jitpack is the only use case where we don't need to set env vars, so we call `buildAll` instead of `build` there.

PR #2893, opened 2019-11-22
Fix long runtimes for e2e-localdb tests

pvannierop:
Fix cBioPortal/cbioportal#6844

This PR corrects long running times for e2e-localdb tests. Cause was output of docker pull command being part of the database contents hash.


pvannierop:
Not at this place in the code. This code will only create the fingerprint. At other points the docker build command is issued again. Here the errors will be visible.

PR #2896, opened 2019-11-23
Allow setting BACKEND on CircleCI

inodb:
On the backend repo the e2e localdb test were giving an error because a
BACKEND variable was being set in custom.sh. This changes the test to
check if the commit itself contains BACKEND such that changes to
custom.sh not checked in are ignored


inodb:
Also fixed the screenshot, there is some issue with updating them: https://github.com/cBioPortal/cbioportal/issues/6855

inodb:
Gonna merge this since it's a pretty trivial fix and would like to have ✅master branches on frontend and backend. Feel free to add comments here or open issues if u find any problems and ill try to address them

PR #2900, opened 2019-11-25
results view oncoprint - fix wrapping behavior in tooltip

adamabeshouse:
this PR just adds "display:inline-block" to spans that are "white-space:nowrap" to make them wrap as blocks

image


zhx828:
@adamabeshouse should it wrap by word?

adamabeshouse:
@zhx828 this is only to address the issue with overflowing gene panel ids as seen in the linked issue:

image

PR #2902, opened 2019-11-25
Refactor Comparison group manager, and fix checkbox alignment

adamabeshouse:
before:
image

after:
image

### DIFF
--- a/src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
+++ b/src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
@@ -98,6 +98,35 @@ export default class GroupCheckbox extends React.Component<IGroupCheckboxProps, 
         );
     }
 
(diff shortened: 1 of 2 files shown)


src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
```diff
@@ -98,6 +98,35 @@ export default class GroupCheckbox extends React.Component<IGroupCheckboxProps,
...
+            >
+                <input
+                    type="text"
+                    value={this.nameInput}
+                    onChange={this.handleNameInputChange}
+                    style={{width:174}}
+                />
+            </DefaultTooltip>
+        );
+    }
```
zhx828: I don't see any content changes for this block. Let me know if it's not the case.

adamabeshouse: thats right

PR #2908, opened 2019-11-29
Update clinical timeline version to v0.0.19

inodb:
New version allows abbreviations in track name if abbreviation shorter than 5 characters and keeps camelcasing if the track name already has that. See also https://github.com/cBioPortal/clinical-timeline/pull/114


Sjoerd-van-Hagen:
Thanks @inodb.

PR #2909, opened 2019-11-30
add separate genome nexus cache for mutation assessor

leexgh:
have a separate cache for mutation assessor, each column in functional impact loading separately, so can show sift and polyphen before mutation assessor was fetched completely

image

fix https://github.com/cBioPortal/cbioportal/issues/6867

### DIFF
--- a/src/shared/components/annotation/genomeNexus/PolyPhen2.tsx
+++ b/src/shared/components/annotation/genomeNexus/PolyPhen2.tsx
@@ -5,8 +5,8 @@ import classNames from 'classnames';
 import tooltipStyles from "./styles/polyPhen2Tooltip.module.scss";
 
 export interface IPolyPhen2Props {
-    polyPhenPrediction: string; // benign, possibly_damaging, probably_damging
-    polyPhenScore: number;
+    polyPhenPrediction: string | undefined; // benign, possibly_damaging, probably_damging
+    polyPhenScore: number | undefined;
 }
 
 export function hideArrow(tooltipEl: any) {
@@ -22,9 +22,14 @@ export default class PolyPhen2 extends React.Component<IPolyPhen2Props, {}> {
         this.tooltipContent = this.tooltipContent.bind(this);
     }
 
-    public static download(polyPhenScore: number, polyPhenPrediction: string): string
+    public static download(polyPhenScore: number | undefined, polyPhenPrediction: string | undefined): string
     {
-        return `impact: ${polyPhenPrediction}, score: ${polyPhenScore}`;
+        if (polyPhenScore || polyPhenPrediction) {
+            return `impact: ${polyPhenPrediction}, score: ${polyPhenScore}`;
+        }
+        else {
+            return "NA";
+        }
     }
 
     public render() {
--- a/src/shared/components/mutationTable/MutationTable.tsx
+++ b/src/shared/components/mutationTable/MutationTable.tsx
@@ -37,6 +37,7 @@ import VariantCountCache from "shared/cache/VariantCountCache";
 import PubMedCache from "shared/cache/PubMedCache";
 import MutationCountCache from "shared/cache/MutationCountCache";
 import GenomeNexusCache from "shared/cache/GenomeNexusCache";
+import GenomeNexusMutationAssessorCache from "shared/cache/GenomeNexusMutationAssessorCache";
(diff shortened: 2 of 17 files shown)


src/shared/lib/MutationTableDownloadDataFetcher.ts
```diff
@@ -86,6 +94,17 @@ export class MutationTableDownloadDataFetcher implements ILazyMobXTableApplicati
         }
     }
 
+    private async fetchAllGenomeNexusFunctionalImpactData()
+    {
+        if (this.mutationData.result)
+        {
+            return await fetchGenomeNexusFunctionalImpactData(this.mutationData.result);
```
alisman: since the function that calls this is going to need to handle the response as a promise (await or .then) there is no reason to put await here.  you can just do return fetchGenomeNexusFunc(...)
Then you can take the async tag off.
In the else condition, you will need to return Promise.resolve(undefined).

src/shared/lib/MutationTableDownloadDataFetcher.ts
```diff
@@ -86,6 +94,17 @@ export class MutationTableDownloadDataFetcher implements ILazyMobXTableApplicati
         }
     }
 
+    private async fetchAllGenomeNexusFunctionalImpactData()
+    {
+        if (this.mutationData.result)
```
alisman: these fetch functions should really take mutationData as a parameter instead of accessing it from instance.  they really ought to just be functions instead of methods.

src/shared/lib/MutationTableDownloadDataFetcher.ts
```diff
@@ -86,6 +94,17 @@ export class MutationTableDownloadDataFetcher implements ILazyMobXTableApplicati
         }
     }
 
+    private async fetchAllGenomeNexusFunctionalImpactData()
```
alisman: @leexgh per your slack question, i think actually that the check for mutationData should be handled outside of this function, for example by the remoteData instance that consumes it.  this function should assume that there is data passed to it (mutationData should be passed as argument).  So ...
1. pass mutationData as arg
2. no check for mutatationData
3. should not be tagged async
4. perhaps this function is not necessary at all since all it will do is return invocation

leexgh:
@inodb got it, will do

leexgh:
@alisman @inodb could you take a look again? thanks

src/shared/components/annotation/genomeNexus/PolyPhen2.tsx
```diff
@@ -22,7 +22,7 @@ export default class PolyPhen2 extends React.Component<IPolyPhen2Props, {}> {
         this.tooltipContent = this.tooltipContent.bind(this);
     }
 
-    public static download(polyPhenScore: number, polyPhenPrediction: string): string
+    public static download(polyPhenScore: number | null, polyPhenPrediction: string | null): string
```
alisman: what happens if these are null?

alisman: i believe that we would actually get the the text "null" printed

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -454,11 +456,11 @@ export default class MutationTable<P extends IMutationTableProps> extends React.
 
         this._columns[MutationTableColumnType.FUNCTIONAL_IMPACT] = {
             name:"Functional Impact",
-            render:(d:Mutation[])=>(this.props.genomeNexusCache
-                ? FunctionalImpactColumnFormatter.renderFunction(d, this.props.genomeNexusCache)
+            render:(d:Mutation[])=>(this.props.genomeNexusCache || this.props.genomeNexusMutationAssessorCache
```
alisman: this is too long for the => shorthant.    lets do => {
....
}

src/shared/components/annotation/genomeNexus/PolyPhen2.tsx
leexgh: i believe it would be "null". actually when I test it I found the functional impact download function was not working, some of my changes might break the function, I'll fix it soon

leexgh: i fix it to show "NA" when there is no data.

alisman:
Merging as https://github.com/cBioPortal/cbioportal-frontend/pull/3020

PR #2920, opened 2019-12-09
Display genes in one column on copy/download

ngocnn1104:
Fix cBioPortal/cbioportal#6889

- Keeping genes in a single column on copy/download

# GIF
gene-panel-modal-download

### DIFF
--- a/src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
+++ b/src/pages/patientView/PatientViewGenePanelModal/GenesList.tsx
@@ -71,7 +71,7 @@ export default class GenesList extends React.Component<
     };
 
     getDownloadData = () => {
-        const downloadData = [['Genes'], ...this.genesDividedToRows(this.props.genePanel.genes)];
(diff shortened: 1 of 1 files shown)


inodb:
@tmazor  you can review it with `localStorage.netlify = deploy-preview-2920--cbioportal-frontend`

I guess it's slightly different from the original request to still have multi columns in the modal, but not in the download. But seems like a good solution for now

tmazor:
@inodb I'm confused... what do you mean this is different? This looks like what I was hoping for - multiple columns in the UI and a single column when I copy it.

I'm not sure if this is just my computer being weird or something about the download, but when I download the table, my computer refuses to let me open it:
image

inodb:
@tmazor sorry, it does indeed do exactly what you requested - my bad. For me the download works fine as well. Not sure if anyone else at Dana Farber can reproduce (CC: @Luke-Sikina) ?

@ngocnn1104 Forgot to say: Really nice work 🎉 & fast ⚡️  !

tmazor:
@inodb the download works just fine for me now also - i think my computer was just having a moment yesterday

PR #2921, opened 2019-12-09
Fix handling of data_priority = profileFilter on submission/modificat…

alisman:
urlWrapper was not correctly handling the legacy data_priority as an alias for profileFilter.  On multi-study queries this broke filtering.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -79,6 +79,7 @@ export type CancerStudyQueryUrlParams = {
     Z_SCORE_THRESHOLD: string;
     RPPA_SCORE_THRESHOLD: string;
     data_priority: '0' | '1' | '2';
+    profileFilter: '0' | '1' | '2';
     case_set_id: string;
     case_ids: string;
     gene_list: string;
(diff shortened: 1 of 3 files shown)


alisman:
@adamabeshouse i'm gonna write a test for this later today

src/shared/components/query/QueryStore.ts
```diff
@@ -2009,7 +2010,7 @@ export class QueryStore {
             : (profileIds.filter(_.identity) as string[]);
         this.zScoreThreshold = params.Z_SCORE_THRESHOLD || '2.0';
         this.rppaScoreThreshold = params.RPPA_SCORE_THRESHOLD || '2.0';
-        this.dataTypePriorityCode = params.data_priority || '0';
+        this.dataTypePriorityCode = params.data_priority || params.profileFilter ||  '0';
```
adamabeshouse: can this parsing be brought into urlwrapper since its handling the aliasing?

alisman: truth be known (is that a phrase? no, but i still like it), we could do away with params.data_priority here because it's an instance of urlWrapper that is being passed.  But then we would have to do some refactoring and i want to just get this out.  we can follow up with refactoring.  won't be too hard

PR #2922, opened 2019-12-09
improve typescript checking for URLWrapper

adamabeshouse:


src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -41,41 +41,40 @@ export type ResultsViewURLQuery = { [key in keyof typeof ResultsViewURLQueryEnum
...
-            { name: "genetic_profile_ids_PROFILE_MUTATION_EXTENDED", isSessionProp:true },
-            { name: "genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION", isSessionProp:true },
-            { name: "genetic_profile_ids_PROFILE_MRNA_EXPRESSION", isSessionProp:true },
-            { name: "genetic_profile_ids_PROFILE_PROTEIN_EXPRESSION", isSessionProp:true },
-            { name: "genetic_profile_ids_PROFILE_GENESET_SCORE", isSessionProp:true },
-            { name: "genetic_profile_ids_GENERIC_ASSAY", isSessionProp:true },
-            { name: "genetic_profile_ids", isSessionProp:true },
-            ],
+                // session props here
+                gene_list: { name: "gene_list", isSessionProp:true },
```
alisman: would it be worth it to get rid of name field and just use the key so as to avoid duplication and also the possibility that someone mis-types the name field?

adamabeshouse: yes, thats a great idea

PR #2924, opened 2019-12-10
Gene query limit supports case sets

Luke-Sikina:
Fix cBioPortal/cbioportal#6898

- Issue:
  - Case sets were ignored when determining if a gene query
  was over the limit
  - This caused case sets that are a proper subset of their
  studies to limit the number of genes too strictly
- Solution:
  - Consider all the ways the user could be restricting
  their sample set when creating a query


src/shared/components/query/QueryStore.ts
```diff
@@ -1784,18 +1785,46 @@ export class QueryStore {
...
 
+    /**
+     * Sample count can come from the following areas:
+     * 1. The sum of the samples in the selected study(s)
+     * 2. The number of cases (patients or samples) in caseIds
+     *     This is _approximate_. Patients can have multiple samples
+     * 3. The number of samples in the selected sample list
+     * 4. The number of samples in the selected profiled samples result
+     */
+    @computed get approxSampleCount(): number {
```
Luke-Sikina: Even though this will be accurate the vast majority of the time, I felt it was important to label it as approximate given that it can be off if the user has a custom case list of patients that have more than one sample.

PR #2925, opened 2019-12-10
Fix overwriting of empty parameters on modify query (and all other updateURL calls)

alisman:
Prior to URLWrapper, modify query cleared all existing query params and replaced them with newly submitted.  But ULWrapper changed that to allow persisting of existing non-URL sessions.  A problem arose when we submit an empty parameter (e.g. mutation profile type).  On submit the code was stripping out empty params which had the affect of retaining the old value.  THis is fixed by accomplishing the overwrite in the routing store where we ALWAYS overwrite all the URL params.  

Fixes: https://github.com/cBioPortal/cbioportal/issues/6894

To reproduce bug, make a single study query and then modify query by updating genomic profile.  You can't update it.

### DIFF
--- a/src/shared/components/query/QueryStoreUtils.ts
+++ b/src/shared/components/query/QueryStoreUtils.ts
@@ -18,6 +18,7 @@ export type MolecularProfileQueryParams = Pick<CancerStudyQueryUrlParams,
 export function currentQueryParams(store:QueryStore) {
     let nonProfileParams = nonMolecularProfileParams(store);
     let profileParams = molecularProfileParams(store);
+    debugger;
     return queryParams(nonProfileParams, profileParams);
 }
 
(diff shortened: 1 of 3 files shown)


src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -90,6 +72,13 @@ export default class ExtendedRouterStore extends RouterStore {
             newQuery = newParams;
         }
 
+        // clear out any undefined props
```
alisman: newQuery will have ALL params in URL, so here we can just delete entrees that are empty. nothing can survive.

src/shared/components/query/QueryStoreUtils.ts
```diff
@@ -25,14 +25,6 @@ export function queryParams(nonMolecularProfileParams:NonMolecularProfileQueryPa
                             molecularProfileParams:MolecularProfileQueryParams) {
     let params:CancerStudyQueryUrlParams = Object.assign({}, nonMolecularProfileParams, molecularProfileParams);
 
-    // Remove params with no value, because they may cause problems.
-    // For example, the server will always transpose if transpose_matrix is present, no matter the value.
-    for (let key in params) {
```
alisman: doing this here meant that the routing store updateURL call would not get a value for this param, which would then allow any OLD param to be retained

PR #2926, opened 2019-12-11
oncoprint - allow adding gaps to string-valued clinical tracks

adamabeshouse:
closes cBioPortal/cbioportal#3336


jjgao:
Nice! Could you connect the issue and use closing keywords in the comment.

jjgao:
For [this query](https://cbioportal-f-gaps-ikexbpaukirv.herokuapp.com/results?Z_SCORE_THRESHOLD=2.0&cancer_study_id=coadread_tcga_pub&cancer_study_list=coadread_tcga_pub&case_ids=&case_set_id=coadread_tcga_pub_nonhypermut&clinicallist=SEX&gene_list=KRAS%20NRAS%20BRAF&gene_set_choice=user-defined-list&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&oncoprint_cluster_profile=&oncoprint_sortby=), if you sort by Sex and then Show gaps, the gap is not in the right place.

image

adamabeshouse:
@jjgao fixed the issue.

It was an interesting bug: I determine the position of each column in the shader, part of this is a binary search which checks the list of gap locations to figure out how many gaps are before a column. 

Here's where it gets tricky - the shader language does not allow while loops. So instead I was using a for loop and estimating the number of loops I would need, at maximum, to fully execute the binary search. I made an error there and so the binary search was terminating too early in this case and not computing the right offset.

I fixed it by rewriting the for loop like this `for (int i=0; i<1/0; i++)` to simulate a while loop (I was surprised that this is allowed).

Sjoerd-van-Hagen:
@adamabeshouse 
> I fixed it by rewriting the for loop like this `for (int i=0; i<1/0; i++)` to simulate a while loop (I was surprised that this is allowed).

In most languages for is just syntactic sugar for while
`for ( int i = 0; i < n; i++
  //do stuff`
equals
`int i = 0
while i < n
  //do stuff
  i++`

So any while-true-loop can be written as 
`for (; true ;)`

I am wondering about the `i<1/0`, is it not easier to just put `true` or `1<2` to make it more obvious what you intention? What do you get from the `1/0`, I am guessing some infinity value, and not a division by zero error?

adamabeshouse:
@Sjoerd-van-Hagen webgl shader language has very specific rules, so `true` and `1<2` are not valid conditions for a for loop. however, i<1/0 is. it can also be done as `for(int i=0; i==0; i+=0)`, but I don't think thats more clear than `for(int i=0; i<1/0; i++)`. yes 1/0 is infinity

adamabeshouse:
https://deploy-preview-2926--cbioportalfrontend.netlify.com/results?Z_SCORE_THRESHOLD=2.0&cancer_study_id=coadread_tcga_pub&cancer_study_list=coadread_tcga_pub&case_set_id=coadread_tcga_pub_nonhypermut&clinicallist=SEX%2CCANCER_TYPE_DETAILED&gene_list=KRAS%20NRAS%20BRAF&gene_set_choice=user-defined-list&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations

PR #2933, opened 2019-12-13
Frontend should consume postData written to JSP

alisman:
External applications may send users to portal using http post (for large queries).  JSP writes this posted data to postData variable.  Frontend needs to consume this data.

### DIFF
--- a/end-to-end-test/local/specs/core/resultspage.spec.js
+++ b/end-to-end-test/local/specs/core/resultspage.spec.js
@@ -1,0 +1,49 @@
+var assert = require('assert');
+var goToUrlAndSetLocalStorage = require('../../../shared/specUtils').goToUrlAndSetLocalStorage;
+var postDataToUrl = require('../../../shared/specUtils').postDataToUrl;
+var parse = require('url-parse');
+var _ = require('lodash');
+
+var useExternalFrontend = require('../../../shared/specUtils').useExternalFrontend;
+
+const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+
+describe('study select page', function() {
+
+    if (useExternalFrontend) {
+
+        describe('error messaging for invalid study id(s)', function(){
+
+            it('show error alert and query form for single invalid study id',function(){
+                var url = `${CBIOPORTAL_URL}/results?localdist=true`;
+
+                const query = {"gene_list":"KRAS NRAS BRAF"
+                    ,"cancer_study_list":"coadread_tcga_pub",
+                    "case_ids":"","case_set_id":"coadread_tcga_pub_nonhypermut",
+                    "Z_SCORE_THRESHOLD":"2.0",
+                    "genetic_profile_ids_PROFILE_MUTATION_EXTENDED":"coadread_tcga_pub_mutations",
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -103,6 +103,8 @@ export default class ResultsViewPage extends React.Component<
 
         handleLegacySubmission(this.urlWrapper);
 
+        handlePostedSubmission(this.urlWrapper);
```
pvannierop: So if I understand this correctly, only the ResultsViewPage can handle POST requests?

end-to-end-test/local/specs/core/resultspage.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+var useExternalFrontend = require('../../../shared/specUtils').useExternalFrontend;
+
+const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, "");
+
+describe('study select page', function() {
+
+    if (useExternalFrontend) {
+
+        describe('error messaging for invalid study id(s)', function(){
```
pvannierop: I am lost here. This file `resutspage.spec.js` has describe `study select page` which belongs to Query Page. So, what does this test actually test? Is there also a test that checks for correct working of the POST method? It is not apparent from the test titles. It seems to only deal with checking of error messages.

alisman: achhh.  terrible, i forgot to update these descriptions.  they are totally wrong.

alisman: i will make PR to correct these.

src/pages/resultsView/ResultsViewPage.tsx
alisman: that's correct.    the server writes the post data in all cases though so any page can in theory access it.

end-to-end-test/local/specs/core/resultspage.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+                const query = {"gene_list":"KRAS NRAS BRAF"
+                    ,"cancer_study_list":"coadread_tcga_pub",
+                    "case_ids":"","case_set_id":"coadread_tcga_pub_nonhypermut",
+                    "Z_SCORE_THRESHOLD":"2.0",
+                    "genetic_profile_ids_PROFILE_MUTATION_EXTENDED":"coadread_tcga_pub_mutations",
+                    "genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION":"coadread_tcga_pub_gistic"
+                };
+
+                postDataToUrl(url, query);
```
alisman: data is posted to page here

end-to-end-test/local/specs/core/resultspage.spec.js
```diff
@@ -0,0 +1,49 @@
...
+
+                browser.waitUntil(()=>{
+                    var url = browser.getUrl();
+
+                    // make sure param in query is passed to url
+                    return _.every(query,(item,key)=>{
```
alisman: @pvannierop here we check that every posted parameter key has made it into the url.  this could only happen as a result of our code transferring the postData variable into the URL.

PR #2589, opened 2019-08-07
Give option to list unselected cases in the Custom Selection component

zhx828:
This solves https://github.com/cBioPortal/cbioportal/issues/6361

PR #2936, opened 2019-12-16
Give option to list unselected cases in the Custom Selection component

zhx828:
This solves https://github.com/cBioPortal/cbioportal/issues/6361

This is an improvement pr of https://github.com/cBioPortal/cbioportal-frontend/pull/2589

### DIFF
--- a/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
+++ b/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
@@ -35,6 +35,9 @@ const GroupByOptions: { value: ClinicalDataType, label: string; }[] = [
     {value: ClinicalDataTypeEnum.PATIENT, label: 'By patient ID'}
 ];
 
+enum SelectMode {
+    SELECTED, UNSELECTED
+}
 @observer
 export default class CustomCaseSelection extends React.Component<ICustomCaseSelectionProps, {}> {
     private validateContent: boolean = false;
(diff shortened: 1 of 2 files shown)


src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
```diff
@@ -76,14 +79,26 @@ export default class CustomCaseSelection extends React.Component<ICustomCaseSele
...
     @autobind
     @action
-    onClick() {
-        let cases = this.props.selectedSamples.map(sample => {
+    onClick(selectMode: SelectMode) {
+        let selectedCases;
+        if (selectMode === SelectMode.SELECTED) {
+            selectedCases = this.props.selectedSamples;
+        } else {
+            const _selectedCaseIds = _.reduce(this.props.selectedSamples, (acc, next) => {
```
kalletlak: may be `_.keyBy(this.props.selectedSamples,sample=>sample.uniqueSampleKey)`

zhx828: @kalletlak I can update. But do they have same performance? Or just for convenience?

kalletlak: just a cleaner way

alisman:
I think the UI is a tiny bit confusing now.    I would suggest something like this it's more clear they are opposites of each other.  

image

zhx828:
@alisman very similar. fontawesome does not have the exact icon in your screenshot
Screen Shot 2019-12-19 at 5 44 18 PM

inodb:
@zhx828 @alisman @kalletlak it seems like this PR didn't go through product review? Let's discuss on engineering meeting how we can avoid these issues

PR #2940, opened 2019-12-18
Bookmark feature should give user fullURL (and shortened version if b…

alisman:
Bookmark feature should allow user to bookmark even non-core (non-session) state params.  Because those params CANNOT be saved to session right now, they are discarded
Fixes https://github.com/cBioPortal/cbioportal/issues/6297


src/shared/lib/ExtendedRouterStore.ts
```diff
@@ -21,6 +21,8 @@ export interface PortalSession {
 }
 
 export function saveRemoteSession(data:any){
+
+    console.log("saving session", data);
```
onursumer: do we want to keep this `console.log`?

alisman: fixed! thanks

PR #2942, opened 2019-12-19
Update study links in results page query summary

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6925

Screen Shot 2019-12-20 at 3 05 33 PM

Screen Shot 2019-12-23 at 2 53 59 PM

Screen Shot 2019-12-20 at 3 05 11 PM

### DIFF
--- a/src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
+++ b/src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
@@ -1,8 +1,8 @@
 import {getPercentage} from "../../../shared/lib/FormatUtils";
 import * as React from "react";
-import { SampleList, Sample } from "shared/api/generated/CBioPortalAPI";
+import { SampleList, Sample, CancerStudy } from "shared/api/generated/CBioPortalAPI";
 import _ from "lodash";
-import { VirtualStudy } from "shared/model/VirtualStudy";
+import {buildCBioPortalPageUrl} from "../../../shared/api/urls";
 
 export function getPatientSampleSummary(
     samples:any[],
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
```diff
@@ -64,18 +64,18 @@ export function getAlterationSummary(
...
+export function submitToStudyViewPage(
+    queriedStudies: Pick<CancerStudy, "studyId">[],
     samples: Pick<Sample, "sampleId" | "studyId">[],
     hasVirtualStudies: boolean,
     sampleLists?: Pick<SampleList, "category">[]) {
     const hasAllCaseLists = _.some(sampleLists, sampleList => sampleList.category === 'all_cases_in_study');
-    let studyViewFilterHash: string | undefined;
     if (samples.length > 0 && (hasVirtualStudies || !hasAllCaseLists)) {
         const sampleIdentifiers = samples.map(sample => ({
             sampleId: sample.sampleId,
```
zhx828: If we are going to remove studyIds later on in study view, we better use uniqueSampleId here

kalletlak: This is equivalent of `sampleIdentifiers` in `StudyViewFilter`, which doesn't support `uniqueSampleIds`

zhx828: Oh right, I misunderstood the structure. Sorry about that.

src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
```diff
@@ -64,18 +64,18 @@ export function getAlterationSummary(
...
             studyId: sample.studyId
         }));
-        studyViewFilterHash = `filterJson=${JSON.stringify({ sampleIdentifiers })}`;
+        window.localStorage.setItem("legacyStudyViewPageFilter", `filterJson=${JSON.stringify({ sampleIdentifiers })}`);
```
alisman: so this is where i think we use the the data passing trick we talked about in slack instead of localStorage.

src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
```diff
@@ -64,18 +64,18 @@ export function getAlterationSummary(
...
     }
-    return studyViewFilterHash;
+    window.open(buildCBioPortalPageUrl(`study`, { id: queriedStudies.map(study => study.studyId).join(',') }), "_blank");
```
alisman: this is where i think we should do 
1. open
2. window.whateverData = data

alisman:
If I do a[ combined query ](https://www.cbioportal.org/results/cancerTypesSummary?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=acc_tcga_pan_can_atlas_2018%2Cblca_dfarber_mskcc_2014&case_set_id=w_cna&data_priority=0&gene_list=TP53&geneset_list=%20&profileFilter=0&tab_index=tab_visualize)and select a subset of patients, i see the wrong count in study name.    

image

kalletlak:
@alisman looks like an existing bug.
@jjgao should I change it to `Combined Study (<all samples count> Samples)`?

alisman:
@kalletlak i'm pretty sure it should be yes.  it's exactly the same scenario (where a user has chosen studies and then selected a case list)

kalletlak:
@alisman updated

PR #2944, opened 2019-12-19
Change the order and put "Matched Samples" above ARM 

victoria34:
Fix https://github.com/oncokb/oncokb/issues/1865


onursumer:
@victoria34 it might be better if we have more descriptive commit messages like "changed the order of..."

victoria34:
> @victoria34 it might be better if we have more descriptive commit messages like "changed the order of..."

Thanks. I have updated commit info and will keep that in mind.

PR #2945, opened 2019-12-19
improve footer version logic

inodb:
Only add v prefix if the API does not return the v prefix

Avoids getting things like `vv3.1.8` in the footer

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -80,7 +80,7 @@ export class AppStore {
             const portalVersionResult = await internalClient.getInfoUsingGET({});
             if (portalVersionResult && portalVersionResult.portalVersion) {
                 let version = portalVersionResult.portalVersion.split('-')[0];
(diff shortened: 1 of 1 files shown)


src/AppStore.ts
```diff
@@ -79,7 +79,11 @@ export class AppStore {
         invoke:async()=>{
             const portalVersionResult = await internalClient.getInfoUsingGET({});
             if (portalVersionResult && portalVersionResult.portalVersion) {
-                return Promise.resolve("v" + portalVersionResult.portalVersion.split('-')[0]);
+                let version = portalVersionResult.portalVersion.split('-')[0];
+                if (!version.startsWith("v")) {
```
alisman: startsWith is probably not supported in IE.  I'll test it.

alisman: nevermind it works and we use this elsewhere

PR #2948, opened 2019-12-20
Patient view mutation table should, by default, show all mutations, n…

alisman:
Change default behavior of patient view mutations table to show by default all mutations, not just mutations which are profiled in all gene panels.
https://github.com/cBioPortal/cbioportal/issues/6890


alisman:
@Sjoerd-van-Hagen @pvannierop not sure why this default was chosen so I wanted to give you an opportunity to approve this.

Sjoerd-van-Hagen:
@alisman it is actually suppose to be false to prevent filtering from being applied.

@pvannierop can you have a look?

alisman:
This PR is wrong.  the property should be default be false.  Problem was in portal.properties.EXAMPLE file

PR #2951, opened 2019-12-23
Update api and custom.sh

dippindots:
update documents and the custom.sh

### DIFF
--- a/env/custom.sh
+++ b/env/custom.sh
@@ -1,3 +1,3 @@
 # set e.g.
 # export CBIOPORTAL_URL="http://localhost:8080"
-export BACKEND=dippindots:override_option_when_import
+export BACKEND=cBioPortal:release-3.2.0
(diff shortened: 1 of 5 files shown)


env/custom.sh
```diff
@@ -1,3 +1,3 @@
 # set e.g.
 # export CBIOPORTAL_URL="http://localhost:8080"
-export BACKEND=dippindots:override_option_when_import
+export BACKEND=cBioPortal:release-3.2.0
```
inodb: @dippindots i think we can remove this line, the backend image is determined based on the branch it is running on

inodb: see e.g. https://github.com/cBioPortal/cbioportal-frontend/blob/master/env/custom.sh

dippindots: Thanks @inodb, sure I can remove this line. Since we decide to postpone the change to the path of the endpoint, should I just change this pr to `update custom.sh`?

dippindots:
I saw you already done that, I will close this pr.

PR #2956, opened 2019-12-26
Update screenshots after expression data revert

inodb:
The expression data z score calculation was reverted for several
studies, so this reverts the screenshots related to that


inodb:
Some stuff still seems to fail, but going to merge regardless since it's not screenshot failures

PR #2963, opened 2019-12-31
Always showing patient icons when there is an alteration

onursumer:
Fix cBioPortal/cbioportal/issues/6891

# Checks


src/pages/patientView/mutation/column/TumorColumnFormatter.tsx
```diff
@@ -33,29 +33,46 @@ export default class TumorColumnFormatter {
...
                 // hide labels for non-existent mutation data
                 // decreased opacity for uncalled mutations
                 // show not-profiled icon when gene was not analyzed
                 const isMutated = sample.id in mutatedSamples;
                 const isProfiled = sample.id in profiledSamples && profiledSamples[sample.id];
-                
+
+                let extraTooltipText = '';
+
+                if (isMutated) {
```
alisman: @onursumer i think this is a great spot where we need a comment to explain why something can be mutated but not profiled.   as @tmazor explained to me, this can indeed happen.  mutations can be read which are not part of the target of the gene panel.  so lets just put that comment here.

PR #2966, opened 2020-01-03
Fix env_vars bool checks

inodb:
Change all bash bool variable checks in the env_var script. This fixes an issue for setting the BRANCH variable when running localdb e2e test on master or release branches in backend repo.


inodb:
See also: https://github.com/cBioPortal/cbioportal/pull/6971

PR #2971, opened 2020-01-07
add hgvsg column and link out to genome nexus

leexgh:
add hgvsg column in mutation table and link out to genome nexus

- hide by default
- click hgvsg will go to genome nexus variant page
- long hgvsg will show `...` and the whole hgvsg will show in tooltip
- compute hgvsg in frontend directly
image


src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
```diff
@@ -0,0 +1,107 @@
...
+        } else if (genomeNexusCacheData.data === null) {
+            status = TableCellStatus.NA;
+        } else {
+            let hgvsgData = HgvsgColumnFormatter.getData(genomeNexusCacheData.data);
+            if (hgvsgData == null) {
+                return null;
+            }
+            else {
+                return (
+                    <a href={`https://www.genomenexus.org/variant/${hgvsgData}`} target="_blank" rel="noopener noreferrer">
```
alisman: lets make a helper in urls.ts to make these links.   the host should only be in once place in our code

src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
```diff
@@ -0,0 +1,107 @@
...
+                        {hgvsgData}
+                        <img
+                            height='16'
+                            width='16'
+                            src={require("./GenomeNexusLogo.png")}
```
alisman: would be best if we accomplished this with css.   so, make a <div> assign it a class and give it a background and dimensions in css.

src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
```diff
@@ -0,0 +1,107 @@
...
+                            alt='Genome Nexus'
+                        />
+                    </a>
+                );
+            }        
+        }
+
+        if (status !== null) {
+            // show loading circle
+            if (status === TableCellStatus.LOADING) {
```
alisman: it's weird that we have a separate case for loading.  one would think that <TableCellStatusIndicator> would take care of loading state.

src/shared/components/mutationTable/column/hgvsg.module.scss
```diff
@@ -0,0 +1,22 @@
+.hgvsg-data a,
```
alisman: lets discuss using SASS for nesting these defs

src/shared/lib/MutationUtils.ts
```diff
@@ -232,7 +232,18 @@ export function extractGenomicLocation(mutation: Mutation)
 // TODO remove when done refactoring mutation mapper
 export function genomicLocationString(genomicLocation: GenomicLocation)
 {
-    return `${genomicLocation.chromosome},${genomicLocation.start},${genomicLocation.end},${genomicLocation.referenceAllele},${genomicLocation.variantAllele}`;
+    // mapping chromosome X with 23
```
alisman: would be good to have a test for this function

src/shared/lib/MutationUtils.ts
```diff
@@ -232,7 +232,18 @@ export function extractGenomicLocation(mutation: Mutation)
...
+    let chromosome = "";
+    if (genomicLocation.chromosome === "X") {
```
alisman: you should use a switch statement instead of if/else when testing against a single value (genomicLocation.chromosome).   Good to get in habit of using switch in these kinds of situations.  You could also use a an enum with a default.  If genomicLocation.chromosome exists in enum, you use it's value.  If not, you use Enum.default.

src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
leexgh: https://github.com/cBioPortal/cbioportal/issues/6988
issue created about the loading state

leexgh:
@alisman fixed, thanks

leexgh:
@tmazor @schultzn could you take a look and give some feedback on the hgvsg column? this column links out to Genome Nexus variant page
https://deploy-preview-2971--cbioportalfrontend.netlify.com/results/mutations?cancer_study_list=5c8a7d55e4b046111fee2296&case_set_id=all&gene_list=EGFR
CC: @jjgao

inodb:
@leexgh as discussed in person: let's hide the hgvsg column for now (make it optional) and do a new PR that adds a tooltip for the protein change column with a link to Genome Nexus.

inodb:
@leexgh looks good to me! Thanks!

We discussed to add a tooltip to the Protein Change column that will also link out to Genome Nexus, which will be addressed in a separate PR

 Since we decided to hide the hgvsg column, maybe ok to release @jjgao @schultzn @tmazor ?

PR #2972, opened 2020-01-07
Update to custom tab support

kalletlak:


alisman:
@kalletlak based on @jjgao feedback from meeting today (1/7), do you think that it would make senes to merge this feature into the portal codebase?   
it's good that we are trying out and refining the custom tab functionality but we should only us it if we are really prepared to support it in the long run.   if there's the possiblity that this feature could be valuable as a native feature, we should consider that.

kalletlak:
yeah sure. The feedback what JJ suggested is bit different from what it is right now, and I think its new feature. This pr specifically fixes updates just custom tab.

alisman:
@kalletlak ah you are right.  so, it's a separate issue whether your feature actually gets implemented as a custom tab.

PR #2975, opened 2020-01-09
use new oncoprint version, which uses numbers for shape specification

adamabeshouse:


src/shared/lib/ObjectManipulation.ts
```diff
@@ -27,3 +27,15 @@ export function dropKeys<T>(dict:T, keys:(keyof T)[]):T {
         return newDict;
     }, {});
 }
+
```
alisman: is this different from _.assign?

adamabeshouse: no, good catch

PR #2976, opened 2020-01-09
Fix cursor issue with oql editor and improve styling and behavior 

alisman:
The cursor would sometimes skip to the end during typing when validation service responded with certain timing in relation to further typing.  
Also applied a minimum height to the validation error tokens the appear below textarea when a user types a nonexistant or ambiguous gene name.  This allows bubbles to appear without forcing other elements to move.

image


src/shared/components/query/GeneSetSelector.tsx
```diff
@@ -133,11 +133,6 @@ export default class GeneSetSelector extends QueryStoreComponent<{}, {}> {
                             <i className={'fa fa-external-link'} />
                         </a>
                     }
-                    promises={[
```
alisman: don't want to show loader anymore.  it's pointless and distracting to typing user

PR #2980, opened 2020-01-13
Study View: Always show selected genes on top and alternate selected rows color

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6896

Fixes
- [x] Always show selected genes on top of table(selected rows will be ignored from sorting)
- [x] Alternate selected rows color

Screen Shot 2020-01-13 at 2 03 28 PM


src/pages/studyView/table/GeneTable.tsx
```diff
@@ -367,6 +368,36 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
+        return _.chain(this.tableData)
+            .filter(data => this.flatterendFilters.includes(data.uniqueKey))
+            .sortBy<GeneTableRow>(data => ifNotDefined(order[data.uniqueKey], Number.POSITIVE_INFINITY))
+            .value();
+    }
+
+    @computed
+    get selectedRowsIndexes(): GeneTableUserSelectionWithIndex[] {
+        return this.selectedRows.map(row => ({
+            uniqueKey: row.uniqueKey
```
alisman: seems strange.  why not just an array of uniqueKey?

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -416,73 +447,40 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
         } else {
-            this.preSelectedRows = _.xorBy(this.preSelectedRows, [record], "rowIndex");
+            this.preSelectedRows = _.xorBy(this.preSelectedRows, [record], "uniqueKey");
         }
     }
 
     @autobind
     @action
     afterSelectingRows() {
+        console.log(JSON.stringify(this.preSelectedRows))
```
alisman: console

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -78,32 +99,53 @@ export default class FixedHeaderTable<T> extends React.Component<IFixedHeaderTab
...
         if (index > -1 && this.isSelectedRow(this._store.dataStore.sortedFilteredData[index])) {
-            return classnames(styles.row, styles.highlightedRow);
+            const classNames: string[] = [styles.row]
+            if (this.props.selectedRowClassName) {
+                const className = this.props.selectedRowClassName(this._store.dataStore.sortedFilteredData[index]);
+                if (className !== null) {
+                    classNames.push(className);
+                }
+            } else {
+                classNames.push(styles.highlightedRow);
```
adamabeshouse: a bit confusing terms - using both highlighted and selected to describe the same thing, right? maybe the prop should be called `highlightedRowClassName`?

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -367,6 +368,36 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
         return this.isFilteredByCancerGeneList ? _.filter(this.props.promise.result, data => data.isCancerGene) : (this.props.promise.result || []);
     }
 
+    @computed get flatterendFilters() {
```
adamabeshouse: flatterend?

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -367,6 +368,36 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
+        }
+        const order = stringListToIndexSet(this.flatterendFilters);
+        return _.chain(this.tableData)
+            .filter(data => this.flatterendFilters.includes(data.uniqueKey))
+            .sortBy<GeneTableRow>(data => ifNotDefined(order[data.uniqueKey], Number.POSITIVE_INFINITY))
+            .value();
+    }
+
+    @computed
+    get selectedRowsIndexes(): GeneTableUserSelectionWithIndex[] {
```
adamabeshouse: shouldn't it be called `selectedRowsKeys`?

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -416,73 +447,40 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
-                []
-            );
-        }
-    }
-
     @autobind
     isSelectedRow(data: GeneTableRow) {
         return !_.isUndefined(
-            _.find(_.union(this.selectedRows, this.preSelectedRows), function (row) {
+            _.find(_.union(this.selectedRowsIndexes, this.preSelectedRows), function (row) {
```
adamabeshouse: this is a linear search through rows, which will run every time isSelectedRow is called, which I'm assuming is once per row to check how to render it? this means quadratic time, which is really slow for long lists. it's better to make selectedRowsIndexes/Keys be a set so that we can check membership in constant time.

adamabeshouse: also, `_.union(this.selectedRowsIndexes, this.preSelectedRows)` doesn't need to be recomputed for each call -  putting that computation into a `@computed` would make it cached and reusable.

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -416,73 +447,40 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
         );
     }
 
+    @autobind
+    selectedRowClassName(data: GeneTableRow) {
+        const index = _.findIndex(this.props.filters, filter => filter.includes(data.uniqueKey));
+        if (index === -1) {
+            return null;
+        }
+        return index % 2 === 0 ? styles.highlightedEvenRow : styles.highlightedOddRow;
```
adamabeshouse: you don't need to do this with code, you can do it with CSS, for eample in simpleTable/styles.scss we do:
```
tbody>tr:nth-of-type(odd).highlighted {
    background-color: #9FAFD1 !important;
  }

  tbody>tr:nth-of-type(even) { // solid background - hard coded hack, react-bootstrap .table-striped makes ODD # rows grey
    background-color: rgb(255,255,255);
  }

  tbody>tr:nth-of-type(even).highlighted {
    background-color: #B0BED9;
  }
```

kalletlak: Here the color is not alternating. alternating color for each selection (there can be more than one row in each selection)

adamabeshouse: ohhh, I see

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -399,88 +423,49 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
         return !!this.props.cancerGeneFilterEnabled && this.props.filterByCancerGenes;
     }
 
+    @computed get allSelectedRowsKeys() {
+        return [...this.selectedRowsKeys, ...this.preSelectedRowsKeys];
```
adamabeshouse: is there any chance these overlap? i.e., its okay not to use _.union?

src/pages/studyView/table/GeneTable.tsx
```diff
@@ -399,88 +423,49 @@ export class GeneTable extends React.Component<GeneTableProps, {}> {
...
+    }
+
     @autobind
     isChecked(uniqueKey: string) {
-        return rowIsChecked(uniqueKey, this.preSelectedRows, this.selectedRows);
+        return  _.some(this.allSelectedRowsKeys, (key) => key === uniqueKey);
```
adamabeshouse: is there a possibility that there will be very many selected rows? if so, then I think we should use a map/set so that checking if uniqueKey is included is constant time

src/pages/studyView/table/GeneTable.tsx
adamabeshouse: is there a chance of many selected rows? if so then this process will be slow - it would be faster to have a map that takes uniquekey to the first index of filter in which it appears, so that you can check this in constant time

src/pages/studyView/table/GeneTable.tsx
kalletlak: there is no chance. even if there is overlap it should be a problem, its just for lookup

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -47,6 +50,24 @@ const RVSDTtoStrType = {
...
 };
 
+export class FixedHeaderTableDataStore extends SimpleGetterLazyMobXTableApplicationDataStore<any> {
+
+    constructor(getData: () => any[], fixedTopRowsData: any[]) {
+        super(getData);
+        this.fixedTopRowsData = fixedTopRowsData;
+    }
+
+    @observable private fixedTopRowsData: any[];
```
alisman: minor but seems like fixedTopRows is actually immutable from perspective of this store.  it's never changed during life of store instance.  if so, it doesn't need to be observable.

PR #2981, opened 2020-01-14
Add civic support for annotation column in react-mutation-mapper

onursumer:
- Moved the implementation of Civic column into `react-mutation-mapper` package so that we can use it in the standalone mutation mapper.
- Partially improved the [original implementation](https://github.com/cBioPortal/cbioportal-frontend/pull/367), there is still room for improvement.

# Checks


packages/react-mutation-mapper/src/component/civic/Civic.tsx
```diff
@@ -19,46 +25,46 @@ export function hideArrow(tooltipEl: any) {
...
+    const variants = _.values(civicEntry.variants);
+    const values: string[] = [];
 
-        // TODO actually this indicates that we have an entry but the evidence is empty
-        if (values.length === 0) {
-            return "NA";
-        }
+    if (variants && variants.length > 0 && variants[0].evidence)
+    {
+        _.toPairs(variants[0].evidence).forEach(pair => {
```
alisman: why bother with toPairs?  _.forEach() can iterate over object's members, passing key and value to callback

packages/react-mutation-mapper/src/component/civic/CivicCard.tsx
```diff
@@ -24,7 +26,7 @@ export default class CivicCard extends React.Component<ICivicCardProps, {}> {
 
         if (variantMap) {
```
alisman: i think we don't needs this outside check because any falsy state of variantMap will be handled by _.isEmpty

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -620,6 +623,35 @@ class DefaultMutationMapperStore implements MutationMapperStore
...
+    readonly civicGenes: MobxPromise<ICivicGene | undefined> = remoteData({
+        await: () => [
+            this.mutationData,
+        ],
+        invoke: async() => this.config.enableCivic ?
+            fetchCivicGenes(this.mutationData.result || [], this.getDefaultEntrezGeneId) : {},
+        onError: () => {
+            // fail silently
+        }
+    }, undefined);
```
alisman: i think the default is undefined by default, so, not sure this makes sense.  the intention of the default was to provide an empty array or object so that consumer code didn't have to do check against undefined

packages/react-mutation-mapper/src/component/civic/Civic.tsx
onursumer: I haven't actually refactored the old implementation, just made the column compatible with `react-mutation-mapper`. Although this line seems like it is newly added, it is probably due to some cleanup / indentation change. Here is the original code: https://github.com/cBioPortal/cbioportal-frontend/pull/2981/files#diff-29ff9c92a56351a845f83f8ab38f6cd0L49

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+import {Mutation} from "../model/Mutation";
+
+export type MutationSpec = {gene:{hugoGeneSymbol: string}, proteinChange: string};
+
+const civicClient = new CivicDataFetcher();
+
+/**
+ * Asynchronously adds the given variant from a gene to the variant map specified.
+ */
+function addCivicVariant(variantMap: ICivicVariant, variantId: number, variantName: string, geneSymbol: string, geneId: number): Promise<void> {
```
alisman: validateAndAddCivicVariantAsync ?

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+ * Asynchronously return a map with Civic information from the genes given.
+ */
+export function getCivicGenes(entrezGeneIds: Array<number>): Promise<ICivicGene> {
+    
+    let civicGenes: ICivicGene = {};
+
+    // Assemble a list of promises, each of which will retrieve a batch of genes
+    let promises: Array<Promise<Array<ICivicGeneData>>> = [];
+    let ids: Array<number> = [];
+    entrezGeneIds.forEach(function(entrezGeneId) {
```
alisman: this logic is a little confusing.  i would just start by splitting the ids into batches of 400 and then use that list to produce the promises.   then you don't have to invoke getCivicGenesBatch twice

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+            ids = [];
+        }
+    });
+    if (ids.length > 0) {
+        let requestIds = ids.join();
+        promises.push(civicClient.getCivicGenesBatch(requestIds));
+    }
+
+    // We're waiting for all promises to finish, then return civicGenes
+    return Promise.all(promises).then(function(responses) {
```
alisman: use reduce

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+            }
+        }
+    } else {
+        for (let geneName in civicGenes) {
+            let geneEntry = civicGenes[geneName];
+            let geneVariants = geneEntry.variants;
+            if (!_.isEmpty(geneVariants)) {
+                for (let variantName in geneVariants) {
+                    // Only retrieve CNA variants
+                    if (variantName == 'AMPLIFICATION' || variantName == 'DELETION') {
```
alisman: can we get this from an enum somewhere?

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+    if (civicGenes[geneSymbol] && civicVariants[geneSymbol] && civicVariants[geneSymbol][mutation.proteinChange]) {
+        let geneVariants: {[name: string]: ICivicVariantData} = {[mutation.proteinChange]: civicVariants[geneSymbol][mutation.proteinChange]};
+        let geneEntry: ICivicGeneData = civicGenes[geneSymbol];
+        civicEntry = buildCivicEntry(geneEntry, geneVariants);
+    }
+
+    return civicEntry;
+}
+
+export function getCivicStatus(civicGenesStatus:"pending" | "error" | "complete",
```
alisman: there is function which aggregates the status of a set of mobxpromises.  looking for it.  i think it could be used here.

alisman: getMobxPromiseGroupStatus

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    return "pending";
+}
+
+export async function fetchCivicGenes(mutations: Partial<Mutation>[],
+                                      getEntrezGeneId: (mutation: Partial<Mutation>) => number)
+{
+    if (mutations.length === 0) {
+        return {};
```
alisman: i think this should be Promise.resolve({})  ?

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,203 @@
...
+    }
+
+    let entrezGeneSymbols: Set<number> = new Set([]);
+
+    mutations.forEach(mutation => {
```
alisman: map?  no reason to produce set and then convert to array

onursumer:
@alisman Thanks for your review. Most of the changes introduced by this PR are actually because of relocating/restructuring the existing code segments. This is mostly the original Civic implementation by The Hyve (https://github.com/cBioPortal/cbioportal-frontend/pull/367). 

I am happy to spend more time to improve the code quality if you think we should address those issues before merging this PR.

packages/react-mutation-mapper/src/util/CivicUtils.ts
onursumer: I think an async function always (implicitly) returns Promise: https://stackoverflow.com/a/35302535. But we can still do an explicit Promise.resolve({}) if you think it is better.

packages/react-mutation-mapper/src/util/CivicUtils.ts
adamabeshouse: agreed. _.chunk would work well for this

packages/react-mutation-mapper/src/util/CivicUtils.ts
adamabeshouse: I think it should be Promise.resolve and dont make the function async and just return getCivicGenes

src/shared/lib/CivicUtils.ts
```diff
@@ -196,3 +39,51 @@ export function getCivicCNAVariants(
     }
     return geneVariants;
 }
+
+export async function fetchCivicGenes(
```
adamabeshouse: all of these functions can be non-async that just return a promise. i think that's more typical to how we've been doing things

packages/react-mutation-mapper/src/util/CivicUtils.ts
```diff
@@ -0,0 +1,224 @@
...
+
+enum AlterationType {
+    AMPLIFICATION = 'AMPLIFICATION',
+    DELETION = 'DELETION',
+}
+
+/**
+ * Asynchronously adds the given variant from a gene to the variant map specified.
+ */
+function addCivicVariant(
```
alisman: I kind of hate this function since it hides the "side effects".

alisman: i see that you are using this multiple times.  maybe change the name of this function to lookupCivicVariantAndAdd ToMap

packages/react-mutation-mapper/src/component/civic/Civic.tsx
```diff
@@ -19,43 +24,41 @@ export function hideArrow(tooltipEl: any) {
...
-        if (civicEntry) {
-            score = 1;
-        }
+    return score;
+}
 
-        return score;
+export function download(civicEntry: ICivicEntry | null | undefined): string {
+    if (!civicEntry) {
+        return 'NA';
```
adamabeshouse: if theres any parsing of these values, it would be better to use and reuse a constant with value "NA". If it's just for display then NVM.

onursumer: probably no parsing for these values, but not 100% sure. Let's just add a constant in `CivicUtils`, it wouldn't hurt.

PR #2982, opened 2020-01-14
Impl. download of OAuth2/OpenID Connect DAT tokens (triggered by backend)

pvannierop:
# Background
cBioPortal provides tokens, also referred to as Data Access Tokens or DATs, that allow programmatic access the cBioPortal API. These tokens, as well as user permissions, are managed by the cBioPortal instance.

# Problem
cBioPortal can be configured to use an external Identity Provider (IDP), such as Keycloak, that handles user authentication and user authorization. This setup is not supported by current DAT implementation.

# Solution
We have extended the frontend so that cBioPortal can provide an OpenID Connect _offline token_ (=OAuth2 refresh token with scope _offline_access_) to the user. Responsibility for triggering the download of all token types (JWT/UUID/OAuth2) has been moved to the cBioPortal backend (is handled by the frontend at present).

# Details
- The backend handles the download because for OpenID Connect this involves communication in the _back channel_ between cBioPortal backend and the Identity Provider (IDP).

### DIFF
--- a/src/pages/staticPages/webAPI/WebAPIPage.tsx
+++ b/src/pages/staticPages/webAPI/WebAPIPage.tsx
@@ -2,14 +2,11 @@ import * as React from 'react';
 import { observer } from 'mobx-react';
 import { observable } from 'mobx';
 import { PageLayout } from '../../../shared/components/PageLayout/PageLayout';
-import internalClient from 'shared/api/cbioportalInternalClientInstance';
 import './styles.scss';
 import Helmet from 'react-helmet';
-import { Link } from 'react-router';
 import AppConfig from 'appConfig';
-import { isNullOrUndefined } from 'util';
-import fileDownload from 'react-file-download';
 import { getBrowserWindow } from 'cbioportal-frontend-commons';
+import { buildCBioPortalAPIUrl } from 'shared/api/urls';
 
 export class UserDataAccessToken {
     @observable token: string;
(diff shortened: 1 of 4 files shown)


src/pages/staticPages/webAPI/WebAPIPage.tsx
```diff
@@ -24,45 +24,19 @@ export class UserDataAccessToken {
...
-        }
-    }
-
     async downloadDataAccessTokenFile() {
-        const dat = this.generateNewDataAccessToken();
-        if (!isNullOrUndefined(dat)) {
-            const fileContents = buildDataAccessTokenFileContents(await dat);
-            fileDownload(fileContents, "cbioportal_data_access_token.txt");
-        }
+        window.open("/api/data-access-token", "_blank");
```
alisman: does this open data access token in a new window?   we want file download, yes?

pvannierop: Yes, this opens a new tab. The backend sends a 'download file' response header to this tab. The browser downloads the DAT token and closes the tab.

n1zea144: @pvannierop This doesn't work in all scenario's.  In my test case, I have a web app running in a local container accessible at the following url: 

> localhost:8080/cbioportal/

When I select the "Download Token" button, I get the error below - seems like the entire app path is not considered...

image

pvannierop:
@n1zea144 well spotted. Issue has been fixed in new commit.

src/pages/staticPages/webAPI/WebAPIPage.tsx
```diff
@@ -29,59 +29,25 @@ export class UserDataAccessToken {
...
-    }
-
     async downloadDataAccessTokenFile() {
-        const dat = this.generateNewDataAccessToken();
-        if (!isNullOrUndefined(dat)) {
-            const fileContents = buildDataAccessTokenFileContents(await dat);
-            fileDownload(fileContents, 'cbioportal_data_access_token.txt');
-        }
+        const tokenUrl =
+            window.location.pathname.replace('webAPI', '') +
```
alisman, line 40: use ```buildCBioPortalAPIUrl``` to create this url.  it should automatically insert /api/

PR #2984, opened 2020-01-14
Use SAMPLE_ACQUISITION track for sample order

inodb:
This is part of prototype for the GENIE BPC study. Order the labels of
the samples on the patient view by whatever the order is in the
SAMPLE_ACQUISITION track. Wait for timeline events, such that the sample
labels aren't shown before they are properly calculated using the
timeline event info.

Related to https://github.com/cBioPortal/cbioportal/issues/6982


inodb:
Merging this as it is a pretty innocuous change related to GENIE BPC prototype that we need by tomorrow

inodb:
FYI: case this applies to is https://triage.cbioportal.mskcc.org/patient?caseId=GENIE-MSK-P-0012393&studyId=genie_bpc_test

PR #2986, opened 2020-01-14
Update oncoprint version for performance improvements

adamabeshouse:


alisman:
@adamabeshouse looks like there's maybe some broken tests that could be associated?

PR #2987, opened 2020-01-15
Show gnomad in number

leexgh:
show gnomad in number with 2 non-zero digits after decimal point

left align with different length of number
image

### DIFF
--- a/packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
+++ b/packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
@@ -13,7 +13,6 @@ import * as React from "react";
 
 import {GnomadSummary} from "../../model/GnomadSummary";
 import GnomadFrequencyTable, {GnomadTableColumnName} from "./GnomadFrequencyTable";
-import { significantDigits } from "../../util/FormatUtils";
 
 const GNOMAD_POPULATION_NAME: {[key:string]: string} = {
     "African" : "afr",
@@ -146,7 +145,11 @@ export default class GnomadFrequency extends React.Component<GnomadFrequencyProp
                 display = <span>0</span>;
             } else {
                 // show frequency as number with 4 significant digits
-                display = <span>{significantDigits(result['Total'].alleleFrequency, 4)}</span>;
+                var significantDigitsFormatter = new Intl.NumberFormat("en", { 
+                    minimumSignificantDigits: 1,
+                    maximumSignificantDigits:4
+                });
+                display = <span>{significantDigitsFormatter.format(result['Total'].alleleFrequency)}</span>;
             }
 
             overlay = () => <GnomadFrequencyTable data={sorted} gnomadUrl={gnomadUrl} />;
--- a/packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
+++ b/packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
@@ -148,8 +148,8 @@ export default class GnomadFrequency extends React.Component<GnomadFrequencyProp
                 display = (
                     <span>
                         {result['Total'].alleleFrequency.toLocaleString(undefined, {
-                            maximumSignificantDigits: 4,
-                            minimumSignificantDigits: 1,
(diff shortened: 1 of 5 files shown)


packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
```diff
@@ -145,7 +145,8 @@ export default class GnomadFrequency extends React.Component<GnomadFrequencyProp
             if (result["Total"].alleleFrequency === 0) {
                 display = <span>0</span>;
             } else {
-                display = <span>{parseFloat(result["Total"].alleleFrequency.toString()).toExponential(1)}</span>;
+                // show frequency as number with 4 significant digits
+                display = <span>{significantDigits(result['Total'].alleleFrequency, 4)}</span>;
```
alisman: we have a library called Numeral.js (google it for docs) which should do this. lets use it instead.

leexgh: @alisman thanks! Basically what I'm looking for is to "round" number to 4 significant digits. I take a look at the Numeral.js doc but I'm not sure if it has the function I need, but I found another way to do it:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat
do you think it's ok to use?

packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
```diff
@@ -146,7 +145,11 @@ export default class GnomadFrequency extends React.Component<GnomadFrequencyProp
                 display = <span>0</span>;
             } else {
                 // show frequency as number with 4 significant digits
-                display = <span>{significantDigits(result['Total'].alleleFrequency, 4)}</span>;
+                var significantDigitsFormatter = new Intl.NumberFormat("en", {
```
alisman: Yes, I think this is fine.  Support looks good.  However, I would subsistute the following https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString

It has the same options and doesn't confuse so much with "Intl" 

Also, it seems that passing the locale ("en") is optional.  If so, we should just not pass anything as we are not localized and we intend to use same formatting for any locale.

packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
alisman: Ok, i see you are right numeral.js doesn't seem to have min/max significant digits.  I put another comment in.    As a general rule, try to avoid writing library code unless necessary and discuss with team if you think you do need to do so.

packages/react-mutation-mapper/src/component/gnomad/GnomadFrequency.tsx
leexgh: @alisman thanks, one minor thing is that it looks like I have to pass a locale anyway, so I pass `undefined` there.

packages/react-mutation-mapper/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "react-mutation-mapper",
-  "version": "0.4.2",
+  "version": "0.4.3",
```
onursumer: we should also increment the version for `cbioportal-frontend` package.json

alisman:
awaiting @jjgao approval of the number column treatment

jjgao:
per discuss on slack, we'll keep two significant digits, e.g. 0.000016.

leexgh:
@jjgao @alisman @onursumer updated to 2 significant digits, could you take a look again? thanks

packages/react-mutation-mapper/src/util/FormatUtils.ts
```diff
@@ -14,4 +14,4 @@ export function formatPercentValue(rate: number, fractionDigits: number = 1)
 
 export function numberOfLeadingDecimalZeros(value: number) {
```
alisman: doesn't look like this is function is used anymore?

PR #2989, opened 2020-01-15
Add comparison tab to results view, including enrichments tabs

adamabeshouse:
TO-DO:
- [x] remove old enrichments tabs code
- [x] add tests


jjgao:
Cool! Thanks, @adamabeshouse. It looks great 👍 

@schultzn @tmazor @cBioPortal/product do you think we should move Survival into Comparison or better leave it separate?

tmazor:
@jjgao Survival does seem to make sense within the Comparison tab, but if we make that change, we may need to help users learn to look there. I have a slight preference to put it in Comparison, but really I can go either way on that decision. @schultzn ?

src/pages/resultsView/survival/SurvivalTab.tsx
```diff
@@ -117,7 +117,11 @@ export default class SurvivalTab extends React.Component<ISurvivalTabProps, {}>
                             yLabelTooltip="Survival estimate"
                             xLabelWithEventTooltip="Time of death"
                             xLabelWithoutEventTooltip="Time of last observation"
-                            fileName="Overall_Survival" />
+                            fileName="Overall_Survival"
```
alisman: do we need this fine anymore?

alisman:
@adamabeshouse just want to ask how session is dealt with in context of resultsview.   We don't have comparison session anymore here, right?  it's just whatever is defined by results view as the two groups?

adamabeshouse:
@alisman right, the sessionId stuff is not put into the superclass ComparisonStore because it is only for GroupComparisonStore. in the results view the groups are just altered/unaltered and we disabled group creation from the overlap tab

inodb:
@adamabeshouse could you maybe add some screenshots or a GIF to this PR? Useful for the release notes + news item.

PR #2991, opened 2020-01-16
Fix error while fetching referenceGenomeSize

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7008

### DIFF
--- a/src/pages/patientView/genomicOverview/tracksHelper.ts
+++ b/src/pages/patientView/genomicOverview/tracksHelper.ts
@@ -92,7 +92,7 @@ export function getChmInfo(genomeBuild:string) {
     if (!referenceGenome || referenceGenome === "") {
         referenceGenome = DEFAULT_GENOME_BUILD;
     }
-    const genomeSize = referenceGenomeSizes[genomeBuild];
+    const genomeSize = referenceGenomeSizes[referenceGenome];
     if (genomeSize) {
         sel.genomeRef = genomeSize;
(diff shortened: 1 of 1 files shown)


src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -92,7 +92,7 @@ export function getChmInfo(genomeBuild:string) {
     if (!referenceGenome || referenceGenome === "") {
         referenceGenome = DEFAULT_GENOME_BUILD;
     }
-    const genomeSize = referenceGenomeSizes[genomeBuild];
+    const genomeSize = referenceGenomeSizes[referenceGenome];
```
alisman: do we have a sense why this working for other patients?  seems like we were using the key instead of the value

kalletlak: for other patients `ncbiBuild` of [first mutation](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/genomicOverview/Tracks.tsx#L58) is in one of `referenceGenomeSizes` keys

khzhu: hi @alisman , @kalletlak , my mistake, should be **_referenceGenome_**. Thanks!

alisman: thanks @khzhu !!!

PR #2994, opened 2020-01-17
Add quick oql edit feature to results view page next to current gene list

alisman:
Add a pencil button near gene list in results page which opens interface for quickly modifying the oql of the query.

image


alisman:
@adamabeshouse gonna add a e2e test for this when approved by product

jjgao:
Since the edit dialog will only be visible after click edit, can we make it bigger, e.g. the same height as in Query?

image

jjgao:
Also, if I click edit and them click "Modify Query", the "quick edit" dialog should be closed.

PR #2995, opened 2020-01-17
Study View: ignore case while applying filters on pie chart

kalletlak:
More updates to fix https://github.com/cBioPortal/cbioportal/issues/5853

### DIFF
--- a/src/pages/studyView/charts/pieChart/PieChart.tsx
+++ b/src/pages/studyView/charts/pieChart/PieChart.tsx
@@ -35,9 +35,18 @@ export default class PieChart extends React.Component<IPieChartProps, {}> implem
         super(props);
     }
 
+    @computed
+    get filters() {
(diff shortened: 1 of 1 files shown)


src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -35,9 +35,18 @@ export default class PieChart extends React.Component<IPieChartProps, {}> implem
         super(props);
     }
 
+    @computed
+    get filters() {
```
alisman: this could really use a comment to explain what's happening

inodb:
Thanks so much @kalletlak !  Would be good to add a test as well: https://github.com/cBioPortal/cbioportal/issues/7016. Could be good GSoC issue maybe for someone to get familiar with e2e tests

PR #2998, opened 2020-01-21
Change footer version logic

inodb:
Use x.y.z if the branch name is like release-x.y.z, otherwise use
portalVersion (same as latest tag). This avoids having to tag the release (or using `PROJECT_VERSION` env variable) to set the footer version. Instead it assumes that when one is building from the `release-x.y.z` branch, that the version is closer to `x.y.z` than the latest tag.


inodb:
Merging this cause it's pretty innocuous. I filed a separate issue to add a more descriptive version page: https://github.com/cBioPortal/cbioportal/issues/7025.

PR #2999, opened 2020-01-21
SV: Fix failing test

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7021


inodb:
@kalletlak so was this expected? Did we change the format of the URL filter for fusions?

PR #3000, opened 2020-01-21
Filter fusion and structural varinat profiles while getting non mutation profiles

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7024

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -126,6 +126,7 @@ import {
     isRNASeqProfile,
     makeEnrichmentDataPromise,
     getMolecularProfiles,
+    excludeMutationAndSVProfiles,
 } from './ResultsViewPageStoreUtils';
 import MobxPromiseCache from '../../shared/lib/MobxPromiseCache';
 import {isSampleProfiledInMultiple} from '../../shared/lib/isSampleProfiled';
@@ -202,6 +203,7 @@ export const AlterationTypeConstants = {
     GENESET_SCORE: 'GENESET_SCORE',
     METHYLATION: 'METHYLATION',
     GENERIC_ASSAY: 'GENERIC_ASSAY',
+    STRUCTURAL_VARIANT: 'STRUCTURAL_VARIANT'
 };
 
 // only show TREATMENT_RESPONSE in the plots tab for now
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1211,11 +1212,7 @@ export class ResultsViewPageStore {
         ],
         invoke: () => {
             // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
-            const profilesWithoutMutationProfile = _.filter(
-                this.selectedMolecularProfiles.result,
-                (profile: MolecularProfile) =>
-                    profile.molecularAlterationType !== 'MUTATION_EXTENDED'
-            );
+            const profilesWithoutMutationProfile = getProfilesWithoutMutationProfile(this.selectedMolecularProfiles.result || []);
```
alisman: probably gauranteed that result will be array here.   i know ts complains so I think trailing ! is sufficient.

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -610,3 +610,7 @@ export function fetchPatients(samples:Sample[]) {
         patientFilter
     });
 }
+
+export function getProfilesWithoutMutationProfile(molecularprofiles: MolecularProfile[]): MolecularProfile[] {
+    return molecularprofiles.filter(profile => !['MUTATION_EXTENDED', "FUSION", "STRUCTURAL_VARIANT"].includes(profile.molecularAlterationType));
```
alisman: can you just check to see if these values are available in an enum somewhere.  i'm pretty sure they are.

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -610,3 +610,8 @@ export function fetchPatients(samples:Sample[]) {
         patientFilter
     });
 }
+
+export function getProfilesWithoutMutationProfile(molecularprofiles: MolecularProfile[]): MolecularProfile[] {
```
inodb: @kalletlak maybe more clear if the name is something like:

`excludeMutationAndSVProfiles`

`FUSION` is a subset of `STRUCTURAL_VARIANT` and will be deprecated in the future anyway

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1211,11 +1213,7 @@ export class ResultsViewPageStore {
         ],
         invoke: () => {
             // we get mutations with mutations endpoint, all other alterations with this one, so filter out mutation genetic profile
```
dippindots: Thanks @kalletlak for this quick fixing! Looks good to me!

Maybe can edit this comment to make it more clear like: 
"we get mutations with mutations endpoint, structural variants and fusions with structural variant endpoint, all other alterations with this one, so filter out mutation genetic profile and structural variant profiles "

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1270,12 +1268,7 @@ export class ResultsViewPageStore {
...
```
dippindots: Also here.

PR #3004, opened 2020-01-22
In localdb e2e tests, don't unnecessarily restart database

alisman:
Unclear why this db restart was left in.  It does not appear to be necessary.

### DIFF
--- a/end-to-end-test/local/docker/setup_docker_containers.sh
+++ b/end-to-end-test/local/docker/setup_docker_containers.sh
@@ -1,6 +1,6 @@
 #!/usr/bin/env bash
 
-set -e 
+set -e
 set -u # unset variables throw error
 set -o pipefail # pipes fail when partial command fails
 shopt -s nullglob # allows files and dir globs to be null - needed in 'for ... do' loops that should not run when no files/dirs are detected by expansion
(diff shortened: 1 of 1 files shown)


end-to-end-test/local/docker/setup_docker_containers.sh
```diff
@@ -51,26 +51,13 @@ build_database_container() {
...
-    (docker stop $DB_HOST && docker rm $DB_HOST) || true
-    docker run -d \
-        --name=$DB_HOST \
-        --net=$DOCKER_NETWORK_NAME \
-        -e MYSQL_ROOT_PASSWORD=$DB_USER \
-        -e MYSQL_USER=$DB_USER \
-        -e MYSQL_PASSWORD=$DB_PASSWORD \
-        -e MYSQL_DATABASE=$DB_PORTAL_DB_NAME \
-        -v "$DB_DATA_DIR:/var/lib/mysql/" \
-        mysql:5.7
```
inodb: @alisman is the restart slow? I guess Q is "how does this part benefit local testing"

alisman: @inodb For some reason this causes things to fail for me.  Everything works fine without it. A while back @pvannierop said he didn't recall why this was necessary and was ok with our removing it

pvannierop: No, its fine to remove this.

PR #3005, opened 2020-01-22
Compile typescript in transpile only mode for optimization in netlify

alisman:
Typescript checking is not important on netlify since we catch this on circleci.   This speeds up build by about ~30%


alisman:
Duplicate

PR #3009, opened 2020-01-23
On circleCI do not run buildModules twice

alisman:
We already run `yarn build` which includes `buildModules`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -89,7 +89,6 @@ jobs:
           - v2-dependencies-{{ checksum "yarn.lock" }}
       # Download and cache dependencies
       - run: yarn
-      - run: yarn buildModules
       - run:
           name: "Make sure lock file is still the same"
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -89,7 +89,6 @@ jobs:
           - v2-dependencies-{{ checksum "yarn.lock" }}
       # Download and cache dependencies
       - run: yarn
-      - run: yarn buildModules
```
inodb: @alisman where else is it run? Can't seem to find it. Is it run as part of some other yarn job?

onursumer: @inodb `yarn build` runs `yarn buildModules` as of the latest setup

inodb: gotcha - thanks!

alisman:
@onursumer @inodb seems that this may be necessary after all for unit test run.  anyhow, it doesn't actually seem to improve Circle CI build time, so gonna close it.

PR #3010, opened 2020-01-24
Study view: support old chart Ids when retrieving user preference settings

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7046

CC: @jjgao

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -88,6 +88,8 @@ import {
     getFilteredAndCompressedDataIntervalFilters,
     getUniqueKeyFromMolecularProfileIds,
     getMolecularProfileIdsFromUniqueKey,
+    getClinicalDataBySamples,
+    updateSavedUserPreferenceChartIds,
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
 import { SingleGeneQuery } from 'shared/lib/oql/oql-parser';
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2512,23 +2512,31 @@ export class StudyViewPageStore {
     private loadSettings(chartSettngs: ChartUserSetting[]){
         this.clearPageSettings();
         _.map(chartSettngs, chartUserSettings => {
+            const regex = /^(?:PATIENT_|SAMPLE_)([a-zA-Z0-9_]+)/i;
```
zhx828: Let's move this to study  view util.  Also the similar logic converting the chart id to clinical attribute should be put together(getClinicalDataBySamples method)

kalletlak: I don't see any similarity between getClinicalDataBySamples and chart id. Do you mean to move getClinicalDataBySamples method to utils?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2417,3 +2418,102 @@ export function getFilteredAndCompressedDataIntervalFilters(
...
+        ) {
+            numberOfClinicalAttributeCharts++;
+            let match = chartSetting.id.match(chartIdWithDataTypeRegex);
+            if (!!match) {
+                numberOfChartRequiringUpdates++;
+            }
+        }
+    });
+
+    // Only update ids if all the chart ids require updates.
```
alisman: this comment could use further elaboration.  i just don't really understand business logic here.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2417,3 +2418,102 @@ export function getFilteredAndCompressedDataIntervalFilters(
...
+            clinicalData[sample.uniqueSampleKey] = {
+                ...(clinicalData[sample.uniqueSampleKey] || {}),
+                [item.clinicalAttributeId]: item.value,
+            };
+        });
+    });
+    return clinicalData;
+}
+
+export function updateChartIds(
```
alisman: i think this name should be more descriptive

PR #3015, opened 2020-01-28
Show error screen in case show_oncokb is true, but no oncokb token defined

onursumer:
Related to cBioPortal/cbioportal/issues/7053

oncokb_not_defined

# Checks


src/appShell/App/Container.tsx
```diff
@@ -65,6 +62,27 @@ export default class Container extends React.Component<IContainerProps, {}> {
...
+                    <ErrorScreen
+                        title={'OncoKB token is not available'}
+                        body={
+                            <p>
+                                As of version 3.2.1, all cBioPortal
+                                installations require an access token to use the latest OncoKB data.
+                                You should either disable OncoKB or register an OncoKB account.
+                                Please review these instructions for how to do so.{' '}
+                                <a href="https://github.com/cBioPortal/cbioportal/blob/master/docs/OncoKB-Data-Access.md">
+                                    https://github.com/cBioPortal/cbioportal/blob/master/docs/OncoKB-Data-Access.md
```
onursumer: - Do not merge this PR, before adding the actual documentation.
- Is this the right location for this documentation?

src/appShell/App/Container.tsx
```diff
@@ -65,6 +62,27 @@ export default class Container extends React.Component<IContainerProps, {}> {
             );
         }
 
+        if (ServerConfigHelpers.oncoKbIsEnabledWithoutToken()) {
+            return (
+                <div className="contentWrapper">
+                    <ErrorScreen
+                        title={'OncoKB token is not available'}
```
onursumer: Should we have longer title like `OncoKB is enabled but token is not available`?

inodb: yes that would be nice

src/appShell/App/Container.tsx
inodb: might be nice to directly mention the properties here as well i.e. `show.oncokb` and `oncokb.token`.

The documentation should prolly point to some link at docs.cbioportal.org instead of the markdown directly. That website is prettier :). Could put the anchor on the `instructions` instead of showing full url

We should change the version prolly. 3.2.1 is already released. We can change it right before merging just made  a comment so we remember to do so

src/config/config.ts
```diff
@@ -148,6 +148,10 @@ export class ServerConfigHelpers {
         return config.serverConfig.sessionServiceEnabled;
     }
 
+    static oncoKbIsEnabledWithoutToken() {
+        return config.serverConfig.show_oncokb && !config.serverConfig.oncoKbTokenDefined;
```
zhx828: Isn't it odd that we use different formats for the variables under serverConfig? @onursumer

src/appShell/App/Container.tsx
onursumer: keeping the full url for consistency: https://github.com/cBioPortal/cbioportal-frontend/blob/05017f04f977acd6ef916fe54362ff1627ef35fe/src/appShell/App/Container.tsx#L56

src/config/config.ts
onursumer: definitely! but this needs to be fixed on the back end.

PR #3017, opened 2020-01-28
Run prettier on some more files

inodb:
This adds some files that seems to have been missed.

If we decide to not revert the prettier commit then it might make sense to add this PR before making everybody rebase their PRs.

This basically applies prettier on these files minus what's in the ignore file:

```
yarn run prettier --write $(git ls-files | grep '\(.js\|.ts\|.scss\|.css\)')
```
(I added this to package.json as `prettierAll`)

I added three more entries to the ignore file:

```
*.template
*.ejs
**/generated/**
```

UPDATE: I kept the types files in prettier as discussed, cause it prolly won't cause an issue. They are edited manually. Removed the `**/generated/**` files with prettierignore


alisman:
@inodb looks like we're hitting types/ folder, which we probably shouldn't.   also the generated clients inside packages.    i think we should add these to the ignore

PR #3025, opened 2020-01-30
Generic assay frontend implementation

dippindots:
Fix cBioPortal/cbioportal#6943
Fix cBioPortal/cbioportal#6366

- a make all generic assay data available in the oncoprint tab as heatmap track
image

    tooltip (link and description)
image

- b make all generic assay data available in the plots tab
image

    Entities selected from the oncoprint (added heatmap tracks) will list as the top options:
image

    info icon (link and description)
image

**Please Note**
The `mutational signatures` data is not the final data, we are still in the process of deciding the `name` and `description` of each signature. Ref: https://github.com/cbioportal/cbioportal/issues/7198
**How to test:**
1. go to https://triage.cbioportal.mskcc.org and select `msk_impact_2017` study.
2. press `F12` to open the browser console and paste this into console then enter: `localStorage.setItem("netlify", "deploy-preview-3025--cbioportalfrontend")`
3. check the new features in oncoprint tab (result view),  plots tab(result view).


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -555,11 +555,22 @@ export class ResultsViewPageStore {
             : [];
     }
 
-    @computed get treatmentList() {
-        return this.urlWrapper.query.treatment_list &&
-            this.urlWrapper.query.treatment_list.trim().length
-            ? this.urlWrapper.query.treatment_list.trim().split(/;/)
+    @computed get parsedGenericAssayGroups() {
+        // empty if no heatmap tracks param
+        const groups = this.urlWrapper.query.generic_assay_groups
```
alisman: we should make this a parsing helper function and have a test for it.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -555,11 +555,22 @@ export class ResultsViewPageStore {
...
+            ? this.urlWrapper.query.generic_assay_groups
+                  .split(';')
+                  .map((x: string) => x.split(','))
             : [];
+
+        const parsedGroups = groups.reduce(
+            (acc: { [molecularProfileId: string]: string[] }, group) => {
+                acc[group[0] as string] = group.slice(1);
```
alisman: why not use group[1] ? instead of slice?  just curious

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3858,22 +3869,46 @@ export class ResultsViewPageStore {
...
+        },
+    });
+
+    readonly selectedGenericAssayEntitiesGroupByGenericAssayType = remoteData<{
+        [genericAssayType: string]: GenericAssayMeta[];
+    }>({
+        await: () => [this.genericAssayEntitiesGroupByGenericAssayType],
         invoke: () => {
-            const treatmentIdFromUrl = this.treatmentList;
+            const selectedGenericAssayEntitiesFromUrl: {
```
alisman: Strange that you have to type this variable.   Should be inferred from parsedGenericAssayGroups.

Also, why do we even bother assigning to local variable rather than just referencing directly.   I actuallyt think "selectedGenericAssayEntities" is a good name for this computed property.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3904,21 +3939,36 @@ export class ResultsViewPageStore {
...
             if (
-                this.molecularProfilesInStudies.result &&
-                this.molecularProfilesInStudies.result.length > 0
+                !_.isEmpty(
+                    this.genericAssayEntitiesGroupByGenericAssayType.result
+                )
             ) {
-                const treatments = await fetchTreatmentByMolecularProfileIds(
-                    this.molecularProfilesInStudies.result
+                const genericAssayTypes = _.keys(
```
alisman: i think the following code can be rewritten more simply as mapValues on genericAssayEntitiesGroupByGenericAssayType.result

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3904,21 +3939,36 @@ export class ResultsViewPageStore {
...
-                treatments.forEach(({ treatmentId, refLink }) => {
-                    linkMap[treatmentId] = refLink;
+                genericAssayTypes.map(genericAssayType => {
+                    const linkMap: { [stableId: string]: string } = {};
+                    this.genericAssayEntitiesGroupByGenericAssayType.result![
+                        genericAssayType
+                    ].forEach(entity => {
+                        // if entity meta contains reference url, add the link into map
+                        linkMap[entity.stableId] =
+                            'URL' in entity.genericEntityMetaProperties
```
alisman: entity.genericEntityMetaProperties.URL || ""

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: Sure!

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: because every group is a string like this `molecular_profile_id,entity_1,entity_2`, the first one in the array stores the value of `molecular_profile_id`, group[1] will just take the `entity_1`, we want the whole array except the first `molecular_profile_id`. That's why I use `group.slice(1)` at here.

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: We don't have a predefined type for `genericEntityMetaProperties`, `genericEntityMetaProperties` is an object returned from the backend API, we don't know what's in it before we try to get the data. So I may not do as you suggested.

dippindots:
@alisman Thanks for the review, updated!

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -173,8 +173,24 @@ export default class ResultsViewOncoprint extends React.Component<
...
+    @computed get selectedGenericAssayEntitiesGroupByGenericAssayTypeFromUrl() {
+        const result = _.reduce(
+            this.props.store.selectedGenericAssayEntities,
+            (acc, value, key) => {
+                if (
+                    this.props.store.molecularProfileIdToMolecularProfile
+                        .result[key]
+                ) {
+                    const type = this.props.store
+                        .molecularProfileIdToMolecularProfile.result[key]
```
alisman: note that we must make sure status of this result is complete.   just make sure that this is not accessed without checking for status

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -207,9 +221,33 @@ export default class OncoprintControls extends React.Component<
 > {
     @observable horzZoomSliderState: number;
     @observable heatmapGenesReady = false;
-    @observable private _selectedTreatmentIds: string[] = [];
-    private textareaTreatmentText = '';
-    @observable treatmentFilter = '';
+    @observable private _selectedGenericAssayEntityIds: string[] =
+        this.props.selectedGenericAssayEntitiesGroupByGenericAssayTypeFromUrl &&
```
alisman: this belongs in constructor

alisman: and in some kind of routine named initialization

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -925,20 +936,60 @@ export default class ResultsViewOncoprint extends React.Component<
...
+        [genericAssayType: string]: ISelectOption[];
+    }>({
+        await: () => [
+            this.props.store.genericAssayEntitiesGroupByGenericAssayType,
+        ],
+        invoke: async () => {
+            return Promise.resolve(
+                genericAssayEntitiesToSelectOptionsGroupByGenericAssayType(
+                    this.props.store.genericAssayEntitiesGroupByGenericAssayType
+                        .result || {}
```
alisman: are you sure that the result can be "falsey" ?

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -925,20 +936,60 @@ export default class ResultsViewOncoprint extends React.Component<
...
+        [genericAssayType: string]: ISelectOption[];
+    }>({
+        await: () => [
+            this.props.store.genericAssayEntitiesGroupByMolecularProfileId,
+        ],
+        invoke: async () => {
+            return Promise.resolve(
+                genericAssayEntitiesToSelectOptionsGroupByGenericAssayType(
+                    this.props.store
+                        .genericAssayEntitiesGroupByMolecularProfileId.result ||
```
alisman: again, i don't think that the result of this promise can return as something that would evaluate to false (false|udefined|null) as long as the await is complete.   please verify that.  this is mostly just code clarity issue.  i.e. i don't think you need do this check.   but i could be wrong.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
dippindots: Oh, It cannot. I should just use this `this.props.store.genericAssayEntitiesGroupByGenericAssayType.result`.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+
+    const genericAssayProfilesGroupByGenericAssayType = _.groupBy(
+        genericAssayProfiles,
+        'genericAssayType'
+    );
+    const genericAssayTypes = _.keys(
+        genericAssayProfilesGroupByGenericAssayType
+    );
+
+    const GenericAssayMetaGroupByGenericAssayType: {
```
alisman: shouldn't be capitalized

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+        [genericAssayType: string]: GenericAssayMeta[];
+    } = {};
+
+    await Promise.all(
```
alisman: this one is kinda weird.  i'm not totally sure why it works because the map function doesn't return a promise as far as I can see.  Promise.all needs an array of promises i think.  i'm not sure why tyepscript allows this.   anyways, you should assign the result of await Promise.all and then use resulting array to populate GenericAssayMetaGroupByGenericAssayType.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+            AlterationTypeConstants.GENERIC_ASSAY
+        );
+    });
+
+    const genericAssayMetaGroupByMolecularProfileId: {
+        [molecularProfileId: string]: GenericAssayMeta[];
+    } = {};
+
+    await Promise.all(
+        genericAssayProfiles.map(async genericAssayProfile => {
```
alisman: use then and make sure the map callback actually returns a promise.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+                // the Swagger-generated type expected by the client method below
+                // incorrectly requires both molecularProfileIds and genericAssayStableIds;
+                // use 'as' to tell TypeScript that this object really does fit.
+            } as GenericAssayMetaFilter,
+        });
+    }
+    return genericAssayMeta;
+}
+
+export async function fetchGenericAssayMetaByEntityIds(entityIds: string[]) {
```
alisman: here, there's no reason to use async/await since it doesn't improve the clarity of your code.  just the the result of client.fetchGenericAssayMetaDataUsingPost which is nicely typed promise.    or, if entetyIds.length = 0, Promise.resolve([])

alisman: same as above.  i think it's a good idea not to use async on fetch functions but instead always return promise.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+        });
+    }
+    return genericAssayMeta;
+}
+
+export async function fetchGenericAssayData(
+    entityIdsByProfile: { [molecularProfileId: string]: string[] },
+    sampleFilterByProfile: { [molecularProfileId: string]: IDataQueryFilter }
+) {
+    const params = Object.keys(entityIdsByProfile).map(profileId => ({
```
alisman: you can use _.map on an object and it will iterate through keys and give you both key and item.  it will result in an array.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -1 +1,132 @@
...
+            ...sampleFilterByProfile[profileId],
+            // the Swagger-generated type expected by the client method below
+            // incorrectly requires both samples and a sample list;
+            // use 'as' to tell TypeScript that this object really does fit.
+        } as GenericAssayDataFilter,
+    }));
+    const dataPromises = params.map(param =>
+        client.fetchGenericAssayDataInMolecularProfileUsingPOST(param)
+    );
+    const results: GenericAssayData[][] = await Promise.all(dataPromises);
```
alisman: if you do this right, TS should be able to infer the return type based on the fact that its an array of whatever is returned by client.fetchGenericAssayDataInMolecularProfileUsingPOST  (Promise<whatever>)

end-to-end-test/local/specs/treatment.screenshot.spec.js
```diff
@@ -30,7 +30,7 @@ describe('treatment feature', () => {
             );
             $('.oncoprint__controls__heatmap_menu textarea').setValue('17-AAG');
             $('div.icon-area div.icon').waitForExist();
-            $('button=Add Treatment Response to Heatmap').click();
+            $('button=Add entity to Heatmap').click();
```
adamabeshouse: is entity a comprehensible word to show to users?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -2129,7 +2198,12 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                 axisSelection.dataType !==
                     AlterationTypeConstants.MUTATION_EXTENDED &&
                 !this.props.store.molecularProfileIdToMolecularProfile
-                    .isComplete)
+                    .isComplete) ||
+            (axisSelection.dataType &&
```
adamabeshouse: when would axisSelection not have a datatype?

src/pages/resultsView/plots/QuickPlots.ts
```diff
@@ -567,7 +567,7 @@ const quickPlots: QuickPlot[] = [
         isApplicableToQuery: (dataTypes: PlotsTabOption[]): boolean => {
             return (
                 dataTypes.find(
-                    dataType => dataType.value === 'GENERIC_ASSAY'
+                    dataType => dataType.value === 'TREATMENT_RESPONSE'
```
adamabeshouse: can we use a constant instead of string TREATMENT_RESPONSE

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -903,9 +914,9 @@ export default class ResultsViewOncoprint extends React.Component<
         const profileMap = this.props.store.molecularProfileIdToMolecularProfile
             .result;
         return (
-            profileMap.hasOwnProperty(this.selectedHeatmapProfile) &&
-            profileMap[this.selectedHeatmapProfile].molecularAlterationType !==
-                'GENESET_SCORE'
+            profileMap.hasOwnProperty(this.selectedHeatmapProfileId) &&
+            profileMap[this.selectedHeatmapProfileId]
+                .molecularAlterationType !== 'GENESET_SCORE'
```
adamabeshouse: lets use a constant instead of the string GENESET_SCORE

end-to-end-test/local/specs/treatment.screenshot.spec.js
inodb: yeah agreed - this should prolly show entity type (i.e. treatment response) or something

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
adamabeshouse: I just checked in the console and passing nonpromises to Promise.all is acceptable.

However that's not what's happening here because `async` causes a function to return a promise

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
adamabeshouse: disagree. async causes the function to return a promise so I think this makes sense

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
adamabeshouse: agree here but not above because in the case up there it was about populating map entries which I don't see as clean a way to do in the way you're suggesting

end-to-end-test/local/specs/treatment.screenshot.spec.js
dippindots: Yeah, I can try to use the existing `generic assay type` at here (e.g. `treatment response` or `mutational signature`). If that's not good enough, we can move forward with this https://github.com/cBioPortal/cbioportal/pull/7132

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
adamabeshouse: @alisman I think the typing for Promise.all is so complicated that the type inference might not work perfectly. might be wrong though.

src/pages/resultsView/plots/PlotsTab.tsx
dippindots: From the definition, I think the dataType can be undefined, so I add this checking
`export type AxisMenuSelection = {
    entrezGeneId?: number;
    genesetId?: string;
    genericAssayEntityId?: string;
    selectedGeneOption?: PlotsTabGeneOption;
    selectedDataSourceOption?: PlotsTabOption;
    selectedGenesetOption?: PlotsTabOption;
    selectedGenericAssayOption?: PlotsTabOption;
    dataType?: string;
    dataSourceId?: string;
    mutationCountBy: MutationCountBy;
    logScale: boolean;
};`

adamabeshouse: ah right, thanks

src/shared/components/oncoprint/controls/OncoprintControls.tsx
adamabeshouse: why not @computed?

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -273,6 +307,45 @@ export default class OncoprintControls extends React.Component<
...
+                .selectedGenericAssayEntitiesGroupByGenericAssayTypeFromUrl[
+                genericAssayType
+            ] &&
+            this.props.state.selectedHeatmapProfileId &&
+            this.props
+                .genericAssayEntitiesSelectOptionsGroupByMolecularProfileId &&
+            this.props
+                .genericAssayEntitiesSelectOptionsGroupByMolecularProfileId[
+                this.props.state.selectedHeatmapProfileId
+            ]
```
adamabeshouse: can you compute this in a separate line/function and give it a name so its easier to follow whats happening

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -509,107 +582,119 @@ export default class OncoprintControls extends React.Component<
...
-            profiles,
-            profile =>
-                profile.molecularAlterationType !==
-                    AlterationTypeConstants.GENERIC_ASSAY ||
-                (profile.molecularAlterationType ===
-                    AlterationTypeConstants.GENERIC_ASSAY &&
-                    profile.genericAssayType ===
-                        GenericAssayTypeConstants.TREATMENT_RESPONSE)
-        );
+    private ononAddAllGenericAssayRemoved(entityId: string) {
```
adamabeshouse: onon

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
dippindots: Updated, TS is able to infer the type from the promised result.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
dippindots: Sure.

src/shared/components/oncoprint/controls/OncoprintControls.tsx
dippindots: oops, will fix.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
dippindots: @alisman I can not use _.map on an object. https://lodash.com/docs/4.17.15#map. The _.map can only be used on collection. So I will change it to chain, maybe that can make this better.

adamabeshouse: @dippindots you can definitely use map on an object, if you look closer the collection can be an Array or an Object. if you test it in the console you can see how it works!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -225,10 +227,9 @@ export const AlterationTypeConstants = {
     STRUCTURAL_VARIANT: 'STRUCTURAL_VARIANT',
 };
 
-// only show TREATMENT_RESPONSE in the plots tab for now
-// TODO: apply to all generic assay profiles when front-end implementation finish
-export const GenericAssayTypeConstants = {
+export const GenericAssayTypeConstants: { [s: string]: string } = {
```
adamabeshouse: lets make a little typescript change here.

step one: remove the typing (`: { [s: string]:string }`)
step two:

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -97,6 +102,26 @@ export const dataTypeDisplayOrder = [
     AlterationTypeConstants.METHYLATION,
     ..._.keys(GenericAssayTypeConstants),
 ];
+
+export function deriveDisplayTextFromGenericAssayType(
+    genericAssayType: string
```
adamabeshouse: step two: `genericAssayType: keyof GenericAssayTypeConstants`

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -643,59 +659,88 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
...
-                    return this._selectedTreatmentOption;
+                    return this._selectedGenericAssayOption;
+                }
+            },
+            set selectedGenericAssayOption(o: any) {
+                this._selectedGenericAssayOption = o;
+            },
+            get isGenericAssayType() {
+                if (!self.dataTypeOptions.isComplete) {
+                    // if there are no options to select a default from, then return false as default
```
adamabeshouse: the comment seems out of date - you're not returning false, right?

src/pages/resultsView/ResultsViewPageStore.ts
dippindots: same as below.

src/pages/resultsView/plots/PlotsTab.tsx
dippindots: no, I am not. I will update this.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
dippindots: Thanks, I will change it to map!

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: ohh okay, thanks for explanation :+1:

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
alisman: @adamabeshouse i would argue that the only reason to use async await is to avoid using a callback.  it's syntatic sugar that allows us to pretend we are programming syncronously.  but we use a callback here anyway in order to populate genericAssayMetaGroupByMolecularProfileId, so i would say that the use of await here actually obscures what's happening.  it's same argument i make for using reduce  or map rather than forEach.    by using assignment, it makes the result of operation clear rather than forcing the developer to look at the internals of the loop to see that it's populating the empty thing declared above.

adamabeshouse: I think I agree with @alisman, and I think that the clearest way to do it would be
```
await Promise.all(
    genericAssayProfiles.map(
        profile=>getData(profile).then(update genericAssayMetaGroupByMolecularProfileId with data)
    )
);
return genericAssayMetaGroupByMolecularProfileId;
```

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
adamabeshouse: After discussion I agree with @alisman and think something like my suggestion below would be most clear here too

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -165,6 +170,11 @@ export type PlotsTabDataSource = {
 };
 
 export type PlotsTabOption = { value: string; label: string };
+export type PlotsTabOptionWithGenericAssayInfo = {
```
alisman: do we have to duplicate PlotsTabOption?   Seems like inheritanace would be good here.  Or just add it as meta node on PlotsTapOption

dippindots: @alisman Should I do 
```
type PlotsTabOptionWithGenericAssayInfo = PlotsTabOption & {isGenericAssayType?: boolean}
```

dippindots: @alisman I think I may just rewrite `PlotsTapOption` to this:
```
export type PlotsTabOption = {
    value: string;
    label: string;
    isGenericAssayType?: boolean;
};
```

PR #3028, opened 2020-01-30
Handle 403 errors properly in global fashion

alisman:
We were showing blank screen on this 403 error.  Should show global error screen

https://genie-private.cbioportal.org/study/summary?id=5e32067be4b01beff45d8174

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -23,7 +23,7 @@ export class AppStore {
                 sendSentryMessage('ERRORHANDLER:' + error);
             } catch (ex) {}
 
-            if (error.status && /400|500/.test(error.status)) {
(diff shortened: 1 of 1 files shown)


src/AppStore.ts
```diff
@@ -23,7 +23,7 @@ export class AppStore {
                 sendSentryMessage('ERRORHANDLER:' + error);
             } catch (ex) {}
 
-            if (error.status && /400|500/.test(error.status)) {
+            if (error.status && /400|500|403/.test(error.status)) {
```
alisman: really we should just handle everything that bubbles up here but i'm not quite ready to do so because there are some endpoints which say 404 to indicate no match (e.g quick search).  So, need to do some more testing

PR #3038, opened 2020-02-04
Add filter bookmark UI to study view page when cohort is filtered

alisman:
When users set filters on study view, they should be able to obtain a shareable link to that filter state:
image


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -183,6 +199,26 @@ export default class StudyViewPage extends React.Component<
         this.props.routing.updateRoute({}, `study/${id}`);
     }
 
+    @observable showBookmarkModal = false;
+
+    @autobind
+    toggleBookmarkModal() {
```
kalletlak: shouldn't this be annotated with `@action` annotation?

alisman: yes

PR #3039, opened 2020-02-04
Study view: fix default group name while creating new group

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7110


inodb:
@kalletlak I created an issue to create a test for this: https://github.com/cBioPortal/cbioportal/issues/7123. Maybe a good GSoC "first issue"? Please add there if you have any pointers how one should go about making a test for this

PR #3040, opened 2020-02-04
exclude mutation uncalled profile and fusion profile

dippindots:
Fix cBioPortal/cbioportal#7099 and https://github.com/cBioPortal/cbioportal/issues/7112

Exclude mutation uncalled profile and fusion profile on download tab.

Note that this temporarily disables downloading fusions in the new fusion data format, which might have not been working in the first place (never tested this afaik). We will re-enable the download of fusions once the fusion tab has been enabled, since there are a few updates to the store there that would make it easier to access the data. There shouldn't be many people that currently have fusion data loaded, so temporarily disabling it should be fine. Filed an issue for this: https://github.com/cBioPortal/cbioportal/issues/7122


dippindots:
@inodb could you review this pr, I want to know if that makes sense to treat `mutation_uncalled` as  `mutation`. We only have this type of profile in the study `glioma_msk_2018`.

inodb:
@dippindots that's correct - the uncalled mutations profile is the same as `mutation` type. It should only be used on the patient view tab. Everywhere else it should be excluded. See also:

https://github.com/cBioPortal/cbioportal/issues/7112

It allows visualizations like this on the patient view:

Screen Shot 2020-02-04 at 5 10 14 PM

Useful for when a mutation wasn't called because there was not enough confidence, but you can still see if there was any signal. We currently only support this on the patient view

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1487,7 +1488,9 @@ export class ResultsViewPageStore {
 
                 if (
                     profile.molecularAlterationType ===
-                    AlterationTypeConstants.MUTATION_EXTENDED
+                        AlterationTypeConstants.MUTATION_EXTENDED ||
+                    profile.molecularAlterationType ===
+                        AlterationTypeConstants.MUTATION_UNCALLED
```
inodb: @dippindots for the `MUTATION_UNCALLED` we should prolly have another `if` category. It should be removed and not show up anywhere on resultsview page.

dippindots: Thanks @inodb, I will update this pr!

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3465,9 +3472,11 @@ export class ResultsViewPageStore {
             ],
             invoke: () => {
                 return Promise.resolve(
-                    _.difference(
-                        this.molecularProfilesInStudies.result!,
-                        this.selectedMolecularProfiles.result!
+                    excludeMutationProfilesForDownload(
```
inodb: @dippindots The name of this `readonly` variable is `nonSelectedMolecularProfiles`, so it seems strange to me to filter it with `excludeForDownload` here. Maybe more comprehensible to have a separate `readonly` variable called `downloadableMolecularProfiles` that depends on this one. Or something like that

inodb:
@dippindots i added an issue to create a suite of tests for the download tab: https://github.com/cBioPortal/cbioportal/issues/7124. It doesn't specifically test the corner case solved here but it would make it easier to test that one as well. Feel free to add any info to that issue that could help newcomers. I think it might be a "good first issue" for potential GSoC students

dippindots:
@inodb That's a great idea, we definitely should have some tests for the download files!

PR #3041, opened 2020-02-05
Fix error while calculating profiled samples count

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7120


kalletlak:
Previously `UniqueKey.MUTATED_GENES_TABLE` were used for chart unique keys. But with recent changes to study-view filter, this has been changed to concatenated molecular profile ids

PR #3046, opened 2020-02-07
Add oncoprint gene track groups to the comparison tab

adamabeshouse:
image

- [x] make screenshot tests


jjgao:
@adamabeshouse thanks! This is awesome. See [this example](http://bit.ly/37hCyFj)

image

jjgao:
Currently, after modifying query, the groups still are not updated. Ideally, we should do the followings:

- "Modify Query"  (study, profile, case list) will rest the comparison tab
- When only genes/tracks are modified, we should update the altered and unaltered groups
- When only genes/tracks are modified, we should try to retain the user-created groups, e.g.
  - if a new track is added (w/o modifying anything else), all previous track groups will remain and a new track group will be added
  - if a track is dropped, the corresponding group (and user-created groups derived from that group) should be removed (is this possible?)

adamabeshouse:
@jjgao ah the fact that it didn't reset was a bug. I fixed it to reset when you submit modify query. 

what you suggested as far as holding onto user-created groups is doable but a little complicated. let's talk about if it's totally necessary.

adamabeshouse:
after talking @jjgao and I decided to never remove user-created groups

tmazor:
This is great @adamabeshouse !

A few small things:
- When including overlapping samples, the KM plot legend in the survival tab is missing the group names:
image
https://www.cbioportal.org/results/comparison?comparison_groupOrder=%5B%22Altered%20group%22%2C%22Unaltered%20group%22%2C%22IDH1%22%2C%22IDH1andEGFR%22%2C%22EGFR%22%2C%22TP53%22%2C%22ATRX%22%2C%22CIC%22%5D&comparison_overlapStrategy=Include&comparison_selectedGroups=%5B%22Unaltered%20group%22%2C%22IDH1%22%2C%22EGFR%22%2C%22TP53%22%2C%22ATRX%22%5D&comparison_subtab=survival&session_id=5e4c253de4b030a3bfcf9076&show_samples=false
- bit.ly URLs aren't working - the link above should be accessible from http://bit.ly/2SOfBUE but that brings me to "This page isn't working"
- There is a horizontal line across the bottom of the page, but the dotted line around the plot extends lower than that horizontal line
https://www.cbioportal.org/results/comparison?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=gbm_tcga_pan_can_atlas_2018%2Clgg_tcga&case_set_id=all&comparison_selectedGroups=%5B%22Unaltered%20group%22%2C%22PTEN%22%2C%22IDH1%22%2C%22ATRX%22%5D&comparison_subtab=clinical&data_priority=0&gene_list=IDH1%2520PTEN%2520ATRX&geneset_list=%20&profileFilter=0&tab_index=tab_visualize
- In the previous link, if you select Tumor Disease Anatomic Site (q<10^-10), the plot shows that all groups are 100% brain, indicating that there are no differences by group. So how is it significant?

cBioPortal appears to be temporarily down, so that's it for now

adamabeshouse:
@tmazor thanks! I fixed two of your issues.

for the second one, regarding bitly, I don't know what to do. it seems like bitly is calling decodeURIComponent too many times on the URL. it sucks. its an issue with bitly that I don't know how to solve. can anyone help?  https://cbioportal.slack.com/archives/C359NNB41/p1582137132262000

the fourth issue looks like its an issue on the backend. @kalletlak can you offer any insight?

kalletlak:
@adamabeshouse looks like plot is based on sample which have clinical data. samples/patients having no clinical data are ignored. I think this can be separate issue.

@jjgao @tmazor should we reduce the denominator to just cases having that clinical data or create a new category(bar) with `NA`?

tmazor:
@kalletlak Are you saying that the calculation of the p/q-values includes samples with NA for an attribute? I'm pretty sure the calculation should only include samples with a value, but maybe @jjgao can confirm that?

tmazor:
@adamabeshouse  The horizontal line is now at the bottom, which is great! But now I'm going to get super nitpicky and ask if you can make it a tiny bit lower? Right now it is essentially even with the bottom of the table controls - can you add some white space between them?

Besides that, this is looking awesome!

In terms of the bitly issues, which part of the URL is causing problems? Is it the group selection settings? Can those be excluded from the bitly encoding? If there's a way to get the bitly URL to go to the right place, even if some settings are lost, I think that's preferable to a completely dead link -- at least as a temporary solution.

adamabeshouse:
I've posted a question on stackoverflow about the bitly issue: https://stackoverflow.com/questions/60513644/keeping-url-params-encoded-in-a-bitly-link

kalletlak:
> @kalletlak Are you saying that the calculation of the p/q-values includes samples with NA for an attribute? I'm pretty sure the calculation should only include samples with a value, but maybe @jjgao can confirm that?

@tmazor Looks like they are included https://github.com/cBioPortal/cbioportal/blob/master/web/src/main/java/org/cbioportal/web/util/ClinicalDataEnrichmentUtil.java#L136

inodb:
@schultzn would you like to test this one?

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -317,6 +317,13 @@ export default class GroupComparisonPage extends React.Component<
         },
     });
 
+    @autobind private isGroupDeletable() {
```
kalletlak: should this be computed?

src/pages/groupComparison/UpSet.tsx
```diff
@@ -60,9 +60,9 @@ const MAX_LABEL_WIDTH = 200;
 
 const BarComponent = (props: any) => (
     <Bar
-        className={`${props.caseType}_${_.sortBy(props.datum.groups).join(
-            '_'
-        )}_bar`}
+        className={`${props.caseType}_${_.sortBy(props.datum.groups)
+            .map(g => g.replace(/ /g, '_'))
```
alisman: curious why you need to make this class

src/pages/groupComparison/groupSelector/GroupSelector.tsx
```diff
@@ -3,24 +3,36 @@ import { observer } from 'mobx-react';
...
 
 @observer
 export default class GroupSelector extends React.Component<
     IGroupSelectorProps,
     {}
 > {
+    @observable collapsedGroupsShown = false;
     private dragging = false;
 
+    @autobind
```
alisman: action

src/pages/groupComparison/groupSelector/GroupSelector.tsx
```diff
@@ -3,24 +3,36 @@ import { observer } from 'mobx-react';
...
+    private toggleCollapsedGroups() {
```
alisman: action

alisman: i know that it's academic because it's only updating one value, but i want to maintain the rule that whenever update state you do it in action because it's a gotcha for other devs

src/pages/resultsView/comparison/ResultsViewComparisonStore.ts
```diff
@@ -52,83 +64,198 @@ export default class ResultsViewComparisonStore extends ComparisonStore {
...
+    readonly nameToUserCreatedGroup = remoteData({
+        await: () => [this._session],
+        invoke: () => {
+            return Promise.resolve(
+                _.keyBy(this._session.result!.groups, g => g.name)
+            );
+        },
+    });
+
+    // <group selection>
```
alisman: errant comment?

alisman: oh, i see, this is some way of grouping things?

src/pages/groupComparison/UpSet.tsx
adamabeshouse: for selecting during e2e test. im not sure why I didn't do data-test. is there a problem with using class?

end-to-end-test/shared/wdio/webdriver-manager.conf.js
```diff
@@ -70,6 +70,7 @@ var config = {
                 args: [
                     '--disable-composited-antialiasing',
                     '--allow-insecure-localhost',
+                    '--no-sandbox',
```
alisman: what's this?

src/pages/resultsView/comparison/ResultsViewComparisonUtils.ts
```diff
@@ -1,6 +1,168 @@
...
     nameOfEnrichmentDirection: string;
     count: number;
 };
+
+export const ALTERED_COLOR = '#dc3912';
+export const UNALTERED_COLOR = '#3366cc';
+export const ALTERED_GROUP_NAME = 'Altered group';
+export const UNALTERED_GROUP_NAME = 'Unaltered group';
+
+export function completeSessionGroups(
```
alisman: maybe a comment explaining what "complete" means

src/pages/resultsView/comparison/ResultsViewComparisonUtils.ts
```diff
@@ -1,6 +1,168 @@
...
+    studyIds: string[],
+    alteredSamples: Sample[],
+    unalteredSamples: Sample[]
+): SessionGroupData[] {
+    return [
+        {
+            name: ALTERED_GROUP_NAME,
+            description: `Number (percentage) of ${
+                patientLevel ? 'patients' : 'samples'
+            } that have alterations in the query gene(s) that also have a deep deletion in the listed gene.`,
```
alisman: not sure if this is existing language,  but might be more clear:  "that have BOTH alteration in query query gene and deep deletion in listed gene."

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -121,19 +122,82 @@ export default class ComparisonStore {
...
+
+    readonly _session: MobxPromise<Session>; // must be implemented in subclasses
     readonly _originalGroups: MobxPromise<ComparisonGroup[]>; // must be implemented in subclasses
     public overlapStrategy: OverlapStrategy; // must be implemented in subclasses
     public usePatientLevelEnrichments: boolean; // must be implemented in subclasses
     readonly samples: MobxPromise<Sample[]>; // must be implemented in subclass
     readonly studies: MobxPromise<CancerStudy[]>; // must be implemented in subclass
     // < / >
 
+    readonly origin = remoteData({
```
alisman: what is origin again?  maybe a comment or more desriptive name

end-to-end-test/shared/wdio/webdriver-manager.conf.js
adamabeshouse: this is the attempt at fixing the circleCI issue

src/pages/resultsView/comparison/ResultsViewComparisonUtils.ts
adamabeshouse: this is existing language, but if you think we should change it we can loop back pretty easily. what do you think, @tmazor ?

src/shared/lib/comparison/ComparisonStore.ts
adamabeshouse: I'll add a comment. it's an existing idiom from the session service, it means the studies that the session (or in this case, the comparison groups) come from

src/pages/groupComparison/GroupComparisonPage.tsx
adamabeshouse: it's implemented as a callback to fit the interface of GroupSelector

PR #3054, opened 2020-02-12
Use selected sample instead of sending complete filter object

kalletlak:
Selected(Filtered) samples is calculated for each study view api with filter object requests. For each selection on study page at least 5 api requests are made (computation logic for finding selected samples is invoked 5 times). Instead of sending complete filter object if we send selected `SampleIdentifier` then this would improve the response time of the api's.

Example for 
Screen Shot 2020-02-12 at 10 11 25 AM

current response times for above selections
Screen Shot 2020-02-12 at 10 10 49 AM

new response times for above selections with this pr
Screen Shot 2020-02-12 at 10 11 18 AM

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -126,7 +126,6 @@ import {
     StudyViewComparisonGroup,
     splitData,
 } from '../groupComparison/GroupComparisonUtils';
-import client from '../../shared/api/cbioportalClientInstance';
 import { LoadingPhase } from '../groupComparison/GroupComparisonLoading';
 import { sleepUntil } from '../../shared/lib/TimeUtils';
 import ComplexKeyMap from '../../shared/lib/complexKeyDataStructures/ComplexKeyMap';
@@ -502,7 +501,7 @@ export class StudyViewPageStore {
                     const entityIdKey = clinicalAttribute.patientAttribute
                         ? 'patientId'
                         : 'sampleId';
-                    let data = await client.fetchClinicalDataUsingPOST({
+                    let data = await defaultClient.fetchClinicalDataUsingPOST({
                         clinicalDataType: clinicalAttribute.patientAttribute
                             ? ClinicalDataTypeEnum.PATIENT
                             : ClinicalDataTypeEnum.SAMPLE,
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -502,7 +501,7 @@ export class StudyViewPageStore {
                     const entityIdKey = clinicalAttribute.patientAttribute
                         ? 'patientId'
                         : 'sampleId';
-                    let data = await client.fetchClinicalDataUsingPOST({
+                    let data = await defaultClient.fetchClinicalDataUsingPOST({
```
alisman: what's the idea of the default client?  it can be switched based on condition?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1933,11 +1932,12 @@ export class StudyViewPageStore {
     }
 
     readonly unfilteredClinicalDataCount = remoteData<ClinicalDataCountItem[]>({
+        await:() => [this.selectedSampleIdentifiersFilter],
         invoke: () => {
             return internalClient.fetchClinicalDataCountsUsingPOST({
                 clinicalDataCountFilter: {
                     attributes: this.unfilteredAttrsForNonNumerical,
-                    studyViewFilter: this.filters,
+                    studyViewFilter: this.selectedSampleIdentifiersFilter.result!,
```
alisman: are these not already contained in this.filters?   it seems strange that we maintain some filter state outside of the filters.

src/pages/studyView/StudyViewPageStore.ts
kalletlak: its just a cleanup `defaultClient` and `client` points to same instance of `shared/api/cbioportalClientInstance`

src/pages/studyView/StudyViewPageStore.ts
kalletlak: `selectedSampleIdentifiersFilter` is a `studyViewFilter` with just `SampleIdentifiers` based on the `this.filters`.  sending `this.filters` for every api request would calculate filtered sample every time. instead if we pass just the `SampleIdentifiers` then this would save lot of computations and improve response time. Updated pr's description

alisman: I see.   Would it make any sense to rename selectedSamples to "filteredSamples" ?

alisman:
@kalletlak there is a study view screenshot failure that looks suspicious on this one.    also, the prettier test is failing.  strange that should happen.

kalletlak:
@alisman true, its because of my changes. Bar chart range varies depending on the filters. Since we are just passing filtered sample identifiers, chart range would be limited to just those filtered samples. I think i'll have to revoke it back for bar chart api's

alisman:
@kalletlak @zhx828 if you approve this PR, please merge, as I will on vacation.

zhx828:
The failing e2e tests are related to oncoprint. Not related to this pr.

PR #3056, opened 2020-02-12
Use IndexDB to cache post requests when in webdriver testing environment

adamabeshouse:
TODO:
- [ ] cache GET requests as well


alisman:
@adamabeshouse not clear if this helps, so gonna close it

PR #3057, opened 2020-02-12
do not clear selected study list when click query tab

dippindots:
Fix cBioPortal/cbioportal#6900

- a do not clear selected study list when click query tab
query_tab


dippindots:
@pvannierop @alisman This pr will fix issue: 6900, we don't need to clear the study list anymore because `download tab` is deprecated, so it's safe to just remove the clear action.

PR #3058, opened 2020-02-13
Add Software acknowledgment page

ecerami:
# Fixes cBioPortal/cbioportal#5667

- Added new software acknowledgment page/component.
- Added new property: ` skin_documentation_software`.
- Added new ` /software` router.
- Added link to `/software` within the footer.

> Note:  the actual content of the software page is a markdown page, which now exists within the back-end repo, and not part of this PR.  See:  https://github.com/cBioPortal/cbioportal/blob/master/docs/Software-Acknowledgments.md

# Any screenshots or GIFs?
- Not applicable

# Notify reviewers
@alisman @inodb or @jjgao

### DIFF
--- a/src/config/serverConfigDefaults.ts
+++ b/src/config/serverConfigDefaults.ts
@@ -44,6 +44,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_authorization_message:
         'Access to this portal is only available to authorized users.',
     skin_documentation_about: 'About-Us.md',
+    skin_documentation_software: 'Software-Acknowledgments.md',
     skin_documentation_baseurl:
         'https://raw.githubusercontent.com/cBioPortal/cbioportal/master/docs/',
     skin_documentation_markdown: true,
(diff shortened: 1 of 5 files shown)


src/config/serverConfigDefaults.ts
```diff
@@ -44,6 +44,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_authorization_message:
         'Access to this portal is only available to authorized users.',
     skin_documentation_about: 'About-Us.md',
+    skin_documentation_software: 'Software-Acknowledgments.md',
```
inodb: Nice US spelling of acknowledgment

FYI - If we want to allow people to update this config using portal.properties it will also need to be added to config_service.jsp in the backend. We already allow people to override the footer in its entirety, so seems overkill to add it there but just so you know

ecerami:
@inodb:  Good tip on imperative.  So, instead of "Added new route for software page" --> "Add new route for software page"?

Write back!  (I think that's imperative :-)

inodb:
Lol yes 🙂!

PR #3062, opened 2020-02-16
Add more descriptive GISTIC and MutSig Tooltips

ecerami:
# Fix cBioPortal/cbioportal#6752

- Added more descriptive tooltips for MutSig and GISTIC.
- The change applies to the patient view and the study view pages.

# Updates from 2/27/2020

- Based on code review from @onursumer, I made the MutSig and GISTIC tooltips into true React Components that use IconLetters.  I also refactored the study view and patient view pages to use these React Components v. the previous state where we had different logic and different code for each view.  The study view was also previously using GISTIC and MutSig PNG graphics, which I have now deleted.
- When I moved to the new IconLetters, three screentshot regression tests failed (as expected).  I have now updated these screenshots.
- With much help from @Luke-Sikina, I fixed my commit history and this PR now consists of just one squashed commit.

# Screencast from Patient View Page
Screencast

# Notify reviewers
@inodb @jjgao or @onursumer


src/shared/components/annotation/Gistic.tsx
```diff
@@ -0,0 +1,47 @@
...
+        </div>
+    );
+}
+
+/**
+ * Makes the GISTIC Icon
+ * @param qValue GISTIC Q-Value
+ * @param peakGeneCount Number of Genes within the GISTIC Peak
+ */
+export function makeGisticIcon(qValue: number, peakGeneCount: number) {
```
onursumer: How about `Gistic` as a proper react component instead of this function? 

It was okay to have this as a static function inside `CohortColumnFormatter`, but since we already moved the code out of the formatter, it is better to have a proper react component here. You can define `qValue` and `peakGeneCount` as `props` of the `Gistic` component.

src/shared/components/annotation/MutSig.tsx
```diff
@@ -0,0 +1,40 @@
...
+            <span> Q-value: {(qValue || 0).toExponential(3)}</span>
+        </div>
+    );
+}
+
+/**
+ * Makes the MutSig Icon
+ * @param qValue Makes the MutSig Icon with Tooltip
+ */
+export function makeMutSigIcon(qValue: number) {
```
onursumer: Similar to `Gistic`, let's have a proper react component here called `MutSig`.

Also contents of this file is very similar to `Gistic.tsx`. I would add some helper functions to help reducing code duplication.

src/shared/components/annotation/Gistic.tsx
ecerami: Thanks, Onur!  And, great suggestion.

Before I do so though, can you clarify one issue?

Right now, the patient view and study view pages actually use different mechanisms for displaying the (M) and (G) icons.  For patient view, we use the LetterIcon component and for the study view page, we use PNG icons, e.g. https://github.com/cBioPortal/cbioportal-frontend/tree/master/src/pages/studyView/table/images.

I didn't want to break the latter, thinking there was probably some good reason for this and I also didn't want to break all the study view screenshot tests.  Do you think it's ok to delete the second mechanism, with the PNG icons?   And, then I need to fix all the screenshot tests?

onursumer: @ecerami I think it is a good idea to replace the study view PNG icons with this one as well. We might have introduced the `LetterIcon` component after implementing the study view tables, or we might have simply missed the `LetterIcon` and added PNGs instead.

Right, you may need to fix the tests if we have the (G) and (M) icons in the reference screenshots.

ecerami: @onursumer I pushed a new commit where I created true React components for MutSig and GISTIC annotation.  I also deleted the PNG icons.  Take a look and let me know what you think.  If it looks good, I will go ahead and squash the 6 commits into one.

src/shared/components/annotation/MutSig.tsx
ecerami: @onursumer see message above.

ecerami:
Hi @inodb @alisman @jjgao @onursumer and @Luke-Sikina -- many thanks for your patient help this past week :-)  I updated this PR to include revised tooltip text from @tmazor and @schultzn.  I also updated the core code, based on feedback from @onursumer, fixed the e2e screenshot tests, and squashed everything into one commit.  All tests on Circle CI also now pass.

I *think* PR ready for final review / merge.  thanks!!

PR #3065, opened 2020-02-18
update to latest GN client

zhx828:


zhx828:
the client has been updated in https://github.com/cBioPortal/cbioportal-frontend/pull/2926

PR #3068, opened 2020-02-18
Study View: group comparison selection update

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7186

### DIFF
--- a/src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
+++ b/src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
@@ -272,7 +272,7 @@ export default class ComparisonGroupManager extends React.Component<
                         this.props.store.updateComparisonGroupsFilter();
                     }}
                 >
-                    View
+                    Filter
                 </button>
             );
         } else {


src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -272,7 +272,7 @@ export default class ComparisonGroupManager extends React.Component<
                         this.props.store.updateComparisonGroupsFilter();
                     }}
                 >
-                    View
+                    Select
```
adamabeshouse: probably should also rename to `selectButton` and look elswhere for `view` to replace it with `select`

kalletlak: @jjgao should I rename it to `selectButton`?
@adamabeshouse I haven't found any `view` in group comparison related pages.

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -272,7 +272,7 @@ export default class ComparisonGroupManager extends React.Component<
                         this.props.store.updateComparisonGroupsFilter();
                     }}
                 >
-                    View
+                    Filter
```
alisman: I thought it was "Select".   I can see "filter" as well, but maybe "Apply to filter" ???  @jjgao can you make an executive decision here?

alisman: And do change the comment above and the name of the getter accordingly

PR #3073, opened 2020-02-20
fix genomeBuild parsing

SRodenburg:
We noticed that the chromosome plot in the patient view was always displaying a human genome, regardless of the `portal.properties` settings. This implies it will never display mouse chromosomes when needed.

When any other reference genome build than 'GRCh37' is used, the genome is not parsed correctly. The frontend derives the genome build through the API, from the dataset itself, and translates the UCSC name into NCBI (GRC) format.
In case the genome is already in GRC format, it will fail and fallback to GRCh37. This is a problem when a non-default genome is used. 

This PR fixes that.

### DIFF
--- a/src/pages/patientView/genomicOverview/tracksHelper.ts
+++ b/src/pages/patientView/genomicOverview/tracksHelper.ts
@@ -91,6 +91,10 @@ export const genomeBuilds: Map<string, string> = new Map([
     ['hg38', 'GRCh38'],
     ['38', 'GRCh38'],
     ['mm10', 'GRCm38'],
+    // If the genome build is already Ensembl (e.g. VEP output), make sure the lookup doesn't fail
+    ['GRCh37', 'GRCh37'],
+    ['GRCh38', 'GRCh38'],
+    ['GRCm38', 'GRCm38'],
 ]);
 
 export type ChromosomeSizes = {
--- a/src/pages/patientView/genomicOverview/tracksHelper.ts
+++ b/src/pages/patientView/genomicOverview/tracksHelper.ts
@@ -91,7 +91,7 @@ export const genomeBuilds: Map<string, string> = new Map([
     ['hg38', 'GRCh38'],
     ['38', 'GRCh38'],
     ['mm10', 'GRCm38'],
-    // If the genome build is already Ensembl (e.g. VEP output), make sure the lookup doesn't fail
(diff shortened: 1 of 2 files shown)


alisman:
I noticed that we have two different constants in separate places:
export const DEFAULT_GENOME_BUILD = 'GRCh37';
export const DEFAULT_GENOME = 'hg19';

Should these be the same?  Also, since there is a map defined to normalize these values, perhaps we should make sure the constant is never referenced directly (it is referenced in several places now).   I.e. there should be a helper that normalizes the value.   Lets take this opportunity to address the mess.    Adding @inodb for input.

pvannierop:
I agree with @adamabeshouse that we have to make sure that the reference genome is evaluated in a constent manner throughout the code.

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -91,6 +91,10 @@ export const genomeBuilds: Map<string, string> = new Map([
     ['hg38', 'GRCh38'],
     ['38', 'GRCh38'],
     ['mm10', 'GRCm38'],
+    // If the genome build is already Ensembl (e.g. VEP output), make sure the lookup doesn't fail
```
inodb: There is an old comment above this variable:
```
/**
 * storing chromesome length info
 */
``` 
can be removed (it's about the variable below it, but the comment is redundant imo)

SRodenburg:
I totally agree it would be more useful to build an object that normalizes the genome build. I am not comfortable enough with the frontend and javascript in general to do this. Maybe @pvannierop can have a look what this would take?

alisman:
@adamabeshouse could you handle the refactoring described in the thread above in next sprint.  seems pretty quick

alisman:
I made the following issue to track the normalization issue https://github.com/cBioPortal/cbioportal/issues/7224

SRodenburg:
There was a missunderstanding. The **genome build** comes from the Genome Reference Consortium (e.g. 'GRCh37'), and the **reference genome** is the _name_ of the assembly (e.g. hg19).

It has nothing to do with Ensembl/NCBI as I described earlier.
So technically, the code was correct. Nonetheless, there is a mixed usage of the _build_ and the _name_  in the code to identify the used genome. That could still be improved.

For now, I suggest not to merge this PR.

PR #3076, opened 2020-02-20
On timeline generalize sample label replacement

inodb:
Instead of only allowing tracks with the name Specimen to have their
circles replaced with the corresponding sample label (the circle with
1/2/3), allow any track that confirms to the following:

If all events in a track contain a SAMPLE_ID in their tooltip table that
matches the patient's corresponding sample id.

Or to elaborate with screenshots:

generalize specimen track sample here:

Screen Shot 2020-02-20 at 5 13 29 PM

to a track with any name:

Screen Shot 2020-02-20 at 5 14 49 PM

Also:

- Don't add DECEASED status point for `mskimpact_test_june` study. Maybe we should change
this to be for all studies. Just afraid this might break things for people relying on this legacy logic

### DIFF
--- a/src/pages/patientView/timeline/legacy.js
+++ b/src/pages/patientView/timeline/legacy.js
@@ -197,53 +197,61 @@ export function buildTimeline(
         });
     }
 
-    // add DECEASED point to Status track using data from
-    // .*OS_MONTHS$ if .*OS_STATUS$ is DECEASED
-    var i;
-    var prefixes;
-    prefixes = Object.keys(patientInfo)
-        .filter(function(x) {
-            // find all keys postfixed with "OS_STATUS"
-            return /OS_STATUS$/.test(x);
-        })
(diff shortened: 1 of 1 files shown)


src/pages/patientView/timeline/legacy.js
```diff
@@ -195,53 +197,61 @@ export function buildTimeline(
...
-    var prefixes;
-    prefixes = Object.keys(patientInfo)
-        .filter(function(x) {
-            // find all keys postfixed with "OS_STATUS"
-            return /OS_STATUS$/.test(x);
-        })
-        .map(function(x) {
-            // get the prefixes
-            return x.substr(0, x.length - 'OS_STATUS'.length);
-        });
```
inodb: this is not removed but rather indented inside `isMskImpactJuneTestStudy`

PR #3077, opened 2020-02-20
Adding bitly option for filter session

dippindots:
Fix cBioPortal/cbioportal#7185

- a Adding bitly option for filter session (study view)
image

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -61,6 +61,7 @@ import './styles.scss';
 import autobind from 'autobind-decorator';
 import { BookmarkModal } from 'pages/resultsView/bookmark/BookmarkModal';
 import { ShareUrls } from 'pages/resultsView/querySummary/ShareUI';
+import request from 'superagent';
 
 export interface IStudyViewPageProps {
     routing: any;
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/querySummary/ShareUI.tsx
```diff
@@ -27,34 +31,38 @@ export interface ShareUrls {
 export class ShareUI extends React.Component<IShareUI, {}> {
     @observable showBookmarkDialog: boolean = false;
 
-    async getUrls(): Promise<ShareUrls> {
-        let sessionUrl = win.location.href;
+    readonly urls = remoteData<ShareUrls>({
```
dippindots: change to use remoteData

src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
```diff
@@ -48,42 +48,44 @@ export default class StudyPageHeader extends React.Component<
...
-                        removeComparisonGroupSelectionFilter={
-                            this.props.store
-                                .removeComparisonGroupSelectionFilter
-                        }
-                        onBookmarkClick={this.props.onBookmarkClick}
-                        clearAllFilters={this.props.store.clearAllFilters}
-                    />
-                )}
+                {this.props.store.clinicalAttributeIdToDataType.isComplete &&
+                    this.props.store.bookmarkUrl.isComplete && (
```
dippindots: Add this line to ensure we have bookmarkUrl data ready, then we render the bookmark button.

kalletlak: I think we should create bookmarkUrl only when the button is clicked. or it would create bitly url for each and every chart filter

dippindots: Thanks @kalletlak , good catch, I will try to update this.

src/pages/resultsView/bookmark/BookmarkModal.tsx
```diff
@@ -11,25 +11,15 @@ var Clipboard = require('clipboard');
 
 @observer
 export class BookmarkModal extends React.Component<
-    { onHide: () => void; urlPromise: Promise<any>; title: string },
+    { onHide: () => void; urlData: ShareUrls; title: string },
```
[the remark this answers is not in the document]
alisman: The reason why it needs to be a passed promise is that we want it to invoke (load) only when the Modal has opened (and the modal can then show load status).    However, the actual promise needs to be defined by external page specific code.  (see @kalletlak's comment below)

src/pages/studyView/studyPageHeader/StudyPageHeader.tsx
alisman: agreed

dippindots:
@alisman Sure, thanks for the clarification!

PR #3078, opened 2020-02-21
update react-table types

inodb:
Fix type error by updating react-table types

https://jitpack.io/com/github/cbioportal/cbioportal-frontend/v3.2.4/build.log
```
[ERROR] react-mutation-mapper: [!] (plugin rpt2) Error: /home/jitpack/build/packages/react-mutation-mapper/src/component/oncokb/OncoKbTreatmentTable.tsx(10,24): semantic error TS7016: Could not find a declaration file for module 'react-table'. '/home/jitpack/build/node_modules/react-table/lib/index.js' implicitly has an 'any' type.
```

### DIFF
--- a/packages/react-mutation-mapper/package.json
+++ b/packages/react-mutation-mapper/package.json
@@ -37,7 +37,6 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
-    "@types/react-table": "^6.8.7",
     "autobind-decorator": "^2.1.0",
     "cbioportal-frontend-commons": "^0.2.0",
     "classnames": "^2.2.5",
@@ -56,7 +55,6 @@
   },
   "devDependencies": {
     "@testing-library/react": "^9.3.2",
-    "@types/react-table": "^6.8.7",
     "react-scripts-ts": "^3.1.0"
   }
 }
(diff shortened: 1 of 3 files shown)


packages/react-mutation-mapper/package.json
```diff
@@ -37,6 +37,7 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "@types/react-table": "^6.8.7",
```
onursumer: It might be better to add this to the root package.json, since we already moved all `@types` dependencies there. Having `@types` in multiple packages is problematic.

packages/react-mutation-mapper/package.json
```diff
@@ -55,7 +56,7 @@
   },
   "devDependencies": {
     "@testing-library/react": "^9.3.2",
-    "@types/react-table": "^6.8.1",
+    "@types/react-table": "^6.8.7",
```
onursumer: I think we can also get rid of this if we move it to the root `package.json`

inodb:
Seems to fix it:

https://jitpack.io/com/github/inodb/cbioportal-frontend/0059653ddfe1336995ecdb247f065db6388c3b55/build.log

@onursumer I updated the location of the react-table types to the root directory. Does that look ok?

inodb:
flaky unit & e2e tests, so gonna merge

PR #3080, opened 2020-02-22
plots tab - store UI selections in URL parameters

adamabeshouse:


src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -67,9 +84,26 @@ export default class ResultsViewURLWrapper extends URLWrapper<
                 treatment_list: { isSessionProp: false },
                 exclude_germline_mutations: { isSessionProp: false },
                 patient_enrichments: { isSessionProp: false },
+
                 comparison_subtab: { isSessionProp: false },
                 comparison_overlapStrategy: { isSessionProp: false },
 
+                plots_selectedGeneOption_horz: { isSessionProp: false },
```
alisman: should we consider keeping these as one object that gets serialized to URL param?   we could probably even adapt urlWrapper to intercept change signaling in the same way as we do at the top level.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -689,13 +694,175 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
...
-            _selectedGeneOption: undefined,
-            _selectedGenesetOption: undefined,
-            _selectedTreatmentOption: undefined,
-            _dataType: undefined,
-            _mutationCountBy: undefined,
-            _logScale: false,
-            _selectedDataSourceOption: undefined,
+
+            get _selectedGeneOption() {
+                const param = vertical
```
alisman: where does vertical come from?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -689,13 +694,175 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
...
+                    const genesetOptions =
+                        (vertical
+                            ? self.vertGenesetOptions
+                            : self.horzGenesetOptions.result) || [];
+
+                    return genesetOptions.find(o => o.value === optionVal);
+                }
+            },
+            set _selectedGenesetOption(o: any) {
+                const newParams: Mutable<Partial<ResultsViewURLQuery>> = {};
```
alisman: What does Mutable do?

alisman:
I want to make sure this dovetails with work that @dippindots is doing to to make plots tab generic (handle any Generic assay profiles).

src/pages/resultsView/ResultsViewURLWrapper.ts
adamabeshouse: A couple of cons to that that are popping up:
- typescript won't be able to check the specific props within the serialized object
- the url will get even uglier

What makes you want to put them into one object?

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: this is within a private helper method that creates the axis selection object with the parameter `vertical:boolean`

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: for some reason Partial<ResultsViewURLQuery> was coming out with readonly props. Mutable just makes it not readonly

src/pages/resultsView/ResultsViewURLWrapper.ts
alisman: Reasons:

1. I can foresee a use case where we want to be able to clear these things as a group in some cases, for example when a query mod renders the configuration invalid. (We got lucky in that oncoprint can handle that circumstance gracefully.) 
2. The fact that prefix these props shows that they are indeed a group. Object is a nicer way of grouping.   If the only drawback is ugliness, i think that we should stop caring about ugliness.  The URL is so long that it's not really composable or readable by a human being anyway.  

Typing
I don't see why we can't type them  as a node of the parent object 
```
plots_config: { ... props, isSessionProp:bool }

File: src/pages/resultsView/ResultsViewURLWrapper.ts [hunk 4]
```diff
@@ -67,9 +96,20 @@ export default class ResultsViewURLWrapper extends URLWrapper<
                 generic_assay_groups: { isSessionProp: false },
                 exclude_germline_mutations: { isSessionProp: false },
                 patient_enrichments: { isSessionProp: false },
+
                 comparison_subtab: { isSessionProp: false },
                 comparison_overlapStrategy: { isSessionProp: false },
 
+                // plots
+                plots_horz_selection: {
+                    isSessionProp: false,
```
alisman: seems like this ought to be a sessionProp because the configuration of the plot is essential to shareability.

src/shared/lib/URLWrapper.ts
```diff
@@ -77,39 +75,28 @@ export default class URLWrapper<
...
+            initValues[property.name] = value as any;
         }
 
         // consumers of wrapper read from this query
-        this.query = observable<QueryParamsType>(initValues as QueryParamsType);
-
-        //per reaction below, any update to URL will result in all properties being reset.
-        //to avoid update signal when properties haven't actually changed (set to existing value)
-        //we intercept changes and cancel them when old === new
-        intercept(this.query, change => {
```
alisman: unless you somehow managed to accomplish this in another way, the intercept performs a crucial role.

PR #3081, opened 2020-02-22
Added Variant Type to Mutation Table

ecerami:
# Overview

This pull request adds variant type to the mutation table.  This fixes cBioPortal/cbioportal#3694.

- Added variant type column formatter
- Modified patient view and results page to include variant type
- Simplified some of the existing mutation type column formatter code to remove legacy attributes that are no longer used.

## Checks
- Added a unit test for the new variant type column formatter.


ecerami:
This seems to have pulled in changes from another branch.  Closing.

PR #3082, opened 2020-02-23
Add variant type column to mutation tables

ecerami:
# Overview 

Added variant type to the mutation tables.  This PR fixes cBioPortal/cbioportal#3694.

- Added a new variant type column formatter and associated unit test.
- Refactored the existing mutation type column formatter to remove legacy attributes that are no longer used (based on discussion with @onursumer).

## Checks
- [X] Yes, I have included a unit test for the variant type column formatter.
- [X] Yes, I have just one commit (thanks to @onursumer for that feedback).

## Screenshot
screencast

## Notify reviewers
@onursumer @inodb

### DIFF
--- a/src/shared/components/mutationTable/MutationTable.tsx
+++ b/src/shared/components/mutationTable/MutationTable.tsx
@@ -23,6 +23,7 @@ import GeneColumnFormatter from './column/GeneColumnFormatter';
 import ChromosomeColumnFormatter from './column/ChromosomeColumnFormatter';
 import ProteinChangeColumnFormatter from './column/ProteinChangeColumnFormatter';
 import MutationTypeColumnFormatter from './column/MutationTypeColumnFormatter';
+import VariantTypeColumnFormatter from './column/VariantTypeColumnFormatter';
 import FunctionalImpactColumnFormatter from './column/FunctionalImpactColumnFormatter';
 import CosmicColumnFormatter from './column/CosmicColumnFormatter';
 import MutationCountColumnFormatter from './column/MutationCountColumnFormatter';
@@ -135,6 +136,7 @@ export enum MutationTableColumnType {
     MUTATION_STATUS,
     VALIDATION_STATUS,
     MUTATION_TYPE,
+    VARIANT_TYPE,
     CENTER,
     TUMOR_ALLELE_FREQ,
     NORMAL_ALLELE_FREQ,
@@ -201,9 +203,7 @@ export function defaultFilter(
         return data.reduce((match: boolean, next: Mutation) => {
             const val = (next as any)[dataField];
             if (val) {
-                return (
-                    match || val.toUpperCase().indexOf(filterStringUpper) > -1
-                );
+                return match || val.toUpperCase().includes(filterStringUpper);
             } else {
                 return match;
             }
@@ -607,7 +607,7 @@ export default class MutationTable<
             ) =>
                 GeneColumnFormatter.getTextValue(d)
                     .toUpperCase()
-                    .indexOf(filterStringUpper) > -1,
+                    .includes(filterStringUpper),
         };
 
         this._columns[MutationTableColumnType.CHROMOSOME] = {
@@ -628,7 +628,7 @@ export default class MutationTable<
             ) =>
                 (ChromosomeColumnFormatter.getData(d) + '')
                     .toUpperCase()
-                    .indexOf(filterStringUpper) > -1,
+                    .includes(filterStringUpper),
             visible: false,
             align: 'right',
         };
@@ -655,7 +655,24 @@ export default class MutationTable<
             ) =>
                 MutationTypeColumnFormatter.getDisplayValue(d)
                     .toUpperCase()
-                    .indexOf(filterStringUpper) > -1,
+                    .includes(filterStringUpper),
+        };
+
+        this._columns[MutationTableColumnType.VARIANT_TYPE] = {
+            name: 'Variant Type',
+            render: VariantTypeColumnFormatter.renderFunction,
+            download: VariantTypeColumnFormatter.getTextValue,
+            sortBy: (d: Mutation[]) =>
+                VariantTypeColumnFormatter.getDisplayValue(d),
+            filter: (
+                d: Mutation[],
+                filterString: string,
+                filterStringUpper: string
+            ) =>
+                VariantTypeColumnFormatter.getDisplayValue(d)
+                    .toUpperCase()
+                    .includes(filterStringUpper),
+            visible: false,
         };
 
         this._columns[MutationTableColumnType.FUNCTIONAL_IMPACT] = {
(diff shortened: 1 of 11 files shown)


ecerami:
Oops @inodb I guess I didn't use the imperative on the commit! :-)

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -658,6 +660,23 @@ export default class MutationTable<
...
+            render: VariantTypeColumnFormatter.renderFunction,
+            download: VariantTypeColumnFormatter.getTextValue,
+            sortBy: (d: Mutation[]) =>
+                VariantTypeColumnFormatter.getDisplayValue(d),
+            filter: (
+                d: Mutation[],
+                filterString: string,
+                filterStringUpper: string
+            ) =>
+                VariantTypeColumnFormatter.getDisplayValue(d)
```
alisman: instead of indexOf, use  .includes() which returns boolean.

src/shared/components/mutationTable/column/CategoricalColumnFormatter.tsx
```diff
@@ -0,0 +1,27 @@
...
+    toolTip?: string;
+    className: string;
+}
+
+export default class CategoricalColumnFormatter {
+    /**
+     * Boilerplate code to create tool tip.
+     */
+    public static createToolTip(content: JSX.Element, toolTip: string) {
+        const arrowContent = <div className="rc-tooltip-arrow-inner" />;
```
alisman: i think this is actually the default arrowContent for DefaultToolTip, so maybe no necessary

src/shared/components/mutationTable/column/VariantTypeColumnFormatter.tsx
```diff
@@ -0,0 +1,152 @@
...
+ * SNP, DNP, TNP, ONP, INS, DEL, or Consolidated.
+ */
+export default class VariantTypeColumnFormatter {
+    public static get MAIN_VARIANT_TYPE_MAP(): {
+        [key: string]: ICategoricalColumn;
+    } {
+        return {
+            SNP: {
+                toolTip: 'Single Nucleotide Polymorphism',
+                className: 'variant-type-snp',
```
alisman: no big deal but you could probably reference the style on styles import itself right here (and not do the lookup later).    then you would get autocomplete on these names in your IDE, assuming it can handle css modules (WebStorm can).       I don't think dashes in class names would work anymore so you'd have to change those.     
so, i.e.,  { ..., className: styles.variantTypeSNP } 
the goal being to avoid repeated and unverifiable strings.

src/shared/components/mutationTable/column/VariantTypeColumnFormatter.tsx
```diff
@@ -0,0 +1,152 @@
...
+    public static renderFunction(data: Mutation[]) {
+        // use text for all purposes (display, sort, filter)
+        const text: string = VariantTypeColumnFormatter.getDisplayValue(data);
+        const className: string = VariantTypeColumnFormatter.getClassName(data);
+
+        // extract the tooltip, if available
+        const toolTip = VariantTypeColumnFormatter.getTooltip(data);
+        let content = <span className={styles[className]}>{text}</span>;
+
+        if (toolTip != undefined) {
```
alisman: you can just do 
`if (toolTip)`
as an aside, we should always use !==  so that we never worry about type coercion.  double equals means no type coercion.    so, just always use === and !== in conditions

ecerami: thanks.  will do.

src/shared/components/mutationTable/column/VariantTypeColumnFormatter.tsx
ecerami: thanks @alisman.  I am not familiar with this.  Can you point me to some other existing code where I can see how this works?

src/shared/components/mutationTable/MutationTable.tsx
ecerami: @alisman general question:  do people usually address code review feedback in a new commit?  and, after the reviewer has re-reviewed, I then squash all the commits into one?

src/shared/components/mutationTable/column/CategoricalColumnFormatter.tsx
```diff
@@ -0,0 +1,27 @@
...
+    displayValue?: string;
+    toolTip?: string;
+    className: string;
+}
+
+export default class CategoricalColumnFormatter {
+    /**
+     * Boilerplate code to create tool tip.
+     */
+    public static createToolTip(content: JSX.Element, toolTip: string) {
```
onursumer: We should probably make this a simple function and get rid of the `CategoricalColumnFormatter` class

```suggestion
    export function createToolTip(content: JSX.Element, toolTip: string) {
```

src/shared/components/mutationTable/MutationTable.tsx
ecerami: ok, found 6 instances of `indexOf()` within `MutationTable.tsx`.  Changed all of these to `includes()`.

src/shared/components/mutationTable/column/CategoricalColumnFormatter.tsx
ecerami: yes, you are right, that this is default `arrowContent`, and this code is not necessary.  Have now removed.

src/shared/components/mutationTable/column/VariantTypeColumnFormatter.tsx
ecerami: Now fixed.

src/shared/components/mutationTable/column/CategoricalColumnFormatter.tsx
ecerami: thanks, @onursumer.  good idea.  this is now done.

src/shared/components/mutationTable/column/VariantTypeColumnFormatter.tsx
ecerami: ah, never mind.  I see what you mean now.  Clever idea!  Will definitely change.

ecerami: this is now done.

ecerami:
@alisman @onursumer thanks for the code review.  I was able to complete all your suggestions.  Everything is passing, and I have squashed down to just one commit.  Ready for final review / merge.  Thanks!

leexgh:
@ecerami Can I ask how do you design the color? In this PR the `Fusion` is blue(#0000bb), and `Other` is set to purple (#8b00c9), but it seems like this purple color #8b00c9 is using for Fusion in [other places](https://github.com/cBioPortal/cbioportal-frontend/blob/594ae37b129bf727899ac371b82869f0986241d6/src/shared/lib/Colors.ts#L20)

ecerami:
@leexgh good question.  It's been a while, but I tried to make the colors consistent.  But, could be the case that I missed a few types, and the colors could be updated.

leexgh:
@ecerami I see, well that happened. I just want to know if the colors are particularly designed to be so, and since they are not, I can create a PR to make the colors consistent

ecerami:
great.  thank you @leexgh!

PR #3083, opened 2020-02-24
Migrate to bitly api v4

alisman:
Bitly has released v4 of their api and is no longer supporting <v4 as of March 1.  This uses new version.


alisman:
@dippindots after this merges, you should update your PR as necessary

PR #3087, opened 2020-02-27
Shorten the alteration text when there are more then 5 alterations associated with the treatment

zhx828:
Fixes https://github.com/cbioportal/cbioportal/issues/7242


onursumer:
@zhx828 should we bump `react-mutation-mapper` version?

zhx828:
> @zhx828 should we bump `react-mutation-mapper` version?

yeah right... do that now

PR #3090, opened 2020-02-27
Study View: New genomic profile samples count table

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6771
Fix https://github.com/cBioPortal/cbioportal/issues/3836

Screen Shot 2020-03-20 at 4 47 32 PM


src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -254,29 +238,35 @@ export class StudySummaryTab extends React.Component<
...
+                );
+                props.getData = () => this.store.getGenesCNADownloadData();
                 props.genePanelCache = this.store.genePanelCache;
                 props.downloadTypes = ['Data'];
                 props.filterByCancerGenes = this.store.filterCNAGenesTableByCancerGenes;
                 props.onChangeCancerGeneFilter = this.store.updateCNAGenesTableByCancerGenesFilter;
                 break;
             }
-            case SURVIVAL: {
+            case ChartTypeEnum.GENOMIC_PROFILES_TABLE: {
```
zhx828: No title for this table?

src/pages/studyView/table/MolecularProfileCountsTable.tsx
```diff
@@ -0,0 +1,401 @@
...
+
+type GeneTableColumn = {
+    columnKey: TableColumnKey;
+    columnWidthRatio?: number;
+};
+
+class GeneTableComponent extends FixedHeaderTable<GenomicDataCount> {}
+
+@observer
+export class MolecularProfileCountsTable extends React.Component<
```
zhx828: Why not use `GeneTable`?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+    let count = 0;
+    if (molecularPorfileSampleCountSet['cna'] !== undefined) {
+        count = molecularPorfileSampleCountSet['cna'].count;
+    } else if (molecularPorfileSampleCountSet['gistic'] !== undefined) {
+        count = molecularPorfileSampleCountSet['gistic'].count;
+    } else if (molecularPorfileSampleCountSet['rae'] !== undefined) {
+        count = molecularPorfileSampleCountSet['rae'].count;
+    } else if (molecularPorfileSampleCountSet['cna_consensus'] !== undefined) {
+        count = molecularPorfileSampleCountSet['cna_consensus'].count;
+    }
```
zhx828: Should we do a for loop instead?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5573,4 +5438,38 @@ export class StudyViewPageStore {
...
+                }
+                case ChartTypeEnum.CNA_GENES_TABLE: {
+                    count = this.molecularProfileSampleCountSet.result['cna']
+                        ? this.molecularProfileSampleCountSet.result['cna']
+                              .count
+                        : 0;
+                    break;
+                }
+            }
+            return `${title} (${count} profiled samples)`;
```
zhx828: Maybe to worth to do a pluralize on the `samples`

zhx828:
We also need to set up the changes somewhere to test the feature.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2430,20 +2426,26 @@ export class ResultsViewPageStore {
         const scatterRequiredParams = _.reduce(
             this.clinicalAttributes.result,
             (acc, next) => {
-                if (MUTATION_COUNT === next.clinicalAttributeId) {
-                    acc[MUTATION_COUNT] = true;
+                if (UniqueKey.MUTATION_COUNT === next.clinicalAttributeId) {
```
alisman: would be nice if this enum had a more descriptive name.  And "Enum" suffix.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3345,6 +3305,28 @@ export class StudyViewPageStore {
     updateChartStats() {
         this.initializeChartStatsByClinicalAttributes();
 
+        if (!_.isEmpty(this.molecularProfiles.result)) {
```
alisman: should updageChartsStats be an action?  I mean, maybe it's called only be things which are actions themselves but seems like it makes a lot of state updates so maybe just tag it as action too?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3829,13 +3823,83 @@ export class StudyViewPageStore {
...
+                profiles =>
+                    _.flatMap(profiles, profile =>
+                        molecularProfileSampleCountSet[profile]
+                            ? molecularProfileSampleCountSet[profile]
+                                  .sampleUniqueKeys
+                            : []
+                    )
+            );
+
+            const sampleSetBykey = this.sampleSetBykey.result || {};
```
alisman: you should add this to away for good measure

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3829,13 +3823,83 @@ export class StudyViewPageStore {
...
+                            sampleId: sample.sampleId,
+                            studyId: sample.studyId,
+                        });
+                    }
+                    return acc;
+                },
+                [] as SampleIdentifier[]
+            );
+        },
+        onError: error => {},
```
alisman: i don't think you want to define onError

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3829,13 +3823,83 @@ export class StudyViewPageStore {
...
     readonly selectedSamples = remoteData<Sample[]>({
-        await: () => [this.samples],
+        await: () => [
+            this.samples,
+            this.sampleIdentifiersFromGenomicProfileFilter,
+        ],
         invoke: () => {
             //fetch samples when there are only filters applied
             if (this.chartsAreFiltered) {
+                const studyViewFilter = _.cloneDeep(toJS(this.filters));
```
alisman: i'm not sure but i think toJS already clones, no?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3829,13 +3823,83 @@ export class StudyViewPageStore {
...
+                                return (
+                                    a.sampleId === b.sampleId &&
+                                    a.studyId === b.studyId
+                                );
+                            }) as any
+                        );
+                        studyViewFilter.sampleIdentifiers = sampleIdentifiers;
+                    } else {
+                        studyViewFilter.sampleIdentifiers = sampleIdentifiersFromGenomicProfileFilter;
+                        delete studyViewFilter.studyIds;
```
alisman: maybe a comment about why we are deleting these

src/pages/studyView/StudyViewUtils.tsx
kalletlak: may be not. I think order is important here

src/pages/studyView/tabs/SummaryTab.tsx
kalletlak: There is a default title for it `'Genomic Profile Sample Counts'` at https://github.com/cBioPortal/cbioportal-frontend/pull/3090/files#diff-48cb532a65b28af956576b588b2ce39fR181

src/pages/studyView/StudyViewPageStore.ts
kalletlak: there is already action annotation to it

kalletlak:
> We also need to set up the changes somewhere to test the feature.

@zhx828 @jjgao @alisman here is there test instance
https://cbioportal-fix-6771-xoffqfv7y6.herokuapp.com/ and setting local storage
`localStorage.setItem("netlify","deploy-preview-3090--cbioportalfrontend")`

zhx828:
Thanks @kalletlak for setting it up. I will keep updating this thread when I see issues.
- [ ]  For this study, https://cbioportal-fix-6771-xoffqfv7y6.herokuapp.com/study/summary?id=acc_2019. It's a bit confusing when the GPSC table shows 0 fusions but there is a fusion table under with some data in it.

jjgao:
Thanks, @kalletlak & @zhx828 !

@tmazor I think it's ready for you to test :)

One question: I am wonder if we should just do intersection instead of union (and therefore no need for "Select samples" button). Alternatively, we could provide intersection and union, eg. as a drop down of the "Select samples" button. @tmazor do you think there are any use cases to select UNION (instead of INTERSECTION) of samples in profiles?

image

For @zhx828's 2nd point, it's hard to address. Different profiles may not be comparable. Also there may be multiple profiles per type per study.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+): GenomicDataCountWithSampleUniqueKeys[] {
+    return _.chain(molecularProfiles)
+        .groupBy(molecularProfile =>
+            molecularProfile.molecularProfileId.replace(
+                molecularProfile.studyId + '_',
+                ''
+            )
+        )
+        .map((profiles, value) => {
+            const uniqueProfiledsamples = _.chain(profiles)
```
alisman: camel case samples

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+            };
+        })
+        .value();
+}
+
+export function getCNASamplesCount(molecularPorfileSampleCountSet: {
+    [id: string]: number;
+}) {
+    let count = 0;
+    if (molecularPorfileSampleCountSet['cna'] !== undefined) {
```
alisman: where do these keys come from?  can we make enum for them?  they are sensitive because would result in wrong counts.   also we could do this tallying programmatically

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+        count = molecularPorfileSampleCountSet['gistic'];
+    } else if (molecularPorfileSampleCountSet['rae'] !== undefined) {
+        count = molecularPorfileSampleCountSet['rae'];
+    } else if (molecularPorfileSampleCountSet['cna_consensus'] !== undefined) {
+        count = molecularPorfileSampleCountSet['cna_consensus'];
+    }
+    return count;
+}
+
+export function getMolecularPorfileSamplesSet(
```
alisman: typo (porfile)

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+    samples: Sample[],
+    genePanelData: GenePanelData[]
+) {
+    const selectedUniqueSampleKeySet = _.keyBy(
```
alisman: seems like from perspective of this function, there is no "selection."  it's just whatever samples are passed, no?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2512,3 +2496,76 @@ export function updateSavedUserPreferenceChartIds(
...
+                selectedUniqueSampleKeySet[next.uniqueSampleKey] !== undefined
+            ) {
+                if (acc[next.molecularProfileId] === undefined) {
+                    acc[next.molecularProfileId] = [];
+                }
+                acc[next.molecularProfileId].push(next.uniqueSampleKey);
+            }
+            return acc;
+        },
+        {} as { [id: string]: string[] }
```
alisman: i think it's better to put type on the argument (acc).   then it will validate what is passed here instead of casting it.  doesn't much matter here because object is empty but sometimes we pass something with empty properties.  
``` _.reduce((acc:type)=>{...},{ collectionA:[], collectionB:[]  })```.

File: src/pages/studyView/UserSelections.tsx [hunk 16]
```diff
@@ -384,6 +386,50 @@ export default class UserSelections extends React.Component<
             components
         );
 
+        if (!_.isEmpty(this.props.filter.genomicProfiles)) {
```
alisman: this is a little too deeply nested

src/pages/studyView/UserSelections.tsx
```diff
@@ -384,6 +386,50 @@ export default class UserSelections extends React.Component<
...
+                                                                profile
+                                                            ] || profile
+                                                        }
+                                                        backgroundColor={
+                                                            STUDY_VIEW_CONFIG
+                                                                .colors.theme
+                                                                .clinicalFilterContent
+                                                        }
+                                                        onDelete={() =>
+                                                            this.props.removeGenomicProfileFilter(
```
alisman: put onDelete handler in bound function to avoid vDOM diffing inefficiency

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -349,18 +352,8 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
-            CNA_GENES_TABLE,
-            TABLE,
-            SCATTER,
-            PIE_CHART,
-            MUTATED_GENES_TABLE,
-            FUSION_GENES_TABLE,
-        } = ChartTypeEnum;
         switch (this.chartType) {
-            case PIE_CHART: {
+            case ChartTypeEnum.PIE_CHART: {
```
alisman: this switch statement is getting pretty huge.   we should instead have a map.   
```
const chartMap = {};
chartMap[enum.PIE_CHART] = ()=><PieChart>
```
then the chart method can just look it up.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1496,6 +1448,26 @@ export class StudyViewPageStore {
         this._geneFilterSet.delete(chartUniqueKey);
     }
 
+    @autobind
+    @action
+    removeGenomicProfileFilter(toBeRemoved: string) {
+        let genomicProfilesFilter = toJS(this.genomicProfilesFilter) || [];
```
alisman: toJS will return falsy if passed something false?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3776,6 +3806,15 @@ export class StudyViewPageStore {
         },
     });
 
+    public readonly sampleSetBykey = remoteData({
```
alisman: type: camel case

tmazor:
This is a great new chart!

A few initial thoughts:
- I do think there is benefit to allowing union. For example, in gbm_tcga there are 3 different mRNA profiles (microarray, U133 microarray and RNA-seq), and I can imagine wanting to select all samples with any expression profile.
- For msk-impact study, the SV profile sample count seems wrong - don't all samples have SV called?
image
Is this an accurate reflection of the data? Is this the profile count for the new data format for the new fusion tab? Is that only provided for now in a subset of samples? 
- I think the cancer studies pie chart is no longer loading by default for combined studies https://www.cbioportal.org/study/summary?id=msk_impact_2017%2Cgbm_tcga%2Clgggbm_tcga_pub%2Clgg_ucsf_2014
- I realize this is a data issue, but I think we need a plan to standardize profile names. In combined studies there can be multiple profiles for the same datatype (row 2 & 3 below are both copy number calls). This is also another reason to let users select the union of multiple profiles.
image

I'll do some more testing tomorrow, but it's looking good!

kalletlak:
@tmazor thanks for testing it out.

Regarding

2nd point: Yes, it exactly reflects the data. Looks like only 1645 samples have structural variant data

3rd point: Looks like Cancer Studies have less priority than other chart. Here is the list https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/studyView/StudyViewConfig.ts#L103. Does this needs to be changed?

cc: @jjgao @zhx828

src/pages/studyView/StudyViewPageStore.ts
kalletlak: toJS would return `undefined` if `this.genomicProfilesFilter` is `undefined`

src/pages/studyView/charts/ChartContainer.tsx
kalletlak: let's do the refactoring in another pr

tmazor:
@kalletlak  regarding point 3, if you load that combined study in production, the Cancer Studies chart loads. But if I load the same combined study in this PR, then it doesn't. I'm not sure how that relates to the priority, but it should always load for a combined study.

kalletlak:
@tmazor Fixed Cancer Studies chart. It is now same as in production.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2614,6 +2562,39 @@ export class StudyViewPageStore {
...
+                    );
+                }
+            );
+
+            const genePanelData = await defaultClient.fetchGenePanelDataInMultipleMolecularProfilesUsingPOST(
+                {
+                    sampleMolecularIdentifiers: sampleMolecularIdentifiers,
+                }
+            );
+            return genePanelData.filter(datum => datum.profiled);
```
alisman: should the fact that we are only returning profiled data be reflected in the name of this promise?  should be have two promises, one with raw data, one with filtered?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3825,13 +3851,84 @@ export class StudyViewPageStore {
         default: [],
     });
 
+    private readonly sampleIdentifiersFromGenomicProfileFilter = remoteData({
+        await: () => [this.samples, this.initialMolecularProfileSampleCounts],
+        invoke: async () => {
+            const molecularProfileSampleCountSet = _.keyBy(
+                this.initialMolecularProfileSampleCounts.result || [],
```
alisman: is there a reason initialMolecularProfileSampleCounts.result would falsy?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3825,13 +3851,84 @@ export class StudyViewPageStore {
...
+                profiles =>
+                    _.flatMap(profiles, profile =>
+                        molecularProfileSampleCountSet[profile]
+                            ? molecularProfileSampleCountSet[profile]
+                                  .sampleUniqueKeys
+                            : []
+                    )
+            );
+
+            const sampleSetBykey = this.sampleSetByKey.result || {};
```
alisman: key should be camel

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4121,6 +4221,25 @@ export class StudyViewPageStore {
...
+                        studyId: sample.studyId,
+                    })),
+                    projection: 'META',
+                })
+                .then(function(response: request.Response) {
+                    const count = parseInt(response.header['total-count'], 10);
+                    return count > 0;
+                });
+        },
+        default: false,
```
alisman: maybe no default value?

src/pages/studyView/StudyViewPageStore.ts
kalletlak: We just need profiled data. I'll rename the promise

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5569,4 +5503,41 @@ export class StudyViewPageStore {
         onError: error => Promise.resolve([]),
         default: [],
     });
+
+    public getChartTitle(chartType: ChartTypeEnum, title?: string) {
```
alisman: i think this needs a good comment to explain why the molecularProfileSampleCountSet result impacts the title

src/pages/studyView/StudyViewPageStore.ts
kalletlak: no

src/pages/studyView/TableUtils.tsx
```diff
@@ -170,3 +168,17 @@ export function getFreqColumnRender(
         </DefaultTooltip>
     );
 }
+
+export function getTooltip(type: FreqColumnType, isPergentage: boolean) {
+    let tooltipContent = `${isPergentage ? 'Percentage' : 'Number'} of samples`;
+    if (type !== 'data') {
+        const detailedTypeInfo =
+            type === 'mutation'
```
alisman: in future please avoid nested ternary.  this should be switch

tmazor:
@kalletlak looks good to me!

schultzn:
This looks great @kalletlak 

I agree with Tali's comment about unifying profile names so that they show up nicely in virtual cohorts. If that is too hard, that would be another reason to offer users and AND an OR option (intersection vs union) option when selecting.

kalletlak:
@schultzn @tmazor could you product review the new selection option(whether its intersection or union) for multi-row selection tables

schultzn:
Looks great. Instead of the mathematical symbols for union and intersection, could we consider showing a two way Venn diagram, in which we indicate by color which set of samples are included in a selection? I.e., everything vs just the intersection.

kalletlak:
@alisman JJ and Niki decided to go with
image (2)

inodb:
Merging this - backend localdb test are passing with this PR

PR #3093, opened 2020-03-04
Consolidate bitly code

adamabeshouse:


src/pages/studyView/StudyViewPage.tsx
```diff
@@ -221,35 +222,18 @@ export default class StudyViewPage extends React.Component<
     }
 
     async getBookmarkUrl(): Promise<ShareUrls> {
-        let bitlyResponse;
-
-        // now lets shorten with bityly, if we have key
-        // WE ARE DISABLING BITLY PENDING DISCUSSION
         console.log(AppConfig.serverConfig.bitly_access_token);
```
alisman: thus is probably my console, but lets kill it.  sorry.

adamabeshouse: sounds good

src/shared/lib/redirectHelpers.ts
```diff
@@ -168,3 +169,11 @@ export function handleIndexDO() {
         );
     }
 }
+
+export function handleEncodedRedirect() {
+    const encodedURL = getBrowserWindow().routingStore.location.query[
+        EncodedURLParam
+    ];
+    const decodedURL = atob(encodedURL);
+    getBrowserWindow().location.href = decodedURL;
```
alisman: using the routingStore updateROute would allow us to replace history entry that contained the encoded thing.  with your solution, user could back button to it and then get redirected again, when they really might want to return to wherever they clicked the original bitly url.

PR #3096, opened 2020-03-05
Study View: Reset chart filters when hidden from Charts dropdown

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7254


alisman:
Out of curiousity, why does this fix the issue?

kalletlak:
> Out of curiousity, why does this fix the issue?

https://github.com/cBioPortal/cbioportal-frontend/pull/3096/files#diff-6266952c7613f639590296d54621e13cR248
Instead of just changing chart visibility we even have to reset chart filters

PR #3099, opened 2020-03-06
Support other types of survival plots

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/7019

Support Overall Survival, Progression-free Survival, Disease-free Survival, Disease-specific Survival, Event-free survival

Study view:
https://deploy-preview-3099--cbioportalfrontend.netlify.com/study/summary?id=brca_tcga_pan_can_atlas_2018
image

Result view:
https://deploy-preview-3099--cbioportalfrontend.netlify.com/results/comparison?Action=Submit&cancer_study_list=brca_tcga_pan_can_atlas_2018&case_set_id=brca_tcga_pan_can_atlas_2018_all&gene_list=TP53&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=brca_tcga_pan_can_atlas_2018_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=brca_tcga_pan_can_atlas_2018_mutations&tab_index=tab_visualize&comparison_subtab=survival
image

Comparison page:
https://deploy-preview-3099--cbioportalfrontend.netlify.com/comparison/survival?sessionId=5d07fa78e4b0ab4137875ff2
image


src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -36,6 +37,36 @@ export type SurvivalPlotFilters = {
...
+export const survivalPlotTooltipxLabelWithEvent: {
+    [prefix: string]: string;
+} = {
+    // TODO text for PFS DSS EFS?
+    OS: 'Time of death',
+    PFS: 'Time of death',
+    DFS: 'Time of Relapse',
+    DSS: 'Time of death',
+    EFS: 'Time of death',
+};
```
leexgh: @jjgao what text would be good for PFS, DSS and EFS? this text will show in the tooltip as an explanation of x label.
image

src/pages/groupComparison/styles.module.scss
```diff
@@ -97,3 +97,43 @@
         margin-right: 30px;
     }
 }
+.survivalPlotHeaderContainer {
+    display: flex;
+    margin-right: 5px;
+}
+
+.survivalPlotHeader {
```
alisman: not certain but i think you can use existing pill styles from bootstrap.  check plots tab implementation

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -36,6 +37,36 @@ export type SurvivalPlotFilters = {
...
+
+// TODO should remove this if we want to be generic
+export const survivalPlotTooltipxLabelWithEvent: {
+    [prefix: string]: string;
+} = {
+    // TODO text for PFS DSS EFS?
+    OS: 'Time of death',
+    PFS: 'Time of death',
+    DFS: 'Time of Relapse',
```
alisman: relapse should not be capitalized.  Alternatively, "death" should be capitalized

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -370,3 +401,22 @@ export function getSurvivalChartDataByAlteredStatus(
...
     };
 }
+
+export function generateSurvivalPlotTitleFromDisplayName(title: string) {
+    return title.replace(/status|survival/gi, '');
+}
+
+export function generateStudyViewSurvivalPlotTitle(title: string) {
+    let result = title.replace(/status/gi, '');
+    return result.toLowerCase().includes('survival')
```
alisman: /survival/i.test(result)
instead of includes when case isensitive.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -370,3 +401,22 @@ export function getSurvivalChartDataByAlteredStatus(
...
+        ? result
+        : `${result} Survival`;
+}
+
+export function getSurvivalAttributes(clinicalAttribute: ClinicalAttribute[]) {
```
alisman: clinicalAttribute should be plural: clinicalAttributes

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -370,3 +401,22 @@ export function getSurvivalChartDataByAlteredStatus(
...
+    return _.chain(clinicalAttribute)
+        .map(attr => attr.clinicalAttributeId)
+        .filter(id => id.endsWith('_STATUS') || id.endsWith('_MONTHS'))
```
alisman: you should use case insisitive regexpression.  /_STATUS$/i.test(...) instead of endsWith

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3466,43 +3467,24 @@ export class StudyViewPageStore {
...
-            dfsMonthsFlag &&
-            getDefaultPriorityByUniqueKey(UniqueKey.DISEASE_FREE_SURVIVAL) !== 0
-        ) {
-            // hide DISEASE_FREE_SURVIVAL chart if cancer type is mixed or have more than one cancer type
-            if (cancerTypeIds.length === 1 && cancerTypeIds[0] !== 'mixed') {
-                this.changeChartVisibility(
-                    UniqueKey.DISEASE_FREE_SURVIVAL,
-                    true
-                );
+        survivalUniqueKeys.map(key => {
```
alisman: this should probably be forEach.  only use map if you are doing something with result of mapping operation

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5569,4 +5534,73 @@ export class StudyViewPageStore {
...
+            const attributes = getSurvivalAttributes(
+                this.clinicalAttributes.result!
+            );
+            let pairedAttributes: {
+                [prefix: string]: { status: string; months: string };
+            } = {};
+            let attributesPrefix: string[] = [];
+            // get paired attributes
+            attributes.map(attribute => {
+                let prefix = attribute.substring(
```
alisman: you should use regexp and string.match for this.  you get case insensitivity and more flexibility

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1181,71 +1230,44 @@ export default class ComparisonStore {
...
-                clinicalAttributeMap[
-                    diseaseFreeSurvivalClinicalAttributeId
-                ].forEach(attr => {
-                    result.push({
-                        studyName: studyIdToStudy[attr.studyId].name,
-                        description: attr.description,
-                    } as ISurvivalDescription);
+            const result: { [prefix: string]: ISurvivalDescription[] } = {};
+
+            if (this.survivalClinicalAttributesPrefix) {
```
alisman: you don't have to test for existence of this.  it's the result which could be undefined.

src/pages/studyView/StudyViewPageStore.ts
leexgh: @alisman Basically this part is to get the attributes' prefix. For example get `OS` from `OS_STATUS`, not sure how can I do this with string.match. I think we can do string.match for `attribute.indexOf('_STATUS')` below, but it doesn't make much sense since all attributes name are in uppercase

src/pages/groupComparison/styles.module.scss
leexgh: will fix in another pr

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -370,3 +401,20 @@ export function getSurvivalChartDataByAlteredStatus(
         patientToAnalysisGroups,
     };
 }
+
+export function generateSurvivalPlotTitleFromDisplayName(title: string) {
+    return title.replace(/status|survival/gi, '');
```
alisman: just want to make sure you know that the "g" modifier makes this global.  so it, will replace all instances of that word in the title.    it seems unlikely that there would be more than one.    you don't have to change this, i just wanted to make sure

src/pages/groupComparison/styles.module.scss
leexgh: I'll create a new pr after merging. Have some code changes based on this pr https://github.com/leexgh/cbioportal-frontend/commit/8d647e93f9a5e5e963404581da3e4ac0cd529f23

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4122,43 +4102,25 @@ export class StudyViewPageStore {
...
-            survivalTypes.push({
-                id: UniqueKey.OVERALL_SURVIVAL,
-                title: 'Overall Survival',
-                associatedAttrs: [OS_STATUS, OS_MONTHS],
-                filter: ['DECEASED'],
-                survivalData: [],
-            });
-        }
-        if (dfsStatusFlag && dfsMonthsFlag) {
+        this.survivalClinicalAttributesPrefix.result.map(prefix => {
```
alisman: let survivalTypes = result.map( iterator returns element of map)

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -971,13 +975,26 @@ export default class ComparisonStore {
...
+        await: () => [
+            this.activeSamplesNotOverlapRemoved,
+            this.survivalClinicalAttributesPrefix,
+        ],
         invoke: async () => {
             if (this.activeSamplesNotOverlapRemoved.result!.length === 0) {
                 return false;
             }
+            const attributeNames: string[] = [];
+            _.reduce(
```
alisman: reduce should always return the product of the operation.   it should be attributeNames = _.reduce(...)

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -997,13 +1014,25 @@ export default class ComparisonStore {
...
-            await: () => [this.activeSamplesNotOverlapRemoved],
+            await: () => [
+                this.activeSamplesNotOverlapRemoved,
+                this.survivalClinicalAttributesPrefix,
+            ],
             invoke: () => {
                 if (this.activeSamplesNotOverlapRemoved.result!.length === 0) {
                     return Promise.resolve([]);
                 }
+                const attributeNames: string[] = _.reduce(
```
alisman: this is correct usage

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1035,6 +1064,34 @@ export default class ComparisonStore {
...
 
+    readonly survivalClinicalAttributesPrefix = remoteData({
+        await: () => [this.activeStudiesClinicalAttributes],
+        invoke: () => {
+            const attributes = getSurvivalAttributes(
+                this.activeStudiesClinicalAttributes.result!
+            );
+            let attributesPrefix: string[] = [];
+            // get paired attributes
+            attributes.forEach(attribute => {
```
alisman: this is actually a reduce or filter operation since you are selectively pushing into attributesPrefix.    code would be clearer if you did ```attributesPrefix = arr.reduce or arr.filter```

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1035,6 +1064,34 @@ export default class ComparisonStore {
...
+                if (!attributesPrefix.includes(prefix)) {
+                    if (attributes.includes(`${prefix}_MONTHS`)) {
+                        attributesPrefix.push(prefix);
+                    }
+                }
+            });
+            // change prefix order based on priority
+            return Promise.resolve(
+                _.sortBy(attributesPrefix, prefix => {
+                    return plotsPriority[prefix] || 999;
```
alisman: 999 is a bit of a magic number here.  make a constant with a name that explains what it is.  like, MAX_PRIORITY or something

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1181,71 +1229,49 @@ export default class ComparisonStore {
...
             this.activeStudiesClinicalAttributes,
             this.activeStudyIdToStudy,
+            this.survivalClinicalAttributesPrefix,
         ],
         invoke: () => {
-            const diseaseFreeSurvivalClinicalAttributeId = 'DFS_STATUS';
-            const clinicalAttributeMap = _.groupBy(
-                this.activeStudiesClinicalAttributes.result,
-                'clinicalAttributeId'
+            return Promise.resolve(
```
alisman: better to assign this to a local variable and then Proimse.resolve(variable) at end.  otherwise, too much nesting

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1047,43 +1104,34 @@ export default class ComparisonStore {
...
+        await: () => [
+            this.survivalClinicalDataGroupByUniquePatientKey,
+            this.activePatientKeysNotOverlapRemoved,
+            this.survivalClinicalAttributesPrefix,
+        ],
+        invoke: () => {
+            return Promise.resolve(
+                _.reduce(
+                    this.survivalClinicalAttributesPrefix.result!,
+                    (result, key) => {
```
alisman: using "result" here is a little confusing because we are processing the result property of the promise.   we usually use "acc" (accumulator) as the name of this argument.

src/pages/resultsView/survival/SurvivalUtil.tsx
leexgh: yes I know that, thanks

jjgao:
from @jjgao: Can we add the info icon to the km plots in study view page?

from @schultzn: 

A couple of thoughts:
For AACR Project GENIE, we can expect dozens of different outcome measures for each study, so having them all show up in the current format might be tricky. Maybe we can show them in a pull down menu once there are >= 10 outcome measures? These could be prioritized by the outcome measures for which more data are available (as we start showing outcome for specific therapies. the data will get sparse and there may be only a few patients with outcome information), similar to what we are already doing in other places for the availability of clinical attributes (OncoPrint, study view).
The description of each outcome measure could be shown in a separate line in black (not bold) on a new line right under the header).

jjgao:
@leexgh based on comments above, can we do the followings?

- [ ] add the info icon to the km plots in study view page
- [ ] The description of each outcome measure could be shown in a separate line in black (not bold) on a new line right under the header)
- [ ] remove axis labels

Supporting many plots is captured [here](https://github.com/cBioPortal/cbioportal/issues/7292). Let's do it in a separate pr.

Resizing charts is captured [here](https://github.com/cBioPortal/cbioportal/issues/7293)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3466,43 +3464,24 @@ export class StudyViewPageStore {
         const cancerTypeIds = _.uniq(
             this.queriedPhysicalStudies.result.map(study => study.cancerTypeId)
         );
+        const survivalUniqueKeys = this.survivalPlotKeys;
```
kalletlak: looks like `this.survivalPlotKeys` can be empty here this.survivalClinicalAttributesPrefix.result is not complete. I think `this.survivalClinicalAttributesPrefix.pending` should be added to `loadingInitialDataForSummaryTab`

@leexgh @alisman @zhx828

PR #3100, opened 2020-03-06
Mutation Table: Fix annotation error when transcript_consequences is not present

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7277


onursumer:
failing tests are unrelated, merging.

PR #3102, opened 2020-03-09
check for relative import from packages directory

onursumer:
Enforcing (1) of https://github.com/cBioPortal/cbioportal-frontend/blob/master/README.md#tips-for-dependency-management

## Checks

### DIFF
--- a/src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
+++ b/src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
@@ -19,6 +19,7 @@ import {
     NumericGeneMolecularData,
 } from '../../../../shared/api/generated/CBioPortalAPI';
 import {
+    generateCopyNumberAlterationQuery,
     generateQueryVariantId,
     getProteinPositionFromProteinChange,
     OncoKbAPI,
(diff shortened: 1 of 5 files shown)


src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
```diff
@@ -27,7 +28,6 @@ import {
     EvidenceType,
     IOncoKbData,
 } from 'cbioportal-frontend-commons';
-import { generateCopyNumberAlterationQuery } from 'cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils';
```
onursumer: these kinds of imports can still sneak in, but they are not as bad as import from `../../../packages`

inodb: is there an easy way to avoid this? E.g. adding a grep for `cbioportal-frontend-commons/.*`? Or maybe a little more fancy with grepping all packages `PACKAGES=$(ls packages/* | cut -d/ -f2)` and then using that to grep

onursumer: we still need import statements like `import 'cbioportal-frontend-commons/dist/styles.css'`, so maybe we only allow importing style files but nothing else?

PR #3109, opened 2020-03-18
Load pathway mapper dynamically to avoid IE11 crash

alisman:
Pathway mapper library is not compatible with IE11.  Merely loading into browswer breaks portal.  This PR loads pathway mapper library dynamically and only when we are not in internet explorer (any version).

### DIFF
--- a/src/pages/resultsView/pathwayMapper/ResultsViewPathwayMapper.tsx
+++ b/src/pages/resultsView/pathwayMapper/ResultsViewPathwayMapper.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import _ from 'lodash';
 import { ResultsViewPageStore } from '../ResultsViewPageStore';
 import { addGenesToQuery, ResultsViewTab } from '../ResultsViewPageHelpers';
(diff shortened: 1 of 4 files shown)


src/pages/resultsView/pathwayMapper/ResultsViewPathwayMapper.tsx
```diff
@@ -30,6 +29,7 @@ import 'cytoscape-panzoom/cytoscape.js-panzoom.css';
 import 'cytoscape-navigator/cytoscape.js-navigator.css';
 import 'react-toastify/dist/ReactToastify.css';
 import styles from './pathwayMapper.module.scss';
+import PathwayMapper, { ICBioData } from 'pathway-mapper';
```
alisman: it's interesting that if you only use these as types, then the library doesn't get bundled.  which is what we want.

adamabeshouse: interesting. I guess that makes sense since types are not included in runtime

alisman:
@adamabeshouse thanks.  you have no idea how long this took me to figure out.

PR #3111, opened 2020-03-18
Fix error in FixedHeaderTable when filter string is not empty

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7246
Fix https://github.com/cBioPortal/cbioportal/issues/7214
Fix https://github.com/cBioPortal/cbioportal/issues/7119

### DIFF
--- a/src/pages/studyView/table/FixedHeaderTable.tsx
+++ b/src/pages/studyView/table/FixedHeaderTable.tsx
@@ -13,7 +13,7 @@ import {
     TableHeaderProps,
 } from 'react-virtualized';
 import 'react-virtualized/styles.css';
-import { action, computed, observable } from 'mobx';
+import { action, computed, observable, toJS } from 'mobx';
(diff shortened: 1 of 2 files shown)


src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -128,12 +127,16 @@ export default class FixedHeaderTable<T> extends React.Component<
...
             return this.props.data;
         }, this.props.fixedTopRowsData || []);
+        const filterString = toJS(this._store.filterString);
         this._store.setProps({
             columns: nextProps.columns,
             dataStore: tableDataStore,
             initialSortColumn: this._sortBy,
             initialSortDirection: this._sortDirection,
         });
+        if (filterString !== undefined) {
```
alisman: interesting.  i think we should have comment explaining why we read and then set this again

PR #3112, opened 2020-03-18
No max width on oncoprint tooltip, and improvements in line breaks

adamabeshouse:


tmazor:
@adamabeshouse I can't find any example of text extending outside the box, so that's great!

But the gene tooltips look very narrow (narrower than on production currently) and there are a lot of line breaks within words in this tooltip, which isn't ideal.
image
Can those gene tooltips be wider also?

src/shared/components/oncoprint/styles.scss
```diff
@@ -1,7 +1,8 @@
 .oncoprintjs__tooltip {
     padding: 5px;
     margin-top: -10px; // to compensate position for padding, since tooltip is positioned above (if positioned at center, wouldnt be problem)
-    width: 250px;
+    //width: 250px;
+    max-width: unset !important;
```
alisman: wow.  never new about unset.

alisman:
@adamabeshouse i'm a little worried this will result in epically long tooltips.  what's reason this was changed?

adamabeshouse:
@alisman theres just no way to square the circle - we want certain elements to not have line breaks inside them, but we also don't want the text to go out of the bounds of the tooltip. somethings gotta give

adamabeshouse:
@tmazor just updated it, let me know what you think! I increased the width of the tooltip, and fixed the issue where cna/mrna/protein alterations were being rendered with line breaks

tmazor:
This is much better, but I can still find the occasional tooltip with a line break mid-word. It's rare, so I could live with it if this is a very hard thing to fix, but it's still not ideal.
image

PR #3118, opened 2020-03-20
saml local logout query parameter included in the url

amirineni1:
fixes cBioPortal/cbioportal#7313 

Added saml logout local query parameter to signout link for saml authentication


n1zea144:
Hi @amirineni1 @dippindots I'm not sure this PR is required - see comment on issue.

amirineni1:
`  this.props.appStore.logoutUrl,
   {
       local: AppConfig.serverConfig.saml_logout_local.toString(),
    }`

for SAML authentication, local=true should be passed in the /saml/logout in order to perform local logout. Please see my additional comments in the issue

dippindots:
@n1zea144 The changes in this pr looks good to me, I am just wondering that there are lots of other saml parameters also, we can do this one time, but we also need to do the same thing for other parameters, am I right? Is there a generic solution to include all saml parameters?

amirineni1:
@n1zea144 @dippindots this particular SAML parameter part of portal.properties file but not handled in the react app so it is kind of missing functionality or a bug so this needs to go first. Other SAML parameters in the portal.properties were used in the spring security context for SAML authentication. My knowledge, I don't see any parameters needs to be considered for now.

src/shared/components/dataAccessTokens/DataAccessTokensDropdown.tsx
```diff
@@ -53,7 +53,10 @@ export class DataAccessTokensDropdown extends React.Component<
                 action: (
                     <a
                         href={buildCBioPortalPageUrl(
-                            this.props.appStore.logoutUrl
+                            this.props.appStore.logoutUrl,
+                            {
+                                local: AppConfig.serverConfig.saml_logout_local.toString(),
```
n1zea144, line 58: @amirineni1 @alisman From what I understand, if saml_logout_local is true, the logout URL is suppose to be extended to something like the following (as opposed to "/saml/logout"):

> /saml/logout?local=true

Where does this happen?  Whats happens when you convert a boolean to a string here?

amirineni1: @n1zea144 src/appShell/App/PortalHeader.tsx is the file will consume DataAccessTokensDropdown to populate "Sign Out" link.  Since parameter "local" is a type of "string", converted boolean value to string. This would match the type and will not show any typescript errors. toString method simply converts a boolean value to string and return either "true" or "false" value.

amirineni1: @n1zea144 could you please review and merge these changes? Let me know if any additional questions. thanks

n1zea144: Sorry I'm being so "thick" here - as I understand it, default logout URL is something like this:

"/saml/logout", but when saml_logout_local is "true" the URL should change to something like this:

"/saml/logout?local=true"

where in PortalHeader.tsx does this happen?

amirineni1: @n1zea144 
PortalHeader.tsx below code calls DataAccessTokensDropdown
`                                        <Dropdown.Menu
                                            style={{
                                                paddingLeft: 10,
                                                overflow: 'auto',
                                                maxHeight: 300,
                                                whiteSpace: 'nowrap',
                                            }}
                                        >
                                            <DataAccessTokensDropdown
                                                appStore={this.props.appStore}
                                            />
                                        </Dropdown.Menu>`

In DataAccessTokensDropdown.tsx signout URL will be constructed using buildCBioPortalPageUrl method. This method requires two parameters 
- one is the actual URL that should return by this.props.appStore.logoutUrl method ("/saml/logout")
- second one is query parameter (local=true|false) that should be retrived from the serverConfig
`
               id: 'signout',
                action: (
                    <a
                        href={buildCBioPortalPageUrl(
                            this.props.appStore.logoutUrl, ---> "/saml/logout"
                            {
                                local: AppConfig.serverConfig.saml_logout_local.toString(), ---> "local=true|false"
                            }
                        )}
                    >
`

buildCBioPortalPageUrl("/saml/logout", {local:true})

n1zea144: I see, thats for the follow-up info.

PR #3119, opened 2020-03-23
Improve Annotation Column Data & Renderer

onursumer:
- Fix cBioPortal/cbioportal/issues/6991: Move the `MyCancerGenome` column renderer as a new component into `react-mutation-mapper` package, and simplify `AnnotationColumnFormatter` by reusing the `Annotation` column component from `react-mutation-mapper`.
- Fix cBioPortal/cbioportal/issues/5167: Remove OncoKB data from `ResultsViewPageStore` and `MutationMapperToolStore`. This allows fetching OncoKB data per gene inside separate `MutationMapperStore` instances, thus reduces loading time for mutation mapper. Currently, we are fetching OncoKB data for all genes at once.

## Checks


packages/react-mutation-mapper/src/util/MyCancerGenomeUtils.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+        it('extracts URL and text from a pre-formatted my cancer genome HTML string', () => {
+            const htmlLink =
+                '<a href="http://mycancergenome.org/content/disease/colorectal-cancer/kras/38/">KRAS c.37G>T (G13C) Mutation in Colorectal Cancer</a>';
+
+            const parsed = parseMyCancerGenomeLink(htmlLink);
+
+            assert.isNotNull(parsed);
+            assert.equal(
+                parsed!.text,
+                'KRAS c.37G>T (G13C) Mutation in Colorectal Cancer'
```
zhx828: Would it be better just put this as a constant?

packages/react-mutation-mapper/src/util/MyCancerGenomeUtils.ts
```diff
@@ -0,0 +1,110 @@
+import { IMyCancerGenome, IMyCancerGenomeData } from '../model/MyCancerGenome';
+import { Mutation } from '../model/Mutation';
+
+import myCancerGenomeJson from '../resources/mycancergenome.json';
+
+export function getMyCancerGenomeLinks(
```
zhx828: Seems like this is the only method to be tested?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -219,51 +215,68 @@ export function sortValue(annotation: IAnnotation): number[] {
...
         oncoKbSortValue(annotation.oncoKbIndicator),
         civicSortValue(annotation.civicEntry),
-        // MyCancerGenome.sortValue(annotationData.myCancerGenomeLinks),
+        myCancerGenomeSortValue(annotation.myCancerGenomeLinks),
         hotspotSortValue(annotation.isHotspot, annotation.is3dHotspot),
         annotation.isOncoKbCancerGene ? 1 : 0,
     ]);
 }
 
+export function renderAnnotation(
```
adamabeshouse: maybe worth using a react functional component for this?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -219,51 +215,68 @@ export function sortValue(annotation: IAnnotation): number[] {
...
+    annotation: IAnnotation,
+    enableOncoKb: boolean,
+    enableCivic: boolean,
+    enableMyCancerGenome: boolean,
+    enableHotspot: boolean,
+    pubMedCache?: MobxCache,
+    userEmailAddress?: string
+): JSX.Element {
+    return (
```
adamabeshouse: whats the point of pulling this out?

adamabeshouse: whats the point of pulling this out?

packages/react-mutation-mapper/src/component/myCancerGenome/MyCancerGenome.tsx
```diff
@@ -0,0 +1,98 @@
...
+    );
+}
+
+export default class MyCancerGenome extends React.Component<
+    IMyCancerGenomeProps,
+    {}
+> {
+    constructor(props: IMyCancerGenomeProps) {
+        super(props);
+        this.state = {};
```
adamabeshouse: why are you using state instead of mobx observer?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -162,6 +165,11 @@ class DefaultMutationMapperStore implements MutationMapperStore {
         );
     }
 
+    @computed
+    public get myCancerGenomeData(): IMyCancerGenomeData {
+        return fetchMyCancerGenomeData();
```
adamabeshouse: I don't think the function should be called `fetch` if its not actually doing any fetching, a little confusing IMO

packages/react-mutation-mapper/src/util/MyCancerGenomeUtils.spec.ts
onursumer: I just left the existing test as is, I think this is not too bad to leave like this.

packages/react-mutation-mapper/src/component/column/Annotation.tsx
onursumer: I wasn't able to directly use the `Annotation` component for the PatientView mutation table (because it doesn't use Mutation data), so instead calling this function to render it there: https://github.com/cBioPortal/cbioportal-frontend/pull/3119/files#diff-a4d36866c4ebe7d81ddb47337d37759eR273. Ideally `Annotation` component should be more generic, but I am leaving that for another PR.

packages/react-mutation-mapper/src/component/myCancerGenome/MyCancerGenome.tsx
onursumer: Hmm, we are not even using that `state` object anywhere. Actually, I just moved this file without modifying it, but for some reason git shows this as a new addition.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: Right, it's not fetching anything, my cancer genome data is all static. I kept the original function name but probably it's better to rename it.

packages/react-mutation-mapper/src/util/MyCancerGenomeUtils.ts
onursumer: Ideally we should test all the methods, but I haven't added any new tests, just moved existing MyCancerGenome related functions (without modifying) into this new file.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -162,6 +165,11 @@ class DefaultMutationMapperStore implements MutationMapperStore {
         );
     }
 
+    @computed
```
alisman: i'm not sure this should be computed.  does it reference any observable values inside of it?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
adamabeshouse: ah okay. I didn't realize you were using it somewhere else. This is fine but we've starting using react functional components for stuff like this, maybe it would be good to do that here, call it `GenericAnnotation` or something

packages/react-mutation-mapper/src/component/myCancerGenome/MyCancerGenome.tsx
```diff
@@ -0,0 +1,96 @@
...
+            <ul className={mcgStyles['link-list']}>{links}</ul>
+        </span>
+    );
+}
+
+export default class MyCancerGenome extends React.Component<
+    IMyCancerGenomeProps
+> {
+    constructor(props: IMyCancerGenomeProps) {
+        super(props);
```
adamabeshouse: unnecessary constructor

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: hmm, right, computed has no effect on this one. I'll just remove it.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -162,6 +165,11 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+    public get myCancerGenomeData(): IMyCancerGenomeData {
+        return getMyCancerGenomeData();
```
adamabeshouse: is there a point to wrapping this call now? I guess speaking to aarons point, maybe you should just declare it as a constant i.e. `public myCancerGenomeData = getMyCancerGenomeData()`

onursumer: right, a constant may work better for this. It's just static data anyway.

jjgao:
Cool! It’s so fast to load up in the Mutations tab!

Annotating OncoPrint seems low though - wondering if it can also be improved?

jjgao:
Also, can we implement the same dialog as OncoPrint: 

image

jjgao:
Discussed on Slack.

Onur:owl:  5:41 PM
OncoPrint data loading part is still the same, I have only improved the mutations tab. Not sure if it is easy to fix it for OncoPrint, but worth investigating.

JJ Gao  5:43 PM
What was the optimization?

Onur:owl:  5:44 PM
Now fetching the OncoKB data only after clicking the corresponding gene tab. But we cannot do that for OncoPrint I guess.

JJ Gao  5:45 PM
so you are now loading annotation 1 gene at a time?
new messages

Onur:owl:  5:48 PM
Yes, each tab sends its own request in its own store. But oncokb data for oncoprint still lives in the main store, and we load everything at once. Not sure if we can do anything about that.

JJ Gao  5:49 PM
Got it, thanks. We will have to load all genes in oncoprint.

PR #3120, opened 2020-03-24
Support grch38 in mutation mapper tool

leexgh:
fix frontend: https://github.com/cBioPortal/cbioportal/issues/7327
Support grch38 in mutation mapper tool. 

**Changes in this pr:**

- Add a check box on the mutation mapper tool page to switch between grch37 and grch38
image

- Example data will change accordingly
image

- Show a warning message when annotated with GRCh38
image

- Annotation results are from grch38 genome nexus if choose `GRCh38`

**Test:**
https://deploy-preview-3120--cbioportalfrontend.netlify.com/mutation_mapper


src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -246,13 +268,64 @@ export default class MutationMapperTool extends React.Component<
...
+                        <strong style={{ paddingRight: '5px' }}>
+                            Select Reference Genome:{' '}
+                        </strong>
+                        <ButtonGroup>
+                            <Radio
+                                checked={
+                                    this.referenceGenomeSelection ===
+                                    referenceGenome.GRCH37
+                                }
+                                onChange={() => {
```
onursumer: Let's have this handler as a separate function if possible.

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -246,13 +268,64 @@ export default class MutationMapperTool extends React.Component<
...
+                                data-value={referenceGenome.GRCH37}
+                            >
+                                {referenceGenome.GRCH37}
+                            </Radio>
+                            <Radio
+                                checked={
+                                    this.referenceGenomeSelection ===
+                                    referenceGenome.GRCH38
+                                }
+                                onChange={() => {
```
onursumer: Let's have this handler as a separate function if possible.

src/shared/cache/Grch38GenomeNexusCache.ts
```diff
@@ -0,0 +1,49 @@
+import { fetchVariantAnnotationsByMutation } from 'shared/lib/StoreUtils';
```
onursumer: Do we really need to implement a whole new cache class? Can't we modify the existing one to make it compatible with different URLs? This is mostly copy/paste from the original `GenomeNexusCahce`, right?

src/shared/cache/Grch38GenomeNexusMutationAssessorCache.ts
```diff
@@ -0,0 +1,50 @@
+import * as _ from 'lodash';
```
onursumer: Same here, do we really need a new implementation?

src/shared/cache/Grch38GenomeNexusMyVariantInfoCache.ts
```diff
@@ -0,0 +1,76 @@
+import { fetchVariantAnnotationsByMutation } from 'shared/lib/StoreUtils';
```
onursumer: Same here, do we really need a new implementation?

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -38,13 +50,23 @@ export default class MutationMapperTool extends React.Component<
...
+    @observable referenceGenomeSelection: string = referenceGenome.GRCH37;
 
     private store: MutationMapperToolStore = new MutationMapperToolStore();
 
     constructor(props: IMutationMapperToolProps) {
         super(props);
 
         this.userSelectionStore = new MutationMapperUserSelectionStore();
+        // store genome build selection to localStorage, so it will not get lost after reload the page
+        if (getBrowserWindow().localStorage.getItem('genomenexus_url_grch38')) {
```
alisman: is this just for testing purposes? the comment says "store" it in localStorage, but all i see is reading from localstorage.

alisman: ok, i see.  this i a user setting

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -587,4 +642,33 @@ export default class MutationMapperTool extends React.Component<
...
     }
+
+    @autobind
+    @action
+    protected handleReferanceGenomeSelectionChange(selection: string) {
+        if (selection === referenceGenome.GRCH37) {
+            this.referenceGenomeSelection = referenceGenome.GRCH37;
+            // if choose grch37, remove genomenexus_url_grch38 from localStorage
+            if (
+                getBrowserWindow().localStorage.getItem(
```
alisman: rather that embedding the genome Id in the localStorage key, I think it would be better to have a single "referenceGenomeId" item that could represent any referenceGenome.

alisman:
@leexgh, @inodb got a few questions on this one.   This seems to affect only standalone mutation mapper tool?  Why is it that other features of cbioportal do not support this toggling?   Should we accomplish this in a more global and reactive way that would not require reloading of the page?  For example, making the current refgenome setting an observable property of appStore?   Also, the current implemenation is not generic.  The ref genome Id is embedded in property names and keys.  Is that a problem?

inodb:
@alisman Thanks for checking!

> This seems to affect only standalone mutation mapper tool? 

Correct. We thought for a first pass, let's try to get this working for mutation mapper tool and see what issues we run into.

> Why is it that other features of cbioportal do not support this toggling? Should we accomplish this in a more global and reactive way that would not require reloading of the page? For example, making the current refgenome setting an observable property of appStore? 

This would be the eventual goal and is a requirement for the HTAN cBioPortal instance

> Also, the current implemenation is not generic. The ref genome Id is embedded in property names and keys. Is that a problem?

yeah it would be good to make it generic but for now we only need to support grch37 and grch38. So maybe it's fine?

To elaborate why we are doing this approach of multiple genome nexus instances. An alternative is to load all the data in one instance and add url parameters in the genome nexus server. Downside of that is that it makes the server logic more complicated. VEP does the same thing of having different instances for different genome builds. E.g. https://grch37.rest.ensembl.org/ and https://rest.ensembl.org. That's one of the reasons we went with that approach.

Also note: I don't think we want to support the use case of comparing studies with different genome builds. Seems complicated. It is prolly easier to do this at the data level and load in multiple coordinates for a single study to achieve the same

leexgh:
@onursumer @inodb @alisman thanks for review, could you take a look again?

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -573,18 +630,35 @@ export default class MutationMapperTool extends React.Component<
     @autobind
     @action
     protected handleLoadExamplePartiallyAnnotated() {
-        this.inputText = require('raw-loader!./resources/standaloneMutationDataExample.txt');
+        if (this.referenceGenomeSelection === referenceGenome.GRCH37) {
+            this.inputText = require('raw-loader!./resources/standaloneMutationDataExample.txt');
```
alisman: just rename this file standaloneMutationDataExampleGrch37?  then you can compose it

src/shared/cache/GenomeNexusCache.ts
```diff
@@ -22,6 +23,20 @@ export function fetch(queries: Mutation[]): Promise<VariantAnnotation[]> {
     }
 }
 
+// fetch function for GRCh38
```
alisman: is there a reason we duplicate this function below?

src/shared/cache/GenomeNexusCache.ts
```diff
@@ -3,13 +3,14 @@ import {
...
 } from 'shared/lib/MutationUtils';
-import { VariantAnnotation } from 'cbioportal-frontend-commons';
+import { VariantAnnotation, GenomeNexusAPI } from 'cbioportal-frontend-commons';
 import { Mutation } from 'shared/api/generated/CBioPortalAPI';
 import LazyMobXCache, { CacheData } from 'shared/lib/LazyMobXCache';
 import AppConfig from 'appConfig';
 
 export type GenomeNexusCacheDataType = CacheData<VariantAnnotation>;
 
+// fetch function for GRCh37
```
alisman: we should really rename this function.  fetch is too generic and should be reserved for the built in fetch function.

src/shared/cache/GenomeNexusCache.ts
alisman: these cache should be able to use the client that you've set up now in the store.  they should not need to know about which GRCh38 or 37.

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -368,20 +396,28 @@ export default class MutationMapperToolStore {
...
     }
 
     @cached get genomeNexusCache() {
-        return new GenomeNexusCache();
+        return new GenomeNexusCache(this.grch38GenomeNexusUrl ? true : false);
     }
 
     @cached get genomeNexusMutationAssessorCache() {
-        return new GenomeNexusMutationAssessorCache();
+        return new GenomeNexusMutationAssessorCache(
```
alisman: pass the caches the client here.

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -86,6 +93,18 @@ export default class MutationMapperToolStore {
         return AppConfig.serverConfig.isoformOverrideSource;
     }
 
+    @computed get genomeNexusClient() {
```
alisman: this is good.  now lets see if we can make this the ONLY point in the code which knows about 38 vs 37

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -405,4 +441,53 @@ export default class MutationMapperToolStore {
     @action public clearCriticalErrors() {
         this.criticalErrors = [];
     }
+
+    // override fetch function in GenomeNexusCache with selected GN client
```
alisman: lets avoid this repetition with a factory for making these fetch functions:
`function fetchFactory(fields, client){
  
  return function(query) {
    console.log(fields);
    console.log(client)
    console.log(query);
  }
  
}

var func1 = makeFetch(["one","two"]);

var func2 = makeFetch(["three", "four"]);

func1("query1");

func2("query2");

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -405,4 +448,19 @@ export default class MutationMapperToolStore {
     @action public clearCriticalErrors() {
         this.criticalErrors = [];
     }
+
+    private fetchFactory(fields: string[], client: GenomeNexusAPI) {
```
alisman: great

src/shared/cache/Grch38GenomeNexusCache.ts
leexgh: Had a discussion about it. We need to use different urls in the fetch function, but fetch function can’t accept other urls, since the GenomeNexusCache is passing the fetch as a function to LazyMobXCache. Ideally we should be able to pass URL to the constructor and use that in the fetch function, but it might be too much work to achieve that at this point

src/shared/cache/GenomeNexusCache.ts
leexgh: same reason as above

src/shared/cache/Grch38GenomeNexusCache.ts
leexgh: Temporary fix we had now is to pass a fetch function from store.

inodb:
We discussed this and came up with following items:

- [ ] info banner with warning about grch38 describing potential issues with oncokb, hotspots, civic, 3d structure, mycancergenome
- [ ] disable myvariant.info columns gnomad, clinvar, dbsnp
- [ ] hgvsg links to grch38.genomenexus.org
- [ ] update ensembl link to ensembl.org instead of grch37.ensembl.org
- [ ] add “beta” indicator next to grch38 bullet (similiar to quick search)
- [ ] ask myvariant.info team about status of directly querying by grch38 variant instead of grch37
- [ ] update mutationassessor link (figure out what URL to use for grhc38 variants)

let's maybe wait with these:

- [ ] wait with disabling tracks based on reference genome selection (would be nice to have this in code maybe, but let’s see what comes out of Hongxin’s analysis)
- [ ] maybe update genomenexus frontend to remove myvariant.info related fields when using grch38

onursumer:
@inodb Do you think we should implement all of these on the portal side or does it make sense to have some of the logic implemented in the `react-mutation-mapper` package? Should `react-mutation-mapper` be aware of the genome build concept at all or it should stay generic and visualize whatever data it receives?

inodb:
@onursumer good question - you can prolly answer that question better than me but here's my two cents. I think ideally it's a visualization layer that doesn't know about grch38/grch37. Based on the response from the API of genome nexus it can figure out what to show / not to show. Say e.g. if you're pointing to grch38 genome nexus instance it will not include annotations in the response that don't support grch38. So instead of having to put a lot of visible/invisible in configuration it could ideally determine what to show from the response. Ofc we can use some show/hide configuration for now, but I think that's what we should be aiming for. That way you can imagine in the future that if we add new annotations in the genome nexus API we could have default visualizations in e.g. the mutations table (like plain text) that would show up without having to first add frontend logic.

leexgh:
follow up with myVariant.info:
the `variant` endpoint supports assembly=hg38 parameter to query on variants on hg38. 
https://myvariant.info/v1/variant/chr7:g.55181378C%3ET?assembly=hg38

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -661,4 +661,36 @@ export default class MutationMapperTool extends React.Component<
         // reload the page to initialize genome nexus api instance with correct url
         getBrowserWindow().location.reload();
     }
+
+    protected grch38Info() {
```
alisman: might as well make this a computed getter.   this will have neglibile impact, but is a best practice

alisman: it basically will cache so React doesn't have to recompute virtual DOM ever time

src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
```diff
@@ -16,7 +20,11 @@ export default class HgvsgColumnFormatter {
         if (!hgvsg) {
             return <span />;
         } else {
-            let genomeNexusUrl = buildGenomeNexusHgvsgUrl(hgvsg);
+            let genomeNexusUrl =
```
alisman: you should put this check inside buildGenomeNexusHgvsgUrl function.   This code shouldn't have to know about 38 vs 37.

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
leexgh: @alisman this is just a block of info and doesn't have anything to compute, I'm not sure is it still makes sense to have `computed get`

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -102,6 +98,7 @@ export default class MutationMapperTool extends React.Component<
                     />
                     {this.store.mutationsNotAnnotated.length > 0 &&
                         this.inputParseWarning()}
+                    {this.shouldShowGrch38Warning && this.grch38Info()}
```
onursumer: We should probably rename either the variable to `shouldShowGrch38Info` or the method to `grch38Warning`

end-to-end-test/remote/specs/core/mutationMapperTool.spec.js
```diff
@@ -296,4 +296,108 @@ describe('Mutation Mapper Tool', function() {
...
+            browser.waitForVisible(
+                '[data-test=GenomicChangesExampleButton]',
+                10000
+            );
+            browser.click('[data-test=GenomicChangesExampleButton]');
+            browser.click('[data-test=MutationMapperToolVisualizeButton]');
+
+            waitForGenomeNexusAnnotation();
+
+            // mutations table should be visiable after oncokb icon shows up,
```
onursumer, line 323:
```suggestion
            // mutations table should be visible after oncokb icon shows up,
```

end-to-end-test/remote/specs/core/mutationMapperTool.spec.js
```diff
@@ -296,4 +296,108 @@ describe('Mutation Mapper Tool', function() {
...
+            let mutationCount = browser.getText(
+                './/*[text()[contains(.,"122 Mutations")]]'
+            );
+            assert.ok(mutationCount.length > 0);
+
+            const brca1 = browser.getText('.//*[text()[contains(.,"BRCA1")]]');
+            assert.ok(brca1.length > 0);
+            browser.elements('.nav-pill').click('a*=BRCA1');
+
+            // mutations table should be visiable after oncokb icon shows up,
```
onursumer, line 350:
```suggestion
            // mutations table should be visible after oncokb icon shows up,
```

end-to-end-test/remote/specs/core/mutationMapperTool.spec.js
```diff
@@ -296,4 +296,108 @@ describe('Mutation Mapper Tool', function() {
...
+            browser.waitForVisible(
+                '[data-test=GenomicChangesExampleButton]',
+                10000
+            );
+            browser.click('[data-test=GenomicChangesExampleButton]');
+            browser.click('[data-test=MutationMapperToolVisualizeButton]');
+
+            waitForGenomeNexusAnnotation();
+
+            // mutations table should be visiable after oncokb icon shows up,
```
onursumer, line 390:
```suggestion
            // mutations table should be visible after oncokb icon shows up,
```

end-to-end-test/remote/specs/core/mutationMapperTool.spec.js
leexgh: @onursumer good catch, I also fix this typo in other tests

end-to-end-test/remote/specs/core/mutationMapperTool.spec.js
```diff
@@ -296,4 +296,108 @@ describe('Mutation Mapper Tool', function() {
...
+            );
+
+            // check total number of mutations (this gets Showing 1-25 of 122
+            // Mutations)
+            let mutationCount = browser.getText(
+                './/*[text()[contains(.,"122 Mutations")]]'
+            );
+            assert.ok(mutationCount.length > 0);
+
+            const brca1 = browser.getText('.//*[text()[contains(.,"BRCA1")]]');
```
alisman, line 346: xpath!

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -18,11 +24,19 @@ import MutationMapperToolStore from './MutationMapperToolStore';
...
 import AppConfig from 'appConfig';
 import { getOncoKbApiUrl } from 'shared/api/urls';
+import { getBrowserWindow } from 'cbioportal-frontend-commons';
+import { referenceGenome } from 'shared/lib/referenceGenomeUtils';
 
 interface IMutationMapperToolProps {
     routing: any;
 }
 
+const ONCOKB_URL = 'https://www.oncokb.org/';
```
alisman, line 34: are these already available in urls.ts?

leexgh: @alisman seems we only have oncokb.org/api/v1

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -449,6 +473,60 @@ export default class MutationMapperTool extends React.Component<
         );
     }
 
+    protected referenceGenomeSelector() {
+        // show reference genome selector if "show_mutation_mapper_tool_grch38" is true
+        if (AppConfig.serverConfig.show_mutation_mapper_tool_grch38) {
```
inodb, line 478: nice! If people want to test this you can use:

```
localStorage.frontendConfig = '{"serverConfig":{"show_mutation_mapper_tool_grch38":false}}'
```
Worked for me :)

src/shared/components/mutationMapper/MutationMapperConfig.ts
```diff
@@ -12,6 +16,8 @@ export interface IMutationMapperConfig {
     uniprot_id_url: string | null;
     ensembl_transcript_url: string | null;
     genomenexus_url: string | null;
+    genomenexus_client?: GenomeNexusAPI;
+    genomenexus_internal_client?: GenomeNexusAPIInternal;
```
onursumer, line 20: These should probably go into `IMutationMapperStoreConfig` with proper camel case names like `genomeNexusClient` and `genomeNexusInternalClient`. `IMutationMapperConfig` is simply a subset of `IServerConfig`. Any value that doesn't come directly from the server side shouldn't be included in this interface.

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
```diff
@@ -511,6 +473,60 @@ export default class MutationMapperTool extends React.Component<
         );
     }
 
+    protected referenceGenomeSelector() {
```
alisman, line 476: should be @computed

also, did you mean to make this protected?  i would think private if anything

alisman:
@leexgh reminder that you'll need to make a PR to backend to add this property to example configuration and also config_service

leexgh:
@alisman yes I have a backend pr https://github.com/cBioPortal/cbioportal/pull/7348

src/pages/staticPages/tools/mutationMapper/MutationMapperTool.tsx
leexgh: @alisman yes `@computed` would be proper, just because other similar methods in this file have `protected` so I declared `protected`  as well. I just asked @onursumer about this, for this particular class we don't have class extending use-case, so I'll make it `@computed` for it

src/shared/components/mutationMapper/MutationMapperConfig.ts
leexgh: discussed with @onursumer , remove this and add a constructor parameter in MutationMapperStore

src/shared/components/mutationMapper/MutationMapperConfig.ts
```diff
@@ -1,4 +1,8 @@
 import { IMutationMapperProps } from './MutationMapper';
+import {
+    GenomeNexusAPI,
+    GenomeNexusAPIInternal,
+} from 'genome-nexus-ts-api-client';
```
onursumer, line 5: unused import

PR #3121, opened 2020-03-24
Add new IE11 browserstack tests to localdb test job

alisman:
Add a new e2e job to test IE11 in browserstack.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,8 +39,6 @@ run_e2e_and_save_artifacts: &run_e2e_and_save_artifacts
           yarn run e2e
         when: always
         environment:
-          JUNIT_REPORT_PATH: ./remote/junit/
-          SCREENSHOT_DIRECTORY: ./remote/screenshots
           FRONTEND_TEST_USE_LOCAL_DIST: true
           NO_PARALLEL: true
     - run:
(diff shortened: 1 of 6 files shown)


end-to-end-test/ie11/specs/ie.js
```diff
@@ -0,0 +1,21 @@
...
+
+const CBIOPORTAL_URL = 'http://localhost:3000';
+
+describe('Application in IE11', function() {
+    it('renders query page study items', function() {
+        goToUrlAndSetLocalStorage(CBIOPORTAL_URL);
+        $('[data-test="StudySelect"]').waitForExist();
+    });
+
+    // it('renders the oncoprint', function() {
```
alisman: for some fucking reason this test won't pass. i'm putting it aside for another day.

alisman: i can confirm the feature works fine in IE11 on browserstack

.circleci/config.yml
```diff
@@ -63,6 +61,62 @@ run_e2e_and_save_artifacts: &run_e2e_and_save_artifacts
         path: /tmp/repo/end-to-end-test/shared/imageCompare.html
         destination: /imageCompare.html
 
+
+run_IE11_e2e_and_save_artifacts: &run_IE11_e2e_and_save_artifacts
```
alisman: this is hugely repetitive of the original e2e task.  i tried to parametize it but at the end of the day just wasn't smart enough to do it.

end-to-end-test/shared/wdio/webdriver-manager.conf.js
```diff
@@ -319,21 +333,17 @@ var config = {
...
+}
 
+if (process.env.TEST_BROWSERSTACK === 'true') {
     config.services = ['visual-regression', 'browserstack'];
-
     config.browserstackLocal = true;
-
-    config.user = process.env.BROWSERSTACK_USER;
-    config.key = process.env.BROWSERSTACK_KEY;
+    config.user = 'aaronlisman2'; //process.env.BROWSERSTACK_USER;
```
alisman: ok, i will hide these in circle.  and change the password since this is now in git history

scripts/serve_dist.sh
```diff
@@ -1,18 +1,36 @@
...
 set -u
 set -o pipefail
 
 SCRIPT_DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
 
+protocol="https"
+
+if [ $# -eq 1 ]
+  then
+    protocol="http"
```
inodb: this seems a little funky for handling arguments, if you want to avoid having to parse arguments properly could also do something like

```
protocol="${PROTOCOL:-https}"
```
If somebody exports PROTOCOL before calling this it would use that, `https` being default. Cleaner would be to do proper arg parsing but get that can be overkill since CI is probably the only one using this

end-to-end-test/shared/wdio/webdriver-manager.conf.js
```diff
@@ -319,21 +333,17 @@ var config = {
...
+    config.key = 'SsvxMVGa2fFVc5bWzswx'; // process.env.BROWSERSTACK_KEY;
```
inodb: maybe good to put these in cirlceci env config instead i.e. not in the github repo? Ofc in practice somebody could still steal the keys by opening a PR and echo'ing these vars but it's a bit more obfuscated

end-to-end-test/local/runtime-config/run_container_screenshot_test.sh
```diff
@@ -15,6 +15,20 @@ docker run -it --rm \
...
+    --name screenshot-e2e \
+    --network="$DOCKER_NETWORK_NAME" \
+    -e CBIOPORTAL_URL="$CBIOPORTAL_URL" \
+    -e SCREENSHOT_DIRECTORY="$SCREENSHOT_DIRECTORY"\
+    -e SPEC_FILE_PATTERN="./local/specs/core/**/ie.js" \
+    -e TEST_IE11="true" \
+    -e TEST_BROWSERSTACK="true" \
+    -e JUNIT_REPORT_PATH="$JUNIT_REPORT_PATH" \
+    -v "$PORTAL_SOURCE_DIR:/cbioportal-frontend/" \
+    $SCREENSHOT_IMAGE_NAME
```
inodb: this seems duplicated with `run_container_ie_test.sh`

end-to-end-test/local/runtime-config/run_container_ie_test.sh
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env bash
```
inodb: this seems duplicated  in `run_container_screenshot_test.sh`

.circleci/config.yml
```diff
@@ -276,15 +343,18 @@ workflows:
...
+#                    - install
+#            - end_to_end_tests:
+#                requires:
+#                    - install
+            - end_to_end_tests_IE11:
                 requires:
                     - install
+#            - end_to_end_tests_localdb:
+#                requires:
+#                    - install
```
inodb: i guess all these ☝️ are temp disabled, but let's not forget to re-enable before merging

alisman: yup

.circleci/config.yml
inodb: Maybe good to make the `Spin up frontend over ssl if necessary and run end to end tests` a bash script so at least that part has no duplication? Not essential but would be nice

end-to-end-test/local/runtime-config/run_container_screenshot_test.sh
inodb: I think having it in separate scripts would be cleaner e.g. in `run_container_ie_test.sh` as you had before? Now the localdb tests is not really as local anymore 

If it's too tricky to make these jobs separate I think we should update the documentation in the `README` to clarify how the localdb tests work?

PR #3124, opened 2020-03-26
Move API clients into separate packages

onursumer:
Separated out OncoKB and Genome Nexus API clients into their own packages. Now we can publish these API clients as standalone packages.

## Checks


packages/genome-nexus-ts-api-client/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "genome-nexus-ts-api-client",
+  "description": "Genome Nexus API Client for TypeScript",
+  "version": "0.1.0",
```
onursumer: This version should reflect the actual API version, ideally the latest tag version on genome-nexus master branch.

packages/oncokb-ts-api-client/package.json
```diff
@@ -0,0 +1,37 @@
+{
+  "name": "oncokb-ts-api-client",
+  "description": "OncoKB API Client for TypeScript",
+  "version": "0.1.0",
```
onursumer: This version should reflect the actual API version, ideally the latest tag version on oncokb master branch

package.json
```diff
@@ -1,6 +1,7 @@
 {
   "name": "cbioportal-frontend",
   "private": true,
+  "version": "3.2.13-alpha.0",
```
onursumer: added version back to make `updatePackageVersion` also update the root `package.json` dependencies. If the version is missing, lerna ignores this file completely when updating other package versions.

README.md
```diff
@@ -336,8 +336,6 @@ then update all package versions with:
 yarn run updatePackageVersion
 ```
 
-After that update the versions of all packages in the root `package.json`
```
onursumer: We don't need to do that anymore, instead we are adding version back to the root `package.json` and let `lerna version` handle dependencies automatically like it does all other versioned packages.

package.json
inodb: when do we need to update this? In every PR?

onursumer:
@inodb @alisman @zhx828 cBioPortal API still lives under `src/shared/api/generated`. Would it be useful to have a separate package for it as well?

inodb:
@onursumer potentially, I guess most times you would want to use the API + components. But maybe it's useful to allow API access as well without using mobx + components. If it's not extra maintenance effort then why not I guess

onursumer:
@inodb ideally root `package.json` should match the frontend tag on master. But, we don't really need to update package.json verison since it is a private package that we never publish. It is good enough to update the dependencies on the root `package.json`. 

We can even remove the version from the root `package.json`, but then we need to update dependencies manually since `lerna version` ignores the root package when there is no `version` field. It's okay if the root package.json version is out of sync, for now we don't really use that version anywhere.

package.json
onursumer: It is not critical to update this version every time. The main reason to have a version in this `package.json` is to make `lerna version` to properly update dependencies. It would be nice to keep this version in sync with the tagged master version though.

packages/oncokb-ts-api-client/package.json
onursumer: after discussing with @zhx828, decided to set the version to the current version in [swagger](https://www.oncokb.org/swagger-ui/index.html) doc, which is `1.0.0`

packages/genome-nexus-ts-api-client/package.json
onursumer: setting this to `1.1.2-beta.0` for now. latest master is 17 commits ahead of the latest release `1.1.1` so it is better set it to the next beta version.

package.json
```diff
@@ -131,7 +132,8 @@
     "bootstrap-sass": "3.3.7",
     "bowser": "^1.7.1",
     "bundle-loader": "^0.5.4",
-    "cbioportal-frontend-commons": "0.2.6",
+    "cbioportal-frontend-commons": "^0.2.7",
+    "cbioportal-ts-api-client": "^1.0.0-beta.0",
```
onursumer, line 136: Current API version in [swagger](https://www.cbioportal.org/api/swagger-ui.html) doc is 1.0 (beta)

PR #3126, opened 2020-03-27
genomic evolution tab - make app state shareable in link

adamabeshouse:


adamabeshouse:
@inodb let me know if you think we should do local storage as well or if this is enough

inodb:
Looks like screenshots are transposed, not sure if that was on purpose?

adamabeshouse:
@inodb accidentally made transpose default to true, updated

PR #3127, opened 2020-03-31
update genome nexus api

inodb:
Add nucleotide context endpoint


onursumer:
hmm, install is failing on circle CI

inodb:
@onursumer it passed first, then i updated frontend commons version (b/c of `check_missing_package_version_update` failure)  and it started crashing on install. Having a look. Do I need to run something after updating the package version?

inodb:
So when updating frontend commons one should first run

`yarn run updatePackageVersion`

this updates all packages, which makes sense b/c all of them rely on frontend commons

and then update the versions in the root `package.json`

We should prolly document this in the README somewhere

PR #3128, opened 2020-03-31
Add study name to quick search sample results

Luke-Sikina:
Fix cBioPortal/cbioportal#7316

- Added query store to quick search component
- Found matching study for each sample in store, added study name
  to search result display.

### DIFF
--- a/src/shared/components/query/QueryAndDownloadTabs.tsx
+++ b/src/shared/components/query/QueryAndDownloadTabs.tsx
@@ -139,7 +139,11 @@ export default class QueryAndDownloadTabs extends React.Component<
                         }}
                     >
                         <div>
-                            <QuickSearch />
+                            {this.store.cancerStudies.isComplete && (
+                                <QuickSearch
+                                    studies={this.store.cancerStudies.result}
+                                />
+                            )}
                         </div>
                     </MSKTab>
                     <MSKTab
--- a/src/shared/components/query/quickSearch/QuickSearch.tsx
+++ b/src/shared/components/query/quickSearch/QuickSearch.tsx
@@ -16,6 +16,8 @@ import { ServerConfigHelpers } from 'config/config';
 import sessionServiceClient from 'shared/api/sessionServiceInstance';
 import { trackEvent } from 'shared/lib/tracking';
 import { PagePath } from 'shared/enums/PagePaths';
+import { QueryStore } from '../QueryStore';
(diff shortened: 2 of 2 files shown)


src/shared/components/query/QueryAndDownloadTabs.tsx
```diff
@@ -139,7 +139,7 @@ export default class QueryAndDownloadTabs extends React.Component<
                         }}
                     >
                         <div>
-                            <QuickSearch />
+                            <QuickSearch queryStore={this.store} />
```
alisman: rather than making querystore a depedency of QuickSearch, lets just pass the study list in as a prop.    So you could test against isCompete out here and then pass in the resulting list.

src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -97,11 +102,21 @@ export default class QuickSearch extends React.Component {
         };
     }
 
+    @autobind
     private sampleToOption(sample: any, index: number): Option {
+        var studyName = 'Loading..';
+        if (this.props.queryStore.cancerStudies.isComplete) {
+            var study = this.props.queryStore.cancerStudies.result.find(
```
alisman: once you pass in studies as a prop, you can make a computed which keys the studies by their id.  That will make lookup faster.   Here we are potentially iterating through every study in list for each sample.   It's probably negligible but since the list gets rendered on keystroke ...

src/shared/components/query/quickSearch/QuickSearch.tsx
```diff
@@ -55,6 +61,16 @@ export default class QuickSearch extends React.Component {
     @observable private inputValue: string = '';
     @observable private isFocusing = false;
     @observable private isLoading = false;
+    private studies: Map<string, string>;
+
+    constructor(props: QuickSearchProps) {
+        super(props);
+
+        this.studies = props.studies.reduce((prev, curr) => {
```
alisman, line 69: This is a bit of an anti-pattern to do this in contructor.  If for some reason the property were to change, we would want this list to recompute.  By doing this in constructor, it would become stale.  In this particular case, it will never happen b/c studies are essentially static, but good to get it right.   So, this should just be a computed property

Luke-Sikina:
None of the e2e test failures are quick search related. I think this is ready for final review. @alisman

zhx828:
Thanks @Luke-Sikina !

PR #3130, opened 2020-04-01
Study view:fix gene table frequency in charts dropdown

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7349

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -4848,14 +4848,25 @@ export class StudyViewPageStore {
         default: [],
     });
 
+    readonly sampleUniqueKeysByMolecularProfileIdSet = remoteData<{
+        [id: string]: string[];
+    }>({
+        await: () => [this.selectedSamples, this.filteredGenePanelData],
+        invoke: async () => {
+            return getMolecularProfileSamplesSet(
+                this.selectedSamples.result,
+                this.filteredGenePanelData.result
+            );
+        },
+    });
+
     readonly molecularProfileSampleCounts = remoteData<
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4848,14 +4848,25 @@ export class StudyViewPageStore {
         default: [],
     });
 
+    readonly molecularProfileSamplesSet = remoteData<{
```
alisman, line 4851: i think we should improve the name of this and also the typing.    this is a problem in a lot of places but the name doesn't tell me what it is.     like, sampleIdsByMolecularProfileIdSet.   Also, is there a reason we don't just stick with samples themselves, e.g.     [profileid] : Sample[]?    I mean the only reason I can see is performance.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5046,6 +5055,58 @@ export class StudyViewPageStore {
                     },
                     {}
                 );
+                const molecularProfileSamplesSet = this
+                    .molecularProfileSamplesSet.result!;
+                if (!_.isEmpty(this.mutationProfiles.result)) {
+                    const uniqueKey = getUniqueKeyFromMolecularProfileIds(
```
alisman, line 5061: this little sum operation should have a comment explaining

src/pages/studyView/StudyViewPageStore.ts
kalletlak: yes, that's the only reason. And I think `sampleUniqueKeysByMolecularProfileIdSet`  would be more describing

PR #3131, opened 2020-04-02
Create a new utility package

onursumer:
Fix cbioportal/cbioportal/issues/7371
Related to knowledgesystems/msk-insight/issues/49

Changes:
- Move certain pure utility functions into a separate package for easier export in external projects. Current packages are not compatible with plain node applications due to some DOM dependencies.
- Make the mutation mapper data fetcher customizable to be able to tweak data source for specific data sources. This is useful, for example, when we want to replace an API call with a static data source.

## Checks


package.json
```diff
@@ -133,6 +133,7 @@
     "bundle-loader": "^0.5.4",
     "cbioportal-frontend-commons": "^0.2.7",
     "cbioportal-ts-api-client": "^0.9.0-beta.0",
+    "cbioportal-utils": "^0.1.0-beta.0",
```
onursumer, line 136: Not sure if this is the best name for the new package. Name suggestions are welcome!

packages/cbioportal-utils/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "test": "cross-env CI=1 react-scripts-ts test --env=jsdom",
+    "test:watch": "yarn run test -- --watch"
+  },
+  "peerDependencies": {
+    "mobx": "^3.0.0 || ^4.0.0 || ^5.0.0"
+  },
+  "dependencies": {
+    "genome-nexus-ts-api-client": "^1.1.2-beta.0",
+    "lodash": "^4.17.11",
+    "typescript": "3.2.2"
```
onursumer, line 37: we should update this to 3.8.3 after merging #3171

package.json
zhx828: Since we have cbioportal-frontend-commons, maybe cbioportal-frontend-utils?

onursumer: I was thinking that too, but this package is intended to be used by node scripts/backend, too. So, "frontend" in the package name might be misleading.

PR #3132, opened 2020-04-02
make study view dropdown not refresh unnecessarily when removing a chart

adamabeshouse:
image

### DIFF
--- a/end-to-end-test/remote/specs/core/studyview.spec.js
+++ b/end-to-end-test/remote/specs/core/studyview.spec.js
@@ -154,12 +154,13 @@ describe('study laml_tcga tests', () => {
             }
             browser.click(ADD_CHART_GENOMIC_TAB);
 
-            const chosenCheckbox = browser.elements(
-                '.addChartTabs .add-chart-option input'
-            ).value[0];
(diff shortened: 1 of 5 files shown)


alisman:
@adamabeshouse we still have situation on first load where we show options before frequency has loaded.  they are not in appropriate sorting order because it depends on frequency.  we need to ask product team whether this is desirable or whether we should just wait until frequency is loaded to show anything (@jjgao).  The only reason I can think to show unsorted is because frequency sometimes takes a long time and a user might want to just click something instead of waiting for frequency.

Also, When we click between tabs inside of the charts dropdown, even once all data is loaded, there is still a loader the flashes.  Can we investigate why?

alisman:
@adamabeshouse i think there's a broken e2e on this one:
chart_in_genomic_tab_can_be_updated - .add_chart

Also, if it's not too much trouble could you center the loading animation and text beneath it?  also maybe push it down 100px.

alisman:
@adamabeshouse also, make sure to rebase because @kalletlak made a fix (just merted) that might possible overlap with yours.  i don't think so, but possible.

PR #3133, opened 2020-04-03
Fix floating numbers while creating groups from bar chart

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7339


onursumer:
@zhx828 I guess we haven't

PR #3137, opened 2020-04-06
oncoprint - fix bug with comparison group track available counts

adamabeshouse:
before:
image

after:
image


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3294,6 +3300,20 @@ export class ResultsViewPageStore {
         []
     );
 
+    readonly sampleSet = remoteData({
+        await: () => [this.samples],
+        invoke: () => {
+            return Promise.resolve(
+                ComplexKeySet.from(
+                    this.samples.result!.map(s => ({
```
alisman, line 3308: couldn't you just have used the unique id?

adamabeshouse: the unique id is not stored in the comparison group object. just a list of study and sample ids

PR #3139, opened 2020-04-08
OncoKB Card improvement

zhx828:
- Get positional alteration matched when the exact match does not exist in alterations column in oncokb treatment table
- Do not sort oncokb card treatment table by default, use the default sorting returned from oncokb data

This fixes https://github.com/cBioPortal/cbioportal/issues/7365

### DIFF
--- a/packages/react-mutation-mapper/src/util/OncoKbUtils.ts
+++ b/packages/react-mutation-mapper/src/util/OncoKbUtils.ts
@@ -261,6 +261,29 @@ export function mergeAlterations(alterations: string | string[]) {
     return regular.join(', ');
 }
 
+/**
+ * Return the positional variant of the missense mutation
+ * @param alteration The missense mutation
+ */
+export function getPositionalVariant(alteration: string) {
+    const regExp = new RegExp('^([A-Z]+)([0-9]+)([A-Z]*)$');
+    const result = regExp.exec(alteration);
+
+    // Only if the alteration is missense mutation or positional variant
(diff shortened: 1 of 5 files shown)


packages/react-mutation-mapper/src/util/OncoKbUtils.ts
```diff
@@ -261,6 +261,19 @@ export function mergeAlterations(alterations: string | string[]) {
     return regular.join(', ');
 }
 
+export function getPositionalAlteration(alteration: string) {
+    const regExp = new RegExp('^(([A-Z]+)([0-9]+))([A-Z]*)$');
+    const result = regExp.exec(alteration);
+    if (
```
onursumer, line 267: maybe add a comment here what `result[2]`, `result[3]`, `result[4]` etc. correspond to? Maybe also add a brief comment on top of the function to summarize what we are doing here.

zhx828: @onursumer updated

alisman: not necessary but check out Array destructuring: https://medium.com/the-andela-way/this-trick-will-help-you-de-structure-javascript-arrays-in-a-better-cleaner-way-a3bbc39ce226

zhx828: @alisman this is really cool! I will bare this in mind and use next time.

PR #3144, opened 2020-04-11
Fully automate package versioning

onursumer:
- Remove `checkMissingPackageVersionUpdate` test in favor of auto versioning by lerna
- Add a Github action which does the following after each merge to master:
  - Look for changes in any of the packages
  - Update version and dependencies for changed packages, and push changes to master 
  - Create and push new tags. Tags are like `cbioportal-frontend-commons@0.2.8`, `cbioportal-frontend@3.2.13`, and `cbioportal-ts-api-client@1.0.0`.

Pros:
- Developer doesn't need to worry about bumping package versions in most cases

Cons:
- Auto generated release tags
- Github action that directly pushes to master

## Checks


.github/workflows/module-publish.yml
```diff
@@ -14,14 +14,18 @@ jobs:
   build:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
```
onursumer, line 19: we need the full commit history, otherwise `lerna version` doesn't work properly

scripts/update_package_version.sh
```diff
@@ -0,0 +1,30 @@
...
+git config --local user.email "LernaPublishBot@GithubActions.com"
+git config --local user.name "Lerna Publish Bot"
+git config --global github.token $GITHUB_TOKEN
+
+# lerna needs tags to be able to determine what's changed
+git fetch --tags
+
+# by default just bump to the next patch
+# so this will, for example, bump 1.0.0 to 1.0.1, 1.0.0-alpha.2 to 1.0.0, 0.1.1-beta.0 to 0.1.1, etc.
+lerna version patch --yes
```
onursumer, line 30: `lerna version` by default pushes directly to master. this might be an issue.

packages/cbioportal-ts-api-client/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "cbioportal-ts-api-client",
   "description": "cBioPortal API Client for TypeScript",
-  "version": "1.0.0-beta.1",
+  "version": "0.9.0-beta.0",
```
onursumer, line 4: This is to prevent lerna from auto publishing version `1.0.0`. When this PR gets merged, version `0.9.0` will be published instead.

onursumer:
@inodb Thanks for your comments! 

Github automatically updates releases once the tags are pushed to master. I think we need to modify the release drafter if possible to work only with the tags starting like `v`. Lerna by default always creates tags with the package name as a prefix. So if we can ignore package releases we are good. I'll look into that.

Regarding a separate repo, I have already created one to test lerna actions: https://github.com/onursumer/lerna-sandbox. Maybe we can add the release drafter there and see what happens?

And regarding pushing to master, I think the damage would be minimal even if something goes wrong. We can always manually fix package versions and delete tags/releases I guess :)

onursumer:
Manually pushed some release tags to (hopefully) prevent unnecessary republishing of certain packages.

onursumer:
failing e2e test is unrelated, merging.

PR #3145, opened 2020-04-13
Study View: Restrict api calls if there are no valid studies in query

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7381

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -386,7 +386,9 @@ export default class StudyViewPage extends React.Component<
                 {this.store.queriedSampleIdentifiers.isComplete &&
                     this.store.invalidSampleIds.isComplete &&
                     this.store.unknownQueriedIds.isComplete &&
-                    this.store.displayedStudies.isComplete && (
+                    this.store.displayedStudies.isComplete &&
+                    this.store.queriedPhysicalStudies.isComplete &&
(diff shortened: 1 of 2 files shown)


alisman:
What UI shows up in this case?   Do we have any idea how these invalid study ids came about?  
We might want to handle this with global error handler.   Lets discuss.

kalletlak:
It might be unauthorized studies. I don't have access to this study.

Screen Shot 2020-04-13 at 1 25 18 PM

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -386,7 +386,9 @@ export default class StudyViewPage extends React.Component<
                 {this.store.queriedSampleIdentifiers.isComplete &&
                     this.store.invalidSampleIds.isComplete &&
                     this.store.unknownQueriedIds.isComplete &&
-                    this.store.displayedStudies.isComplete && (
+                    this.store.displayedStudies.isComplete &&
+                    this.store.queriedPhysicalStudies.isComplete &&
```
alisman, line 390: i'm curious to know why this wouldn't be caught by unknownQueriedIds (above).  Maybe the study list service returns the ID even if user doesn't have access to it?

alisman:
I think this is ok.  Would sort of be nice to have more explicit validation routine so that reader can understand exactly what needs to be true for studies to be valid.

PR #3147, opened 2020-04-14
support grch38 in portal

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/7368

If the study is grch38, we will use grch38 genome nexus. 

We decide to not support mixed studies and the warning will be in another pr

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -33,11 +33,16 @@ import DiscreteCNACache from 'shared/cache/DiscreteCNACache';
 import {
     getDarwinUrl,
     getDigitalSlideArchiveMetaUrl,
+    getGenomeNexusHgvsgUrl,
 } from '../../../shared/api/urls';
 import PubMedCache from 'shared/cache/PubMedCache';
 import GenomeNexusCache from 'shared/cache/GenomeNexusCache';
 import GenomeNexusMutationAssessorCache from 'shared/cache/GenomeNexusMutationAssessorCache';
 import GenomeNexusMyVariantInfoCache from 'shared/cache/GenomeNexusMyVariantInfoCache';
+import {
+    GenomeNexusAPI,
+    GenomeNexusAPIInternal,
+} from 'genome-nexus-ts-api-client';
 import {
     getMyCancerGenomeData,
     ICivicGene,
@@ -87,6 +92,7 @@ import {
     mergeMutationsIncludingUncalled,
     noGenePanelUsed,
     ONCOKB_DEFAULT,
+    getGenomeNexusUrl,
 } from 'shared/lib/StoreUtils';
 import {
     fetchCivicGenes,
@@ -120,6 +126,7 @@ import { AppStore, SiteError } from 'AppStore';
 import { getGeneFilterDefault } from './PatientViewPageStoreUtil';
 import { checkNonProfiledGenesExist } from '../PatientViewPageUtils';
 import autobind from 'autobind-decorator';
+import { createVariantAnnotationsByMutationFetcher } from 'shared/components/mutationMapper/MutationMapperUtils';
 
 type PageMode = 'patient' | 'sample';
 
@@ -641,7 +648,8 @@ export class PatientViewPageStore {
                         this.uncalledMutationData
                     ),
                     ['annotation_summary', 'hotspots'],
-                    AppConfig.serverConfig.isoformOverrideSource
+                    AppConfig.serverConfig.isoformOverrideSource,
+                    this.genomeNexusClient
                 ),
             onError: (err: Error) => {
                 // fail silently, leave the error handling responsibility to the data consumer
@@ -655,7 +663,8 @@ export class PatientViewPageStore {
         invoke: async () => {
             return fetchHotspotsData(
                 this.mutationData,
-                this.uncalledMutationData
+                this.uncalledMutationData,
+                this.genomeNexusInternalClient
             );
         },
         onError: () => {
@@ -1516,15 +1525,30 @@ export class PatientViewPageStore {
     }
 
     @cached get genomeNexusCache() {
-        return new GenomeNexusCache();
+        return new GenomeNexusCache(
+            createVariantAnnotationsByMutationFetcher(
+                ['annotation_summary'],
+                this.genomeNexusClient
+            )
+        );
     }
 
     @cached get genomeNexusMyVariantInfoCache() {
-        return new GenomeNexusMyVariantInfoCache();
+        return new GenomeNexusMyVariantInfoCache(
+            createVariantAnnotationsByMutationFetcher(
+                ['my_variant_info'],
+                this.genomeNexusClient
+            )
+        );
     }
 
     @cached get genomeNexusMutationAssessorCache() {
-        return new GenomeNexusMutationAssessorCache();
+        return new GenomeNexusMutationAssessorCache(
+            createVariantAnnotationsByMutationFetcher(
+                ['annotation_summary', 'mutation_assessor'],
+                this.genomeNexusClient
+            )
+        );
     }
 
     @cached get pubMedCache() {
@@ -1623,4 +1647,24 @@ export class PatientViewPageStore {
         },
         []
     );
+
+    @computed get referenceGenomeBuild() {
+        if (!this.studies.isComplete) {
+            throw new Error('Failed to get studies');
+        }
+        return getGenomeNexusUrl(this.studies.result);
+    }
+
+    @autobind
+    generateGenomeNexusHgvsgUrl(hgvsg: string) {
+        return getGenomeNexusHgvsgUrl(hgvsg, this.referenceGenomeBuild);
+    }
+
+    @computed get genomeNexusClient() {
+        return new GenomeNexusAPI(this.referenceGenomeBuild);
+    }
+
+    @computed get genomeNexusInternalClient() {
+        return new GenomeNexusAPIInternal(this.referenceGenomeBuild);
+    }
 }
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -38,9 +38,14 @@ import {
     IOncoKbData,
     remoteData,
     stringListToSet,
+    getBrowserWindow,
 } from 'cbioportal-frontend-commons';
 import { getProteinPositionFromProteinChange } from 'cbioportal-utils';
-import { VariantAnnotation } from 'genome-nexus-ts-api-client';
+import {
+    VariantAnnotation,
+    GenomeNexusAPI,
+    GenomeNexusAPIInternal,
+} from 'genome-nexus-ts-api-client';
 import { CancerGene, IndicatorQueryResp } from 'oncokb-ts-api-client';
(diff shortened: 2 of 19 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4814,15 +4869,24 @@ export class ResultsViewPageStore {
      * For annotations of Genome Nexus we want to fetch lazily
      */
     @cached get genomeNexusCache() {
-        return new GenomeNexusCache();
+        return new GenomeNexusCache(
+            fetchFactory(['annotation_summary'], this.genomeNexusClient)
```
alisman, line 4873: this fetchFactory is actually very specific to fetchVariantAnnotationsByMutation.   but here in resultsviewpagestore, it would be very hard to understand that.  it sounds like a generic thing.  I would actually change the name fetchFactory to something very specific:    createVariantAnnotationsByMutationFetcher

alisman: i mean change it where it's defined

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -511,6 +520,14 @@ export class ResultsViewPageStore {
             },
             { fireImmediately: true }
         );
+
+        // this localStorage variable will be used to get relevant hgvsg url
+        if (this.grch38GenomeNexusUrl) {
+            getBrowserWindow().localStorage.setItem(
```
alisman, line 526: i forget, why are we puting this in localstorage?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1623,4 +1648,37 @@ export class PatientViewPageStore {
         },
         []
     );
+
+    // if we use grch37(default), grch38GenomeNexusUrl will be undefined
+    @computed get grch38GenomeNexusUrl() {
```
alisman, line 1653: This value is serving as the "source of truth" of the current build.  I don't that's abstract enough.  In other words, I'm a developer, i want to know what the current build is.  I wouldn't think to check whether grch38GenomeNexusUrl is undefined.  I would be looking for "currentReferenceGenemeBuild" or some such abstract property. It can contain the same logic here (based on study result).

src/pages/resultsView/ResultsViewPageStore.ts
leexgh: @alisman this comes from mutation mapper tool.
In mutation mapper tool, when people choose "GRCh37" or "GRCh38", we need to provide separate genome nexus instances. And in order to update genome nexus clients, we need to reload the page when people change their selection, so we need this variable to store their selection.
In this pr, I use this variable to create [hgvsg links](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/shared/api/urls.ts#L219-L224)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3187,6 +3206,45 @@ export class ResultsViewPageStore {
         []
     );
 
+    // if we use grch37(default), grch38GenomeNexusUrl will be undefined
+    @computed get grch38GenomeNexusUrl() {
```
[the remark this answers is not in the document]
leexgh: @onursumer where would be a good place to move to?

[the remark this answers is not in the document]
onursumer: We have `StoreUtils.ts`. We can maybe add a function there if we don't have a more specific utility file.

src/shared/lib/StoreUtils.ts
```diff
@@ -1251,3 +1252,25 @@ export async function getHierarchyData(
         sampleListId,
     });
 }
+
+export function currentReferenceGenemeBuild(studies: CancerStudy[]) {
```
onursumer, line 1256:
```suggestion
export function currentReferenceGenomeBuild(studies: CancerStudy[]) {
```

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
leexgh: updated

src/pages/resultsView/ResultsViewPageStore.ts
alisman: but this seems like something that belongs in the mutation mapper page.  It's particular to that page, not to the this store, right?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3187,6 +3210,25 @@ export class ResultsViewPageStore {
...
+    @computed get ensemblLink() {
+        return currentReferenceGenemeBuild(this.studies.result) ===
+            AppConfig.serverConfig.genomenexus_url_grch38
+            ? AppConfig.serverConfig.ensembl_transcript_grch38_url
+            : AppConfig.serverConfig.ensembl_transcript_url;
+    }
+
+    @computed get genomeNexusClient() {
+        return new GenomeNexusAPI(
+            currentReferenceGenemeBuild(this.studies.result)
```
alisman, line 3222: this should be in a computed property @computed referenceGenomeBuild() that references this.studies.

alisman: and then this will simple reference that computed value instead of calculating it here (and again below).  It's not the calculation is slow, it's that conceptually we want the source of truth to live in a place that is clear to everyone.

alisman: also, note the typo "GenemeBuild"

alisman:
@leexgh so, MutationTable component should have a property, buildWhateverLink:(blah)=>string;

This function will be called inside the column renderer for our column.   We define the function ... maybe in ReusltsViewMutationMapper which has access to the store.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -511,6 +520,17 @@ export class ResultsViewPageStore {
             },
             { fireImmediately: true }
         );
+
+        // this localStorage variable will be used to get relevant hgvsg url
```
alisman, line 524: can we get rid of this now?

src/shared/components/mutationTable/column/HgvsgColumnFormatter.tsx
```diff
@@ -2,21 +2,29 @@ import * as React from 'react';
...
+    private static getHgvsgDataViz(
+        mutation: Mutation,
+        generateGenomeNexusHgvsgUrl?: (hgvsg: string) => string
+    ) {
         let hgvsg = HgvsgColumnFormatter.getData(mutation);
         if (!hgvsg) {
             return <span />;
         } else {
-            let genomeNexusUrl = buildGenomeNexusHgvsgUrl(hgvsg);
+            const genomeNexusUrl = generateGenomeNexusHgvsgUrl!(hgvsg);
```
alisman, line 27: it doesn't seem like generateGenomeNexusHgvsgUrl is actually an optional argument.  It must be there or we'll get an error.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1623,4 +1647,21 @@ export class PatientViewPageStore {
         },
         []
     );
+
+    @computed get referenceGenomeBuild() {
+        return currentReferenceGenomeBuildUrl(this.studies.result);
```
alisman, line 1652: you should throw an error here if this.studies.isComplete === false.

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -62,7 +65,10 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
         >,
         public uniqueSampleKeyToTumorType: {
             [uniqueSampleKey: string]: string;
-        }
+        },
+        public generateGenomeNexusHgvsgUrl: (hgvsg: string) => string,
+        protected genomenexus_client?: GenomeNexusAPI,
+        protected genomenexus_internal_client?: GenomeNexusAPIInternal
```
onursumer, line 71: Let's use camelcase for variable naming:
```suggestion
        protected genomenexusClient?: GenomeNexusAPI,
        protected genomenexusInternalClient?: GenomeNexusAPIInternal
```

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -72,7 +78,9 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
             indexedHotspotData,
             indexedVariantAnnotations,
             oncoKbCancerGenes,
-            uniqueSampleKeyToTumorType
+            uniqueSampleKeyToTumorType,
+            genomenexus_client,
+            genomenexus_internal_client
```
onursumer, line 83:
```suggestion
            genomenexusClient,
            genomenexusInternalClient
```

src/shared/lib/referenceGenomeUtils.ts
```diff
@@ -1,4 +1,10 @@
...
-    GRCH37: 'GRCh37',
-    GRCH38: 'GRCh38',
+    GRCH37: {
+        NCBI: 'GRCh37',
+        UCSC: 'hg19',
+    },
+    GRCH38: {
+        NCBI: 'GRCh38',
+        UCSC: 'hg38',
+    },
```
onursumer, line 9: Might be better to name the constant all uppercase, and keys lowercase:
https://github.com/airbnb/javascript#naming--uppercase

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3192,6 +3204,33 @@ export class ResultsViewPageStore {
         []
     );
 
+    @computed get referenceGenomeBuild() {
+        if (this.studies.isComplete === false) {
```
onursumer, line 3208: Should we just do this instead?
```suggestion
        if (!this.studies.isComplete) {
```

src/shared/lib/referenceGenomeUtils.ts
leexgh: @onursumer  for those keys, using lowercase would be a good name from naming side, but I think`NCBI` and `UCSC` is the abbreviation of those organization names, so uppercase might make sense from semantic side. So in this case should we make good naming a priority?
Uppercase for constant sounds good to me though.

leexgh: @onursumer 
```
export const REFERENCE_GENOME = {
    grch37: {
        NCBI: 'GRCh37',
        UCSC: 'hg19',
    },
    grch38: {
        NCBI: 'GRCh38',
        UCSC: 'hg38',
    },
};
```
how do you like this?

src/shared/lib/StoreUtils.ts
```diff
@@ -1251,3 +1252,25 @@ export async function getHierarchyData(
         sampleListId,
     });
 }
+
+export function currentReferenceGenomeBuildUrl(studies: CancerStudy[]) {
```
inodb, line 1256: this is specific to genome nexus, so since this util function lives in a general namespace (not genome nexus specific), maybe good to use something like `currentGenomeNexusReferenceGenomeBuildUrl`

PR #3150, opened 2020-04-14
Dont include CNA gain or hetloss when querying DRIVER in OQL

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7376

When querying `EGFR: DRIVER`

From:

image

to:

Screen Shot 2020-04-16 at 9 47 48 AM


src/shared/lib/oql/oqlfilter.ts
```diff
@@ -1,21 +1,111 @@
 /* eslint camelcase: "off" */
```
inodb, line 1: @adamabeshouse what's the main change here - adding types (js -> ts)? Anything else?

src/shared/lib/oql/oqlfilter.d.ts
```diff
@@ -1,83 +0,0 @@
-//export function filterCBioPortalWebServiceData(oql_query, data, opt_default_oql, opt_by_oql_line, opt_mark_oql_regulation_direction) {
```
inodb, line 1: @adamabeshouse why can this be removed now?

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -400,30 +447,33 @@ var isDatumWantedByOQLAlterationCommand = function(alt_cmd, datum, accessors) {
...
             case 'DRIVER':
-            default:
-                datumWanted = isDatumWantedByOQLAlterationModifier(
-                    modifier,
-                    datum,
-                    accessors
-                );
+                const cnaValue = accessors.cna(datum);
+                datumWanted =
+                    cnaValue !== 'gain' && // dont include gains or hetloss here
```
adamabeshouse, line 470: only logic change is here

src/shared/lib/oql/oqlfilter.ts
adamabeshouse: that plus just the logic change I just indicated below

src/shared/lib/oql/oqlfilter.d.ts
adamabeshouse: we no longer need a type declaration file because the actual file is now in typescript. before, this served as like a typescript interface.

src/shared/lib/oql/AccessorsForOqlFilter.ts
```diff
@@ -107,7 +115,14 @@ export default class AccessorsForOqlFilter {
         );
     }
 
-    public gene(d: Mutation) {
+    private isMutation(d: Datum): d is Mutation {
+        return (
+            this.molecularAlterationType(d.molecularProfileId) ===
+            'MUTATION_EXTENDED'
```
alisman, line 121: Don't we have this in enum somewhere?

adamabeshouse: yes, good catch, I was just copy pasting code here

PR #3152, opened 2020-04-14
files & links tabs in patient view and study view

adamabeshouse:
Study view:
https://beta.cbioportal.org/study/filesAndLinks?id=ucec_tcga_pub and put `localStorage.netlify = "deploy-preview-3152--cbioportalfrontend"`

https://beta.cbioportal.org/patient/filesAndLinks?studyId=ucec_tcga_pub&caseId=TCGA-BK-A0CC and put `localStorage.netlify = "deploy-preview-3152--cbioportalfrontend"`
(ignore the duplicate tabs for Tissue Image and Pathology Report)

### DIFF
--- a/packages/cbioportal-frontend-commons/src/components/WindowWrapper.ts
+++ b/packages/cbioportal-frontend-commons/src/components/WindowWrapper.ts
@@ -32,4 +32,14 @@ export default class WindowWrapper {
             height: this.windowObj.innerHeight,
         };
     }
+
+    @autobind
+    public getWindowWidth() {
+        return this.size.width;
+    }
+
+    @autobind
+    public getWindowHeight() {
+        return this.size.height;
+    }
 }
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -276,12 +276,11 @@ export default class PatientViewPage extends React.Component<
         );
     }
 
-    private shouldShowResources(
-        patientViewPageStore: PatientViewPageStore
-    ): boolean {
-        if (patientViewPageStore.resourceIdToResourceData.isComplete) {
+    @computed
+    private get shouldShowResources(): boolean {
+        if (this.patientViewPageStore.resourceIdToResourceData.isComplete) {
             return _.some(
-                patientViewPageStore.resourceIdToResourceData.result,
+                this.patientViewPageStore.resourceIdToResourceData.result,
                 data => data.length > 0
             );
         } else {
@@ -289,16 +288,16 @@ export default class PatientViewPage extends React.Component<
         }
     }
 
-    private shouldShowPathologyReport(
-        patientViewPageStore: PatientViewPageStore
-    ): boolean {
+    @computed
+    private get shouldShowPathologyReport(): boolean {
         return (
-            patientViewPageStore.pathologyReport.isComplete &&
-            patientViewPageStore.pathologyReport.result.length > 0
+            this.patientViewPageStore.pathologyReport.isComplete &&
+            this.patientViewPageStore.pathologyReport.result.length > 0
         );
     }
 
-    hideTissueImageTab() {
+    @computed
+    private get hideTissueImageTab() {
         return (
             this.patientViewPageStore.hasTissueImageIFrameUrl.isPending ||
             this.patientViewPageStore.hasTissueImageIFrameUrl.isError ||
@@ -307,13 +306,12 @@ export default class PatientViewPage extends React.Component<
         );
     }
 
-    private shouldShowTrialMatch(
-        patientViewPageStore: PatientViewPageStore
-    ): boolean {
+    @computed
+    private get shouldShowTrialMatch(): boolean {
         return (
             getBrowserWindow().localStorage.trialmatch === 'true' &&
-            patientViewPageStore.detailedTrialMatches.isComplete &&
-            patientViewPageStore.detailedTrialMatches.result.length > 0
+            this.patientViewPageStore.detailedTrialMatches.isComplete &&
+            this.patientViewPageStore.detailedTrialMatches.result.length > 0
         );
     }
 
@@ -1438,11 +1436,7 @@ export default class PatientViewPage extends React.Component<
                                     key={4}
                                     id={PatientViewPageTabs.FilesAndLinks}
                                     linkText={RESOURCES_TAB_NAME}
-                                    hide={
-                                        !this.shouldShowResources(
-                                            this.patientViewPageStore
-                                        )
-                                    }
+                                    hide={!this.shouldShowResources}
                                 >
                                     <div>
                                         <ResourcesTab
@@ -1457,11 +1451,7 @@ export default class PatientViewPage extends React.Component<
                                     key={3}
                                     id={PatientViewPageTabs.PathologyReport}
                                     linkText="Pathology Report"
-                                    hide={
-                                        !this.shouldShowPathologyReport(
-                                            this.patientViewPageStore
-                                        )
-                                    }
+                                    hide={!this.shouldShowPathologyReport}
                                 >
                                     <div>
                                         <PathologyReport
@@ -1480,7 +1470,7 @@ export default class PatientViewPage extends React.Component<
                                     key={5}
                                     id={PatientViewPageTabs.TissueImage}
                                     linkText="Tissue Image"
-                                    hide={this.hideTissueImageTab()}
+                                    hide={this.hideTissueImageTab}
                                 >
                                     <div>
                                         <IFrameLoader
@@ -1520,9 +1510,7 @@ export default class PatientViewPage extends React.Component<
                                         </MSKTab>
                                     )}
 
-                                {this.shouldShowTrialMatch(
-                                    this.patientViewPageStore
-                                ) && (
+                                {this.shouldShowTrialMatch && (
                                     <MSKTab
                                         key={7}
                                         id={PatientViewPageTabs.TrialMatchTab}
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -1,7 +1,11 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import GenomicOverview from './genomicOverview/GenomicOverview';
-import { CancerStudy, ClinicalData, ResourceData } from 'cbioportal-ts-api-client';
+import {
+    CancerStudy,
+    ClinicalData,
+    ResourceData,
+} from 'cbioportal-ts-api-client';
 import {
     ClinicalDataBySampleId,
     RequestStatus,
@@ -838,6 +842,7 @@ export default class PatientViewPage extends React.Component<
                                     this.urlWrapper.setTab(id)
                                 }
                                 className="mainTabs"
+                                getPaginationWidth={WindowStore.getWindowWidth}
                             >
                                 <MSKTab
                                     key={0}
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import GenomicOverview from './genomicOverview/GenomicOverview';
-import { CancerStudy, ClinicalData } from 'cbioportal-ts-api-client';
+import { CancerStudy, ClinicalData, ResourceData } from 'cbioportal-ts-api-client';
 import {
     ClinicalDataBySampleId,
     RequestStatus,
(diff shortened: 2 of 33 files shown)


inodb:
@adamabeshouse @alisman Actually one thing I noticed: https://68401-66571349-gh.circle-artifacts.com/0/imageCompare.html. Looks like the borders around the tabs are removed, is that on purpose?

packages/cbioportal-frontend-commons/src/components/WindowWrapper.ts
```diff
@@ -32,4 +32,14 @@ export default class WindowWrapper {
             height: this.windowObj.innerHeight,
         };
     }
+
+    @autobind
+    public getWindowWidth() {
```
alisman, line 37: maybe make these computed getters since they reference observables?

alisman: i say this only so that it's clear to callers that they are getting observables

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -112,6 +124,13 @@ export default class PatientViewPage extends React.Component<
         );
         getBrowserWindow().patientViewPageStore = this.patientViewPageStore;
 
+        const openResourceId =
```
alisman, line 127: maybe add this as constructor parameter of store?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -257,6 +276,19 @@ export default class PatientViewPage extends React.Component<
         );
     }
 
+    private shouldShowResources(
```
alisman, line 279: why not computed getter?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -404,6 +436,72 @@ export default class PatientViewPage extends React.Component<
...
+        },
+    });
+
+    @autobind
+    @action
+    private openResource(resource: ResourceData) {
+        // open tab
+        this.patientViewPageStore.setResourceTabOpen(resource.resourceId, true);
+        // go to tab
+        this.urlWrapper.setTab(
```
alisman, line 481: i think this needs comment explaining why these two things are necessary (setTab and then setResourceUrl)

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -404,6 +436,72 @@ export default class PatientViewPage extends React.Component<
...
+    }
+
+    @autobind
+    @action
+    private closeResourceTab(tabId: string) {
+        // close tab
+        const resourceId = extractResourceIdFromTabId(tabId);
+        if (resourceId) {
+            this.patientViewPageStore.setResourceTabOpen(resourceId, false);
+            // go to resources tab if we're currently on that tab
```
alisman, line 495: need better comment explaining where we're going from here

src/pages/patientView/PatientViewUrlWrapper.ts
```diff
@@ -47,4 +51,15 @@ export default class PatientViewUrlWrapper extends URLWrapper<
     public setTab(tab: string): void {
         this.updateURL({}, `${PagePath.Patient}/${tab}`);
     }
+
+    @computed public get tabId() {
```
alisman, line 55: activeTabId? 
activeSecondaryTabId?

src/pages/patientView/PatientViewUrlWrapper.ts
```diff
@@ -47,4 +51,15 @@ export default class PatientViewUrlWrapper extends URLWrapper<
...
+        const regex = new RegExp(`${PagePath.Patient}\/(.+)`);
```
alisman, line 56: seems like a better strategy to split on slash and then use index.  That would hew better to the definition of the path protocol and also handle a future case where we might introduce deeper paths.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -435,6 +451,123 @@ export class PatientViewPageStore {
...
     );
 
+    readonly resourceDefinitions = remoteData({
+        invoke: () =>
+            client.getAllResourceDefinitionsInStudyUsingGET({
+                studyId: this.studyId,
+            }),
+        onResult: defs => {
+            if (defs) {
+                for (const def of defs)
```
alisman, line 461: might want to run this loop in runInAction.   (it's possible onResult is already an action but not sure).

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -435,6 +451,123 @@ export class PatientViewPageStore {
         []
     );
 
+    readonly resourceDefinitions = remoteData({
+        invoke: () =>
+            client.getAllResourceDefinitionsInStudyUsingGET({
+                studyId: this.studyId,
+            }),
+        onResult: defs => {
+            if (defs) {
```
alisman, line 460: lets add a comment on what this does.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -435,6 +451,123 @@ export class PatientViewPageStore {
...
+                    client
+                        .getAllStudyResourceDataInStudyUsingGET({
+                            studyId: this.studyId,
+                            resourceId: resource.resourceId,
+                            projection: 'DETAILED',
+                        })
+                        .then(data => ret.push(...data))
+                );
+            }
+            return Promise.all(promises).then(() => ret);
```
alisman, line 487: I think promise.all should already return an array of the results of promises in the order called?   you could then just do 
`Promise.all(studyResourceDefinision.map(()=>clintCalls)`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -435,6 +451,123 @@ export class PatientViewPageStore {
...
+                            })
+                            .then(data => {
+                                ret[sample.sampleId] =
+                                    ret[sample.sampleId] || [];
+                                ret[sample.sampleId].push(...data);
+                            })
+                    );
+                }
+            }
+            return Promise.all(promises).then(() => ret);
```
alisman, line 524: if what i said is true above ...

src/shared/components/resources/styles.module.scss
```diff
@@ -0,0 +1,25 @@
...
+            opacity: 0.7;
+        }
+
+        &:active {
+            opacity: 0.9;
+        }
+    }
+}
+
+:global(.cbioportal-frontend) h1.blackHeader,
```
alisman, line 18: lets but this in global.scss and also, i think it can be just .blackText.   I.e. not applied only to h1.    Possible it won't be strong enough and you would need to add h1....    You could also, try using existing bootstrap style .text-dark.  Again, not sure if it will overcome the h1 colors.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -194,10 +210,21 @@ export default class StudyViewPage extends React.Component<
             false,
             true
         );
+
+        this.toolbarLeftUpdater = setInterval(() => {
```
alisman, line 214: what's dis?

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -354,6 +419,42 @@ export default class StudyViewPage extends React.Component<
         );
     }
 
+    readonly openResourceTabs = MakeMobxView({
```
alisman, line 422: i would prefer just resourceTabs.  I read "open" as a verb.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -354,6 +419,42 @@ export default class StudyViewPage extends React.Component<
...
+        ],
+        render: () => {
+            const openDefinitions = this.store.resourceDefinitions.result!.filter(
+                d => this.store.isResourceTabOpen(d.resourceId)
+            );
+            const sorted = _.sortBy(openDefinitions, d => d.priority);
+            const resourceDataById = this.store.resourceIdToResourceData
+                .result!;
+
+            const tabs: JSX.Element[] = [];
```
alisman, line 435: i've been telling everyone to use reduce for these kind of filteredMap operations.  or do a filter and then a map.   i just think its more readable.   what do you think?  i know there are lots of these in this PR but i just decided to put comment down here.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -397,6 +402,17 @@ export class StudyViewPageStore {
         }
     }
 
+    private _isResourceTabOpen = observable.map<boolean>();
```
alisman, line 405: this sounds like boolean when it's map of boolean.  how about openResourceTabMap?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2854,6 +2866,76 @@ export class StudyViewPageStore {
...
+                        .then(data => {
+                            ret.push(...data);
+                        })
+                );
+            }
+            return Promise.all(promises).then(() => ret);
+        },
+        onResult: defs => {
+            if (defs) {
+                for (const def of defs)
```
alisman, line 2898: perhaps runInAction

src/pages/studyView/styles.scss
```diff
@@ -106,3 +106,16 @@ iframe.mdacc-heatmap-iframe {
     height: calc(100vh - 200px);
     min-height: 700px;
 }
+
+.resourcesTab {
+    .resourcesSection {
+        padding: 10px;
+        border-radius: 5px;
+        border: 1px dashed #ddd;
```
alisman, line 114: lets try to accomplish this with bordered chart.   If we don't want to call it bordered chart then we can use some sass extend/inherit sorcery to apply this styling to both border-chart and this.  see sass basics

packages/cbioportal-frontend-commons/src/components/WindowWrapper.ts
adamabeshouse: I specifically made them of this form for uses like this: https://github.com/cBioPortal/cbioportal-frontend/pull/3152/files#diff-b02249405d34f258fb2e74c6a0e83c73R845

users can use .size.width if they want a straight value. and I think if someone is already looking at this file then @computed doesnt add anything that the @observables above don't already

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: but then it would be the only parameter that we pass directly into the store rather than setting later

src/pages/patientView/PatientViewPage.tsx
adamabeshouse: just following the conventions. but they should all be computed actually

src/pages/patientView/PatientViewUrlWrapper.ts
adamabeshouse: not sure what you mean

src/shared/components/resources/OpenResourceTab.tsx
```diff
@@ -0,0 +1,166 @@
...
+export interface IOpenResourceTabProps {
+    resourceData: ResourceData[];
+    urlWrapper: {
+        setResourceUrl: (resourceUrl: string) => void;
+        query: { resourceUrl?: string };
+    };
+}
+
+@observer
+export default class OpenResourceTab extends React.Component<
```
alisman, line 22: Isn't it be definition "open" if we're rendering it?

src/pages/patientView/PatientViewUrlWrapper.ts
alisman: I mean, tabId is maybe not quite descritive enough.  Active tells me that this is the tab that is currently showing.  Sorry if I misundersatnd.

adamabeshouse: activeTabId makes sense. I dont get the secondary part though

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: looks like onResult is always run inside an action
https://github.com/adufilie/MobxPromise/blob/master/src/MobxPromise.ts#L176

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: its a flat list in the end. could be `return _.flatten(await Promise.all(...))` but idk if that's easier to read

src/pages/studyView/StudyViewPage.tsx
adamabeshouse: thats how i figure out how big to make the msk tabs. it keeps track of the leftmost coordinate of the study view toolbars

src/pages/studyView/styles.scss
adamabeshouse: good idea. see my new addition mixins.scss

src/pages/studyView/StudyViewPage.tsx
alisman: would it make any sense for this to be handled inside MSKTabs?  isn't it just getting position of itself?  or perhaps it gets passed a function as a prop which returns the element when invoked?

adamabeshouse: I feel like that's making MSKTabs too complicated for an unusual use case. I like just passing `width` to MSKTabs and handling that outside.

PR #3153, opened 2020-04-15
Run module publish action only on the main repo

onursumer:
This is to prevent auto versioning & publishing under forked repos

## Checks

### DIFF
--- a/.github/workflows/module-publish.yml
+++ b/.github/workflows/module-publish.yml
@@ -12,6 +12,8 @@ on:
 
 jobs:
   build:
+    # do not attempt versioning on forked repos, only trigger on the main repo
+    if: github.repository == 'cBioPortal/cbioportal-frontend'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2


scripts/update_package_version.sh
```diff
@@ -10,6 +10,12 @@
 # Prompt changed packages to manually set the new version
 
 if [[ "$GITHUB_RUN_ID" ]]; then
+  # do not attempt versioning if the NODE_AUTH_TOKEN is missing
+  if [[ -z "$NODE_AUTH_TOKEN" ]]; then
```
zhx828, line 14: Just a note: this is under assumption that no one has NODE_AUTH_TOKEN except the people actually wants to publish it.

onursumer: Seems like there is some sort of placeholder for `NODE_AUTH_TOKEN` under forked repos as well. So, this check does NOT work. We need to come up with a better solution.

onursumer: Now using a job level conditional within the workflow yaml file

PR #3154, opened 2020-04-15
Show timeline in genomic evolution tab

adamabeshouse:
https://deploy-preview-3154--cbioportalfrontend.netlify.app/patient/genomicEvolution?caseId=P04&genomicEvolutionSettings=%7B%22clusterHeatmap%22%3A%22false%22%7D&studyId=lgg_ucsf_2014&tab=summaryTab

image

image


inodb:
Thanks @adamabeshouse! Now that I see it in action I think actually mabye better to have the timeline at the top, because the selection controls (i.e. mutation table) or kinda far now from the vaf/heatmap plot. When it's at the top prolly better to hide by default. Would be nice if to state of hide/show is kept when clicking between patients, so maybe some URL parameter again

inodb:
@adamabeshouse thanks! Looks great!

Two minor things:

- [ ] the timeline button bar is is a bit large so on my macbook screen I only see the first two lines of the mutations table. Not essential to fix for this PR tho imo
- [ ] there is something off with the left padding when making the window size smal. I guess the timeline is centered or something? On the summary view it goes to the left, but on genomic evolution tab it shows a border on the left:

    Screen Shot 2020-04-16 at 4 11 53 PM

tmazor:
This is awesome! One request - can the "Show/Hide Timeline" button be on the left side of the page above the Line Chart/Heatmap tabs rather than centered? I feel like it's just floating all by itself out there

image

src/pages/patientView/mutation/PatientViewMutationsTab.tsx
```diff
@@ -46,6 +50,30 @@ export default class PatientViewMutationsTab extends React.Component<
...
 > {
+    get showTimeline() {
+        return (
+            this.props.urlWrapper.query.genomicEvolutionSettings
+                .showTimeline === 'true'
+        );
+    }
+    set showTimeline(o: boolean) {
+        this.props.urlWrapper.updateURL({
+            genomicEvolutionSettings: Object.assign(
```
alisman, line 61: i feel like there's got to a nicer way to do this.  it's basically the old immutable issue.  I think m.westtrate has some library which handles stuff like this.  we should look into it.  not necessary for this PR though

alisman:
The button for show timeline is easily lost.  Since it takes up vertical space either way, what if we surround it with a chart border like so:

image

adamabeshouse:
updated @alisman @tmazor 

image

image

src/pages/patientView/mutation/PatientViewMutationsTab.tsx
adamabeshouse: ill do this in a different PR

tmazor:
Looks good to me @adamabeshouse !

alisman:
@adamabeshouse this is REALLY picky, but I don't like how the show timeline button moves position when click it open.  seems to occur because margin/padding is different around the lone button than around the open state.    
Also, I would like to make the button style btn-default

image

adamabeshouse:
@alisman just pushed changes addressing your comment

jjgao:
Looks good! 

One thought: should we try to adopt the same style for Line Chart and HeatMap too? ie. three buttons to toggle, `Timeline`,  `Line Chart` and `HeatMap`.

PR #3155, opened 2020-04-16
Study view: Add Case Lists table

kalletlak:
Frontend changes for https://github.com/cBioPortal/cbioportal/issues/7358

Example
This chart is not show by default. It should be checked from charts dropdown
Screen Shot 2020-04-16 at 1 47 58 PM

Screen Shot 2020-04-16 at 12 47 12 PM

To test
https://cbioportal-fix-7358-fwzkhgccnq.herokuapp.com/ and set `localStorage.netlify = "deploy-preview-3155--cbioportalfrontend"`


jjgao:
Thanks, @kalletlak! It looks great!

@schultzn @tmazor could you also take a look?

I think it's probably a good idea that the Case List table is currently hidden by default, because most of the case lists are related to data availability which is already covered by the profiles table.

I think for the next step we may want to consider redesigning/consolidating these concepts: case list, profile availability, groups, and clinical attributes.

tmazor:
I can't seem to load it? I just see this:
image

kalletlak:
@tmazor I had to close that pr. Here is the new test site
https://cbioportal-fix-7358-fwzkhgccnq.herokuapp.com/ and set `localStorage.netlify = "deploy-preview-3155--cbioportalfrontend"`

PR #3156, opened 2020-04-16
Refine global/local http error handling

alisman:
We have thus far handled errors in mobxpromise.  This works well only if the developer does in fact handle the error condition in UI.  It doesn't work well for global error handling.  This PR forces developer to choose when actually calling the API client or using http client (superagent) directly.  

There are three ways of error handling:
1. local: errors fail silently unless dealt with in UI (e.g. annotation services we show error icon in table cell)
2. global: dialog.  We have an error which is serious but not catastrophic and we still want to show normal UI but with a dimissiable error dialog (or alert?).
3. global: screen.  We want to hide the site entirely and render only an error screen

### DIFF
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -1,39 +1,18 @@
 import { action, computed, observable } from 'mobx';
-import {
-    addServiceErrorHandler,
-    getBrowserWindow,
-    remoteData,
-} from 'cbioportal-frontend-commons';
-import { initializeAPIClients } from './config/config';
+import { remoteData } from 'cbioportal-frontend-commons';
 import * as _ from 'lodash';
 import internalClient from 'shared/api/cbioportalInternalClientInstance';
-import { sendSentryMessage } from './shared/lib/tracking';
+import autobind from 'autobind-decorator';
 
 export type SiteError = {
     errorObj: any;
     dismissed: boolean;
     title?: string;
+    mode?: 'dialog' | 'screen';
+    customMessage?: string;
 };
 
 export class AppStore {
-    constructor() {
-        getBrowserWindow().me = this;
-        addServiceErrorHandler((error: any) => {
-            try {
-                sendSentryMessage('ERRORHANDLER:' + error);
-            } catch (ex) {}
-
-            if (
-                error.status &&
-                parseInt(error.status) &&
-                parseInt(error.status) >= 400
-            ) {
-                sendSentryMessage('ERROR DIALOG SHOWN:' + error);
-                this.siteErrors.push({ errorObj: error, dismissed: false });
-            }
-        });
-    }
-
     @observable private _appReady = false;
 
     @observable siteErrors: SiteError[] = [];
@@ -54,12 +33,21 @@ export class AppStore {
         }
     }
 
-    @computed get undismissedSiteErrors() {
+    @computed get undismissedSiteErrors(): SiteError[] {
         return _.filter(this.siteErrors.slice(), err => !err.dismissed);
     }
 
     @computed get isErrorCondition() {
+        // does it have at least one error of type screen (total failure as opposed to recoverable
+        // which we will represent as dialog
         return this.undismissedSiteErrors.length > 0;
+    }
+
+    @computed get dialogErrors() {
+        return _.filter(
+            () => this.undismissedSiteErrors,
+            (err: SiteError) => err.mode === 'dialog'
+        );
     }
 
     @action
@@ -73,6 +61,18 @@ export class AppStore {
     @action
     public setAppReady() {
         this._appReady = true;
+    }
+
+    @autobind
+    handleServiceError(error: SiteError): void {
+        if (
+            error.errorObj.status &&
+            parseInt(error.errorObj.status) &&
+            parseInt(error.errorObj.status) >= 400
+        ) {
+            error.dismissed = false;
+            this.siteErrors.push(error);
+        }
     }
 
     public get appReady() {
--- a/src/AppStore.ts
+++ b/src/AppStore.ts
@@ -23,7 +23,11 @@ export class AppStore {
                 sendSentryMessage('ERRORHANDLER:' + error);
             } catch (ex) {}
 
-            if (error.status && /400|500|403/.test(error.status)) {
+            if (
+                error.status &&
+                parseInt(error.status) &&
+                parseInt(error.status) >= 400
+            ) {
                 sendSentryMessage('ERROR DIALOG SHOWN:' + error);
                 this.siteErrors.push({ errorObj: error, dismissed: false });
             }
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -17,6 +17,8 @@ import {
 import { buildCBioPortalPageUrl } from 'shared/api/urls';
 import ErrorScreen from 'shared/components/errorScreen/ErrorScreen';
 import { ServerConfigHelpers } from 'config/config';
+import { AppStore } from 'AppStore';
+import { Modal } from 'react-bootstrap';
 
 interface IContainerProps {
     location: Location;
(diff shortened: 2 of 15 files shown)


src/AppStore.ts
```diff
@@ -71,6 +63,18 @@ export class AppStore {
         this._appReady = true;
     }
 
+    @autobind
+    handleServiceError(error: SiteError): void {
+        if (
+            error.errorObj.status &&
+            parseInt(error.errorObj.status) &&
+            parseInt(error.errorObj.status) >= 400
```
alisman, line 71: this probably needs refinement

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -267,17 +267,19 @@ export default class ClinicalData extends React.Component<
...
-                        : 'SAMPLE',
-                    clinicalDataMultiStudyFilter: {
-                        attributeIds: [attribute.clinicalAttributeId],
-                        identifiers: attribute.patientAttribute
-                            ? patientidentifiers
-                            : sampleIdentifiers,
-                    },
-                });
+                let clinicalData = await client
+                    .handleErrorsGlobally()
```
alisman, line 271: i'm torn whether to make global handling the default and allow people just to use the client directly (client.fetchClincal ...) or actually make them write out the method, which woudl force them to think about it.

src/AppStore.ts
```diff
@@ -50,14 +33,23 @@ export class AppStore {
         }
     }
 
-    @computed get undismissedSiteErrors() {
+    @computed get undismissedSiteErrors(): SiteError[] {
         return _.filter(this.siteErrors.slice(), err => !err.dismissed);
     }
 
     @computed get isErrorCondition() {
```
adamabeshouse, line 40: as per your edit, maybe lets call this `isFatalError` or such

src/AppStore.ts
```diff
@@ -50,14 +33,23 @@ export class AppStore {
...
+        // does it have at least one error of type screen (total failure as opposed to recoverable
+        // which we will represent as dialog
         return this.undismissedSiteErrors.length > 0;
     }
 
+    @computed get dialogErrors() {
+        return _.filter(
```
adamabeshouse, line 47: small thing but `this.undismissedSiteErrors.filter(...)` is more direct to read, IMO

src/AppStore.ts
adamabeshouse: I'm not sure I understand this condition
```
parseInt(error.errorObj.status) &&
parseInt(error.errorObj.status) >= 400
```

would it not be `error.errorObj.status && parseInt(..)>=400`? otherwise `parseInt(.) && parseInt(..)>=400` is redundant?

src/appShell/App/Container.tsx
```diff
@@ -104,37 +106,33 @@ export default class Container extends React.Component<IContainerProps, {}> {
...
+                            </>
+                        </Else>
+                    </If>
+                </div>
             </div>
         );
     }
 }
+
+function showSiteErrors(appStore: AppStore) {
```
adamabeshouse, line 126: lets rename this `showFatalErrors`? in general site error->fatal error, to distinguish from "dialog errors"?

src/pages/groupComparison/ClinicalData.tsx
adamabeshouse: i think global handling/fatal error is a good default. its more unusual for an error to be tolerable

adamabeshouse: as discussed, I think it should be mandatory to choose an error handling method

src/shared/lib/extendCBioPortalAPI.ts
```diff
@@ -0,0 +1,50 @@
...
+            return this;
+        }
+
+        constructor(...args: any[]) {
+            super(...args);
+        }
+    };
+}
+
+class Nada {}
```
adamabeshouse, line 50: ?

src/shared/lib/extendCBioPortalAPI.ts
```diff
@@ -0,0 +1,50 @@
...
+            };
+            return self;
+        }
+
+        handleErrorsLocally() {
+            return this;
+        }
+
+        constructor(...args: any[]) {
+            super(...args);
```
adamabeshouse, line 45: unecessary constructor?

src/shared/lib/extendCBioPortalAPI.ts
```diff
@@ -0,0 +1,50 @@
...
+
+            function restoreRequest() {
+                // @ts-ignore (it's private, but we CAN access it)
+                self.request = proxiedRequest;
+            }
+
+            // @ts-ignore (it's private, but we CAN access it)
+            self.request = function() {
+                const args = Array.from(arguments);
+                if (args.length >= 9) {
```
adamabeshouse, line 23: whats going on here?

src/shared/lib/extendCBioPortalAPI.ts
```diff
@@ -0,0 +1,50 @@
...
+                        function(error: any) {
+                            errorConfig.errorObj = error;
+                            getBrowserWindow().globalStores.appStore.handleServiceError(
+                                errorConfig as SiteError
+                            );
+                        },
+                    ]);
+                }
+                // @ts-ignore
+                const res = proxiedRequest(...args);
```
adamabeshouse, line 34: whats `res` for?

src/shared/lib/SuperAgentWrapped.ts
```diff
@@ -0,0 +1,39 @@
+import { Request } from 'superagent';
+import request from 'superagent';
+import { SiteError } from 'AppStore';
+import { getBrowserWindow } from 'cbioportal-frontend-commons';
+
+class SuperAgentClient {
```
adamabeshouse, line 6: this file could use a few more comments

src/pages/groupComparison/ClinicalData.tsx
onursumer: I think it is okay to force the developer. We should think about every new API call we add to the application anyway.

src/AppStore.ts
alisman: It's not necessarily a fatal error.  It could be a dialog error (any i may introduce an alert error also) which is dimissiable and allows user to continue to use site

src/AppStore.ts
alisman: actualy this code is defunct. removing

src/AppStore.ts
alisman: i think i didn't want to do comparison on what will be NaN (parseInt(undefind)).  It turns out to work.

adamabeshouse: it's a little hard to read. can you do `!isNaN(error.errorObj.status)`?

src/shared/lib/extendCBioPortalAPI.ts
alisman: I will change the name but the reason for this is that if and when we decide we don't want people calling methods directly on client, we can inherit from this instead.  I'll get rid of it for now.

src/shared/lib/extendCBioPortalAPI.ts
```diff
@@ -0,0 +1,50 @@
...
+            // @ts-ignore (it's private, but we CAN access it)
+            const proxiedRequest = this.request;
+
+            function restoreRequest() {
+                // @ts-ignore (it's private, but we CAN access it)
+                self.request = proxiedRequest;
+            }
+
+            // @ts-ignore (it's private, but we CAN access it)
+            self.request = function() {
```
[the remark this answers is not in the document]
alisman: so, superagent makes requests by invoking the client's instance of request.  I need to monkey with that request in order to insert an error handler.  But I only want to do this on an individual method call basis.  So, as soon as the call is made, I restore the request to what it was so that any future requests won't be affected.     I don't love this but after a lot of tinkering, it was the only way I could figure out to get the dev interface I wanted.  I could also have proxied all methods on the client and added another parameter, but it would half been harder i think from typescript standpoint.

src/shared/lib/oql/oql-parser.js
```diff
@@ -4,3031 +4,2561 @@
  * http://pegjs.org/
  */
 
-'use strict';
+"use strict";
```
onursumer, line 7: we probably don't need to regenerate this one, do we?

src/shared/lib/extendCBioPortalAPI.ts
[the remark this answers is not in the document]
onursumer: It might be possible to inject code using decorators instead. I haven't done much research on that stuff though, maybe it is not worth the effort.

PR #3158, opened 2020-04-17
URLWrapper.updateURL gets a more powerful interface, and use it to clean up nested object updates

adamabeshouse:


src/shared/lib/URLWrapper.ts
```diff
@@ -221,16 +221,25 @@ export default class URLWrapper<
...
+        _updatedParams:
+            | Partial<QueryParamsType>
+            | ((currentQuery: QueryParamsType) => Partial<QueryParamsType>),
         path: string | undefined = undefined,
         clear = false,
         replace = false
     ) {
+        const updatedParams: Partial<QueryParamsType> =
+            typeof _updatedParams === 'function'
+                ? _updatedParams(JSON.parse(JSON.stringify(this.query)))
```
alisman, line 233: why not use mobx's toJS (with recurse everything on)?

src/shared/lib/URLWrapper.ts
```diff
@@ -221,16 +221,25 @@ export default class URLWrapper<
...
+                : _updatedParams;
+
         // get the names of the props that are designated as session props
         const sessionProps = this.getSessionProps();
 
         // overwrite params onto the existing params
-        const mergedParams = _.assign({}, this.query, updatedParams);
+        const mergedParams = clear
```
alisman, line 240: hmm. so nothing was being done with clear before?

alisman:
@adamabeshouse, this looks good.    i think you and i know should go over the relationship between urlWrapper.query and the actual url just to make we understand it.  And also document it.

src/shared/lib/URLWrapper.ts
adamabeshouse: right, if there was a session involved

PR #3162, opened 2020-04-20
Dont hide query genes in mrna and protein enrichments

adamabeshouse:


adamabeshouse:
now, should we determine what to hide in mutation/cna enrichments based on which groups are selected?
@jjgao @kalletlak @schultzn

PR #3167, opened 2020-04-22
Oncoprint: do not unnecessarily re-render clinical dropdown, losing scroll position

adamabeshouse:
fixes #7426

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -760,6 +760,28 @@ export class ResultsViewPageStore {
         },
     });
 
+    @computed.struct get comparisonGroupsReferencedInURL() {
+        const clinicalTracksParam = this.urlWrapper.query.clinicallist;
+        if (clinicalTracksParam) {
+            const groupIds = clinicalTracksParam
+                .split(',') // split by comma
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -760,6 +760,28 @@ export class ResultsViewPageStore {
         },
     });
 
+    @computed.struct get comparisonGroupsReferencedInURL() {
```
alisman, line 763: hey, can you explain struct to me?  i looked it up.   it says "ignore new values that are structurally equal" but i don't quite understand what structurally equal means?

alisman:
Curious why the study view files are involved

adamabeshouse:
@alisman the study view components are used in the AddClinicalTracks component - they're just the tabs of the dropdown.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: somethings lost in translation there but it means deep equal

alisman: got it.  thanks.   i guess struct comes from structurally equal

PR #3171, opened 2020-04-23
Update typescript and fix arising type errors

adamabeshouse:
Fix #7432


onursumer:
For consistency, we should also update TS version for all the packages. Seems like module build is failing anyway.

PR #3173, opened 2020-04-24
Study view: stop unnecessary api call when querying virtual study

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7407


alisman:
@kalletlak we're still calling the  resource definition service with the vs id:
image

kalletlak:
@alisman updated pr to fix it

PR #3176, opened 2020-04-24
Dont show plots treatment data error before options are loaded

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/7428


dippindots:
@adamabeshouse Thanks for fixing this!

PR #3179, opened 2020-04-29
filters + 

Luke-Sikina:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6034,4 +6134,117 @@ export class StudyViewPageStore {
...
+            _.isArray(filters.sampleTreatmentFilters.filters)
+        ) {
+            this.setSampleTreatmentFilters(filters.sampleTreatmentFilters);
+        }
+    }
+
+    // a row represents a list of patients that either have or have not recieved
+    // a specific treatment
+    @computed
+    public get sampleTreatments() {
```
kalletlak, line 6207: any specific reason to get create this promise with computed. why not create it directly

PR #3181, opened 2020-04-29
Filter samples having molecular profiles while requesting for enrichment data

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7451

### DIFF
--- a/src/shared/lib/comparison/ComparisonStore.ts
+++ b/src/shared/lib/comparison/ComparisonStore.ts
@@ -12,6 +12,7 @@ import {
     IOverlapComputations,
     isGroupEmpty,
     partitionCasesByGroupMembership,
+    EnrichmentAnalysisComparisonGroup,
 } from '../../../pages/groupComparison/GroupComparisonUtils';
 import { GroupComparisonTab } from '../../../pages/groupComparison/GroupComparisonTabs';
 import { remoteData, stringListToIndexSet } from 'cbioportal-frontend-commons';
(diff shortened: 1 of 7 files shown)


kalletlak:
@adamabeshouse issue is some of the studies doesn't have particular molecular profiles(mutation, cna, etc.) in multi-study analysis. Solution is to remove those sample which doesn't have molecular profile

adamabeshouse:
@kalletlak ah okay thanks. can you write that down in comments?

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -557,100 +558,157 @@ export default class ComparisonStore {
...
+        },
+    });
+
+    readonly mutationEnrichmentDataRequestGroups = remoteData({
+        await: () => [
+            this.mutationEnrichmentAnalysisGroups,
+            this.selectedStudyMutationEnrichmentProfileMap,
+        ],
+        invoke: async () => {
+            let groups: MolecularProfileCasesGroupFilter[] = _.chain(
```
alisman, line 600: chaining with these big statements makes things kind of hard to read.  granted prettier tends to bloat everything vertically.   
i would either separate the filter statement ```const nonEmptyGroups = groups.filter(..)``` or use reduce to accomplish the filtering

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -557,100 +558,157 @@ export default class ComparisonStore {
...
     });
 
+    readonly copyNumberEnrichmentAnalysisGroups = remoteData({
+        await: () => [
+            this.selectedStudyCopyNumberEnrichmentProfileMap,
+            this.enrichmentAnalysisGroups,
+        ],
+        invoke: () => {
+            return Promise.resolve(
+                _.chain(this.enrichmentAnalysisGroups.result)
```
alisman, line 657: ditto

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -557,100 +558,157 @@ export default class ComparisonStore {
...
-                                                .molecularProfileId,
-                                    };
-                                }
-                            );
-                        }
+            let groups: MolecularProfileCasesGroupFilter[] = _.chain(
+                this.copyNumberEnrichmentAnalysisGroups.result
+            )
+                .map(group => {
+                    const molecularProfileCaseIdentifiers = group.samples.map(
```
alisman, line 693: ditto

src/shared/lib/comparison/ComparisonStore.ts
alisman: fyi, we don't need lodash here because map, reduce, filter are available natively on array now (and chaineable)

PR #3182, opened 2020-04-29
Allow cbioportal use oncokb biological data without token

zhx828:
This is a part of fix for https://github.com/oncokb/oncokb/issues/1975

- we no longer require a token to display biological data from oncokb(it connects to public.api.oncokb.org)
- we allow users to configure the cbioportal instance to use www.oncokb.org to include clinical data
- Levels column will be disabled for mutation mapper annotation track

Here is the tooltip when portal connects to the public instance:
Screen Shot 2020-04-29 at 7 51 54 PM

### DIFF
--- a/packages/react-mutation-mapper/src/component/column/Annotation.tsx
+++ b/packages/react-mutation-mapper/src/component/column/Annotation.tsx
@@ -29,6 +29,7 @@ import OncoKB, { sortValue as oncoKbSortValue } from '../oncokb/OncoKB';
 import HotspotAnnotation, {
     sortValue as hotspotSortValue,
 } from './HotspotAnnotation';
+import { USE_DEFAULT_PUBLIC_INSTANCE_FOR_ONCOKB } from '../../util/DataFetcherUtils';
 
 export type AnnotationProps = {
     mutation?: Mutation;
@@ -39,6 +40,7 @@ export type AnnotationProps = {
     hotspotData?: RemoteData<IHotspotIndex | undefined>;
     oncoKbData?: RemoteData<IOncoKbData | Error | undefined>;
     oncoKbCancerGenes?: RemoteData<CancerGene[] | Error | undefined>;
+    usingPublicOncoKbInstance: boolean;
     pubMedCache?: MobxCache;
     resolveEntrezGeneId?: (mutation: Mutation) => number;
     resolveTumorType?: (mutation: Mutation) => string;
@@ -66,6 +68,7 @@ export interface IAnnotation {
     oncoKbStatus: 'pending' | 'error' | 'complete';
     oncoKbGeneExist: boolean;
     isOncoKbCancerGene: boolean;
+    usingPublicOncoKbInstance: boolean;
     myCancerGenomeLinks: string[];
     civicEntry?: ICivicEntry | null;
     civicStatus: 'pending' | 'error' | 'complete';
@@ -77,6 +80,7 @@ export const DEFAULT_ANNOTATION_DATA: IAnnotation = {
     oncoKbStatus: 'complete',
     oncoKbGeneExist: false,
     isOncoKbCancerGene: false,
+    usingPublicOncoKbInstance: false,
     isHotspot: false,
     is3dHotspot: false,
     hotspotStatus: 'complete',
@@ -100,6 +104,7 @@ export function getAnnotationData(
     hotspotData?: RemoteData<IHotspotIndex | undefined>,
     myCancerGenomeData?: IMyCancerGenomeData,
     oncoKbData?: RemoteData<IOncoKbData | Error | undefined>,
+    usingPublicOncoKbInstance?: boolean,
     civicGenes?: RemoteData<ICivicGene | undefined>,
     civicVariants?: RemoteData<ICivicVariant | undefined>,
     resolveTumorType: (mutation: Mutation) => string = getDefaultTumorType,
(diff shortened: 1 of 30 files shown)


packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
```diff
@@ -239,6 +240,13 @@ export default class MutationMapper<
             : getDefaultWindowInstance();
     }
 
+    @computed
+    get usingOncoKbPublicInstance() {
+        return this.store.oncoKbInfo.result
+            ? this.store.oncoKbInfo.result.publicInstance
+            : ONCOKB_DEFAULT_INFO_PUBLIC_INSTANCE;
```
onursumer, line 247: Isn't it possible here to just do:
```suggestion
        return this.store.usingOncoKbPublicInstance;
```

zhx828: Oh good catch. This is must been there before I wrote the method in store

alisman: i feel like it's odd to have something in all caps that is not a string or number constant.

onursumer: Hmm, I never thought about restricting all caps to string and number constants. We already have similar default values in `react-mutation-mapper` though.

onursumer: According to airbnb javascript style guideline, it seems okay to have all caps for other types of constants as well: https://github.com/airbnb/javascript/#naming--uppercase

onursumer: Oh, `ONCOKB_DEFAULT_INFO_PUBLIC_INSTANCE` is a boolean variable: https://github.com/cBioPortal/cbioportal-frontend/pull/3182/files#diff-c27e93a0aa919b85e6be54dc7efb1c85R33
Having the name `instance` for a boolean variable is a bit confusing here.

packages/react-mutation-mapper/src/component/oncokb/OncoKbTooltip.tsx
```diff
@@ -14,6 +14,7 @@ export interface IOncoKbTooltipProps {
     hugoSymbol: string;
     isCancerGene: boolean;
     geneNotExist: boolean;
+    usingPublicInstance: boolean;
```
alisman, line 17: lets keek this usingPublicOncoKBInstance

alisman:
@zhx828 I'm curious why we have to call the info service to find out whether we're using public instance or now.  

Code looks fine.  I do wonder if this bit of state (using public instance) or not is kind of a global state and could be read directly off the stores (which have reference to the appstore) by components rather than passed down to them though the component heirarchy.   The downside of this might be that "leaf" components would need reference to the stores, but perhaps they already have it?  If so, then maybe this could be accomplished with less code change.  Just an idea.

zhx828:
@alisman thanks for the suggestion. There are two ways of implementing this as I know.
- On the backend, similar to oncokbToken, we could return the status of the instance in the serverConfig. But that status needs to be managed once the application is alive since it still needs to reach out to the info endpoint. The timing is tricky. At the moment, we only configure portal instance with few oncokb properties, we do not manage any actually oncokb data on the backend. Thought that would complicate the situation.
- I never thought about adding a global info state, but that's probably a good idea. The main places where use oncokb is the react-mutation-mapper(RMM) package. I don't think our store is integrate with it much. We still need to pass info status to the sub components. Or maybe I'm wrong. @onursumer

onursumer:
@zhx828 @alisman `react-mutation-mapper` does not know anything about the stores defined in `cbioportal-frontend`, but there is `MutationMapperStore` that we pass down to the certain components, but probably not all the way down to the leaf components.

Having access to a global state info might be a handicap for modularity though. For example, if one day we want to move OncoKB related components into a separate package, importing a dependency from `react-mutation-mapper` would be a problem. Of course we can always refactor the code when we need to do these kind of separation.

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -39,6 +40,7 @@ export type AnnotationProps = {
     hotspotData?: RemoteData<IHotspotIndex | undefined>;
     oncoKbData?: RemoteData<IOncoKbData | Error | undefined>;
     oncoKbCancerGenes?: RemoteData<CancerGene[] | Error | undefined>;
+    usingOncoKbPublicInstance: boolean;
```
onursumer, line 43: should we also rename this to `usingPublicOncoKbInstance` for consistency?

packages/react-mutation-mapper/src/component/track/OncoKbTrackTooltip.tsx
```diff
@@ -8,12 +8,16 @@ import { IndicatorQueryResp } from 'oncokb-ts-api-client';
 import OncoKbSummaryTable from '../oncokb/OncoKbSummaryTable';
 
 type OncoKbTrackTooltipProps = {
+    usingOncoKbPublicInstance: boolean;
```
onursumer, line 11: this one too?

```suggestion
    usingPublicOncoKbInstance: boolean;
```

alisman:
@zhx828 not sure i understand ```But that status needs to be managed once the application is alive since it still needs to reach out to the info endpoint.```.  you mean it can change during the life of the application and we don't want to have redeploy to handle that?   

@zhx828 @onursumer on issue of global state, understood.  sounds like you made the right decision

zhx828:
@alisman I doubt the status, whether it's a public oncokb instance, will change. But it does not feel right to me when a portal backend state relies on an API and that API call needs to be triggered when the portal is deployed. What if at that time, the endpoint failed, should we manage to refetch the info?
I think it's better handle in the frontend, the only downside is we need to aware of the logic whenever oncokb is used.

@jjgao @alisman @onursumer any thoughts?

PR #3183, opened 2020-04-30
survival data frontend implementation

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/7433

- a support old data format and new (we should merge this before backend, then remove old format support) 

**screenshots:**
**oncoprint:** 
oncoprint
**patient view:**
patient_view
**study view:**
study_view
**clinical data:**
clinical_data
**survival plot:**
image

### DIFF
--- a/src/pages/resultsView/SurvivalStoreHelper.ts
+++ b/src/pages/resultsView/SurvivalStoreHelper.ts
@@ -1,8 +1,12 @@
 import { Patient, ClinicalData } from 'cbioportal-ts-api-client';
 import { PatientSurvival } from '../../shared/model/PatientSurvival';
+import _ from 'lodash';
+import { isNullSurvivalClinicalDataValue } from './survival/SurvivalUtil';
 
 export function getPatientSurvivals(
-    survivalClinicalDataGroupByUniquePatientKey: any,
+    survivalClinicalDataGroupByUniquePatientKey: {
+        [patientKey: string]: ClinicalData[];
+    },
     targetUniquePatientKeys: string[],
     statusAttributeId: string,
     monthsAttributeId: string,
@@ -11,26 +15,26 @@ export function getPatientSurvivals(
     if (targetUniquePatientKeys) {
         return targetUniquePatientKeys.reduce(
             (patientSurvivals: PatientSurvival[], uniquePatientKey: string) => {
-                const clinicalData: ClinicalData[] =
+                const clinicalData =
                     survivalClinicalDataGroupByUniquePatientKey[
                         uniquePatientKey
                     ];
                 if (clinicalData) {
-                    const statusClinicalData:
-                        | ClinicalData
-                        | undefined = clinicalData.find(
+                    const statusClinicalData = clinicalData.find(
                         c => c.clinicalAttributeId === statusAttributeId
                     );
-                    const monthsClinicalData:
-                        | ClinicalData
-                        | undefined = clinicalData.find(
+                    const monthsClinicalData = clinicalData.find(
                         c => c.clinicalAttributeId === monthsAttributeId
                     );
                     if (
                         statusClinicalData &&
                         monthsClinicalData &&
-                        statusClinicalData.value != 'NA' &&
-                        monthsClinicalData.value != 'NA' &&
+                        !isNullSurvivalClinicalDataValue(
+                            statusClinicalData.value
+                        ) &&
+                        !isNullSurvivalClinicalDataValue(
+                            monthsClinicalData.value
+                        ) &&
                         !Number.isNaN(Number(monthsClinicalData.value))
                     ) {
                         patientSurvivals.push({
@@ -43,6 +47,54 @@ export function getPatientSurvivals(
                     }
                 }
                 return patientSurvivals;
+            },
+            []
+        );
+    } else {
+        return [];
+    }
+}
+
+export function getClinicalDataOfPatientSurvivalStatus(
+    survivalClinicalDataGroupByUniquePatientKey: {
+        [patientKey: string]: ClinicalData[];
+    },
+    targetUniquePatientKeys: string[],
+    statusAttributeId: string,
+    monthsAttributeId: string
+): ClinicalData[] {
+    if (targetUniquePatientKeys) {
+        return targetUniquePatientKeys.reduce(
+            (
+                patientSurvivalStatusData: ClinicalData[],
+                uniquePatientKey: string
+            ) => {
+                const clinicalData =
+                    survivalClinicalDataGroupByUniquePatientKey[
+                        uniquePatientKey
+                    ];
+                if (clinicalData) {
+                    const statusClinicalData = clinicalData.find(
+                        c => c.clinicalAttributeId === statusAttributeId
+                    );
+                    const monthsClinicalData = clinicalData.find(
+                        c => c.clinicalAttributeId === monthsAttributeId
+                    );
+                    if (
+                        statusClinicalData &&
+                        monthsClinicalData &&
+                        !isNullSurvivalClinicalDataValue(
+                            statusClinicalData.value
+                        ) &&
+                        !isNullSurvivalClinicalDataValue(
+                            monthsClinicalData.value
+                        ) &&
+                        !Number.isNaN(Number(monthsClinicalData.value))
+                    ) {
+                        patientSurvivalStatusData.push(statusClinicalData);
+                    }
+                }
+                return patientSurvivalStatusData;
             },
             []
         );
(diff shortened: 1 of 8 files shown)


dippindots:
This front-end pr should be merged before backend prs, the second commit of this pr need to be reverted after related back-end codes deployed.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -202,12 +208,28 @@ function getSpanElementsFromCleanData(
     let spans = [];
     return Object.keys(clinicalAttributesCleanDerived).map(key => {
         let value = clinicalAttributesCleanDerived[key];
+        let attrValue = value;
+        // update survival data
```
alisman, line 212: can you improve this comment to explain bussiness log better?  also, i think the following logic should have a test.

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -50,3 +51,46 @@ export function getPatientSurvivals(
...
+                    );
+                    const monthsClinicalData:
+                        | ClinicalData
+                        | undefined = clinicalData.find(
+                        c => c.clinicalAttributeId === monthsAttributeId
+                    );
+                    if (
+                        statusClinicalData &&
+                        monthsClinicalData &&
+                        statusClinicalData.value != 'NA' &&
```
alisman, line 82: !==

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -50,3 +51,46 @@ export function getPatientSurvivals(
         return [];
     }
 }
+
+export function getPatientSurvivalUniqueData(
+    survivalClinicalDataGroupByUniquePatientKey: any,
```
alisman, line 56: seems like this is survivalClinicalDataGroupByUniquePatientKey:ClinicalData[]. ?

alisman: or sorry, [key:string] : ClinicalData[]

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -50,3 +51,46 @@ export function getPatientSurvivals(
...
+                        statusClinicalData &&
+                        monthsClinicalData &&
+                        statusClinicalData.value != 'NA' &&
+                        monthsClinicalData.value != 'NA' &&
+                        !Number.isNaN(Number(monthsClinicalData.value))
+                    ) {
+                        patientSurvivalData.push(statusClinicalData.value);
+                    }
+                }
+                return _.uniq(patientSurvivalData);
```
alisman, line 89: i think you're calling uniq inside the reduce loop.  why not call it once on final result of reduce?

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -50,3 +51,46 @@ export function getPatientSurvivals(
         return [];
     }
 }
+
+export function getPatientSurvivalUniqueData(
```
alisman, line 55: this function returns an array of strings but it's not super clear what they are based on function name _Data.  status values?    I also wonder if it would be better to return array of the clinical data objects and then having calling code pluck off the values.  This makes this more reausable.

alisman, line 55: lets get a test on this

src/pages/resultsView/SurvivalStoreHelper.ts
alisman: i think if you just cast this input property, then all the casting inside will be unnecessary.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +120,103 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
     return estimates;
 }
 
+export function parseSurvivalData(s: string) {
+    const splitedStatusAndLabel = s.split(':').map(text => text.trim());
```
alisman, line 124: splitStatusAndLabel

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +120,103 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
...
+    }
+
+    // if status not exists, status will be undefined
+    return {
+        status,
+        label,
+    } as ParsedSurvivalData;
+}
+
+export function getStatus(s: string, prefix: string) {
```
alisman, line 143: I would make this name more descriptive.  Maybe getSurvivalStatusBoolean?

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +120,103 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
...
+            }
+            return attributesPrefix;
+        },
+        [] as string[]
+    );
+    return attributesPrefix;
+}
+
+export function getSurvivalStatusAttributePrefix(attributeId: string) {
+    if (attributeId.indexOf('_STATUS')) {
```
alisman, line 180: this would be a good application of regular expression (with capture groups) because you could accomplish this without seeking _STATUS twice.   "BLAH_STATUS".match(/([^_]*)_STATUS/)

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +120,103 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
...
+): string {
+    // if not found set as 'Event'
+    let vocabulary = 'Event';
+
+    if (prefix in survivalClinicalDataVocabulary) {
+        vocabulary = survivalClinicalDataVocabulary[prefix][0];
+    } else {
+        // find first data with '1:' prefix
+        _.forEach(uniqueSurvivalData, data => {
+            const splitedData = data.split(':');
```
alisman, line 209: splitData

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +120,103 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
     return estimates;
 }
 
+export function parseSurvivalData(s: string) {
```
alisman, line 123: we should get tests on all the new functions in this file.   also, keep an eye toward performance.  there are some optimizations that could be made.  but it's only worth making them IF these function are going to be run for every sample in result set.   if not, then it's negligible.    you would know better what the "n" is here.

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -50,3 +49,47 @@ export function getPatientSurvivals(
...
+                    const statusClinicalData = clinicalData.find(
+                        c => c.clinicalAttributeId === statusAttributeId
+                    );
+                    const monthsClinicalData = clinicalData.find(
+                        c => c.clinicalAttributeId === monthsAttributeId
+                    );
+                    if (
+                        statusClinicalData &&
+                        monthsClinicalData &&
+                        statusClinicalData.value !== 'NA' &&
```
alisman, line 81: gauranteed to be NA?  i mean could it be lowercase?

adamabeshouse: either way lets make it a constant and put it somewhere, maybe even in the /api/ folder in an api constants file

dippindots: @alisman @adamabeshouse No, it's not guaranteed to be NA. I should make a constant in SurvivalUtil. The validator will allow the following values (ignore case and ignore whitespaces): `"[not applicable]", "[not available]", "[pending]", "[discrepancy]", "[completed]", "[null]", "", "na"`

https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py#L2156
https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/scripts/importer/validateData.py#L2468-L2470

dippindots: @alisman @adamabeshouse Added a constant and a function in `SurvivalUtil` to check if a value is null. Can I get your review again? Thanks!

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -82,6 +140,69 @@ export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
...
+        const parsedSurvivalData = parseSurvivalData(s);
+        if (parsedSurvivalData.status) {
+            return parsedSurvivalData.status === '1';
+        }
+    }
+    // return false as default for values that cannot get mapped
+    return false;
+}
+
+export function getStatusCasesHeaderVocabulary(
```
alisman, line 179: i don't think "vocabulary" is the right word here.  it's just a string so it should be "text"

PR #3193, opened 2020-05-05
Fix bilibili link

dippindots:
bilibili link is case sensitive, replace with the correct link.


inodb:
Good catch @dippindots ! Thanks!

dippindots:
@inodb np!

PR #3198, opened 2020-05-06
add warning for studies with mixed genome build

leexgh:
adding warning for studies with mixed genome build
Fix: https://github.com/cBioPortal/cbioportal/issues/7482

all studies in cbioportal are hg19, here are some mockup:
- when select multi studies in mixed reference genome:
image

- mixed studies on result view page:
image

- mixed studies on study view page:
image


src/shared/lib/referenceGenomeUtils.tsx
```diff
@@ -0,0 +1,76 @@
...
+}
+
+export function isGrch38(genomeBuild: string) {
+    return (
+        REFERENCE_GENOME.grch38.NCBI.match(new RegExp(genomeBuild, 'i')) ||
+        REFERENCE_GENOME.grch38.UCSC.match(new RegExp(genomeBuild, 'i'))
+    );
+}
+
+export function normalizeStudyReferenceGenome(genomeBuild: string) {
```
inodb, line 66: the name of this function is maybe better if it's something like `printReferenceGenome` or `formatReferenceGenome`? It's not really normalizing but rather a formatting / toString() style function

leexgh: @inodb good point, thanks

PR #3200, opened 2020-05-06
For localdb e2e tests use ping from busybox

inodb:
The code was using ping from the backend container to check whether the
container is alive. The java 11 slim image does not have ping. Busybox is a
very small image that has tons of common unix utilities, so maybe a
decent choice of image to check for liveness. That said I'm not sure if
there is a more common approach in e.g. docker itself to check for
liveness of a container in a network.

See backend PR that introduces java 11: https://github.com/cBioPortal/cbioportal/pull/7464

### DIFF
--- a/end-to-end-test/local/docker/setup_docker_containers.sh
+++ b/end-to-end-test/local/docker/setup_docker_containers.sh
@@ -101,7 +101,7 @@ run_cbioportal_container() {
 
     sleeptime=0
     maxtime=180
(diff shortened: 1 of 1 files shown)


inodb:
@nr23730 good point that it is probably related to slim image 👍. Seems like a good idea regardless to use the `slim` image

I looked at the e2e test failures here but they are not related to this PR so gonna merge this

PR #3202, opened 2020-05-07
Enrichments: fix p-value column tooltip

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7475


src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
```diff
@@ -24,6 +24,7 @@ export interface IAlterationEnrichmentTableProps {
     onCheckGene?: (hugoGeneSymbol: string) => void;
     onGeneNameClick?: (hugoGeneSymbol: string) => void;
     checkedGenes?: string[];
+    groupsCount: number;
```
adamabeshouse, line 27: I think this can be derived from `props.data` as `_.values(props.data[0].groupsSet).length`

PR #3206, opened 2020-05-07
update oncokb ts client

zhx828:


zhx828:
I forgot the update the oncokb api version at the oncokb side. Sigh... it's too much work tagging then update again. I will bump up two versions next time.

PR #3207, opened 2020-05-07
upgrade yarn to v1.22.4

inodb:
Upgrade yarn to v1.22.4

Side note: originally created this because jitpack was failing due to missing yarn binary, but turned out the issue was on the yarn S3 server that hosted binaries.


inodb:
test failures are unrelated (api sync has been upgraded in master)

PR #3209, opened 2020-05-11
improve coloring in plots tab - better UI, and enable coloring by clinical data

adamabeshouse:
image

Also, legends can now be clicked on to highlight data
image

image


tmazor:
Awesome @adamabeshouse ! This is a great addition!

I do have a few thoughts -- everything below is using this query: https://www.cbioportal.org/results?=&session_id=5a1eef06498eb8b3d560f502&show_samples=false

1) Looks like when the page first loads, it's defaulting to color by both Mutation & Copy Number? I think we used to start with just Mutation by default?
2) "Mutation Type" should have an * after, to link to the footnote about driver annotation settings
3) If I color by Diagnosis Age, the legend just shows two colored dots, I assume these are the min & max? Can we instead show a continuous color spectrum? I think that's a common way to show a label for something like this.
4) If I then select "Log Scale", the text "(log scale)" appears somewhat awkwardly in the middle of the legend:
image
I can also see that some samples are colored slightly differently, but it's hard to appreciate without seeing a labeled continuous color gradient in the legend.
5) When the legend is horizontal and below the chart, can you add some space between the values? In this example, "NO" is closer to the green circle that means "YES" than it is to it's own red circle (plot: Mut# (log scaled) vs Dx, colored by Family History of Primary Brain Tumor):
image
6) For OS Status, the legend text is actually overlapping (plot: Mut# (log scaled) vs Dx, colored by OS Status):
image
7) This tooltip is cut off by the plot boundary (plot: Mut# (log scaled) vs Dx, colored by Prior Cancer Diagnosis Occurence)
image
8) When I color by a gene, the legend includes the gene name. When I color by a clinical attribute, that name should also be part of the legend.

adamabeshouse:
@tmazor regarding (1), yes that's because @jjgao requested to default color by CNA now as well.

adamabeshouse:
@tmazor okay, I made all of the requested changes, please give it another lookover when you can

tmazor:
looks good to me @adamabeshouse !

src/pages/staticPages/webAPI/WebAPIPage.tsx
```diff
@@ -119,10 +119,10 @@ export default class WebAPIPage extends React.Component<{}, {}> {
                 </Helmet>
 
                 <h1>Web API</h1>
```
alisman, line 121: rebase

src/shared/cache/ClinicalDataCache.ts
```diff
@@ -64,6 +67,14 @@ export function clinicalAttributeIsLocallyComputed(attribute: {
     );
 }
 
+export type ClinicalDataCacheEntry = {
```
alisman, line 70: maybe a comment explaining why we include things like color state in data cache

src/shared/cache/ClinicalDataCache.ts
```diff
@@ -259,15 +270,81 @@ export default class ClinicalDataCache extends MobxPromiseCache<
...
                         studyToMutationMolecularProfilePromise.result!,
                         studyIdToStudyPromise.result!,
                         coverageInformationPromise.result!
-                    ),
+                    );
+                    let categoryToColor;
+                    let numericalValueToColor;
+                    let logScaleNumericalValueToColor;
+                    let numericalValueRange;
+                    if (q.datatype === 'STRING') {
```
alisman, line 286: can we move this out into a function

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -899,6 +903,8 @@ export function makeClinicalTracksMobxPromise(
                     }
                 } else if (attribute.datatype === 'STRING') {
                     ret.datatype = 'string';
+                    (ret as any).category_to_color =
```
alisman, line 906: how come any?

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -272,46 +282,99 @@ export default class BoxScatterPlot<
...
+
     private get legend() {
         if (this.props.legendData && this.props.legendData.length) {
             let legendData = this.props.legendData;
             if (this.legendLocation === 'bottom') {
                 // if legend is at bottom then flatten labels
                 legendData = legendData.map(x => {
-                    let name = x.name;
-                    if (Array.isArray(x.name)) {
+                    let { name, ...rest } = x;
```
alisman, line 323: neat.  this assigns everything but x.name to an object called "rest"?

src/shared/components/plots/PlotUtils.ts
```diff
@@ -1,18 +1,121 @@
...
     IThreshold1D,
-    IWaterfallPlotData,
-    IScatterPlotData,
-    IBoxScatterPlotPoint,
     IThreshold2D,
-    IValue1D,
-    IValue2D,
 } from 'pages/resultsView/plots/PlotsTabUtils';
-import { IBoxScatterPlotData } from './BoxScatterPlot';
+import Timer = NodeJS.Timer;
```
alisman, line 8: What's Timer for?

src/shared/components/plots/BoxScatterPlot.tsx
adamabeshouse: yep, im not sure the technical name for this but it rules

src/shared/components/oncoprint/OncoprintUtils.ts
adamabeshouse: typescript doesnt handle this well, union type

PR #3211, opened 2020-05-13
Add transcript dropdown on result view

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/7479
Add transcript dropdown on result view, click dropdown to switch to other transcripts, the lollipop and mutation table will be updated based on the selection.

- Canonical transcript(default selection): the Protein Change and Mutation Type data are from database. The results view page should be exactly the same as the page without transcripts dropdown.
- Non-canonical transcripts: update mutations by genome nexus annotated results, so Protein Change and Mutation Type date are from Genome Nexus. Annotation column, Oncokb track and Cancer Hotspots tracks are disabled.

image

Mutation type column will show original data when it's `Other`
image

**How to test**:
https://deploy-preview-3211--cbioportalfrontend.netlify.app/
Try with any query, go to results view page, then go to Mutations tab. There should be a transcript switch dropdown on the right. (If no dropdown, run this in browser console)
```
localStorage.frontendConfig = '{"serverConfig":{"show_transcript_dropdown":true}}'
```
Switch transcripts by clicking the dropdown, after switching, check lollipop and mutation table, especially Protein Change, Mutation Type and HGVSc data.

It will be good to check Mutation Mapper Tool as well, same logic applied to it too.

**TODO**
- [x] Split out `Other` button to `Other` and `Fusion`
- [x] If the mutation is categorized as `Other`, show original data in Mutation Type column
- [x] Fixing protein change with `*` (like `* 280 *`) in server


src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -57,13 +57,13 @@ export default class Mutations extends React.Component<
 
     public render() {
         // use routing if available, if not fall back to the observable variable
-        const activeTabId = this.props.routing
```
leexgh, line 60: can't find where do we use routing here, so remove routing and use gene symbol instead

src/shared/lib/MutationUtils.ts
```diff
@@ -365,3 +365,24 @@ export function isValidGenomicLocation(mutation: Partial<Mutation>): boolean {
 
     return false;
 }
+
+export function extractAnnotatedMutation(
```
onursumer, line 369: this is more like `finding` than `extracting`, so maybe
```suggestion
export function findMatchingAnnotatedMutation(
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -1,20 +1,51 @@
...
+} from 'shared/lib/MutationUtils';
+import MutationTypeColumnFormatter from './MutationTypeColumnFormatter';
+import { VariantAnnotation } from 'genome-nexus-ts-api-client';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
 
 export default class ProteinChangeColumnFormatter {
-    public static getSortValue(d: Mutation[]): number | null {
+    public static getSortValue(
+        d: Mutation[],
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 22:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -1,20 +1,51 @@
...
+                indexedAnnotatedMutationByGenomicLocation,
+                isCanonicalTranscript
+            )
         );
     }
 
-    public static getTextValue(data: Mutation[]): string {
+    public static getTextValue(
+        data: Mutation[],
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 38:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -26,9 +57,17 @@ export default class ProteinChangeColumnFormatter {
     public static getFilterValue(
         data: Mutation[],
         filterString: string,
-        filterStringUpper: string
+        filterStringUpper: string,
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 61:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -43,41 +82,91 @@ export default class ProteinChangeColumnFormatter {
         return filterValue.toUpperCase().indexOf(filterStringUpper) > -1;
     }
 
-    public static getDisplayValue(data: Mutation[]): string {
+    public static getDisplayValue(
+        data: Mutation[],
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 87:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -43,41 +82,91 @@ export default class ProteinChangeColumnFormatter {
...
+            data,
+            indexedAnnotatedMutationByGenomicLocation,
+            isCanonicalTranscript
+        );
     }
 
-    public static getData(data: Mutation[]) {
+    public static getData(
+        data: Mutation[],
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 102:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -43,41 +82,91 @@ export default class ProteinChangeColumnFormatter {
...
-        // use text as display value
-        const text: string = ProteinChangeColumnFormatter.getDisplayValue(data);
-
-        return <span>{text}</span>;
-    }
-
-    public static renderWithMutationStatus(data: Mutation[]) {
+    public static renderWithMutationStatus(
+        data: Mutation[],
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 120:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -95,4 +184,68 @@ export default class ProteinChangeColumnFormatter {
...
+            }
+            // non-canonical and GN doesn't have data, show data from database
+        }
+        // canonical, show database result
+        return originalMutation.proteinChange;
+    }
+
+    private static shouldShowWarningForDataDifference(
+        originalMutation: Mutation,
+        indexedAnnotatedMutationByGenomicLocation?: {
```
onursumer, line 219:
```suggestion
        indexedAnnotatedMutationsByGenomicLocation?: {
```

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -522,6 +522,20 @@ export class ResultsViewPageStore {
             },
             { fireImmediately: true }
         );
+
+        reaction(
```
onursumer, line 526: Not sure if ResultsViewPageStore is the right place to trigger the gene selection.

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -101,7 +101,8 @@ export function filterMutationByTranscriptId(
 export function getMutationToTranscriptId(
     mutation: Mutation,
     ensemblTranscriptId: string,
-    indexedVariantAnnotations: { [genomicLocation: string]: VariantAnnotation }
+    indexedVariantAnnotations: { [genomicLocation: string]: VariantAnnotation },
+    overwrite?: boolean
```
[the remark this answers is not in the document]
leexgh: good point, I change them to `overwriteWithAnnotatedMutation`, what do you think?

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -664,7 +667,16 @@ export default class MutationTable<
             download: ProteinChangeColumnFormatter.getTextValue,
             sortBy: (d: Mutation[]) =>
                 ProteinChangeColumnFormatter.getSortValue(d),
-            filter: ProteinChangeColumnFormatter.getFilterValue,
```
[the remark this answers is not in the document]
leexgh:
```
Type '(data: Mutation[], filterString: string, filterStringUpper: string, indexedAnnotatedMutationsByGenomicLocation?: { [genomicLocation: string]: Mutation; } | undefined, isCanonicalTranscript?: boolean | undefined) => boolean' is not assignable to type '(data: Mutation[], filterString: string, filterStringUpper?: string | undefined, filterStringLower?: string | undefined) => boolean'.
  Types of parameters 'indexedAnnotatedMutationsByGenomicLocation' and 'filterStringLower' are incompatible.
    Type 'string | undefined' is not assignable to type '{ [genomicLocation: string]: Mutation; } | undefined'.
      Type 'string' is not assignable to type '{ [genomicLocation: string]: Mutation; } | undefined'.ts(2322)
```
I think tt's because in this method, I add fourth optional parameter `indexedAnnotatedMutationsByGenomicLocation`, while `filter` in LazyMobXTable has fourth optional parameter `filterStringLower`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2789,78 +2805,63 @@ export class ResultsViewPageStore {
...
+        if (
+            this.genes.isComplete &&
+            this.oncoKbCancerGenes.isComplete &&
+            this.oncoKbCancerGenes &&
+            this.uniqueSampleKeyToTumorType &&
+            this.mutations &&
+            this.mutationsByGene &&
+            this.selectedGene
+        ) {
+            return new ResultsViewMutationMapperStore(
```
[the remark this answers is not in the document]
leexgh: @onursumer yeah it's not that ideal, the purpose of this is to help not overwhelm genome nexus, so it can only send one gene instead of all genes, do you have better suggestions to do this?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -577,6 +591,8 @@ export class ResultsViewPageStore {
         return this.urlWrapper.tabId || ResultsViewTab.ONCOPRINT;
     }
 
+    @observable public selectedGeneSymbol: string | undefined = undefined;
```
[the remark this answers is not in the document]
leexgh: @onursumer are you saying this selected gene shouldn't be in ResultsViewPageStore?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2789,78 +2805,63 @@ export class ResultsViewPageStore {
...
+        }
+        return undefined;
+    }
 
-    public getMutationMapperStore(
-        hugoGeneSymbol: string
-    ): ResultsViewMutationMapperStore | undefined {
-        return this.mutationMapperStores.result[hugoGeneSymbol];
+    public getMutationMapperStore():
+        | ResultsViewMutationMapperStore
```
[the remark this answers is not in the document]
leexgh: actually prettier added this...

src/shared/components/mutationTable/MutationTable.tsx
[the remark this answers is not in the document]
onursumer: I see, makes sense.

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
onursumer: I think we can handle this by lazy initializing `MutationMapperStore`s in the `getMutationMapperStore(hugoSymbol: string)` function. That would eliminate the need to recompute the store upon a new gene selection action. We can define a cache of `[hugoGeneSymbol: string]: ResultsViewMutationMapperStore` in this store to make sure that each store is initialized only once.

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
onursumer: I think this should be defined in `Mutations.tsx`

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
onursumer: prettier can be annoying sometimes :)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4528,4 +4529,9 @@ export class ResultsViewPageStore {
             return isMixedReferenceGenome(this.studies.result);
         }
     }
+
+    @action
+    public setSelectedGeneSymbol(hugoGeneSymbol: string) {
```
onursumer, line 4534: I think we should move this to `Mutations.tsx`

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -173,13 +121,96 @@ export default class Mutations extends React.Component<
...
-        else {
-            this.mutationsGeneTab = id;
+        this.props.store.setSelectedGeneSymbol(id);
+    }
+
+    @computed get geneTabContent() {
+        if (
+            this.props.store.selectedGene &&
+            this.props.store.mutationMapperStoreForSelectedGene
+        ) {
```
onursumer, line 131: Assuming that we lazy init the mutation mapper stores within `getMutationMapperStore`, we can do something like this:
```suggestion
        if (this.selectedGene) {
            const mutationMapperStore = this.props.store.getMutationMapperStore(this.selectedGene)
```

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: Instead of doing this we can wait `genes` to complete before rendering the `Mutations` tab, and then set an initial value for the observable selectedGene inside `Mutations.tsx` constructor. We should avoid using reactions anyway unless it is absolutely necessary.

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
leexgh: moved to `Mutations.tsx`

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
leexgh: thanks!

leexgh:
@inodb @onursumer @jjgao I think based our discussion, we got a conclusion on how to handle the protein change, but the number of mutations remains the same when you switching transcripts, do you think that's ok? 
And this will also affect mutation mapper tool because they are using same methods, so in mutation mapper tool when you switch transcripts, only the data inside the table will change, while currently in production we will filter some mutations out when switching

src/config/serverConfigDefaults.ts
```diff
@@ -41,6 +41,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     show_civic: false,
     show_pathway_mapper: true,
     show_mutation_mapper_tool_grch38: true,
+    show_transcript_dropdown: true,
```
inodb, line 44: @leexgh wonder if we should make it false for now so we can test in production before making true the default for everyone

inodb: it is convenient with testing tho for now, but might make sense to change before merging

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -101,7 +101,8 @@ export function filterMutationByTranscriptId(
 export function getMutationToTranscriptId(
```
alisman, line 101: i think this function should actually be called ```getMutationByTranscriptId```

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -213,16 +210,24 @@ export function getMutationFromSummary(
 export function getMutationsToTranscriptId(
```
alisman, line 210: getMutationsByTranscriptId

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -127,43 +128,39 @@ export function getMutationToTranscriptId(
...
         }
+        return annotatedMutation as Mutation;
     } else {
-        return undefined;
+        // if mutation is not annotatable or can't map back (due to genomic location normalization or other reasons), return original mutation
+        return mutation;
     }
 }
 
 export function getMutationFromSummary(
```
alisman, line 147: @onursumer @leexgh this function appears to be adding annotation data to the passed mutation.  i think the name of the function should communicate that.   @leexgh i realize this is existing code, but lets improve

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -213,16 +210,24 @@ export function getMutationFromSummary(
...
 ): Mutation[] {
-    return _.compact(
-        mutations.map(mutation =>
-            getMutationToTranscriptId(
-                mutation,
-                ensemblTranscriptId,
-                indexedVariantAnnotations
+    const fusionMutation = getFusionMutations(mutations);
+    // only non-fusion mutations need to get mutation with transcript id
+    const annotatableMutation = _.difference(mutations, fusionMutation);
```
alisman, line 218: pluralize: annotatableMutations

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -337,3 +342,11 @@ export function genomicLocationStringFromVariantAnnotation(
         variantAllele,
     });
 }
+
+export function getFusionMutations(mutations: Mutation[]) {
+    return mutations.filter(
+        (m: Mutation) =>
+            m.mutationType &&
+            new RegExp('fusion', 'i').test(m.mutationType.toLowerCase())
```
alisman, line 350: you don't need to lowercase mutationType since the regex is case insensitive.      Also, this regexp only needs to be instantiated once.  Making a new one inside loop is a performance drain.  Since there might well be tens of thousands of mutations, it's worth optimizing.  Just make the regexp outside the function.

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -101,7 +101,8 @@ export function filterMutationByTranscriptId(
 export function getMutationToTranscriptId(
     mutation: Mutation,
     ensemblTranscriptId: string,
-    indexedVariantAnnotations: { [genomicLocation: string]: VariantAnnotation }
+    indexedVariantAnnotations: { [genomicLocation: string]: VariantAnnotation },
+    overwriteWithAnnotatedMutation?: boolean
```
onursumer, line 105: Ideally we should add new tests for the updated logic in this function and all the other functions modified in this utility file. Not sure if it is critical for this PR though

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -268,10 +268,12 @@ export default class ResultsViewPage extends React.Component<
                             id={ResultsViewTab.MUTATIONS}
                             linkText="Mutations"
                         >
-                            <Mutations
-                                store={store}
-                                appStore={this.props.appStore}
-                            />
+                            {store.genes.isComplete && (
+                                <Mutations
```
alisman, line 272: we seem to do this checking for loadind inside of tab components, not here.  ALso make sure we are showing loader

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2789,78 +2793,56 @@ export class ResultsViewPageStore {
...
-                                return map;
-                            },
-                            {}
-                        )
-                    );
-                } else {
-                    return Promise.resolve({});
-                }
+    public createMutationMapperStoreForSelectedGene(gene: Gene) {
+        console.log('created a new store');
```
alisman, line 2797: clean up

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2789,78 +2793,56 @@ export class ResultsViewPageStore {
...
-        hugoGeneSymbol: string
+        gene: Gene
     ): ResultsViewMutationMapperStore | undefined {
-        return this.mutationMapperStores.result[hugoGeneSymbol];
+        if (
+            this.genes.isComplete &&
+            this.oncoKbCancerGenes.isComplete &&
+            this.oncoKbCancerGenes &&
+            this.uniqueSampleKeyToTumorType &&
+            this.mutations &&
```
alisman, line 2838: is there a reason we aren't checking for some mutation isComplete?  same for samples

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -28,12 +30,19 @@ export default class Mutations extends React.Component<
 > {
     private userSelectionStore: MutationMapperUserSelectionStore;
 
-    @observable mutationsGeneTab: string;
+    @observable public selectedGeneSymbol: string = this.props.store.genes
+        .result![0].hugoGeneSymbol;
```
onursumer, line 34: let's do this in the constructor if possible. I think it is okay to assume that result may be undefined, so here:
```suggestion
    @observable public selectedGeneSymbol: string | undefined;
```

in the constructor:
```suggestion
    this.selectedGeneSymbol = props.store.genes.result ? props.store.genes.result[0].hugoGeneSymbol: undefined;
```

src/pages/resultsView/mutation/Mutations.tsx
alisman: @inodb  We probably want to save the selected transcript id in the url? so we can also save selected gene.   lets discuss in a chat

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -73,24 +79,31 @@ export default class Mutations extends React.Component<
...
+            [hugoGeneSymbol: string]: Mutation[];
+        }
+    ) {
         const tabs: JSX.Element[] = [];
 
         genes.forEach((gene: string) => {
-            const mutationMapperStore = this.props.store.getMutationMapperStore(
-                gene
-            );
+            const mutations = mutationsByGene
```
alisman, line 101: if it's true that mutationsByGene is sometimes undefined, then it should probably be listed as an optional argument above or |undefined, so that ts will enforce this

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -73,24 +79,31 @@ export default class Mutations extends React.Component<
...
+                ? mutationsByGene[gene]
+                : undefined;
 
-            if (mutationMapperStore) {
-                const tabHasMutations =
-                    mutationMapperStore.mutations.length > 0;
+            if (mutations) {
```
alisman, line 105: just make this condition part of the logic above ```if (mutationsByGene) ...  else```

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -173,13 +127,98 @@ export default class Mutations extends React.Component<
...
+                            this.props.store.generateGenomeNexusHgvsgUrl
+                        }
+                        showTranscriptDropDown={
+                            AppConfig.serverConfig.show_transcript_dropdown
+                        }
+                    />
+                </div>
+            );
         }
+        return <div></div>;
```
alisman, line 222: return null

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -102,4 +103,19 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
             this.getDiscreteCNACache
         );
     }
+
+    @computed
+    get isCanonicalTranscript(): boolean | undefined {
```
alisman, line 108: should only return boolean.  checks for loading status should be handled outside of this check.

src/pages/resultsView/mutation/Mutations.tsx
inodb: yes agreed

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -241,6 +243,27 @@ export default class MutationMapperStore extends DefaultMutationMapperStore {
         return (this.mutationData.result || []).map(mutation => [mutation]);
     }
 
+    @computed get indexedAnnotatedMutationsByGenomicLocation(): {
+        [genomicLocation: string]: Mutation;
+    } {
+        if (
```
alisman, line 249: this is a problem in many places in our codebase but we should really not be doing these checks inside of this function.    if there is not mutationData.result, this getter should not ever be called.  the loading status should be handled outside.

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -65,6 +73,14 @@ export default class Mutations extends React.Component<
             .props.store.mutationsTabFilteringSettings.excludeGermline;
     }
 
+    @action
+    public setSelectedGeneSymbol(hugoGeneSymbol: string) {
+        this.selectedGeneSymbol = hugoGeneSymbol;
```
alisman, line 78: these two lines duplicate the source of truth, we should not need to keep the hugoGeneSymbol in both url and as instance property of the Mutations component.

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -42,7 +44,13 @@ export default class Mutations extends React.Component<
 
     constructor(props: IMutationsPageProps) {
         super(props);
-        this.selectedGeneSymbol = this.props.store.hugoGeneSymbols[0];
+        this.selectedGeneSymbol =
```
alisman, line 47: this will only fire on instantiation of component so it isn't "reactive".

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
leexgh: if transcript dropdown is disabled(https://github.com/cBioPortal/cbioportal-frontend/pull/3211/files#diff-3a3df9949f97de541d314ef00760e1d2R62), then this `isCanonicalTranscript` should be undefined. So I return boolean | undefined here.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2835,9 +2844,13 @@ export class ResultsViewPageStore {
             this.mutations.isComplete &&
             this.mutationsByGene.isComplete
         ) {
-            return this.mutationMapperStoreByGene[gene.hugoGeneSymbol]
+            let store = this.mutationMapperStoreByGene[gene.hugoGeneSymbol]
                 ? this.mutationMapperStoreByGene[gene.hugoGeneSymbol]
                 : this.createMutationMapperStoreForSelectedGene(gene);
+
+            autorun(() => store.setSelectedTranscript(selectedTranscriptId));
```
leexgh, line 2851: I know this is not a good way to handle it, but I got errors without `autorun`. Do you have better ideas on handling this?

image

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
```diff
@@ -444,7 +445,9 @@ export default class MutationMapper<
         return (
             this.store.mutationData.isPending ||
             this.isMutationPlotDataLoading ||
-            this.isMutationTableDataLoading
+            this.isMutationTableDataLoading ||
+            (!!this.store.activeTranscript &&
```
alisman, line 449: i think store should always have activeTranscript, right?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+                const canonicalTranscriptId = this.canonicalTranscript.result
+                    ? this.canonicalTranscript.result.transcriptId
+                    : undefined;
 
-        // TODO this.activeTranscript = activeTranscript?
-        return activeTranscript;
-    }
+                const selectedTranscript =
+                    this._selectedTranscript &&
+                    this.allTranscripts.result &&
```
alisman, line 114: since you are awaiting this.allTranscripts, you do not have to test for this.allTranscripts.result.  You can just use it.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+                    this.allTranscripts.result
+                        .map(transcript => transcript.transcriptId)
+                        .includes(this._selectedTranscript)
+                        ? this._selectedTranscript
+                        : undefined;
+
+                if (selectedTranscript) {
+                    activeTranscript = selectedTranscript;
+                } else if (this._activeTranscript !== undefined) {
+                    activeTranscript = this._activeTranscript;
```
alisman, line 124: there shouldn't be a _activeTranscript anymore

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+                            .result[0];
+                    } else {
+                        activeTranscript = canonicalTranscriptId;
+                    }
+                } else {
+                    activeTranscript = canonicalTranscriptId;
+                }
 
-    public set activeTranscript(transcript: string | undefined) {
+                // TODO this.activeTranscript = activeTranscript?
```
alisman, line 145: no need.  this is the one and only place where the app should look to find the active transcript

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
-        } else {
-            activeTranscript = canonicalTranscriptId;
-        }
+    readonly activeTranscript: MobxPromise<string | undefined> = remoteData(
+        {
+            await: () => [this.allTranscripts, this.canonicalTranscript],
+            invoke: async () => {
+                let activeTranscript;
+                const canonicalTranscriptId = this.canonicalTranscript.result
+                    ? this.canonicalTranscript.result.transcriptId
```
alisman, line 109: since you await it above, you can just reference result.transcriptId without checking.   if TS gives you an error, you can use ```this.canonicalTranscript.result!``` which tells TS you KNOW it will be !undefined

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+                    : undefined;
 
-        // TODO this.activeTranscript = activeTranscript?
-        return activeTranscript;
-    }
+                const selectedTranscript =
+                    this._selectedTranscript &&
+                    this.allTranscripts.result &&
+                    this.allTranscripts.result
+                        .map(transcript => transcript.transcriptId)
```
alisman, line 116: rather than doing map and then includes, look at _.some method (lodash).    it's a little more readable and performance is better (although here it's totally negligible).

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -92,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
 
-    public set activeTranscript(transcript: string | undefined) {
+                // TODO this.activeTranscript = activeTranscript?
+                return activeTranscript;
+            },
+        },
+        undefined
+    );
+
+    public setActiveTranscript(transcript: string | undefined) {
```
alisman, line 152: this should be gone

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -676,10 +697,12 @@ class DefaultMutationMapperStore implements MutationMapperStore {
 
     readonly ptmData: MobxPromise<PostTranslationalModification[]> = remoteData(
         {
-            await: () => [this.mutationData],
+            await: () => [this.mutationData, this.activeTranscript],
             invoke: async () => {
-                if (this.activeTranscript) {
-                    return this.dataFetcher.fetchPtmData(this.activeTranscript);
+                if (this.activeTranscript.result) {
```
alisman, line 702: is it actually possible that there can be no active transcript result (after isComplete)

src/pages/resultsView/ResultsViewPageStore.ts
alisman: let me think about this for a bit.

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -42,15 +49,12 @@ export default class Mutations extends React.Component<
         );
     }
 
+    @computed get selectedTranscripId() {
```
alisman, line 52: to fix the error above, try removing @computed here.  it should not be necessary.

packages/react-mutation-mapper/src/model/MutationMapperStore.ts
```diff
@@ -23,7 +23,7 @@ export interface MutationMapperStore {
     gene: Gene;
     dataStore: DataStore;
     uniprotId: RemoteData<string | undefined>;
-    activeTranscript?: string;
```
leexgh, line 26: @onursumer why `activeTranscript` is optional here?

onursumer: I think this should have been defined as `activeTranscript: string | undefined;` 

Maybe we update this to make sure that it is never undefined. But there might be cases where the gene is invalid or for some reason no transcript can be found for a given gene.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -90,50 +91,72 @@ class DefaultMutationMapperStore implements MutationMapperStore {
     @observable
     private _activeTranscript: string | undefined = undefined;
 
+    @observable
+    private _selectedTranscript: string | undefined = undefined;
```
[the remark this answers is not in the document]
leexgh: _selectedTranscript is the transcript id we got from url, and _activeTranscript is the current transcript id we use in mutation mapper, if _selectedTranscript is a valid transcript, we will set activeTranscript = selectedTranscript

src/pages/resultsView/mutation/Mutations.tsx
leexgh: @alisman I removed the `@computed` but still got the errors

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
leexgh: the `activeTranscript` might be undefined, so I have the check here

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
leexgh: @alisman `this.allTranscripts` might be undefined, so probably I need to keep this to avoid breaking something? @onursumer mentioned the transcripts should always be defined but we probably got empty result from the remote data

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
leexgh: same as `allTranscripts`, the`canonicalTranscript` could be undefined sometimes, probably I need to keep this to avoid breaking something?

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
leexgh: because in in `GeneSummary` we use activeTranscript!.result  (https://github.com/cBioPortal/cbioportal-frontend/pull/3211/files#diff-8cec5864c80004992d57fdcaf961294aR365), and activeTrancript is optoinal in MutationMapperStore, so I think we should check activeTrancript here

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -2,11 +2,12 @@ import autobind from 'autobind-decorator';
 import { IOncoKbData, remoteData } from 'cbioportal-frontend-commons';
 import {
     genomicLocationString,
-    getMutationsToTranscriptId,
     groupMutationsByProteinStartPos,
     uniqueGenomicLocations,
+    getMutationsByTranscriptId,
```
onursumer, line 7: let's move this back between `genomicLocationString` and `groupMutationsByProteinStartPos`

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: are we okay with an `autorun` here?

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
onursumer: Isn't it always the canonical transcript if the dropdown is disabled? It feels like this function should return true if the dropdown is disabled. Also the store should ideally only know about data not the state (enabled/disabled) of a UI component. So probably it is a good idea to move the `undefined` logic out of this function.

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -95,4 +184,69 @@ export default class ProteinChangeColumnFormatter {
 
         return content;
     }
+
+    private static getProteinChangeData(
```
onursumer, line 188: This function has a lot of custom logic in it. Might be better to move into a utility file and add a few tests.

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
```diff
@@ -95,4 +184,69 @@ export default class ProteinChangeColumnFormatter {
...
+                // non-canonical, GN has data, show GN annotated result
+                return annotatedMutation.proteinChange;
+            }
+            // non-canonical and GN doesn't have data, show data from database
+        }
+        // canonical, show database result
+        return originalMutation.proteinChange;
+    }
+
+    private static shouldShowWarningForDataDifference(
```
onursumer, line 217: This function is also a good candidate for some unit tests.

src/pages/resultsView/ResultsViewPageStore.ts
leexgh: @onursumer Do you have any suggestions? I spend a lot of time but didn't find a good solution... @alisman is helping me with it

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
leexgh: @onursumer good point. There were some tests for `MutationAnnotator`, we might also update them to fit the new changes

inodb:
@leexgh Looks really good! One more thing:

- would it be possible to update hgvsc column as well, such that it displays the hgvsc info for the selected transcript?

leexgh:
@inodb good catch, will do

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
leexgh: we removed the logic here, so no need for tests

src/shared/components/mutationTable/column/ProteinChangeColumnFormatter.tsx
leexgh: we removed the logic here, so no need for tests

PR #3219, opened 2020-05-18
Trim off html entities coming from NCI API

zhx828:
This is a part of fixes for https://github.com/oncokb/oncokb/issues/1815

### Before
Screen Shot 2020-05-18 at 1 06 31 PM

### After
Screen Shot 2020-05-19 at 10 37 03 AM

### DIFF
--- a/packages/cbioportal-frontend-commons/src/lib/StringUtils.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/StringUtils.ts
@@ -72,3 +72,24 @@ export function isUrl(str: string) {
     const pattern = /^http[s]?:\/\//;
     return pattern.test(str);
 }
+
+export const LESS_THAN_HTML_ENTITY = '&lt;';
+export const GREATER_THAN_HTML_ENTITY = '&gt;';
+
+// This method is only designed to trim off html tags <**> </**>
+// A full solution could be https://www.npmjs.com/package/string-strip-html
+// But there is an issue related to template literals (https://gitlab.com/codsen/codsen/-/issues/31)
+export function trimOffHtmlTagEntities(str: string) {
+    // I previously explored Regex but does not work really well in multiple matches
+    return (str || '')
+        .split(LESS_THAN_HTML_ENTITY)
+        .map(match => {
+            const elements = match.split(GREATER_THAN_HTML_ENTITY);
+            if (elements.length > 1) {
+                return elements[1];
+            } else {
+                return elements.length === 0 ? '' : elements[0];
+            }
+        })
+        .join('');
+}
--- a/packages/cbioportal-frontend-commons/src/lib/StringUtils.spec.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/StringUtils.spec.ts
@@ -6,6 +6,9 @@ import {
     stringListToSet,
     unquote,
     isUrl,
+    trimOffHtmlTagEntities,
+    LESS_THAN_HTML_ENTITY,
+    GREATER_THAN_HTML_ENTITY,
 } from './StringUtils';
 
 describe('longestCommonStartingSubstring', () => {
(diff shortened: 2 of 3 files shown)


alisman:
@zhx828 why are they encoded in the first place?  Seems wrong if they are meant to be displayed as HTML.  I understand this is coming from NCI system.

zhx828:
@alisman NCI allows the title including HTML entities, but later seems encoded in their system.  The API response from NCI has the title encoded already. Not sure what their intension is, probably better to store in the database. But when we display it, we should convert it back the original HTML.

onursumer:
@zhx828 @alisman looks like a poor API design to me. We have a similar issue for MyCancerGenome data as well. My solution was to actually parse and cleanup the data instead of using `dangerouslySetInnerHTML`: https://github.com/cBioPortal/cbioportal-frontend/blob/e469080ab05ffb826665d8667db3138b24b8f2c6/packages/react-mutation-mapper/src/component/myCancerGenome/MyCancerGenome.tsx#L34

But seems like we already use `dangerouslySetInnerHTML` in multiple other places in the codebase. So, not sure if it is worth the effort to implement a custom data parser for this one.

zhx828:
@onursumer not sure what the cleanup you are referring to. Do you mean remove all HTML entities from the title? I'm actually ok with that.

onursumer:
@zhx828 yes I meant removing all html entities and keeping only the text. But if it is too much work we can live with `dangerouslySetInnerHTML `. How dangerous can a title be, right? :)

zhx828:
> @zhx828 yes I meant removing all html entities and keeping only the text. But if it is too much work we can live with `dangerouslySetInnerHTML `. How dangerous can a title be, right? :)

I was hoping NCI can be trusted. I don't think trim off the html entities would be a lot of work. Maybe generalize your solution?

onursumer:
> I was hoping NCI can be trusted. I don't think trim off the html entities would be a lot of work. Maybe generalize your solution?

This might be it: https://stackoverflow.com/a/47436220

alisman:
@zhx828 @onursumer i think we should strip out HTML if it's not essential.    mostly i was reacting to the fact that's it's encoded by the api, which doesn't make a lot of sense.

zhx828:
> @zhx828 @onursumer i think we should strip out HTML if it's not essential. mostly i was reacting to the fact that's it's encoded by the api, which doesn't make a lot of sense.

I agree. OK, I will update he pr and let you guys know.

packages/cbioportal-frontend-commons/src/lib/StringUtils.ts
```diff
@@ -72,3 +72,17 @@ export function isUrl(str: string) {
     const pattern = /^http[s]?:\/\//;
     return pattern.test(str);
 }
+
+export const LESS_THAN_ENTITY = '&lt;';
+export const GREATER_THAN_ENTITY = '&gt;';
+
+export function trimOffHtmlTagsByEntity(htmlStr: string) {
```
onursumer, line 79: Should we just use [this package](https://www.npmjs.com/package/string-strip-html) instead of implementing our own custom function?

zhx828: @onursumer I thought about that, but we also need to decode it first before using this package. Thought that would be overkill to include two new packages to the project.

onursumer: [Here](https://www.npmjs.com/package/string-strip-html#features) it says: 

```
Uses recursive HTML decoding, so there's no way to cheat this library by using any kind of HTML encoding (unless you turn the decoding off via opts.skipHtmlDecoding)
```

This sounds like it actually handles encoded strings by default, so maybe we don't need any other package.

zhx828: Great find! Let me give it a try.

zhx828: Works really well, also passed my tests.
Gonna remove all my script and tests.

zhx828:
The failing tests are not related.

PR #3220, opened 2020-05-18
comparison -  download sample group membership data in overlap tab

adamabeshouse:
image


adamabeshouse:
@jjgao can you confirm this is what you had in mind?

jjgao:
Please add "No" as well.

image

PR #3223, opened 2020-05-19
oncoprint - make comparison group clinical tracks based on queried studies (even virtual)

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/7113.

When querying a virtual study, show the group in that virtual study:

Screen Shot 2020-05-20 at 9 54 29 AM
similar to study view:

Screen Shot 2020-05-15 at 10 59 59 AM

Before this PR the oncoprint showed for the same virtual study:

Screen Shot 2020-05-15 at 10 53 19 AM

Of note: groups are associated to a study set, not to each study individually

There were a few options about what groups to show given a virtual study V which comes from physical studies A,B,C:

1. V groups, A groups, B groups, C groups
2. V groups, ABC groups
3. V groups, A, B, C, AB, AC, BC, ABC
4. V groups

Since the study view only shows 4, we chose to do the same thing for Oncoprint

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -790,14 +790,14 @@ export class ResultsViewPageStore {
     }
 
     readonly comparisonGroups = remoteData<Group[]>({
-        await: () => [this.studyIds],
+        await: () => [this.queriedStudies],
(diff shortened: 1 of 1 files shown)


adamabeshouse:
ooh very detailed, thanks!

PR #3227, opened 2020-05-22
oncoprint - open data in oncoprinter

adamabeshouse:
image


jjgao:
Very cool!

Would it be hard to support mutational spectrum? Maybe they can be supplied as clinical data in the same format as in download (I couldn't find the issue about it somehow).

When heatmap is added, maybe also add a warning that heatmap will not be supported. (or can we support heatmap as well?

src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
```diff
@@ -339,7 +329,7 @@ function makeGeneticTrackDatum_Data_Type3(oncoprinterInputLine:OncoprinterInputL
     return ret as OncoprinterGeneticTrackDatum_Data;
 }*/
 
-export function makeGeneticTrackDatum_Data_Type2(
+export function makeGeneticTrackDatum_Data(
```
onursumer, line 332: was there a specific reason to have mixed case function names?

src/pages/staticPages/tools/oncoprinter/OncoprinterImportUtils.ts
```diff
@@ -0,0 +1,200 @@
...
+): OncoprinterGeneticInputLine {
+    const alteration = getOncoprinterGeneticAlteration(d);
+    if (alteration) {
+        const oncoprinterInput: Partial<OncoprinterGeneticInputLineType2> = {};
+        oncoprinterInput.sampleId = caseId;
+        oncoprinterInput.hugoGeneSymbol = d.hugoGeneSymbol;
+        oncoprinterInput.alteration = alteration;
+        oncoprinterInput.proteinChange = d.proteinChange;
+        oncoprinterInput.isGermline = !isNotGermlineMutation(d);
+        oncoprinterInput.isCustomDriver = d.driverFilter === 'Putative_Driver';
```
onursumer, line 59: should we define `const PUTATIVE_DRIVER = Putative_Driver` somewhere?

src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
adamabeshouse: I guess its just supposed to make these huge type names a little easier to comprehend

PR #3231, opened 2020-05-22
Pick default profile for any alteration types included in OQL

adamabeshouse:
If OQL is submitted which involves an alteration type that does not have a selected profile, a default profile for that type will automatically be selected.

e.g. if OQL is `BRCA1: EXP>0` but no mrna profile is selected, submission will now be allowed and an mrna profile will automatically be selected.

On the other hand, if there is no mrna profile at all, then there will be an error indicating that.

This PR standardizes the OQL submission flow, including this new feature, for the home page query form, the modify query form, the quick gene edit (pencil) on the results view, and the quick gene query box in the study view.

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -56,6 +56,7 @@ import { GroupComparisonTab } from '../groupComparison/GroupComparisonTabs';
 import NotUsingGenePanelWarning from './NotUsingGenePanelWarning';
 import Survival from '../groupComparison/Survival';
 import ResultsViewComparisonStore from './comparison/ResultsViewComparisonStore';
+import { QueryStore } from '../../shared/components/query/QueryStore';
 
 export function initStore(
     appStore: AppStore,
@@ -542,13 +543,18 @@ export default class ResultsViewPage extends React.Component<
         return isRoutedTo || (!isExcludedInList && !isExcluded);
     }
 
-    @computed get quickOQLSubmitButtion() {
+    @autobind
+    private getQuickOQLSubmitButton(store: QueryStore) {
         return (
             <>
                 <button
                     className={'btn btn-primary btn-sm'}
                     style={{ marginLeft: 10 }}
-                    onClick={this.handleQuickOQLSubmission}
+                    onClick={() => {
+                        //this.handleQuickOQLSubmission();
+                        store.submit();
+                    }}
+                    disabled={!store.submitEnabled}
                 >
                     Submit Query
                 </button>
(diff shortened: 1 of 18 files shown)


src/pages/studyView/studyPageHeader/rightPanel/RightPanel.tsx
```diff
@@ -53,16 +57,88 @@ export default class RightPanel extends React.Component<IRightPanelProps, {}> {
         );
     }
 
+    @computed get isQueryLimitReached(): boolean {
```
alisman, line 60: did we just not check for this in study view query in past?

src/shared/components/query/QueryStore.ts
```diff
@@ -1535,6 +1539,51 @@ export class QueryStore {
         return result;
     }
 
+    @computed get defaultMutationProfile() {
```
alisman, line 1542: seems like there is a lot of repetition here.  why not have defaultProfiles object populate and extract out the "find" logic into a function.

src/pages/studyView/studyPageHeader/rightPanel/RightPanel.tsx
```diff
@@ -53,16 +57,88 @@ export default class RightPanel extends React.Component<IRightPanelProps, {}> {
...
+            return 0;
+        }
+    }
+
+    @computed get oqlMessages() {
+        return getOqlMessages(this.props.store.geneQueries);
+    }
+
+    @computed get oqlSubmitError() {
+        if (this.props.store.isSingleNonVirtualStudyQueried) {
```
alisman, line 89: what happens if it's multiple studies?

adamabeshouse: that case is captured in the ensuing if blocks

src/pages/studyView/studyPageHeader/rightPanel/RightPanel.tsx
adamabeshouse: if we did I couldn't find it

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -542,34 +543,49 @@ export default class ResultsViewPage extends React.Component<
...
-                &nbsp;
-                <button
-                    className={'btn btn-link btn-sm'}
-                    onClick={this.toggleOQLEditor}
-                >
-                    Cancel
-                </button>
-            </>
-        );
+    @computed get quickOQLSubmitButton() {
```
alisman, line 546: i think submitting store (instead of just changing the gene_list param) will cause us to lose settings kept in url?

adamabeshouse: good catch. fixed this by passing an extra arg to make the URL not clear here: https://github.com/cBioPortal/cbioportal-frontend/pull/3231/files#diff-3cdd9cf2b536b6b0a606ab1d87e2d04aR578

adamabeshouse:
Hey @tmazor , I'm going to write some e2e's for this but if you have time could you please give this a hand test?

jjgao:
Thanks @adamabeshouse! Looks good!

PR #3232, opened 2020-05-22
Remove filterJson hash from study-view url

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7187


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1008,6 +1008,7 @@ export class StudyViewPageStore {
 
         if (!_.isEqual(toJS(this.initialFiltersQuery), filters)) {
             this.initialFiltersQuery = filters;
+            this.urlWrapper.updateURL({}, this.currentTab);
```
alisman, line 1011: if you look at signature for updateURL, you'll see there's a param for "replace."   We want to pass true for this because we don't want to have the url with filterjson in history (accessible by back button).

Also, is there a reason we update url here in the store?  I think it would be more appropriate out in the view.

PR #3235, opened 2020-05-26
Show super small p/q values as <10^-x rather than 0

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7526

Example
Screen Shot 2020-05-26 at 5 54 48 PM


adamabeshouse:
wait why do we want to do this again? if we actually know the p-value is 0 @tmazor @jjgao  it seems like reporting an arbitrary error interval

PR #3241, opened 2020-05-28
add ability to query mutations by position range in OQL

adamabeshouse:
- [x]  add to OQL documentation https://github.com/cBioPortal/cbioportal/pull/7552


src/shared/lib/oql/oqlfilter.ts
```diff
@@ -672,10 +693,44 @@ function isDatumWantedByOQLMutationModifier<T>(
     datum: T,
     accessors: IAccessorsForOqlFilter<T>
 ) {
-    switch (modifier) {
+    switch (modifier.type) {
         case 'GERMLINE':
         case 'SOMATIC':
-            return accessors.mut_status(datum) === modifier.toLowerCase();
+            return accessors.mut_status(datum) === modifier.type.toLowerCase();
+        case 'RANGE':
```
onursumer, line 700: This logic might worth to modularize in `cbioportal-utils` for future use. We can define some generic RangeUtils function. At this point it might be over-engineering though.

There might also be node packages for this. For example, I was using https://guava.dev/releases/19.0/api/docs/com/google/common/collect/Range.html for backend. But of course, that's java :)

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -261,7 +262,11 @@ export function doesQueryContainMutationOQL(oql_query: string): boolean {
                     break;
                 } else if (alteration.alteration_type === 'any') {
                     // any DRIVER specification, which includes mutation
-                    if (alteration.modifiers.indexOf('DRIVER') > -1) {
+                    if (
+                        alteration.modifiers.findIndex(
+                            m => m.type === 'DRIVER'
+                        ) > -1
```
onursumer, line 268: I prefer `find` over `findIndex` if we don't need to use the index anywhere else:
```suggestion
                        alteration.modifiers.find(
                            m => m.type === 'DRIVER'
                        )
```

PR #3244, opened 2020-05-28
Check intragenic before regular fusion & Do not set intragenic as functionalFusion

zhx828:
Fix https://github.com/cBioPortal/cbioportal/issues/7210


zhx828:
The failing e2e tests are not related to this pr.

PR #3246, opened 2020-05-28
Add webinar 5 + remove webinar banner 😢

inodb:
Screen Shot 2020-05-28 at 6 25 27 PM


onursumer:
No need to 😢 , I am sure there will be more webinars in the future.

PR #3192, opened 2020-05-04
Add typed-css-modules-webpack-plugin to support type checking

onursumer:
Fix cBioPortal/cbioportal#7455

- Configure `typed-css-mdoules-webpack-plugin` to auto generate typings for every scss module (files matching the pattern `*.module.scss`)
- For each scss module file, it generates a TS type definition file in the form of `*.module.scss.d.ts`
- Works for both build and watch modes.

## Checks
- [ ] Has tests or has a separate issue that describes the types of test that should be created. If no test is included it should explicitly be mentioned in the PR why there is no test.
- [ ] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!

PR #3248, opened 2020-05-31
Split OncoKbCard to small components && Add e2e tests for OncoKB

zhx828:
This is an effort to separate out the business logics of OncoKbCard before integrating Dx/Px


end-to-end-test/shared/specUtils.js
```diff
@@ -191,6 +191,17 @@ function waitForNetworkQuiet(timeout) {
     }, timeout);
 }
 
+function closeDevModeBanner() {
```
alisman, line 194: i think this is fine although a cleaner way might be to hide it using css alone.   we have a special stylesheet called e2etest.scss which gets applied only in e2e context.  You could hide it that way. If you go down this route, i would try to give that alert a specific class, or use a specific selector.  like ```body > div:n-th(0) .alert``` so as not to mess with other alerts

alisman: btw the selector i put up there is just psuedo.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCard.tsx
```diff
@@ -1,131 +1,47 @@
...
-
     @computed
     get oncokbLinkOut() {
         let link: string | undefined = undefined;
-        if (this.props.gene) {
-            link = `https://oncokb.org/gene/${this.props.gene}`;
-            if (this.props.variant) {
-                link = `${link}/${this.props.variant}`;
+        if (this.props.hugoSymbol) {
+            link = `https://oncokb.org/gene/${this.props.hugoSymbol}`;
```
alisman, line 32: should be using proxy?   either way maybe we use a function from urls.ts

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    render() {
+        return (
+            <>
+                {!this.props.geneNotExist && (
+                    <div>
+                        {this.props.indicator &&
+                            this.getBody(this.props.indicator)}
+                        {this.props.usingPublicOncoKbInstance ? (
```
alisman, line 86: i don't think you need ternary here?  just add this last condition to the one above it?   also, for returning nothing, i like null better than <></>.  Unless is says otherwise somewhere else.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
```diff
@@ -0,0 +1,122 @@
...
+                            <></>
+                        ) : (
+                            <>
+                                {/*Use tab pane style for the disclaimer to keep the consistency since the info is attached right under the tab pane*/}
+                                <div className={tabsStyles['tab-pane']}>
```
alisman, line 91: @zhx828 @onursumer is this method of referecing styles (rather than dot notation) compliant with new typings?

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardTreatmentContent.tsx
```diff
@@ -0,0 +1,280 @@
...
+    variantSummary: string;
+    tumorTypeSummary: string;
+    biologicalSummary: string;
+    treatments: IndicatorQueryTreatment[];
+    pmidData: ICache;
+    usingPublicOncoKbInstance: boolean;
+};
+
+const ONCOKB_DATA_ACCESS_PAGE_LINK =
+    'https://docs.cbioportal.org/2.4-integration-with-other-webservices/oncokb-data-access';
```
alisman, line 32: i see a few urls embedded in these views.  Can we forsee these needing instance specific configuration?  is it worth trying to externalize them to urls.ts?

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardTreatmentContent.tsx
```diff
@@ -0,0 +1,280 @@
...
+export default class OncoKbCardTreatmentContent extends React.Component<
+    OncoKbCardTreatmentContentProps
+> {
+    @observable activeTab: ActiveTabEnum = ActiveTabEnum.ONCOGENICITY;
+
+    // TODO we should replace the tabs with an actual ReactBootstrap Tab,
+    public render() {
+        return (
+            <div className={mainStyles['oncokb-card']} data-test="oncokb-card">
+                <span>
```
alisman, line 71: span?

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
zhx828: @alisman I maybe new on this topic. I think I use both types. Generally I tried to use a-b in non modules, but camel in modules. I didn't do much of the refactoring in this pr since the intension is to separate the logics. But I'm happy to do that. Do we have any guidance on which style should we use?

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardTreatmentContent.tsx
```diff
@@ -0,0 +1,280 @@
...
+                        {this.activeTab === ActiveTabEnum.MUTATION_EFFECT && (
+                            <div
+                                className={classnames(tabsStyles['tab-pane'])}
+                                data-test={`${ActiveTabEnum.MUTATION_EFFECT}-pane`}
+                                style={{
+                                    maxHeight: 200,
+                                    overflowY: 'auto',
+                                }}
+                            >
+                                {this.props.biologicalSummary !== undefined &&
```
alisman, line 205: i find these nested ternary expressions to be too hard to read and very easy to mess up.  i would rather see <IF/Else> used here.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardTreatmentContent.tsx
zhx828: It's a general guidance.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardTreatmentContent.tsx
```diff
@@ -0,0 +1,280 @@
...
+
+        content.push(parts[0]);
+
+        const comp = (
+            <a href={link.link} target={link.target || '_blank'}>
+                {link.keyword}
+            </a>
+        );
+
+        for (let i = 1; i < parts.length; i++) {
```
alisman, line 273: parts.forEach. or even   parts reduce(..., content).   i just don't like plain for loops unless the concern is performance.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCard.tsx
onursumer: Unfortunately we cannot use a function from `urls.ts` in any of the packages. We need to decouple it from the main project and make it generic first. We can pass the url as a property to this component though. But probably we still need a default value.

Also this is just a static link-out, not an API endpoint. Not sure if a proxy url is necessary here.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
onursumer: @alisman good point, it is not complaint with the plugin introduced in #3192. But that's a webpack plugin and only works for the main package (`cbioportal-frontend`). We are still missing css typings for modular packages, because we build them with rollup. I'll open a separate ticket for that.

alisman: Ok, so, @zhx828, when we introduce the typed css modules to packages, then we will need to bring these into compliance.   I believe that means dot notation and no dashes (styles.someClass).   I would just take opportunity now to adjust this here.

onursumer: @alisman sorry, I was wrong. Apparently both dot notation and brackets are valid, here is an example: https://github.com/cBioPortal/cbioportal-frontend/blob/9bf8a68e8ad0b36d75da3a3b47bd4eeeff3dd50e/src/pages/resultsView/download/DownloadTab.tsx#L677

This doesn't change the fact that we still need to introduce type checking for the packages. I have created a ticket for that: cBioPortal/cbioportal#7559

end-to-end-test/shared/specUtils.js
zhx828: I will open a separate PR for this since the changes need to be done at both frontend/backend repos.

zhx828: https://github.com/cBioPortal/cbioportal/issues/7561

zhx828:
The two failing tests are not related.
All tests in this pr have been run five times in a row and all green.

PR #3249, opened 2020-06-01
Fix relative path on institute image logo

alisman:
The path to logo (or any image) cannot every be relative because we have nested paths to subpages (results/oncoprint).    This PR allows installers to specify a path that starts with /images/ or an absolute path.   The former will allow things to work as they have in the past but will fix the image on interior pages.  The only time it will fail is when a portal is hosted in a sub folder.  In that case, the only option is to use the absolute path.

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -206,6 +206,20 @@ export function getOncoKbApiUrl() {
     return buildCBioPortalAPIUrl(`proxy/oncokb`);
 }
 
+export function getInstituteLogoUrl() {
+    if (AppConfig.serverConfig.skin_right_logo) {
+        if (/^http/.test(AppConfig.serverConfig.skin_right_logo)) {
+            return AppConfig.serverConfig.skin_right_logo;
+        } else {
+            return buildCBioPortalPageUrl(
+                `images/${AppConfig.serverConfig.skin_right_logo}`
+            );
+        }
+    } else {
+        return undefined;
+    }
+}
+
 export function getGenomeNexusApiUrl() {
     let url = AppConfig.serverConfig.genomenexus_url;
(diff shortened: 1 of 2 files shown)


src/shared/api/urls.ts
```diff
@@ -206,6 +206,18 @@ export function getOncoKbApiUrl() {
     return buildCBioPortalAPIUrl(`proxy/oncokb`);
 }
 
+export function getInstituteLogoUrl() {
+    if (AppConfig.serverConfig.skin_right_logo) {
+        if (/^http/.test(AppConfig.serverConfig.skin_right_logo)) {
+            return AppConfig.serverConfig.skin_right_logo;
+        } else {
+            return `/images/${AppConfig.serverConfig.skin_right_logo}`;
```
inodb, line 214: @alisman this should prolly be relative to root url of app for portals running in subfolder, e.g. pdx.cbioportal.org/pdx

alisman:
@inodb i don't know how many installations use the logo configuration option.  Without any doubt, their logo will be broken in the same way as ours is.    This will fix things for them only if they run their portal at root of domain.  If not, it will not help them and they need to specify an absolute path. 

Another option would be to  stop forcing this "images" folder on them and just accept whatever they put in configuration.  That would require US to adjust our configuration.  

@Luke-Sikina i add you to see what you guys have configured at DFC

inodb:
@alisman it shouldn't be too tricky to support non root config right? Maybe:

```
return `${frontendConfig.baseUrl}/images/${AppConfig.serverConfig.skin_right_logo}`
```

we are using these subfolders ourselves so would be good to support

src/shared/api/urls.ts
alisman: ah, yeah, didn't think about that

PR #3252, opened 2020-06-01
Add css modules typings for packages

onursumer:
Fix cBioPortal/cbioportal#7559

## Checks


packages/react-mutation-mapper/src/util/FilterUtils.ts
```diff
@@ -152,8 +152,10 @@ export function applyDefaultMutationFilter(
     const filterPredicates = filter.values.map((value: MutationFilterValue) => {
         const valuePredicates = Object.keys(value).map(key =>
             includesSearchTextIgnoreCase(
-                mutation[key] ? mutation[key].toString() : undefined,
-                value[key] ? value[key.toString()] : undefined
+                (mutation as any)[key]
```
zhx828, line 155: It's not a `Mutation` type?

packages/cbioportal-frontend-commons/package.json
```diff
@@ -18,15 +18,17 @@
...
-    "build": "cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=2048 yarn run rollup",
+    "build": "yarn run tcm && cross-env NODE_ENV=production NODE_OPTIONS=--max-old-space-size=2048 yarn run rollup",
     "start": "yarn run watch",
-    "watch": "yarn run rollup:watch",
+    "watch": "concurrently \"yarn run tcm:watch\" \"yarn run rollup:watch\"",
     "watchSSL": "yarn run watch",
     "rollup": "rollup -c rollup.config.ts",
-    "rollup:watch": "rollup -c rollup.config.ts -w",
+    "rollup:watch": "yarn run rollup --watch",
+    "tcm": "tcm -p src/**/*.module.scss",
```
zhx828, line 27: Too bad we cannot add a comment in JSON. It took me a while to understand what tcm is. Maybe it's ok.

packages/react-mutation-mapper/src/util/FilterUtils.ts
onursumer: It is, but `key` can be technically anything here. This was passing previously because `suppressImplicitAnyIndexErrors` was set to true for some reason. I have updated that in this PR to make the packages more type-safe.

packages/cbioportal-frontend-commons/package.json
onursumer: Hmm, actually we can rename it to `typed-css` or something like that.

PR #3253, opened 2020-06-02
remove console log on comparison page

dippindots:
remove console log on comparison page


dippindots:
Remove a console log line in comparison page store.

PR #3258, opened 2020-06-04
Update screenshots for release-3.4.0

dippindots:


inodb:
@dippindots do we want to show 0:LIVING and 1:DECEASED in the oncoprint?

Screen Shot 2020-06-04 at 7 49 38 AM

Seems like an implementation detail not that relevant to the user. But I guess it can also be confusing to clean the data. Thoughts @adamabeshouse ?

inodb:
Same for this one: 
Screen Shot 2020-06-04 at 8 51 22 AM

zhx828:
> Same for this one:
> Screen Shot 2020-06-04 at 8 51 22 AM
> 
> And on study view some custom coloring seems to have been lost:
> 
> was
> Screen Shot 2020-06-04 at 8 52 57 AM
> is now
> Screen Shot 2020-06-04 at 8 53 01 AM
> CC @zhx828 is this coloring important to keep?

The original color is for `true`. If the value of that status is changed, then we should update the screenshot. Otherwise, it's important to keep.

zhx828:
Actually for Living. But I saw the value is changed to `0:LIVING`. Is this intentional? @dippindots @inodb

src/shared/lib/Colors.ts
```diff
@@ -83,6 +83,36 @@ export let RESERVED_CLINICAL_VALUE_COLORS: { [value: string]: string } = {
...
+    '0:no progression': CLI_YES_COLOR,
+    '0:not progressed': CLI_YES_COLOR,
+    '0:censored': CLI_YES_COLOR,
+    '0:censor': CLI_YES_COLOR,
+    '0:alive or censored': CLI_YES_COLOR,
+    '0:alive or dead tumor free': CLI_YES_COLOR,
+    '0:no relapse': CLI_YES_COLOR,
+    '0:progression free': CLI_YES_COLOR,
+    '0:censure': CLI_YES_COLOR,
+    '0:ned': CLI_YES_COLOR,
```
zhx828, line 115: I think you are missing the space in the middle, that's why the color is not matched.

dippindots:
Thanks @inodb @zhx828 , I actually changed color in this pr, so in this deployment the color looks as same as previous, https://deploy-preview-3258--cbioportalfrontend.netlify.app/study/summary?id=acc_tcga. Can you confirm that?

src/shared/lib/Colors.ts
zhx828: nvm, just saw the message thread in slack. This looks good.

inodb: wondering if it isn't easier in this case to match on `0:` and `1:`? Guess that would involve more code changes, but kinda feels like one of the reasons for introducing the `0/1` stuff is to normalize the data?

zhx828:
@dippindots  @inodb  the api sync is failing as well. Is it related to this release?

inodb:
@zhx828 yes, i just updated it. So if this PR is rebased that should solve the problem

dippindots:
Thanks @inodb @zhx828. I will rebase this pr.

PR #3259, opened 2020-06-04
Combine Comparison and Survival navigation buttons into one button reading "Comparison/Surivval"

alisman:
We want users to access survival data as a sub tab of comparison.  This change hopefully ensures they won't miss the presence of survival data because it has lost it's top level status.

Fixes https://github.com/cBioPortal/cbioportal/issues/7583


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -317,7 +317,7 @@ export default class ResultsViewPage extends React.Component<
                         <MSKTab
                             key={10}
                             id={ResultsViewTab.COMPARISON}
-                            linkText={'Comparison'}
+                            linkText={'Comparison/Survival'}
```
adamabeshouse, line 320: should the tab name only show survival if there is survival data? (see the `hide` condition for SURVIVAL_REDIRECT)

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -310,14 +310,19 @@ export default class ResultsViewPage extends React.Component<
             {
                 id: ResultsViewTab.COMPARISON,
                 hide: () => {
-                    return !this.resultsViewPageStore.studies.isComplete;
+                    return this.resultsViewPageStore.survivalClinicalDataExists
```
alisman, line 313: we could show it and then CHANGE the text when this kicks in, but ...

zhx828:
@alisman there are few oncokb screenshots in this pr needed to be updated. You can ignore `reference/patient_view_-_without_levels_-_mutation_effect_tab_shows_correctly_element_chrome_1600x1000.png` it just flaky. I will need to spend some time on these tests.

PR #3260, opened 2020-06-04
(Patient page header) Remove prefix for OS_STATUS and DFS_STATUS

dippindots:


src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -53,6 +53,14 @@ function clean(clinicalData) {
                     }
                     cleanClinicalData[key] = value;
                     break;
+                // remove prefix ("1:" or "0:") for OS_STATUS and DFS_STATUS
```
alisman, line 56: great. i'm glad we can do this "cleaning" here.

PR #3264, opened 2020-06-08
Hide "Days since diagnosis label" on clinical timeline

alisman:
Fixes https://github.com/cbioportal/cbioportal/issues/7570

### DIFF
--- a/src/pages/patientView/timeline/styles.scss
+++ b/src/pages/patientView/timeline/styles.scss
@@ -10,9 +10,10 @@
     font-size: 10px;
 }
 
-.timeline-label {
-    font-family: sans-serif;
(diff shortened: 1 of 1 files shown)


src/pages/patientView/timeline/styles.scss
```diff
@@ -10,9 +10,10 @@
     font-size: 10px;
 }
 
-.timeline-label {
-    font-family: sans-serif;
-    font-size: 12px;
+#timeline > svg:first-child {
+    .timeline-label:last-of-type {
```
alisman, line 14: @inodb that label seems to be reliably the last child element (even though it appears first).  this is pretty icky but ...

PR #3266, opened 2020-06-08
Option to exclude unprofiled samples in results view

adamabeshouse:
image


tmazor:
Love this new feature @adamabeshouse ! A few comments:

1) In the settings menu, it says filters don't apply to Plots, Coexpression, CN Segments. But this new filter does apply to Plots. And I'm not sure if it applies to Co-expression because that tab was taking too long to load. But I can see it doesn't impact CN segments. We might need to rethink how the filters are labeled.
2) In contrast, looks like the mutations tab is not impacted by this filter - but I would expect it to be
3) Looks like Pathways is impacted by the filter (?), but the "xxx samples are excluded" warning doesn't appear
4) Download tab has the "xxx samples are excluded" message, but it looks like that is done unconsistently? The Gene Alteration Frequency is excluding samples but Types of Genetic Alterations Across All Samples is not?
5) I think there is something off about the overall frequencies being calculated. Or else maybe I'm not understanding something. Here's what I did:
     - I ran this query: https://bit.ly/3dXqXPJ  Let's call it A. I compared A with and without the new filter applied.
     - I also tried to replicate what A+filter should give me. I used study view to pull out samples with the relevant genomic profiles and then queried the same genes in that subset of samples. Let's call that B: https://bit.ly/2zsyuXH
     - So now I can compare A-filter vs A+filter vs B. And my expectation is that A+filter should be identical to B -- but that's not what I'm seeing. Sometimes they are the same, but sometimes A+filter does not match B. In general, I think the discrepancies are in cases where there's a denominator involved -- e.g. OncoPrint gene track frequencies or cancer type summary. My guess is the denominator that's being used is all samples in the query, rather than all samples in the query that are profiled for all profiles.
     - Am I understanding the behavior correctly? If so, I think the correct thing to report is the adjusted denominator that only includes samples that are profiled. Do you agree? Or am I misunderstanding the point of this?

adamabeshouse:
@tmazor thanks for the feedback!
1. Good call. I think the best thing to do is just remove that info icon. We show alert banners on each tab anyway that indicate if theyre affected.

2. Good point. Should those be included in the same banner as for VUS, and toggled together? 
image

3. Oops, you're right, good catch.

4. Good catch, I'll fix that.

5. Yes, you're absolutely right, that calculation was not taking the filtering into account. Good catch!

tmazor:
1. I'm worried that if we only have the single (i), people might get confused and think it applies to the filter settings also -- so how about we update the text for the Annotate Data info tooltip to say "Putative driver vs VUS settings apply to...." ? 

2. I think we should be consistent across tabs, and on other tabs the VUS filter is its own banner, so let's do the same on Mutations.

2b. Now that I'm testing with the VUS filter also, I'm realizing that we should probably be a littler more verbose with this new filter banner. Instead of:
"xx samples are excluded from analysis."
how about:
"xx samples that are not profiled in all select genomic profiles are excluded from analysis."
There may be cleaner phrasing we can use, but that's the best I have right now.

3. Looks good! Although now I realize Pathways also don't have a notification for VUS filtering. I don't think I ever tested for these banners when we introduced Pathways. I'll make another ticket about that.

4 & 5. These are kinda the same thing, and I think this is only partly fixed. But I think the underlying question is - given the example I laid out in my previous comment, should A+filter be 100% identical to B? I see differences right now in the following places:
- Header: case count on left and overall % altered on right
- Cancer Types Summary: when y-axis is frequency, those frequencies are different
- Mutations: I realize the filter just isn't implemented here at all yet, but there is a Somatic Mutation Frequency to the right of the lollipop so I'm adding it to the list in case it also has this happen
- Download: I haven't validated any actual downloads yet, but just from looking at the tables, the number of rows in the bottom table corresponds to the unfiltered data

For the tabs, I feel comfortable saying those should all be updated to fully reflect the filter. But I'm torn about the header.

A+filter:
image


B:
image


On one hand, that sample/patient count is correct, it is the number of cases that were queried. On the other hand, it is not the number of samples/patients that are being shown in most of the analyses below. And there's no indication in the header that the samples have been filtered. But the numerator of the frequency on the left is adjusted for the filter, so now that frequency is half filtered and half not.

I'm very torn. Anyone else have thoughts? @adamabeshouse @alisman @jjgao @schultzn (for full context, start with point 5 in my earlier comment)

tmazor:
Also it appears this PR is hiding the CN Segments tab? When I load the link called A without the PR, I see CN Segments. But when I use that magic bookmark thing to reload with this PR, the tab disappears.

adamabeshouse:
@tmazor I can't repro the hidden CN segments?

adamabeshouse:
seems like CN segments issue is somehow due to using netlify frontend on cbioportal backend

adamabeshouse:
@tmazor let me know if it looks good now

jjgao:
Thanks, @adamabeshouse !

I have some comments about the message: "xxx patients that are not profiled in all selected genomic profiles are excluded from analysis."
- Should it be "samples"?
- The message is accurate when it's just about whether samples are included in profiles, but for whether a gene is included in gene panels (e.g. for [this query](https://deploy-preview-3266--cbioportalfrontend.netlify.app/results/oncoprint?Action=Submit&cancer_study_list=msk_impact_2017&case_set_id=msk_impact_2017_all&gene_list=WT1%250AFLCN%250ANUP93&hide_unprofiled_samples=true&tab_index=tab_visualize)), we may need a different message. 
- How about we change it to "xxx samples are excluded from analysis since they are not profiled for one or more genes in one of the selected profiles. To include them, please see the ||| menu."

jjgao:
I don't see the message to the CN Segments tab. Why don't we apply the filter there? Any other tabs we don't apply the filter?

adamabeshouse:
@jjgao ✔️  

shouldn't it be 'patients' when viewing the oncoprint in 'patient' mode?

adamabeshouse:
@tmazor @jjgao please see updated version when the deploy finishes

tmazor:
This is looking great!

Just one minor thing - when I click "xxx patients/samples" in header, it should bring me to study view with a filter applied to show only those same patients/samples, but that filter isn't getting applied in study view.

For the text in the menu & the banner, I'm wondering if a slightly change in wording might be clearer. Below is the current text and then my proposed change. Which is clearer to you @adamabeshouse @jjgao ?

Exclude samples that are not profiled for one or more genes in any of the profiles (current)
Exclude samples that are not profiled for all genes in all profiles (proposal)

433 patients/samples are excluded from analysis since they are not profiled for one or more queried genes in one of the selected profiles. (current)
433 patients/samples are excluded from analysis since they are not profiled for all genes in all profiles. (proposal)

jjgao:
@tmazor's proposal sounds good to me!

adamabeshouse:
@tmazor @jjgao okay I made those changes, except I still added the word "queried profiles" because I think that's clearer. Should I take that out?

tmazor:
@adamabeshouse you're totally right - and actually it should probably say "queried genes" in addition to "queried profiles"

PR #3268, opened 2020-06-09
results view comparison tab - improve description of altered/unaltered groups and put them in tooltips

adamabeshouse:
image

### DIFF
--- a/src/pages/groupComparison/ClinicalData.tsx
+++ b/src/pages/groupComparison/ClinicalData.tsx
@@ -325,7 +325,7 @@ export default class ClinicalData extends React.Component<
 
     private readonly groupMembershipAxisData = remoteData({
         await: () => [
-            this.props.store.sampleSet,
+            this.props.store.sampleMap,
             this.props.store.activeGroups,
         ],
         invoke: async () => {
@@ -339,7 +339,7 @@ export default class ClinicalData extends React.Component<
                 categoryOrder,
             } as IStringAxisData;
             const sampleSet =
-                this.props.store.sampleSet.result ||
+                this.props.store.sampleMap.result ||
                 new ComplexKeyMap<Sample>();
 
             const sampleKeyToGroupSampleData = _.reduce(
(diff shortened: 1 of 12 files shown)


adamabeshouse:
@inodb @jjgao does this address the issue well enough?

adamabeshouse:
@jjgao this is about showing the altered/unaltered group in the oncoprint (https://github.com/cbioportal/cbioportal/issues/6893) there was discussion about making it a clinical track. so I made it so that comparison groups (of which altered/unaltered will always be two) will be available in the oncoprint now

jjgao:
I see. Ideally "altered vs unaltered" should be just one track, which can be loaded by default... @inodb thoughts?

inodb:
Thanks @adamabeshouse! I think one track for "Altered" makes sense. By definition "Unaltered should be the inverse right?

I like that this makes it easy to find patients where not all samples are altered (the Mixed group):

Screen Shot 2020-06-19 at 4 27 08 PM

I'm actually not sure if there's another way to easily do this currently.

In regards to the original issue: I mostly meant to address that people don't always immediately grasp what altered vs unaltered is on the comparison tab. E.g. when you open comparison tab here it's not obvious to everyone that altered means all cases/samples with a mutation in one of the query genes (or oql) 

Screen Shot 2020-06-19 at 4 33 07 PM

Maybe a better way to address that particular issue is to change the tooltip of altered/unaltered on the comparison tab to be something like:

`Altered group (x/n). Number of cases or samples with an alteration in (the OQL specification of) your queried genes`

`Unaltered group (x/n). Number of cases or samples without an alteration in (the OQL specification of) your queried genes`

adamabeshouse:
I like the tooltip idea.

jjgao:
Tooltip is a good idea.

adamabeshouse:
@jjgao @inodb is this feature as i've implemented it still useful or should i scrap it?

adamabeshouse:
@inodb @jjgao how does it look now?

jjgao:
@adamabeshouse would you please change the tooltip a bit?

Altered group (x/n) - samples with at least one alteration in your queried genes in the selected profiles

Unaltered group (x/n) - samples without any alteration in your queried genes in the selected profiles

jjgao:
I think adding the gene/altered/untered "Groups" in the OncoPrint menu can be potentially confusing (e.g. a user can create a group called KRAS from study view) and I don't see much use cases. Maybe we should remove them instead? @inodb thoughts?

adamabeshouse:
okay, how does it look now @jjgao @inodb

PR #3270, opened 2020-06-09
Add structural variant table in patient page

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7545

Screen Shot 2020-06-09 at 5 56 27 PM

To test

https://cbioportal-fix-7545-z0u0djy00p.herokuapp.com and set `localStorage.netlify = "deploy-preview-3270--cbioportalfrontend"`

Example, sample having structural variant data
https://cbioportal-fix-7545-z0u0djy00p.herokuapp.com/patient?studyId=msk_impact_2017&caseId=P-0000373

### DIFF
--- a/packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
@@ -76,6 +76,19 @@ export function generateQueryVariant(
     return query;
 }
 
+export function generateQueryStructuralVariantId(
+    site1EntrezGeneId: number,
+    site2EntrezGeneId: number,
+    tumorType: string | null
+): string {
+    let id = `${site1EntrezGeneId}_${site2EntrezGeneId}`;
+    if (tumorType) {
+        id = `${id}_${tumorType}`;
+    }
+
+    return id.trim().replace(/\s/g, '_');
+}
+
 export function generateQueryVariantId(
     entrezGeneId: number,
     tumorType: string | null,
@@ -221,6 +234,45 @@ export function generateAnnotateStructuralVariantQuery(
         structuralVariantType: 'FUSION',
         functionalFusion: false,
         tumorType: tumorType,
+        evidenceTypes: evidenceTypes,
+    } as AnnotateStructuralVariantQuery;
+}
+
+export enum StructuralVariantType {
+    DELETION = 'DELETION',
+    TRANSLOCATION = 'TRANSLOCATION',
+    DUPLICATION = 'DUPLICATION',
+    INSERTION = 'INSERTION',
+    INVERSION = 'INVERSION',
+    FUSION = 'FUSION',
+    UNKNOWN = 'UNKNOWN',
+}
+
+export function generateAnnotateStructuralVariantQueryFromGenes(
+    site1EntrezGeneId: number,
+    site2EntrezGeneId: number,
+    tumorType: string | null,
+    structuralVariantType: keyof typeof StructuralVariantType,
+    evidenceTypes?: EvidenceType[]
+): AnnotateStructuralVariantQuery {
+    return {
+        id: generateQueryStructuralVariantId(
+            site1EntrezGeneId,
+            site2EntrezGeneId,
+            tumorType
+        ),
+        geneA: {
+            entrezGeneId: site1EntrezGeneId,
+        },
+        geneB: {
+            entrezGeneId: site2EntrezGeneId,
+        },
+        structuralVariantType,
+        functionalFusion: !(
+            site1EntrezGeneId === site2EntrezGeneId ||
+            site2EntrezGeneId === null
+        ),
+        tumorType,
         evidenceTypes: evidenceTypes,
     } as AnnotateStructuralVariantQuery;
 }
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -78,6 +78,7 @@ import {
 import ResourcesTab, { RESOURCES_TAB_NAME } from './resources/ResourcesTab';
 import { MakeMobxView } from '../../shared/components/MobxView';
 import ResourceTab from '../../shared/components/resources/ResourceTab';
+import PatientViewStructuralVariantTable from './structuralVariant/PatientViewStructuralVariantTable';
 
 export interface IPatientViewPageProps {
     params: any; // react route
@@ -192,13 +193,6 @@ export default class PatientViewPage extends React.Component<
             },
             { fireImmediately: true }
         );
-
-        this.onMutationTableColumnVisibilityToggled = this.onMutationTableColumnVisibilityToggled.bind(
-            this
-        );
-        this.onCnaTableColumnVisibilityToggled = this.onCnaTableColumnVisibilityToggled.bind(
-            this
-        );
     }
 
     componentDidMount() {
@@ -266,7 +260,9 @@ export default class PatientViewPage extends React.Component<
         );
     }
 
-    @action private onCnaTableColumnVisibilityToggled(
+    @autobind
+    @action
+    private onCnaTableColumnVisibilityToggled(
         columnId: string,
         columnVisibility?: IColumnVisibilityDef[]
     ) {
@@ -277,7 +273,9 @@ export default class PatientViewPage extends React.Component<
         );
     }
 
-    @action private onMutationTableColumnVisibilityToggled(
+    @autobind
+    @action
+    private onMutationTableColumnVisibilityToggled(
         columnId: string,
         columnVisibility?: IColumnVisibilityDef[]
     ) {
@@ -372,42 +370,6 @@ export default class PatientViewPage extends React.Component<
         this.patientViewPageStore.copyNumberTableGeneFilterOption = option;
     }
 
-    mutationTableShowGeneFilterMenu(sampleIds: string[]): boolean {
-        const entrezGeneIds: number[] = _.uniq(
-            _.map(
-                this.patientViewPageStore.mergedMutationDataIncludingUncalled,
-                mutations => mutations[0].entrezGeneId
-            )
-        );
-        return (
-            sampleIds.length > 1 &&
-            checkNonProfiledGenesExist(
-                sampleIds,
-                entrezGeneIds,
-                this.patientViewPageStore.sampleToMutationGenePanelId.result,
-                this.patientViewPageStore.genePanelIdToEntrezGeneIds.result
-            )
-        );
-    }
-
-    cnaTableShowGeneFilterMenu(sampleIds: string[]): boolean {
-        const entrezGeneIds: number[] = _.uniq(
-            _.map(
-                this.patientViewPageStore.mergedDiscreteCNAData,
-                alterations => alterations[0].entrezGeneId
-            )
-        );
-        return (
-            sampleIds.length > 1 &&
-            checkNonProfiledGenesExist(
-                sampleIds,
-                entrezGeneIds,
-                this.patientViewPageStore.sampleToDiscreteGenePanelId.result,
-                this.patientViewPageStore.genePanelIdToEntrezGeneIds.result
-            )
-        );
-    }
-
     @autobind
     @action
     toggleGenePanelModal(genePanelId?: string | undefined) {
@@ -421,29 +383,6 @@ export default class PatientViewPage extends React.Component<
         return this.patientViewPageStore.genePanelIdToPanel.result[
             this.genePanelModal.genePanelId
         ];
-    }
-    @computed get sampleManager() {
-        if (
-            this.patientViewPageStore.patientViewData.isComplete &&
-            this.patientViewPageStore.studyMetaData.isComplete
-        ) {
-            const patientData = this.patientViewPageStore.patientViewData
-                .result;
-
-            if (
-                this.patientViewPageStore.clinicalEvents.isComplete &&
-                this.patientViewPageStore.clinicalEvents.result!.length > 0
-            ) {
-                return new SampleManager(
-                    patientData.samples!,
-                    this.patientViewPageStore.clinicalEvents.result
-                );
-            } else {
-                return new SampleManager(patientData.samples!);
-            }
-        } else {
-            return null;
-        }
     }
 
     readonly resourceTabs = MakeMobxView({
@@ -522,7 +461,10 @@ export default class PatientViewPage extends React.Component<
     }
 
     public render() {
-        const sampleManager = this.sampleManager;
+        let sampleManager: SampleManager | null = null;
+        if (this.patientViewPageStore.sampleManager.isComplete) {
+            sampleManager = this.patientViewPageStore.sampleManager.result!;
+        }
         let sampleHeader: (JSX.Element | undefined)[] | null = null;
         let cohortNav: JSX.Element | null = null;
         let studyName: JSX.Element | null = null;
@@ -1381,6 +1323,14 @@ export default class PatientViewPage extends React.Component<
                                                 />
                                             </div>
                                         )}
+
+                                    <hr />
+                                    <PatientViewStructuralVariantTable
+                                        store={this.patientViewPageStore}
+                                        onSelectGenePanel={
+                                            this.toggleGenePanelModal
+                                        }
+                                    />
                                 </MSKTab>
                                 {this.patientViewPageStore.sampleIds.length >
                                     1 &&
@@ -1472,7 +1422,10 @@ export default class PatientViewPage extends React.Component<
                                     <div>
                                         <ResourcesTab
                                             store={this.patientViewPageStore}
-                                            sampleManager={this.sampleManager}
+                                            sampleManager={
+                                                this.patientViewPageStore
+                                                    .sampleManager.result!
+                                            }
                                             openResource={this.openResource}
                                         />
                                     </div>
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -15,6 +15,7 @@ import {
     ResourceData,
     Sample,
     SampleMolecularIdentifier,
+    StructuralVariantFilter,
 } from 'cbioportal-ts-api-client';
 import client from '../../../shared/api/cbioportalClientInstance';
 import internalClient from '../../../shared/api/cbioportalInternalClientInstance';
@@ -67,8 +68,6 @@ import {
     fetchCopyNumberSegments,
     fetchCosmicData,
     fetchDiscreteCNAData,
-    fetchGenePanel,
-    fetchGenePanelData,
     fetchGisticData,
     fetchMutationalSignatureData,
     fetchMutationalSignatureMetaData,
@@ -95,6 +94,8 @@ import {
     noGenePanelUsed,
     ONCOKB_DEFAULT,
     getGenomeNexusUrl,
+    generateStructuralVariantId,
+    fetchStructuralVariantOncoKbData,
 } from 'shared/lib/StoreUtils';
 import {
     fetchCivicGenes,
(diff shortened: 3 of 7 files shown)


packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
```diff
@@ -224,3 +237,39 @@ export function generateAnnotateStructuralVariantQuery(
         evidenceTypes: evidenceTypes,
     } as AnnotateStructuralVariantQuery;
 }
+
+export function generateAnnotateStructuralVariantQueryFromGenes(
+    site1EntrezGeneId: number,
+    site2EntrezGeneId: number,
+    tumorType: string | null,
+    structuralVariantType:
+        | 'DELETION'
```
alisman, line 246: seems like this ought to exposed for use elsewhere

alisman: i mean the variant type enum

packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
```diff
@@ -224,3 +237,39 @@ export function generateAnnotateStructuralVariantQuery(
         evidenceTypes: evidenceTypes,
     } as AnnotateStructuralVariantQuery;
 }
+
+export function generateAnnotateStructuralVariantQueryFromGenes(
```
alisman, line 241: should this be generateAnnotatedStruct...  (see the "d" at end of annotated)

alisman: and if so, seems like there are other places where you use this name

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -1364,6 +1317,7 @@ export default class PatientViewPage extends React.Component<
                                                     }
                                                     onFilterGenes={
                                                         this
+                                                            .patientViewPageStore
```
alisman, line 1320: are you sure this handler belongs on store?

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+            columns.push({
+                name: 'Samples',
+                render: (d: StructuralVariant[]) =>
+                    TumorColumnFormatter.renderFunction(
+                        d.map(datum => ({
+                            sampleId: datum.sampleId,
+                            entrezGeneId: datum.site1EntrezGeneId,
+                        })),
+                        this.props.store.sampleManager.result!,
+                        this.props.store.sampleToStructuralVariantGenePanelId
```
alisman, line 56: just want to make sure this is awaited somewhere.

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+                            }
+                        />
+                    </HeaderIconMenu>
+                );
+            },
+            visible: true,
+            order: 35,
+        });
+
+        const GenePanelProps = (d: StructuralVariant[]) => ({
```
alisman, line 159: no initial caps?

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+            name: 'Variant Class',
+            render: (d: StructuralVariant[]) => (
+                <span>{d[0].variantClass}</span>
+            ),
+            filter: (
+                d: StructuralVariant[],
+                filterString: string,
+                filterStringUpper: string
+            ) => {
+                return d[0].variantClass.indexOf(filterStringUpper) > -1;
```
alisman, line 224: i know we do this in a lot of places already, but i dislike depending on case sensitive detection.     devs need to know that variantClass is always upper case.  instead we should just use RegExp(... , i).test

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+            this.props.store.sampleManager,
+            this.props.store.structuralVariantTableShowGeneFilterMenu,
+            this.props.store.sampleToStructuralVariantGenePanelId,
+            this.props.store.genePanelIdToEntrezGeneIds,
+        ],
+        render: () => {
+            if (
+                this.props.store.structuralVariantProfileId.result === undefined
+            ) {
+                return (
```
alisman, line 377: can we be sure this message won't show in pending state?  i guess if mobx promise has default.  but would be better to check for both isComplete and result undefined (if indeed the result can be undefined when call has completed

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,179 @@
+import * as React from 'react';
```
alisman, line 1: typeo in folder name (structuralVairnt)

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,179 @@
...
+
+        if (structuralVariantData) {
+            let oncoKbIndicator: IndicatorQueryResp | undefined = undefined;
+            let oncoKbStatus: IAnnotation['oncoKbStatus'] = 'complete';
+
+            let oncoKbGeneExist = false;
+            let isOncoKbCancerGene = false;
+            if (oncoKbCancerGenes && !(oncoKbCancerGenes instanceof Error)) {
+                oncoKbGeneExist =
+                    _.find(
```
alisman, line 39: _.some will return boolean and free you from testing against undefined

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,179 @@
...
+                    oncoKbData.status === 'complete'
+                ) {
+                    oncoKbIndicator = this.getIndicatorData(
+                        structuralVariantData,
+                        oncoKbData.result,
+                        uniqueSampleKeyToTumorType,
+                        studyIdToStudy
+                    );
+                }
+                oncoKbStatus = oncoKbData ? oncoKbData.status : 'pending';
```
alisman, line 78: are we sure it's really pending if that thing doesn't exist?

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,179 @@
...
+            if (oncoKbData && oncoKbData.result instanceof Error) {
+                oncoKbStatus = 'error';
+            } else if (oncoKbGeneExist) {
+                // actually, oncoKbData.result shouldn't be an instance of Error in this case (we already check it above),
+                // but we need to check it again in order to avoid TS errors/warnings
+                if (
+                    oncoKbData &&
+                    oncoKbData.result &&
+                    !(oncoKbData.result instanceof Error) &&
+                    oncoKbData.status === 'complete'
```
alisman, line 69: does it not have isComplete property?

src/shared/lib/StoreUtils.ts
```diff
@@ -799,6 +768,44 @@ export async function fetchCnaOncoKbData(
...
+                    datum.site1EntrezGeneId,
+                    datum.site2EntrezGeneId,
+                    cancerTypeForOncoKb(
+                        datum.uniqueSampleKey,
+                        uniqueSampleKeyToTumorType
+                    ),
+                    datum.variantClass as any
+                );
+            }),
+            'id'
```
alisman, line 803: no string shortcuts.   (n)=>n.id

src/shared/lib/StoreUtils.ts
```diff
@@ -799,6 +768,44 @@ export async function fetchCnaOncoKbData(
...
+        structuralVariantData.result.length === 0
+    ) {
+        return ONCOKB_DEFAULT;
+    } else {
+        const alterationsToQuery = _.filter(
+            structuralVariantData.result,
+            d =>
+                d.site1EntrezGeneId &&
+                d.site2EntrezGeneId &&
+                (!!annotatedGenes[d.site1EntrezGeneId] ||
```
alisman, line 788: i think you can just use ```annotatedGenes[...]``` (no !!) because the values of that object are not falsy

src/shared/lib/StoreUtils.ts
```diff
@@ -939,6 +946,20 @@ export async function queryOncoKbCopyNumberAlterationData(
     return toOncoKbData(oncokbSearch);
 }
 
+export async function queryOncoKbStructuralVariantData(
```
alisman, line 949: i think we usually call this async data fetchers "fetch..."

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -267,6 +274,24 @@ export class PatientViewPageStore {
         this.openResourceTabMap.set(resourceId, open);
     }
 
+    @autobind
```
alisman, line 277: i think these bound actions should really be on view, not on store.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -311,14 +340,36 @@ export class PatientViewPageStore {
...
-            if (profile) {
-                return profile.molecularProfileId;
+            if (structuralVariantProfiles.length > 0) {
+                return structuralVariantProfiles[0];
+            }
+            return undefined;
+        },
+    });
+
+    readonly structuralVariantProfileId = remoteData({
```
alisman, line 368: seems kind silly to make a whole mobxpromise wrapper just to reference the id.  any calling code will still have to check the status of this, so we don't save really anything

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -740,6 +791,26 @@ export class PatientViewPageStore {
         invoke: async () => client.getStudyUsingGET({ studyId: this.studyId }),
     });
 
+    public sampleManager = remoteData<SampleManager>({
```
alisman, line 794: do we not already have a sampleManager on patient page somewhere?  should basically be a singleton unless there's something i'm missing.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1227,6 +1412,32 @@ export class PatientViewPageStore {
...
+    readonly structuralVariantOncoKbData = remoteData<IOncoKbData>(
+        {
+            await: () => [
+                this.oncoKbAnnotatedGenes,
+                this.structuralVariantData,
+                this.clinicalDataForSamples,
+                this.studies,
+            ],
+            invoke: async () => {
+                if (AppConfig.serverConfig.show_oncokb) {
```
alisman, line 1424: i would hope that we should check this property outside of this resource and not reference this promise if it was false.    i.e. this promise should not know about an app setting

src/shared/components/mutationTable/column/ChromosomeColumnFormatter.tsx
```diff
@@ -4,12 +4,12 @@ import { Mutation } from 'cbioportal-ts-api-client';
  * @author Selcuk Onur Sumer
  */
 export default class ChromosomeColumnFormatter {
-    public static getSortValue(data: Mutation[]): number | null {
-        const chromosome = ChromosomeColumnFormatter.getData(data);
+    public static getSortValue(data: Pick<Mutation, 'chr'>[]): number | null {
+        const chromosome = this.getData(data);
```
adamabeshouse, line 8: Im surprised this works, I thought `this` wasnt allowed in static functions?

src/shared/lib/StoreUtils.ts
adamabeshouse: lol I read this comment and was so confused for a second until i realized you meant "no `!!`" and not NO!! 😆

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
```diff
@@ -0,0 +1,403 @@
...
+    StructuralVariant[]
+> {}
+
+@observer
+export default class PatientViewStructuralVariantTable extends React.Component<
+    IPatientViewStructuralVariantTableProps,
+    {}
+> {
+    constructor(props: IPatientViewStructuralVariantTableProps) {
+        super(props);
```
adamabeshouse, line 39: unnecessary constructor

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
adamabeshouse: yes as a matter of fact i think its better for `columns` to be a `remoteData` and then just `await` it down in `tableUI` before using it

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
adamabeshouse: yes or just do variantClass.toUpperCase().indexOf

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
adamabeshouse: I agree ^^

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
adamabeshouse: agreed

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -959,7 +1044,7 @@ export class PatientViewPageStore {
         []
     );
 
-    readonly coverageInformation = remoteData<CoverageInformation>(
+    readonly genePanelData = remoteData(
```
adamabeshouse, line 1047: lets keep the name `coverageInformation` to align w our convention?

packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
kalletlak: `AnnotateStructuralVariantQuery` is the type. Do you still want me to change it?

src/pages/patientView/PatientViewPage.tsx
kalletlak: Actions on structural variant table are in store. I thought it better to have similar filters at one place

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
kalletlak: Yes, it is awaited at other place. I was following the same model as other tables in patient page. Yeah, even I prefer to be `remoteData`. I'll update it

src/pages/patientView/structuralVairnt/PatientViewStructuralVariantTable.tsx
kalletlak: `groupedStructuralVariantData` is in await, which already awaits for `structuralVariantProfileId`. Wouldn't it be redundant to add again?

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
kalletlak: yeah, I tried to copy code from existing annotation file which is used for cna table https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx. I'll update it

src/pages/patientView/structuralVairnt/column/AnnotationColumnFormatter.tsx
kalletlak: Looks like it is. Ref: https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/copyNumberAlterations/column/AnnotationColumnFormatter.tsx#L86

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
kalletlak: Actually I copied it here so I can use it in structural variant table component. I forgot to remove it from patient page

src/shared/components/mutationTable/column/ChromosomeColumnFormatter.tsx
kalletlak: it is working. I think with the class it can be accessed in non-static way

src/pages/patientView/PatientViewPage.tsx
kalletlak: reverted back this code

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1522,6 +1692,35 @@ export class PatientViewPageStore {
...
 
+    readonly structuralVariantTableShowGeneFilterMenu = remoteData({
+        await: () => [
+            this.samples,
+            this.sampleToStructuralVariantGenePanelId,
+            this.genePanelIdToEntrezGeneIds,
+            this.groupedStructuralVariantData,
+        ],
+        invoke: () => {
+            // TODO: should we include site2EntrezGeneId?
```
zhx828, line 1703: I don' see why not?

src/pages/patientView/structuralVariant/column/AnnotationColumnFormatter.tsx
```diff
@@ -0,0 +1,177 @@
...
+                        oncoKbData.result,
+                        uniqueSampleKeyToTumorType,
+                        studyIdToStudy
+                    );
+                }
+                oncoKbStatus = oncoKbData ? oncoKbData.status : 'pending';
+            }
+
+            value = {
+                hugoGeneSymbol: structuralVariantData[0].site1HugoSymbol, //TODO: which site hugo gene symbol should be used?
```
zhx828, line 80: We probably need to redesign the interface for annotation, the oncokb card content needs to updated to since we need to validate both genes are not in oncokb.

zhx828: But for now, we should check both sites and pick the one with higher order. geneIsAnnotated>geneIsCancerGene>geneNotCancerGene

packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
alisman: if what it means is "a structural varianty query that has been annotated" then yes.

alisman: is it an API type?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -311,20 +328,31 @@ export class PatientViewPageStore {
...
-        await: () => [this.molecularProfilesInStudy],
+        await: () => [this.mutationMolecularProfile],
         invoke: async () => {
-            const profile = findMutationMolecularProfile(
-                this.molecularProfilesInStudy,
-                this.studyId
-            );
-            if (profile) {
-                return profile.molecularProfileId;
+            if (this.mutationMolecularProfile.result) {
```
alisman, line 333: you can assume that result is populated because it's in await.   so, we only need this check if undefined is a valid

packages/cbioportal-frontend-commons/src/lib/oncokb/OncoKbUtils.ts
kalletlak: yes

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -522,7 +461,7 @@ export default class PatientViewPage extends React.Component<
     }
 
     public render() {
-        const sampleManager = this.sampleManager;
+        const sampleManager = this.patientViewPageStore.sampleManager.result!;
```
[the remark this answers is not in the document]
kalletlak: sampleManager is used even outside if block

PR #3273, opened 2020-06-11
Show survival charts by default for pan-cancer studies

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7518


zhx828:
@kalletlak thanks! But I'm surprised that there is no study view screenshot failing. Can we add one if not exists?

PR #3274, opened 2020-06-11
(Front-end) Update endpoint name for generic-assay

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/6955

Update reference of generic-assay endpoint.


inodb:
Thanks for updating this! I've created a new issue for this #7647. I think we should think a bit more about the naming and the structure before merging. We can re-open this one later

PR #3280, opened 2020-06-15
ASCN E2E Tests

averyniceday:
- Added E2E Tests
- Requested changes by Niki
	- ASCN Copy # Icon
	- CCF Significant Digits
	- Rename Clonal to Clonality

This is going fail here because this PR is to the feature branch. But previously we were testing by making a PR to master and seeing if they would complete. This is a screenshot of the CircleCI tests.

Screen Shot 2020-06-15 at 4 31 24 PM

This is the same code, squashed and rebased. 

Co-authored-by: Manda Wilson 1458628+mandawilson@users.noreply.github.com
Co-authored-by: Avery Wang 18199796+averyniceday@users.noreply.github.com

## Checks

## Any screenshots or GIFs?

## Notify reviewers


averyniceday:
@zhx828 it's mapping elements to the array of sampleIds coming from the props

env/custom.sh
```diff
@@ -1,2 +0,0 @@
-# set e.g.
```
onursumer, line 1: we should not be deleting this one

end-to-end-test/local/specs/core/patientview.screenshot.spec.js
```diff
@@ -6,16 +6,46 @@ var assertScreenShotMatch = require('../../../shared/lib/testUtils')
...
+    CBIOPORTAL_URL + '/patient?studyId=ascn_test_study&caseId=FAKE_P001';
 
 describe('patient view page', function() {
+    describe('mutation table for study with ascn data', () => {
+        beforeEach(() => {
+            goToUrlAndSetLocalStorage(ascnPatientViewUrl);
+            waitForPatientView();
+        });
+
+        it('ascn columns are displayed', () => {
```
onursumer, line 21: Convention for the `it` function is that it reads like a proper sentence, so while `it ascn columns are displayed` is not a proper sentence `it displays ascn columns` is. This comment applies most of the `it` functions introduced in this PR.

averyniceday: @onursumer I'll update this today!

env/custom.sh
averyniceday: ditto here

zhx828:
> @zhx828 it's mapping elements to the array of sampleIds coming from the props

Does it mean the order has no meaning?

averyniceday:
@zhx828 yeah, the order should just match the order of the samples. Want it to match the `1,2,3,4,5,6..` sample icons in the first column when there's multiple samples per patient.

zhx828:
> @zhx828 yeah, the order should just match the order of the samples. Want it to match the `1,2,3,4,5,6..` sample icons in the first column when there's multiple samples per patient.

I see. It's not so intuitive especially when you have the two columns very far away from each other. But that's not in the scope of this pr.

averyniceday:
@zhx828 yeah, we can always refine down the line 😄  Right now we're running this through Niki and Mark (his group is planning on looking at FACETS data). The original idea was to follow the allele freq column as an example - it's using a bar chart where each bar is a sample.

PR #3281, opened 2020-06-16
Add Mutation Aligner cache

onursumer:
Fix cBioPortal/cbioportal#4909

## Checks

### DIFF
--- a/packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
+++ b/packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
@@ -1,0 +1,31 @@
+import { action, observable } from 'mobx';
+
+import { CacheData, MobxCache } from '../model/MobxCache';
+
+export abstract class DefaultStringQueryCache<D>
+    implements MobxCache<D, string> {
+    protected _cache = observable.shallowMap<CacheData>();
+
+    @action
+    public get(query: string) {
+        if (!this._cache.get(query)) {
+            this._cache.set(query, { status: 'pending' });
+
+            this.fetch(query)
+                .then((d: D) =>
+                    this._cache.set(query, { status: 'complete', data: d })
+                )
+                .catch(() => this._cache.set(query, { status: 'error' }));
+        }
+
+        const data = this._cache.get(query);
+
+        return data ? data : null;
+    }
+
+    public abstract fetch(query: string): Promise<D>;
+
+    public get cache() {
+        return this._cache.toJS();
+    }
+}
--- a/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
+++ b/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
@@ -40,6 +40,7 @@ import {
 import TrackPanel from '../track/TrackPanel';
 
 import './lollipopMutationPlot.scss';
+import DomainTooltip from '../lollipopPlot/DomainTooltip';
 
 const DEFAULT_PROTEIN_LENGTH = 10;
 
@@ -47,6 +48,7 @@ export type LollipopMutationPlotProps = {
     store: MutationMapperStore;
     controlsConfig?: LollipopPlotControlsConfig;
     pubMedCache?: MobxCache;
+    mutationAlignerCache?: MobxCache<string>;
     getLollipopColor?: (mutations: Partial<Mutation>[]) => string;
     getMutationCount?: (mutation: Partial<Mutation>) => number;
     topYAxisSymbol?: string;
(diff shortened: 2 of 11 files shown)


packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
```diff
@@ -0,0 +1,33 @@
+import { action, observable } from 'mobx';
+
+import { CacheData, MobxCache } from '../model/MobxCache';
+
+export abstract class DefaultStringQueryCache<D>
```
onursumer, line 5: Abstracted and moved from `DefaultPubMedCache` for reusability purposes.

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -325,55 +327,6 @@ export default class LollipopMutationPlot extends React.Component<
         return specs;
     }
 
-    private mutationAlignerLink(pfamAccession: string): JSX.Element | null {
```
onursumer, line 328: Moved into a new `DomainTooltip` component

packages/react-mutation-mapper/src/component/lollipopPlot/DomainTooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+@observer
+export default class DomainTooltip extends React.Component<
+    IDomainTooltipProps,
+    {}
+> {
+    @computed
+    get mutationAlignerCacheData() {
+        return this.props.mutationAlignerCache &&
+            this.props.mutationAlignerCache.cache
+            ? this.props.mutationAlignerCache.cache[this.props.pfamDomainId]
```
onursumer, line 24: accessing the `cache` field is the only way to trigger rerendering for some reason, using the `get` function doesn't work.

packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
```diff
@@ -0,0 +1,33 @@
...
+
+        return data ? data : null;
+    }
+
+    public fetch(query: string): Promise<D> {
+        return Promise.reject('No Default Implementation');
+    }
+
+    public get cache() {
+        return this._cache.toJS();
```
zhx828, line 31: The key is to prevent the cache to be updated?

packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
```diff
@@ -0,0 +1,33 @@
...
+        }
+
+        const data = this._cache.get(query);
+
+        return data ? data : null;
+    }
+
+    public fetch(query: string): Promise<D> {
+        return Promise.reject('No Default Implementation');
+    }
```
zhx828, line 28: Can we use abstract here so all implementations need to override this method? https://stackoverflow.com/questions/13333489/declaring-abstract-method-in-typescript

packages/react-mutation-mapper/src/component/lollipopPlot/DomainTooltip.tsx
```diff
@@ -0,0 +1,71 @@
...
+            : undefined;
```
zhx828, line 25: Here the default is undefined, but the cache manager returns null as default if cache is not available. Not a big deal, but we should consolidate the defaults.

packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
onursumer: Oh, I should have known that TS was as advanced as Java now :)

packages/react-mutation-mapper/src/cache/DefaultStringQueryCache.ts
onursumer: Previously we were returning the cache as is, but that is a `Map` instance, so it doesn't actually match the function signature, we need to return a proper JS object. Not sure if this is the best way to do it though.

packages/react-mutation-mapper/src/component/lollipopPlot/DomainTooltip.tsx
onursumer: good point, I haven't even noticed that

inodb:
@onursumer I'm actually wondering if we even need this service? It only returns this for instance for each domain:

```
{linkToMutationAligner: "http://mutationaligner.org/domains/PF00757"}
```

see also:
https://github.com/cBioPortal/cbioportal/blob/master/core/src/main/java/org/mskcc/cbio/portal/servlet/MutationAlignerServlet.java

I think it just checks whether it gets a 200 response code to determine if it should add a mutation aligner link. Why don't we do this on the frontend? And maybe only when you mouse over the actual domain?

onursumer:
@inodb We don't need `MutationAlignerServlet.java`, we can replace it with the proxy and move the check to the frontend. We cannot actually directly use `http://mutationaligner.org/api/domains/` on the frontend because it is not a secure service. I was actually thinking to create a separate PR to address that part (cBioPortal/cbioportal#6743) once we merge this one.

This PR actually addresses the `when you mouse over the actual domain` part. No prior call to the mutation aligner service until you mouse over a domain.

inodb:
@onursumer gotcha - ok that makes sense! I remember now you mentioned the SSL issue

onursumer:
Failing tests are unrelated, merging.

PR #3285, opened 2020-06-18
Implement heatmap tracks in oncoprinter, and export heatmaps to oncoprinter from results page

adamabeshouse:


jjgao:
@adamabeshouse Looks good. Thanks!

For the format text, can we still do it in the old way, ie. expand and collapse?

adamabeshouse:
@jjgao we had a whole long convo about that the other day and decided on this 😅  is it okay to leave it like this and change it if theres complaints?

jjgao:
@adamabeshouse Do you remember the reasoning that it was decided on this? 

I don't feel strongly but would like to listen to input from others.

adamabeshouse:
@jjgao I think the main idea was that it may be easier to use if you can open the instructions in a new window. But I don't remember very well. @inodb @alisman do you remember?

adamabeshouse:
@tmazor how does it look to you?

tmazor:
The new layout looks great!

I think I have a slight preference for the old approach of expand/collapse for the format descriptions rather than a separate page, but I don't feel super strongly either way. Totally ok with me to leave as is with the new page.

Just one comment about the heatmaps - can the legend name be the same as the track name? When I add the example data, the legend says "Heatmap" 3 times but it would be clearer if they had the same name as the data tracks:
image

adamabeshouse:
Great, thanks @tmazor ! the only concern with that is probably often people will have more than one heatmap track on the same color scale. In the results view, the legend will have the name of the alteration type (e.g. Expression Heatmap), as opposed to the names of the tracks (e.g. KRAS / PTEN / BRAF). I guess we could make it different for the special case of one heatmap track but... idk. What do you think?

tmazor:
Ohhh I see, that makes sense. Maybe it's ok to leave as is - the 3 color schemes are different enough that's it's fairly obvious which is which, and this is user-provided data so they should be at least a little familiar with the data they provided. 

Or, what about naming the legend similar to the data format? e.g. Heatmap (z-scores), Heatmap (0-1), Heatmap

adamabeshouse:
I think naming by color scale type, as you suggested, would work well!

adamabeshouse:
@tmazor okay, I made all those changes!

tmazor:
looks good to me @adamabeshouse !

src/pages/staticPages/tools/oncoprinter/styles.module.scss
```diff
@@ -1,17 +1,16 @@
...
+    max-width: 800px;
 
     .dataFormatHelpBorder {
         border-left: 1px solid black;
         margin-left: 15px;
         margin-right: 15px;
     }
 }
+
+.inputSection {
```
alisman, line 15: ?

src/pages/staticPages/tools/oncoprinter/OncoprinterClinicalAndHeatmapUtils.ts
```diff
@@ -0,0 +1,572 @@
...
+
+type OncoprinterHeatmapTrackSpec = {
+    trackName: string;
+    datatype: HeatmapTrackDataType;
+};
+
+type OncoprinterClinicalTrackDatum = Pick<
+    ClinicalTrackDatum,
+    'attr_id' | 'uid' | 'attr_val_counts' | 'na' | 'attr_val'
+> & { sample: string };
```
alisman, line 36: sampleId?

src/pages/staticPages/tools/oncoprinter/OncoprinterClinicalAndHeatmapUtils.ts
```diff
@@ -0,0 +1,572 @@
...
+type OncoprinterHeatmapTrackDatum = Pick<
+    IBaseHeatmapTrackDatum,
+    'profile_data' | 'uid' | 'na'
+> & { sample: string };
+
+// Output mimics mobxpromise form but thats just because its a nice way to package status and result.
+// This isn't meant to be plugged into mobxpromise machinery i.e. with `await`
+type ParseInputResult<TrackSpecType> =
+    | {
+          status: 'complete';
```
alisman, line 47: 'complete' implies to me some kind of async process and could lead to confusion given mobxpromise.    could we use 'parsedSucess' or 'success' or 'valid' ?

src/pages/staticPages/tools/oncoprinter/OncoprinterConstants.ts
```diff
@@ -592,373 +597,406 @@ TCGA-25-2392-01
...
-TCGA-25-1877-01\t43\t${ONCOPRINTER_CLINICAL_VAL_NA}\t24\t${ONCOPRINTER_CLINICAL_VAL_NA}
-TCGA-25-1878-01\t33\tProstate\t104\t${ONCOPRINTER_CLINICAL_VAL_NA}
-TCGA-25-2042-01\t50\tProstate\t49\t${ONCOPRINTER_CLINICAL_VAL_NA}
-TCGA-25-2390-01\t21\tProstate\t17\t${ONCOPRINTER_CLINICAL_VAL_NA}
-TCGA-25-2396-01\t51\tBreast\t1\t${ONCOPRINTER_CLINICAL_VAL_NA}
-TCGA-25-2397-01\t33\tBreast\t40\t${ONCOPRINTER_CLINICAL_VAL_NA}
-`;
+// generate using seeded random number so its the same every time but we don't have to type it out manually
+const rng = new DeterministicRNG();
+const samples = [
```
alisman, line 602: exampleSampleList?

src/pages/staticPages/tools/oncoprinter/OncoprinterClinicalAndHeatmapUtils.ts
adamabeshouse: I'm just going along with the existing oncoprint type name which is `sample`, this just makes `sample` mandatory whereas in the oncoprint data type its optional

src/pages/staticPages/tools/oncoprinter/OncoprinterClinicalAndHeatmapUtils.ts
adamabeshouse: good point, I will change that

src/pages/staticPages/tools/oncoprinter/OncoprinterConstants.ts
adamabeshouse: sure

PR #3286, opened 2020-06-18
Update message on PlotsTab.tsx with more information

jagnathan:
Update PlotsTab.tsx by Setting flag isUnaffected same as Mutations.tsx so that messages are identical

Fix cBioPortal/cbioportal#6059 

- isUnaffected flag is set to the same value as Mutations.tsx
- Based on the isUnaffected flag the message shown is identical to the message shown in Mutations tab


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -4555,6 +4555,9 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                         className="plots-oql-status-banner"
                         store={this.props.store}
                         tabReflectsOql={false}
+                        isUnaffected={
```
alisman, line 4558: @adamabeshouse what is the difference between tabReflectsOql and isUnaffected props?

alisman:
@jagnathan this looks good.  i tested it.  i wonder though about the code which was producing the old message with the inconsistent icon.  Can you investigate?  Should that code be removed?  This may dovetail with the question I posed to @adamabeshouse about the seemingly redundant props.

src/pages/resultsView/plots/PlotsTab.tsx
adamabeshouse: the naming is not great. It's for the use case of toggling oql filtering in a tab. If isUnaffected, then the tab is not affected by oql no matter what. If !isUnaffected then tabReflectsOql just means that the tab is currently not affected by oql. This difference is only meaningful if onToggle is set, meaning that the oql filtering can be toggled and there'll be a little button to do so.

inodb:
Thanks @jagnathan - as discussed at standup looks like I just didn't read the info message well enough, so it wasn't really a bug but a feature.

@adamabeshouse @alisman after re-reading the message I think it all makes sense. Maybe we can just close this and leave it as is

jagnathan:
hi @inodb, I can add to the message if needed. So that the message is self-explanatory.

inodb:
Thanks @jagnathan ! Maybe we can change it to this:

1. The results below **do not** reflect the OQL specification from your query.

2. The results below are unaffected by the OQL specification from your query.

instead of 

1. The results below do not reflect the OQL specification from your query.

2. The results below are not affected by the OQL specification from your query.

We could also maybe use a yellow banner for 1. instead of blue, like https://getbootstrap.com/docs/4.0/components/alerts/#examples. Tho maybe these colors were picked on purpose to not be as attention grabbing

What do you think @adamabeshouse ? `

adamabeshouse:
@inodb the colors were picked on purpose so that it wouldn't seem as if there was any error or need to take action to fix something, as yellow might suggest.

How about using the word "filter" instead of the vaguer word "affected". So 

1. The results below *are not filtered* by the OQL specification from your query.

2. The results below are unaffected by the OQL specification from your query.

@tmazor @jjgao @alisman ?

inodb:
Thanks @adamabeshouse! I see, in that case let's keep the coloring as is

I believe what you're proposing is to have these two messages:

1. The results below **do not reflect** the OQL specification from your query.
2. The results below **are not filtered** by the OQL specification from your query.

On quick read I think it is hard to see the difference between the two sentences because they look very similar and both use the word "not". Any ideas for a sentence where (2) does not use "not"?

E.g.

2. The results below are the same with and without applying the OQL specification from your query.

Then we can bold **do not** in (1) and no bolding in (2)

adamabeshouse:
@inodb no, I deliberately am only using the "filtered" word for tabs that have data thats affected by OQL, whether OQL affects the results or not (i.e. your sentence #1). I think "filter" captures that better. then "unaffected" (sentence #2) can mean "this data is irrelevant to your OQL". maybe not different enough but that's my thought.

inodb:
@adamabeshouse just listing out all the options so I can better understand. I think there are 5?

1. 

2. 

4. 

5. 

5. 

My question is how to make (2) and (5) above appear more distinct. Although the coloring is different I think the meaning of "not affected" wasn't immediately obvious to me without the context of all other options. In addition because the sentence structure is similar it was hard to spot that 2 & 5 were different. Maybe the sentence could be:

`Although your query contains OQL, the result below is unaffected i.e. it would be identical without the OQL specification.`

In some ways (5) is only used as a visual indicator to the user, that they are using OQL in the query. Another option might therefore be to remove the message.

Thoughts?

adamabeshouse:
Right. So my suggestion is to make the distinction between 1-4 and 5 clearer, use language like "filtered" instead of "reflected" in 1-4. I think 5 is useful because all the other tabs will have banners as well.

inodb:
@adamabeshouse I see! Ok got it, yeah that makes sense

inodb:
@adamabeshouse so basically this right:

1+3. The results below are filtered by the OQL specification from your query
2+4. The results below **are not filtered** by the OQL specification from your query 
5. The results below are not affected by the OQL specification from your query.

Do you like 5 as is or do you think it makes sense to elaborate what "not affected" means e.g.

`The results below are not affected by the OQL specification from your query i.e. they would be identical without the OQL specification.`

adamabeshouse:
@inodb yes, exactly. if "not affected" isn't clear enough I think we should choose a different word because having a full explanation seems kind of long/more confusing. Maybe 'Your OQL specification is not relevant to the results below"?

inodb:
@adamabeshouse I like that

inodb:
@jagnathan could you help update the PR per the discussion ☝️ . Please let us know if anything is unclear

adamabeshouse:
@jagnathan sorry I think we weren't clear enough - the language we're going with is "Your OQL specification is not relevant to the results below"

adamabeshouse:
@jagnathan @inodb I just had a talk with @alisman who felt the #5 message was confusing, and I think I agree. how about this?
"Your OQL specification does not affect the results below, because you have not specified any filters for mutations in this gene."

inodb:
Thanks @jagnathan !

PR #3287, opened 2020-06-19
update oncokb client

zhx828:


onursumer:
@zhx828 Should we also manually set the version to `1.1.0-beta.1`? Otherwise publish bot will publish `1.0.5` instead of `1.1.0`.

zhx828:
@onursumer updated.

PR #3290, opened 2020-06-22
Fix selection of custom driver annotations source in Results View config menu

pvannierop:
Fix cBioPortal/cbioportal#7616

# Problem
The selection of custom driver annotations in the configuration menu of Results View (see figure below) was broken  (checkboxes could not be de-selected). 

# Fix
This PR  fixes this feature and add serveral e2e-tests to make sure it will stay intact in the future.

Example of the custom driver selection in Results View configuration menu.
image

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -465,10 +465,10 @@ export class ResultsViewPageStore {
             cosmicCountThreshold: 0,
             driverTiers: observable.map<boolean>(),
 
-            _customBinary: undefined,
             _hotspots: false,
             _oncoKb: false,
             _excludeVUS: false,
+            _customBinary: undefined,
 
             set hotspots(val: boolean) {
                 this._hotspots = val;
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -500,13 +500,15 @@ export class ResultsViewPageStore {
                 return anySelected;
             },
 
+            set customBinary(val: boolean) {
```
alisman, line 503: maybe time improve this name.   have no idea what customBinary is.  i realize this is legacy code

pvannierop: @alisman  I agree that the name is very uninformative. `customBinary` refers to _custom_ diver annotations that can be used or not-used for VUS filtering, hence `Binary`. Refactoring of names is realy needed here, yes.

Sjoerd-van-Hagen:
@inodb can we get this merged?

inodb:
@pvannierop just want to double check: is this done?

src/pages/resultsView/settings/DriverAnnotationControls.tsx
```diff
@@ -107,6 +104,23 @@ export default class DriverAnnotationControls extends React.Component<
         }
     }
 
+    @autobind
+    private onCustomDriverTierCheckboxClick(
+        event: React.MouseEvent<HTMLInputElement>
+    ) {
+        let a = 1;
```
alisman, line 111: is a used?

pvannierop: @alisman No I have removed the line. I think that all e2e-tests that fail are not because of my code. Can we merge?

PR #3292, opened 2020-06-22
Replace getMutationAligner.json servlet with proxy

onursumer:
Fix cBioPortal/cbioportal#6743

## Checks

### DIFF
--- a/packages/react-mutation-mapper/src/util/DataFetcherUtils.ts
+++ b/packages/react-mutation-mapper/src/util/DataFetcherUtils.ts
@@ -15,7 +15,9 @@ import { OncoKbAPI, OncoKBInfo } from 'oncokb-ts-api-client';
 import _ from 'lodash';
 
 export const DEFAULT_MUTATION_ALIGNER_URL_TEMPLATE =
-    'https://www.cbioportal.org/getMutationAligner.json?pfamAccession=<%= pfamDomainId %>';
+    'http://mutationaligner.org/api/domains/<%= pfamDomainId %>?metadata=true';
+export const DEFAULT_MUTATION_ALIGNER_PROXY_URL_TEMPLATE =
+    'https://www.cbioportal.org/proxy/mutationaligner.org/api/domains/<%= pfamDomainId %>?metadata=true';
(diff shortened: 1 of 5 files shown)


src/shared/api/urls.ts
```diff
@@ -173,10 +173,6 @@ export function getUniprotIdUrl(swissProtAccession: string) {
     });
 }
 
-export function getMutationAlignerUrl() {
```
onursumer, line 176: Seems like this function was not used anymore.  Currently it always points to the public instance. Do we still need to worry about making the proxy url customizable per instance? Since this feature is not very important, maybe it is okay to just remove this.

packages/react-mutation-mapper/src/util/DataFetcherUtils.ts
```diff
@@ -15,7 +15,7 @@ import { OncoKbAPI, OncoKBInfo } from 'oncokb-ts-api-client';
 import _ from 'lodash';
 
 export const DEFAULT_MUTATION_ALIGNER_URL_TEMPLATE =
-    'https://www.cbioportal.org/getMutationAligner.json?pfamAccession=<%= pfamDomainId %>';
+    'https://www.cbioportal.org/proxy/mutationaligner.org/api/domains/<%= pfamDomainId %>?metadata=true';
```
inodb, line 18: @onursumer actually would be good to use the specific instance's proxy (i.e. genie uses genie.cbioportal.org/proxy). Can use this function: https://github.com/cBioPortal/cbioportal-frontend/blob/b0b2d3852ccafdc3f5fa41ab611bd4de26a2c129/src/shared/api/urls.ts#L191-L203

onursumer: To be able to use mutation aligner outside the cbioportal context, we should still keep the default proxy URL. However, each portal instance should use its own proxy URL. PR has been updated accordingly.

src/shared/api/urls.ts
onursumer: Actually, we decided to keep this to be able to use instance specific proxy URLs.

PR #3293, opened 2020-06-23
Add mutation assessor definition to functional impact tooltip

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/6208

image

Add mutation assessor definition to functional impact tooltip.
Mutation assessor was down so we comment it out: https://github.com/cBioPortal/cbioportal-frontend/pull/978, now add it back


leexgh:
Mutation Assessor is down, we'll remove it until it's back

PR #3298, opened 2020-06-26
Skip all oncokb tests

zhx828:
OncoKB tests are based on the tooltip screenshots but often they are unstable due to
- Tooltip shifting
- Data change resulted in the table layout change

Disable for now until finding a better solution


zhx828:
@alisman I'm gonna merge this pr. Don't think the falling tests are related.

PR #3302, opened 2020-06-30
redesign the UI for pulling generic assay data in OncoPrint

dippindots:
Fix cBioPortal/cbioportal#7656

- Change to multiple select, add `select all filtered options` option
image

- Warning for options over 100 (disable button when showing warning)
image
image

Test Link: [link](https://deploy-preview-3302--cbioportalfrontend.netlify.app/results/oncoprint?genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ccle_broad_2019_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=ccle_broad_2019_cna&cancer_study_list=ccle_broad_2019&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&profileFilter=0&case_set_id=ccle_broad_2019_cnaseq&gene_list=TP53&geneset_list=%20&tab_index=tab_visualize&Action=Submit)


dippindots:
Thanks for the review and reminder, I will add some tests for this new component!

dippindots:
@jjgao Get I get your feedback on this new component and the warning text? Thanks!

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -694,8 +672,32 @@ export default class OncoprintControls extends React.Component<
...
+        ) {
+            candidateOptions = _.union(
+                this.filteredGenericAssayOptions,
+                candidateOptions
+            );
+        }
+        // map to id
+        let candidateIds = candidateOptions.map(o => o.id);
+        // remove select all option from the candidate id list
+        if (candidateIds.includes('select_all_filtered_options')) {
```
alisman, line 694: might as well just use filter function to make this simpler

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -694,8 +672,32 @@ export default class OncoprintControls extends React.Component<
...
+        selectInfo: any
+    ) {
+        // selectedOptions can be null if delete the last selected option
+        let candidateOptions = selectedOptions ? selectedOptions : [];
+        // if choose select all option, add all filtered options
+        if (
+            selectInfo.action === 'select-option' &&
+            selectInfo.option.id === 'select_all_filtered_options'
+        ) {
+            candidateOptions = _.union(
```
alisman, line 686: why do you do this union?  is it because there maybe previously selected options that you want to retain?  if so, that should be explained in comment

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -719,8 +721,20 @@ export default class OncoprintControls extends React.Component<
         }));
     }
 
-    @computed get filteredGenericAssayOptions() {
-        const allOptionsInSelectedProfile =
+    private isOptionIncludingText(text: string, option: ISelectOption) {
+        let result = false;
+        if (
+            !text ||
+            option.label.toLowerCase().includes(text.toLowerCase()) ||
```
alisman, line 728: We should use case insensitive regexp to avoid all this casing stuff.  

(new RegExp(option.label,"i")).test(text)

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -226,8 +218,7 @@ export default class OncoprintControls extends React.Component<
     @observable horzZoomSliderState: number;
     @observable heatmapGenesReady = false;
     @observable private _selectedGenericAssayEntityIds: string[];
```
alisman, line 220: you can make _selectedGenericAssayEntityIds @observable.ref
You do this if every time it gets updated, you replace array entirely (which you seem to do).  
This is more preformance b/c mbox doesn't have to make all elements of the array observable

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -730,41 +736,54 @@ export default class OncoprintControls extends React.Component<
                       .genericAssayEntitiesSelectOptionsGroupByMolecularProfileIdPromise!
                       .result[this.props.state.selectedHeatmapProfileId]
                 : [];
-        if (this.genericAssayEntityFilter) {
-            const regex = new RegExp(this.genericAssayEntityFilter, 'i');
-            return allOptionsInSelectedProfile.filter(
-                option => regex.test(option.label) || regex.test(option.value)
+
+        // add select all option
```
alisman, line 740: you should comment why we only allow select all when filtered

alisman:
@dippindots can we put the count of filtered options beside select all text?  So "Select all filtered options (26)
image

pvannierop:
@dippindots In the example image there is truncation of the names for each generic assay in the "selection buttons". Would it be possible to reveal the full name when hovering the selection tile with the mouse?

dippindots:
I tried the tooltip, but seems like it's not compatible with our DeaultTooltip solution, do you think reveal the whole text is good enough?
image

pvannierop:
@dippindots Yes, great!

PR #3304, opened 2020-07-01
Study page: updates to comparison group manager

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7280

Other Updates
- [x] Allow group creation whether the user is logged in or not
    - [x] For logged in users -> groups would be saved to his account
    - [x] For non-logged in users-> created groups would have page scope
        - A warning should show: "The groups above will not be saved without log-in"
- [x] Always show shared groups in groups panel
    - [x] For logged in users-> shared groups are saved to his account and are highlighted
- [x] Groups panel should be visible by default on page load if there are any shared groups in the URL
    - [x] Highlight the shared groups

### DIFF
--- a/src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
+++ b/src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
@@ -17,7 +17,6 @@ import {
 import styles from '../styles.module.scss';
 import { DefaultTooltip, remoteData } from 'cbioportal-frontend-commons';
 import {
-    addSamplesParameters,
     getGroupParameters,
     getSelectedGroups,
 } from './ComparisonGroupManagerUtils';
@@ -30,9 +29,13 @@ import { sleepUntil } from '../../../shared/lib/TimeUtils';
 import { LoadingPhase } from '../GroupComparisonLoading';
 import _ from 'lodash';
 import { serializeEvent } from 'shared/lib/tracking';
+import AppConfig from 'appConfig';
+import { openSocialAuthWindow } from 'shared/lib/openSocialAuthWindow';
+import classnames from 'classnames';
 
 export interface IComparisonGroupManagerProps {
     store: StudyViewPageStore;
+    shareGroups: (groups: StudyViewComparisonGroup[]) => void;
 }
 
 @observer
@@ -97,21 +100,6 @@ export default class ComparisonGroupManager extends React.Component<
     @action
     private deleteGroup(group: StudyViewComparisonGroup) {
         this.props.store.toggleComparisonGroupMarkedForDeletion(group.uid);
-    }
-
-    @autobind
-    @action
-    private async addSamplesToGroup(group: StudyViewComparisonGroup) {
-        if (this.props.store.selectedSamples.result) {
-            await comparisonClient.updateGroup(
-                group.uid,
-                addSamplesParameters(
-                    group,
-                    this.props.store.selectedSamples.result
-                )
-            );
-            this.props.store.notifyComparisonGroupsChange();
-        }
     }
 
     @autobind
@@ -181,16 +169,17 @@ export default class ComparisonGroupManager extends React.Component<
     }
 
     @autobind
-    private async renameGroup(
-        newName: string,
-        group: StudyViewComparisonGroup
-    ) {
-        const { name, ...rest } = group;
-        await comparisonClient.updateGroup(group.uid, {
-            name: newName,
-            ...rest,
-        });
-        this.props.store.notifyComparisonGroupsChange();
+    private shareSingleGroup(group: StudyViewComparisonGroup) {
+        this.props.shareGroups([group]);
+    }
+
+    @autobind
+    private shareGroups() {
+        const selectedGroups = getSelectedGroups(
+            this.props.store.comparisonGroups.result || [],
+            this.props.store
+        );
+        this.props.shareGroups(selectedGroups);
     }
 
     private readonly groupsSection = MakeMobxView({
@@ -202,9 +191,6 @@ export default class ComparisonGroupManager extends React.Component<
         render: () => {
             if (this.props.store.comparisonGroups.result!.length > 0) {
                 // show this component if there are groups, and if filteredGroups is complete
-                const allGroupNames = this.props.store.comparisonGroups.result!.map(
-                    group => group.name
-                );
                 return (
                     <div className={styles.groupCheckboxes}>
                         {this.filteredGroups.result!.length > 0 ? (
@@ -216,10 +202,9 @@ export default class ComparisonGroupManager extends React.Component<
                                         group.uid
                                     )}
                                     restore={this.restoreGroup}
-                                    rename={this.renameGroup}
                                     delete={this.deleteGroup}
-                                    addSelectedSamples={this.addSamplesToGroup}
-                                    allGroupNames={allGroupNames}
+                                    studyIds={this.props.store.studyIds}
+                                    shareGroup={this.shareSingleGroup}
                                 />
                             ))
                         ) : (
@@ -280,18 +265,47 @@ export default class ComparisonGroupManager extends React.Component<
         }
     }
 
+    private get shareButton() {
+        if (
+            this.props.store.comparisonGroups.isComplete &&
+            this.props.store.comparisonGroups.result.length > 0
+        ) {
+            // only show select button if there are any groups
+            return (
+                <button
+                    className="btn btn-sm btn-default"
+                    disabled={
+                        getSelectedGroups(
+                            this.props.store.comparisonGroups.result,
+                            this.props.store
+                        ).length === 0
+                    }
+                    style={{ marginLeft: 7 }}
+                    onClick={this.shareGroups}
+                >
+                    Share
+                </button>
+            );
+        } else {
+            return null;
+        }
+    }
+
+    @computed get allGroupNames() {
+        return this.props.store.comparisonGroups.isComplete
+            ? this.props.store.comparisonGroups.result!.map(group => group.name)
+            : undefined;
+    }
+
     @computed get submitNewGroupDisabled() {
         const selectedSamples = this.props.store.selectedSamples.isComplete
             ? this.props.store.selectedSamples.result
             : undefined;
-        const allGroupNames = this.props.store.comparisonGroups.isComplete
-            ? this.props.store.comparisonGroups.result!.map(group => group.name)
-            : undefined;
         return (
             !selectedSamples ||
             this.inputGroupName.length === 0 ||
-            !allGroupNames ||
-            allGroupNames.includes(this.inputGroupName)
+            !this.allGroupNames ||
+            this.allGroupNames.includes(this.inputGroupName)
         );
     }
 
@@ -308,8 +322,8 @@ export default class ComparisonGroupManager extends React.Component<
                 this.props.store.studyIds
             )
         );
-        this.props.store.setComparisonGroupSelected(id); // created groups start selected
         this.props.store.notifyComparisonGroupsChange();
+        this.props.store.setComparisonGroupSelected(id); // created groups start selected
         this.cancelAddGroup();
     }
 
@@ -401,6 +415,7 @@ export default class ComparisonGroupManager extends React.Component<
                     >
                         {this.compareButton}
                         {this.viewButton}
+                        {this.shareButton}
                     </div>
                 );
             } else {
@@ -411,13 +426,9 @@ export default class ComparisonGroupManager extends React.Component<
 
     private get addGroupPanel() {
         let contents: any;
-        const inputWidth = 235;
         const createOrAddButtonWidth = 58;
         const selectedSamples = this.props.store.selectedSamples.isComplete
             ? this.props.store.selectedSamples.result
-            : undefined;
-        const allGroupNames = this.props.store.comparisonGroups.isComplete
-            ? this.props.store.comparisonGroups.result!.map(group => group.name)
             : undefined;
         if (this.addGroupPanelOpen) {
             contents = [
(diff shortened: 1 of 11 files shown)


src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -280,18 +252,95 @@ export default class ComparisonGroupManager extends React.Component<
         }
     }
 
+    @observable private copied = false;
```
adamabeshouse, line 255: I'm a little confused, can you explain what's happening with `copied`?

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -534,6 +581,57 @@ export default class ComparisonGroupManager extends React.Component<
         this.props.store.deleteMarkedComparisonGroups();
     }
 
+    @computed get anyThereAnySharedGroups() {
```
adamabeshouse, line 584: usually I've been phrasing this like `existSharedGroups` but I'm not sure that there's an established idiom for this. maybe let's go with `existSharedGroups` to start a trend?

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -534,6 +581,57 @@ export default class ComparisonGroupManager extends React.Component<
...
+                                openSocialAuthWindow(this.props.store.appStore)
+                            }
+                        >
+                            Login
+                        </button>
+                        <br />
+                    </>
+                );
+            } else if (this.anyThereAnySharedGroups) {
+                message = <>{message} They would be saved to your profile.</>;
```
adamabeshouse, line 629: I'm not sure I understand the meaning of this message. Can you explain it more and maybe we can figure out another way to express it?

src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
```diff
@@ -169,73 +112,11 @@ export default class GroupCheckbox extends React.Component<
                         )}
                         onClick={this.onCheckboxClick}
                     />
-                    {this.editingName ? this.editingNameUI : this.label}
+                    {this.label}
                 </div>
             );
         }
 
-        let editingRelatedIcons;
```
adamabeshouse, line 177: we can no longer rename groups?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -501,26 +537,129 @@ export class StudyViewPageStore {
...
+                            groupsToSave.push(sharedGroup.uid);
+                        }
+                    }
+                });
+            }
+
+            if (groupsToSave.length > 0) {
+                const addPromises: Promise<void>[] = [];
+                groupsToSave.forEach(groupId =>
+                    addPromises.push(comparisonClient.addGroupToUser(groupId))
```
adamabeshouse, line 618: I don't think `invoke` should be executing changes like this. unless it's necessary for the logic of `invoke` it should be in `onResult`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1035,6 +1176,11 @@ export class StudyViewPageStore {
                 this.studyIds = studyIds;
             }
         }
+        if (query.sharedGroups) {
+            this.sharedGroupSet = stringListToSet(
```
adamabeshouse, line 1180: can you add a comment explaining that this can be done in constructor because it comes from the URL

src/shared/api/ComparisonGroupClient.ts
```diff
@@ -27,19 +26,19 @@ export type Group = {
...
-    // edge case: user deletes a group, then opens the panel again,
-    //          and the getGroups request responds before the deleteGroup
+    // edge case: user deletes/add a group, then opens the panel again,
+    //          and the getGroups request responds before the deleteGroup/addGroup
     //          request completes, thus showing a group that should be
     //          (and soon will be) deleted. We fix this by waiting
-    //          until deletions are done before allowing getGroups requests.
-    private _pendingDeletions: PromisePlus<any>[] = [];
+    //          until deletions/additions are done before allowing getGroups requests.
+    private _pendingDecisions: PromisePlus<any>[] = [];
```
adamabeshouse, line 34: good idea but I think a better word would be `_pendingChanges`

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -280,18 +252,95 @@ export default class ComparisonGroupManager extends React.Component<
...
+    }
+
+    @autobind
+    @action
+    private onTooltipVisibleChange(visible: boolean) {
+        this.copied = !visible;
+    }
+
+    @autobind
+    private copyLinkRef(el: HTMLAnchorElement | null) {
```
alisman, line 270: i think these should have names more specific to group sharing feature?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -350,6 +354,41 @@ export class StudyViewPageStore {
             )
         );
 
+        this.reactionDisposers.push(
```
alisman, line 357: i think i understand why you made this a reaction (we only want to force the tooltip on when it changes, not after user closes it)?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -350,6 +354,41 @@ export class StudyViewPageStore {
...
+            reaction(
+                () => this.sharedGroupSet,
+                groups => {
+                    if (Object.keys(groups).length > 0) {
+                        this.showGroupsTooltip = true;
```
alisman, line 362: by groupsTooltip do we mean the whole groups UI dropdown?   If so, I think we should call it comparisonGroupUI or something.   tooltip suggests something less significnat.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -350,6 +354,41 @@ export class StudyViewPageStore {
...
+                    }
+                }
+            )
+        );
+
+        this.reactionDisposers.push(
```
alisman, line 368: again, need explaination.  when user logs in, if there are shared groups, save them, right?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -501,26 +537,129 @@ export class StudyViewPageStore {
...
+        invoke: async () => {
+            let groups: StudyViewComparisonGroup[] = _.cloneDeep(
+                this.userSavedGroups.result
+            );
+            let groupIdSet: { [s: string]: boolean } = {};
+
+            const groupsToSave: string[] = [];
+            if (this.sharedGroups.result.length > 0) {
+                groupIdSet = stringListToSet(groups.map(group => group.uid));
+                this.sharedGroups.result.forEach(sharedGroup => {
```
alisman, line 604: clear if you did groupsToSave = result.reduce()

src/pages/studyView/StudyViewPageStore.ts
alisman: agreed.    i think onresult is the way to go.   could we reuse logic from the login reaction?

alisman: yeah, if that reaction just referenced comparisonGroups, would that not cover all the cases?

src/shared/api/ComparisonGroupClient.ts
```diff
@@ -27,19 +26,19 @@ export type Group = {
 export type GroupData = Omit<VirtualStudyData, 'studyViewFilter'>;
 
 export default class ComparisonGroupClient {
-    // edge case: user deletes a group, then opens the panel again,
-    //          and the getGroups request responds before the deleteGroup
+    // edge case: user deletes/add a group, then opens the panel again,
```
alisman, line 29: doesn't feel quite right to deal with this at client level.  our clients thus far don't have any such state (as far as I know).   i would deal with this in the store mobxpromises.  @adamabeshouse what do you think?

src/pages/groupComparison/comparisonGroupManager/GroupCheckbox.tsx
kalletlak: yes, this is because we are allowing sharing of groups

src/pages/studyView/StudyViewPageStore.ts
kalletlak: yes

src/pages/studyView/StudyViewPageStore.ts
kalletlak: yes

src/shared/api/ComparisonGroupClient.ts
adamabeshouse: @alisman I disagree, I think it's an issue the client should hide, too nitty gritty for every application to handle and rehandle

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
kalletlak: Updated the logic. May be its clear now

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
kalletlak: Updated logic and added comments

src/pages/studyView/StudyViewPageStore.ts
kalletlak: I tried referencing `comparisonGroups` in reaction but it wasn't working as expected. Anyways I ended removing the reaction as it is no more required. And I ended up using `onResult` which works fine for all cases

src/pages/studyView/StudyViewPageStore.ts
kalletlak: This reaction is no longer required

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
adamabeshouse: sorry, I still don't get it. what does `copied` represent? maybe a different variable name would make it clearer, e.g. what is copied?

kalletlak: forgot to clean it.  it's updated now

jjgao:
@kalletlak thanks, this looks great!

Some small comments:

Should `sharedGroups` in the main URL or add to # portion? @alisman 

In this text, should it be "They have been saved to your profile"?
image

kalletlak:
@jjgao updated text message

src/shared/api/ComparisonGroupClient.ts
alisman: @kalletlak i'm gonna go with my gut here.  when i make a request to a client, i am not expecting that request to be dependent on status of other requests.   we do not want to start cluttering our clients with this kind of business logic.   this belongs in the store.

jjgao:
@cBioPortal/product please help test. You may want to use the public portal or any private portal us run `localStorage.netlify = "deploy-preview-3304--cbioportalfrontend"`.

jjgao:
@kalletlak it looks great!

Just one last comment: is it possible to have the sharedGroups after #? Or is there a reason for the current design?

https://www.cbioportal.org/study?id=msk_impact_2017#sharedGroups=5d092670e4b0ab4137876374,5c99454ee4b0ab4137873dc5

kalletlak:
> @kalletlak it looks great!
> 
> Just one last comment: is it possible to have the sharedGroups after # in the url? Or is there a reason for the current design?

@alisman and I had a discussion and decided to go with main url for now as the shareable groups are limited and url length with groups included would be in limit, unless user creates thousands of groups and share them.

jjgao:
@kalletlak @alisman I think it's better to keep it consistent (even though shared_groups is short) since we have the filter_json after #, unless there is a particular technical reason.

src/pages/groupComparison/comparisonGroupManager/ComparisonGroupManager.tsx
```diff
@@ -542,6 +612,17 @@ export default class ComparisonGroupManager extends React.Component<
             >
                 {this.header}
                 {this.groupsSection.component}
+                {this.notificationMessages.length > 0 && (
+                    <>
+                        <br />
+                        <i
+                            className="text-warning"
+                            style={{ fontSize: '12px' }}
```
alisman, line 620: what's reason for inline fontsize?

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -220,6 +222,30 @@ export default class StudyViewPage extends React.Component<
         this.showBookmarkModal = !this.showBookmarkModal;
     }
 
+    @observable showShareBookmarkModal = false;
```
alisman, line 225: maybe we just call it "shareLinkModal"?    i think the idea of sharing encompasses bookmark.

src/shared/api/ComparisonGroupClient.ts
alisman: @kalletlak are we going to do this (move it to store)?

alisman:
I think we need a line of instruction text here.  Note, I've also removed the redundant label above input and pluralized the title.  If it's easy enough, make the title dynamic so that it reads either "Share group" for single group or "Shared 4 groups"

image

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: I'm confused though - why would the shared groups change after initial page load? i.e. why a reaction, why not just in constructor?

kalletlak: true. I'll remove it

src/shared/api/ComparisonGroupClient.ts
kalletlak: moved.

kalletlak:
@alisman done with updates as suggested.
@jjgao moved sharedGroups parameter to hash string. ex: http://localhost:8080/study?id=acc_tcga#sharedGroups=5f11bcaee84a8951f4ac6980

inodb:
@alisman going to merge this, but please follow up if are were more comments that needed to be addressed

Test failures unrelated

PR #3306, opened 2020-07-02
Refactored patient view clinical attributes summary using JavaScript …

jtquach1:
…instead of original CSS ordering

Fix cBioPortal/cbioportal#7631

- Replicated ordering logic from CSS in JS by sorting the clinical attributes in ascending order of their mapped integer values
- Removed an unreferenced line of code, `let spans = [];` 
- Removed the original ordering logic in CSS.

## Any screenshots or GIFs?
Before: 
screenshot_original_patient_view
screenshot_original_patient_view_2

After: 
screenshot_refactor_patient_view
screenshot_refactor_patient_view_2


src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,54 @@ function getSpanElementsFromCleanData(
...
+
+    return Object.keys(clinicalAttributesCleanDerived)
+        .sort(compare)
+        .map(key => {
+            let value = clinicalAttributesCleanDerived[key];
+            return (
+                <span
+                    is
+                    class="clinical-attribute"
+                    attr-id={key}
```
Luke-Sikina, line 250: Let's check on Monday, but I don't think we need this field any more.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,56 @@ function getSpanElementsFromCleanData(
...
-                attr-value={value}
-                study={cancerStudyId}
-            >
-                {value}
-            </span>
-        );
-    });
+    // Order patient clinical attributes
+    const getOrder = key => {
+        switch (key) {
```
alisman, line 212: instead of a switch statement, i would make a map: {  'SEX':1, ....   }, then check in this function for undefined and return the default if it's undefined.

alisman:
Hi @jtquach1, this is a good start.  If you look at this patient:
https://www.cbioportal.org/patient?studyId=ucec_tcga_pub&caseId=TCGA-BK-A0CC
you'll see there's some coloring and even some characters () and , which are actually being placed into the content via by css (using :before and :after pseudo selectors).  We also want to accomplish this in js instead of in css.   So, we need to figure out the logic that is embedded in that css and reproduce it.   We can discuss this further in chat if necessary.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
jtquach1: Ok, I'm currently refactoring the code to use an object instead of a switch statement. 
(I'm currently having trouble since I can't subscript the object with the function argument, but I will keep you updated!)

jtquach1: I've just pushed another commit: "Edited patient view logic based on Aaron's suggestion."

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,89 @@ function getSpanElementsFromCleanData(
...
+        return getOrder(a) - getOrder(b);
+    };
+
+    // Styles before and after value
+    function stringBuilder(value, key) {
+        var builder = [];
+
+        const before = {
+            /* attributes with opening parenthesis */
+            OS_MONTHS: '\xa0(',
```
alisman, line 240: @jtquach1 @inodb, i wonder if we can make this more general.    @inodb could we have a list of primary attributes and secondary attributes?  Secondary would always be wrapped in parens.   lastly, we would need a map of "formatters", e.g.   { OS_MONTHS:  `${s} months`   }.   In fact, the formatters could be values in both primary and secondary maps.

alisman: another thought, @jtquach1 i think we should also make color configurable via data.  so, it should be struck from css too and represented in data structure.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -282,9 +325,8 @@ function getSpanElementsFromCleanData(
                 <span
                     is
                     class="clinical-attribute"
-                    attr-id={key}
```
jtquach1: Removing attr-id and study along with  lines 21-24 of the CSS makes the patient view much more verbose

jtquach1: Restoring the attr-id, study HTML attributes and CSS lines 21-24 makes the patient view closer to the reference. However the "Copy #" is still missing from under the "33 Mutations" header.

jtquach1: I refreshed https://www.cbioportal.org/patient?studyId=ucec_tcga_pub&caseId=TCGA-BK-A0CC and it looks like the "Copy #" column is gone from there too, so please disregard my comments about it above!

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,132 @@ function getSpanElementsFromCleanData(
...
-                attr-value={value}
-                study={cancerStudyId}
-            >
-                {value}
-            </span>
-        );
-    });
+    // Order patient clinical attributes
+    const getOrder = key => {
+        const map = {
```
Luke-Sikina, line 212: I'm not a huge fan of having these big `const`s just hanging around in these functions. I think it makes things a little less readable. Maybe they'd be better off at the top of this file, with better names. Don't change it yet, though, I'm curious what @alisman thinks.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,132 @@ function getSpanElementsFromCleanData(
...
+            DERIVED_SAMPLE_LOCATION: ')',
+            SAMPLE_DISPLAY_NAME: ')',
+
+            /* text after an attribute */
+            OS_MONTHS: ' months)',
+            DFS_MONTHS: ' months)',
+            AGE: ' years old',
+        };
+
+        /* Show comma before clinical attributes */
```
Luke-Sikina, line 267:
```suggestion
        // Show comma before clinical attributes
```
You have this in several places. Using block comment syntax for 1 liners probably isn't great. I'm just going to leave this 1 comment, but fix all the 1 line block comments.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,132 @@ function getSpanElementsFromCleanData(
...
+        // Prefix is never used, but its assignment allows the side effect of push to happen
+        const prefix =
+            before[key] == '\xa0(' ? builder.push('') : builder.push(', ');
```
Luke-Sikina, line 270:
```suggestion
            before[key] === '\xa0(' ? builder.push('') : builder.push(', ');
```

Luke-Sikina, line 270: What? I'm confused. What bug occurs if you do something like this:

```suggestion
        const prefix = before[key] === '\xa0(' ? '' :', ';
        builder.push(prefix);
```

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +207,132 @@ function getSpanElementsFromCleanData(
...
+                Primary: colors.sample_color_primary,
+                Progressed: colors.sample_color_recurrence,
+                Recurrence: colors.sample_color_recurrence,
+                Metastasis: colors.sample_color_metastasis,
+            },
+        };
+
+        let color = 'black';
+        if (keys[key] != undefined) {
+            if (keys[key]['_'] != undefined) {
```
Luke-Sikina, line 309: Magic constants like `'_'` are a bit worrisome. What does `'_'` represent? The code might be more readable if you create a const near the top of the file `const NAME_THAT_EXPLAINS_VAR = '_';` and then use that here.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
jtquach1: That looks nicer, thanks!

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
Luke-Sikina: Thinking about this a little more, my preferred solution would be to pull these consts into their own file. clinicalAttributesStyleConsts.js or something.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
jtquach1: what is the difference between using == and === here?

Luke-Sikina: `===` is equality, `==` is identity. In this case, they are functionally the same, but in general, `==` can lead to unexpected true values because it does some type coercion. In general, always use `===` in Javascript / Typescript

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
jtquach1: I moved the const functions into their own file and documented them with comments just now

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,136 @@
...
+        },
+        DERIVED_NORMALIZED_CASE_TYPE: {
+            Primary: colors.sample_color_primary,
+            Progressed: colors.sample_color_recurrence,
+            Recurrence: colors.sample_color_recurrence,
+            Metastasis: colors.sample_color_metastasis,
+        },
+    };
+
+    if (keys[key] != undefined) {
```
alisman, line 126: you should always use !== instead of !=

alisman: also, you could use ```key in keys```

jtquach1: does !== also check for value and type?

alisman: btw, we use !== for same reason as Luke cites above.  != does type coercion which we want to avoid.    in this case, it's not important, but just be in the habit of always using !== and ===.   safe to say, never use != and ==

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,183 @@
...
+    sample_color_primary: 'black',
+    sample_color_recurrence: 'orange',
+    sample_color_metastasis: 'red',
+    sample_color_cfdna: 'blue',
+    sample_color_xenograft: 'pink',
+};
+
+// Attributes with special colors
+export const keys = {
+    OS_STATUS: {
```
alisman, line 166: how come these keys are not included in the above config object?

jtquach1: good question: I did not think of it initially but I shall refactor that!

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,201 @@
...
+
+/**
+ * Object mapping of keys to objects indicating
+ * whether an attribute is primary or secondary,
+ * the formatter to display the inputted value t,
+ * and font coloring based on the key and value.
+ */
+export const config = {
+    OS_MONTHS: {
+        primary: false,
```
alisman, line 101: for secondary attributes, the parens and space should be outside of the formatter and in once place.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,201 @@
...
+    ERG_FUSION_ACGH: {
+        primary: true,
+        formatter: t => {
+            return `, ERG-fusion aCGH: ${t}`;
+        },
+    },
+    TMPRSS2_ERG_FUSION_STATUS: {
+        primary: true,
+        formatter: t => {
+            return `, TMPRSS2-ERG Fusion: ${t}`;
```
alisman, line 151: again the comma doesn't need to be embedded in the data (and duplicated).  it should go in the getSpanElementsFromCleanData function down below

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,201 @@
...
+
+    return existsKey && value in config[key].values
+        ? config[key].values[value]
+        : existsKey && NO_VALUE in config[key].values
+        ? config[key].values[NO_VALUE]
+        : color;
+};
+
+// Colors taken from clinicalAttributes.scss
+export const colors = {
```
alisman, line 85: this map isn't necessary.  just put this in the config structure and use strings (e.g. 'black')

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,201 @@
...
+ * Functions for ordering and styling clinical attributes.
+ */
+/**
+ * Order patient clinical attributes by an object mapping.
+ * If the key does not exist in the map, then it shall be ordered last.
+ * @param {string} key
+ */
+export const getOrder = key => {
+    // Keys with lower numbers appear earlier
+    const map = {
```
alisman, line 11: see if you can get this in config also

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
jtquach1: Thanks! Implemented in latest commit!

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,234 @@
...
+            values: {
+                DECEASED: '#f00',
+                DEAD: '#f00',
+                LIVING: 'rgb(0, 128, 0)',
+                ALIVE: 'rgb(0, 128, 0)',
+            },
+        },
+        SAMPLE_DISPLAY_NAME: {
+            primary: false,
+            formatter: t => {
```
alisman, line 119: might as well just leave formatter out when it just prints itself.  i believe you already handle case of no formatter in logic

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,21 +208,28 @@ function getSpanElementsFromCleanData(
...
+                    attr-value={value}
+                    style={{
+                        color: styleConsts.colorGetter(
+                            value,
+                            key,
+                            styleConsts.config
+                        ),
+                    }}
+                >
+                    {styleConsts.stringBuilder(value, key, styleConsts.config)}
```
alisman, line 229: you can handle the comma out here where it will be possible to tell if it's last element.   one more potential issue, you don't want color to be applied to comma. maybe you can put comma in its own span and apply the black color.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
...
+/**
+ * Functions for ordering and styling clinical attributes.
+ */
+/**
+ * Order patient clinical attributes by the object mapping config.
+ * If the key does not exist in config, then it shall be ordered last.
+ * @param {string} key
+ * @param {object} config
+ */
+export const getOrder = (key, config) => {
```
Luke-Sikina, line 172:
```suggestion
export function getOrder(key, config){
```
Let's declare this function like we do everywhere else. Makes it a bit faster to differentiate between a function being exported and an object being exported.

Luke-Sikina: Same goes for all other top level func decls

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
+/**
```
Luke-Sikina, line 1: This is a lot of Javascript to add to a codebase that is almost entirely Typescript. I take it we're doing JS here because it's being used by an existing JS file. Why not convert that file to Typescript and then make this typescript too? All this `@param {object}` documentation is killing me.

Don't change this unless Aaron agrees, because it's a fair bit of work.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
...
+        order: 6,
+    },
+    GENDER: {
+        order: 1,
+    },
+    HISTOLOGY: {
+        order: 6,
+    },
+    KNOWN_MOLECULAR_CLASSIFIER: {
+        order: 5,
```
Luke-Sikina, line 104: Why is `primary` an optional field? Is `primary: undefined` different than `primary: false`?

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
...
+        primary: false,
+        order: 2,
+    },
+    DFS_STATUS: {
+        primary: true,
+        values: {
+            _: '#f00',
+            DiseaseFree: 'rgb(0, 128, 0)',
+            Yes: 'rgb(0, 128, 0)',
+        },
```
Luke-Sikina, line 59: This is really minor and very debatable, but if you are always iterating through all the attributes in `values`, I'd rather it be an array of key value pairs. I think an array of key value pairs shows your intent a bit better.
```suggestion
        values: [
           {key: '_', value: '#f00'},
           {key: 'DiseaseFree', value: 'rgb(0, 128, 0)'}
        },
```

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
...
+ * - leading comma on attribute after sample id
+ * - trailing comma on last displayed attribute
+ * on Patient row and Samples row. *
+ * @param {string} value
+ * @param {string} key
+ * @param {object} config
+ * @param {string[]} keys
+ */
+export const stringBuilder = (value, key, config, keys) => {
+    var builder = [];
```
Luke-Sikina, line 202: It'd be cool to mess around with monads here, but let's not.

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
```diff
@@ -0,0 +1,285 @@
...
+/**
+ *  Gets a color for the given key, value, and config. 
+    If there a specified color for the key and value, return it.
+    If there is no such value, then get the color only specified for the key. 
+    Otherwise, return a default color. 
+ * @param {string} value 
+ * @param {string} key 
+ * @param {object} config 
+ */
+export const colorGetter = (value, key, config) => {
```
Luke-Sikina, line 240: Not a huge fan of nouns for function names. Maybe `calculateColor`?

src/pages/patientView/clinicalInformation/lib/clinicalAttributesUtil.js
```diff
@@ -207,18 +208,28 @@ function getSpanElementsFromCleanData(
     clinicalAttributesCleanDerived,
     cancerStudyId
 ) {
-    let spans = [];
-    return Object.keys(clinicalAttributesCleanDerived).map(key => {
+    const config = styleConsts.config;
```
Luke-Sikina, line 211: This looks good - nice job

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
jtquach1: I forgot to specify `primary` which is what led to the parentheses wrapping around the GENDER and SEX attributes, but this has been corrected for the latest commit

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
jtquach1: Ah, I figured to use the @param documentation after typing /** */ in VS Code, then it showed up behind the function I was defining

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
jtquach1: However wherever primary is undefined, the logic in stringBuilder should return false

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
jtquach1: how do you mean?

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
alisman: i think the map is fine @Luke-Sikina, it's a lookup table.  we do that a lot

src/pages/patientView/clinicalInformation/lib/clinicalAttributesStyleConsts.js
alisman: yeah, Luke, what's monad!?    i think we should get this merged soon so lets hold off.  but i am interested in  what that is!

Luke-Sikina: Oh, this was a joke. Please do not solve this using monads.

But to quote Saunders Mac Lane:
> All told, a monad in X is just a monoid in the category of endofunctors of X, with product × replaced by composition of endofunctors and unit set by the identity endofunctor.

jtquach1: aha, I know this meme (very infamously and ironically spouted by a classmate in our Structure of Higher Level languages class)

jtquach1:
> Hi @jtquach1 we still need to get rid of attr-id and any styling that is dependend on it.

Done! Removed references to attr-id, .clinical-attributes, and attr-value since their logic is taken care of in clinicalAttributesStyleConsts.ts

PR #3311, opened 2020-07-09
Migrate fusion to structrual variant in oncoprint

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/6965
Fix https://github.com/cBioPortal/cbioportal/issues/1874

To test
https://cbioportal-fix-6965-7m35ffcob2.herokuapp.com and set `localStorage.netlify = "deploy-preview-3311--cbioportalfrontend"`

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -30,6 +30,8 @@ import {
     SampleIdentifier,
     SampleList,
     SampleMolecularIdentifier,
+    StructuralVariant,
+    StructuralVariantFilter,
 } from 'cbioportal-ts-api-client';
 import client from 'shared/api/cbioportalClientInstance';
 import { action, computed, observable, ObservableMap, reaction } from 'mobx';
@@ -38,7 +40,7 @@ import {
     IOncoKbData,
     remoteData,
     stringListToSet,
-    getBrowserWindow,
+    generateQueryStructuralVariantId,
 } from 'cbioportal-frontend-commons';
 import { getProteinPositionFromProteinChange } from 'cbioportal-utils';
 import {
@@ -76,6 +78,7 @@ import {
     getGenomeNexusUrl,
     fetchSurvivalDataExists,
     getSurvivalClinicalAttributesPrefix,
+    fetchStructuralVariantOncoKbData,
 } from 'shared/lib/StoreUtils';
 import { IHotspotIndex, indexHotspotsData } from 'react-mutation-mapper';
 import { fetchHotspotsData } from 'shared/lib/CancerHotspotsUtils';
@@ -132,6 +135,9 @@ import {
     isRNASeqProfile,
     OncoprintAnalysisCaseType,
     parseGenericAssayGroups,
+    filterAndAnnotateStructuralVariants,
+    compileStructuralVariants,
+    FilteredAndAnnotatedStructuralVariantsReport,
 } from './ResultsViewPageStoreUtils';
 import MobxPromiseCache from '../../shared/lib/MobxPromiseCache';
 import { isSampleProfiledInMultiple } from '../../shared/lib/isSampleProfiled';
@@ -258,7 +264,10 @@ export type SamplesSpecificationElement =
     | { studyId: string; sampleId: string; sampleListId: undefined }
     | { studyId: string; sampleId: undefined; sampleListId: string };
 
-export interface ExtendedAlteration extends Mutation, NumericGeneMolecularData {
+export interface ExtendedAlteration
+    extends Mutation,
+        NumericGeneMolecularData,
+        StructuralVariant {
     hugoGeneSymbol: string;
     molecularProfileAlterationType: MolecularProfile['molecularAlterationType'];
     // TODO: what is difference molecularProfileAlterationType and
@@ -273,6 +282,14 @@ export interface AnnotatedMutation extends Mutation {
     oncoKbOncogenic: string;
     isHotspot: boolean;
     simplifiedMutationType: SimplifiedMutationType;
+}
+
+export interface AnnotatedStructuralVariant extends StructuralVariant {
+    putativeDriver: boolean;
+    oncoKbOncogenic: string;
+    isHotspot: boolean;
+    entrezGeneId: number;
+    hugoGeneSymbol: string;
 }
 
 export interface AnnotatedNumericGeneMolecularData
@@ -1400,6 +1417,7 @@ export class ResultsViewPageStore {
         await: () => [
             this.filteredAndAnnotatedMutations,
             this.filteredAndAnnotatedMolecularData,
+            this.filteredAndAnnotatedStructuralVariants,
             this.selectedMolecularProfiles,
             this.entrezGeneIdToGene,
         ],
@@ -1408,13 +1426,11 @@ export class ResultsViewPageStore {
                 this.selectedMolecularProfiles.result!
             );
             const entrezGeneIdToGene = this.entrezGeneIdToGene.result!;
-            let result: (
-                | AnnotatedMutation
-                | AnnotatedNumericGeneMolecularData)[] = [];
-            result = result.concat(this.filteredAndAnnotatedMutations.result!);
-            result = result.concat(
-                this.filteredAndAnnotatedMolecularData.result!
-            );
+            let result = [
+                ...this.filteredAndAnnotatedMutations.result!,
+                ...this.filteredAndAnnotatedMolecularData.result!,
+                ...this.filteredAndAnnotatedStructuralVariants.result!,
+            ];
             return Promise.resolve(
                 result.map(d => {
                     const extendedD: ExtendedAlteration = annotateAlterationTypes(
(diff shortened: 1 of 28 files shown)


src/shared/lib/oql/oqlfilter.ts
```diff
@@ -820,14 +825,15 @@ function filterData<T extends Datum>(
                 gene: query_line.gene,
                 parsed_oql_line: query_line,
                 oql_line: unparseOQLQueryLine(query_line),
-                data: data.filter(datum =>
-                    isDatumWantedByOQLLine(
+                data: data.filter(datum => {
+                    const temp = isDatumWantedByOQLLine(
```
adamabeshouse, line 829: unnecessary variable

src/shared/components/oncoprint/geneticrules.ts
```diff
@@ -308,13 +308,13 @@ const non_mutation_rule_params: GeneticAlterationRuleParams = {
...
-            // tall inset purple rectangle for fusion
+        // structural variant
+        disp_structuralVariant: {
+            // tall inset purple rectangle for structural variant
             true: {
-                shapes: [shapeBank[ShapeId.fusionRectangle]],
-                legend_label: 'Fusion',
-                legend_order: FUSION_LEGEND_ORDER,
+                shapes: [shapeBank[ShapeId.structuralVariantRectangle]],
+                legend_label: 'Structual Variant',
```
adamabeshouse, line 316: typo

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2713,6 +2737,52 @@ export class ResultsViewPageStore {
...
+                        memo.push({
+                            molecularProfileId:
+                                studyIdToProfileMap[sample.studyId]
+                                    .molecularProfileId,
+                            sampleId: sample.sampleId,
+                        });
+                    }
+                    return memo;
+                },
+                [] as any[]
```
alisman, line 2769: i think it might be possible to type this as StructuralVariantFilter.sampleMolecularIdentifiers?  worth a try

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2713,6 +2737,52 @@ export class ResultsViewPageStore {
...
+
+            const data = {
+                entrezGeneIds: _.map(
+                    this.genes.result,
+                    (gene: Gene) => gene.entrezGeneId
+                ),
+                sampleMolecularIdentifiers: filters,
+            } as StructuralVariantFilter;
+
+            return await client.fetchStructuralVariantsUsingPOST({
```
alisman, line 2780: just remove async/await and return client.fetch...    since we aren't actually awaiting anything

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3943,6 +4087,67 @@ export class ResultsViewPageStore {
         },
     });
 
+    readonly getStructuralVariantPutativeDriverInfo = remoteData({
+        await: () => {
+            const toAwait = [];
```
alisman, line 4092: @adamabeshouse @kalletlak i'm just thinkng about these dynamic awaits.  i know we've long used them.  but do they actually work?  when does await get called (only at setup time or is it reactive?). just kind of curious.  it seems possible to me that we think it's working but it actually doesn't do what we think.  it would have to re-evaluate what it needs to await every time something changes, but if one of the conditions evaluates false initially, it won't actually be registered with its change event.

adamabeshouse: @alisman in my opinion the fact that we're not seeing bugs is proof that it does work, but we could do a more purposive test if necessary. 

In theory it does work - `await` is not `@computed` but it is referenced by `@computed status`, which means that any changed observables in `await` will trigger recomputation of `status` and `await`. `driverAnnotationSettings` is an observable object so even if the conditions evaluate false they will trigger a recomputation when they change. The only gotcha is to make sure in `invoke` that you're not expecting a promise to be ready that wasn't actually awaited.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3943,6 +4087,67 @@ export class ResultsViewPageStore {
...
+            if (this.driverAnnotationSettings.oncoKb) {
+                toAwait.push(
+                    this.getOncoKbStructuralVariantAnnotationForOncoprint
+                );
+            }
+            if (this.driverAnnotationSettings.hotspots) {
+                toAwait.push(this.isHotspotForOncoprint);
```
adamabeshouse, line 4099: why are you awaiting these (hotspots, cbioportalCount, cosmicCount), they dont' seem to be used?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4133,6 +4371,43 @@ export class ResultsViewPageStore {
         },
     });
 
+    readonly getOncoKbStructuralVariantAnnotationForOncoprint = remoteData<
```
alisman, line 4374: seems like mostly we don't name these mobxpromises "get..."   which implies a function?

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -158,6 +160,35 @@ export function annotateMutationPutativeDriver(
...
+): AnnotatedStructuralVariant {
+    const putativeDriver = !!(
+        putativeDriverInfo.oncoKb ||
+        putativeDriverInfo.hotspots ||
+        putativeDriverInfo.cbioportalCount ||
+        putativeDriverInfo.cosmicCount ||
+        putativeDriverInfo.customDriverBinary ||
+        putativeDriverInfo.customDriverTier
+    );
+    return Object.assign(
```
alisman, line 182: curious why you need Object.assign?  why not just add structuralVariant in the same way as putativeDriver?

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -232,6 +316,26 @@ export function compileMutations<
     return mutations;
 }
 
+export function compileStructuralVariants<
```
alisman, line 319: mergeStructuralVariants?  concatStructuralVariants?

src/shared/components/query/QueryStore.ts
```diff
@@ -1885,6 +1895,11 @@ export class QueryStore {
                 !this.defaultMutationProfile
             )
                 return 'Mutation data query specified in OQL, but no mutation profile is available for the selected study.';
+            if (
+                this.alterationTypesInOQL.haveStructuralVariantInQuery &&
+                !this.defaultStructuralVariantProfile
+            )
+                return 'Structural Variant data query specified in OQL, but no structural variant profile is available for the selected study.';
```
alisman, line 1902: second word (variant) probably should be lower case.  Also, could it be more than one study?

alisman:
@kalletlak can you comment on backward compatibility issues?  are there fusion profiles which will no longer be valid?  anything else you foresee?   it might not be important but i just want to get your thoughts.

src/shared/lib/oql/AccessorsForOqlFilter.ts
```diff
@@ -123,7 +127,13 @@ export default class AccessorsForOqlFilter
...
 
     public gene(d: Datum) {
-        return d.gene.hugoGeneSymbol;
+        if (
+            this.molecularAlterationType(d.molecularProfileId) ===
+            AlterationTypeConstants.STRUCTURAL_VARIANT
+        ) {
+            return (d as StructuralVariant).site1HugoSymbol;
+        }
+        return (d as any).gene.hugoGeneSymbol;
```
alisman, line 136: why does it now have to be d as any?  seems like typing should be same for this case?

src/shared/lib/oql/annotateAlterationTypes.ts
```diff
@@ -17,18 +19,16 @@ export function annotateAlterationTypes(
...
-                datum.alterationSubType = accessors.mut_type(
-                    datum as AnnotatedMutation
-                ) as any;
-            }
+            datum.alterationType = AlterationTypeConstants.MUTATION_EXTENDED;
+            datum.alterationSubType = accessors.mut_type(
+                datum as AnnotatedMutation
+            ) as any;
+            break;
+        case AlterationTypeConstants.STRUCTURAL_VARIANT: {
```
alisman, line 27: not sure i understand these curly brackets in context of switch syntax.  is this syntax i just don't know about?

src/pages/resultsView/ResultsViewPageStore.ts
kalletlak: I'll remove them

src/pages/resultsView/ResultsViewPageStoreUtils.ts
kalletlak: We could. I was copy pasting logic from `annotateMutationPutativeDriver` (function above this)

src/pages/resultsView/ResultsViewPageStoreUtils.ts
kalletlak: I was using the same format what was used for mutations `compileMutations`(line 297)

src/shared/components/query/QueryStore.ts
kalletlak: it could be more than one study

kalletlak: these messages are shown for single study

src/shared/lib/oql/AccessorsForOqlFilter.ts
kalletlak: `gene` object doesn't exist in `StructuralVariant` and `AnnotatedStructuralVariant` types.

src/shared/lib/oql/annotateAlterationTypes.ts
kalletlak: curly brackets are not required in this case. but sometimes there might be a same variables which needs to be declared in multiple cases. It would complain that variable is already declared. In that case adding curly brackets around case would declare those variables with block scope. example at https://github.com/cBioPortal/cbioportal-frontend/pull/3311/files#diff-8f5a13488274fc21f61a85ce616cf599R664, if curly brackets are not added then would complain at https://github.com/cBioPortal/cbioportal-frontend/pull/3311/files#diff-8f5a13488274fc21f61a85ce616cf599R689

PR #3323, opened 2020-07-21
Hide mutation assessor on functional impact column

leexgh:
Partial of https://github.com/cBioPortal/cbioportal/issues/7691

- Because mutation assessor server is not stable, adding a parameter `SHOW_MUTATION_ASSESSOR` to hide or show mutation assessor
- by default the `SHOW_MUTATION_ASSESSOR` is set to false since the mutation assessor is down for a while

In the future if we decide to show mutation assessor again, just set `SHOW_MUTATION_ASSESSOR = true`, or remove `SHOW_MUTATION_ASSESSOR` parts completely if mutation assessor is stable.

### DIFF
--- a/src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
@@ -46,6 +46,9 @@ interface FunctionalImpactData {
     polyPhenScore: number | undefined;
     polyPhenPrediction: string | undefined;
 }
+// mutation assessor server is not stable
+// set "SHOW_MUTATION_ASSESSOR = true" to enable mutation assessor
+const SHOW_MUTATION_ASSESSOR = false;
 
 class FunctionalImpactColumnTooltip extends React.Component<
     FunctionalImpactColumnTooltipProps,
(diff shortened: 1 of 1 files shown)


src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
```diff
@@ -46,6 +46,9 @@ interface FunctionalImpactData {
     polyPhenScore: number | undefined;
     polyPhenPrediction: string | undefined;
 }
+// mutation assessor server is not stable
+// set "SHOW_MUTATION_ASSESSOR = true" to enable mutation assessor
+const SHOW_MUTATION_ASSESSOR = false;
```
leexgh, line 51: not sure this is a good place to add the parameter

inodb: @leexgh this can work, but shouldn't the choice of using mutation assessor or not be at a higher level? Like if one disables the use of mutation assessor, we prolly should use a different `field` parameter for genome nexus. If it is enabled it should not

Instead of hardcoding, it might also make sense to introduce the enabling/disabling of mutation assessor as a property. So people can choose what resources of genome nexus to enable/disable. Maybe we can have some property `show.genome_nexus.annotation_sources=hotspots,mutation_assessor,clinvar,gnomad` etc. We can set the default to those we already support. No need to handle all of the various enabling/disabling of sources, maybe just mutation_assessor for now. But this way we can easily add handling enabling/disabling of new annotation sources in the future without having to introduce new properties

PR #3326, opened 2020-07-23
Oncoprint clinical tracks based on custom charts from study view

adamabeshouse:
related to: https://github.com/cBioPortal/cbioportal/issues/7619

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -126,6 +126,7 @@ import {
     computeCustomDriverAnnotationReport,
     DEFAULT_GENOME,
     excludeSpecialMolecularProfiles,
+    ExtendedClinicalAttribute,
     fetchPatients,
     fetchQueriedStudies,
     filterAndAnnotateMutations,
@@ -137,6 +138,7 @@ import {
     groupDataByCase,
     initializeCustomDriverAnnotationSettings,
     isRNASeqProfile,
+    makeCustomChartData,
     OncoprintAnalysisCaseType,
     parseGenericAssayGroups,
 } from './ResultsViewPageStoreUtils';
@@ -993,13 +995,12 @@ export class ResultsViewPageStore {
         },
     });
 
-    readonly clinicalAttributes = remoteData<
-        (ClinicalAttribute & { molecularProfileIds?: string[] })[]
-    >({
+    readonly clinicalAttributes = remoteData<ExtendedClinicalAttribute[]>({
         await: () => [
             this.studyIds,
             this.clinicalAttributes_profiledIn,
             this.clinicalAttributes_comparisonGroupMembership,
+            this.clinicalAttributes_customCharts,
             this.samples,
             this.patients,
         ],
@@ -1043,12 +1044,13 @@ export class ResultsViewPageStore {
                     patientAttribute: true,
                 } as ClinicalAttribute);
             }
-            return serverAttributes
-                .concat(specialAttributes)
-                .concat(this.clinicalAttributes_profiledIn.result!)
-                .concat(
-                    this.clinicalAttributes_comparisonGroupMembership.result!
-                );
+            return [
+                ...serverAttributes,
+                ...specialAttributes,
+                ...this.clinicalAttributes_profiledIn.result!,
+                ...this.clinicalAttributes_comparisonGroupMembership.result!,
+                ...this.clinicalAttributes_customCharts.result!,
+            ];
         },
     });
 
@@ -1072,6 +1074,7 @@ export class ResultsViewPageStore {
             this.studyToDataQueryFilter,
             this.clinicalAttributes_profiledIn,
             this.clinicalAttributes_comparisonGroupMembership,
+            this.clinicalAttributes_customCharts,
         ],
         invoke: async () => {
             let clinicalAttributeCountFilter: ClinicalAttributeCountFilter;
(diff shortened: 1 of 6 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1021,12 +1022,15 @@ export class ResultsViewPageStore {
                     patientAttribute: true,
                 } as ClinicalAttribute);
             }
-            return serverAttributes
-                .concat(specialAttributes)
-                .concat(this.clinicalAttributes_profiledIn.result!)
-                .concat(
-                    this.clinicalAttributes_comparisonGroupMembership.result!
-                );
+            const ret: ExtendedClinicalAttribute[] = [];
```
kalletlak, line 1025: May be directly return value instead of creating new variable.
`return [... serverAttributes, ... specialAttributes, ...]`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1126,6 +1131,12 @@ export class ResultsViewPageStore {
                     }
                 );
             }
+            // add counts for custom chart clinical attributes
+            for (const attr of this.clinicalAttributes_customCharts.result!) {
+                ret[attr.clinicalAttributeId] = attr.data!.filter(d => {
+                    return sampleMap.has(d as any, ['sampleId', 'studyId']);
```
kalletlak, line 1137: Looks like samples are already filtered in `makeCustomChartData`, do you still need this?

adamabeshouse:
thanks @kalletlak , just pushed changes

adamabeshouse:
@kalletlak what do you think?

@tmazor thoughts?

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -48,6 +49,12 @@ import { isSampleProfiled } from 'shared/lib/isSampleProfiled';
 import { AlteredStatus } from './mutualExclusivity/MutualExclusivityUtil';
 import { Group } from '../../shared/api/ComparisonGroupClient';
 import { isNotGermlineMutation } from '../../shared/lib/MutationUtils';
+import ComplexKeySet from 'shared/lib/complexKeyDataStructures/ComplexKeySet';
```
dippindots, line 52: Seems this `ComplexKeySet` has not been used here.

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
```diff
@@ -150,9 +156,14 @@ export default class AddClinicalTracks extends React.Component<
...
 
+            sortedAttributes.customCharts = _.sortBy<ExtendedClinicalAttribute>(
+                sortedAttributes.customCharts,
+                x => x.displayName.toLowerCase()
+            );
+
             return Promise.resolve(
                 _.mapValues(sortedAttributes, attrs => {
-                    return attrs.map(attr => ({
+                    return (attrs as ClinicalAttribute[]).map(attr => ({
```
dippindots, line 166: Do we need to cast this? I thought `attrs` will derived the type from `sortedAttributes`.

adamabeshouse: its this typescript issue where when the array type is ambiguous then it cant resolve the array methods type (here its either ClinicalAttribute[] or ExtendedClinicalAttribute[], but either one works for the return type)

tmazor:
@adamabeshouse I'm not seeing anything, but maybe I'm testing it wrong. I used Charts/Custom Data to add a chart with a name in Study View. Then I used the query box on study view to run a query. When OncoPrint loaded, I couldn't find that custom chart name/data anywhere. What am I doing wrong?

tmazor:
For the record - logging in makes this work 😄 

I do want to test a bit more, but so far it looks great @adamabeshouse ! Just two things:

1) Custom charts are appearing in both the Clinical and Custom Charts lists but should probably only appear in the Custom Charts tab:
image
image
2) Plots tab won't load: https://www.cbioportal.org/results/plots?Action=Submit&cancer_study_list=lgg_ucsf_2014&case_set_id=lgg_ucsf_2014_all&clinicallist=NUM_SAMPLES_PER_PATIENT%2CCUSTOM_FILTERS_3%2CCUSTOM_FILTERS_4&comparison_subtab=clinical&gene_list=TP53&show_samples=true&tab_index=tab_visualize
image

adamabeshouse:
@tmazor I fixed the first error, and couldn't reproduce the second one (maybe it was related to the first somehow?). Could you please check my update when you have time?

adamabeshouse:
@tmazor I fixed it, at the expense of making old plots tab URLs, which pointed to a specific "color by" setting, no longer point to that. Is that acceptable?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -289,7 +289,8 @@ export function getColoringMenuOptionValue(
     option: Omit<ColoringMenuOmnibarOption, 'value'>
 ) {
     return `${option.info.entrezGeneId}_${JSON.stringify(
-        option.info.clinicalAttribute
+        option.info.clinicalAttribute,
+        ['clinicalAttributeId', 'patientAttribute', 'studyId']
```
adamabeshouse, line 293: @alisman @tmazor @jjgao I should have done this from the start. But this will cause plots tab URLs which people have saved to no longer point to the "color by" setting that they used to point to. Is that acceptable? However this will also make the URL much more concise.

PR #3327, opened 2020-07-23
map variant annotation with mutation by originalVariantQuery

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/7707

- map variant annotation with mutation by `originalVariantQuery`

Example:
https://deploy-preview-3327--cbioportalfrontend.netlify.app/results/mutations?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=msk_impact_2017&case_set_id=msk_impact_2017_cnaseq&data_priority=0&gene_list=ABL1&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=msk_impact_2017_cna&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=msk_impact_2017_mutations&mutations_transcript_id=ENST00000318560&profileFilter=0&tab_index=tab_visualize

Filter by `P-0006124-T01-IM5`. 

The genomic location of this mutation is 
```
9,133759441,133759501,GCGCCTTCTCCCCAAAGACAAAAAGACCAACTTGTTCAGCGCCTTGATCAAGAAGAAGAAG,GCGCCTTCTCCCCAAAGACAAAAAGACCAACTTGTTCAGCGCCTTGATCAAGAAGAAGAAGA
```
But in variant annotation is
```
9,133759501,133759502,-,A
```
Mapping with`originalVariantQuery` can make this variant show up in the mutation table

### DIFF
--- a/packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
+++ b/packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
@@ -128,7 +128,6 @@ export function getMutationByTranscriptId(
             (tc: TranscriptConsequenceSummary) =>
                 tc.transcriptId === ensemblTranscriptId
         );
-
     if (
         variantAnnotation &&
         transcriptConsequenceSummaries &&
@@ -148,7 +147,6 @@ export function getMutationByTranscriptId(
         }
         return annotatedMutation as Mutation;
     } else {
-        // if mutation is not annotatable or can't map back (due to genomic location normalization or other reasons), return undefined
         return undefined;
     }
 }
@@ -325,10 +323,8 @@ export function indexAnnotationsByGenomicLocation(
     variantAnnotations: VariantAnnotation[]
 ): { [genomicLocation: string]: VariantAnnotation } {
     return _.keyBy(variantAnnotations, annotation =>
-        annotation.annotation_summary
-            ? genomicLocationString(
-                  annotation.annotation_summary.genomicLocation
-              )
+        annotation.originalVariantQuery
+            ? annotation.originalVariantQuery
             : genomicLocationStringFromVariantAnnotation(annotation)
     );
 }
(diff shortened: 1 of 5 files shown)


leexgh:
@inodb good point! This pr only fix mapping in mutation table, it doesn't touch Hotspots fetching in oncoprint.

leexgh:
@inodb Cancer Hotspots data fetching is using POST /cancer_hotspots/genomic, but only /annotation/ endpoints have `originalVariantQuery`, so I think the hotspots are fine.

Since we use genomic location to map annotations in a lot of places, I change the implementation to have both genomic locations from mutations and `originalVariantQuery` as keys when they are different, so we don't need to change other logic in the codebase.

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -324,12 +322,23 @@ export function findCanonicalTranscript(
...
+                      annotation.annotation_summary.genomicLocation
+                  )
+                : genomicLocationStringFromVariantAnnotation(annotation);
+            indexAnnotationsByGenomicLocation[key] = annotation;
+            if (key !== annotation.originalVariantQuery) {
+                indexAnnotationsByGenomicLocation[
+                    annotation.originalVariantQuery
+                ] = annotation;
+            }
+            return indexAnnotationsByGenomicLocation;
```
inodb, line 339: @leexgh could you add some comments in the code that describe what you mentioned in the PR? From what I understand this is mostly to avoid changing the logic in other places. Using genomicLocation to map back won't always be correct, so don't you think it's best to use `originalVariantQuery` everywhere? E.g. we could use the annotation endpoint instead to get the hotspots. If it's too much work to change everywhere, one option is to add a `// TODO:` explaining what needs to be done before we can use `originalVariantQuery` everywhere

leexgh: @inodb I see, I found we also use genomic location to query myVariantInfo so I did this temp fix to avoid too much code changes. But yes using `originalVariantQuery` is more clear. I'm not exactly sure how many places we need to change, shouldn't be too much I guess. I'll try to change them all using annotation endpoint and mapping with `originalVariantQuery`

PR #3329, opened 2020-07-27
Add a study view warning

Leowisd:
Fix https://github.com/cBioPortal/cbioportal/issues/4065

Default value will be removed when it comes from backend

### DIFF
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -17,6 +17,7 @@ import {
 import { buildCBioPortalPageUrl } from 'shared/api/urls';
 import ErrorScreen from 'shared/components/errorScreen/ErrorScreen';
 import { ServerConfigHelpers } from 'config/config';
+import StudyViewWarning from 'pages/studyView/studyPageHeader/studyViewWarning/StudyViewWarning';
 
 interface IContainerProps {
     location: Location;
@@ -90,6 +91,15 @@ export default class Container extends React.Component<IContainerProps, {}> {
 
                 <div className="pageTopContainer">
                     <UserMessager />
+                    {localStorage.getItem('private_study_link_warning') !==
+                        'Yes' &&
+                        AppConfig.serverConfig.skin_study_view_blurb && (
+                            <StudyViewWarning
+                                message={
+                                    AppConfig.serverConfig.skin_study_view_blurb
+                                }
+                            />
+                        )}
                     <div className="contentWidth">
                         <PortalHeader appStore={this.appStore} />
                     </div>
--- a/src/config/serverConfigDefaults.ts
+++ b/src/config/serverConfigDefaults.ts
@@ -45,6 +45,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_description:
         'The cBioPortal for Cancer Genomics provides visualization, analysis and download of large-scale cancer genomics data sets',
     show_genomenexus: true,
+    skin_study_view_blurb:
+        'All URLs in this website are private - do NOT include in manuscripts.',
     skin_authorization_message:
         'Access to this portal is only available to authorized users.',
     skin_documentation_about: 'About-Us.md',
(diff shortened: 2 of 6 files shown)


src/config/serverConfigDefaults.ts
```diff
@@ -45,6 +45,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_description:
         'The cBioPortal for Cancer Genomics provides visualization, analysis and download of large-scale cancer genomics data sets',
     show_genomenexus: true,
+    skin_study_view_blurb:
```
zhx828, line 48: This default will be removed before merging. The value should come from the server which https://github.com/cBioPortal/cbioportal/pull/7718 is trying to address that.

alisman:
@Leowisd could you guys put a screenshot in here so i can see appearance of warning?

Leowisd:
Like this
image

alisman:
@jjgao @zhx828 do we want to do a little extra work to make this warning dismissable (stored in localstorage)

alisman:
@Leowisd the spacing around that alert isn't great.   lets wait to see what JJ says about dismissability.  if we user can dismiss warning, then it could be above the study and that would ease positioning.

inodb:
Not sure if the warning should be this big/visible, what do you think @zhx828 @jjgao ? Another option would be to append the text to the description of the study. E.g. similar to https://cbioportal.mskcc.org/study/summary?id=mskimpact. There is some red text warning

zhx828:
> Not sure if the warning should be this big/visible, what do you think @zhx828 @jjgao ? Another option would be to append the text to the description of the study. E.g. similar to https://cbioportal.mskcc.org/study/summary?id=mskimpact. There is some red text warning

Good point, I believe that red text is very similar to the message we are about to add. I will defer the decision to the @cBioPortal/product team.

Leowisd:
So if like this
image

alisman:
@inodb @zhx828 i would be in favor of a dismisseable warning.  That we can get people to notice without taking up real estate permanently.

schultzn:
But then they will dismiss the warning and forget about it, and use the URL in a paper again. I am not sure that we have a better choice here.

Leowisd:
image
image
image

src/appShell/App/Container.tsx
```diff
@@ -90,6 +91,14 @@ export default class Container extends React.Component<IContainerProps, {}> {
 
                 <div className="pageTopContainer">
                     <UserMessager />
+                    {localStorage.getItem('warning_dismiss') !== 'Yes' &&
```
alisman, line 94: that key should be more specific to this particular case.   'private_study_link_warning'

src/pages/studyView/studyPageHeader/studyViewWarning/StudyViewWarning.tsx
```diff
@@ -0,0 +1,186 @@
...
+        super(props);
+
+        this.handleHide = this.handleHide.bind(this);
+        this.handleShow = this.handleShow.bind(this);
+        this.handleDismiss = this.handleDismiss.bind(this);
+        this.handleModalShow = this.handleModalShow.bind(this);
+        this.handleModalHide = this.handleModalHide.bind(this);
+        this.handleChecked = this.handleChecked.bind(this);
+
+        this.state = {
```
alisman, line 29: we don't use React state.  We use mobx observables.   Not a big deal but you should familiarize yourself with how it works.   Look at any other component and you will see see @observable decorator on instance properties.  And you'll see @observer decorator on the class.  This wires up render to observe changes to observable properties exactly as setState does.

src/pages/studyView/studyPageHeader/studyViewWarning/StudyViewWarning.tsx
```diff
@@ -0,0 +1,186 @@
...
+}
+
+export default class StudyViewWarning extends React.Component<
+    IStudyViewWarningBannerProps,
+    IStudyViewWarningBannerStates
+> {
+    constructor(props: any) {
+        super(props);
+
+        this.handleHide = this.handleHide.bind(this);
```
alisman, line 22: you can do this with @autobind decorator

alisman: that way, you don't have to bother with this stuff in constructor

PR #3333, opened 2020-07-27
fix colors change when filters are applied

Leowisd:
Fix https://github.com/cBioPortal/cbioportal/issues/5893

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -104,6 +104,7 @@ import {
     getGenomicDataAsClinicalData,
     convertGenomicDataBinsToClinicalDataBins,
     getGenomicChartUniqueKey,
+    pickNewColorForClinicData,
 } from './StudyViewUtils';
 import MobxPromise from 'mobxpromise';
 import { SingleGeneQuery } from 'shared/lib/oql/oql-parser';
@@ -288,6 +289,9 @@ export type OncokbCancerGene = {
 export class StudyViewPageStore {
     private reactionDisposers: IReactionDisposer[] = [];
 
+    private chartItemToColor: Map<string, string>;
+    private chartToUsedColors: Map<string, Set<string>>;
+
     public studyViewQueryFilter: StudyViewURLQuery;
 
     @observable showComparisonGroupUI = false;
@@ -297,6 +301,8 @@ export class StudyViewPageStore {
         private sessionServiceIsEnabled: boolean,
         private urlWrapper: StudyViewURLWrapper
     ) {
+        this.chartItemToColor = new Map();
+        this.chartToUsedColors = new Map();
         this.reactionDisposers.push(
             reaction(
                 () => this.loadingInitialDataForSummaryTab,
@@ -2488,18 +2494,74 @@ export class StudyViewPageStore {
                             .clinicalAttributeId,
                     } as ClinicalDataCountItem);
                     let counts: ClinicalDataCount[] = [];
+                    let attributeId: string = '';
                     if (data !== undefined) {
                         counts = data.counts;
+                        attributeId = data.attributeId;
+                        if (!this.chartToUsedColors.has(attributeId))
+                            this.chartToUsedColors.set(attributeId, new Set());
                     }
-                    return getClinicalDataCountWithColorByClinicalDataCount(
+
+                    let res = getClinicalDataCountWithColorByClinicalDataCount(
                         counts
                     );
+                    res.forEach(item => {
+                        let colorMapKey = this.generateColorMapKey(
+                            attributeId,
+                            item.value
+                        );
+                        // If the item doesn't has an assigned color
+                        if (!this.chartItemToColor.has(colorMapKey)) {
+                            // If the color has not been used
+                            if (
+                                !this.chartToUsedColors
+                                    .get(attributeId)
+                                    ?.has(item.color)
+                            ) {
+                                this.chartItemToColor.set(
+                                    colorMapKey,
+                                    item.color
+                                );
+                                this.chartToUsedColors
+                                    .get(attributeId)
+                                    ?.add(item.color);
+                            } else {
+                                // Pick up a new color if the color has been used
+                                let d = {
+                                    value: item.value,
+                                    count: item.count,
+                                };
+                                let newColor = pickNewColorForClinicData(
+                                    d,
+                                    this.chartToUsedColors.get(attributeId) ||
+                                        new Set()
+                                );
+                                this.chartItemToColor.set(
+                                    colorMapKey,
+                                    newColor
+                                );
+                                this.chartToUsedColors
+                                    .get(attributeId)
+                                    ?.add(newColor);
+                                item.color = newColor;
+                            }
+                        } else {
+                            item.color = this.chartItemToColor.get(
+                                colorMapKey
+                            )!;
+                        }
+                    });
+                    return res;
                 },
                 onError: error => {},
                 default: [],
             });
         }
         return this.clinicalDataCountPromises[uniqueKey];
+    }
+
+    private generateColorMapKey(id: string, value: string) {
+        return `${id}.${value}`;
     }
 
     public getClinicalDataBin(chartMeta: ChartMeta) {
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -1843,6 +1843,39 @@ export function pickClinicalAttrColorsByIndex(
     );
 }
 
+export function pickUnusedColor(
+    data: ClinicalDataCount[],
+    usedColors: Set<string>,
+    availableColors: string[] = COLORS
+): { [attribute: string]: string } {
+    let colorIndex = 0;
+
+    return _.reduce(
+        data,
+        (acc: { [id: string]: string }, slice) => {
+            if (
+                !isNAClinicalValue(slice.value) &&
+                !getClinicalValueColor(slice.value)
+            ) {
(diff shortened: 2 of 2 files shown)


src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -354,6 +359,26 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+                    this.props.promise.result.map((item: any) => {
+                        this.colorMap.set(
+                            this.props.title + '.' + item.value,
+                            item.color
+                        );
+                    });
+                } else if (this.colorMap.size > 0) {
+                    for (let i = 0; i < pieData.length; i++) {
+                        pieData[i].color = this.colorMap.get(
+                            this.props.title + '.' + pieData[i].value
```
zhx828, line 377: Let's use string template literals for future string concatenation.

zhx828: Also you need a method to generate the key so you would not need to duplicate the logics.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -354,6 +359,26 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
     get chart() {
         switch (this.chartType) {
             case ChartTypeEnum.PIE_CHART: {
+                let pieData = this.props.promise.result;
+
+                if (
```
zhx828, line 364: This condition is not valid when the initial call is already filtered, so the dataset will be smaller. But when user resets the filter, you will not get the color anymore for values that previous filtered out.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -136,6 +139,8 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
 
         this.chartType = this.props.chartType;
 
+        this.colorMap = new Map();
```
zhx828, line 142: @alisman what's our take on the Map vs Object in the project? I barely used Map, is it something we can mix?

adamabeshouse:
I agree with @zhx828 's feedback

Leowisd:
@zhx828 New commit, please check.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2485,12 +2493,49 @@ export class StudyViewPageStore {
...
+                                        count: item.count
+                                    }]
+                                    let newColor = pickNewColorForClinicData(d, this.chartColorMap.get(attributeId) || new Set())
+                                    this.colorMap.set(colorMapKey, newColor[0])
+                                    this.chartColorMap.get(attributeId)?.add(newColor[0])
+                                    item.color = newColor[0]
+                                }
+                            }
+                        }
+                        else item.color = this.colorMap.get(colorMapKey)
```
adamabeshouse, line 2535: lets just put this in braces, we don't really use one line else blocks

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1894,6 +1894,22 @@ export function pickClinicalDataColors(
...
+    d: ClinicalDataCount[],
+    usedColors: Set<string>,
+    colors: string[] = COLORS
+) {
+    let availableColors: string[] = _.cloneDeep(colors);
+    for (let item of usedColors) {
+        let index = availableColors.indexOf(item);
+        if (index > -1) availableColors.splice(index, 1);
+    }
+    const fixedColors = pickClinicalAttrFixedColors(d);
```
adamabeshouse, line 1907: lets keep an eye on this because I think that repeatedly cloning this array could be slow if this function is run many times. is there another way to do it, e.g. by rewriting `pickClinicalAttrColorsByIndex` to just skip over used colors?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -288,6 +289,10 @@ export type OncokbCancerGene = {
 export class StudyViewPageStore {
     private reactionDisposers: IReactionDisposer[] = [];
 
+    private colorMap : Map<string, string>;
+    private chartSet : Set<string>;
```
adamabeshouse, line 293: I do feel like we already have an object keeping track of which charts are active, by chart id rather than by clinical attribute id. so I dont think chartSet is necessary. right @zhx828 ?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -288,6 +289,10 @@ export type OncokbCancerGene = {
 export class StudyViewPageStore {
     private reactionDisposers: IReactionDisposer[] = [];
 
+    private colorMap : Map<string, string>;
```
adamabeshouse, line 292: how about `chartItemToColor`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -288,6 +289,10 @@ export type OncokbCancerGene = {
...
+    private chartSet : Set<string>;
+    private chartColorMap : Map<string, Set<string>>;
```
adamabeshouse, line 294: how about `chartToUsedColors`

src/pages/studyView/StudyViewPageStore.ts
zhx828: Yeah, but the logic here is that you just want to keep the color consistent. Why bother to check whether the chart is available? You could generate color counts using getClinicalDataCountWithColorByClinicalDataCount, then do a quick check whether the (attribtue id + attribute value) exists in the color cache, if it is then replace.

adamabeshouse:
something went wrong with the latest commit, it includes a lot of random stuff

Leowisd:
Yeah, I'm fixing it.

Leowisd:
Discard this commit, git commit log occurs some problem. I'm fixing it...

Leowisd:
Current commit is correct

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2485,12 +2491,43 @@ export class StudyViewPageStore {
...
+                        attributeId = data.attributeId;
+                        if (!this.chartToUsedColors.has(attributeId))
+                            this.chartToUsedColors.set(attributeId, new Set());
                     }
-                    return getClinicalDataCountWithColorByClinicalDataCount(
+
+                    let res = getClinicalDataCountWithColorByClinicalDataCount(
                         counts
                     );
+                    res.map((item: any) => {
```
adamabeshouse, line 2505: while the effect of `map` as you're using it will be the same as `forEach`, it's more semantic to use `forEach` since it more easily indicates that the purpose of the loop is to iterate rather than to return a new array

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1843,6 +1843,35 @@ export function pickClinicalAttrColorsByIndex(
     );
 }
 
+export function pickClinicalAttrColorByIndex(
```
adamabeshouse, line 1846: thanks for making this change! I think the name is too detailed for users, maybe just `pickUnusedColor`

Leowisd:
Thanks for your comments! I've fixed those.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1894,6 +1923,15 @@ export function pickClinicalDataColors(
     };
 }
 
+export function pickNewColorForClinicData(
```
zhx828, line 1926: This method is designed to pick color for one ClinicalDataCount not the list of ClinicalDataCount. Otherwise the line 1931 is not correct. Maybe change the signature to one datum?

Leowisd: I reused the 'pickClinicalAttrFixedColors' which needs a list of ClinicalDataCount so I used a list of ClinicalDataCount as signature. And I think this function also return a list of  color string. ._values() will return a list right?

zhx828: `pickClinicalAttrFixedColors ` does not mean you need a list as input.
Also it returns the fixed colors, imagine you have a list of ClinicalDataCount(more than 1 elements), if only one is fixed, then your method returns only one color. The number does not match the length of ClinicalDataCount

Leowisd: The ClinicalDataCount list of this function's input must only contain one element. So if there is one fixed color, just return it. It won't appear such length not match problem I think.

zhx828: This is not enforced though right? The type is a list, so you cannot guarantee whoever uses this method will only pass one element array.

Leowisd: Oh, I got you. I will fix it.

Leowisd: Please check

Leowisd:
Upgrade the version of Prettier to 1.19.0 to support new feature, Optional Chaining, from Typescript 3.7 and above

package.json
```diff
@@ -319,7 +319,7 @@
     "karma-webpack": "^3.0.5",
     "mocha": "^5.2.0",
     "pre-commit": "^1.2.2",
-    "prettier": "1.18.2",
+    "prettier": "1.19.0",
```
zhx828, line 322: @Leowisd can we actually upgrade to the latest patch version, in this case, 1.19.1?

Leowisd: Oh, sure.

Leowisd:
There is a separate PR https://github.com/cBioPortal/cbioportal-frontend/pull/3340 to upgrade the version of Prettier only.

PR #1232, opened 2018-06-06
Fixed LazyMobxTable filter reset issue 

onursumer:
# What? Why?
Fix https://github.com/cBioPortal/cbioportal-frontend/issues/1071.

Changes proposed in this pull request:
- Restoring the `store.dataStore.filterString` value for `LazyMobXTable` after each re-render

# Checks
- [ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!
- [ ] Follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [ ] Follows the [Airbnb React/JSX Style guide](https://github.com/airbnb/javascript/tree/master/react).
- [ ] Make sure your commit messages end with a Signed-off-by string (this line
  can be automatically added by git if you run the `git-commit` command with
  the `-s` option)

PR #3334, opened 2020-07-28
Add clear-filter-X button to datasets page search box

adamabeshouse:
image

### DIFF
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.tsx
@@ -82,6 +82,7 @@ type LazyMobXTableProps<T> = {
     itemsLabel?: string;
     itemsLabelPlural?: string;
     showFilter?: boolean;
+    showFilterClearButton?: boolean;
     showCopyDownload?: boolean;
     copyDownloadProps?: ICopyDownloadControlsProps;
     showPagination?: boolean;
@@ -249,7 +250,7 @@ function getDownloadObject<T>(columns: Column<T>[], rowData: T) {
 }
 
 export class LazyMobXTableStore<T> {
-    public filterString: string | undefined;
+    @observable public filterString: string | undefined;
     @observable private _itemsLabel: string | undefined;
     @observable private _itemsLabelPlural: string | undefined;
     @observable public sortColumn: string;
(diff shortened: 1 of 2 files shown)


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -249,7 +250,7 @@ function getDownloadObject<T>(columns: Column<T>[], rowData: T) {
 }
 
 export class LazyMobXTableStore<T> {
-    public filterString: string | undefined;
+    @observable public filterString: string | undefined;
```
adamabeshouse, line 253: I wonder why this wasnt observable before... hopefuly nothing breaks

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -954,10 +963,32 @@ export default class LazyMobXTable<T> extends React.Component<
                                 className="form-control tableSearchInput"
                                 style={{ width: 200 }}
                             />
+                            {this.props.showFilterClearButton &&
```
alisman, line 966: maybe we we swap it with the magnifying glass instead of having two next to each other?

adamabeshouse: good idea, updated

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
alisman: i actually feel like there was a reason.  let me test it out and think on it.

alisman: @adamabeshouse unless there is a strong reason to make it observable (like, we are fixing something known to be broken), lets just leave this as it was.

onursumer: @alisman @adamabeshouse according to my comment [here](https://github.com/cBioPortal/cbioportal-frontend/pull/1232/files#diff-c2cf724d243fce1017b7b24de3932de8R197) we were not observing that variable anywhere. I guess it is okay to make it observable as long as we don't end up with infinite rendering or performance issues.

PR #3335, opened 2020-07-28
Redirect the linkout endpoint "ln" to curated non-redundant list of studies

adamabeshouse:


src/shared/lib/redirectHelpers.ts
```diff
@@ -107,17 +111,54 @@ export function handleCaseDO() {
...
+        (studyIds, buttonName) => {
+            const [buttonText, tooltipText] = buttonName.split('|');
+            return buttonText
+                .toLowerCase()
+                .includes('curated set of non-redundant studies');
+        }
+    );
+    if (curatedNonRedundantStudyIdsArray) {
+        // filter out studies that user doesnt have access to
+        const allStudies = await client.getAllStudiesUsingGET({});
```
alisman, line 138: way to anticipate this.  did you manage to test this?

adamabeshouse: yeah, there's actually one in the curated list that's not available in the public portal lol

alisman: so, has that been failing?

tmazor:
I see this is already merged and on production but I'm testing anyway 😄 

It looks good, but there are a few small things that I noticed:
1) Using ln?q=TP53 brings me to the Mutations tab rather than Cancer Type Summary -- that's not necessarily an issue, just a thing I've noticed
2) When I tried to run the same query from the homepage for comparison, I noticed that selecting the curated study list says 178 studies & 46959 samples
image
But after running the quer, I have 47005 samples?
image
3) We should update the last line on the Web API page to say "curated set of non-redundant studies":
image

@adamabeshouse @alisman

alisman:
@tmazor sorry i should have given you an opportunity to test it.   @adamabeshouse could the discrepancy have something to do with the study that isn't actually publically accessible?

alisman:
Also, @tmazor @adamabeshouse i will check but i think the sample count on homepage comes from counts stored in studies table which are not kept in sync with dynamic counts.  I will bring this up in standup

adamabeshouse:
@alisman Just to summarize what we said in stand-up, I don't think that's a factor because the full list is 179 studies, minus the non-accessible one is 178, which is the number we're seeing in those screenshots.

@tmazor yes, I also thought it was weird to send to mutations, but that's actually the existing behavior

tmazor:
@adamabeshouse I don't think I'd ever used this ln?... link before, so I don't know what it did before. Maybe there's a historical reason why that behavior was desired? @jjgao do you know why the ln?... link goes to the Mutations tab? In principle I would prefer if it brought a user to Cancer Types Summary to be consistent with running a query from the homepage, but I don't really think it matters all that much so it's probably ok to leave as is.

PR #3338, opened 2020-07-30
Disable custom GA tracking events for all except MSK managed portals…

alisman:
… (leaves out-of-box GA)

Fixes https://github.com/cBioPortal/cbioportal/issues/7671

### DIFF
--- a/src/shared/lib/tracking.ts
+++ b/src/shared/lib/tracking.ts
@@ -37,14 +37,18 @@ export function initializeTracking() {
 }
 
 export function trackEvent(event: GAEvent) {
-    getGAInstance()(
-        'send',
-        'event',
-        event.category,
-        event.action,
-        event.label,
-        event.fieldsObject
-    );
+    // trackEvent is used to send custom UI events which may depend on custom configuration within GA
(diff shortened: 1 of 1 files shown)


src/shared/lib/tracking.ts
```diff
@@ -37,14 +37,18 @@ export function initializeTracking() {
...
-        'send',
-        'event',
-        event.category,
-        event.action,
-        event.label,
-        event.fieldsObject
-    );
+    // trackEvent is used to send custom UI events which may depend on custom configuration within GA
+    // for other installers, we want to shut these off, leaving them with bare-bones out-of-box GA implemenation
+    if (/cbioportal\.org$/.test(getBrowserWindow().location.host)) {
```
alisman, line 42: @inodb i think this should catch all our portals, or at least those we are tracking with GA. whachu think?

inodb: maybe also track *.mskcc.org portals?

alisman: @inodb good catch! updated

PR #3344, opened 2020-08-03
Add study id to plots tooltip when theres more than one study

adamabeshouse:


src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -2110,8 +2111,36 @@ function tooltipClinicalDataSection(
     }
 }
 
+function sampleIdForTooltip<D extends IPlotSampleData>(
+    d: D,
+    studiesInQuery: CancerStudy[]
```
kalletlak, line 2116: may be this should be a`Map` since it is lookup for study

adamabeshouse: yeah I guess you're right, got lazy

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -470,7 +470,11 @@ export default class ClinicalData extends React.Component<
...
             if (this.boxPlotData.isComplete) {
-                content = boxPlotTooltip(d, this.boxPlotData.result.horizontal);
+                content = boxPlotTooltip(
+                    d,
+                    this.props.store.studies.result!,
+                    this.boxPlotData.result.horizontal
+                );
             } else {
                 content = (
                     <span>
```
alisman, line 480: should we take out this strange message ("shouldn't appear")

PR #3354, opened 2020-08-07
Fix bug in venn diagram when it cant find a good layout

adamabeshouse:
This fixes https://github.com/cbioportal/cbioportal/issues/7735


zhx828:
@adamabeshouse I updated your comment to include the issue.

PR #3356, opened 2020-08-12
Make sure reference genome json is parsed regardless of contentype

alisman:
On public portal we load json from a deployed file (to save hit on api).  Load is dependent on content type of response which is server dependent.


inodb:
Thanks! Looks like this fixed it

PR #3361, opened 2020-08-18
Include splice hotspots on mutation table

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/3789
image

Test: https://deploy-preview-3361--cbioportalfrontend.netlify.app/results/mutations?session_id=595d61a7498e5df2e2921610&show_samples=false&plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D

Splice mutations will show hotspots if available.

- Include splice hotspots on mutation table and oncoprint. Add a new category `splice hotspot`. If the mutation type is `splice`, then get splice hotspots data
- remove redundant hotspot util functions in `AnnotationUtils.ts`, using replacement functions from `cbioportal-utils` package
- move tests in `AnnotationUtils.spec.ts` to `CancerHotspotsUtils.spec.ts`, since we are using hotspot util functions from package


packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
```diff
@@ -128,6 +144,9 @@ export function isHotspot(
 export function defaultHotspotFilter(hotspot: Hotspot) {
     const type = hotspot.type.toLowerCase();
     return (
-        type.includes('single') || type.includes('indel') || type.includes('3d')
+        type.includes('single') ||
+        type.includes('indel') ||
+        type.includes('3d') ||
+        type.includes('splice')
```
leexgh, line 150: not sure if I should add `splice` into default filter

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -170,7 +171,10 @@ export function getAnnotationData(
                 hotspotData &&
                 hotspotData.result &&
                 hotspotData.status === 'complete'
-                    ? isRecurrentHotspot(mutation, hotspotData.result)
+                    ? // if mutation type is splice, get splice hotspot, otherwise get recurrent hotspot
+                      mutation.mutationType?.toLowerCase().includes('splice')
+                        ? isSpliceHotspot(mutation, hotspotData.result)
+                        : isRecurrentHotspot(mutation, hotspotData.result)
```
leexgh, line 177: Could you confirm the logic here? If the mutation type is `splice`, then check if the hotspots data has `splice` type, for all other types of mutations check if it has recurrent hotspots(`single` and `indel`)

src/shared/lib/AnnotationUtils.ts
```diff
@@ -35,7 +35,8 @@ export function recurrentHotspotFilter(hotspot: Hotspot) {
     // only single and indel mutations are regular hotspots
     return (
         hotspot.type.toLowerCase().includes('single') ||
-        hotspot.type.toLowerCase().includes('indel')
+        hotspot.type.toLowerCase().includes('indel') ||
+        hotspot.type.toLowerCase().includes('splice')
```
leexgh, line 39: This function only been used in getting oncoprint hotspots data (https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/ResultsViewPageStore.ts#L4283), not sure if I should add `splice` as one of the filter type?

leexgh:
There are some screenshot tests failed that I don't really get why, maybe it's because oncoprint hotspot was changed?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
onursumer: Should we just change the `isRecurrentHotspot` logic and include splice there instead? That would simplify the logic. Also I think you can simply do this: 
```suggestion
                    ? isSpliceHotspot(mutation, hotspotData.result) || isRecurrentHotspot(mutation, hotspotData.result)
```

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
onursumer: I think it is okay If splice hotspots are considered like any other recurrent hotspot

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
```diff
@@ -106,6 +106,15 @@ export function filter3dHotspotsByMutations(
     );
 }
 
+export function filterSpliceHotspotsByMutations(
+    mutations: Mutation[],
+    index: IHotspotIndex
+): Hotspot[] {
+    return filterHotspotsByMutations(mutations, index, (hotspot: Hotspot) =>
+        hotspot.type.toLowerCase().includes('splice')
```
onursumer, line 114: We may also consider adding this condition into `filterRecurrentHotspotsByMutations`. Is there any specific reason that we may want to treat splice hotspots as a separate category? We needed to do this for 3d hotspots for visualization purposes, but not sure if this is necessary for splice mutations.

src/shared/lib/AnnotationUtils.ts
onursumer: This should be okay, assuming that splice hotspots are recurrent hotspots. I think this filter was added mainly to exclude 3d hotspots.

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
leexgh: although splice hotspots are not considered like recurrent hotspot, I think we still should add it here. Because this is the default filter, and we use this filter on getting overall hotspots data like `hotspotCount` or `hotspotsByPosition`, we have other filters like `filterSpliceHotspotsByMutations` when we need a particular type of hotspot data

packages/react-mutation-mapper/src/component/column/Annotation.tsx
leexgh: as we discussed, only splice mutations should match splice hotspots

src/shared/lib/AnnotationUtils.ts
leexgh: This is fixed now. We need to apply the same "splice logic" to oncoprint hotspot data

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
leexgh: because only splice mutations should match splice hotspots as we discussed

packages/react-mutation-mapper/src/component/column/Annotation.tsx
onursumer: Makes sense.

Can we avoid using nested ternary operators though? They are not very easy to read, and quite easy to mess up.

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.spec.ts
```diff
@@ -209,6 +376,77 @@ describe('CancerHotspotsUtils', () => {
...
+                'SMURF1 R101N should be a splice hotspot mutation.'
+            );
+        });
+    });
+
+    describe('isNon3dHotspot', () => {
+        it('checks if a mutation is a hotspot mutation by isNon3dHotspot()', () => {
+            assert.isTrue(
+                isNon3dHotspot(hotspotSpliceMutation, spliceHotspotIndex),
+                'SMURF1 R101N should be a non-3d(splice) hotspot mutation.'
```
inodb, line 440: @leexgh this one I don't fully understand. `SMURF1 R101N` means amino acid R changes to N and by definition is not splice but missense. It also does not seem to be hotpot in https://www.cancerhotspots.org/#/home or https://www.3dhotspots.org/#/home. So shouldn't this return false?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4284,7 +4284,8 @@ export class ResultsViewPageStore {
                 const indexedHotspotData = this.indexedHotspotData.result;
                 if (indexedHotspotData) {
                     return Promise.resolve((mutation: Mutation) => {
-                        return isRecurrentHotspot(mutation, indexedHotspotData);
+                        // if mutation type is splice, get splice hotspot, otherwise get recurrent hotspot
```
inodb, line 4287:
```suggestion
```
this comment is now handled inside of `isNon3dHotspot`

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.spec.ts
```diff
@@ -209,6 +376,77 @@ describe('CancerHotspotsUtils', () => {
...
+            );
+        });
+    });
+
+    describe('isSpliceHotspot', () => {
+        it('checks if a mutation is a hotspot mutation by isSpliceHotspot()', () => {
+            assert.isFalse(
+                isSpliceHotspot(notHotspotMutation, spliceHotspotIndex),
+                'MET X1028_splice should not be a splice hotspot mutation.'
+            );
```
inodb, line 427: maybe good to add MET X1010 as well as an expected hotspot in the tests. Just b/c the originally issue specifically mentions that one as the one missing. Good to proof that it's working in a test. Maybe you already have this somewhere but I didn't see it (I did see it in the test data)

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.spec.ts
leexgh: @inodb oops I messed up the description message😢 , the logic and tests are correct - it tests `MET X1010`, and this message should be 
```
'MET X1010 should be a non-3d(splice) hotspot mutation.'
```

onursumer: @inodb I think this is a made up mock data that I added long time ago, that's why the gene and amino acid combination do not make any sense :)

inodb: ah ok that makes sense, i should've checked the code as well 😅

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.spec.ts
leexgh: @inodb yes I have a test for it but got messes up with messages... let me update it

packages/cbioportal-utils/src/hotspot/CancerHotspotsUtils.ts
```diff
@@ -106,6 +106,15 @@ export function filter3dHotspotsByMutations(
...
+    );
+}
+
 export function isRecurrentHotspot(
```
inodb, line 118: I know naming of this already existed, but might be worth revisiting. The 3d hotspots were also determined using recurrence, so maybe LinearHotspot or LinearClusterHotspot is better? Using terminology from Lopez-Bigas lab: https://twitter.com/nlbigas/status/1295351606855245825/photo/1.

leexgh:
we still use `recurrentHotspot` as a name in other places, e.g. https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/react-mutation-mapper/src/component/hotspot/HotspotInfo.tsx#L61, but I guess it's fine to leave them there

PR #3362, opened 2020-08-19
Fix reload of ResultsView components after opening annotation config menu when custom driver annotations are present

pvannierop:
# Problem
When custom driver annotations are present and activated in portal.properties, opening the configuration menu in ResultsView triggers re-rendering of tab components (see video 1). Also, custom driver annotation data is not incorporated when loading the oncoprint for the first time (see incorrect legend labels in video 1).

# Analysis
The re-rendering is triggered by a store update via the variable `customDriverAnnotationReport` in the store that is first observed when the config menu is opened.

# Partial solution
I have added `customDriverAnnotationReport` as observed variable to ResultsViewOncoprint and PlotsTab. This solves the problem for these componens.

# Assistance required
Since this solution has to be applied to all ResultsView components individually, I think a more elegant solution that does this evaluation at a more central place would be preferred. I do not know what such a solution would look like. Help and advise here would be welcomed.

## Video 1 (before)
Peek 2020-08-19 08-32

## Video 2 (after)
Peek 2020-08-19 08-36

### DIFF
--- a/src/pages/resultsView/plots/PlotsTab.tsx
+++ b/src/pages/resultsView/plots/PlotsTab.tsx
@@ -4809,7 +4809,11 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                                 <Observer>{this.controls}</Observer>
                             ) : (
                                 <LoadingIndicator
-                                    isLoading={true}
+                                    isLoading={
+                                        this.props.store
+                                            .customDriverAnnotationReport
+                                            .isPending
+                                    }
                                     center={true}
                                     size={'big'}
                                 />
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -1553,7 +1553,8 @@ export default class ResultsViewOncoprint extends React.Component<
                 this.props.store.molecularProfileIdToMolecularProfile,
                 this.alterationTypesInQuery,
                 this.alteredKeys,
-                this.heatmapTrackHeaders
+                this.heatmapTrackHeaders,
+                this.props.store.customDriverAnnotationReport
             ) === 'pending'
         );
     }


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -4809,7 +4809,11 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                                 <Observer>{this.controls}</Observer>
                             ) : (
                                 <LoadingIndicator
-                                    isLoading={true}
+                                    isLoading={
+                                        this.props.store
```
adamabeshouse, line 4813: why?

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -4809,7 +4809,11 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                                 <Observer>{this.controls}</Observer>
                             ) : (
                                 <LoadingIndicator
-                                    isLoading={true}
+                                    isLoading={
```
alisman, line 4812: we should probably package this up with other loading states (e.g. dataTypeOptions.pending) 
 i.e. it's odd that we show don't show loader for those promises, but do for this one.

adamabeshouse:
generally, hacking mobx isnt the best solution, our code should reflect the actual logic of the situation if possible. so now my question is, why aren't the oncoprint and plots tab reflecting their dependency on this? what mobxpromise or observer which actually should `await` customDriverAnnotationReport, and thus put it into the chain of dependency, isn't awaiting it?

pvannierop:
@adamabeshouse @alisman I agree fully that this is not the correct solution. This 'fix' only covers Oncoprint and Plots Tab, but the problem persists in other Results View component. I hope you guys can prose a better solution here. My knowledge of MobX and cBioPortral frontend architecture is not sufficient here.

adamabeshouse:
@pvannierop I understand. the beauty of mobx is that if you arrange the code logically then things will rerender/recompute exactly when their dependencies change. so my analysis here is that there is some error, because these results view components logically rely on `customDriverAnnotationReport` but are not depending on them in the code, which is why your fix to manually insert this dependency in is fixing the issue. so what I am proposing is that you look into where exactly the components rely on `customDriverAnnotationReport` and insert that dependency logically there. I predict that there is somewhere that is implicitly relying on it but not explicitly awaiting it from within an observer. Does that make sense?

pvannierop:
@adamabeshouse Sorry for not responding sooner. Yes, some help from your side in this analysis would be very welcome. I will close this PR since we are refactoring this aspect in another PR.

PR #3364, opened 2020-08-19
Fix e2e tests that were broken by new data

adamabeshouse:
Fixes https://github.com/cbioportal/cbioportal/issues/7776

### DIFF
--- a/end-to-end-test/remote/specs/core/studyview.spec.js
+++ b/end-to-end-test/remote/specs/core/studyview.spec.js
@@ -627,10 +627,10 @@ describe('the gene panel is loaded properly', () => {
     });
 });
 
-describe.skip('submit genes to results view query', () => {
+describe('submit genes to results view query', () => {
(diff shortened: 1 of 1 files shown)


inodb (commented):
Thanks a ton!

There a few more in `home.spec.js` e.g.:

- [ ] https://github.com/adamabeshouse/cbioportal-frontend/blob/e2e/end-to-end-test/remote/specs/core/home.spec.js#L613
- [ ] https://github.com/adamabeshouse/cbioportal-frontend/blob/e2e/end-to-end-test/remote/specs/core/home.spec.js#L634
- [ ] https://github.com/adamabeshouse/cbioportal-frontend/blob/e2e/end-to-end-test/remote/specs/core/home.spec.js#L696

Note that for some of these it might make sense to use `localdb` tests at some point in the future. Those are a more reliable. These tests seem to mostly test the functionality of OQL. For this case there doesn't seem to be much added benefit to testing live data

adamabeshouse:
@inodb okay cool, lets address those in a different issue/PR

PR #3367, opened 2020-08-20
Add support for Uniprot PTM data

onursumer:
Fix cBioPortal/cbioportal#7815

## Checks

To test multiple PTM tracks in cBioPortal type this in console and refresh the page:

```JavaScript
localStorage.frontendConfig = JSON.stringify({serverConfig: {ptmSources: ["Uniprot", "dbPTM"]}});
```


packages/react-mutation-mapper/src/component/ptm/PtmAnnotationTable.tsx
```diff
@@ -46,34 +56,44 @@ export default class PtmAnnotationTable extends React.Component<
...
         const pubmedRender = !_.isEmpty(this.pmidData)
-            ? (props: { original: PostTranslationalModification }) => (
-                  <PtmReferenceList
-                      pmidData={this.pmidData}
-                      pubmedIds={props.original.pubmedIds}
-                  />
-              )
-            : () =>
+            ? (props: { original: PostTranslationalModification }) =>
+                  !_.isEmpty(props.original.pubmedIds) ? (
```
onursumer, line 60: this is to prevent empty tooltip in case the pubmed id list is empty

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -724,9 +790,9 @@ class DefaultMutationMapperStore implements MutationMapperStore {
         {
             await: () => [this.cancerHotspotsData],
             invoke: async () =>
-                this.ptmData.result
```
onursumer, line 727: this seems like a bug, we shouldn't be waiting for ptm data to process cancer hotspots data.

packages/cbioportal-utils/src/model/Uniprot.ts
```diff
@@ -0,0 +1,31 @@
+// Temporary model for Uniprot data
```
onursumer, line 1: We will remove this file once we merge genome-nexus/genome-nexus#442

packages/cbioportal-utils/src/ptm/PtmUtils.ts
```diff
@@ -0,0 +1,368 @@
+import * as _ from 'lodash';
+
+import { PostTranslationalModification as DbPtm } from 'genome-nexus-ts-api-client';
```
onursumer, line 3: We should probably rename `PostTranslationalModification` to `DbPtm` on the GenomeNexus side

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -649,10 +659,66 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
+        },
+        []
+    );
+
+    readonly dbPtmData: MobxPromise<DbPtm[]> = remoteData(
         {
             await: () => [this.mutationData, this.activeTranscript],
             invoke: async () => {
-                if (this.activeTranscript.result) {
+                if (this.activeTranscript.result && this.swissProtId.result) {
```
leexgh, line 721: Why do we need to check `swissProtId` here?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -649,10 +659,66 @@ class DefaultMutationMapperStore implements MutationMapperStore {
     );
 
     readonly ptmData: MobxPromise<PostTranslationalModification[]> = remoteData(
+        {
+            await: () => [this.uniprotPtmData],
```
leexgh, line 663: Do we need to wait for `dbPtmData` here?

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -62,6 +62,8 @@ import {
 
 export interface IMutationMapperStoreConfig {
     filterMutationsBySelectedTranscript?: boolean;
+    // TODO make ptmSource part of AppConfig? (IMutationMapperConfig)
+    ptmSource?: string;
```
leexgh, line 66: I'm thinking since we are supporting two sources, should we add an info icon or indicator to show which source of data you are seeing?

packages/react-mutation-mapper/src/component/ptm/PtmAnnotationTable.tsx
```diff
@@ -46,34 +56,44 @@ export default class PtmAnnotationTable extends React.Component<
...
+                      <PtmReferenceList
+                          pmidData={this.pmidData}
+                          pubmedIds={props.original.pubmedIds || []}
+                      />
+                  ) : null
+            : (props: { original: PostTranslationalModification }) =>
+                  !_.isEmpty(props.original.pubmedIds) &&
                   this.props.pubMedCache ? (
                       <i className="fa fa-spinner fa-pulse" />
                   ) : null;
```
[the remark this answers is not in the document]
onursumer: Right :) Actually, this is even more confusing because the ternary operation is not nested but what we set is a function that has ternary operation within. This is a workaround to trigger re-render when the pmid data updates. I'll try to make this more clear.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: swissProtId is needed to get the uniprot accession. But this is temporary, we should be only using the transcript id when querying for the PTMs.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: Nice catch! We should probably only wait for the PTM source set with `this.ptmSource`.

src/shared/components/mutationMapper/MutationMapperStore.ts
onursumer: Good point. I guess we can also update the existing PTM info tooltip to indicate the PTM data source.

packages/cbioportal-utils/src/model/PostTranslationalModification.ts
```diff
@@ -0,0 +1,35 @@
+export interface PostTranslationalModification {
+    type: string;
+    residue: {
+        start: number;
+        end?: number;
```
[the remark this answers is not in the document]
onursumer: Hmm, I guess we can always set `start` and `end` to the same value if there is only one position value provided (which is the case for dbPTM). So, it seems like there is no need to make `end` optional.

onursumer:
To test multiple PTM tracks in cBioPortal type this in console and refresh the page:
```JavaScript
localStorage.frontendConfig = JSON.stringify({serverConfig: {ptmSources: ["Uniprot", "dbPTM"]}});
```

packages/cbioportal-utils/src/ptm/PtmUtils.ts
```diff
@@ -0,0 +1,377 @@
...
+
+    if (!ptmType) {
+        if (ptm.description.toLowerCase().includes('ubiquitin')) {
+            ptmType = PtmType.Ubiquitination;
+        } else if (ptm.description.toLowerCase().includes('sumo')) {
+            ptmType = PtmType.Sumoylation;
+        }
+    }
+
+    return ptmType || PtmType.Other;
```
leexgh, line 333: maybe like this? I feel like `PtmType.Other` probably better to be in `else` logic
```
    if (!ptmType) {
        if (ptm.description.toLowerCase().includes('ubiquitin')) {
            ptmType = PtmType.Ubiquitination;
        } else if (ptm.description.toLowerCase().includes('sumo')) {
            ptmType = PtmType.Sumoylation;
        } else {
            ptmType = PtmType.Other
        }
    }

    return ptmType;
```

packages/cbioportal-utils/src/ptm/PtmUtils.ts
```diff
@@ -0,0 +1,377 @@
...
+    // if the description is empty
+    if (_.isEmpty(description)) {
+        description = ptm.type;
+
+        if (!_.isEmpty(description)) {
+            // assuming type is almost always all uppercase,
+            // capitalize the first letter, and set the rest to lowercase
+            description =
+                description.charAt(0).toUpperCase() +
+                description.slice(1).toLowerCase();
```
leexgh, line 348: _.capitalize(description) can also do this job :)

packages/react-mutation-mapper/src/component/track/TrackPanel.tsx
```diff
@@ -25,7 +31,7 @@ type TrackPanelProps = {
 @observer
 export default class TrackPanel extends React.Component<TrackPanelProps, {}> {
     public static defaultProps: Partial<TrackPanelProps> = {
-        tracks: [TrackName.CancerHotspots, TrackName.OncoKB, TrackName.PTM],
+        tracks: [TrackName.CancerHotspots, TrackName.OncoKB, TrackName.dbPTM],
```
leexgh, line 34: Just wanted to make sure that only `dbPTM` is default and `UniprotPTM` is not?

packages/react-mutation-mapper/src/component/track/TrackPanel.tsx
```diff
@@ -102,4 +98,42 @@ export default class TrackPanel extends React.Component<TrackPanelProps, {}> {
...
+            this.props.trackVisibility[trackName] === 'visible' ? (
+            <PtmTrack
+                store={this.props.store}
+                pubMedCache={this.props.pubMedCache}
+                ensemblTranscriptId={
+                    this.props.store.activeTranscript &&
+                    this.props.store.activeTranscript.result
+                        ? this.props.store.activeTranscript.result
+                        : undefined
+                }
```
leexgh, line 113: Maybe let's shorten it to:
```
ensemblTranscriptId={
                    this.props.store.activeTranscript?.result
                }
```

packages/react-mutation-mapper/src/component/track/TrackSelector.tsx
```diff
@@ -41,7 +41,7 @@ export default class TrackSelector extends React.Component<
         tracks: [
             TrackName.CancerHotspots,
             TrackName.OncoKB,
-            TrackName.PTM,
+            TrackName.dbPTM,
```
leexgh, line 44: Same here, just wanted to make sure that only `dbPTM` is default and `UniprotPTM` is not?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -685,9 +695,87 @@ class DefaultMutationMapperStore<T extends Mutation>
...
+                this.allTranscripts,
+            ],
             invoke: async () => {
+                let uniprotId: string | undefined;
+
                 if (this.activeTranscript.result) {
+                    const transcript = this.transcriptsByTranscriptId[
+                        this.activeTranscript.result
+                    ];
+                    uniprotId = transcript ? transcript.uniprotId : undefined;
```
leexgh, line 753: Same here, maybe shorten it to 
`uniprotId = transcript?.uniprotId;`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -56,6 +56,7 @@ import {
     IHotspotIndex,
     indexHotspotsData,
     IOncoKbData,
+    PtmSource,
```
leexgh, line 59: Are we using this?

leexgh (commented):
Thanks! This is great!

packages/cbioportal-utils/src/ptm/PtmUtils.ts
```diff
@@ -0,0 +1,377 @@
...
+export function groupPtmDataByTypeAndPosition(
+    ptmData: PostTranslationalModification[]
+) {
+    const groupedByType = _.groupBy(ptmData, p => p.type);
+
+    const groupedByTypeAndPosition: {
+        [type: string]: { [position: number]: PostTranslationalModification[] };
+    } = {};
+
+    _.keys(groupedByType).forEach(type => {
```
alisman, line 261: minor but _.map will iterate over an object and give you (key,value)=>{}

packages/cbioportal-utils/src/ptm/PtmUtils.ts
```diff
@@ -0,0 +1,377 @@
...
+    } else if (a > b) {
+        return 1;
+    } else if (b > a) {
+        return -1;
+    }
+
+    return 0;
+}
+
+function getResiduesFromUniportFeature(ptm: UniprotFeature) {
```
alisman, line 301: looks like these three functions have type on their name "uniport" ?   should it be "uniprot"?

packages/cbioportal-utils/src/ptm/PtmUtils.ts
alisman: agree.

packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
```diff
@@ -99,18 +134,33 @@ export default class PtmTrack extends React.Component<PtmTrackProps, {}> {
...
-                    codon: Number(position),
-                    color: ptmColor(ptmDataByProteinPosStart[Number(position)]),
-                    tooltip: (
-                        <PtmTooltip
-                            ptms={ptmDataByProteinPosStart[Number(position)]}
-                            pubMedCache={this.props.pubMedCache}
-                        />
-                    ),
-                }));
+            return _.compact(
```
alisman, line 137: this whole thing could be written as a single reduce function.  which would save 3 iteration cycles

packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
```diff
@@ -128,27 +178,39 @@ export default class PtmTrack extends React.Component<PtmTrackProps, {}> {
...
-                                        ptmDataByTypeAndProteinPosStart[type][
-                                            Number(position)
-                                        ]
-                                    }
-                                    pubMedCache={this.props.pubMedCache}
-                                />
-                            ),
-                        })),
-                }));
+                    specs: _.compact<TrackItemSpec>(
```
alisman, line 181: same here

packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
```diff
@@ -161,15 +223,25 @@ export default class PtmTrack extends React.Component<PtmTrackProps, {}> {
...
                     destroyTooltipOnHide={true}
                 >
                     <span
                         style={{ marginLeft: 4 }}
                         onClick={this.handleToggleExpand}
                     >
-                        PTM Sites <i className="fa fa-info-circle" />
+                        PTM{' '}
+                        {this.props.dataSource ? (
+                            <>({this.props.dataSource})</>
```
alisman, line 240: i don't think you need these <></>.  i don't think they do anything here

packages/react-mutation-mapper/src/component/track/TrackPanel.tsx
alisman: yeah, is there really a case where activeTranscript would be undefined itself?

packages/react-mutation-mapper/src/component/track/TrackPanel.tsx
onursumer: Correct, `dbPTM` is the default for cBioPortal.

src/pages/resultsView/ResultsViewPageStore.ts
onursumer: Good catch. I'll remove it.

packages/cbioportal-utils/src/model/Uniprot.ts
onursumer: I think eventually we want to get Uniprot features from Genome Nexus instead of directly using Uniprot API. genome-nexus/genome-nexus#442 does not implement a cache yet so I guess we can update this part later.

packages/react-mutation-mapper/src/component/track/TrackPanel.tsx
onursumer: Theoretically yes. `activeTranscript` is defined as an optional field for `MutationMapperStore` interface: https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/react-mutation-mapper/src/model/MutationMapperStore.ts#L30

Can't remember if we have an actual store without an `activeTranscript` implementation though.

inodb:
Unrelated failing e2e tests also failing in master

PR #3370, opened 2020-08-21
Support structural variant profile in result page plots tab

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/1879

To test
https://cbioportal-fusion-sv-mi-btpzf1.herokuapp.com/ and setting `localStorage.netlify = "deploy-preview-3370--cbioportalfrontend"`

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1502,7 +1502,9 @@ export class ResultsViewPageStore {
                     profile.molecularAlterationType ===
                         AlterationTypeConstants.GENESET_SCORE ||
                     profile.molecularAlterationType ===
-                        AlterationTypeConstants.GENERIC_ASSAY
+                        AlterationTypeConstants.GENERIC_ASSAY ||
+                    profile.molecularAlterationType ===
+                        AlterationTypeConstants.STRUCTURAL_VARIANT
                 ) {
                     // geneset profile, we dont have the META projection for geneset data, so just add it
                     /*promises.push(internalClient.fetchGeneticDataItemsUsingPOST({
@@ -2664,16 +2666,14 @@ export class ResultsViewPageStore {
         [studyId: string]: MolecularProfile;
     }>(
         {
-            await: () => [this.molecularProfilesInStudies],
+            await: () => [this.mutationProfiles],
             invoke: () => {
-                const ret: { [studyId: string]: MolecularProfile } = {};
-                for (const profile of this.molecularProfilesInStudies.result) {
-                    const studyId = profile.studyId;
-                    if (!ret[studyId] && isMutationProfile(profile)) {
-                        ret[studyId] = profile;
-                    }
-                }
-                return Promise.resolve(ret);
+                return Promise.resolve(
+                    _.keyBy(
+                        this.mutationProfiles.result,
+                        (profile: MolecularProfile) => profile.studyId
+                    )
+                );
             },
         },
         {}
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -1131,6 +1165,25 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                 });
             },
 
+            get _structuralVariantCountBy() {
+                const urlSelection =
+                    (vertical
```
Luke-Sikina, line 1170: 99% sure this exceeds the scope of this PR, but I don't like all the `vertical ? {} : {}` boilerplate we have. I think a better solution would be two have a type that represents all the axis dependant state, then maintain a Map<Boolean, ThatState>, with two entries, one for vertical == true, and one for vertical == false.

That way statements like this just turn into `this.state.axisState[vertical].plots_selection`.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -5103,6 +5115,46 @@ export class ResultsViewPageStore {
...
+                            molecularProfileId:
+                                studyIdToProfileMap[sample.studyId]
+                                    .molecularProfileId,
+                            sampleId: sample.sampleId,
+                        });
+                    }
+                    return memo;
+                },
+                [] as StructuralVariantFilter['sampleMolecularIdentifiers']
+            );
```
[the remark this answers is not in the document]
kalletlak: I remember we used to do this before. but for performance reason(iterate once instead of twice) we decided to use `.reduce`

src/pages/resultsView/plots/PlotsTab.tsx
kalletlak: Yeah, this should a separate effort

src/pages/resultsView/ResultsViewPageStore.ts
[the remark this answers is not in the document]
Luke-Sikina: Ok, fair enough. Maybe one day Javascript will support lazy evaluation ootb, and we won't have to choose.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -827,6 +848,17 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
...
             },
+            get structuralVariantCountBy() {
+                if (this._structuralVariantCountBy === undefined) {
+                    // default
+                    return StructuralVariantCountBy.VariantType;
+                } else {
+                    return this._structuralVariantCountBy;
+                }
+            },
+            set structuralVariantCountBy(m: StructuralVariantCountBy) {
```
onursumer, line 859: Very minor suggestions:
```suggestion
            set structuralVariantCountBy(s: StructuralVariantCountBy) {
```

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -5103,6 +5115,46 @@ export class ResultsViewPageStore {
...
+        invoke: async () => {
+            const studyIdToProfileMap = this
+                .studyToStructuralVariantMolecularProfile.result!;
+
+            if (_.isEmpty(studyIdToProfileMap)) {
+                return Promise.resolve([]);
+            }
+
+            const filters = this.samples.result.reduce(
+                (memo, sample: Sample) => {
```
dippindots, line 5135: Do we need to specify this type `Sample` at here? Can't it be derived from the `this.samples.result`?

kalletlak: Not required. It is mostly for us to know what type it is.

dippindots: Got it!

PR #3382, opened 2020-08-31
timeline - Special render in sample tracks for multiple simultaneous samples

adamabeshouse:
fixes https://github.com/cBioPortal/cbioportal/issues/7849

-----
#### SCREENSHOTS

https://genie-private.cbioportal.org/patient?studyId=nsclc_genie_bpc&caseId=GENIE-DFCI-000036
image

https://genie-private.cbioportal.org/patient?caseId=GENIE-MSK-P-0012057&studyId=nsclc_genie_bpc
image

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
@@ -165,8 +165,8 @@ export function renderPoint(
     //  we'll show that render in the tooltip.
     if (events.length === 1 && events[0].render) {
         return events[0].render(events[0]);
-    } else if (events.length === 1 && trackData.renderEvent) {
-        return trackData.renderEvent(events[0]);
+    } else if (trackData.renderEvents) {
+        return trackData.renderEvents(events);
     } else {
         return (
             <g>
--- a/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
@@ -213,6 +213,12 @@ export const TimelineTrack: React.FunctionComponent<ITimelineTrackProps> = obser
 
         if (trackData.items) {
             eventsGroupedByPosition = groupEventsByPosition(trackData.items);
+            if (trackData.sortSimultaneousEvents) {
+                eventsGroupedByPosition = _.mapValues(
+                    eventsGroupedByPosition,
+                    trackData.sortSimultaneousEvents
+                );
+            }
         }
 
         let trackValueRange: { min: number; max: number };
--- a/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
@@ -82,7 +82,7 @@ function getLineChartYCoordinate(
     return padding + (1 - plottingProportion) * plottingHeight; // 1-p because SVG y axis points down
 }
 
-function renderSuperscript(number: number) {
+export function renderSuperscript(number: number) {
     return (
         <g transform={'translate(3 -8)'}>
             <text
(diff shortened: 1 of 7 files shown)


packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
```diff
@@ -82,9 +82,9 @@ function getLineChartYCoordinate(
     return padding + (1 - plottingProportion) * plottingHeight; // 1-p because SVG y axis points down
 }
 
-function renderSuperscript(number: number) {
+export function renderSuperscript(number: number) {
```
alisman, line 85: i think we should start adopting this to functional components when we can.

alisman: but you don't need to change this

src/pages/patientView/timeline2/TimelineWrapperUtils.ts
```diff
@@ -0,0 +1,80 @@
...
+        return DEFAULT_LABEL;
+    }
+
+    // aggregate into ranges
+    const ranges: { start: number; end: number }[] = [];
+    let currentRange = {
+        start: sortedNumbers[0],
+        end: sortedNumbers[0],
+    };
+    for (let i = 1; i < sortedNumbers.length; i++) {
```
alisman, line 35: i think we should establish a rule of never using native for loops unless there's an optimization case.   i find them harder to read because you have to use the index to get element and also, there's classic issue where length is read every time unecessarily. granted here it's negligible.

adamabeshouse: I agree with you but in this case I need to start with the second element. would be sweet if there was a nicer way to do that besides if index ===0 continue

src/pages/patientView/timeline2/TimelineWrapper.tsx
```diff
@@ -344,7 +366,7 @@ const TimelineWrapper: React.FunctionComponent<ITimeline2Props> = observer(
                             { re: /improving/i, color: 'rgb(44, 160, 44)' },
                             { re: /worsening/i, color: 'rgb(214, 39, 40)' },
                         ];
-                        cat.items.forEach((event, i) => {
+                        function getEventColor(event: TimelineEvent) {
```
alisman, line 369: lets just pull this out since it's used again below.  and pass in the string constant

alisman (approved):

inodb:
@adamabeshouse This looks awesome!! Some small things I noticed:

- [ ] in the example https://genie-private.cbioportal.org/patient?studyId=nsclc_genie_bpc&caseId=GENIE-DFCI-000036. When I go through the different tooltips of the sample acquisition, it goes `1 -> 3 -> 2` instead of `1 -> 2 -> 3`. And for sequencing it does `3 -> 2` instead of `2 -> 3`. See screenshot, would expect `(2) 2 of 3`:
    Screen Shot 2020-09-03 at 11 08 55 AM
- [ ] there should maybe be some more slack when deciding whether something is on the same day. See e.g. here: https://genie-private.cbioportal.org/patient?caseId=GENIE-MSK-P-0002917&studyId=nsclc_genie_bpc. 
    Screen Shot 2020-09-03 at 11 14 07 AM
    Maybe better that at a certain zoom level it's more of a range instead of the exact same day?
- [ ] the top right number of the stacked circle icon gets cut off here: https://genie-private.cbioportal.org/patient?caseId=GENIE-DFCI-003865&studyId=nsclc_genie_bpc
    Screen Shot 2020-09-03 at 11 19 15 AM

First one would be nice to fix, the other ones maybe not as essential

alisman:
@inodb i'm not sure i agree about the slack.   i think it might be hard to do well.  and that's what zoom is for.    maybe zoom is not obvious enough?

PR #3385, opened 2020-09-01
Enhancement #7742 download option to lollipop plot on mutation tab

jagnathan:

- a
- b

## Checks

## Any screenshots or GIFs?

## Notify reviewers


src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+    protected getDownloadData(dataType?: DataType): string {
+        var flatdata: any = '';
+        if (
+            this.selectedGene &&
+            this.props.store.getMutationMapperStore(this.selectedGene)
+        ) {
+            var mutationMapperStore: ResultsViewMutationMapperStore;
+            mutationMapperStore = this.props.store.getMutationMapperStore(
+                this.selectedGene
+            )!;
```
onursumer, line 256:
```suggestion
            const mutationMapperStore = this.props.store.getMutationMapperStore(
                this.selectedGene
            )!;
```

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+            let sidToStudy = mutationMapperStore.studyIdToStudy.result!;
+            mutations.forEach(m => {
+                let study: any;
+                study = sidToStudy[m.studyId];
+                Object.assign(m, { StudyName: study.name });
+                downloadData.push(m);
+            });
+
+            flatdata = tsvFormat(downloadData);
+        }
```
onursumer, line 268: Let's move this into `ResultsViewMutationMapperStore` and just do this: 
```suggestion
            flatData = mutationMapperStore.getDownloadData();
        }
```

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
         }
     }
 
+    @autobind
+    @action
+    protected getDownloadData(dataType?: DataType): string {
+        var flatdata: any = '';
```
onursumer, line 248:
```suggestion
        let flatdata: string = '';
```

alisman: yes, lets always use const or let

alisman: and ts will infer the type, so  `const flatdata = ''`  is sufficient

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+        if (
+            this.selectedGene &&
+            this.props.store.getMutationMapperStore(this.selectedGene)
+        ) {
+            var mutationMapperStore: ResultsViewMutationMapperStore;
```
alisman, line 253: const

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+            mutationMapperStore = this.props.store.getMutationMapperStore(
+                this.selectedGene
+            )!;
+            let mutations = mutationMapperStore.getMutationData();
+            let downloadData: any[] = [];
+            let sidToStudy = mutationMapperStore.studyIdToStudy.result!;
+            mutations.forEach(m => {
```
alisman, line 260: this is actually a mapping operation, so should be expressed 
```
const downloadData = mutations.map((m)=>...)
```

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+                let study: any;
+                study = sidToStudy[m.studyId];
+                Object.assign(m, { StudyName: study.name });
```
alisman, line 263: this is actually going to mutate the mutation data.  would be better to do 
```
Object.assign({},m,{...})
```
that will create new object with properties of both 2nd and 3rd args without mutating m

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
         }
     }
 
+    @autobind
+    @action
```
alisman, line 246: this doesn't need to be @action because it doesn't do update mobxstate.  it just returns data, so it's actually not an "action".

src/pages/resultsView/mutation/Mutations.tsx
alisman: yeah, since it's an abstract operation we're doing here, i agree it belongs in the store where it could be used elsewhere if necessary

src/pages/resultsView/mutation/Mutations.tsx
onursumer: Let's name the field `studyName`, not `StudyName`

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -238,6 +242,33 @@ export default class Mutations extends React.Component<
...
+            var mutationMapperStore: ResultsViewMutationMapperStore;
+            mutationMapperStore = this.props.store.getMutationMapperStore(
+                this.selectedGene
+            )!;
+            let mutations = mutationMapperStore.getMutationData();
+            let downloadData: any[] = [];
+            let sidToStudy = mutationMapperStore.studyIdToStudy.result!;
+            mutations.forEach(m => {
+                let study: any;
+                study = sidToStudy[m.studyId];
```
onursumer, line 262:
```suggestion
                const study = sidToStudy[m.studyId];
```

alisman: onur is right, you can just use type inference here.   in general, try to avoid using 'any' at all costs.  there are sometimes it can't be avoided or is very convenient.  but it should be used sparingly.

alisman:
@jagnathan  Noticed a couple more things: 
1. gene column is not properly populated.  It's getting an object when it should get a string
image

2. The downloaded file should have .tsv extension i think.  not .txt.

onursumer:
@alisman @jagnathan Not sure if it is a big deal but column names are not as user friendly as the output file generated by the table download option. Also, there are some overlapping columns but this is not the exact data that we get from the table download function. I don't know if getting the exact same output is important though.

jjgao:
I guess we should close this one?

jagnathan:
The chart is part of the package react-mutation-mapper used by other projects. Further the table below the chart provides the option of downloading TSV with some additional info. So this PR is closed.

PR #3389, opened 2020-09-02
Remove deprecated DAT-related property

ao508:
The (optional) argument for revoking existing tokens before issuing a
new one to a user is being removed from the backend. This PR brings the
frontend up-to-date.

Also added the updated generated cbio api docs.

Relevant PR to backend: https://github.com/cBioPortal/cbioportal/pull/7859

Signed-off-by: Angelica Ochoa <15623749+ao508@users.noreply.github.com>

### DIFF
--- a/packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal-docs.json
+++ b/packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal-docs.json
@@ -943,13 +943,6 @@
             "type": "object"
           },
           {
-            "name": "myAllowRevocationOfOtherTokens",
-            "in": "query",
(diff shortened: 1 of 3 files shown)


sheridancbio (approved):
There was a time when this property was used in the frontend code. If now there are no references to it anywhere else (I assume we have scanned the frontend), then the deletion is likely when we broke the compatibility with UUID tokens. 

Also, once this code is deployed we may have to update our deployment files to not attempt to pass this setting in the application configuration. (even if we don't have to, we should clean up property files and deployment files to remove references to this non-existent property)

adamabeshouse (approved):

inodb (approved):

PR #3391, opened 2020-09-03
Timeline: scrollbar overlaps bottom row when timeline is zoomed

alisman:
Fixes https://github.com/cbioportal/cbioportal/issues/7812

Before:
image

After:
image

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/Timeline.tsx
+++ b/packages/cbioportal-clinical-timeline/src/Timeline.tsx
@@ -249,10 +249,12 @@ const Timeline: React.FunctionComponent<ITimelineProps> = observer(function({
     width,
 }: ITimelineProps) {
     const tracks = store.data;
-    const height =
+    const SCROLLBAR_PADDING = 15;
+    let height =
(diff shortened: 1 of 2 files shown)


adamabeshouse (approved):

packages/cbioportal-clinical-timeline/src/Timeline.tsx
```diff
@@ -249,10 +249,12 @@ const Timeline: React.FunctionComponent<ITimelineProps> = observer(function({
...
 }: ITimelineProps) {
     const tracks = store.data;
-    const height =
+    const SCROLLBAR_PADDING = 15;
+    let height =
         TICK_AXIS_HEIGHT +
         _.sumBy(tracks, t => t.height) +
-        _.sumBy(customTracks || [], t => t.height(store));
+        _.sumBy(customTracks || [], t => t.height(store)) +
+        SCROLLBAR_PADDING;
```
alisman, line 257: @adamabeshouse this does mean that there will be buffer of white space below timeline when there's no scrollbar.  but that's better than making everything jump when we zoom

PR #3394, opened 2020-09-08
Improve setup of development environment for localdb e2e-tests 

pvannierop:
Now promoted the use of the docker-compose solution and uses steps consistent with tests for remote e2e-tests.


inodb (approved):
Great! Thank you!

README.md
```diff
@@ -174,63 +174,94 @@ If all goes well, you will soon seen the end-to-end tests running via your Chrom
...
+export PATH=$PATH:$(yarn global dir)/node_modules/webdriver-manager/bin
+```
+
+and start webdriver-manager:
 ```
+webdriver-manager update
+webdriver-manager start
+```
+
+6. In a second terminal, build and serve the frontend version under test:
```
alisman, line 245: I might add here that this the normal setup for portal development, so most people should already have this set up.   i.e. if one is already developing portal, it doesn't need to be separate

pvannierop: I know. But under step 7 there are slight differences so I reasoned it to be better to show all steps. You think we should remove these steps?

alisman: Not sure what you mean here.  I would just say 
```
If you do not already have the frontend dev server running at localhost:3000: In a second terminal ...
```

pvannierop: @alisman I updated the test to:

```
6. If not already running at _localhost:3000_, open a second terminal and start the frontend dev server:
```

inodb:
@alisman @pvannierop is this good to go?

pvannierop:
@inodb @alisman It is as far as I am concerned.

PR #3396, opened 2020-09-09
Fix localdb e2e-tests caused by changes in study_es_0

pvannierop:
Fix localdb e2e-tests caused by changes in study_es_0. Refactored a test class a bit so tests are not-interdependent.


alisman:
@pvannierop looks like still one failure

pvannierop:
@alisman @inodb I fixed the localdb tests. Other tests are failing (API, remote e2e), but this seems unrelated to updates of study_es_0.

inodb (approved):
Thanks so much!

PR #3397, opened 2020-09-09
Dynamic x-axis for GSVA scores in volcano plot selector

pvannierop:
Rather than hard-coded -1 to 1, the x-axis range is determined based on the data. The range is determined by:

1. Get largest absolute score value
2. Round to first integer
3. The range is -(rounded value) to +(rounded value)
3. Ticks are placed at [ -(rounded value), -0.5*(rounded value), 0, +0.5*(rounded value), +(rounded value) ]

For default score range the menu still looks the same:
image


adamabeshouse (approved):

src/shared/components/query/GenesetsVolcanoSelector.tsx
```diff
@@ -87,6 +74,46 @@ export default class GenesetsVolcanoSelector extends QueryStoreComponent<
         );
     }
 
+    // assumes that values are symmetrically centered around 0
+    @computed get xAxisTickSize() {
+        if (this.props.plotData && this.props.plotData.length > 0) {
```
alisman, line 79: this kinda looks testworthy, but since this is really a hyve feature, you can decide

alisman (approved):

PR #3399, opened 2020-09-10
Study page: hide gene specific tab by default when data is still loading

kalletlak:
Fixes https://github.com/cBioPortal/cbioportal/issues/7867


src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -192,6 +192,7 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
     get hideGeneSpecificTab() {
         return (
             this.props.disableGeneSpecificTab ||
+            this.props.store.molecularProfileOptions.isPending ||
```
alisman, line 195: Should it just be !isComplete so that we don't show it in error state?

alisman (approved):

PR #3404, opened 2020-09-15
Timeline: roll out patient view Timeline 2.0 for all portals

alisman:
Turn on new timeline for all portals (except where legacy is configured)


inodb (approved):
LGTM, thanks!

PR #3408, opened 2020-09-17
update oncokb and portal clients

zhx828:


zhx828:
@inodb @alisman I also noticed a change in portal. Let me know if this is not expected to be changed.

inodb (approved):
Thanks this looks good to me. The cBioPortal changes are expected

PR #3410, opened 2020-09-18
Add iframe on homepage sidebar that leads to offsite installation map

jtquach1:
Resolves [#8003](https://github.com/cBioPortal/cbioportal/issues/8003)

- Render clickable iframe that leads to offsite installation map
- Added server configurations for installation map URL, whether iframe should be rendered

## Any screenshots or GIFs?
Before (live site)
image

After (local)
after_links_to_map
after_demo

### DIFF
--- a/src/appShell/App/PortalHeader.tsx
+++ b/src/appShell/App/PortalHeader.tsx
@@ -90,6 +90,14 @@ export default class PortalHeader extends React.Component<
                 hide: () =>
                     AppConfig.serverConfig.skin_show_about_tab === false,
             },
+
+            {
+                id: 'installation-map',
+                text: 'cBioPortal Installations',
+                address: '/installations',
+                internal: false,
+                hide: () => !AppConfig.serverConfig.installation_map_url,
+            },
         ];
     }
 
--- a/src/config/serverConfigDefaults.ts
+++ b/src/config/serverConfigDefaults.ts
@@ -168,6 +168,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
 
     saml_logout_local: false,
     patient_view_use_legacy_timeline: false,
+
+    installation_map_url: 'https://installationmap.netlify.app/',
 };
 
 export default ServerConfigDefaults;
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import * as _ from 'lodash';
-import BarGraph from '../barGraph/BarGraph';
 import { observer } from 'mobx-react';
 import { TypeOfCancer as CancerType } from 'cbioportal-ts-api-client';
 import Testimonials from '../testimonials/Testimonials';
@@ -8,7 +7,7 @@ import AppConfig from 'appConfig';
 import { QueryStore } from 'shared/components/query/QueryStore';
 import { Link } from 'react-router';
 import LoadingIndicator from '../loadingIndicator/LoadingIndicator';
-import { redirectToStudyView } from '../../api/urls';
+import { buildCBioPortalPageUrl, redirectToStudyView } from '../../api/urls';
 import { ResultsViewTab } from '../../../pages/resultsView/ResultsViewPageHelpers';
 
 interface IRightBarProps {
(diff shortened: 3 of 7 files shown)


src/config/serverConfigDefaults.ts
```diff
@@ -61,6 +61,9 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_query_max_tree_depth: '3',
     skin_right_nav_show_data_sets: true,
     skin_right_nav_show_examples: true,
+    skin_right_nav_show_installations: true,
+    installations_url: 'http://localhost:3006',
```
Luke-Sikina, line 65: My bet is that we don't need to make this configurable. The same url should work for everyone.

src/shared/components/rightbar/RightBar.tsx
```diff
@@ -157,6 +157,29 @@ export default class RightBar extends React.Component<
         }
     }
 
+    private getInstallationMap() {
+        return AppConfig.serverConfig.skin_right_nav_show_installations &&
```
Luke-Sikina, line 161: So I think rather than make this a configuration, we're going to show the map only if this is cbioportal.org. To do this, you can just do something like this:

```suggestion
        return AppConfig.serverConfig.app_name === 'public-portal' &&
```

Luke-Sikina: Good start. We can talk about this on Monday, mostly after we meet with JJ

src/config/serverConfigDefaults.ts
jtquach1: I just removed the two lines and hardcoded installations_url inside RightBar.tsx and PortalHeader.tsx.

jtquach1:
Current view of internal page displaying installation map iframe:
update_internal_link

jtquach1:
Current view:
image

src/appShell/App/PortalHeader.tsx
```diff
@@ -90,6 +90,14 @@ export default class PortalHeader extends React.Component<
                 hide: () =>
                     AppConfig.serverConfig.skin_show_about_tab === false,
             },
+
+            {
+                id: 'installation-map',
+                text: 'cBioportal Installations',
```
Luke-Sikina, line 96: Just noticed this
```suggestion
                text: 'cBioPortal Installations',
```

jtquach1:
Current view of the demo homepage:
image

src/appShell/App/PortalHeader.tsx
```diff
@@ -90,6 +90,14 @@ export default class PortalHeader extends React.Component<
                 hide: () =>
                     AppConfig.serverConfig.skin_show_about_tab === false,
             },
+
+            {
+                id: 'installation-map',
+                text: 'cBioPortal Installations',
+                address: '/installations',
+                internal: false,
+                hide: () => AppConfig.serverConfig.app_name !== 'public-portal',
```
jjgao, line 99: Maybe we should add a configuration in the properties file (turn off by default)? @alisman

alisman: @jtquach1 yeah we're trying not to use the name property like this.  reach out to me and i can tell you how to add configuration

jtquach1:
Current view (revised wording)
image

Luke-Sikina (approved):
Don't forget to squash your commits or give a good reason why you're keeping them seperate.

jtquach1:
Current view: 
image
image

src/appShell/App/PortalHeader.tsx
jtquach1: @alisman would this be done in `src/config/serverConfigDefaults.ts`?

jtquach1:
Current view:
image
image
image

Issue:
image
- Linked Google Form in iframe is not scrollable
  - Scrolling is disabled by default in the iframe

jtquach1:
> Current view:
> image
> image
> image
> 
> Issue:
> image
> 
> * Linked Google Form in iframe is not scrollable
>   
>   * Scrolling is disabled by default in the iframe

Reenabled scrolling in the static page iframe and [disabled scrolling in the installation map body](https://github.com/jtquach1/installation-map/pull/24) so one can scroll through the Google Forms link when clicked through the iframe.

PR #3412, opened 2020-09-18
Add oncoprint track data for patient view page

onursumer:
Related to cBioPortal/cbioportal#7869

## Checks


src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -250,152 +241,6 @@ export function getOncoKbOncogenic(response: IndicatorQueryResp): string {
     }
 }
 
-export function computeGenePanelInformation(
```
onursumer, line 253: moved to a new shared utility class `GenePanelUtils.ts`

src/pages/resultsView/ResultsViewPageStoreUtils.ts
```diff
@@ -250,152 +241,6 @@ export function getOncoKbOncogenic(response: IndicatorQueryResp): string {
...
-            }
-        }
-    }
-    return {
-        samples: sampleInfo,
-        patients: patientInfo,
-    };
-}
-
-export function annotateMolecularDatum(
```
onursumer, line 371: moved to shared `StoreUtils.ts`

src/shared/lib/StoreUtils.ts
```diff
@@ -1415,3 +1425,209 @@ export function getDefaultProfilesForOql(profiles: MolecularProfile[]) {
             })
     );
 }
+
```
onursumer, line 1428: these new functions below are mostly extracted from Results View related pages

adamabeshouse (approved):
nice!

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -419,6 +448,22 @@ export class PatientViewPageStore {
         default: [],
     });
 
+    readonly samplesWithUniqueKeys = remoteData(
+        {
+            await: () => [this.samples, this.allSamplesForPatient],
+            invoke: () =>
+                Promise.resolve(
+                    this.samples.result.map(
```
alisman, line 456: i'm not sure i get this operation here.  seems like you are just replacing

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -313,6 +336,12 @@ export class PatientViewPageStore {
             ),
     });
 
+    readonly discreteMolecularProfile = remoteData({
```
alisman, line 339: couldn't there be more than one discrete molecular profile?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -877,6 +922,20 @@ export class PatientViewPageStore {
         []
     );
 
+    readonly molecularData = remoteData<NumericGeneMolecularData[]>(
```
alisman, line 925: seems odd that this is called molecularData (general) when it derives from discreteCnaData?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1696,6 +1788,173 @@ export class PatientViewPageStore {
...
 
+    readonly oncoKbDataForOncoprint = remoteData<IOncoKbData | Error>(
+        {
+            await: () => [this.mutationData, this.oncoKbAnnotatedGenes],
+            invoke: async () =>
+                fetchOncoKbDataForOncoprint(
+                    this.oncoKbAnnotatedGenes,
+                    this.mutationData
+                ),
+            onError: (err: Error) => {
```
alisman, line 1799: lets get rid of this onError

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1696,6 +1788,173 @@ export class PatientViewPageStore {
...
+                // fail silently, leave the error handling responsibility to the data consumer
+            },
+        },
+        ONCOKB_DEFAULT
+    );
+
+    readonly cnaOncoKbDataForOncoprint = remoteData<IOncoKbData | Error>(
+        {
+            await: () => [
+                this.clinicalDataForSamples,
```
alisman, line 1809: how come we await things that aren't used in invoke?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1696,6 +1788,173 @@ export class PatientViewPageStore {
...
+                hotspots: boolean;
+                cbioportalCount: boolean;
+                cosmicCount: boolean;
+                customDriverBinary: boolean;
+                customDriverTier?: string;
+            } => {
+                const getOncoKbMutationAnnotationForOncoprint = this
+                    .getOncoKbMutationAnnotationForOncoprint.result!;
+                const oncoKbDatum:
+                    | IndicatorQueryResp
```
alisman, line 1843: wow.  why is this type so wide open?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1696,6 +1788,173 @@ export class PatientViewPageStore {
...
+                    // TODO support these too?
+                    cbioportalCount: false,
+                    cosmicCount: false,
+                    customDriverBinary: false,
+                };
+            });
+        },
+    });
+
+    readonly getOncoKbMutationAnnotationForOncoprint = remoteData<
```
alisman, line 1874: it's probably too late but i don't think we should name these promises "get" prefix

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: Right, `this.samples` do not have unique keys, whereas `this.allSamplesForPatient` do. That's why we are replacing the samples here. We should probably make it more clear.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: There could be, but we already have `readonly molecularProfileIdDiscrete`. I just followed the same logic.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: This is to make the cna data compatible with existing utility functions which accept molecular data only. We should probably refactor the store to use the molecular data endpoint to fetch cna data (if possible) to avoid this type of hacky data conversion.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: good catch. this might be some leftover code.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: Cannot remember why, this is partially copied from https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/ResultsViewPageStore.ts#L4176

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: Right, we have a promise with the same name in the results view page store.

adamabeshouse: the result of the remoteData is a function, so that's where the `get` comes from. I remember we had discussed this and you didn't like it. Im not crazy about it either but what I like about it is it combines a function with its asynchronous dependencies so that a caller only has to ask whether the function is ready, without having to reference its dependencies separately. @alisman

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
alisman: seems like we should have a comment explaining it.  or maybe just called discreteCNAMolecularData?

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1428,6 +1443,85 @@ export class PatientViewPageStore {
         );
     }
 
+    @computed
+    get annotatedExtendedAlterationData(): AnnotatedExtendedAlteration[] {
```
alisman, line 1447: this logic doesn't exist in resultsView?

src/shared/lib/oql/AccessorsForOqlFilter.ts
```diff
@@ -32,6 +32,22 @@ export type SimplifiedMutationType =
     | 'inframe'
     | 'other';
 
+export function getMutationSubType(d: {
+    mutationType: string;
+    proteinChange: string;
+}) {
+    if (d.mutationType && d.mutationType.toLowerCase() === 'fusion') {
```
alisman, line 39: i feel like these exist somewhere in enums

onursumer: This code is not new, I just extracted it from `public mut_type(d: Datum)` to be able to reuse it. Most of the string references in this file is hardcoded already. If we really want to fix this specific fusion reference only, I think we can use this one: https://github.com/cBioPortal/cbioportal-frontend/blob/10928633533e2c55c9f4530cf545af5e3f11da74/packages/cbioportal-frontend-commons/src/lib/getCanonicalMutationType.ts#L16

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
onursumer: Sort of. This one is tricky. We rely on OQL filter functions (oqlfilter.ts) in the results view. It is not straightforward to completely isolate that logic. It might be possible to create a fake OQL query on the patient view page store but instead I decided to implement a simplified version of this logic for the patient view.

PR #3425, opened 2020-09-24
Treatments Bugfixes

Luke-Sikina:
Fix cBioPortal/cbioportal#7904
- Treatment chart search bar is no longer banished to behind the
  the chart below it
- Treatment charts are now prioritized more highly in default display


kalletlak (approved):

jjgao:
Thanks, @Luke-Sikina. It looks good.

For priority, it works if we "reset charts." But if I don't reset, the chart remains the same due to the remembers charts. I am wondering what would be the solution - is there a way to do a force reset?

jjgao:
@Luke-Sikina could you also take case of this btw? Thanks.

alisman (approved):

PR #3429, opened 2020-09-29
Add timeline data to patient clinical data tab

alisman:
image

Fixes https://github.com/cbioportal/cbioportal/issues/1783


adamabeshouse (approved):

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -78,6 +78,7 @@ import ResourcesTab, { RESOURCES_TAB_NAME } from './resources/ResourcesTab';
 import { MakeMobxView } from '../../shared/components/MobxView';
 import ResourceTab from '../../shared/components/resources/ResourceTab';
 import TimelineWrapper from './timeline2/TimelineWrapper';
+import ClinicalEventsTables from 'pages/patientView/timeline2/ClinicalEventsTables';
```
onursumer, line 81: Minor but I think this is better:

```suggestion
import ClinicalEventsTables from './timeline2/ClinicalEventsTables';
```

src/pages/patientView/timeline2/ClinicalEventsTables.tsx
```diff
@@ -0,0 +1,56 @@
+import React from 'react';
+import { ClinicalEvent } from 'cbioportal-ts-api-client';
+import { groupTimelineData } from 'pages/patientView/timeline2/timeline-data-utils';
```
onursumer, line 3: Not sure if we should use `-` in filenames. I would name this file `TimelineDataUtils` or something like that.

src/pages/patientView/timeline2/timeline-data-utils.ts
```diff
@@ -17,8 +17,23 @@ export function downloadZippedTracks(events: ClinicalEvent[]) {
     });
 }
 
+export function groupTimelineData(events: ClinicalEvent[]) {
+    const groupedData = _.groupBy(events, d => d.eventType);
+
+    return _.mapValues(groupedData, (data, eventType) => {
+        return getRows(data);
+    });
```
onursumer, line 25: Should we just do this instead?

```suggestion
    return _.mapValues(groupedData, data => getRows(data));
```

src/pages/patientView/timeline2/timeline-data-utils.ts
```diff
@@ -17,8 +17,23 @@ export function downloadZippedTracks(events: ClinicalEvent[]) {
...
+}
+
 function toTSV(events: ClinicalEvent[]): string {
     // First get the extra columns
```
onursumer, line 29: This comment should be inside the `getRows` method, right?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -1455,17 +1456,10 @@ export default class PatientViewPage extends React.Component<
                                     key={2}
                                     id={PatientViewPageTabs.ClinicalData}
                                     linkText="Clinical Data"
+                                    className={'patient-clinical-data-tab'}
```
onursumer, line 1459: Another minor one. We don't need the curly brackets here.
```suggestion
                                    className="patient-clinical-data-tab"
```

onursumer (commented):
Looks good. Just some minor suggestions.

src/pages/patientView/PatientViewPage.tsx
alisman: i actually gonna leave these because this is how my IDE adds them automatically.  i will look to see if it can use relative instead

src/pages/patientView/PatientViewPage.tsx
alisman: i could go either way on this.  i don't think it makes a difference

PR #3431, opened 2020-09-30
update oncokb client

zhx828:


zhx828:
The change would not have any affect.

PR #3434, opened 2020-10-01
Study view: use mutation profiled sample count for fusion/sv

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/7910

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -6007,7 +6007,7 @@ export class StudyViewPageStore {
                         )
                     );
                     // samples countaing this data would be the samples profiled for these molecular profiles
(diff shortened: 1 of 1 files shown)


zhx828 (approved):
LGTM 💯

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6007,7 +6007,7 @@ export class StudyViewPageStore {
                         )
                     );
                     // samples countaing this data would be the samples profiled for these molecular profiles
-                    ret[uniqueKey] = _.sumBy(
+                    let count = _.sumBy(
```
alisman, line 6010: logic in this if block seems odd.  i think you can put the uniqueKey logic at 6004 in the else block of count===0 since you otherwise don't use it.

alisman (approved):

PR #3435, opened 2020-10-01
Add compare button to pie chart tooltip, and make changes to select/deselect buttons

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7533

image

image

image


alisman:
Looks great.  One thing. The select all button at bottom of checkbox table now shows always (as opposed to on hover as in current production).   While you're at it, maybe bump that button leftward so it isn't right in corner
image

PR #3436, opened 2020-10-02
(GC) Fix error when group data is not available

dippindots:
Fix cBioPortal/cbioportal#7884

- (sort) when data is not available for selected group, do not filter with it's value
- (download) when data is not available for selected group, return `NA`


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -759,9 +759,13 @@ export function getEnrichmentColumns(
...
             sortBy: (d: ExpressionEnrichmentRow) =>
-                d.groupsSet[group.name].meanExpression,
+                d.groupsSet[group.name]
+                    ? d.groupsSet[group.name].meanExpression
+                    : null,
             download: (d: ExpressionEnrichmentRow) =>
-                d.groupsSet[group.name].meanExpression.toFixed(2),
+                d.groupsSet[group.name]
+                    ? d.groupsSet[group.name].meanExpression.toFixed(2)
+                    : 'NA',
```
adamabeshouse, line 768: is there an `NA` constant?

adamabeshouse (approved):
looks good, just please use an NA constant there if it exists

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
dippindots: Sure, we already has a constant that I can use, updated.

alisman (approved):

PR #3437, opened 2020-10-02
oncoprint - hide menu dots unless hovering

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/6313


jjgao (approved):
Looks good from the product perspective. Thanks!

PR #3438, opened 2020-10-05
Timeline vaf grouprows

alisman:

- a
- b

## Checks

## Any screenshots or GIFs?

## Notify reviewers


pvannierop:
@alisman There are a few small issues left:

1. The sequential view has problems with the groupBy row:
image

2. With long names the labels in the groupBy header are truncated. Can we perhaps implement wrapping easily. The samples icons should be centered in the vertical dimension.
image

src/pages/patientView/timeline2/VAFChart.tsx
```diff
@@ -0,0 +1,801 @@
...
+                    renderHeader: () => this.groupByTrackLabel(index),
+                    renderTrack: () => this.sampleIconsGroupByTrack(sampleIds),
+                    height: () => this.groupIndexToTrackHeight[index],
+                    labelForExport: this.clinicalValuesForGrouping[index],
+                });
+            });
+        this.props.wrapperStore.groupByTracks = tracks;
+    }
+
+    yAxisHeaderReaction = autorun(() => {
```
pvannierop, line 642: Please review the use of an autorun here.

alisman: Hi Pim.  So, here's my analysis of the situation.   The problem is that the track header is a computation of this.ticks, but it doesn't have access to it because it's a property of a sibling component.  (This is just an unfortunate consequence of the fact that though header is conceptually a child of track, it cannot be rendered as part of the track.)   So, these autoruns observe local state and update the wrapperStore, which is effectively a parent of both track header and this component.   But I believe the right pattern would be just to bump ticks itself into the store. That way both header and chart can observe it properly and we don't need any of this autorun shenanigans.   In general, we should always try to avoid reactions (autorun) and rely instead on one way flow of data from parent to child.

PR #3450, opened 2020-10-08
Implement comparison groups based on mutated genes table and treatments tables in study view

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7831

## Patients
image
image

## Samples
image
image

[Link](https://genie-private.cbioportal.org/study/summary?id=nsclc_genie_bpc)


kalletlak (approved):

Luke-Sikina (approved):

inodb (approved):
Awesome! Thanks!

One small thing. I saw that the compare groups option is grayed out by default when nothing is selected. For pie charts it does a comparison across all groups. Maybe it makes sense if these behave in the same way? Can be addressed in follow up PR.

Thanks for the quick implementation 🚀 !

PR #3452, opened 2020-10-12
Fix bugs in survival types table and add median survival columns

adamabeshouse:
Fixes  https://github.com/cBioPortal/cbioportal/issues/7952

[Link1](https://www.cbioportal.org/results/comparison?Action=Submit&RPPA_SCORE_THRESHOLD=2.0&Z_SCORE_THRESHOLD=2.0&cancer_study_list=brca_tcga_pan_can_atlas_2018&case_set_id=brca_tcga_pan_can_atlas_2018_cnaseq&comparison_subtab=survival&data_priority=0&gene_list=TP53&geneset_list=%20&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=brca_tcga_pan_can_atlas_2018_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=brca_tcga_pan_can_atlas_2018_mutations&profileFilter=0&tab_index=tab_visualize)
[Link2](https://genie-private.cbioportal.org/results/comparison?Action=Submit&cancer_study_list=nsclc_genie_bpc&case_set_id=nsclc_genie_bpc_all&gene_list=TP53&tab_index=tab_visualize&comparison_subtab=survival)

### DIFF
--- a/src/pages/groupComparison/Survival.tsx
+++ b/src/pages/groupComparison/Survival.tsx
@@ -363,6 +363,7 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
             const patientSurvivals = this.props.store.patientSurvivals.result!;
             const analysisGroups = this.analysisGroupsComputations.result!
                 .analysisGroups;
+            const uidToAnalysisGroup = _.keyBy(analysisGroups, g => g.value);
             const patientToAnalysisGroups = this.analysisGroupsComputations
                 .result!.patientToAnalysisGroups;
             const pValues = this.pValuesByPrefix.result!;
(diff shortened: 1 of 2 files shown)


src/pages/groupComparison/Survival.tsx
```diff
@@ -389,19 +390,27 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
                                 }
                                 return {
                                     prefix,
                                     displayText,
-                                    numPatients:
-                                        patientSurvivals[prefix].length,
+                                    numPatients: _.sumBy(
+                                        patientSurvivals[prefix],
+                                        s =>
+                                            +(
```
alisman, line 409: what does that plus operator do?

alisman: seems weird adding a boolean?

adamabeshouse: + converts true to 1 and false to 0

adamabeshouse:
@cBioPortal/product could you please take a look at this one?

src/pages/groupComparison/Survival.tsx
alisman: I feel like we should avoid this kind of wizardry/coercion in favor of + (s ? 1 : 0)

jjgao:
The table overlaps with the plot when there are many columns. I am not sure how to address this though...

image

alisman:
@jjgao i think the only thing to do is to make the table horizontally scrollable.   that is if we don't want to push the chart rightward.  which i don't think we want to do.   that raises our favorite issue of whether scrollbar can be on bottom (runs risk that user isn't aware of hidden columns)

alisman (approved):

PR #3455, opened 2020-10-13
Independent display logic for sample treatments chart

Luke-Sikina:
Backend changes here: https://github.com/cBioPortal/cbioportal/pull/7960

Fix cBioPortal/cbioportal#7928

- Sample treatments sometimes should not be shown when patient treatments
are shown
- This PR uses a new API endpoint to determine when sample treatments should be shown
- Also, unknown treatments are no longer shows on the frontend

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -3832,7 +3832,7 @@ export class StudyViewPageStore {
             _chartMetaSet
         );
 
-        if (this.displayTreatments.result) {
+        if (this.displaySampleTreatments.result) {
             _chartMetaSet['SAMPLE_TREATMENTS'] = {
                 uniqueKey: 'SAMPLE_TREATMENTS',
                 dataType: ChartMetaDataTypeEnum.CLINICAL,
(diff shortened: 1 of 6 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6924,6 +6926,17 @@ export class StudyViewPageStore {
         });
     }
 
+    @computed
```
alisman, line 6929: why do we put these in computed (instead of treating them like patientTreatments below as an instance property)?   @Luke-Sikina you might just have copied an existing thing, but i'm curious.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6924,6 +6926,17 @@ export class StudyViewPageStore {
...
+    public get displaySampleTreatments() {
+        return remoteData({
+            invoke: () => {
+                return defaultClient.getContainsSampleTreatmentDataUsingPOST({
+                    studyViewFilter: this.initialFilters,
```
kalletlak, line 6934: Question, isn't studyIds not enough to check they have treatments data or not? do we need studyViewFilter?

src/pages/studyView/StudyViewPageStore.ts
Luke-Sikina: I'm not going to resolve this comment just in case it's still relevant, but I think these do need to be computed now, as `this.studyIds` is an observable.

alisman: yes but they would be computed either way.  i'm just curious why it is we treat these different than we normally treate remoteData, at least in other stores.    probably there isn't an answer.  
normal way (note that this will have same computed behavior):
```
readonly localPro = remotedata(...)
```

alisman (approved):

PR #3459, opened 2020-10-14
create react-variant-view package

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/7946

react-variant-view package using for signal.
image

https://www.npmjs.com/package/react-variant-view
This is a testing version, it has a lot of things could be cleaned up or updated


leexgh:
Netlify shows error:
```
(plugin rpt2) Error: /opt/build/repo/packages/react-variant-view/src/util/AnnotationSummaryUtil.ts(9,9): semantic error TS2739: Type '{ aminoAcidAlt: string; aminoAcidRef: string; aminoAcids: string; codonChange: string; consequenceTerms: string; entrezGeneId: string; exon: string; hgvsc: string; hgvsp: string; hgvspShort: string; ... 4 more ...; variantClassification: string; }' is missing the following properties from type 'TranscriptConsequenceSummary': 'polyphenPrediction', 'polyphenScore', 'siftPrediction', 'siftScore'
```
but local is fine

leexgh:
This is a testing version, it has a lot of things could be cleaned up or updated. We can update it later

packages/react-variant-view/package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "react-variant-view",
+  "version": "0.0.0",
```
onursumer, line 3: Let's set this one to `0.1.0-beta.0`

packages/react-variant-view/src/util/ApiClientUtils.ts
```diff
@@ -0,0 +1,17 @@
+import { GenomeNexusAPI } from 'genome-nexus-ts-api-client';
+import { OncoKbAPI } from 'oncokb-ts-api-client';
+
+// TODO customize domain?
+export const genomeNexusDomain = 'https://www.genomenexus.org/';
+const genomeNexusInternalClient = new GenomeNexusAPI(genomeNexusDomain);
```
onursumer, line 6: Right, most likely we would like to customize this. How do we do the genome nexus client customization for `react-mutation-mapper`? Is it possible to do it the same way?

packages/react-variant-view/src/util/VariantUtil.ts
```diff
@@ -0,0 +1,44 @@
+import {
+    getProteinPositionFromProteinChange,
+    Mutation,
+} from 'cbioportal-utils';
+import { VariantAnnotationSummary } from 'genome-nexus-ts-api-client';
+import { getTranscriptConsequenceSummary } from './AnnotationSummaryUtil';
+
+export function variantToMutation(
```
onursumer, line 8: We may want to move some of these utility functions into `cbioportal-utils`, but I guess it is okay to leave them here at this point.

packages/react-variant-view/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "test:watch": "yarn run test --watch"
+  },
+  "peerDependencies": {
+    "mobx": "^3.0.0 || ^4.0.0 || ^5.0.0",
+    "mobx-react": "^4.0.0 || ^5.0.0",
+    "prop-types": "^15.5.4",
+    "react": "^15.0.0 || ^16.0.0",
+    "react-dom": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
```
onursumer, line 40: We need to make sure that these dependencies are complete. This package may still work inside the monorepo even if some dependencies are not listed here. But when we include the package in another project we may get missing dependency errors. For now we can test it with signal to see if it works fine (after publishing the beta version to npm).

onursumer (commented):
Looks good, let's make sure that it works in Signal

leexgh:
Update ideas:
1. Do we like the styles? (not located in the center without mutation mapper)
2. Click gene to go to gene page

packages/react-variant-view/src/util/AnnotationSummaryUtil.ts
```diff
@@ -0,0 +1,34 @@
+import {
+    TranscriptConsequenceSummary,
+    VariantAnnotationSummary,
+} from 'genome-nexus-ts-api-client';
+
+export function getTranscriptConsequenceSummary(
+    data: VariantAnnotationSummary | undefined
+): TranscriptConsequenceSummary {
+    let transcriptConsequenceSummary: TranscriptConsequenceSummary = {
```
onursumer, line 9:
```suggestion
    let transcriptConsequenceSummary = {
```

packages/react-variant-view/src/util/AnnotationSummaryUtil.ts
```diff
@@ -0,0 +1,34 @@
...
+        },
+        refSeq: '',
+        transcriptId: '',
+        variantClassification: '',
+    };
+
+    if (data !== undefined && data.transcriptConsequenceSummary) {
+        transcriptConsequenceSummary = data.transcriptConsequenceSummary;
+    }
+    return transcriptConsequenceSummary;
```
onursumer, line 33: As a workaround for the TS error, we can do this for now:

```suggestion
    return transcriptConsequenceSummary as TranscriptConsequenceSummary;
```

packages/react-variant-view/package.json
```diff
@@ -0,0 +1,70 @@
...
+    "react-if": "^2.1.0",
+    "react-motion": "^0.5.2",
+    "react-mutation-mapper": "^0.6.10",
+    "react-rangeslider": "^2.2.0",
+    "react-select": "^3.0.4",
+    "react-spinkit": "^3.0.0",
+    "react-table": "^6.10.0",
+    "seamless-immutable": "^7.1.4",
+    "superagent": "^3.8.3",
+    "typescript": "3.8.3"
```
onursumer, line 64: let's upgrade this to `4.0.3` for consistency. I have another PR (#3472) to deal with the rest of the packages.

packages/react-variant-view/package.json
```diff
@@ -0,0 +1,70 @@
...
+  "peerDependencies": {
+    "mobx": "^3.0.0 || ^4.0.0 || ^5.0.0",
+    "mobx-react": "^4.0.0 || ^5.0.0",
+    "prop-types": "^15.5.4",
+    "react": "^15.0.0 || ^16.0.0",
+    "react-dom": "^15.0.0 || ^16.0.0"
+  },
+  "dependencies": {
+    "autobind-decorator": "^2.1.0",
+    "bootstrap": "^4.3.1",
```
onursumer, line 42: Not sure if it is a good idea to have direct dependency to bootstrap 4, since the main package still uses bootstrap 3. Maybe add this as <strong>peer dependency</strong> and let the consumer decide which version to use, something like:
```suggestion
    "bootstrap": "^3.0.0 || ^4.0.0",
```

onursumer (changes requested):

packages/react-variant-view/package.json
leexgh: good point

leexgh:
@onursumer Thank you! I've updated the version and no more complain about it

onursumer (approved):

PR #3460, opened 2020-10-15
plots - gracefully handle log scale with rna_seq profiles that have negative numbers

adamabeshouse:
Fixes issue reported by @jjgao:
"Do you why the log scale checkbox is not checked but y-axis is log2? https://bit.ly/3k4iYTR"

This issue was due to the fact that some rna_seq profiles now have small negative values as a result of batch-correction. The solution is that we will not disable the log-scale checkbox for rna_seq profiles even if they have negative values. In addition, log scale rendering for mrna profiles now sets negative values to 0, and then renders log2(value+1)

image


jjgao (approved):
Looks good from product.

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -3045,9 +3046,23 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
     }
 
     isDisabledAxisLogCheckbox(vertical: boolean): boolean {
-        return vertical
+        const axisSelection = vertical
+            ? this.vertSelection
+            : this.horzSelection;
+
+        // dont disable the log box for rna_seq even if there are
```
alisman, line 3053: not sure if i understand this comment.  this method seems to always disable checkbox if there are negative numbers detected (hasNegativeNumbers)

adamabeshouse: no its `!isValidMrnaProfile && hasNegativeNumbers` so if `isValidMrnaProfile` is true then it will not disable the checkbox

alisman (approved):

PR #3461, opened 2020-10-16
study view - add timeout before rerendering study view charts

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7977


src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -376,7 +377,14 @@ export class StudySummaryTab extends React.Component<
         //      issue will be solved.
         return (
             <div key={chartMeta.uniqueKey}>
-                <ChartContainer key={chartMeta.uniqueKey} {...(props as any)} />
+                <DelayedRender>
```
alisman, line 380: would be an interesting experiment:  if you basically debounce render so that in a space of say x ms, it only actually fires the last render.    i would be curious to see if we are in fact getting a lot of these unnecessary render cycles and would benefit from this kind of throttling

alisman (approved):

PR #3465, opened 2020-10-19
(Oncoprint) generic assay color range from light to dark

dippindots:
Fix cBioPortal/cbioportal#7959

- a generic assay track data color range from light to dark (except `Treatment Response`)

before:
image

after:
image


adamabeshouse (approved):

pvannierop (commented):
First, can you provide some rationale on why this is needed? Second, I remember long discussions on the colors used for treatment data (predecessor of generic assay data). If I recall well this related to another feature of generic assay data, the 'pivot threshold'. When provided, the pivot threshold represents an externally defined value that sets a color boundary. For treatment data the color would be dark red-light red-white-light blue-dark blue. So white color would represent the pivot threshold. I light of this I would recommend to expand your PR to handle this situation because there is more to it then simply changing the blue color gradient.

dippindots:
Hi @pvannierop , thanks for the comment. The reason of this is because we noticed that when the `pivot threshold` is not specified or is not applicable, the color range is from `dark` to `light` which seems not a normal expression for `low` to `high`. And this is only affect the profile which don't specify `pivot threshold`. But I do noticed that for `Treatment IC50` has been affected by this change, just curious that for this case (in the following screenshot), the range from `0 - 100` should use color from `light` to `dark`?
image

pvannierop:
@alisman I see this is merged. However, I think there is still an ongoing discussion [here](https://github.com/cBioPortal/cbioportal/issues/8004). Can this be reverted?

pvannierop:
@alisman I see this has been merged. This issue is still under discussion [here](https://github.com/cBioPortal/cbioportal/issues/8004).

PR #3466, opened 2020-10-19
Upgrade typescript to latest version, and fix new errors

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -273,8 +273,10 @@ export enum SampleListCategoryType {
 }
 
 export enum GeneticEntityType {
-    'GENE' = 'gene',
-    'GENESET' = 'geneset',
+    'GENE' = 'GENE',
```
adamabeshouse, line 276: @dippindots you worked on geneset stuff right? could you take a look at these changes I made relating to GeneticEntityType and make sure I haven't done anything weird?

dippindots: @adamabeshouse This looks good to me! It's in sync with the backend.

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -611,7 +611,7 @@ export function getAlterationEnrichmentColumns(
         enrichedGroupColum.tooltip = (
             <table>
                 <tr>
-                    <td>Log ratio > 0</td>
+                    <td>Log ratio {'>'} 0</td>
```
adamabeshouse, line 614: officially, JSX does not allow >, <, {, } as plaintext characters, and the new version of typescript catches that

src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
```diff
@@ -707,8 +710,10 @@ export function annotateGeneticTrackData(
             const newObj = _.clone(object);
             newObj.data = newObj.data.filter(d => {
                 // clear previous annotations
-                delete d.oncoKbOncogenic;
-                delete d.putativeDriver;
+                delete (d as Partial<OncoprinterGeneticTrackDatum_Data>)
```
adamabeshouse, line 713: the new typescript has a check which does not allow you to use `delete` on a member which is not potentially undefined in the type

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -170,6 +170,13 @@ export default class ComparisonStore {
     protected async saveAndGoToSession(newSession: Session) {
         throw new Error(`saveAndGoToSession must be implemented in subclass`);
     }
+    abstract _session: MobxPromise<Session>;
+    abstract _originalGroups: MobxPromise<ComparisonGroup[]>;
+    abstract overlapStrategy: OverlapStrategy;
+    abstract usePatientLevelEnrichments: boolean;
```
adamabeshouse, line 176: the new typescript catches if the superclass declares an instance property and the subclass overrides it using a `get` accessor, so after some difficulty I realized that the solution is to declare these as abstract properties

src/shared/components/GeneSelectionBox/GeneSymbolValidatorMessage.tsx
```diff
@@ -246,10 +246,7 @@ export default class GeneSymbolValidatorMessage extends React.Component<
 
         return (
             <div id="geneBoxValidationStatus">
-                <GeneSymbolValidatorMessageChild
-                    replaceGene={this.props.replaceGene}
-                    {...this.props}
-                />
+                <GeneSymbolValidatorMessageChild {...this.props} />
```
alisman, line 249: what happened to replaceGene?

src/shared/lib/comparison/ComparisonStore.ts
alisman: abstract forces the subclass to define it?

alisman (approved):

src/shared/components/GeneSelectionBox/GeneSymbolValidatorMessage.tsx
adamabeshouse: it's already in `this.props`

src/shared/lib/comparison/ComparisonStore.ts
adamabeshouse: I believe so, yes. and it allows the property to be implemented using an instance property or a getter/setter

PR #3474, opened 2020-10-26
Improve GSVA volcano plot selector layout

pvannierop:
Small fixes to GSVA volcano-plot selector component include:
- Hiding y scrollbar when not needed
- Making the modal a tad wider
- Using flex for more consistent layout 

# Before
image

# After
image

### DIFF
--- a/src/shared/components/oncoprint/DeltaUtils.ts
+++ b/src/shared/components/oncoprint/DeltaUtils.ts
@@ -684,8 +684,8 @@ function transitionTracks(
 
     // find the max and min GSVA score value in the next heatmap track group
     // max and min value is used to create a custom legend for the track group
-    const genesetScoreValueMax = _.max(genesetScoreProfileData) || undefined;
-    const genesetScoreValueMin = _.min(genesetScoreProfileData) || undefined;
+    const genesetScoreValueMax = _.max(genesetScoreProfileData);
+    const genesetScoreValueMin = _.min(genesetScoreProfileData);
 
     // Transition genetic tracks
     const prevGeneticTracks = _.keyBy(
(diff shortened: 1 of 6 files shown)


src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -672,6 +682,11 @@ function transitionTracks(
         }
     );
 
+    // find the max and min GSVA score value in the next heatmap track group
+    // max and min value is used to create a custom legend for the track group
+    const genesetScoreValueMax = _.max(genesetScoreProfileData) || undefined;
+    const genesetScoreValueMin = _.min(genesetScoreProfileData) || undefined;
```
alisman, line 688: not sure that this does.    if it's 0 we use undefined?

pvannierop: No, _max_ and _min_ can return _null_. The interface for _maxProfileValue_ and _minProfileValue_ explains that these props can be a number or undefinedL

```
    maxProfileValue?: number;
    minProfileValue?: number;
```
The ` _.max(genesetScoreProfileData) || undefined` formulation in effect translates _null_ to _undefined_.

But I see that this translation also translated _0_ to _undefined_. This is not the intention. I have reverted the code you mentioned and added the check when the value is actually used:

```
        track.maxProfileValue = genesetScoreValueMax === null? undefined : genesetScoreValueMax;
        track.minProfileValue = genesetScoreValueMin === null? undefined : genesetScoreValueMin;
```

alisman (approved):

alisman (approved):

alisman:
@pvannierop you can merge this whenever tests are passing.  if you need me to merge, let me know

jjgao:
It seems like this is ready to be merged. Adding to this sprint.

PR #3480, opened 2020-10-29
Skip annotating canonical transcript only for cBioPortal mutations

onursumer:
Fix cBioPortal/cbioportal#8019. MutationMapper is used in multiple projects (cBioPortal, Genome Nexus & SIGNAL). In cBioPortal the mutations are pre-annotated for the canonical transcripts, which is why we only annotate mutations for non-canonical transcripts. In other projects where this component is used we do want to annotate the canonical transcript using Genome Nexus. This introduces a property `skipAnnotationForCanonicalTranscript` to allow developers that use this package to choose to enable/disable annotation of the canonical transcript.

## Checks

### DIFF
--- a/packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
+++ b/packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
@@ -113,7 +113,8 @@ export function getMutationByTranscriptId(
     mutation: Mutation,
     ensemblTranscriptId: string,
     indexedVariantAnnotations: { [genomicLocation: string]: VariantAnnotation },
-    isCanonicalTranscript: boolean
+    isCanonicalTranscript: boolean = false,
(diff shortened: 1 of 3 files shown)


leexgh (approved):
@onursumer Thanks!

alisman:
not sure i understand how this identifies the "cbioportal mutations"

onursumer:
@alisman Sorry for the confusing PR title. We don't actually identify the cbioportal mutations. We just skip annotation for the cbioportal instance by setting the corresponding parameter to true.

Previously the annotator code was accidentally updated to always skip annotation for canonical transcript in the library level. But, this is something we only want for the cbioportal instance. This PR fixes the library function so that the skipping annotation is optional. It doesn't skip annotation for canonical transcript by default anymore. Instead we only skip it for the cbioportal instance.

PR #3482, opened 2020-10-30
Send physical study ids instead of virtual ids from treatments api

kalletlak:
https://github.com/cBioPortal/cbioportal/issues/8020


inodb (approved):
🚀

alisman (approved):

PR #3485, opened 2020-11-02
patient view - sample manager always has data on all samples for patient

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8013


onursumer (approved):
LGTM

inodb (approved):
Thanks! LGTM!

Just a side note: it is not super clear one is looking at a single sample view, so that might be slightly confusing. However I think that was already an issue before this change

dippindots (approved):
LGTM!

PR #3489, opened 2020-11-03
react-variant-view update with existing signal data

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/7989
image


leexgh:
I copied over some files from signal, those files have a comment on the top: `// this component comes from signal`. We could skip those files when review. Later we can refactor those copied files somewhere

packages/react-variant-view/src/component/cancerPatientPopulation/CancerPatientPrevalence.tsx
```diff
@@ -0,0 +1,195 @@
...
+                <div key={'cancer-patient-prevalence-no-data'}>
+                    No data available
+                </div>
+            );
+        }
+        return content;
+    }
+
+    // duplicate of FrequencyCell.tsx mainContent()
+    private fixDisplayValue(value: number | null) {
```
onursumer, line 141: We should move this to a utility/helper file.

packages/react-variant-view/src/component/frequencyTable/FrequencyCell.tsx
```diff
@@ -0,0 +1,58 @@
+// this component comes from signal
```
onursumer, line 1: We should either move this to `cbioportal-frontend-commons` or the depending implementation into signal.

packages/react-variant-view/src/component/frequencyTable/MutationTumorTypeFrequencyTable.tsx
```diff
@@ -0,0 +1,117 @@
+// this component comes from signal
```
onursumer, line 1: Similar to `FrequencyCell`, we should either move this to `cbioportal-frontend-commons` or the depending implementation into signal.

packages/react-variant-view/src/component/generalPopulationPrevalence/GeneralPopulationPrevalence.tsx
```diff
@@ -19,7 +31,27 @@ interface IVcf {
 }
 
 @observer
-class PopulationPrevalence extends React.Component<IPopulationPrevalenceProps> {
+class GenralPopulationPrevalence extends React.Component<
+    IGenralPopulationPrevalenceProps
+> {
+    @computed get indexHotspot() {
```
onursumer, line 37: we should probably name this `indexedHotspots` or `hotspotIndex`.

packages/react-variant-view/src/component/pathogenicity/ClinVarInterpretation.tsx
```diff
@@ -0,0 +1,105 @@
...
+            const rcvCountMap = _.mapValues(rcvMap, rcvs =>
+                _.countBy(rcvs, rcv => rcv.clinicalSignificance)
+            );
+            return rcvCountMap;
+        } else {
+            return undefined;
+        }
+    }
+
+    @computed get clinVarContent() {
```
onursumer, line 42: This function is a bit difficult to follow. Seems like here we are doing both data transformation and rendering at the same time. It might be better to have a separate method to transform the data into the structure that we need, and a simpler render function consuming the transformed data.

packages/react-variant-view/src/util/FilterUtils.ts
```diff
@@ -0,0 +1,199 @@
...
+export const CANCER_TYPE_FILTER_ID = '_signalCancerTypeFilter_';
+export const MUTATION_STATUS_FILTER_ID = '_signalMutationStatusFilter_';
+export const PROTEIN_IMPACT_TYPE_FILTER_ID = '_signalProteinImpactTypeFilter_';
+export const PENETRANCE_FILTER_ID = '_signalPenetranceFilter_';
+export const HUGO_SYMBOL_FILTER_TYPE = 'signalHugoSymbol';
+export const MUTATION_STATUS_FILTER_TYPE = 'signalMutationStatus';
+export const MUTATION_COUNT_FILTER_TYPE = 'signalMutationCount';
+export const PENETRANCE_FILTER_TYPE = 'signalPenetrance';
+
+export enum MutationStatusFilterValue {
```
onursumer, line 26: At this point we only need `MutationStatusFilterValue` from this file, right? Maybe add a new model into `cbioportal-utils` named `SignalMutationStatus` and use that instead? We can update signal to use the new model, too.

packages/react-variant-view/src/util/MutationDataUtils.ts
```diff
@@ -0,0 +1,140 @@
+// functions in this component come from signal
```
onursumer, line 1: We should probably rename this to `SignalMutationUtils` and move into `cbioportal-utils`

packages/react-variant-view/src/model/GeneFrequencySummary.ts
```diff
@@ -0,0 +1,24 @@
+// this component comes from signal
```
onursumer, line 1: move to `model/signal` under `cbioportal-utils`?

packages/react-variant-view/src/model/Mutation.ts
```diff
@@ -0,0 +1,37 @@
+// this component comes from signal
```
onursumer, line 1: We should move this to `model/signal` under `cbioportal-utils`.

Also maybe directly use `SignalMutation` coming from Genome Nexus instead of `IMutation`, and rename `IExtendedMutation` to `IExtendedSignalMutation`.

onursumer (changes requested):
I think we should try to refactor/relocate duplicates before merging this PR

packages/react-variant-view/src/component/frequencyTable/MutationTumorTypeFrequencyTable.tsx
leexgh: @onursumer This table is not done yet, we'll add more columns when new data is available. So I guess maybe leave it in the `react-variant-view` package would be easier to customize?

packages/react-variant-view/src/component/cancerPatientPopulation/CancerPatientPrevalence.tsx
leexgh: I move this one to `cbioportal-frontend-commons` too, because the `FrequencyCell.tsx` there also needs this function

leexgh: Moved to `cbioportal-utils` in the end, this is a util function so might be more proper there

packages/react-variant-view/src/component/frequencyTable/FrequencyCell.tsx
leexgh: this is moved to `cbioportal-frontend-commons`

packages/react-variant-view/src/component/cancerPatientPopulation/CancerPatientPrevalence.tsx
onursumer: Makes sense. We should probably find a more descriptive name for the function. Its main purpose is to convert values like `0.004` to `< 0.01`. I don't know which name describes this conversion best though.

Also, we should add a second parameter `fractionDigits: number = 1` to make the function more generic.

packages/cbioportal-utils/src/model/SignalMutationStatus.ts
```diff
@@ -0,0 +1,7 @@
+export enum MutationStatusFilterValue {
```
onursumer, line 1: should we also rename this to `enum SignalMutationStatus`?

packages/cbioportal-utils/src/signal/DisplayUtils.ts
```diff
@@ -0,0 +1,11 @@
+export function fixDisplayValue(value: number | null) {
```
onursumer, line 1: Actually, looks like we already have a very similar function here: https://github.com/cBioPortal/cbioportal-frontend/blob/97d211b37627ddc9f46d7044d2f4f653db49576c/packages/cbioportal-utils/src/format/FormatUtils.ts#L1

We should probably reuse that one. In any case this function is not really signal specific, so we should probably move the implementation into `FormatUtils`.

packages/cbioportal-utils/src/signal/GeneFrequencySummary.ts
```diff
@@ -1,5 +1,3 @@
-// this component comes from signal
```
onursumer: I think this belongs to model. Let's move this under model and rename it `SignalGeneFrequencySummary`? Also prefix the model names with Signal to make it clear, like `ISignalGeneFrequencySummary`, `ISignalTumorTypeFrequencySummary`, etc?

onursumer (changes requested):
Thanks for moving the duplicate code under appropriate packages! Just added minor suggestions regarding naming and location of the functions/models.

packages/cbioportal-utils/src/signal/DisplayUtils.ts
leexgh: Good point! The one we have has one more checking for null value, and if it's null we show `-`. So I move the function to `FormatUtils` and resue the `formatPercentValue`

packages/cbioportal-utils/src/signal/GeneFrequencySummary.ts
leexgh: @onursumer Do you think it makes sense to rename all signal related model with a `signal` prefix? For example https://github.com/cBioPortal/cbioportal-frontend/pull/3489/files#diff-a4120f05d3b6d91af87296f694be727c063beab1990c7c317cd66da0662ba43bR1

packages/react-variant-view/src/model/Mutation.ts
leexgh: Good point, `ICountByTumorType` could be replaced by `CountByTumorType` from Genome Nexus also.

onursumer: Right, we can use the existing model from genome nexus directly. I guess this is also true for `SignalMutation`. We can always go over what we have and clean up duplicate models later though.

packages/cbioportal-utils/src/format/FormatUtils.ts
```diff
@@ -14,3 +14,9 @@ export function formatPercentValue(rate: number, fractionDigits: number = 1) {
 export function numberOfLeadingDecimalZeros(value: number) {
     return -Math.floor(Math.log10(value) / Math.log10(10) + 1);
 }
+
+export function displayValue(value: number | null, fractionDigits: number = 1) {
```
onursumer, line 18: I think `displayValue` is too generic. How about `formatFrequencyValue`?

packages/cbioportal-utils/src/index.tsx
```diff
@@ -19,6 +19,8 @@ export * from './model/MyCancerGenome';
 export * from './model/MyVariantInfo';
 export * from './model/OncoKB';
 export * from './model/RemoteData';
+export * from './model/signal/SignalMutation';
+export * from './model/signal/SignalMutationStatus';
```
onursumer, line 23: I think it is okay to put these directly under model:

```suggestion
export * from './model/SignalMutation';
export * from './model/SignalMutationStatus';
```

packages/cbioportal-utils/src/index.tsx
```diff
@@ -29,3 +31,6 @@ export * from './myCancerGenome/MyCancerGenomeUtils';
 export * from './oncokb/OncoKbUtils';
 
 export * from './promise/PromiseUtils';
+
+export * from './model/signal/SignalGeneFrequencySummary';
```
onursumer, line 35: move this export statement next to the other models?

packages/react-variant-view/src/component/frequencyTable/FrequencyTable.css
```diff
@@ -0,0 +1,18 @@
+/* this component comes from signal */
```
onursumer, line 1: Should we remove this comment? After merging this PR, we are going to remove the one in Signal repo and import this one, right?

onursumer (changes requested):
Almost there! Just a few minor requests.

onursumer (approved):
Looks great, thanks!

inodb (approved):

PR #3490, opened 2020-11-05
Improve study view loading indicator

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8012

### DIFF
--- a/src/pages/studyView/tabs/SummaryTab.tsx
+++ b/src/pages/studyView/tabs/SummaryTab.tsx
@@ -428,6 +428,9 @@ export class StudySummaryTab extends React.Component<
                     this.store.initialVisibleAttributesClinicalDataCountData,
                 ],
             },
+            {
+                label: 'Rendering',
+            },
         ];
     }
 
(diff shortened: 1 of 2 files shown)


src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -519,51 +522,46 @@ export class StudySummaryTab extends React.Component<
...
-                                        }
-                                        onResizeStop={this.onResize}
-                                    >
-                                        {this.store.visibleAttributes.map(
-                                            this.renderAttributeChart
-                                        )}
-                                    </ReactGridLayout>
-                                )}
-                            </div>
+                {!this.store.loadingInitialDataForSummaryTab && (
```
adamabeshouse, line 525: the only change here is moving the `this.props.store.selectedSamples.result.length > 0` condition into `loadingInitialDataForSummaryTab`

alisman: cool.  this definately seems to solve the problem.    is there a good reason you think we don't just test status of selectedSamples?   i.e. selectedSamples.isPending  ?

alisman (approved):

PR #3494, opened 2020-11-09
Fixed display of column values in downloaded tsv

jtquach1:
Fix cBioPortal/cbioportal#7911

- Added download function for Cytoband
- Removed columns and their headers that don't have a download function defined

## Before
image
[table.tsv.txt](https://github.com/cBioPortal/cbioportal-frontend/files/5513249/table.tsv.txt)

## After
image
[table (15).tsv.txt](https://github.com/cBioPortal/cbioportal-frontend/files/5513250/table.15.tsv.txt)

### DIFF
--- a/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
+++ b/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
@@ -242,6 +242,8 @@ export default class CopyNumberTableWrapper extends React.Component<
             ),
             sortBy: (d: DiscreteCopyNumberData[]) =>
                 this.getCytobandForGene(d[0].gene.hugoGeneSymbol),
+            download: (d: DiscreteCopyNumberData[]) =>
+                this.getCytobandForGene(d[0].gene.hugoGeneSymbol),
             visible: true,
             order: 60,
         });
--- a/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
+++ b/src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
@@ -7,6 +7,7 @@ import {
     lazyMobXTableSort,
     default as LazyMobXTable,
     Column,
+    LazyMobXTableStore,
 } from './LazyMobXTable';
 import SimpleTable from '../simpleTable/SimpleTable';
 import { DefaultTooltip } from 'cbioportal-frontend-commons';
@@ -396,6 +397,25 @@ describe('LazyMobXTable', () => {
 
     after(() => {
         clock.uninstall();
+    });
+
+    describe('downloadData', () => {
+        it('only displays columns with defined download functions', () => {
+            assert.deepEqual(
+                new LazyMobXTableStore({
+                    columns: columns,
+                    data: data,
+                }).downloadData,
+                [
+                    ['Name', 'Number', 'String', 'Initially invisible column'],
+                    ['0', '0', 'asdfj', '0HELLO123456'],
+                    ['1', '6', 'kdfjpo', '1HELLO123456'],
+                    ['2', 'null', 'null', '2HELLO123456'],
+                    ['3', '-1', 'zijxcpo', '3HELLO123456'],
+                    ['4', '90', 'zkzxc', '4HELLO123456'],
+                ]
+            );
+        });
     });
 
     describe('utils', () => {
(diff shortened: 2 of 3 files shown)


alisman (approved):

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -378,7 +378,29 @@ export class LazyMobXTableStore<T> {
                 }
             );
         });
-        return tableDownloadData;
+
```
Luke-Sikina, line 381: I don't like that we process all the columns and _then_ filter out columns without the download function. Futhermore, you're just filtering out empty columns, which could theoretically be legit data. 

Can you spend some time seeing if you can instead filter out columns before we create tableDownloadData?

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -242,6 +242,8 @@ export default class CopyNumberTableWrapper extends React.Component<
             ),
             sortBy: (d: DiscreteCopyNumberData[]) =>
                 this.getCytobandForGene(d[0].gene.hugoGeneSymbol),
+            download: (d: DiscreteCopyNumberData[]) =>
```
Luke-Sikina, line 245: Seems good to me

Luke-Sikina (changes requested):

jtquach1:
Current output (same):
image

[table (3).tsv.txt](https://github.com/cBioPortal/cbioportal-frontend/files/5527391/table.3.tsv.txt)

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
jtquach1: That's true- I just realized it's inefficient to filter at the very end (and a specified download function could return the empty string). I just saw that the header and an empty string for the download object rows are displayed for columns without specified download functions.

Luke-Sikina (changes requested):

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -398,6 +398,12 @@ describe('LazyMobXTable', () => {
         clock.uninstall();
     });
 
+    describe('downloadData', () => {
```
alisman, line 401: probably you're working on this

jtquach1: that's correct, I'm in the process of mocking up a props.columns and props.data to pass into a LazyMobXTable constructor call

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -364,7 +357,7 @@ export class LazyMobXTableStore<T> {
             //rowDownloadData is list of lists, containing all the elements per column.
             //processedRowsDownloadData becomes the transposed of rowDownloadData.
             let processedRowsDownloadData = rowDownloadData[0].map(function(
-                row: string,
+                _row: string,
```
jtquach1, line 360: VS Code suggested prefixing unused variables with an underscore

Luke-Sikina: Let's not do that. I guess that's vscode's default naming scheme for unused variables, but that clashes with how we use underscore prefixed variables in our stores. Just leave it as `row`.

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
Luke-Sikina: This looks good, nice job. Given the amount of boilerplate we'll have to add if we do multiple tests, I think this is good enough. Thank you for adding the test.

src/shared/components/lazyMobXTable/LazyMobXTable.spec.tsx
```diff
@@ -398,6 +399,25 @@ describe('LazyMobXTable', () => {
...
+        it.only('only displays columns with defined download functions', () => {
```
Luke-Sikina, line 403: Oop, you left an `only`
```suggestion
        it('only displays columns with defined download functions', () => {
```

Luke-Sikina (approved):
Tiny changes. This looks really good, nice job.

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -787,7 +780,7 @@ export default class LazyMobXTable<T> extends React.Component<
                 // populate the cache instances with all available data for the lazy loaded columns
                 this.store.downloadDataFetcher
                     .fetchAndCacheAllLazyData()
-                    .then(allLazyData => {
+                    .then(_allLazyData => {
```
Luke-Sikina, line 783: couldn't you just replace this one with `()`?

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -872,7 +865,7 @@ export default class LazyMobXTable<T> extends React.Component<
         );
         this.pageToHighlightReaction = reaction(
             () => this.store.firstHighlightedRowIndex,
-            (index: number) => {
+            (_index: number) => {
```
Luke-Sikina, line 868: Same, doesn't `()` work?

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
jtquach1: yes, it looks like passing in a no-arg lambda works (no type mismatch)

Luke-Sikina (approved):

PR #3501, opened 2020-11-13
Updates to fusion-structural variant migration based on feedback

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8048

### DIFF
--- a/packages/react-mutation-mapper/src/store/DefaultMutationMapperDataFetcher.ts
+++ b/packages/react-mutation-mapper/src/store/DefaultMutationMapperDataFetcher.ts
@@ -11,6 +11,7 @@ import {
     IOncoKbData,
     Mutation,
     uniqueGenomicLocations,
+    StructuralVariantType,
 } from 'cbioportal-utils';
 import {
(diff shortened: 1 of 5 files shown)


packages/react-mutation-mapper/src/store/DefaultMutationMapperDataFetcher.ts
```diff
@@ -314,7 +314,8 @@ export class DefaultMutationMapperDataFetcher
         const structuralQueryVariants: AnnotateStructuralVariantQuery[] = _.uniqBy(
             _.map(
                 queryVariants.filter(
-                    mutation => mutation.mutationType === 'Fusion'
+                    mutation =>
+                        mutation.mutationType?.toUpperCase() === 'FUSION'
```
alisman, line 318: sorta shocking if we don't have an enum for this somewhere?

kalletlak: Found one, updated

alisman (approved):

PR #3504, opened 2020-11-16
Update some e2e tests broken by data changes

alisman:


end-to-end-test/shared/specUtils.js
```diff
@@ -327,13 +327,32 @@ function checkElementWithTemporaryClass(
 }
 
 function checkElementWithMouseDisabled(selector, pauseTime, options) {
-    return checkElementWithTemporaryClass(
+    browser.execute(function() {
+        $("<div id='blockUI'></div>")
```
adamabeshouse, line 331: width: 100%, height:100% is implicit?

also blockUI seems not bizarre enough to never collide with anything. unlikely but not impossible enough IMO. how about using Math.random() as the id instead

alisman: sure

alisman: if you put something positon absolute and then do left:0, top:0, right:0, left:0, it stretches to fill portal.

PR #3505, opened 2020-11-16
Fix default page size for RMM data table

onursumer:
Related to cBioPortal/cbioportal#8062

## Checks

### DIFF
--- a/packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
+++ b/packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
@@ -163,17 +163,6 @@ export default class DataTable<T> extends React.Component<
     }
 
     @computed
-    get defaultPageSize() {
-        const initialItemsPerPage = this.props.initialItemsPerPage;
-
-        return this.tableData
-            ? this.tableData.length > initialItemsPerPage!
-                ? initialItemsPerPage
-                : this.tableData.length
-            : 1;
-    }
-
-    @computed
     get initialColumnDataStatus() {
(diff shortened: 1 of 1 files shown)


packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
```diff
@@ -162,17 +162,6 @@ export default class DataTable<T> extends React.Component<
...
-    @computed
-    get defaultPageSize() {
-        const initialItemsPerPage = this.props.initialItemsPerPage;
-
-        return this.tableData
-            ? this.tableData.length > initialItemsPerPage!
-                ? initialItemsPerPage
-                : this.tableData.length
-            : 1;
-    }
```
inodb, line 174: @onursumer how come we can get rid of this logic now?

inodb (approved):
LGTM - minor comment: could be nice to add some clarification as to why the old logic is no longer necessary

packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
onursumer: @inodb Apparently this wasn't necessary in the first place. I thought this was required to get rid of the empty padding rows, but actually setting `minRows` to a small number is good enough. See the `minRows` property here: https://github.com/tannerlinsley/react-table/tree/v6#props.

Setting `defaultPageSize` to a computed value, especially to zero, messes things for the pagination.

leexgh (approved):

onursumer:
Failing e2e tests are not related, they are also failing on master. Merging.

PR #3511, opened 2020-11-19
Fix localdb tests

alisman:


end-to-end-test/local/specs/gsva.screenshot.spec.js
```diff
@@ -73,6 +73,8 @@ describe('gsva feature', () => {
             vertDataSelect.$('.Select-arrow-zone').click();
             vertDataSelect.$('.Select-option=Gene Sets').click();
 
+            browser.debug();
```
adamabeshouse, line 76: do you mean to commit this?

PR #3515, opened 2020-11-24
Add cohort and mrna expr download data to patient view cna table

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/6915

### DIFF
--- a/src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
+++ b/src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
@@ -173,6 +173,24 @@ export default class CohortColumnFormatter {
         }
     }
 
+    public static getDownloadValue(
+        data: DiscreteCopyNumberData[],
+        copyNumberCountCache: CopyNumberCountCache
+    ) {
+        const copyNumberCount = CohortColumnFormatter.getCopyNumberCount(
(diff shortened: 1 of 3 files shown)


src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
```diff
@@ -173,6 +173,24 @@ export default class CohortColumnFormatter {
...
     }
 
+    public static getDownloadValue(
+        data: DiscreteCopyNumberData[],
+        copyNumberCountCache: CopyNumberCountCache
+    ) {
+        const copyNumberCount = CohortColumnFormatter.getCopyNumberCount(
+            data,
+            copyNumberCountCache
+        )!;
```
alisman, line 183: seems like would be better to allow for undefined in the type rather than use ! since you check for falsy below.

onursumer (approved):

Luke-Sikina (approved):

src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.tsx
inodb: @adamabeshouse do you want to update this? Or merge as is?

PR #3520, opened 2020-11-25
Upgrade mobx

adamabeshouse:
[Official Mobx Migration Guide](https://mobx.js.org/migrating-from-4-or-5.html)

# cBioPortal mobx migration guide:

- you have to add a `makeObservable(this)` call in the constructor of classes that use any mobx decorators, like `@observable`, `@computed`, `@action`. That's because these decorators no longer do anything on their own, they only mark class members so that `makeObservable` knows how to handle them. The `@observer` decorator on react components does not require calling `makeObservable`

- For the same reason (that mobx decorators aren't really real decorators anymore), `@action` no longer plays nice with the `@autobind` decorator. Instead you must use `@action.bound`, i.e. `@autobind @action myAction() {` becomes `@action.bound myAction() {`

- You can no longer override decorated methods with decorators, this will cause a world of debugging pain for you. For example, this will break:
```
class MySuperClass {
    constructor() {
        makeObservable(this);
    }
    @computed get myComputed() {
    }
}

class MySubClass extends MySuperClass {
    constructor() {
        makeObservable(this);
    }
    @computed get myComputed() {
        // Hope you have several hours today to debug the confusing bugs this causes!!
    }
}
```
The workaround we've been using is as follows:
```
class MySuperClass {
    constructor() {
        makeObservable(this);
    }
    protected getMyComputed() {
    }
    @computed get myComputed() {
        return this.getMyComputed();
    }
}

class MySubClass extends MySuperClass {
    constructor() {
        makeObservable(this);
    }
    protected getMyComputed() {
        // Subclass-specific implementation
    }
}
```


alisman:
@adamabeshouse 
there is one legit broken test.  it's the number in the Genomic Profile Sample Counts. weird.
https://89990-66571349-gh.circle-artifacts.com/0/imageCompare.html

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6081,16 +6081,10 @@ export class StudyViewPageStore {
...
-            if (this.isInitialFilterState) {
-                molecularProfileOptions = this
-                    .initialMolecularProfileSampleCounts.result!;
-            } else {
-                molecularProfileOptions = getMolecularProfileOptions(
-                    this.molecularProfiles.result,
-                    this.sampleUniqueKeysByMolecularProfileIdSet.result!
-                );
-            }
+            const molecularProfileOptions: GenomicDataCountWithSampleUniqueKeys[] = getMolecularProfileOptions(
```
alisman, line 6084: @kalletlak is this ok to lose this condition?

kalletlak: This should be ok

PR #3521, opened 2020-11-25
generic assay continuous data study view implementation

dippindots:
Fix cBioPortal/cbioportal#7657

- **New add chart tabs**
image

image

image

- **Increased width of Add chart tooltip (50px)**
previous: 
image

after: 
image

--------------------------**Testing**--------------------------
Multiple studies (contains 2 `GENERIC ASSAY TYPE` data: `Microbiome Signature` and `Treatment Response`)
https://deploy-preview-3521--cbioportalfrontend.netlify.app/study/summary?id=brca_tcga_pan_can_atlas_2018%2Clgg_tcga_pan_can_atlas_2018%2Cccle_broad_2019

tests can be performed:
1. Click `Add Chart`, and click the arrow, then select `Microbiome Signature` or `Treatment Response` to add a new chart
2. After add a chart, select a range in the new added chart see if selected samples updated correctly
3. download chart data or go to comparison page from the selected chart
4. refresh the page and see if added charts still in place


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -174,6 +180,15 @@ export type RectangleBounds = {
     yStart?: number;
 };
 
+export type MolecularProfileOption = {
+    value: string;
+    count: number;
+    label: string;
+    description: string;
+    sampleUniqueKeys: string[];
+    dataType?: string;
```
dippindots, line 189: dataType is optional, this is to support `Generic Assay`. `Generic Assay` has three dataType: `CONTINUOUS`, `CATEGORY` and `BINARY`, different dataType will have different chart type. For example, `CONTINUOUS` should create a bar chart while `CATEGORY` and `BINARY` should create a pie chart.

src/pages/studyView/addChartButton/geneLevelSelection/GeneLevelSelection.tsx
```diff
@@ -17,16 +17,10 @@ import { MakeMobxView } from 'shared/components/MobxView';
...
+import { MolecularProfileOption } from 'pages/studyView/StudyViewUtils';
 
 export interface IGeneLevelSelectionProps {
-    molecularProfileOptionsPromise: MobxPromise<
-        {
-            value: string;
-            label: string;
-            count: number;
-            description: string;
-        }[]
```
dippindots, line 28: created a type in studyViewUtils: `MolecularProfileOption`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1384,6 +1428,22 @@ export class StudyViewPageStore {
             });
         }
 
+        if (
+            _.isArray(filters.genericAssayDataFilters) &&
```
kalletlak, line 1432: what is this for?
`_.isEmpty` checks for both undefined and empty array

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3595,6 +3784,97 @@ export class StudyViewPageStore {
...
+                        )
+                        .value()
+                );
+            },
+            default: {},
+        },
+        {}
+    );
+
+    readonly genericAssayEntityOptionsByType = remoteData(
```
kalletlak, line 3846: looks similar to `genericAssayProfileOptionsByType` (line 3814)

kalletlak (approved):
LGTM

src/pages/studyView/StudyViewPageStore.ts
dippindots: Oh thanks, this is an outdated function, it shouldn't be here, deleted.

src/pages/studyView/StudyViewUtils.tsx
alisman: maybe we should set this dataType to enum to enforce this?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -793,6 +815,22 @@ export function getVirtualStudyDescription(
...
             });
 
+            _.each(
+                filter.genericAssayDataFilters || [],
+                genericAssayDataFilters => {
+                    const uniqueKey = getGenericAssayChartUniqueKey(
+                        genericAssayDataFilters.stableId,
+                        genericAssayDataFilters.profileType
+                    );
+                    const name = attributeNamesSet[uniqueKey];
```
alisman, line 825: how come you can be sure that uniqueKey exists in attributNamesSet?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2672,13 +2724,25 @@ export function getMolecularProfileOptions(
...
                 .value();
-            return {
-                value: value,
-                count: uniqueProfiledSamples.length,
-                label: profiles[0].name,
-                description: profiles[0].description,
-                sampleUniqueKeys: uniqueProfiledSamples,
-            };
+
+            if (withDataType) {
```
alisman, line 2728: kinda repetative here.  is there a reason that we cannot always return dataType even if it doesn't exist and is undefined?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2766,6 +2830,86 @@ export async function getGenomicDataAsClinicalData(
     });
 }
 
+export function convertGenericAssayDataBinsToClinicalDataBins(
```
alisman, line 2833: why do we have to convert to "clinical data" bins?   would it not be better to generalize bins?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2766,6 +2830,86 @@ export async function getGenomicDataAsClinicalData(
...
+            gaDataBin.profileType
+        ),
+        count: gaDataBin.count,
+        end: gaDataBin.end,
+        specialValue: gaDataBin.specialValue,
+        start: gaDataBin.start,
+    }));
+}
+
+export async function getGenericAssayDataAsClinicalData(
```
alisman, line 2848: again, just wondering why we do this "conversion"

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2798,3 +2942,21 @@ export function getMolecularProfileSamplesSet(
         {}
     );
 }
+
+export function getNonZeroUniqueBins(dataBins: ClinicalDataBin[]) {
```
alisman, line 2946: this is a little opaque.  how does this filter for non zero?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2798,3 +2942,21 @@ export function getMolecularProfileSamplesSet(
...
+    return _.uniq(
+        _.reduce(
+            dataBins,
+            (acc, next) => {
+                if (next.start) {
```
alisman, line 2951: if you are relying on 0 being falsy, we should just test for zero directly

src/pages/studyView/UserSelections.tsx
```diff
@@ -318,6 +323,54 @@ export default class UserSelections extends React.Component<
...
+        _.reduce(
+            this.props.filter.genericAssayDataFilters || [],
+            (acc, genericAssayDataFilter) => {
+                const uniqueKey = getGenericAssayChartUniqueKey(
+                    genericAssayDataFilter.stableId,
+                    genericAssayDataFilter.profileType
+                );
+                const chartMeta = this.props.attributesMetaSet[uniqueKey];
+                if (chartMeta) {
+                    acc.push(
```
alisman, line 336: lets assign to a variable and then push it at the end to avoid nesting.  tough to read

src/pages/studyView/StudyViewUtils.tsx
dippindots: we can do that, I can update this.

src/pages/studyView/StudyViewUtils.tsx
dippindots: @kalletlak I think there may have some discussions when you implemented the `gene specific charts`. I am wondering whether you know the reason why we decided to convert other data bins to `clinical data bins`?

alisman: @kalletlak if perhaps the first implementation of this was for clinical data and now we are using that for other forms of data, would be a lot of work to make the implementation generic?

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -162,6 +171,36 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+            chartMeta => chartMeta.genericAssayType
+        );
+
+        return _.mapValues(groupedChartMetaByGenericAssayType, chartMeta => {
+            let genericAssayDataOptions = getOptionsByChartMetaDataType(
+                chartMeta,
+                this.selectedAttrs,
+                this.props.store.chartsType.toJS()
+            );
+            if (
```
alisman, line 191: it's not a good idea to hide a check like this inside of a computed like this.    it hides the business logic.  also, the name of this computed doesn't seems to fully describe it.  it seems it is actually returning chart types, right?  if that's true, chart types should be in the name.   then we can have another layer of computeds that name the tabs and which do the filtering.   e.g.   chartTypeOptionsForClinicalDataTab.

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -307,7 +360,7 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
         return (
             <div
                 style={{
-                    width: '400px',
+                    width: '420px',
```
alisman, line 363: you can just use integer 420.  react will supply the units automatically.

src/pages/studyView/StudyViewUtils.tsx
kalletlak: Initially it was just for clinical data. And then I added for gene specific charts. Since it is evolving it make sense to implement it in a generic way.

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -325,6 +381,7 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
                         className="addClinicalChartTab"
                     >
                         <AddChartByType
+                            width={420}
```
alisman, line 384: lets make a constant for this 420 since it gets repeated

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -428,6 +488,101 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+                                                .genericAssayEntitiesGroupByGenericAssayType
+                                                .result![type],
+                                            entity =>
+                                                makeGenericAssayOption(
+                                                    entity,
+                                                    false
+                                                )
+                                        )}
+                                        onSubmit={(
+                                            charts: GenericAssayChart[]
```
alisman, line 518: lets take this callback (onSubmit) and make it a bound method

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -428,6 +488,101 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+                                    />
+                                    {this
+                                        .genericAssayDataOptionsByGenericAssayType[
+                                        type
+                                    ] &&
+                                        this
+                                            .genericAssayDataOptionsByGenericAssayType[
+                                            type
+                                        ].length > 0 && (
+                                            <div style={{ marginTop: '10px' }}>
```
alisman, line 561: integer 10

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -428,6 +488,101 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
                             </div>
                         )}
                     </MSKTab>
+                    {!this.hideGenericAssayTabs &&
+                        _.map(
+                            this.props.store.genericAssayProfileOptionsByType
+                                .result,
+                            (options, type) => (
+                                <MSKTab
```
alisman, line 496: this tab spec is too deeply nested.   lets make a wrapper component for it and extract some of the logic inside of attributes to local variables.

src/pages/studyView/StudyViewUtils.tsx
kalletlak: We need clinical data for download and comparison feature. These methods/functions only support `ClincalData` objects. This method is for converting genomic data to clinical data and use those methods directly. I think it can also be implemented in a generic way

src/pages/studyView/StudyViewUtils.tsx
dippindots: Got it, thanks @kalletlak, will update this to a generic solution.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2299,38 +2441,18 @@ export class StudyViewPageStore {
...
-                        if (next.start) {
-                            acc.push(next.start);
-                        }
-                        if (next.end) {
-                            acc.push(next.end);
-                        }
-                        return acc;
-                    },
-                    [] as number[]
-                )
```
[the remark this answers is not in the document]
dippindots: @kalletlak I have another question about this part, we are using `if (next.start)` and `if (next.end)` at here. This will exclude the cases that `next.start === 0` and `next.end === 0`, is that what we want? Or we just need `start` and `end` are not `undefined`?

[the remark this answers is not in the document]
kalletlak: I think this logic came from Hongxin originally, I just reused it. Looking at the function we need all non-zero numbers which is required for custom data bins, which looks correct to me.

src/pages/studyView/addChartButton/genericAssaySelection/GenericAssaySelection.tsx
```diff
@@ -0,0 +1,290 @@
...
+    value: string;
+    label: string;
+}
+
+@observer
+export default class GenericAssaySelection extends React.Component<
+    IGenericAssaySelectionProps,
+    {}
+> {
+    @observable private _selectedProfileOption?: MolecularProfileOption & {
```
alisman, line 29: you should always assign something to an observable.  even if    = undefined

src/pages/studyView/addChartButton/genericAssaySelection/GenericAssaySelection.tsx
```diff
@@ -0,0 +1,290 @@
...
+    @action
+    private handleProfileSelect(option: any) {
+        if (option && option.value) {
+            this._selectedProfileOption = option;
+        }
+    }
+
+    @computed
+    private get selectedProfileOption() {
+        if (this._selectedProfileOption !== undefined) {
```
alisman, line 67: should really be if else else.   i think it's better to get in habit of writing out conditionals.

src/pages/studyView/addChartButton/genericAssaySelection/GenericAssaySelection.tsx
```diff
@@ -0,0 +1,290 @@
...
+        return filteredSelectedGenericAssayEntityIds.map(
+            o => this.genericAssayEntitiesOptionsByValueMap[o]
+        );
+    }
+
+    @computed get selectedGenericAssaysJS() {
+        return toJS(this.selectedGenericAssayEntities);
+    }
+
+    private isOptionIncludingText(text: string, option: ISelectOption) {
```
alisman, line 147: doesOptionMatchSearchText

src/pages/studyView/StudyViewUtils.tsx
dippindots: Updated to use a generic type `DataBin`

src/pages/studyView/StudyViewUtils.tsx
dippindots: Updated to use a generic type `DataBin`

src/pages/studyView/StudyViewUtils.tsx
dippindots: @alisman backend didn't expose the `dataType`, maybe we should leave as it? Or we may need to create one and update along with backend?

src/pages/studyView/UserSelections.tsx
dippindots: Not sure if this looks better?

src/pages/studyView/addChartButton/AddChartButton.tsx
dippindots: It's actually returns the `chartOptions for generic assay`, so rename to `genericAssayChartOptionsByGenericAssayType`

src/pages/studyView/addChartButton/AddChartButton.tsx
dippindots: Made a `@computed` component and separated logics into variables: https://github.com/cBioPortal/cbioportal-frontend/pull/3521/files#diff-2c76ee08342f5cf09e54594c2c9ea1b268689a000376f4b1e1e76d2f9fb048f5R232-R271

alisman (approved):

PR #3522, opened 2020-12-01
Fix filtering error on custom chart containing unselected samples

kalletlak:
https://github.com/cBioPortal/cbioportal/issues/8077

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -2432,7 +2432,7 @@ export class StudyViewPageStore {
             }) as any
         );
 
-        if (_sampleIdentifiers && _sampleIdentifiers.length > 0) {
+        if (!_.isEmpty(sampleIdentifiersFilterSets)) {
             filters.sampleIdentifiers = _sampleIdentifiers;
         } else {
             if (_.isEmpty(this.queriedSampleIdentifiers.result)) {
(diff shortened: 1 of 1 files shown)


zhx828 (approved):
LGTM

alisman:
@kalletlak seems kinda odd that we need to duplicate this  same check in so many places.   what are the conditions where it's true?  (empty studies and samples).    if they are a precondition for the promises returning any data at all, then that would suggest that perhaps the condition should be in a view layer so that the promises would not be referenced under that condition?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2798,3 +2798,13 @@ export function getMolecularProfileSamplesSet(
         {}
     );
 }
+
+export function conditionalInvocation<T>(
```
alisman, line 2802: not really what i meant.   this is just a replacement for "if" which does not improve on it.  also, you have a mystery return type [T] which isn't suggested by the name.  
what i mean is  instead of this: 
```this.studyViewFilterWithFilteredSampleIdentifiers.result!.sampleIdentifiers.length > 0```
this:
```this.hasSamplesInCurrentFilter ```
or some name which spells out the condition.

Comment (alisman) [reply src/pages/studyView/StudyViewUtils.tsx:2802]: the name should suggest why we don't have to fetch anything

File: src/pages/studyView/StudyViewPageStore.ts [hunk 2]
```diff
@@ -4925,6 +4954,12 @@ export class StudyViewPageStore {
         invoke: () => {
             //fetch samples when there are only filters applied
             if (this.chartsAreFiltered) {
+                if (
+                    _.isEmpty(this.filters.studyIds) &&
```
alisman, line 4958: should this use it?

alisman (approved):

PR #3523, opened 2020-12-03
Switch Study Download to Download from S3

averyniceday:
- Dataset download will now pull from S3
- Corresponding change has been made in datahub repo. Pushes to master will automatically push zipped and updates study into S3 bucket

## Checks

## Any screenshots or GIFs?

## Notify reviewers

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -324,7 +324,7 @@ export function getDarwinUrl(sampleIds: string[], caseId: string) {
 
 export function getStudyDownloadListUrl() {
     return buildCBioPortalAPIUrl(
-        'proxy/download.cbioportal.org/study_list.json'
+        'proxy/cbioportal-datahub.s3.amazonaws.com/study_list.json'
     );
 }
 
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -323,9 +323,7 @@ export function getDarwinUrl(sampleIds: string[], caseId: string) {
 }
 
 export function getStudyDownloadListUrl() {
-    return buildCBioPortalAPIUrl(
-        'proxy/cbioportal-datahub.s3.amazonaws.com/study_list.json'
-    );
+    return 'https://cbioportal-datahub.s3.amazonaws.com/study_list.json';
 }
 
 export function getMDAndersonHeatmapPatientUrl(patientId: string) {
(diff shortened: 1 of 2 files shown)


inodb:
@averyniceday Thanks so much for fixing this! Maybe it's a good idea to move `study_list.json` over as well? See: http://download.cbioportal.org/study_list.json

Not sure if this is the  best mechanism to determine what studies are in the S3 bucket, but I believe that's how it currently works. Alternatively I guess we can assume all studies in the public portal are in the S3 bucket.

See also code here:

https://github.com/cBioPortal/cbioportal-frontend/blob/ed842150ffc3a8740bf50ecaa1f99d30580b12a5/src/shared/api/urls.ts#L325-L329

src/shared/api/urls.ts
```diff
@@ -324,7 +324,7 @@ export function getDarwinUrl(sampleIds: string[], caseId: string) {
 
 export function getStudyDownloadListUrl() {
     return buildCBioPortalAPIUrl(
-        'proxy/download.cbioportal.org/study_list.json'
+        'proxy/cbioportal-datahub.s3.amazonaws.com/study_list.json'
```
inodb, line 327: don't need the proxy anymore since S3 is on https, so can directly use `https://cbioportal-datahub.s3.amazonaws.com/study_list.json`. Before we were using the proxy to be able to pull from http on a https domain

sheridancbio (approved):
I think this looks good. We are updating the study_list.json every time there is an update to datahub/master now (via github actions). And we do upload a .zip file for every touched public study (and have backfilled all zip files so that there is a currently complete set).

averyniceday:
@inodb is this good to merge? Saw you updated the code with your suggestion.

inodb:
@averyniceday looks like CORS is breaking the datasets page (it can't get the study JSON). I added the CORS to the S3 bucket: https://s3.console.aws.amazon.com/s3/buckets/cbioportal-datahub?region=us-east-1&tab=permissions. But it still seems to fail atm (could be an update issue tho)

averyniceday:
@inodb I can see the downloads on the deploy preview site. Does that mean it's working?

sheridancbio:
> I added the CORS to the S3 bucket But it still seems to fail atm (could be a temp issue tho, i just updated CORS)
> See: https://deploy-preview-3523--cbioportalfrontend.netlify.app/datasets

Hi @inodb .. I tried the link above and download a couple of studies without problem. Do you want to try again?
for example, I got this 100MB file pretty quickly : https://cbioportal-datahub.s3.amazonaws.com/stad_tcga_pub.tar.gz

inodb:
Thanks for checking @sheridancbio and @averyniceday - it does seem to be working now! 🎉 

For me the datasets was showing a blank page before because of the CORS header missing, but looks like those changes have propagated now. Thanks again for fixing this! Super useful!

PR #3525, opened 2020-12-04
Compress problematically large request bodies

Luke-Sikina:
Fix cBioPortal/cbioportal#8026

Problem:
- Some request bodies were so large (20+ Mb) that they resulted in
significantly degraded performance for people with poor upload speeds

Solution:
- Intercept requests for problematic endpoints
- gZip their request bodies using pako
- Add a Content-Encoding: gzip header
- Create a feature toggle for this


src/config/IAppConfig.ts
```diff
@@ -130,4 +130,5 @@ export interface IServerConfig {
     saml_logout_local: boolean;
     patient_view_use_legacy_timeline: boolean;
     installation_map_url: string;
+    enable_request_body_compression: boolean;
```
alisman, line 133: gzipping?  just to be specific

src/config/config.ts
```diff
@@ -190,6 +191,76 @@ export function initializeAPIClients() {
...
+ * @param apiClient
+ * @param urlsToCompress
+ * @param domain
+ */
+function compressRequestBodies(
+    apiClient: any,
+    urlsToCompress: string[],
+    domain: string
+): void {
+    if (!AppConfig.serverConfig.enable_request_body_compression) {
```
alisman, line 222: would rather see this check outside of the function

src/config/config.ts
```diff
@@ -190,6 +191,76 @@ export function initializeAPIClients() {
     cachePostMethods(GenomeNexusAPI, [], /POST$/);
     cachePostMethods(GenomeNexusAPIInternal, [], /POST$/);
     cachePostMethods(OncoKbAPI);
+
+    compressRequestBodies(
```
alisman, line 195: i think we should say applyGZIPToFetchMothods.  just be super specific about what we are doing in the name of the method.   also, we should comment on the fact that this required server to be specially configured

Luke-Sikina: Sure - anything particular you want me to say about the server being specially configured? As long as it's on the right release and has `enable_request_body_gzip_compression` set to true, it will work.

alisman (approved):

PR #3530, opened 2020-12-09
Attempt to fix ErrorScreen for missing session service

alisman:
There something very strange about session service error message.  It only fails for frontend.cbioportal.org or embedded frontend.  Not any other  netlify preview, local dev version etc. even when we alter the configuration frontend.url property so that loading takes place exactly as in production.  So it seems we need to merge something to master to more carefully debug it.

### DIFF
--- a/src/shared/components/errorScreen/ErrorScreen.tsx
+++ b/src/shared/components/errorScreen/ErrorScreen.tsx
@@ -21,13 +21,20 @@ export default class ErrorScreen extends React.Component<
     {}
 > {
     @autobind
-    copyToClipRef(copyToClip: HTMLButtonElement | null) {
-        if (copyToClip) {
-            new Clipboard(copyToClip, {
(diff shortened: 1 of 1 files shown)


src/shared/components/errorScreen/ErrorScreen.tsx
```diff
@@ -21,13 +21,20 @@ export default class ErrorScreen extends React.Component<
...
-        if (copyToClip) {
-            new Clipboard(copyToClip, {
+    copyToClipRef(copyToClipButton: HTMLButtonElement) {
+        this.copyToClipButton = copyToClipButton;
+    }
+
+    copyToClipButton: HTMLButtonElement | undefined = undefined;
+
+    componentDidMount() {
+        console.log('copytoclip', this.copyToClipButton);
```
alisman, line 31: i need to leave this here for debugging purposes

PR #3534, opened 2020-12-11
Render splice mutations separately

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7857

image

image

[Link](https://beta.cbioportal.org/results/oncoprint?session_id=5bd3ccdf498eb8b3d56887fb&plots_horz_selection=%7B%22dataType%22%3A%22MUTATION_EXTENDED%22%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D)

I don't think the color choice is ideal, let's discuss


jjgao:
Thanks, @adamabeshouse. I think the color is a bit too close to inframe. Maybe we can try (dark) navy blue for splice driver and a light version for splice vus?

adamabeshouse:
@jjgao wouldnt that totally blend in with CNA del?

adamabeshouse:
here are all the current oncoprint colors
oncoprint colors

jjgao:
maybe we can try #000080 for splice driver? It's close to black. 

image

Not sure what would be a good one for splice vus though.. maybe #8093b0? it's close to truncting vus.
image

adamabeshouse:
@jjgao another idea: would splice be considered very close conceptually to truncating? maybe we can use truncating colors and add a line or something

jjgao:
@adamabeshouse yes, it's close to truncating. I think it might be better to choose a color that is close to truncating instead of creating more styles.

adamabeshouse:
@jjgao I understand your point but I also think it's confusing to add more colors that are extremely close to each other.. I'm not sure which is more confusing/useful.

schultzn:
How about shades of orange? I think the below would work -- and sorry if you have already tried that

image

schultzn:
Fixed the legend
image

adamabeshouse:
thanks @schultzn but isn't it very close to inframe?

adamabeshouse:
@schultzn I think those colors are different enough, but visually they are a bit closer to inframe (brown) than to splice (black), right? if its not a problem then I think those colors would work

schultzn:
I think that is fine - as long as they are distinct enough.

adamabeshouse:
@schultzn @jjgao @alisman I made the change using the colors that @schultzn suggested. What do you think?

alisman (approved):

PR #3536, opened 2020-12-16
Timeline line-chart track recognition is broken when value=0 exists in track data

alisman:
We treat as a line-chart any track with where all RESULT attributes have numerical values.  This detection was breaking when any value was equal 0.

### DIFF
--- a/src/pages/patientView/timeline2/timeline_helpers.spec.tsx
+++ b/src/pages/patientView/timeline2/timeline_helpers.spec.tsx
@@ -1,0 +1,53 @@
+import { allResultValuesAreNumerical } from 'pages/patientView/timeline2/timeline_helpers';
+import { assert } from 'chai';
+import { TimelineEvent } from 'cbioportal-clinical-timeline';
+
+describe('#allResultValuesAreNumerical', () => {
+    var events = [] as TimelineEvent[];
+
+    beforeEach(() => {
+        events = [
+            {
+                event: {
+                    attributes: [{ key: 'RESULT', value: '0.6' }],
+                },
+            },
+            {
+                event: {
+                    attributes: [{ key: 'RESULT', value: '10' }],
+                },
+            },
+        ] as TimelineEvent[];
+    });
+
+    it('catches all numerical condition', () => {
+        assert.isTrue(allResultValuesAreNumerical(events));
+    });
+
+    it('catches non numerical', () => {
+        events.push({
+            event: {
+                attributes: [{ key: 'RESULT', value: 'S' }],
+            },
+        } as TimelineEvent);
+        assert.isFalse(allResultValuesAreNumerical(events));
+    });
+
+    it('handles zero as numeric', () => {
+        events.push({
+            event: {
+                attributes: [{ key: 'RESULT', value: '0' }],
+            },
+        } as TimelineEvent);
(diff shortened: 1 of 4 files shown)


src/pages/patientView/timeline2/timeline_helpers.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+            },
+        ] as TimelineEvent[];
+    });
+
+    it('catches all numerical condition', () => {
+        assert.isTrue(allResultValuesAreNumerical(events));
+    });
+
+    it('catches non numerical', () => {
+        events.push({
```
adamabeshouse, line 28: this makes test run order brittle. how about `const events2 = [...events, { blah }];`

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+            event: {
+                attributes: [{ key: 'RESULT', value: 'S' }],
+            },
+        } as TimelineEvent);
+        assert.isFalse(allResultValuesAreNumerical(events));
+    });
+
+    it('handles zero as numeric', () => {
+        events.push({
```
adamabeshouse, line 37: same here

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+            event: {
+                attributes: [{ key: 'RESULT', value: '0' }],
+            },
+        } as TimelineEvent);
+        assert.isTrue(allResultValuesAreNumerical(events));
+    });
+
+    it('catches missing RESULT attribute', () => {
+        events.push({
```
adamabeshouse, line 46: and here

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
```diff
@@ -0,0 +1,53 @@
...
+        assert.isFalse(allResultValuesAreNumerical(events));
+    });
+
+    it('handles zero as numeric', () => {
+        events.push({
+            event: {
+                attributes: [{ key: 'RESULT', value: '0' }],
+            },
+        } as TimelineEvent);
+        assert.isTrue(allResultValuesAreNumerical(events));
```
adamabeshouse, line 42: actually I'm not even sure how this passed..

adamabeshouse (approved):
nitpick with testing but otherwise lgtm!

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
alisman: no, events are refreshed in beforeEach

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
alisman: see above, events is restored to initial state prior to each test

src/pages/patientView/timeline2/timeline_helpers.spec.tsx
adamabeshouse: oh oops

PR #3538, opened 2020-12-17
Fix formatting/calculation of percentages; Threshold at 1%

alisman:
Fix https://github.com/cBioPortal/cbioportal/issues/8102


package.json
```diff
@@ -50,6 +50,7 @@
     "testMain": "export NODE_OPTIONS=--max-old-space-size=2048; karma start karma.conf.js",
     "testModules": "lerna run test --ignore=cbioportal-frontend --stream",
     "test:watch": "yarn run test -- --watch",
+    "testMain:watch": "yarn run testMain -- --watch",
```
alisman, line 53: @onursumer is there an existing way to run only the main tests?

src/shared/lib/FormatUtils.tsx
```diff
@@ -61,8 +61,15 @@ export function toFixedWithThreshold(value: number, digits: number): string {
     return fixed;
 }
 
-export function getPercentage(proportion: number, digits: number = 1) {
-    return `${toFixedWithThreshold(100 * proportion, digits)}%`;
+export function getPercentage(proportion: number) {
```
alisman, line 64: @jjgao confirmed we can show <1%, which means we don't have to sorry about decimals anymore.

package.json
onursumer: @alisman you mean to watch only the main tests, right? Seems like we don't. We need to add `testMain:watch`.

Ideally main package should be under some directory like `webapps/cbioportal`. We could just do `yarn workspace cbioportal test:watch` in that case. For the current setup though we still need workarounds like `testMain` and `testMain: watch` because we use root `package.json` for both managing the main package and the monorepo setup.

adamabeshouse (approved):

inodb (approved):
LGTM

Fine to merge after fixing other unit tests

PR #3540, opened 2020-12-18
Make sure we use es5 version of pako to ensure compatibility with ie11

alisman:
Will fix ie tests (and application in IE11

fixes https://github.com/cbioportal/cbioportal/issues/8223


alisman:
@Luke-Sikina this was my fault.  i downplayed the failure of ie11 test, when in fact it was actually broken!  looks like pako uses es6 by default.  this PR ensures we load compatible version

Luke-Sikina (approved):

PR #3545, opened 2020-12-22
Fix bug when creating comparison session from study view

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8103

Gets rid of warning when trying to make a group comparison session with more than 20 groups. The reason for this is that the code cannot at that time determine how many groups there will be. 

Discussion: Should we show that notification once the user arrives at the group comparison page?
@jjgao @schultzn @alisman

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -1078,14 +1078,6 @@ export class StudyViewPageStore {
                     const groups: SessionGroupData[] = _.chain(
                         clinicalAttributeValues
                     )
-                        .sortBy(attrVal => -attrVal.count)
-                        // do not slice for comparison on cancer studies chart
-                        .slice(
-                            0,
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1095,6 +1087,15 @@ export class StudyViewPageStore {
                                 sampleIdentifiers.length > 0
                             );
                         })
+                        .sortBy(attrVal => -attrVal.count)
```
alisman, line 1090: so essence of bug was that we were slicing before sort and therefor throwing out desired elements?

alisman (approved):

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: exactly

PR #3547, opened 2020-12-22
add signal column to mutation table

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/8217

- [x] Move `defaultSortMethod` to `cbioportal-utils`
- [x] Move `MutationTumorTypeFrequencyTable` to `cbioportal-frontend-commons`
- [x] Add `SIGNAL` column into mutations table: `0.001`
    - add signal column to `mutation mapper`
    - only germline mutations have data in Signal column. 
    - show `total` row at the bottom

image


onursumer:
Nice work, thanks!

I think for consistency we should also add this column into `DefaultMutationTable` to make it available by default. It doesn't need to be in this PR though. We can have a separate PR for that.

jjgao:
Nice, thanks, @leexgh!

- [ ] change the tooltip to the SIGNAL column header to "Prevalence of germline mutations in cancer patients from [SIGNAL](https://www.signaldb.org)"
- [ ] Let's show the prevalence only, e.g. `0.2%` instead of `0.2% prevalence (85.7% ballelic)` to simplify the interface
- [ ] For germline prevalence, 0.1 is a big number, so we should add more precision. Let's use the same number format as gnomAD column.
- [ ] In tooltip, add a summary sentence above the table (similar to COSMIC): "29 of 17152 (0.17%) cancer patients in [SIGNAL - link to the signal variant page] carried this germline mutation (BRCA1 Q1756Pfs*74). 24 of the 29 carriers (82.8%) had a somatic biallelic inactivating alteration to the same gene (BRCA1) in their tumors."
- [ ] For the table in cBioPortal, can we only keep the columns up to "% Biallellic", ie. remove/hide all columns starting `N Prevalence`?
- [ ] (good to apply to signaldb.org too) Can we render the rows with prevalence > 0 differently, e.g. bold? It may also be good to bold the header.
- [ ] (good to apply to signaldb.org too) Can we add tooltip to explain each header?
- [ ] (good to apply to signaldb.org too) maybe rename "# Samples" to "# Patients" and "# Variants" to "# Carriers" - please confirm with Chai.

leexgh:
@jjgao Thanks for review, there is still something I wanted to confirm with you:

> * [ ]  Let's show the prevalence only, e.g. `0.2%` instead of `0.2% prevalence (85.7% ballelic)` to simplify the interface
How about the download? Do we still want to have `0.2% prevalence (85.7% biallelic)` in download file?


> * [ ]  In tooltip, add a summary sentence above the table (similar to COSMIC): "29 of 17152 (0.17%) cancer patients in [SIGNAL - link to the signal variant page] carried this germline mutation (BRCA1 Q1756Pfs*74). 24 of the 29 carriers (82.8%) had a somatic biallelic inactivating alteration to the same gene (BRCA1) in their tumors."

Are you referring https://www.signaldb.org/variant/17:g.41209079_41209080insG? Could you explain more about what are these numbers? 
What is 29 and 17152 in `29 of 17152 (0.17%)`?
What is 24 and 29 in `24 of the 29 carriers (82.8%)`?

jjgao:
> > * [ ]   Let's show the prevalence only, e.g. `0.2%` instead of `0.2% prevalence (85.7% ballelic)` to simplify the interface
> 
> How about the download? Do we still want to have `0.2% prevalence (85.7% biallelic)` in download file? For now I still keep the `0.2% prevalence (85.7% biallelic)` format in download

Let's just include the prevalence number in download.
 
> > * [ ]   In tooltip, add a summary sentence above the table (similar to COSMIC): "29 of 17152 (0.17%) cancer patients in [SIGNAL - link to the signal variant page] carried this germline mutation (BRCA1 Q1756Pfs*74). 24 of the 29 carriers (82.8%) had a somatic biallelic inactivating alteration to the same gene (BRCA1) in their tumors."
> 
> Are you referring https://www.signaldb.org/variant/17:g.41209079_41209080insG? Could you explain more about what are these numbers?
> What is 29 and 17152 in `29 of 17152 (0.17%)`? My guess is 29 = sum(Carriers) which is sum(Variants), and 17152 = sum(Patients) which is sum(Samples). Is that right?
> What is 24 and 29 in `24 of the 29 carriers (82.8%)`? My guess is 29 = sum(Carriers) which is sum(Variants), while 24 = sum(Variants have none-zero biallelic). Right?

All correct.

> > * [ ]   (good to apply to signaldb.org too) Can we add tooltip to explain each header?
> 
> Could you give me some descriptions for columns? Here is what I have:

* Tumor Type: (let's change it to Cancer Type to be consistent with cbioportal) Main cancer type based on OncoTree
* Mutation Status: Mutations status can be: Somatic, Pathogenic Germline, or Rare Benign/VUS Germline
* `# Patients`: Total number of patients who were diagnosed with the corresponding cancer type
* `# Carriers`: Total number of patients who were diagnosed with the corresponding cancer type and carried the mutation
* % Prevalence: # Carrers / # Patients
* % Biallelic: Percent of pathogenic germline carriers with somatic biallelic inactivation in the corresponding tumor sample.
* N Prevalence: ?
* Median Age at Dx: Median patient age of patients when diagnosed
* Median TMB: Median tumor somatic mutational burden (TMB)
* MSI Score: Median microsatellite instability (MSI) score determined by MSIsensor
* Median HRD LST: ?
* Median HRD ntelomeric AI: ?
* Median HRD Fraction LOH %: ?

@leexgh please ask Chai for the missing ones and double-check the others.

leexgh:
@jjgao Thanks! After checking the logic again, I found the biallelic ratio is calculated by `biallelicCounts / qcPassCounts`
https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/cbioportal-utils/src/signal/SignalMutationUtils.ts#L69-L72

So for my question above: "What is 24 and 29 in `24 of the 29 carriers (82.8%)`", if we follow this logic, the answer would be: 24 is biallelicCounts, 29 is qcPassCounts. So it's not what I wrote in the last comment(9 = sum(Carriers), 24 = sum(Variants have none-zero biallelic)). @onursumer Do you think this makes sense?

But I still got some questions:

For [this mutation 17:g.41246061C>T](https://www.signaldb.org/variant/17:g.41246061C%3ET) in signal genes page, it has `-` in `% Biallelic` column, and the biallelic ratio value in this column is getting from [here](https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/cbioportal-utils/src/signal/SignalMutationUtils.ts#L65-L73), which is `null` and we transfer it to a `-`:
image

But in its table, the `% Biallelic` column is not always 0. The biallelic ratio in table is getting from [here](https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/cbioportal-utils/src/signal/SignalMutationUtils.ts#L107-L113).

So if I look up this mutation in cbioportal mutation table signal column, then it looks like this:
image
The problem is the summary on the top doesn't make sense in this way: "**4** of the **15** carriers (**0%**) had a somatic biallelic inactivating alteration to the same gene (BRCA1) in their tumors."


I think the main reason is this mutation is not pathogenic, so when we calculate total biallelic ratio from [here](https://github.com/cBioPortal/cbioportal-frontend/blob/master/packages/cbioportal-utils/src/signal/SignalMutationUtils.ts#L107-L113), it checks if it's pathogenic and returns `null` for this particular case. 

@jjgao @onursumer What would be a good way to handle this conflict in tooltip summary?

packages/cbioportal-utils/src/signal/SignalMutationUtils.ts
```diff
@@ -195,3 +196,16 @@ export function calculateTotalVariantRatio(
 ) {
     return totalVariants(counts1) / totalVariants(counts2);
 }
+
+export function getMutationStatus(mutation: SignalMutation) {
```
onursumer, line 200: maybe rename this as `getSignalMutationStatus`? Theoretically we may have a `getMutationStatus` function in `MutationUtils` too. There would be an export conflict in that case. Not a big deal though, we can always rename functions when we have such conflicts.

```suggestion
export function getSignalMutationStatus(mutation: SignalMutation): SignalMutationStatus {
```

packages/cbioportal-utils/src/signal/SignalMutationUtils.ts
```diff
@@ -195,3 +196,16 @@ export function calculateTotalVariantRatio(
...
+    if (isGermlineMutation(mutation)) {
+        if (isPathogenicMutation(mutation)) {
+            return SignalMutationStatus.PATHOGENIC_GERMLINE;
+        } else {
+            return SignalMutationStatus.BENIGN_GERMLINE;
+        }
+    } else if (isSomaticMutation(mutation)) {
+        return SignalMutationStatus.SOMATIC;
+    }
+    return 'Unknown';
```
onursumer, line 210: Would it make sense to add `SignalMutationStatus.UNKNOWN`?

packages/cbioportal-frontend-commons/src/components/signal/MutationTumorTypeFrequencyTable.tsx
```diff
@@ -0,0 +1,380 @@
+import {
```
onursumer, line 1: This was moved from `packages/react-variant-view/src/component/frequencyTable/MutationTumorTypeFrequencyTable.tsx`, right? It is weird that it shows up as a new file.

onursumer (changes requested):
Looks good in general. Just a few minor requests.

packages/react-mutation-mapper/src/component/mutationTable/DefaultMutationTable.tsx
```diff
@@ -151,6 +152,17 @@ export default class DefaultMutationTable extends React.Component<
                   );
     }
 
+    @computed
+    get signalAccessor() {
+        return this.hgvsgColumnDataStatus === 'pending'
```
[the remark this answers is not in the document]
leexgh: I change the name to `indexedVariantAnnotationDataStatus` because `hgvsg` and `hgvsc` are also using this function,  no need to create another duplication.

packages/cbioportal-frontend-commons/src/components/signal/MutationTumorTypeFrequencyTable.tsx
leexgh: I moved this to cbioportal-commons because now we use it in both cbioportal and signal

leexgh:
@jjgao I discussed with @onursumer, for the biallelic ratio problem I mentioned above, we thought the temp solution would be only showing biallelic description(24 of the 29 carriers...) for pathogenic mutations. On signal we only show overall biallelic ratio for pathogenic mutations, so I believe this also makes sense to do the same here. What do you think? Do you have other ideas to avoid confusion?

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -0,0 +1,147 @@
...
+import { observer } from 'mobx-react';
+import { errorIcon, loaderIcon } from '../StatusHelpers';
+type SignalProps = {
+    mutation: Mutation;
+    indexedVariantAnnotations?: RemoteData<
+        { [genomicLocation: string]: VariantAnnotation } | undefined
+    >;
+};
+
+export function getSignalcColumnData(
```
onursumer, line 28: There is a typo here, right?

```suggestion
export function getSignalColumnData(
```

We can maybe just name it like this:

```suggestion
export function getSignalData(
```

onursumer (commented):
Looks good. Thank you! Just added a couple more minor comments about variable/function naming.

onursumer (approved):
Thanks!

onursumer:
Standalone mutation mapper Grch38 e2e test seems flaky. Merging this one, but we may need to address the issue if it persists.

leexgh:
There is a test failing probably related to this pr(fetching signal makes it slow), but it works fine locally so ok to merge, but need to investigate.
https://github.com/cBioPortal/cbioportal/issues/8283

PR #3548, opened 2020-12-22
Allow '_SURVIVAL' word to occur in custom survival clin vars

pvannierop:
# Problem
Clinical variable name _OVERALL_SURVIVAL_ for survival data is not accepted.

# Cause
Frontend code creates a unique key for this by appending __SURVIVAL_ to the name. The original is back translated by removing the __SURVIVAL_ element. When the __SURVIVAL_ word occurs in the name of the clinical var this results in an incorrect name.

# Solution
The correct name is obtained when using `lastIndexOf('_SURVIVAL')` instead of `indexOf('_SURVIVAL')`.

### DIFF
--- a/src/pages/studyView/tabs/SummaryTab.tsx
+++ b/src/pages/studyView/tabs/SummaryTab.tsx
@@ -315,7 +315,7 @@ export class StudySummaryTab extends React.Component<
                     ? this.store.survivalDescriptions.result![
                           chartMeta.uniqueKey.substring(
                               0,
-                              chartMeta.uniqueKey.indexOf('_SURVIVAL')
+                              chartMeta.uniqueKey.lastIndexOf('_SURVIVAL')
                           )
                       ][0]
                     : undefined;


dippindots:
Hi @pvannierop, yes, if the `_SURVIVAL` appears in the clinical attribute, we will see errors.

Could you explain why using lastIndexOf('_SURVIVAL') instead of indexOf('_SURVIVAL') would solve the problem? I don't see it would have a different result for `OVERALL_SURVIVAL`.

dippindots:
This change looks good to me, it's better to get the lastIndex. However, I don't know why this would be a problem for the `OVERALL_SURVIVAL` attribute since it would never touch this part of the code.

dippindots (approved):

jjgao:
@pvannierop could you answer @dippindots's question before we merge?

pvannierop:
@jjgao @dippindots Ok, apparently internally the code adds a suffix __SURVIVAL_ to any user defined name for survival clinical data. For instance, when defining _MY_SURV_DATA_ in the clinical file, the internal name becomes _MY_SURV_DATA_SURVIVAL_. To derive the original name for the clinical variable the logic is reversed by stripping the __SURVIVAL_ suffix using position 0 to indexOf('_SURVIVAL'). For _MY_SURV_DATA_SURVIVAL_ this would correctly be _MY_SURV_DATA_. However, when using __SURVIVAL_ in the name of the variable this logic is not correct. For instance, when using name _MY_OVERALL_SURVIVAL_ the frontend labels it to _MY_OVERALL_SURVIVAL_SURVIVAL_ and then incorrectly reverses it to _MY_OVERALL_. Using this name to access the clinical data results in null pointer exception. The use of lastIndex('_SURVIVAL') solves this problem.

dippindots:
Thanks @pvannierop!

PR #3551, opened 2020-12-22
plots tab - add driver-vs-vus mode for mutation data

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/6742


src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1093,29 +1097,28 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
-        _.groupBy(mutations, m => m.uniqueSampleKey),
-        sampleMuts =>
-            _.uniq(
-                sampleMuts.map(
-                    m => mutationTypeToDisplayName[getOncoprintMutationType(m)]
-                )
+    const sampleToMutationTypes = _.mapValues(sampleToMutations, sampleMuts =>
+        _.uniq(
+            sampleMuts.map(
+                m => mutationTypeToDisplayName[getOncoprintMutationType(m)]
```
alisman, line 1112: should we generalize the function name here?   seems like the mutationtype is not just oncoprint related?

adamabeshouse: I agree but for now I think it's okay as a shorthand for a certain taxonomy. we already have "simplified mutation type" which is something else. do you have another idea?

src/pages/resultsView/plots/PlotsTabUtils.tsx
```diff
@@ -1138,13 +1141,28 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
                     // if more than one type, its "Multiple"
                     if (sampleMutTypes.length > 1) {
                         value = MUT_PROFILE_COUNT_MULTIPLE;
                     } else {
                         value = sampleMutTypes;
                     }
                     break;
+                case MutationCountBy.DriverVsVUS:
+                    if (_.some(sampleMutations, m => m.putativeDriver)) {
+                        value = MUT_PROFILE_COUNT_DRIVER;
```
alisman, line 1161: i know this is existing code but would be nice to have a more descriptive nam for "value"

alisman (approved):

src/pages/resultsView/plots/PlotsTabUtils.tsx
adamabeshouse: good idea. I'll do that before merging

PR #3557, opened 2020-12-30
add isoform mskcc for signal variant page GN query

leexgh:
Fix https://github.com/knowledgesystems/signal/issues/93

### DIFF
--- a/packages/react-variant-view/src/component/headerAnnotation/HeaderAnnotation.tsx
+++ b/packages/react-variant-view/src/component/headerAnnotation/HeaderAnnotation.tsx
@@ -31,6 +31,7 @@ interface IHeaderAnnotationProps {
     isCanonicalTranscriptSelected?: boolean | undefined;
     allValidTranscripts: string[];
     onTranscriptSelect(transcriptId: string): void;
+    isoformOverrideSource?: string;
 }
 
 interface IMutationTypeFormat {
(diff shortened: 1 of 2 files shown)


packages/react-variant-view/src/component/headerAnnotation/HeaderAnnotation.tsx
```diff
@@ -363,7 +363,9 @@ export default class HeaderAnnotation extends React.Component<
                 <a
                     href={`https://www.genomenexus.org/annotation/${
                         this.props.variant
-                    }?fields=${ANNOTATION_QUERY_FIELDS.join(',')}`}
+                    }?isoformOverrideSource=mskcc&fields=${ANNOTATION_QUERY_FIELDS.join(
```
onursumer, line 366: Should we make `isoformOverrideSource` a prop with the default value of `mskcc`?

packages/react-variant-view/src/store/VariantStore.ts
```diff
@@ -33,6 +33,7 @@ export class VariantStore {
             return await this.genomeNexusClient.fetchVariantAnnotationGET({
                 variant: this.variant,
                 fields: ANNOTATION_QUERY_FIELDS,
+                isoformOverrideSource: 'mskcc',
```
onursumer, line 36: add `@observable public isoformOverrideSource: string = 'mskcc'` and refer it here as `this.isoformOverrideSource`?

onursumer (commented):
Not sure if we have a use case at this point but it might be better to make `isoformOverrideSource` customizable instead of hardcoding `mskcc`.

onursumer (approved):
LGTM!

PR #3560, opened 2021-01-02
Fix routing issue for nested portals

alisman:


alisman:
@inodb got to bottom of it.  in fact, we had not upgraded history package to 5.0 and it still supports basePath.  will do a bit more testing.

inodb:
@alisman Nice! Thanks for figuring it out!

PR #3566, opened 2021-01-07
Display interpretation instead of ClinVar ID

onursumer:
Fix cBioPortal/cbioportal#8308
Related to knowledgesystems/signal/issues/78

- Display clinical significance as the column value
- Display detailed clinical interpretation (as in the variant view) as a tooltip
- Linkout (ClinVar ID) in the tooltip as well

ClinVar


packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
```diff
@@ -0,0 +1,148 @@
...
+        clinicalSignificance: string;
+        count: number;
+    }[];
+};
+
+export type RcvCountMap = {
+    [origin: string]: { [clinicalSignificance: string]: number };
+};
+
+export function getClinVarInterpretation(rcvData: RcvData[]) {
```
alisman, line 33: might as well make this React.FunctionComponent

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
```diff
@@ -0,0 +1,148 @@
...
+    // first map by origin, then count evidence number for each origin group
+    const rcvMap = _.groupBy(filteredRcv, d => d.origin);
+
+    return _.mapValues(rcvMap, rcvs =>
+        _.countBy(rcvs, rcv => rcv.clinicalSignificance)
+    );
+}
+
+export function getRcvData(rcvCountMap: RcvCountMap): RcvData[] {
+    return _.map(
```
alisman, line 66: pretty sure _.map will already iterate over object.    (val, key)=>{}

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
```diff
@@ -0,0 +1,148 @@
...
+                    ([clinicalSignificance, count]) => {
+                        return { clinicalSignificance, count };
+                    }
+                ),
+            };
+        }
+    );
+}
+
+export function getClinicalSignificance(rcvData: RcvData[]) {
```
alisman, line 82: formatClinicalSignificanceText?

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
```diff
@@ -0,0 +1,148 @@
...
+    return _.uniq(
+        _.flatten(
+            rcvData.map(d => d.evidences.map(e => e.clinicalSignificance))
+        )
+    ).join(', ');
+}
+
+@observer
+export default class ClinVarSummary extends React.Component<
```
alisman, line 91: might as well make this functional component.   don't forget to wrap with observer()

packages/react-mutation-mapper/src/component/column/ClinVar.tsx
```diff
@@ -17,10 +20,13 @@ export function download(myVariantInfo?: MyVariantInfo): string {
     return value ? value.toString() : '';
 }
 
-export function sortValue(myVariantInfo?: MyVariantInfo): number | null {
-    const id = getClinVarId(myVariantInfo);
+export function sortValue(myVariantInfo?: MyVariantInfo): string | null {
```
alisman, line 23: seems pretty heavy for a sort.  will this be used on a large collection?

packages/react-variant-view/src/component/pathogenicity/ClinVarInterpretation.tsx
```diff
@@ -1,93 +1,66 @@
...
 
-interface IClinVarInterpretationProps {
+export interface IClinVarInterpretationProps {
     clinVar: ClinVar | undefined;
 }
 
-enum ClinVarOrigin {
-    GERMLINE = 'germline',
-    SOMATIC = 'somatic',
+function clinVarTooltip() {
```
alisman, line 18: function component

packages/react-variant-view/src/component/pathogenicity/ClinVarInterpretation.tsx
```diff
@@ -1,93 +1,66 @@
...
         return undefined;
     }
 
     @computed get clinVarContent() {
-        // first map by origin, then count evidence number for each origin group
         return this.rcvDisplayData ? (
             <div className={featureTableStyle['feature-table-layout']}>
                 <div className={featureTableStyle['data-source']}>
-                    {this.clinVarTooltip()}
+                    {clinVarTooltip()}
```
alisman, line 54: this is where we can then just use normal component synax if we just make it a function componet.

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
onursumer: good point, I'll update this

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
onursumer: Hmm. We extract unique clinical significance values from the RcvData here. So it is not a simple formatting. Probably still better than `get` :)

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
onursumer: Functional component is a good idea. We may not even need the `observer()` actually, this component seems pretty static to me.

packages/react-mutation-mapper/src/component/column/ClinVar.tsx
onursumer: This one is used to sort the column in the table. So, the collection size is the number of mutations. For a table with ~5000 mutations sorting is pretty fast. If we happen to have issues at some point, we can either try to improve this function or disable sorting for this column. Since this column is hidden by default not sure if it is worth to optimize sort at this point.

leexgh: Does it sort by alphabetical order?

src/shared/components/mutationTable/column/ClinVarColumnFormatter.tsx
```diff
@@ -29,7 +27,7 @@ export default class ClinVarColumnFormatter {
         return (
             <div data-test="clinvar-data">
                 <ClinVar
-                    className={generalStyles['integer-data']}
+                    className={''}
```
leexgh, line 30: why do we need a className here?

packages/react-mutation-mapper/src/component/clinvar/ClinVarSummary.tsx
```diff
@@ -0,0 +1,138 @@
...
+
+const ClinVarSummary = (props: ClinVarSummaryProps) => {
+    const clinVar = props.myVariantInfo
+        ? props.myVariantInfo.clinVar
+        : undefined;
+
+    if (!clinVar) {
+        return (
+            <DefaultTooltip
+                placement="topRight"
```
leexgh, line 96: Probably good to move this to left
image

leexgh (commented):
Thanks!

packages/react-mutation-mapper/src/component/column/ClinVar.tsx
onursumer: yes, but it might be better to sort like `Oncogenic > Likely Oncogenic > ...`

src/shared/components/mutationTable/column/ClinVarColumnFormatter.tsx
onursumer: Good catch. We should just update this one instead: https://github.com/cBioPortal/cbioportal-frontend/blob/573529f93b08fab807b624518f4d89402d0c9cdc/packages/react-mutation-mapper/src/component/column/ClinVar.tsx#L39

jjgao (approved):
looks great from ui.

inodb (approved):

alisman (approved):

leexgh (approved):

PR #3567, opened 2021-01-07
(GC) Remove survival months attributes from the clinical table

dippindots:
Partial of cBioPortal/cbioportal#8254
Remove survival attributes(`_STATUS` and `_MONTHS`) from the clinical table

- Before
image
image

- After
image
image

Testing:
https://deploy-preview-3567--cbioportalfrontend.netlify.app/results/comparison?cancer_study_list=blca_msk_tcga_2020&tab_index=tab_visualize&case_set_id=blca_msk_tcga_2020_all&Action=Submit&gene_list=TP53&comparison_subtab=clinical


dippindots:
Might need to update some screenshots.

inodb (approved):
@dippindots Thanks! Excellent work!

Fine to merge IMO after updating screenshots

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -191,7 +192,14 @@ export default class ClinicalData extends React.Component<
 
     private tableDataStore = new ClinicalDataEnrichmentStore(
         () => {
-            return this.props.store.clinicalDataEnrichmentsWithQValues.result;
+            // exclude survival months and status from the data store
+            return this.props.store.clinicalDataEnrichmentsWithQValues.result.filter(
+                d =>
+                    !getSurvivalAttributeIdsFromPrefix(
```
alisman, line 198: this is not really optimal.  you are recreating this array of attributeIds every cycle, when i think it will always tbe the same.   also, using the "in" operator would be faster.     so, make an object { '??_SURVIVAL':null,  ...  } and then condition:     
```d.clincialAttribute.clinicalAttributeId in obj```

File: src/pages/groupComparison/ClinicalData.tsx [hunk 2]
```diff
@@ -191,7 +192,14 @@ export default class ClinicalData extends React.Component<
 
     private tableDataStore = new ClinicalDataEnrichmentStore(
         () => {
-            return this.props.store.clinicalDataEnrichmentsWithQValues.result;
+            // exclude survival months and status from the data store
```
alisman, line 195: are we sure we don't want to filter these out higher up in our application?  are they useable elsewhere, just not in this table?   we should include in the comment WHY we are excluding them.

inodb: @alisman good point - the tests in the clinical tab are not appropriate for survival data according to the biostats team

I think for now we only want to exclude it from the comparison tab but we might want to exclude it from the plots tab as well in the future

src/pages/groupComparison/ClinicalData.tsx
dippindots: updated

src/pages/groupComparison/ClinicalData.tsx
dippindots: @alisman Added more comments to explain why: 
```
// ClinicalData is only used in Group Comparison Page and Group Comparison Tab
// And survival related attributes are used to create KM plots in survival tab
// We are doing the proper statistical tests in the survival tab and the used tests for the clinical tab are not appropriate for that type of data
// There is no need to still keep them in clinical tab
// So, exclude them from the data store
```

PR #3570, opened 2021-01-08
Add statistical caution info to comparison clinical tab

inodb:
Show statistical caution info on both survival and clinical tab. Before it was only on survival tab

Screen Shot 2021-01-08 at 3 54 24 PM

Screen Shot 2021-01-08 at 3 55 12 PM

In addition add another info icon for genie bpc specific data:

Screen Shot 2021-01-08 at 5 57 15 PM

Related to https://github.com/cBioPortal/cbioportal/issues/8254


src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -143,7 +144,10 @@ export default class ClinicalData extends React.Component<
                 );
             } else {
                 content.push(
-                    <OverlapExclusionIndicator store={this.props.store} />
+                    <div className={'tabMessageContainer'}>
```
inodb, line 147: needed this class to get proper margin around the messages

alisman: that class was meant for a particular spot on the page so might be problematic here in future.

inodb:
unrelated screenshot test failure

PR #3571, opened 2021-01-09
(GC )hide columns for clinical tab

dippindots:
Fix cBioPortal/cbioportal#8263

Quick fix for hiding columns for clinical tab on comparison page / comparison tab
We should create it's own config parameter later, for now, we are reuse parameter `survival_show_p_q_values_in_survival_type_table`


inodb (approved):
Thanks so much for adding this!

PR #3573, opened 2021-01-11
Fix error in loading chart data that are in user preference

kalletlak:
Some tables where showing up empty in https://genie-private.cbioportal.org/study/summary?id=crc_genie_bpc


Luke-Sikina (approved):

inodb (approved):
Thanks for the quick fix!

alisman (approved):

PR #3581, opened 2021-01-12
remove build of image from fingerprint step

inodb:
Move build image step before fingerprinting so we can see the output of building the image


inodb:
merged this b/c it just adds more logging, currently having a hard time debugging the backend localdb tests which relies on this change

PR #3585, opened 2021-01-13
Include OncoKB diagnostic and prognostic implications

zhx828:
## Feature changes

- Show OncoKB icons separately. In the previous compact mode, we only have biological effect and therapeutic implications. We like to keep the consistency for all data types(Biological Effect, Therapeutic Implication, Diagnostic Implication, Prognostic Implication).

- Use the same interface for all data types but show the tab based on the data type user hovers.

- Do not reserve icon space for data types that not available for all variants.

## Interface changes
|  | Before | After |
|-|-|-|
| Icons | Icons Before | Icons After |
| Tooltip | Tooltip Before | Tooltip After |

## Issues to be fixed before merging

- [x] The biological effect tab needs a loading indicator when it fetches ncbi for publications

- [x] The screenshots are failing, seems like some of the tables haven't been updated with the `availableDataTypes`

- [x] Dx/Px implication model does not include references. The summary is not released but rather the summary.

- [x] The tab needs a max-height when there are too many references https://cbioportal.mskcc.org/patient?studyId=mskimpact&sampleId=P-0033455-T02-IH3 SMARCB1 Deletion


packages/react-mutation-mapper/src/component/oncokb/OncoKbHelper.tsx
```diff
@@ -57,6 +88,145 @@ export default class OncoKbHelper {
...
+                    </DefaultTooltip>
+                </span>
+            );
+        }
+        return (
+            <div style={{ whiteSpace: 'normal', lineHeight: '1rem' }}>
+                {content}
+            </div>
+        );
+    };
```
zhx828, line 231: No need to review, this is from previous codebase.

packages/react-mutation-mapper/src/component/oncokb/OncoKbHelper.tsx
```diff
@@ -57,6 +88,145 @@ export default class OncoKbHelper {
...
 }
```
zhx828, line 232: No need to review, this is from previous codebase.

packages/react-mutation-mapper/src/component/oncokb/OncoKbHelper.tsx
```diff
@@ -57,6 +88,145 @@ export default class OncoKbHelper {
...
+                    Header: <span>Alteration(s)</span>,
+                    accessor: 'alterations',
+                    minWidth: 80,
+                    sortMethod: (a: string[], b: string[]) =>
+                        defaultArraySortMethod(a, b),
+                };
+        }
+    }
+
+    public static getAlterationsColumnCell = (
```
zhx828, line 189: This is from previous codebase.

packages/react-mutation-mapper/src/component/oncokb/OncoKbHelper.tsx
```diff
@@ -57,6 +88,145 @@ export default class OncoKbHelper {
...
+
+    public static levelTooltipContent = (level: string) => {
+        return (
+            <div style={{ maxWidth: '200px' }}>
+                {OncoKbHelper.LEVEL_DESC[level]}
+            </div>
+        );
+    };
+
+    public static getDefaultColumnDefinition(
```
zhx828, line 140: This is from previous codebase.

Leowisd (approved):

onursumer:
@zhx828 Now we are showing multiple icons for OncoKB would it make sense to have a separate `OncoKB` column?

zhx828:
> @zhx828 Now we are showing multiple icons for OncoKB would it make sense to have a separate `OncoKB` column?

We tried it but the group does not like oncokb as a separate column. We might need to upgrade our table component to allow subgrouping.  @onursumer

yarn.lock
```diff
@@ -16180,16 +16180,10 @@ oncokb-styles@~0.1.2:
   resolved "https://registry.yarnpkg.com/oncokb-styles/-/oncokb-styles-0.1.2.tgz#8b26c0a0829787cdc1b595d3a021b3266607102b"
   integrity sha512-tuy5s3qFxgf1ogMATQSRPNgLlAMrvOOTCAN1dm/wJ+VZoStbJ7g36/qHwc99UPfh3vrB05broLodF+k58p5tUw==
 
-oncoprintjs@4.2.3:
```
zhx828, line 16183: Not sure why this is deleted but the oncoprintjs is still in the yarn.lock

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -203,12 +213,19 @@ export function getAnnotationData(
                     resolveTumorType,
                     resolveEntrezGeneId
                 );
+                oncoKbAvailableDataTypes = _.uniq([
```
alisman, line 216: seems like we're calling uniq twice once inside of calculateOncoKBAlavailabe... function ?  maybe just move this above if statement and allow it to handle empty onckbData.result?

zhx828: We need to make sure OncoKbCardDataType.BIOLOGICAL, 
 is always attached whether the if statement is passed or `_.values(oncoKbData.result.indicatorMap)` is not empty. Keeping it out of if statement would just make the logic more complex. We are scarifying the performance a bit here.

packages/react-mutation-mapper/src/component/oncokb/OncoKB.tsx
```diff
@@ -114,29 +127,75 @@ export default class OncoKB extends React.Component<IOncoKbProps, {}> {
...
-                        {oncoKbContent}
-                    </DefaultTooltip>
-                );
             }
         }
 
         return oncoKbContent;
     }
 
+    private getAnnotationIcon(type: OncoKbCardDataType) {
```
onursumer, line 136: Make a functional component `AnnotationIcon` instead?

packages/react-mutation-mapper/src/component/oncokb/OncoKbTreatmentTable.tsx
```diff
@@ -20,6 +20,7 @@ import SummaryWithRefs from './SummaryWithRefs';
 
 import mainStyles from './main.module.scss';
 import './oncoKbTreatmentTable.scss';
+import { EvidenceReferenceContent } from './oncokbCard/EvidenceReferenceContent';
```
onursumer, line 23: Very minor but can we move this above the style imports? And also while we are at it can we merge
```
import { LEVELS } from 'cbioportal-utils';
```
and
```
import { defaultArraySortMethod, defaultSortMethod } from 'cbioportal-utils';
```
into a single import statement?

packages/react-mutation-mapper/src/component/oncokb/ReferenceList.tsx
```diff
@@ -7,15 +7,34 @@ import ArticleAbstractItem from './ArticleAbstractItem';
 import PmidItem from './PmidItem';
 
 import styles from './listGroupItem.module.scss';
+import { observer } from 'mobx-react';
+import { computed } from 'mobx';
```
onursumer, line 11: again very minor. can we move these to the topmost import block, since these are external dependecies?

yarn.lock
onursumer: probably this version not used anymore?

onursumer (commented):
Looks good! Just a few minor suggestions

packages/react-mutation-mapper/src/component/oncokb/OncoKbTreatmentTable.tsx
zhx828: Weird they are not automatically updated. I will update.

yarn.lock
zhx828: Yeah I suppose so.

PR #3593, opened 2021-01-19
fix mutation mapper tool test

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8283
Reason:
Query `signal` data makes it a little slower than before.

Solution:
- Only fetch `signal` when it's enabled
- Add waiting time

Changes in this pr:
- Fix mutation mapper tool test
- Using `GENOME_NEXUS_ARG_FIELD_ENUM` on query fields

Fix tests:
1. `should show all transcripts when using combination of genomic and protein changes`
2. `should correctly annotate the genomic changes example with GRCh38 and display the results`

### DIFF
--- a/src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
+++ b/src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
@@ -50,6 +50,7 @@ import defaultGenomeNexusClient from 'shared/api/genomeNexusClientInstance';
 import defaultGenomeNexusInternalClient from 'shared/api/genomeNexusInternalClientInstance';
 import autobind from 'autobind-decorator';
 import { getGenomeNexusHgvsgUrl } from 'shared/api/urls';
+import { GENOME_NEXUS_ARG_FIELD_ENUM } from 'shared/constants';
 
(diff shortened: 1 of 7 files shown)


onursumer (approved):
LGTM

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -219,7 +220,13 @@ export default class MutationMapperToolStore {
             invoke: async () =>
                 await fetchVariantAnnotationsIndexedByGenomicLocation(
                     this.rawMutations,
-                    ['annotation_summary', 'hotspots', 'signal'],
+                    [
+                        GENOME_NEXUS_ARG_FIELD_ENUM.ANNOTATION_SUMMARY,
+                        GENOME_NEXUS_ARG_FIELD_ENUM.HOTSPOTS,
+                        AppConfig.serverConfig.show_signal
+                            ? GENOME_NEXUS_ARG_FIELD_ENUM.SIGNAL
+                            : '',
```
alisman, line 228: better to return null than empty string.

alisman (approved):

PR #3595, opened 2021-01-20
Add y-axis to line chart timeline tracks

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8219

image

[Link](https://www.cbioportal.org/patient?caseId=MPCProject_0muduPhG&studyId=prad_mpcproject_2018)

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
@@ -1,11 +1,16 @@
 import React, { useCallback } from 'react';
-import { TimelineTrackSpecification } from './types';
+import { TimelineTrackSpecification, TimelineTrackType } from './types';
 import { TICK_AXIS_HEIGHT } from './TickAxis';
 import { CustomTrackSpecification } from './CustomTrack';
 import { TimelineStore } from './TimelineStore';
-import { useLocalStore, useObserver } from 'mobx-react-lite';
+import { useObserver } from 'mobx-react-lite';
 import { TruncatedText } from 'cbioportal-frontend-commons';
 import { isTrackVisible } from './lib/helpers';
+import LineChartAxis, {
+    LINE_CHART_AXIS_SVG_WIDTH,
+    LINE_CHART_AXIS_TICK_WIDTH,
+} from './LineChartAxis';
+import ReactDOM from 'react-dom';
 
 interface ITrackHeaderProps {
     store: TimelineStore;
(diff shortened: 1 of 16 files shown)


packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
```diff
@@ -44,6 +52,17 @@ const TrackHeader: React.FunctionComponent<ITrackHeaderProps> = function({
                 <span>
                     <TruncatedText text={getTrackLabel(track)} maxLength={20} />
                 </span>
+                {isLineChartTrack && (
```
alisman, line 55: i would like to see if there's any way we can make this generic, so TrackHeader doesn't know about linechart type.   the problem is we need to be able to render a child of the track into this region, but  somehow from within the track component.  lets think on this.

alisman:
I think we should make the axis text and tick color match the colors of the horizontal timeline ticks perfectly.  making it a little lighter will distinguish it from label.

also, axis font and alignment seems to get lost a bit in export svg.   this is probably because style info needs to be inlined in order to make it to export?

packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
adamabeshouse: Im not sure I agree. so far the only thing we need to bring into the track headers is an axis. The generic route as far as I see is to make it more like doesTrackHaveAxis which we can do in the future if/when we add other types of tracks with axes. Or we could even embed that in the track specification regardless of track type

alisman:
Looks good except for the decimal issue discussed last week.

adamabeshouse:
@alisman updated

alisman (approved):

src/pages/studyView/charts/scatterPlot/StudyViewDensityScatterPlot.tsx
```diff
@@ -14,7 +14,7 @@ import {
 } from 'cbioportal-frontend-commons';
 import { computed, makeObservable, observable } from 'mobx';
 import autobind from 'autobind-decorator';
-import { tickFormatNumeral } from '../../../../shared/components/plots/TickUtils';
+import { tickFormatNumeral } from 'cbioportal-frontend-commons/src/lib/TickUtils';
```
alisman, line 17: this is not right way to import from package.  you have to expose it via index.tsx so that you can just do 
```
import blah from 'cbioportal-frontend-commons'
```

.   there's a test that catches this.

adamabeshouse: ah yeah, I think the refactoring function messed that up

PR #3596, opened 2021-01-20
Treatment counts display correctly on GENIE

Luke-Sikina:
Fix cBioPortal/cbioportal#8288

- The counts for treatments in the chart select were not calculated
correctly... or at all. How this is working on the public portal
is a mystery to me
- Made it so the counts are the number of unique sample id + study id
pairs in the respective treatment responses.

## Any screenshots or GIFs?
Screenshot from 2021-01-20 14-03-49

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -7304,11 +7304,34 @@ export class StudyViewPageStore {
                     );
                 }
 
+                const calculateSampleCount = (
+                    result:
+                        | (SampleTreatmentRow | PatientTreatmentRow)[]
+                        | undefined
+                ) => {
+                    if (!result) {
+                        return 0;
+                    }
+
+                    return result
+                        .map(row => row.samples)
+                        .reduce((allSamples, samples) => {
+                            samples.forEach(s =>
+                                allSamples.add(s.sampleId + '::' + s.studyId)
+                            );
+                            return allSamples;
+                        }, new Set<String>()).size;
+                };
+
                 if (!_.isEmpty(this.sampleTreatments.result)) {
-                    ret['SAMPLE_TREATMENTS'] = 1;
+                    ret['SAMPLE_TREATMENTS'] = calculateSampleCount(
+                        this.sampleTreatments.result
+                    );
                 }
                 if (!_.isEmpty(this.patientTreatments.result)) {
-                    ret['PATIENT_TREATMENTS'] = 1;
+                    ret['PATIENT_TREATMENTS'] = calculateSampleCount(
+                        this.patientTreatments.result
+                    );
                 }
 
                 if (!_.isEmpty(this.structuralVariantProfiles.result)) {


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7305,10 +7305,24 @@ export class StudyViewPageStore {
...
                 if (!_.isEmpty(this.patientTreatments.result)) {
-                    ret['PATIENT_TREATMENTS'] = 1;
+                    ret['PATIENT_TREATMENTS'] = this.patientTreatments.result
+                        ?.map(row => row.samples)
+                        .reduce((allSamples, samples) => {
+                            samples.forEach(s =>
+                                allSamples.add(s.sampleId + '::' + s.studyId)
+                            );
+                            return allSamples;
+                        }, new Set<String>()).size as any;
```
inodb, line 7325: as discussed maybe good to make this a function just to make it clear they are the same apart from using `this.sampleTreatments.result` vs `this.patientTreatments.result`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7476,6 +7480,23 @@ export class StudyViewPageStore {
...
     });
 
+    private calculateSampleCount(
+        result: (SampleTreatmentRow | PatientTreatmentRow)[] | undefined
+    ) {
+        if (!result) {
+            return 0;
+        }
+
+        return result
```
alisman, line 7490: this is confusing.  look like you are just returning an integer count, so why then bother populating allSamples with ids?   and why return size as any?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7476,6 +7480,23 @@ export class StudyViewPageStore {
         default: {},
     });
 
+    private calculateSampleCount(
```
alisman, line 7483: this is a pure function (no reference to this), so might as well just have this in module scope (not a method).

src/pages/studyView/StudyViewPageStore.ts
Luke-Sikina: Because a patient or sample can receive multiple treatments, it is possible for the same unique ID to appear in multiple rows. We want to count the number of unique samples being considered by this table, not the summation of each row's sample count. In order to do this, we have to construct a unique collection of sample ids and then return its count, rather than simply counting.

You're right about the `any` cast. I'll fix that.

inodb:
@alisman will take a look

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7304,11 +7304,34 @@ export class StudyViewPageStore {
...
+                    }
+
+                    return result
+                        .map(row => row.samples)
+                        .reduce((allSamples, samples) => {
+                            samples.forEach(s =>
+                                allSamples.add(s.sampleId + '::' + s.studyId)
+                            );
+                            return allSamples;
+                        }, new Set<String>()).size;
```
alisman, line 7323: I see, so we're using Set in order to handle duplicates

alisman (approved):

PR #3599, opened 2021-01-21
(Patient View) Validate mutational signatures data after fetching

dippindots:
Fix cBioPortal/cbioportal#8295

- validate data will show up in pair (has both `contribution` and `pvalue`)

Testing:
https://triage.cbioportal.mskcc.org/patient?studyId=msk_impact_2021&caseId=P-0035874

set `localStorage.netlify ="deploy-preview-3599--cbioportalfrontend"` for testing.

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -197,6 +197,7 @@ import { GenericAssayTypeConstants } from 'shared/lib/GenericAssayUtils/GenericA
 import {
     MutationalSignaturesVersion,
     MutationalSignatureStableIdKeyWord,
+    validateMutationalSignatureRawData,
 } from 'shared/lib/GenericAssayUtils/MutationalSignaturesUtils';
 
 type PageMode = 'patient' | 'sample';
@@ -459,7 +460,7 @@ export class PatientViewPageStore {
                 this.samples,
                 this.mutationalSignatureMolecularProfiles,
             ],
-            invoke: () => {
+            invoke: async () => {
                 const mutationalSignatureMolecularProfileIds = this.mutationalSignatureMolecularProfiles.result.map(
                     profile => profile.molecularProfileId
                 );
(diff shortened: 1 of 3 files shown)


dippindots:
@inodb @alisman I added a testing link in the description. Please let me know if you need to know anything about it. Thanks!

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -472,13 +473,18 @@ export class PatientViewPageStore {
...
                     );
-                    return client.fetchGenericAssayDataInMultipleMolecularProfilesUsingPOST(
+                    const genericAssayRawData = await client.fetchGenericAssayDataInMultipleMolecularProfilesUsingPOST(
                         {
                             genericAssayDataMultipleStudyFilter: {
                                 sampleMolecularIdentifiers,
                             } as GenericAssayDataMultipleStudyFilter,
                         }
                     );
+                    if (
```
alisman, line 483: is the case where we have invalid data something that should immediately be fixed?   i ask because we fail silently here

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -132,3 +133,23 @@ export function getSignificantMutationalSignatures(
...
     );
 }
+
+export function validateMutationalSignatureRawData(
+    mutationalSignatureData: GenericAssayData[]
+): boolean {
+    const regex = new RegExp(
+        `(${MutationalSignatureStableIdKeyWord.MutationalSignatureContributionKeyWord}|${MutationalSignatureStableIdKeyWord.MutationalSignatureConfidenceKeyWord})`
+    );
+    const uniqueProfileIdGroupByVersion = _.chain(mutationalSignatureData)
```
alisman, line 143: is the N here likely to be large?   if so, the following logic could certainly be optimized by using reduce and a dictionary to handle uniqueness.   right now you are looping through the whole thing 5 times when actually only once is necessary.

inodb (approved):
This seems good to me as a quick fix. For a follow up it would be good to show the signatures if only contribution is available (which is the most common case, p value is very specific to our signature calculation method)

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
dippindots: updated with one round solution.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
dippindots: Data fixing is on the way, it's not that urgent since this data is not public yet.

alisman (approved):

PR #3600, opened 2021-01-21
Use MSI_TYPE to determine the MSI-H annotation in patient view

zhx828:
This is to fix https://github.com/cBioPortal/cbioportal/issues/8100

There are only two studies at the moment having MSI_TYPE attribute.
The mskimpact has the following values `Stable`, `Instable`, `Do not report`, `Indeterminate`, `NA`
The coadread_vakianie_impact_cmo_2018 has the following values `Yes`, `No`, `Stable`, `Do not report`, `Indeterminate`
Note: the clinical attribute MSI_TYPE does not have controlled vocabulary.

We could map value `Instable` to OncoKB MSI-H.
The value `No` from coadread_vakianie_impact_cmo_2018 study is correspoding to MSI-H, but I don't think the value itself is generic enough to be used to map to MSI-H. Thoughts? @inodb @jjgao

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -101,6 +101,7 @@ import {
     generateUniqueSampleKeyToTumorTypeMap,
     getGenomeNexusUrl,
     getOtherBiomarkersQueryId,
+    getSampleClinicalDataMapByKeywords,
     getSampleClinicalDataMapByThreshold,
     getSampleTumorTypeMap,
     groupBySampleId,
@@ -178,6 +179,7 @@ import {
 } from 'shared/lib/oql/AccessorsForOqlFilter';
 import {
     CLINICAL_ATTRIBUTE_ID_ENUM,
+    MIS_TYPE_VALUE,
     GENOME_NEXUS_ARG_FIELD_ENUM,
     MSI_H_THRESHOLD,
     TMB_H_THRESHOLD,
@@ -2360,10 +2362,10 @@ export class PatientViewPageStore {
     }
 
     @computed get sampleMsiHInfo() {
-        return getSampleClinicalDataMapByThreshold(
+        return getSampleClinicalDataMapByKeywords(
             this.clinicalDataForSamples.result,
-            CLINICAL_ATTRIBUTE_ID_ENUM.MSI_SCORE,
-            MSI_H_THRESHOLD
+            CLINICAL_ATTRIBUTE_ID_ENUM.MSI_TYPE,
+            [MIS_TYPE_VALUE.INSTABLE]
         );
     }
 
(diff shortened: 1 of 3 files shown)


inodb:
> We could map value `Instable` to OncoKB MSI-H.

that sounds good to me

For follow up it might be good to:

- Add in the oncokb integration docs that we are using MSI_TYPE?
- Add a warning in the validator when using `MSI_TYPE` attribute that it is a controlled vocabulary

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -2350,10 +2351,10 @@ export class PatientViewPageStore {
     }
 
     @computed get sampleMsiHInfo() {
-        return getSampleClinicalDataMapByThreshold(
+        return getSampleClinicalDataMapByKeywords(
             this.clinicalDataForSamples.result,
-            CLINICAL_ATTRIBUTE_ID_ENUM.MSI_SCORE,
-            MSI_H_THRESHOLD
+            CLINICAL_ATTRIBUTE_ID_ENUM.MSI_TYPE,
+            ['Instable']
```
alisman, line 2357: curious if this word is really "Instable" or if it's type?  Unstable?    Either way, maybe make this a constant?

src/shared/lib/StoreUtils.ts
```diff
@@ -1627,6 +1627,25 @@ export function getSampleClinicalDataMapByThreshold(
...
+        clinicalData,
+        (acc, next) => {
+            if (next.clinicalAttributeId === clinicalAttributeId) {
+                if (keywords.includes(next.value)) {
+                    acc[next.sampleId] = next;
+                }
+            }
+            return acc;
+        },
+        {} as { [key: string]: ClinicalData }
```
alisman, line 1645: i believe if you type the acc argument (1637), instead of here, you can avoid using "as" which is stronger.
```
(acc: {...}, next) => {...}

```

alisman (approved):

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
zhx828: @alisman yeah, we use Instable/Stable to describe it https://en.wikipedia.org/wiki/Microsatellite_instability. I will update to constant.

jjgao:
@zhx828 we should add some documentation about this so that other instances know how to format their data to link to OncoKB.

zhx828:
> @zhx828 we should add some documentation about this so that other instances know how to format their data to link to OncoKB.

@jjgao I added some info https://github.com/cBioPortal/cbioportal/pull/8298 But I agree we should prioritize https://github.com/cBioPortal/cbioportal/issues/8084

zhx828:

PR #3602, opened 2021-01-22
Add splice to oncoprinter

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8290


src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
```diff
@@ -414,6 +414,14 @@ export function makeGeneticTrackDatum_Data(
                 proteinChange: 'promoter',
             });
             break;
+        case 'splice':
```
alisman, line 417: seems kind of shocking we don't have an enum for all these.

alisman (approved):

src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
alisman: lets find one or make it

PR #3606, opened 2021-01-25
Comparison tab performance improvements

Luke-Sikina:
Fix cBioPortal/cbioportal#8287

**Problem**:
- For the GENIE dataset, the coparison page was taking
20+ minutes to load from when the spinner appears

**Solution**:
- Added tests for `StudyViewUtils.getPatientIdentifiers`
- Modified `StudyViewUtils.getPatientIdentifiers` to add objects
to a set rather than constructing and then filtering a list
- Modified `GroupCompairsonUtils.getSampleIdentifiers` in the
same manner
  - This method was already well covered, so no tests were added

**Result**:
- Spinner time down to ~38 seconds on my local.

### DIFF
--- a/src/pages/groupComparison/GroupComparisonUtils.tsx
+++ b/src/pages/groupComparison/GroupComparisonUtils.tsx
@@ -247,24 +247,29 @@ export function getStudyIds(groups: Pick<SessionGroupData, 'studies'>[]) {
     );
 }
 
+/**
+ * Gets all unique sample identifiers from the groups
+ * provided. A unique key can be generated for a sample
+ * by appending the sample's study id to the sample's sample id, separated
+ * by a character not allowed in either id- in this case a newline symbol.
+ */
 export function getSampleIdentifiers(
     groups: Pick<SessionGroupData, 'studies'>[]
 ) {
-    return _.uniqWith(
-        _.flattenDeep<SampleIdentifier>(
-            groups.map(group =>
-                group.studies.map(study => {
-                    const studyId = study.id;
-                    return study.samples.map(sampleId => ({
-                        studyId,
-                        sampleId,
-                    }));
-                })
-            )
-        ),
-        (id1, id2) =>
-            id1.sampleId === id2.sampleId && id1.studyId === id2.studyId
-    );
+    const sampleIds: { [key: string]: SampleIdentifier } = {};
+
+    groups.forEach(group => {
+        group.studies.forEach(study => {
+            study.samples.forEach(sample => {
+                sampleIds[study.id + '\n' + sample] = {
+                    studyId: study.id,
+                    sampleId: sample,
+                };
+            });
+        });
+    });
+
+    return Object.values(sampleIds);
 }
 
 export function getNumSamples(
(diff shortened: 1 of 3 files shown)


src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -250,21 +250,20 @@ export function getStudyIds(groups: Pick<SessionGroupData, 'studies'>[]) {
...
-        ),
-        (id1, id2) =>
-            id1.sampleId === id2.sampleId && id1.studyId === id2.studyId
-    );
+    const sampleIds: { [key: string]: SampleIdentifier } = {};
+
+    groups.forEach(group => {
+        group.studies.forEach(study => {
+            study.samples.forEach(sample => {
+                sampleIds[study.id + '\n' + sample] = {
```
alisman, line 258: why linebreak instead of say, underscore?  just curious.  seems strange but i get that it isn't used.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -250,21 +250,20 @@ export function getStudyIds(groups: Pick<SessionGroupData, 'studies'>[]) {
...
-                    }));
-                })
-            )
-        ),
-        (id1, id2) =>
-            id1.sampleId === id2.sampleId && id1.studyId === id2.studyId
-    );
+    const sampleIds: { [key: string]: SampleIdentifier } = {};
+
+    groups.forEach(group => {
```
alisman, line 255: maybe add a comment explaining how we are obtaining uniqueness

src/pages/studyView/StudyViewUtils.spec.ts
```diff
@@ -3442,4 +3443,108 @@ describe('StudyViewUtils', () => {
             ]);
         });
     });
+
```
alisman, line 3446: do we test the uniquness anywhere?

alisman:
@Luke-Sikina curious why we don't use the unique identifier that i believe server provides (a hash of studyid and sampleid).  i realize this predates your PR.

src/pages/groupComparison/GroupComparisonUtils.tsx
Luke-Sikina: Really, we could use any character that we exclude from study, sample and patient ids. I chose to use `\n` because I know that if that character was included in an id, the study would fail to import. I was concerned that other characters may technically be excluded according to our specifications, but that we may not actually enforce that during import.

src/pages/studyView/StudyViewUtils.spec.ts
Luke-Sikina: Yes. The third test has study `S01` and all of its samples in both groups. Without proper uniqueness enforcement, I would expect the three S01 samples to appear twice.

src/pages/groupComparison/GroupComparisonUtils.tsx
Luke-Sikina: Sure. Let me know if the comment on lines 250-255 is what you were looking for. Wasn't sure if you wanted me to explain what constitutes a unique sample, or how the samplieIds data structure enforces uniqueness.

PR #3611, opened 2021-01-26
Improve mobx v6 compatibility of certain components

onursumer:
Related to knowledgesystems/signal#99

## Checks

### DIFF
--- a/packages/cbioportal-frontend-commons/src/components/gradientLegend/GradientLegend.tsx
+++ b/packages/cbioportal-frontend-commons/src/components/gradientLegend/GradientLegend.tsx
@@ -61,6 +61,7 @@ export class GradientLegend extends React.Component<IGradientLegendProps> {
 
         return [
             <text
+                key="gradientLegendLabelMax"
                 fontSize={this.props.fontSize}
                 fontFamily={this.props.fontFamily}
                 x={this.rectX + rectWidth + 4}
@@ -70,6 +71,7 @@ export class GradientLegend extends React.Component<IGradientLegendProps> {
                 {this.props.max.toLocaleString()}
             </text>,
             <text
+                key="gradientLegendLabelMin"
                 fontSize={11}
                 x={this.rectX + rectWidth + 4}
                 y={this.rectY + rectHeight}
--- a/packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
+++ b/packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
@@ -97,7 +97,6 @@ function getColumnVisibilityDef<T>(
     return colVisProp;
 }
 
-@observer
 export default class DataTable<T> extends React.Component<
     DataTableProps<T>,
     {}
--- a/packages/react-mutation-mapper/src/component/filter/DropdownSelector.tsx
+++ b/packages/react-mutation-mapper/src/component/filter/DropdownSelector.tsx
@@ -1,4 +1,3 @@
-import autobind from 'autobind-decorator';
 import { CheckedSelect, Option } from 'cbioportal-frontend-commons';
 import classNames from 'classnames';
 import { action, computed, makeObservable } from 'mobx';
@@ -31,15 +30,30 @@ export type DropdownSelectorProps = {
     options?: { label?: string | JSX.Element; value: string }[];
 };
 
+const SelectionIndicator: React.FunctionComponent<{
+    allValues: string[];
+    selectedValues: { value: string }[];
+    selectionIndicatorClassNames: string;
+}> = observer(props => {
+    return (
+        <div
+            style={{
+                marginRight: 5,
+                marginTop: 'auto',
+                marginBottom: 'auto',
+            }}
+            className={props.selectionIndicatorClassNames}
+        >
+            {props.selectedValues.length}/{props.allValues.length}
+        </div>
+    );
+});
+
 @observer
 export class DropdownSelector extends React.Component<
     DropdownSelectorProps,
     {}
 > {
-    constructor(props: any) {
-        super(props);
-        makeObservable(this);
-    }
     public static defaultProps: Partial<DropdownSelectorProps> = {
         showNumberOfSelectedValues: true,
         selectionIndicatorClassNames: {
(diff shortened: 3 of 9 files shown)


packages/react-mutation-mapper/src/component/dataTable/DataTable.tsx
```diff
@@ -97,7 +97,6 @@ function getColumnVisibilityDef<T>(
     return colVisProp;
 }
 
-@observer
```
onursumer, line 100: mobx is not happy with the the `@observer` decoration added to the base DataTable. Instead we tag the components that extend `DataTable` with the decorator. Ideally we should get rid of inheritance based component implementation, but that's beyond the scope of this PR.

packages/react-mutation-mapper/src/component/mutationTable/DefaultMutationTable.tsx
```diff
@@ -61,6 +61,7 @@ export type DefaultMutationTableProps = {
     appendColumns?: boolean;
 } & DataTableProps<Partial<Mutation>>;
 
+@observer
```
onursumer, line 64: we make this component observer instead of the base `DataTable` class

adamabeshouse (approved):
LGTM

packages/cbioportal-frontend-commons/src/components/gradientLegend/GradientLegend.tsx
```diff
@@ -61,6 +61,7 @@ export class GradientLegend extends React.Component<IGradientLegendProps> {
 
         return [
             <text
+                key="gradientLegendLabelMax"
```
alisman, line 64: is this something that coule appear more than once on a page at the same time?

packages/react-mutation-mapper/src/component/filter/DropdownSelector.tsx
```diff
@@ -31,15 +30,30 @@ export type DropdownSelectorProps = {
     options?: { label?: string | JSX.Element; value: string }[];
 };
 
+const SelectionIndicator = (props: {
```
alisman, line 33: observer?  and i think we should start type this as FunctionComponents.  
```
const SelectionIdicator: React.FunctionComponent<{type}> = function(props) { ... }
```

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
```diff
@@ -188,34 +188,54 @@ export default class MutationMapper<
...
     }
 
+    protected getPlotTopYAxisSymbol(): string | undefined {
+        return this.props.plotTopYAxisSymbol;
+    }
+
     @computed
     protected get plotTopYAxisSymbol(): string | undefined {
-        return this.props.plotTopYAxisSymbol;
+        return this.getPlotTopYAxisSymbol();
```
alisman, line 197: why do we wrap this?

alisman: and all the rest?  just curious

alisman (approved):

packages/cbioportal-frontend-commons/src/components/gradientLegend/GradientLegend.tsx
adamabeshouse: arent keys only relevant within a single list, for react?

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
adamabeshouse: pretty sure this is the workaround for having inheritance/overriding, since we can no longer override a @computed, we have to have subclasses override the method, in this case `getPlotTopYAxisSymbol`. right @onursumer ?

packages/cbioportal-frontend-commons/src/components/gradientLegend/GradientLegend.tsx
onursumer: Just double checked. I see this error message in the console if I remove the key: 
```Warning: Each child in a list should have a unique "key" prop.```

I think it is because we are returning an array here.

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
onursumer: Actually @leexgh implemented this part. I think this is because of the issue with overriding the `@computed` values. mobx 6 doesn't allow that anymore. @leexgh can you confirm?

Ideally we should refactor mutation mapper so that we don't use inheritance anymore. Again, beyond the scope of this PR.

onursumer: Oh, actually, I guess these didn't need to be wrapped as `@computed` values at the first place. @alisman Did you mean that? It's probably my bad to implement this initially this way.

leexgh: @onursumer @alisman Yes this is the workaround for having overriding as @adamabeshouse said above.

leexgh (approved):
Thanks!

leexgh:
@onursumer Just found a bug, on mutations tab, if you add annotation tracks and hover any of dots on tracks, the page will crash.

image

packages/react-mutation-mapper/src/component/mutationMapper/MutationMapper.tsx
onursumer: Just removed the `@computed` decoration for these functions. I don't think they need to be computed. This way we don't need to introduce the additional `get...` functions.

onursumer:
> @onursumer Just found a bug, on mutations tab, if you add annotation tracks and hover any of dots on track, the page will crash. This bug probably not related to this pr, it also happens on current cbioportal.
> 
> image

@leexgh Thanks for pointing this out. @zhx828 noticed this one earlier, there is a separate PR for this: #3614

alisman (approved):

PR #3614, opened 2021-01-27
Fix annotation tracks bug

zhx828:
Update components to be functional components to resolve mobx 6 upgrade issue

### DIFF
--- a/packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
+++ b/packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
@@ -26,12 +26,14 @@ type PtmTrackProps = TrackProps & {
     subTrackMargin?: number;
 };
 
-export function ptmTooltip(
-    ptms: PostTranslationalModification[],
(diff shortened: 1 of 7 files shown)


packages/react-mutation-mapper/src/component/track/PtmTrack.tsx
```diff
@@ -26,12 +26,14 @@ type PtmTrackProps = TrackProps & {
     subTrackMargin?: number;
 };
 
-export function ptmTooltip(
-    ptms: PostTranslationalModification[],
-    pubMedCache?: MobxCache
-) {
-    return <PtmAnnotationTable data={ptms} pubMedCache={pubMedCache} />;
```
zhx828, line 33: @adamabeshouse could you comment on this? If a function returns the component, then the component will throw error in the construction method. The error is `TypeError: Cannot set property 'props' of undefined`

alisman (approved):

onursumer (approved):
LGTM

PR #3623, opened 2021-02-02
Fix webdriver manager not starting selenium

inodb:
For some reason the update command was hanging. I guess some URL is no
longer working. Tried a few different things, not entirely sure what
fixed the issue but I made these changes:

- Change to specific version of chrome in the cirlceci
image (google-chrome --version)
- use an older version of selenium instead of the latest alpha release
- use older version of webdriver-manager

UPDATE:

- there was a minor issue with selenium timing for localdb tests (it's not ready yet when the tests start so added a wait)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -22,8 +22,8 @@ run_e2e_and_save_artifacts: &run_e2e_and_save_artifacts
           yarn serveDist & \
           cd end-to-end-test && \
           yarn install --frozen-lockfile && \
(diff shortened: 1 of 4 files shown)


inodb:
Also noticed there's a third place where the webdriver manager version is defined: https://github.com/inodb/cbioportal-frontend/commits/fix-e2e-webdriver-manager/scripts/setup_localdb.sh. It's not used in CI nor mentioned in the docs. Should keep/edit this in some way?

inodb:
The failing e2e tests are real issues but not related to this pr: https://github.com/cBioPortal/cbioportal/issues/8323

PR #3624, opened 2021-02-03
Improve mutation annotation performance

onursumer:
Addresses the performance issues caused by the latest mobx update.

Related to knowledgesystems/signal#99

## Checks


packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -290,17 +292,15 @@ export function annotateMutation(
 export function initAnnotatedMutation(
     mutation: Partial<Mutation>
 ): Partial<Mutation> {
-    const deepClone = toJS(mutation);
```
onursumer, line 293: `toJS` is now significantly slower. we don't really need to deep clone here, just a shallow copy should be good enough.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -226,15 +227,32 @@ class DefaultMutationMapperStore implements MutationMapperStore {
         }
     }
 
-    public getAnnotatedMutations(transcriptId: string): Mutation[] {
+    @memoize
```
onursumer, line 230: memoize the annotation calls to avoid redundant annotation of the same set of mutations

packages/cbioportal-utils/src/mutation/MutationAnnotator.ts
```diff
@@ -290,17 +292,15 @@ export function annotateMutation(
...
-
     return {
-        ...deepClone,
+        ...mutation,
```
alisman, line 296: How do we handle typing of these extended mutations?   Should there be something called AnnotatedMutation?

onursumer: I don't think we introduce new fields when annotating a mutation. We just update/override existing `Mutation` model fields with annotated values. In this function we just set default (empty) values for those fields.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
alisman: i think we already have a memoize decorator.  what is the "weak" indicate?   also, why not just use computed?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
```diff
@@ -226,15 +227,32 @@ class DefaultMutationMapperStore implements MutationMapperStore {
...
-              )
-            : this.getMutations();
+        return getMutationsByTranscriptId(
+            mutations,
+            transcriptId,
+            indexedVariantAnnotations
+        );
+    }
+
+    public getAnnotatedMutations(transcriptId: string): Mutation[] {
```
alisman, line 246: lets take this opportunity to comment this code.  like it doesn't make a lot of sense to me.  if indexedVariantAnnotations is not complete yet, then we're just going to return un-annotated mutations?  that doesn't feel right.   i would think we would not want to use mutations before we know whether or not they are annotated.   that's an argument for passing indexedVariantAnnotations.result in as an argument.

src/shared/components/mutationMapper/MutationMapperStore.ts
```diff
@@ -121,22 +115,26 @@ export default class MutationMapperStore extends DefaultMutationMapperStore {
         );
     };
 
-    public getAnnotatedMutations(transcriptId: string): SimpleMutation[] {
-        // overriding the base method (DefaultMutationMapperStore.getAnnotatedMutations)
+    @memoize
```
alisman, line 118: again curious why we don't use computed?

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: Not sure what weak indicates but as far as I can see this is the only memoize decorator we use everywhere else. See for example https://github.com/cBioPortal/cbioportal-frontend/blob/2675e99f1eafdeeb0366ff1d18161f427c0fa3af/src/shared/components/query/CancerStudySelector.tsx#L18 and https://github.com/cBioPortal/cbioportal-frontend/blob/2675e99f1eafdeeb0366ff1d18161f427c0fa3af/src/shared/components/query/QueryStore.ts#L38

The reason this function one is not computed is because we call it in an `invoke` function: https://github.com/cBioPortal/cbioportal-frontend/blob/f206b6dc2d0299db6bb3dedfdf629ae5e7df5c12/packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts#L671

Also, there is this dynamic `getMutations` function which makes things trickier.  It might be possible to refactor things to be able to use computed here.

packages/react-mutation-mapper/src/store/DefaultMutationMapperStore.ts
onursumer: Right, this might confusing. I can't really remember why we did it this way, but there might be cases where we don't have any variant annotation data. I'll try to figure out the reason and add a comment for this code.

leexgh (approved):
Transcript switch works great, thanks!

inodb:
Let's merge this

inodb (approved):

PR #3636, opened 2021-02-16
Lollipop plot improvements

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8334

(1) Render splice mutations
(2) Render driver vs vus colors

[Link](https://www.cbioportal.org/results/mutations?cancer_study_list=ccle_broad_2019&tab_index=tab_visualize&case_set_id=ccle_broad_2019_all&Action=Submit&gene_list=TP53)

### DIFF
--- a/packages/cbioportal-frontend-commons/src/index.tsx
+++ b/packages/cbioportal-frontend-commons/src/index.tsx
@@ -55,3 +55,5 @@ export * from './lib/TickUtils';
 export { default as CBIOPORTAL_VICTORY_THEME } from './theme/cBioPortalTheme';
 export * from './theme/cBioPortalTheme';
 export { default as TruncatedText } from './components/truncatedText/TruncatedText';
+
+export * from './lib/AlterationColors';
--- a/packages/cbioportal-frontend-commons/src/index.tsx
+++ b/packages/cbioportal-frontend-commons/src/index.tsx
@@ -37,6 +37,7 @@ export { default as MutationTumorTypeFrequencyTable } from './components/signal/
 export * from './components/signal/SignalHelper';
 export * from './api/remoteData';
 
+export * from './lib/AlterationColors';
 export * from './lib/ColumnVisibilityResolver';
 export * from './lib/findFirstMostCommonElt';
 export { default as getBrowserWindow } from './lib/getBrowserWindow';
@@ -55,5 +56,3 @@ export * from './lib/TickUtils';
 export { default as CBIOPORTAL_VICTORY_THEME } from './theme/cBioPortalTheme';
 export * from './theme/cBioPortalTheme';
 export { default as TruncatedText } from './components/truncatedText/TruncatedText';
-
-export * from './lib/AlterationColors';
--- a/packages/react-mutation-mapper/src/util/MutationTypeUtils.ts
+++ b/packages/react-mutation-mapper/src/util/MutationTypeUtils.ts
@@ -68,11 +68,11 @@ export function mutationTypeSort(
     }
 }
 
-export function getColorForProteinImpactType(
-    mutations: Partial<Mutation>[],
+export function getColorForProteinImpactType<T extends Mutation>(
+    mutations: Partial<T>[],
     colors: IProteinImpactTypeColors = DEFAULT_PROTEIN_IMPACT_TYPE_COLORS,
-    getMutationCount: (mutation: Partial<Mutation>) => number = () => 1,
-    isPutativeDriver?: (mutation: Partial<Mutation>) => boolean
+    getMutationCount: (mutation: Partial<T>) => number = () => 1,
+    isPutativeDriver?: (mutation: Partial<T>) => boolean
 ): string {
     const processedMutations = mutations.map(m => {
         return {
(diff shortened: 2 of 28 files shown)


packages/cbioportal-frontend-commons/src/index.tsx
```diff
@@ -55,3 +55,5 @@ export * from './lib/TickUtils';
 export { default as CBIOPORTAL_VICTORY_THEME } from './theme/cBioPortalTheme';
 export * from './theme/cBioPortalTheme';
 export { default as TruncatedText } from './components/truncatedText/TruncatedText';
+
+export * from './lib/AlterationColors';
```
onursumer, line 59: for purely OCD reasons, can we move this export right before `export * from './lib/ColumnVisibilityResolver';` (line 40)?

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -35,11 +35,13 @@ import MutationMapperDataStore, {
     PROTEIN_IMPACT_TYPE_FILTER_ID,
 } from './MutationMapperDataStore';
 import WindowStore from '../window/WindowStore';
+import { Mutation } from 'cbioportal-utils';
```
onursumer, line 38: unused import?

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -211,7 +213,14 @@ export default class MutationMapper<
                 trackVisibility={this.trackVisibility}
                 trackDataStatus={this.trackDataStatus}
                 onTrackVisibilityChange={this.onTrackVisibilityChange}
-                getLollipopColor={getColorForProteinImpactType}
+                getLollipopColor={(mutations: any) =>
```
onursumer, line 216: why do we need to make `mutations` any, here?

packages/react-mutation-mapper/src/util/MutationTypeUtils.ts
```diff
@@ -61,41 +69,66 @@ export function mutationTypeSort(
 }
 
 export function getColorForProteinImpactType(
-    mutations: Partial<Mutation>[],
+    mutations: Partial<Mutation & { putativeDriver: boolean }>[],
```
onursumer, line 72: `putativeDriver` field is cBioPortal specific, right? Instead of adding a new field to the model, should we make it more generic by passing another function to this function? 

```suggestion
    mutations: Partial<Mutation>[],
    isPutativeDriver(mutation: Partial<Mutation>) => boolean = () => false,
```

On the cBioPortal side we can pass a function like this:
```JavaScript
m => m.putativeDriver
```

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -35,11 +35,13 @@ import MutationMapperDataStore, {
     PROTEIN_IMPACT_TYPE_FILTER_ID,
 } from './MutationMapperDataStore';
 import WindowStore from '../window/WindowStore';
+import { Mutation } from 'cbioportal-utils';
 
 import styles from './mutationMapper.module.scss';
 
 export interface IMutationMapperProps {
     store: MutationMapperStore;
+    colorDriverVsVus?: boolean;
```
onursumer, line 44: should we also support this in the library level by adding this property into default mutation mapper inside `react-mutation-mapper`?

src/shared/components/mutationMapper/MutationMapper.tsx
adamabeshouse: I tried to make it more specific but, for whatever reason, theres some weird stuff going on with which `Mutation` is being used. I spent about 30 minutes trying to untangle it then just decided "if it works now then whatever, just get TS off my back"

packages/react-mutation-mapper/src/util/MutationTypeUtils.ts
adamabeshouse: sure, thats a good idea

src/shared/components/mutationMapper/MutationMapper.tsx
adamabeshouse: I have no opinion on it, whatever you think is best

src/shared/components/mutationMapper/MutationMapper.tsx
adamabeshouse: yes this was related to my failed attempts to make the typing more specific (see other response)

packages/cbioportal-frontend-commons/src/index.tsx
adamabeshouse: no problem

src/shared/components/mutationMapper/MutationMapper.tsx
onursumer: Let's keep it this way for now, we can add support in the library level when we need it.

adamabeshouse:
@onursumer updated

onursumer:
@adamabeshouse seems like a unit test related to `getColorForProteinImpactType ` is broken now.

src/shared/components/mutationMapper/MutationMapper.tsx
alisman: i think we should spend a little more time figuring it out since it's an important type.  maybe pair program?

alisman (approved):
Approved but lets see if we can figure out the typing issue

adamabeshouse:
@alisman @onursumer okay, the typing issue is fixed now

PR #3638, opened 2021-02-17
Treatments Selection Fix

Luke-Sikina:
Fix cBioPortal/cbioportal#8348

- The treatments selection methods were being passed a chart meta
that they did not expect
- The selections have been updated to expect a chart meta


kalletlak (approved):

inodb (approved):
Can we add a e2e test that makes a treatment selection and takes a screenshot? Could be in a follow up PR. It's quite critical functionality now so I think it would be good to add some regression testing

PR #3643, opened 2021-02-22
fix frequency formatting in signal table

leexgh:
Fix: https://github.com/knowledgesystems/signal/issues/107

Reason: floating-point
https://stackoverflow.com/questions/1458633/how-to-deal-with-floating-point-number-precision-in-javascript

`formatNumberValueInSignificantDigits` will handle floating-point. So to avoid this formatting issue, apply `* 100` inside `formatNumberValueInSignificantDigits`.

### DIFF
--- a/packages/cbioportal-frontend-commons/src/components/signal/SignalHelper.tsx
+++ b/packages/cbioportal-frontend-commons/src/components/signal/SignalHelper.tsx
@@ -175,12 +175,12 @@ export const FREQUENCY_COLUMNS_DEFINITION = {
                     sum + row.variantCount,
                 0
             );
-            const frequency =
-                (formatNumberValueInSignificantDigits(
-                    variantCount / tumorTypeCount,
+            const percent =
+                formatNumberValueInSignificantDigits(
+                    (variantCount * 100) / tumorTypeCount,
                     2
-                ) || 0) * 100;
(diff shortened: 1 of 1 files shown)


packages/cbioportal-frontend-commons/src/components/signal/SignalHelper.tsx
```diff
@@ -176,10 +176,10 @@ export const FREQUENCY_COLUMNS_DEFINITION = {
                 0
             );
             const frequency =
```
onursumer, line 178: this is actually percentage not frequency, right? maybe rename `frequency` as `percent`?

packages/cbioportal-frontend-commons/src/components/signal/SignalHelper.tsx
```diff
@@ -203,10 +203,10 @@ export const FREQUENCY_COLUMNS_DEFINITION = {
...
```
onursumer, line 205: same as above, rename `frequency` as `percent`?

onursumer (approved):
Looks good, minor variable renaming suggestion.

onursumer (approved):
Thanks!

onursumer:
failing e2e tests are not related, merging.

PR #3645, opened 2021-02-22
Dont show study url warning in MSKCIS setting

alisman:


src/appBootstrapper.jsx
```diff
@@ -96,6 +96,7 @@ window.FRONTEND_COMMIT = COMMIT;
 // portal header
 window.postLoadForMskCIS = function() {
     AppConfig.hide_login = true;
+    window.isMSKCIS = true;
```
alisman, line 99: kinda hacky but i don't think we want to formalize something that is already hacked in for MSK only

inodb (approved):
I can't test it, but this seems like it should work to me. Thanks for fixing!

PR #3647, opened 2021-02-23
Timeline fixes

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8310 
Fixes https://github.com/cBioPortal/cbioportal/issues/8358

image

[Link](https://cbioportal.mskcc.org/patient?studyId=metastatic_tropism_msk_2021&caseId=P-0000686)


inodb (commented):
Seems like now all the tracks are just lower case?

I think the decision was to use all the data as is from the database. So you can show whatever you get returned from the API:

Screen Shot 2021-03-03 at 6 01 40 PM

E.g. `DIAGNOSIS` and `Lung`

inodb (approved):
LGTM! Thanks!

PR #3648, opened 2021-02-23
Alterations: fix error in filtering samples

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8371


alisman (approved):

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -743,22 +743,19 @@ export default abstract class ComparisonStore {
             return Promise.resolve(
                 this.enrichmentAnalysisGroups.result!.reduce(
                     (acc: EnrichmentAnalysisComparisonGroup[], group) => {
-                        let filteredSamples: Sample[] = [];
+                        let filteredSamples: Sample[] = group.samples;
```
alisman, line 746: @kalletlak is it possible this doesn't work when we're in patient mode?  see Tali's ticket. https://github.com/cBioPortal/cbioportal/issues/8383

PR #3651, opened 2021-02-25
Re-add cancer studies bar chart visualization

oplantalech:
Reintroduce the cancer studies bar chart visualization which was removed in #3410.

### DIFF
--- a/src/shared/components/barGraph/BarGraph.tsx
+++ b/src/shared/components/barGraph/BarGraph.tsx
@@ -1,0 +1,276 @@
+import * as React from 'react';
+import * as _ from 'lodash';
+import wordwrap from 'word-wrap';
+import { ThreeBounce } from 'better-react-spinkit';
+import { CancerStudy } from 'cbioportal-ts-api-client';
+import { ChartTooltipItem } from 'chart.js';
+const convertCssColorNameToHex = require('convert-css-color-name-to-hex');
+import Chart from 'chart.js';
+
+interface ICancerTypeStudy {
+    shortName: string;
+    color: string;
+    studies: CancerStudy[];
+    caseCount?: number;
+}
+
+export interface IBarGraphProps {
+    data: ICancerTypeStudy[];
+    openStudy: (studyId: string) => void;
+}
+
+export default class BarGraph extends React.Component<IBarGraphProps, {}> {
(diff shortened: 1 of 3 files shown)


alisman:
https://github.com/cBioPortal/cbioportal-frontend/pull/3644/files

@oplantalech you'll see in above recent PR, i removed chart.js from vendor file.  Lets add it back in this PR.

oplantalech:
@alisman Done!

alisman (approved):

alisman (changes requested):
I just found that the configuration property defaults to true in our properties example file.  I have to check with JJ whether we should make a new property so as to avoid turning it on for everyone again.

Sjoerd-van-Hagen:
@alisman Hi Aaron,
I do not really understand what is happening here. For all the current installations your change will require them to update the portal properties to keep the functionality when they update. Now you are arguing that for new installations your approach is easier, but to me keeping this working in existing installations seems more important. People who set up a new installation are going to change the portal properties anyways so they can easily change the default value to something else.

To me the default behaviour should be the behaviour that has been the default for all this time, and not the new behaviour.

alisman:
@jjgao can you comment on @Sjoerd-van-Hagen's feedback.   My understanding was that we retired this feature because we didn't think it was particularly valuable.   It has a bit of negative impact on performance of homepage and requires a library we would rather not load.  But if we were wrong about its' value to some then we should reconsider.

jjgao:
@Sjoerd-van-Hagen Thanks for the feedback. I vaguely remember this was discussed in a (probably steering) call and a couple of us decided this chart can be dropped due to the reason @alisman has mentioned. If you feel strongly that this chart has great value and should be kept, we can reopen the discussion.

I do think that we should have the freedom to remove features and code if they are not providing significant product value especially when they become a technical burden (e.g. previous Network tab) - inevitably, some users will be upset... I am curious how many of your clients asked for this feature to be restored?

Sjoerd-van-Hagen:
Hi @jjgao , let me check with the client to see how important this is. Perhaps they have a use case.

Sjoerd-van-Hagen:
@jjgao @alisman Discussed this with the client, it is fine to remove.

PR #3654, opened 2021-02-26
Enrichments: Fix group cases count

kalletlak:
Fix https://github.com/cbioportal/cbioportal/issues/8386

### DIFF
--- a/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
+++ b/src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
@@ -40,7 +40,7 @@ import {
 
 export interface IAlterationEnrichmentContainerProps {
     data: AlterationEnrichmentWithQ[];
-    groups: Omit<EnrichmentAnalysisComparisonGroup, 'samples'>[];
+    groups: EnrichmentAnalysisComparisonGroup[];
     alteredVsUnalteredMode?: boolean;
     headerName: string;
     store?: ResultsViewPageStore;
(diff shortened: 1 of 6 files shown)


src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -88,6 +88,28 @@ export default class AlterationEnrichmentContainer extends React.Component<
         return this.props.groups[1];
     }
 
+    @computed get group1QueriedCasesCount() {
+        let caseIds: Set<string> = new Set(
```
alisman, line 92: you use Set only to enforce uniqueness?

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
```diff
@@ -88,6 +88,28 @@ export default class AlterationEnrichmentContainer extends React.Component<
...
+            this.group1.samples.map(sample =>
+                this.props.patientLevelEnrichments
+                    ? sample.uniquePatientKey
+                    : sample.uniqueSampleKey
+            )
+        );
+        return caseIds.size;
+    }
+
+    @computed get group2QueriedCasesCount() {
```
alisman, line 102: it seems confusing that "cases" will sometimes mean samples, right? i mean, if we're looking at sampleLevelENrichment, this count will a count of samples.  i think "case" typically means patient in our system.

alisman: i know we're racing to get this done and we already use "case" generically below.  just want to make sure i understand.

alisman (approved):

src/pages/resultsView/enrichments/AlterationEnrichmentsContainer.tsx
kalletlak: yes

PR #3656, opened 2021-02-26
Change genomic profile filtering logic to backend

kalletlak:
This improves initial page load. But it would degrade the response time/performance in the subsequent filtering on genomic profiles table


Luke-Sikina:
This is merged to a branch named ?, but is not in master. This will be complete when it is merged into master, and it would be nice to address some of the performance concerns in another ticket.

inodb:
Closing this for now

PR #3660, opened 2021-02-27
Update default values of MutationEnrichmentEventTypes

kalletlak:
Frontend changes to fix https://github.com/cbioportal/cbioportal/issues/8389

### DIFF
--- a/src/shared/lib/comparison/ComparisonStoreUtils.ts
+++ b/src/shared/lib/comparison/ComparisonStoreUtils.ts
@@ -189,7 +189,8 @@ export const cnaEventTypeSelectInit: {
 export function mutationEventTypeSelectInit() {
     const object = {} as { [key in MutationEnrichmentEventType]?: boolean };
     Object.keys(MutationEnrichmentEventType).forEach(type => {
-        object[type as MutationEnrichmentEventType] = true;
+        object[type as MutationEnrichmentEventType] =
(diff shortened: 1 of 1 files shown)


src/shared/lib/comparison/ComparisonStoreUtils.ts
```diff
@@ -189,7 +189,8 @@ export const cnaEventTypeSelectInit: {
 export function mutationEventTypeSelectInit() {
     const object = {} as { [key in MutationEnrichmentEventType]?: boolean };
     Object.keys(MutationEnrichmentEventType).forEach(type => {
-        object[type as MutationEnrichmentEventType] = true;
+        object[type as MutationEnrichmentEventType] =
```
alisman, line 192: not sure i understand this logic.  it's unselected if it's of type any?    comment would be good to explain why.  but i can merge.

kalletlak: Not exactly. `MutationEnrichmentEventType.any` is never used/updated in frontend, but its default value is set to `true`. With my recent backend api changes I am trying to leverage this option to filter data.

If `any` is set to `true` then no mutation type filtering is done. If it is `false` filtering is done based on the other mutation type selection

Similarly for `other` option, which is used in the frontend has a default value `true` - meaning to include other mutation types.

PR #3662, opened 2021-03-01
Survival prefix table - use slider for setting min patient threshold

adamabeshouse:
Also only show p-value for survival when all groups have at least 10 cases.

Fixes https://github.com/cBioPortal/cbioportal/issues/8381
8381-survival-slider


inodb (approved):
LGTM - thanks!

PR #3669, opened 2021-03-03
Do not assign default values to alteration types

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8413

### DIFF
--- a/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
+++ b/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
@@ -481,6 +481,18 @@ export function pickMutationEnrichmentProfiles(profiles: MolecularProfile[]) {
     );
 }
 
+export function pickStructuralVariantEnrichmentProfiles(
+    profiles: MolecularProfile[]
+) {
+    return _.filter(
+        profiles,
+        (profile: MolecularProfile) =>
+            profile.molecularAlterationType ===
+                AlterationTypeConstants.STRUCTURAL_VARIANT ||
+            AlterationTypeConstants.FUSION
+    );
+}
+
 export function pickCopyNumberEnrichmentProfiles(profiles: MolecularProfile[]) {
     return _.filter(
         profiles,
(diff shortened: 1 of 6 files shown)


src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -438,6 +446,30 @@ export default abstract class ComparisonStore {
                     this.molecularProfilesInActiveStudies.result!
                 )
             ),
+        onResult: profiles => {
+            if (!_.isEmpty(profiles)) {
```
alisman, line 450: this kind of seems like it is defining default values?   i think we need more comment here in order to understand what this is doing.   also,   since we may be mutation state multiple times (forEach) we should run this whole in thing inside runInAction.   

but lets see if there's any way to do this without using onResult  would componentDidMount not work?

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -481,6 +481,18 @@ export function pickMutationEnrichmentProfiles(profiles: MolecularProfile[]) {
     );
 }
 
+export function pickStructuralVariantEnrichmentProfiles(
+    profiles: MolecularProfile[]
+) {
+    return _.filter(
+        profiles,
+        (profile: MolecularProfile) =>
+            profile.molecularAlterationType ===
```
alisman, line 490: @kalletlak , problem was here.  you probably know this but you can't use OR operator like this.    has to be   
```
x == y || x === z
```

alisman: fixed in subsequent commit

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -430,23 +438,71 @@ export default abstract class ComparisonStore {
...
         invoke: () =>
             Promise.resolve(
-                pickMutationEnrichmentProfiles(
-                    this.molecularProfilesInActiveStudies.result!
-                )
+                this.alterationEnrichmentProfiles.result!.mutationProfiles
+            ),
+    });
+    //
+    public readonly structuralVariantProfiles = remoteData({
```
alisman, line 491: i could have change all referencing code to access the single promise 
```
alterationEnrichmentProfiles.whatever
```
but decided it was better to tread lightly

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
kalletlak: thanks for fixing

PR #3678, opened 2021-03-10
fiix build.sh for checking out a commit

inodb:
see attempt at updating localdb tests for backend https://github.com/cBioPortal/cbioportal/pull/8441

it's possible the error is actually further upstream in figuring out the environment parameters when running this as a backend PR


inodb:
actually this prolly doesn't work

PR #3682, opened 2021-03-15
Add signal additional columns back

leexgh:
Fix https://github.com/knowledgesystems/signal/issues/105

Adding additional signal columns back because we have new data for pathogenic variants. Please review and merge https://github.com/genome-nexus/genome-nexus-importer/pull/41 first

Will show new data after merging https://github.com/genome-nexus/genome-nexus-importer/pull/41 and update database.

image
https://www.signaldb.org/variant/17:g.7578263G%3EA


onursumer:
@leexgh Does this also add these columns for Signal feature in cBioPortal or only for Signal webpage?

leexgh:
@onursumer Good point, add to cbioportal signal feature as well. Now it should apply to both signal and cbioportal

onursumer (approved):

PR #3685, opened 2021-03-16
(KM plot) Confidence Interval and Control Limit implementation

dippindots:
Fix cBioPortal/cbioportal#8427
- Calculate standard error for all data points ([Greenwood formula)](https://www.math.wustl.edu/~sawyer/handouts/greenwood.pdf)
- Calculate 95% Confidence Interval (upper and lower) for all data points (Reference: https://github.com/cran/survival/blob/master/R/survfit.R#L307-L322)
- Control Limit
image

Tests:
Validated with survival package lung dataset: https://rdrr.io/cran/survival/man/lung.html


src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+                    nEventDict[patientSurvival.months] /
+                    (adjustedAtRisk *
+                        (adjustedAtRisk - nEventDict[patientSurvival.months]));
+                const error = estimate * estimate * (previousError + v1);
+                previousError = previousError + v1;
+                standardError = Math.sqrt(error);
+
+                // calculate confidence intervals
+                const conficenceInterval = 0.95;
+                const zValue = jStat.normal.inv(
```
dippindots, line 155: Z value for 95% confidence is around Z=1.96

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+                    0,
+                    1
+                );
+                // Reference: https://github.com/cran/survival/blob/master/R/survfit.R#L308-L316
+                const xx = estimate === 0 ? NaN : estimate;
+                const se2 = (standardError / estimate) * zValue;
+                const temp1 = Math.exp(Math.log(xx) - se2);
+                const temp2 = Math.exp(Math.log(xx) + se2);
+                low95ConfidenceInterval = temp1;
+                high95ConfidenceInterval = Math.min(temp2, 1);
```
dippindots, line 166: based on log function used in survival package: https://github.com/cran/survival/blob/master/R/survfit.R#L308-L316

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -25,6 +26,13 @@ export type GroupedScatterData = {
     [key: string]: SurvivalCurveData;
 };
 
+export type SurvivalSummary = {
+    estimate: number;
```
alisman, line 30: estimate of what?   maybe a more descriptive name.  like somethingEstimate?

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+    patientSurvivals: PatientSurvival[]
+): SurvivalSummary[] {
+    let summaries: SurvivalSummary[] = [];
     let previousEstimate: number = 1;
+    let previousMonths: number | undefined = undefined;
+    let isErrorCalculatable: boolean = false;
+    let previousError: number = 0;
+
+    // calculate number of events
+    let numberOfEvenrts: number = 0;
```
alisman, line 98: what is rts?    came case Rts

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+    const nEventDict: Dictionary<number> = _.reduce(
```
alisman, line 99: seems like you are producing "counts" so i would include the word count in naming

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: This is a estimate of [Survival Function](https://en.wikipedia.org/wiki/Survival_function), maybe change to `survivalFunctionEstimate` makes more sense?

src/pages/resultsView/survival/SurvivalUtil.tsx
alisman: if this stretch of code is logic that we've taken from somewhere else, maybe put in a helper funcion

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+                const v1 =
+                    nEventDict[patientSurvival.months] /
+                    (adjustedAtRisk *
+                        (adjustedAtRisk - nEventDict[patientSurvival.months]));
+                const error = estimate * estimate * (previousError + v1);
+                previousError = previousError + v1;
+                standardError = Math.sqrt(error);
+
+                // calculate confidence intervals
+                const conficenceInterval = 0.95;
```
alisman, line 154: typo

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -151,16 +239,55 @@ export function isNullSurvivalClinicalDataValue(value: string) {
...
+            survivalSummaries[i].high95ConfidenceInterval! <= 0.5
+        ) {
+            upperControlLimit = patientSurvivals[i].months.toFixed(2);
+        }
+    }
+
+    if (median === undefined) {
+        return 'NA';
+    } else if (
+        median !== undefined &&
```
alisman, line 271: seems like median !== undefined is already take care of in first condition

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -77,20 +85,100 @@ export const survivalClinicalDataNullValueSet = new Set([
     'na',
 ]);
 
-export function getEstimates(patientSurvivals: PatientSurvival[]): number[] {
-    let estimates: number[] = [];
+export function getSurvivalSummaries(
```
alisman, line 88: @adamabeshouse could you look at this code especially closely. thanks

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: changed to `eventCountsByMonths`

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: created a helper function

dippindots:
@cBioPortal/product Could I get your review of this feature? You can use this link: https://deploy-preview-3685--cbioportalfrontend.netlify.app/ and see the survival tab (result view page or comparison page). In the survival summary table, the last column has been updated to include the `95% CI`. Thank you!

src/pages/resultsView/survival/SurvivalUtil.spec.tsx
```diff
@@ -312,8 +3142,8 @@ describe('SurvivalUtil', () => {
 
         it('returns correct stats for the example data', () => {
             assert.deepEqual(
-                getStats(examplePatientSurvivals, exampleEstimates),
-                [6, 3, '182.19']
+                getStats(examplePatientSurvivals, exampleSurvivalSummaries),
+                [6, 3, '182.19 (63.83 - NA)']
```
adamabeshouse, line 3146: so you computed the confidence intervals using R and you're checking your code against it?

src/pages/resultsView/survival/SurvivalHelper.tsx
```diff
@@ -0,0 +1,26 @@
+import jStat from 'jStat';
+
+export type ConfidenceIntervalSet = {
+    low: number;
+    high: number;
+};
+
+export function calculateLogConfidenceIntervals(
```
adamabeshouse, line 8: what does it mean "log confidence interval"? why not "confidence interval"?

src/pages/resultsView/survival/SurvivalUtil.tsx
adamabeshouse: I don't have a prayer of analyzing this off the top of my head :) @dippindots is there a reference link that you used to write this?

dippindots: Yes, For the standard error calculation, you can see this reference: https://sphweb.bumc.bu.edu/otlt/mph-modules/bs/bs704_survival/BS704_Survival4.html, and for the confidence interval, I am using the `code from R survival package` as the reference: https://github.com/cran/survival/blob/master/R/survfit.R#L308-L316

src/pages/resultsView/survival/SurvivalHelper.tsx
dippindots: There are several ways to calculate confidence interval, I found this may answer the question about `log confidence interval`: https://stats.stackexchange.com/questions/361354/choosing-conf-type-for-survfit-in-r/361387

src/pages/resultsView/survival/SurvivalUtil.spec.tsx
dippindots: Thanks for mention this, I do checked the new dataset I put in again R package. But I checked our previous test again and found out that our previous dataset was not correct since we didn't sort the `survival data` before passing them into the function, thus we got incorrect estimates and median, I am working on fixing the tests.

dippindots: tests fixed, also compared with survival package, get the same result.

```
Call: survfit(formula = Surv(time, status) ~ 1)

      n  events  median 0.95LCL 0.95UCL 
    6.0     3.0    63.8    63.8      NA 
```

```
Call: survfit(formula = Surv(time, status) ~ 1)

   time n.risk n.event survival std.err lower 95% CI upper 95% CI
   0.09      5       1      0.8   0.179       0.5161            1
  63.83      2       1      0.4   0.297       0.0935            1
 182.19      1       1      0.0     NaN           NA           NA
```

inodb:
@dippindots I noticed for the first example query it gives NA: https://deploy-preview-3685--cbioportalfrontend.netlify.app/comparison/survival?sessionId=5cf89323e4b0ab413787436c

Screen Shot 2021-03-22 at 9 26 33 AM

dippindots:
@inodb which one you think is incorrect? I think `Primary` gives `NA` is make sense since the curve never drop down to 50%. And the `Metastasis`'s high confidence interval could be `NA` also since the curve ends near 50%, so it's likely that high confidence interval cannot reach under 50% (which will result in `NA`). I will double check the value and update at here.

dippindots:
@inodb @jjgao Just tested those data in R survival package, here is the result:
```
Call: survfit(formula = Surv(time, status) ~ 1)

      n  events  median 0.95LCL 0.95UCL 
    200      60      26      22      NA 
```

adamabeshouse (approved):
As I mentioned in stand-up, the math code looks reasonable to me, I'm sure it's correct if the results match up with R. great job!

adamabeshouse (changes requested):
actually @dippindots sorry let me ask one thing - can you add another tests against R computations where there's less data? this is to test more cases, makes it more reliable

dippindots:
Thanks @adamabeshouse, that's a good idea. I will create some small tests.

dippindots:
Converted back to draft now. I will create another test and update the screenshots before merge.

adamabeshouse (approved):

PR #3689, opened 2021-03-23
update GN api client

leexgh:
Related to: https://github.com/knowledgesystems/signal/issues/112
Update genome nexus api client


onursumer (approved):

onursumer:
API sync is failing though. Do we still need to deploy the latest Genome Nexus?

leexgh:
@onursumer yes I think so, but we might need to merge other related GN PRs first

PR #3663, opened 2021-03-01
Label 1 lollipop by default for diagrams that originally have no lollipops

jtquach1:
Fix cBioPortal/cbioportal#8319

Describe changes proposed in this pull request:
- Label 1 lollipop for lollipops on the diagram based on the best candidate that:
  - has the highest mutation count
  - is in a domain (if at all)
  - is found in the middle-most position
- Added unit tests for existing and new functions in LollipopPlotUtils

## Any screenshots or GIFs?
If this is a new visual feature please add a before/after screenshot or gif
here with e.g. [Giphy CAPTURE](https://giphy.com/apps/giphycapture) or [Peek](https://github.com/phw/peek)

#### Case 1
image
#### Case 2
image
#### Case 3
image

#### Case 1
image
#### Case 2
image
#### Case 3
image

## Links tested:
- [1](http://localhost:3000/results/mutations?cancer_study_list=metastatic_solid_tumors_mich_2017%2Cchol_msk_2018&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&profileFilter=0&case_set_id=all&gene_list=STAR&geneset_list=%20&tab_index=tab_visualize&Action=Submit&comparison_subtab=cna&mutations_gene=STAR)
- [2](http://localhost:3000/results/mutations?genetic_profile_ids_PROFILE_MUTATION_EXTENDED=msk_impact_2017_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=msk_impact_2017_cna&cancer_study_list=msk_impact_2017&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&profileFilter=0&case_set_id=msk_impact_2017_cnaseq&gene_list=EGFR&geneset_list=%20&tab_index=tab_visualize&Action=Submit&mutations_transcript_id=ENST00000420316)
- [3](http://localhost:3000/results/mutations?cancer_study_list=metastatic_solid_tumors_mich_2017%2Cchol_msk_2018&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&data_priority=0&profileFilter=0&case_set_id=all&gene_list=CIR1&geneset_list=%20&tab_index=tab_visualize&Action=Submit&comparison_subtab=cna&mutations_gene=CIR1)

PR #3690, opened 2021-03-23
fix empty lollipop specs error

leexgh:
Related to: https://github.com/knowledgesystems/signal/issues/112
Page will crash when `spec` is empty, for example: https://www.signaldb.org/gene/ABL1 (This link won't crash because it's not using the latest mutation mapper package, but will crash when using local latest package)
image
In this example, it has `somatic` and `germline`. `somatic` has data but `germline` is empty, the spec will be `[]` for germline. If no data available it shouldn't label any lollipop, but in this example - `noLabelsAreShown` will still be `true` and enter `labelOneLollipopByDefault`, and will break on `specs[candidateIndex].label!.show = true;`

### DIFF
--- a/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
+++ b/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
@@ -228,8 +228,8 @@ export default class LollipopMutationPlot<
             group,
             placement
         );
-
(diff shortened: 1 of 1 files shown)


onursumer:
@leexgh thanks for fixing this!

I still have mixed feelings about `show` being part of the specs (https://github.com/cBioPortal/cbioportal-frontend/pull/3663#discussion_r587836024) But I guess that's beyond the scope of this PR.

onursumer (approved):

Luke-Sikina (approved):

PR #3697, opened 2021-04-02
Plots tab: hide messages when there is no data

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8483

### DIFF
--- a/src/pages/resultsView/plots/PlotsTab.tsx
+++ b/src/pages/resultsView/plots/PlotsTab.tsx
@@ -468,6 +468,10 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
         );
     }
 
+    @computed get showPlot(): boolean {
+        return this.plotDataExistsForTwoAxes || this.waterfallPlotIsShown;
+    }
+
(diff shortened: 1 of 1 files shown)


adamabeshouse (approved):

src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -476,6 +476,9 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
             .length;
         const vertAxisDataSampleCount = this.vertAxisDataPromise.result!.data
             .length;
+        if (!(this.plotDataExistsForTwoAxes || this.waterfallPlotIsShown)) {
```
alisman, line 479: how about return null and a comment.  returning components is confusing for the reader (and could leader to potential issues above)

alisman:
It strikes me it would better if possible to deal with this condition higher up in components.  Like, don't show the chart wrapper at all instead of this message nested inside chart wrapper:

image

alisman (approved):

PR #3698, opened 2021-04-05
Add oncokb annotation for fusion genes

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8475

### DIFF
--- a/packages/react-mutation-mapper/src/component/column/Annotation.tsx
+++ b/packages/react-mutation-mapper/src/component/column/Annotation.tsx
@@ -32,6 +32,7 @@ import HotspotAnnotation, {
     sortValue as hotspotSortValue,
 } from './HotspotAnnotation';
 import { USE_DEFAULT_PUBLIC_INSTANCE_FOR_ONCOKB } from '../../util/DataFetcherUtils';
+import { CanonicalMutationType } from 'cbioportal-frontend-commons';
 
 export type AnnotationProps = {
     mutation?: Mutation;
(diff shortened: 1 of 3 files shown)


zhx828 (changes requested):
Can we have use a constant to make the string comparison? I believe they will all be `Fusion`?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -131,7 +132,9 @@ export function getAnnotationData(
                 _.find(
                     oncoKbCancerGenes.result,
                     (gene: CancerGene) =>
-                        gene.oncokbAnnotated &&
+                        (mutation.mutationType ===
+                            MAIN_MUTATION_TYPE_MAP['fusion'].label ||
```
onursumer, line 136: use this one instead? https://github.com/cBioPortal/cbioportal-frontend/blob/cd06ecf60e07cb8ce2af6c9d0beda5225dee20fc/packages/cbioportal-frontend-commons/src/lib/getCanonicalMutationType.ts#L16

onursumer (approved):
LGTM

zhx828 (approved):

zhx828 (commented):
I think the proper fix is to find out whether any of the gene partner in the fusion is oncokb annotated. If none of it is annotated, then skip the oncokb call. Could you check whether this could be an easy fix? If not, we can merge this pr first then do the next fix as second step.

zhx828:
My intention in the above comment is to improve the performance and avoid unnecessary call to oncokb. But after discussion with Karthik, we realize in the current production, we don't have a quick way to get both gene entrez ids in the mutations tab. So we cannot do the filtering based on that. If we filter based on hugo symbol, we might filtered out gene that use different hugo symbols in two places.
In the new implementation of SV, we can get both ids in the result's tab but not in the mutations tab.

PR #3718, opened 2021-04-20
Fix issues related to #3358

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8534
Also fix issue related to default selected profiles on query page

### DIFF
--- a/src/shared/lib/getDefaultMolecularProfiles.ts
+++ b/src/shared/lib/getDefaultMolecularProfiles.ts
@@ -1,13 +1,8 @@
 import { MolecularProfile } from 'cbioportal-ts-api-client';
 import * as _ from 'lodash';
 import { AlterationTypeConstants } from '../../pages/resultsView/ResultsViewPageStore';
-import {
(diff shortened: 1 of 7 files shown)


src/shared/lib/getDefaultMolecularProfiles.ts
```diff
@@ -123,32 +117,23 @@ export function getDefaultGeneSetProfile(profiles: MolecularProfile[]) {
 
 export function getFilteredMolecularProfiles(
     profiles: MolecularProfile[],
-    profileFilterSet?: { [profileType: string]: boolean },
+    queriedProfileSuffixes?: string[],
     dataPriority: number = 0 // this is legacy and necessary only for backward compatibility
```
alisman, line 121: @jjgao do you think we can drop support for dataPriority at this point?

kalletlak: @alisman I think we cannot. there might be old/saved urls which still uses it

PR #3719, opened 2021-04-21
Fix error downloading data for non-selected profiles

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8535


dippindots (approved):
Thanks for fixing this. Seems this pr also fixed another ticket we have: 
Fix https://github.com/cBioPortal/cbioportal/issues/8528

PR #3722, opened 2021-04-22
exclude `showNA` while checking for changes in attribute from default state

kalletlak:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3799,7 +3799,10 @@ export class StudyViewPageStore {
                     }
                     const attributeChanged =
                         isDefaultAttr &&
-                        !_.isEqual(toJS(attribute), initDataBinFilter);
+                        !_.isEqual(
```
alisman, line 3802: we can probably just merge this, but would it not be better to test for what we actually want to compare?

alisman: as in, what properties of the attribute need to be different in order for this to be regarded as changed

kalletlak: we need to compare for everything except `showNA`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6355,7 +6364,7 @@ export class StudyViewPageStore {
     }
 
     @computed private get hasSampleIdentifiersInFilter(): boolean {
-        return !(
+        return (
```
alisman, line 6367: this seems weird

adamabeshouse (approved):

inodb:
@kalletlak could you help update the description to clarify what this change does? This avoids the duplicate binning related HTTP request on study view?

kalletlak:
@inodb yes

PR #3724, opened 2021-04-27
Fetch cached gene-panel-data from s3 if available

alisman:
Try fetching gene_panel_data from s3 and go to web server if it doesn't exist

### DIFF
--- a/packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
@@ -1,5 +1,6 @@
 import getBrowserWindow from './getBrowserWindow';
 import sizeof from 'object-sizeof';
+import fileDownload from 'react-file-download';
 
 function hash(str: string) {
     var hash = 0,
@@ -97,6 +98,7 @@ export function cachePostMethod(
     targetObj[methodName] = function(arg: any) {
         //console.log('posted data',arg);
         const hash = getHash(arg);
+
         // remove this from sizeQueue, if it exists, because we just used it so it should get bumped back in line,
         //  away from deletion
         const sizeQueueIndex = sizeQueue.findIndex(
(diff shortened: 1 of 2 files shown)


packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
```diff
@@ -128,6 +129,39 @@ export function cachePostMethod(
...
+
+    if (methodName in cachedMethods) {
+        const oldMethod = targetObj[methodName];
+
+        targetObj[methodName] = function(arg: any) {
+            const hash = getHash(arg);
+            const args = arguments; // keep them locally for closure
+            const fileName = `${methodName}-${hash}.json`;
+
+            return $.get(
```
alisman, line 151: i got sick of wrestling with fetch api which drives me nuts.

packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
```diff
@@ -142,13 +176,23 @@ export function cachePostMethodsOnClient(
...
             obj.prototype,
             methodName,
             apiCacheLimit,
             log,
             sentryLog
-        )
-    );
+        );
+
+        if (window.location.hostname.includes('genie-private')) {
```
alisman, line 188: we can expand this as necessary.  obviously is this temporary solution

adamabeshouse (approved):
interesting

packages/cbioportal-frontend-commons/src/lib/apiClientCache.ts
inodb: can we include `genie.cbioportal.org` as well? For `genie_public`

inodb (changes requested):
seeing some strange errors after adding some filters: https://bit.ly/3gQyJyW
 
Screen Shot 2021-04-27 at 4 18 47 PM

EDIT: looks like this is by design (a 403 will result in calling the actual endpoint instead of getting the data from S3)

PR #3728, opened 2021-04-30
Remove makeObservable call from ProteinImpactTypeBadgeSelector

onursumer:
When trying to update to the latest `react-mutation-mapper` this isse causes `Signal` to crash in dev mode (and potentially in prod mode).

## Checks

### DIFF
--- a/packages/react-mutation-mapper/src/component/filter/ProteinImpactTypeBadgeSelector.tsx
+++ b/packages/react-mutation-mapper/src/component/filter/ProteinImpactTypeBadgeSelector.tsx
@@ -1,6 +1,4 @@
 import { Option, ProteinImpactType } from 'cbioportal-frontend-commons';
-import _ from 'lodash';
-import { computed, makeObservable } from 'mobx';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 import { CSSProperties } from 'react';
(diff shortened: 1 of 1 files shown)


packages/react-mutation-mapper/src/component/filter/ProteinImpactTypeBadgeSelector.tsx
```diff
@@ -63,8 +61,8 @@ export class ProteinImpactTypeBadgeSelector<
 > extends React.Component<P, {}> {
     constructor(props: any) {
         super(props);
-        makeObservable(this);
```
inodb, line 66: @onursumer don't think i fully get what's the issue here

inodb: but if that fixes it - sounds good to me :)

inodb (approved):

onursumer:
Error message for future reference:

```
Error: [MobX] No annotations were passed to makeObservable, but no decorator members have been found either
▶ 2 stack frames were collapsed.
new ProteinImpactTypeBadgeSelector
/Users/onur/IdeaProjects/cbioportal-frontend/packages/react-mutation-mapper/dist/index.es.js:3655
  3652 | __extends(ProteinImpactTypeBadgeSelector, _super);
  3653 | function ProteinImpactTypeBadgeSelector(props) {
  3654 |     var _this = _super.call(this, props) || this;
> 3655 |     makeObservable(_this);
  3656 |     return _this;
  3657 | }
  3658 | Object.defineProperty(ProteinImpactTypeBadgeSelector.prototype, "optionDisplayValueMap", {
View compiled
▶ 23 stack frames were collapsed.
```

PR #3730, opened 2021-05-03
Remove gene panel data calls for mutations tab

inodb:
Should only be mutation rate really that's relying on gene panel data but there were a bunch of different things that had the same dependency (e.g. SVs and some warning related to showing only profiled cases).

Note: this also disables a lot of the other tabs. This is not really meant for review but pure for hackathon example

See also https://docs.google.com/presentation/d/16yEuaNJu0mmvxOvF481402lDtUjHtER_s6HL9pDRrEg/edit#slide=id.gd6f8e9ff63_0_0

Further ideas:

- only need clinical sample information (cancer type) for samples with mutations. Don't need detailed samples projection

    Screen Shot 2021-05-03 at 5 13 46 PM

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -3275,9 +3275,7 @@ export class ResultsViewPageStore {
         await: () => [
             this.selectedMolecularProfiles,
             this.defaultOQLQuery,
-            this.mutationsReportByGene,
-            this.filteredSampleKeyToSample,
-            this.structuralVariantsReportByGene,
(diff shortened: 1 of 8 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3571,6 +3581,7 @@ export class ResultsViewPageStore {
                     }
                 );
                 let promises: Promise<Sample[]>[] = [];
+                debugger;
```
alisman, line 3584: debugger

inodb:
superseded by #3746

PR #3731, opened 2021-05-03
Performance improvements for initial oncoprint load

adamabeshouse:
### Results

Genie query with EGFR loading time goes from ~56s to ~46s, and without the backwards-progress-in-loading phenomenon

### Details 
1) Fix issue that was causing unnecessary delay on initial oncoprint load, loading data that isn't necessary yet as per default mutation annotation settings

2) Dont do unnecessary data copying in OQL filter code, which was costly for large data 

3) Optimize annotation with cosmic and cbio count - turn O(n^2) into O(n) for some parts of the code

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -4796,13 +4796,16 @@ export class ResultsViewPageStore {
                     getOncoKbMutationAnnotationForOncoprint(mutation);
 
                 const isHotspotDriver =
+                    this.driverAnnotationSettings.hotspots &&
                     !(this.isHotspotForOncoprint.result instanceof Error) &&
                     this.isHotspotForOncoprint.result!(mutation);
                 const cbioportalCountExceeded =
+                    this.driverAnnotationSettings.cbioportalCount &&
                     this.getCBioportalCount.isComplete &&
                     this.getCBioportalCount.result!(mutation) >=
                         this.driverAnnotationSettings.cbioportalCountThreshold;
                 const cosmicCountExceeded =
+                    this.driverAnnotationSettings.cosmicCount &&
                     this.getCosmicCount.isComplete &&
                     this.getCosmicCount.result!(mutation) >=
                         this.driverAnnotationSettings.cosmicCountThreshold;
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -255,6 +255,7 @@ import {
     ANNOTATED_PROTEIN_IMPACT_FILTER_TYPE,
     createAnnotatedProteinImpactTypeFilter,
 } from 'shared/lib/MutationUtils';
+import ComplexKeyCounter from 'shared/lib/complexKeyDataStructures/ComplexKeyCounter';
 
 type Optional<T> =
     | { isApplicable: true; value: T }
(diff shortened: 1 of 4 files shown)


packages/cbioportal-utils/src/mutation/ProteinChangeUtils.ts
```diff
@@ -12,46 +14,48 @@ const PROTEIN_CHANGE_MATCHERS = [
  *
  * @return  protein location
  */
-export function getProteinPositionFromProteinChange(
-    proteinChange?: string
-): { start: number; end: number } | undefined {
-    if (!proteinChange) {
-        return undefined;
-    }
+export const getProteinPositionFromProteinChange = memoize({
```
inodb, line 17: @adamabeshouse just curious: why does oncoprint use `getProteinPositionFromProteinChange`? Is it for annotation of OncoKB or something?

inodb: feels slightly funky to me to `memoize` this function but not fix the calling logic (i.e. why would this be called multiple times for the same protein changes)? Seems to indicate it might be calling other functions multiple times too for the same mutation? Unless all the speed up comes from mutations that have different genes but the same protein change (this is prolly not the case, because I believe you were testing on a single gene query?)

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4798,13 +4798,16 @@ export class ResultsViewPageStore {
...
+                    this.driverAnnotationSettings.hotspots &&
                     !(this.isHotspotForOncoprint.result instanceof Error) &&
                     this.isHotspotForOncoprint.result!(mutation);
                 const cbioportalCountExceeded =
+                    this.driverAnnotationSettings.cbioportalCount &&
                     this.getCBioportalCount.isComplete &&
                     this.getCBioportalCount.result!(mutation) >=
                         this.driverAnnotationSettings.cbioportalCountThreshold;
                 const cosmicCountExceeded =
+                    this.driverAnnotationSettings.cosmicCount &&
```
inodb, line 4810: so this is mostly to get the speed up when driver annotations by cosmic are disabled right (i guess that's the default)

packages/cbioportal-utils/src/mutation/ProteinChangeUtils.ts
adamabeshouse: @inodb check this out https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/ResultsViewPageStore.ts#L5215

This function is used to get the protein position of a COSMIC mutation record, for comparison with a given mutation for annotating. the COSMIC mutation records are grouped by keyword, so a mutation is checked based on its keyword and then the COSMIC records are looped through. So this would happen repeatedly for mutations with the same keyword (which, in the case of truncating, are not that descriptive).

This discussion has just made me realize how inefficient this process actually is, though. Maybe there is an algorithmic optimization to make instead of memoizing this.

src/pages/resultsView/ResultsViewPageStore.ts
adamabeshouse: right. we shouldnt be using this if cosmic is disabled (this is also, I think, the cause of the issue where the loading goes backwards - its because were using a promise in `invoke` that wasn't in `await`

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -5153,27 +5164,16 @@ export class ResultsViewPageStore {
...
-                } else {
-                    return -1;
-                }
+                return this.cbioportalMutationCountData.result![key] || -1;
             });
         },
     });
     //COSMIC count
-    readonly cosmicCountData = remoteData<CosmicMutation[]>({
+    readonly cosmicCountData = remoteData<ComplexKeyCounter>({
```
alisman, line 5174: i think maybe this name should reflect it's somewhat involved typing better?   so we can more easily understand what the nature of the complex key is?

adamabeshouse: you mean like `indexedCosmicCounts`? `cosmicCountsByKeywordAndPosition`?

alisman: right

alisman (approved):

inodb (approved):
LGTM from product perspective (i tried cosmic driver annotations with different counts)

It would be nice to add some info in the description how much this actually improves performance. E.g. querying EGFR on mskimpact: x time is spent on loading, y time was spent on oncoprint processing which is now z

PR #3732, opened 2021-05-04
proto cache

Luke-Sikina:
DO NOT MERGE


alisman:
I would perhaps be in favor of keeping this on the downlow.  Not updating the documentation.  I think it should be possible without breaking types

PR #3733, opened 2021-05-04
Adjustments to loading indicators in results view for smoother initial load

adamabeshouse:


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -703,10 +704,8 @@ export default class ResultsViewPage extends React.Component<
                                 {// we don't show the result tabs if we don't have valid query
                                 this.showTabs &&
                                     !this.resultsViewPageStore.genesInvalid &&
-                                    !this.resultsViewPageStore.isQueryInvalid &&
-                                    this.resultsViewPageStore
```
alisman, line 707: I think @pvannierop added this for a reason.  We talked about moving it deeper but ...

alisman: it means we load all mutation data before we show tabs

adamabeshouse: why do we want to do that? the tabs handle their own loading state/indicators

alisman: @pvannierop lets talk about whether we can put this further down or not.   but @adamabeshouse maybe we just need to add a loader when these conditions aren't true?

alisman: @adamabeshouse i agree.  i know this came up before but i can't remember why it ended up here.

src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -609,6 +609,13 @@ export default class ResultsViewPage extends React.Component<
                 </div>
             );
         } else {
+            // we don't show the result tabs if we don't have valid query
+            const tabsReady =
```
alisman, line 613: i'm gonna merge this right now, but why don't we put this in showTabs?

alisman (approved):

PR #3735, opened 2021-05-05
Remove year related attributes from comparison

inodb:
Fix https://github.com/cBioPortal/cbioportal/issues/8445. Decided to
hardcode removal of a few GENIE attributes. The stats team mentioned
that it does not make sense to compare on these year related attributes.
This removes them for all studies, but that probably makes sense anyway
given their year related names.

Before:
Screen Shot 2021-05-05 at 3 48 39 PM

After:
Screen Shot 2021-05-05 at 3 48 05 PM

Note - not entirely sure if we should also exclude `GENIE Sequence year` and `Age at sequencing`


inodb:
Test failures are unrelated - this is a pretty innocuous and urgent request so gonna merge. Can follow up with improvements

PR #3737, opened 2021-05-06
optimize filtering process in mutations tab

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8576

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -3616,20 +3616,18 @@ export class ResultsViewPageStore {
 
     readonly germlineConsentedSamples = remoteData<SampleIdentifier[]>(
         {
-            await: () => [this.studyIds, this.samples],
+            await: () => [this.studyIds, this.sampleMap],
             invoke: async () => {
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3616,20 +3616,18 @@ export class ResultsViewPageStore {
...
-                const sampleIds = this.samples.result
-                    ? this.samples.result.map(s => s.sampleId)
-                    : [];
 
                 // do not simply return all germline consented samples,
                 // only include the ones matching current sample selection
+                const sampleMap = this.sampleMap.result!;
                 return germlineConsentedSamples.filter(s =>
-                    sampleIds.includes(s.sampleId)
+                    sampleMap.has(s, ['sampleId', 'studyId'])
```
alisman, line 3630: @adamabeshouse  don't samples have unique id? in which case we could halve this lookup time?   i realize this would be a small optimization compared to what you just did

adamabeshouse: `germlineConsentedSamples` dont have the unique id attached to them, but this lookup time is still just one level of indexing

PR #3747, opened 2021-05-17
Fetch and filter all samples from a study when using a custom case list

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8593

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -256,6 +256,7 @@ import {
     createAnnotatedProteinImpactTypeFilter,
 } from 'shared/lib/MutationUtils';
 import ComplexKeyCounter from 'shared/lib/complexKeyDataStructures/ComplexKeyCounter';
+import SampleSet from 'shared/lib/sampleDataStructures/SampleSet';
 
 type Optional<T> =
     | { isApplicable: true; value: T }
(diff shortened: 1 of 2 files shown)


Luke-Sikina (approved):

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3647,32 +3648,45 @@ export class ResultsViewPageStore {
...
+                                    sampleListIds: customSampleListStudyIds.map(
+                                        studyId => `${studyId}_all`
+                                    ),
+                                } as SampleFilter,
+                                projection:
+                                    REQUEST_ARG_ENUM.PROJECTION_DETAILED,
+                            })
+                            .then(samples => {
+                                return samples.filter(s =>
+                                    customSampleListIds.has(s, [
```
alisman, line 3684: this could be a fuckload of samples, i take it that .has is pretty optimal?

alisman: i always forget, why don't these guys have unique ids that we made on server specifically for this reason?  maybe we stopped doing that?

adamabeshouse: they do, it's that the custom list specification is specified by sample id and study id, not by unique id, so we won't have them here for filtering

adamabeshouse: Im going to do benchmarking to see how fast it is compared to a more specific solution for this

alisman (approved):

PR #3748, opened 2021-05-17
Fix VAF in genomic evolution tooltip

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8453


inodb (approved):
Thanks! Confirmed that this fixes the bug 
Screen Shot 2021-05-17 at 8 02 02 PM

Note that it would be nice to show the actual read counts as well (can be in follow up PR)

Since the read coverage can vary a lot between positions and samples, it is useful to see the numerator/denominator

inodb:
failing tests unrelated

PR #3751, opened 2021-05-18
Introduce batch utility

alisman:
Introduce utility for batching calls to a service (multiple simultaneous calls beats single big call)

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -102,6 +102,7 @@ import {
     fetchStructuralVariantOncoKbData,
     cancerTypeForOncoKb,
     getOncoKbOncogenic,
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3105,12 +3106,24 @@ export class ResultsViewPageStore {
                 sampleMolecularIdentifiers: filters,
             } as MutationMultipleStudyFilter;
 
-            return await client.fetchMutationsInMultipleMolecularProfilesUsingPOST(
+            const mutations = await batchCalls<Mutation>(
+                client,
+                'fetchMutationsInMultipleMolecularProfilesUsingPOST',
```
alisman, line 3111: maybe ```(client:T)=>cliient.fetchM...``` ?

PR #3752, opened 2021-05-18
chunkCalls method to chunk large arguments to API calls

adamabeshouse:


src/shared/lib/StoreUtils.ts
```diff
@@ -120,6 +120,18 @@ export const MolecularAlterationType_filenameSuffix: {
     STRUCTURAL_VARIANT: 'structural_variants',
 };
 
+export async function chunkCalls<ParamData, ResponseData>(
```
alisman, line 123: we need to move this out to utils package because it may need to be used by other packages that fetch their own data

src/shared/lib/StoreUtils.ts
```diff
@@ -942,12 +943,14 @@ export async function queryOncoKbData(
         'id'
     );
 
-    const mutationQueryResult =
-        mutationQueryVariants.length === 0
-            ? []
-            : await client.annotateMutationsByProteinChangePostUsingPOST_1({
-                  body: mutationQueryVariants,
-              });
+    const mutationQueryResult: IndicatorQueryResp[] = await chunkCalls(
```
alisman, line 946: ok, good. thanks

PR #3758, opened 2021-05-20
Add VAF numerator and denominator to VAF chart tooltip

adamabeshouse:
image


inodb (commented):
Thanks for adding this! Some minor tweaks:

- [x] It might be nice to spell it out i.e. `(xx reads of yy total)` instead of `(x/y)`. Same as in the mutations table
- [x] looks like it doesn't work for the heatmap:

    Screen Shot 2021-05-21 at 4 46 49 PM

adamabeshouse:
@inodb thanks! updated now

inodb (approved):
Neat - thanks!!

PR #3759, opened 2021-05-21
Fetch gene panel data by molecular profile ids

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8603

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -7,10 +7,11 @@ import {
     ClinicalDataMultiStudyFilter,
     ClinicalDataSingleStudyFilter,
     CopyNumberSeg,
-    CosmicMutation,
     DiscreteCopyNumberData,
     DiscreteCopyNumberFilter,
     Gene,
+    GenePanelData,
+    GenePanelDataMultipleStudyFilter,
     GenericAssayData,
     GenericAssayDataMultipleStudyFilter,
     GenericAssayMeta,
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2034,7 +2029,7 @@ export class ResultsViewPageStore {
         },
     });
 
-    readonly coverageInformation = remoteData<CoverageInformation>(
+    /*readonly coverageInformation = remoteData<CoverageInformation>(
```
adamabeshouse, line 2032: oops I forgot to delete this

src/shared/lib/GenePanelUtils.ts
```diff
@@ -143,58 +144,52 @@ export function computeGenePanelInformation(
...
-                    molecularProfileId: profile.molecularProfileId,
-                    sampleId,
-                });
-            }
-        }
-    });
-    // query for gene panel data using sample molecular identifiers
-    let genePanelData: GenePanelData[];
-    if (sampleMolecularIdentifiers.length && genes.result!.length) {
+    // fetch all gene panel data for profiles
```
alisman, line 152: amazing commentry (commentary?)
thank you

alisman: commenting

alisman:
@adamabeshouse now that we fetch by profile, i believe we should be able to fire the fetch earlier in the "timeline" because we no longer need to wait for samples.    this could potentially shave off seconds from load time of page.  could you investigate if this is true?    we would pass this data into getCoverageInformation

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -2053,7 +2048,22 @@ export class ResultsViewPageStore {
...
+    readonly coverageInformation = remoteData<CoverageInformation>({
+        await: () => [
+            this.molecularProfilesInStudies,
+            this.sampleKeyToSample,
+            this.patients,
+            this.genes,
+        ],
+        invoke: () =>
+            getCoverageInformation(
+                this.molecularProfilesInStudies.result!,
```
kalletlak, line 2061: Do we need gene-panel data for all molecular profiles in studies or just the selected profiles?

adamabeshouse: good catch, yeah I think you're right

adamabeshouse:
@kalletlak @alisman updated

kalletlak (approved):

PR #3761, opened 2021-05-21
Remove dependency on clinical data in results page initial load

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8621


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -5124,7 +5142,7 @@ export class ResultsViewPageStore {
                             structuralVariant.site2EntrezGeneId,
                             cancerTypeForOncoKb(
                                 structuralVariant.uniqueSampleKey,
-                                this.uniqueSampleKeyToTumorType.result || {}
+                                {}
```
alisman, line 5145: could we eliminate this paremeter on this function?

alisman (approved):

PR #3763, opened 2021-05-25
Get clinvar from GN annotation endpoint

leexgh:
Part of:https://github.com/cBioPortal/cbioportal/issues/8470

- Get clinvar from genome nexus annotation endpoint.
- Be consistent with clinvar website - show one clinvar interpretation(unless it's conflicting interpretation). Previously was multiple interpretations and evidences in "Germline" and "Somatic"
- Update clinvar on Signal

Previous:
image

Now:
image

### DIFF
--- a/packages/react-mutation-mapper/src/component/column/ClinVar.tsx
+++ b/packages/react-mutation-mapper/src/component/column/ClinVar.tsx
@@ -1,18 +1,11 @@
-import {
-    Clinvar,
-    MyVariantInfo,
-    VariantAnnotation,
-} from 'genome-nexus-ts-api-client';
+import { Clinvar, VariantAnnotation } from 'genome-nexus-ts-api-client';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
 import { defaultSortMethod, Mutation, RemoteData } from 'cbioportal-utils';
 import ClinVarSummary from '../clinvar/ClinVarSummary';
-import {
-    MyVariantInfoProps,
-    renderMyVariantInfoContent,
-} from './MyVariantInfoHelper';
 import { getClinVarData } from './ClinVarHelper';
+import { errorIcon, loaderIcon } from '../StatusHelpers';
 
 type ClinVarProps = {
     mutation: Mutation;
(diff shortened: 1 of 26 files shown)


leexgh:
There will be some tests failed because genome nexus call is slow, after solving slowness it should pass

packages/react-mutation-mapper/src/component/column/ClinVar.tsx
```diff
@@ -1,41 +1,66 @@
...
-    getRcvCountMap,
-    getRcvData,
-} from '../clinvar/ClinVarSummary';
-import {
-    MyVariantInfoProps,
-    renderMyVariantInfoContent,
-} from './MyVariantInfoHelper';
+import { defaultSortMethod, Mutation, RemoteData } from 'cbioportal-utils';
+import ClinvarSummary from '../clinvar/ClinVarSummary';
+import { getClinvarData } from './ClinVarHelper';
```
onursumer, line 7: Should we also rename files as `ClinvarSummary` and `ClinvarHelper` for consistency?

packages/react-mutation-mapper/src/component/dbsnp/DbsnpId.tsx
```diff
@@ -4,12 +4,12 @@ import { MyVariantInfo } from 'genome-nexus-ts-api-client';
 import { observer } from 'mobx-react';
 import * as React from 'react';
 
-export type ClinVarIdProps = {
+export type DbsnpIdProps = {
```
onursumer, line 7: Good catch :)

packages/react-variant-view/src/component/pathogenicity/ClinVarInterpretation.tsx
```diff
@@ -2,82 +2,88 @@ import * as React from 'react';
...
+                    ClinVar Interpretation
+                </a>
+            </DefaultTooltip>
+        );
     }
 
-    @computed get rcvDisplayData() {
-        if (this.rcvCountMap) {
-            return getRcvData(this.rcvCountMap);
+    @computed get clinvarData() {
```
onursumer, line 37: The name `clinvarData` is a bit confusing because we are returning a `div` here. Let's find another name for this if possible.

onursumer (approved):
Looks good, just a few naming suggestions. We should also make sure that the changes here do not break the Signal website.

inodb (approved):
Looks good to me! Thanks for pushing this over the finish line!

PR #3764, opened 2021-05-26
Fix Patient View PathwayMapper node tooltips

msalihaltun:
Fix https://github.com/cBioPortal/cbioportal/issues/8527.

Changes tooltip libraries used by PathwayMapper as mentioned [here](https://github.com/cBioPortal/cbioportal/issues/8527#issuecomment-847781063).

### DIFF
--- a/package.json
+++ b/package.json
@@ -213,7 +213,7 @@
     "oncoprintjs": "4.2.4",
     "pako": "2.0.2",
     "parameter-validator": "^1.0.2",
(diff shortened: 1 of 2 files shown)


onursumer (approved):
Looks good, thanks! Result view pathways e2e tests are failing but that's a known issue which can be addressed in another PR (cBioPortal/cbioportal#8622). I think it is okay to merge this one.

PR #3765, opened 2021-05-26
Added "Deselect all" button: Resolved Issue #8541

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8541

Changes:
- adds a "Deselect all" button to the study view which is always there unless no charts are selected

No tests have been created since the change is simple enough to be tested by manual inspection.

**Before:**
before

**After:**
after


src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -351,37 +351,48 @@ export default class FixedHeaderTable<T> extends React.Component<
...
-            onClick = this.onAddAll;
-            content = `Select all${
+        let selectAllContent: string;
+        let showDeselectAll = false;
+        let showSelectAll = false;
+
+        if (!noneSelected && this.props.removeAll) {
+            showDeselectAll = true;
+        }
+        if (!allSelected && this.props.addAll) {
```
Luke-Sikina, line 365: We can clean this up a bit by leveraging JS's truthieness a bit earlier

 ```suggestion
        const showDeselectAll = !noneSelected && !!this.props.removeAll;
        const showSelectAll = !allSelected && !!this.props.addAll;

       
        if (showSelectAll) {
```

File: src/pages/studyView/table/FixedHeaderTable.tsx [hunk 2]
```diff
@@ -351,37 +351,48 @@ export default class FixedHeaderTable<T> extends React.Component<
...
+            showSelectAll = true;
+            selectAllContent = `Select all${
                 this.props.showSelectableNumber
                     ? ` (${this._store.dataStore.sortedFilteredData.length})`
                     : ''
```
Luke-Sikina, line 370: I appreciate the sentiment of only calculating this string when needed, but I don't think it's necessary. The performance savings is negligible, and I'd rather not spend time thinking when that string is or isn't computed. Let's move this logic to the declaration of `selectAllContent` and make that variable a const

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -351,37 +351,48 @@ export default class FixedHeaderTable<T> extends React.Component<
...
             }`;
-            showButton = true;
         }
 
-        if (showButton) {
+        if (showDeselectAll || showSelectAll) {
```
Luke-Sikina, line 374: Pretty sure it's impossible for this to be false. You'd have to have a study view with 0 potential charts. If that happens, frankly, I want something to blow up. Let's get rid of the `if` and the null return block.

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -351,37 +351,48 @@ export default class FixedHeaderTable<T> extends React.Component<
...
+                        </button>
+                    )}
+                    {showDeselectAll && (
+                        <button
+                            className="btn btn-default btn-xs"
+                            data-test={'fixed-header-table-remove-all'}
+                            onClick={this.onRemoveAll}
+                        >
+                            {'Deselect all'}
+                        </button>
```
Luke-Sikina, line 393: This all looks correct and easy to read; nice job.

Luke-Sikina (approved):

alisman (approved):

alisman:
@Rajat-Sirohi @Luke-Sikina looks like there's a single legit broken screenshot.  @Rajat-Sirohi, luke can show how you how to update

https://100983-66571349-gh.circle-artifacts.com/0/imageCompare.html

PR #3767, opened 2021-05-27
display driver annotation setting button when hover legend panel

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8560
May-27-2021 18-25-27


alisman:
@leexgh i notied the existing settings button doesn't have a tooltip either.  lets add one?

leexgh:
@alisman Yes we could add one, but personally I feel we have too many tooltips, and that icon is obvious enough to tell people it's a setting button, so I'm not sure if it's necessary. But let's see what other people think. 
Related ticket: https://github.com/cBioPortal/cbioportal/issues/8651

src/pages/resultsView/mutation/DriverAnnotationProteinImpactTypeBadgeSelector.tsx
```diff
@@ -220,9 +235,17 @@ export default class DriverAnnotationProteinImpactTypeBadgeSelector extends Prot
...
+        e.stopPropagation(); // Prevent click being applied to parent element
+        this.settingMenuVisible = !this.settingMenuVisible;
+        this.props.onClickSettingMenu &&
+            this.props.onClickSettingMenu(this.settingMenuVisible);
+    }
+
     public render() {
         return (
-            <div>
+            <div className="legend-panel">
```
alisman, line 248: this is annoying but we should use css modules approach here b/c legend-panel is too general.

alisman: that will involve converting mutations.scss to mutations.module.scss below.

src/pages/resultsView/mutation/mutations.scss
```diff
@@ -3,3 +3,11 @@
         margin-right: 5px;
     }
 }
+
```
alisman, line 6: both of these are too general because they could appear elsewhere and we wouldn't want this behavior.   the best thing to do would be to convert this into a css module.

alisman:
@leexgh i think our current tooltip is too clunky.  we need something more discreet for small bits of text.  but in general i don't think tooltips are problematic.

alisman (approved):

PR #3769, opened 2021-05-28
Load molecular data using profileIds instead of samples ids

alisman:
In the past we waited until samples loaded to fetch molecular data.  Instead, use profileIds to load molecular data and then filter using samples. In this way, we load in parallel instead of serially.

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1304,44 +1304,42 @@ export class ResultsViewPageStore {
     );
 
     readonly molecularData = remoteData<NumericGeneMolecularData[]>({
-        await: () => [
-            this.studyToDataQueryFilter,
-            this.genes,
-            this.selectedMolecularProfiles,
-            this.samples,
-        ],
+        await: () => [this.samples, this.molecularData_preload],
         invoke: () => {
-            // we get mutations with mutations endpoint, structural variants and fusions with structural variant endpoint, generic assay with generic assay endpoint.
-            // filter out mutation genetic profile and structural variant profiles and generic assay profiles
-            const profilesWithoutMutationProfile = excludeSpecialMolecularProfiles(
-                this.selectedMolecularProfiles.result!
+            return Promise.resolve(
+                this.molecularData_preload.result.filter(
+                    m => m.uniqueSampleKey in this.sampleKeyToSample
+                )
             );
-            const genes = this.genes.result;
+        },
+    });
 
-            if (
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1304,61 +1304,66 @@ export class ResultsViewPageStore {
...
-            this.selectedMolecularProfiles,
-            this.samples,
-        ],
+        await: () => [this.samples, this.molecularData_preload],
         invoke: () => {
-            // we get mutations with mutations endpoint, structural variants and fusions with structural variant endpoint, generic assay with generic assay endpoint.
-            // filter out mutation genetic profile and structural variant profiles and generic assay profiles
-            const profilesWithoutMutationProfile = excludeSpecialMolecularProfiles(
-                this.selectedMolecularProfiles.result!
+            const samplesByUniqueKey = _.keyBy(
```
adamabeshouse, line 1309: we already have this in the store - sampleKeyToSample

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1304,61 +1304,66 @@ export class ResultsViewPageStore {
...
-                                });
-                            }
-                        });
-                    }
+    // loading molecular data based on profileId (instead of individual samples) provides
+    // two optimizations.
+    // 1. we can load this data before we know samples
+    // 2. backend can cache based on finite set of profiles
+    // we then have to filter this using samples, which can be loaded concurrently instead of serially
+    readonly molecularData_preload = remoteData<NumericGeneMolecularData[]>(
```
adamabeshouse, line 1326: How about `molecularDataForSelectedProfiles`? to make it same as how it is for gene panel data

alisman: (I actually had to undo that optimization (molecularDataForSelectedProfiles) because in fact we need the data for ALL profiles and there was a test failing for coexpression.)

Do you mean molecularDataForAllProfiles_preload and molecularDataForAllProfiles?  I want to use the suffix because I don't believe this data should be used directly as it is not filtered by sample selection.

adamabeshouse: ah okay, that makes sense. for the other one, that would be a one line fix to make the preload use all molecular profiles instead of selected molecular profiles

kalletlak (approved):
LGTM. one comment what Adam specified on `samplesByUniqueKey`

PR #3773, opened 2021-06-01
Remove app_name configuration checks

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8419:
Adds explicit booleans for configuration checks instead of indirectly checking `app_name` property.
(Need to merge corresponding backend PR#8656 before this.)

### DIFF
--- a/src/appShell/App/PortalFooter.tsx
+++ b/src/appShell/App/PortalFooter.tsx
@@ -156,8 +156,8 @@ export default class PortalFooter extends React.Component<
                                     </If>
                                     <If
                                         condition={
-                                            AppConfig.serverConfig.app_name ===
-                                            'public-portal'
+                                            AppConfig.serverConfig
+                                                .skin_right_nav_show_twitter
                                         }
                                     >
                                         <li>
@@ -174,8 +174,7 @@ export default class PortalFooter extends React.Component<
                         </If>
                         <If
                             condition={
-                                AppConfig.serverConfig.app_name ===
-                                'public-portal'
+                                AppConfig.serverConfig.skin_footer_show_dev
                             }
                         >
                             <div className="footer-elem">
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -77,6 +77,7 @@ export interface IServerConfig {
     skin_examples_right_column_html: string | null;
     skin_documentation_faq: string | null;
     skin_footer: string | null;
+    skin_footer_show_dev: boolean;
     skin_login_contact_html: string | null;
     skin_login_saml_registration_html: string | null;
     skin_citation_rule_text: string | null;
@@ -88,6 +89,7 @@ export interface IServerConfig {
     skin_right_nav_show_examples: boolean;
     skin_right_nav_show_testimonials: boolean;
     skin_right_nav_show_whats_new: boolean;
+    skin_right_nav_show_twitter: boolean;
     skin_right_nav_whats_new_blurb: string | null;
     skin_show_about_tab: boolean;
     skin_show_data_tab: boolean;
--- a/src/config/serverConfigDefaults.ts
+++ b/src/config/serverConfigDefaults.ts
@@ -63,6 +63,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_documentation_markdown: true,
     skin_email_contact: 'cbioportal at googlegroups dot com',
     skin_documentation_faq: 'FAQ.md',
+    skin_footer_show_dev: false,
     skin_login_saml_registration_html: 'Sign in with MSK',
     skin_documentation_news: 'News.md',
     skin_documentation_oql: 'Onco-Query-Language.md',
@@ -71,6 +72,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_right_nav_show_examples: true,
     skin_right_nav_show_testimonials: true,
     skin_right_nav_show_whats_new: true,
+    skin_right_nav_show_twitter: false,
     skin_citation_rule_text:
         'Please cite: <a href="http://cancerdiscovery.aacrjournals.org/content/2/5/401.abstract" target="_blank">Cerami et al., 2012</a> &amp; <a href="http://www.ncbi.nlm.nih.gov/pubmed/23550210" target="_blank">Gao et al., 2013</a>',
     skin_show_about_tab: true,


src/config/IAppConfig.ts
```diff
@@ -26,6 +26,9 @@ export type StudyView = {
 
 export interface IServerConfig {
     app_name: string | null;
+    app_is_public_portal: boolean;
```
alisman, line 29: @Rajat-Sirohi @inodb @Luke-Sikina this doesn't feel right to me.

alisman: I think maybe what we should do is hide this yucky stuff behind feature toggle type thing.  So, there's a module FeatureToggles which has methods e.g. showPersistenceWarnings and we can then do whatever we want in there. For example read the url or whatnot.    If it's a real feature then we could of course add it properly to backend configuration

alisman (changes requested):
see comments

src/pages/staticPages/datasetView/DatasetPage.tsx
```diff
@@ -21,7 +21,7 @@ export class DatasetPageStore {
     });
 
     readonly downloadList = remoteData(() => {
-        if (AppConfig.serverConfig.app_name === 'public-portal') {
+        if (AppConfig.serverConfig.app_is_public_portal) {
```
Luke-Sikina, line 24:
```suggestion
        if (AppConfig.serverConfig.show_study_download_list) {
```

src/appShell/App/PortalFooter.tsx
```diff
@@ -156,8 +156,8 @@ export default class PortalFooter extends React.Component<
                                     </If>
                                     <If
                                         condition={
-                                            AppConfig.serverConfig.app_name ===
-                                            'public-portal'
+                                            AppConfig.serverConfig
+                                                .app_is_public_portal
```
Luke-Sikina, line 160:
```suggestion
                                            AppConfig.serverConfig
                                                .show_twitter
```

src/appShell/App/PortalFooter.tsx
```diff
@@ -174,8 +174,7 @@ export default class PortalFooter extends React.Component<
                         </If>
                         <If
                             condition={
-                                AppConfig.serverConfig.app_name ===
-                                'public-portal'
+                                AppConfig.serverConfig.app_is_public_portal
```
Luke-Sikina, line 177:
```suggestion
                                AppConfig.serverConfig.show_dev_channels
```

src/appShell/App/usageAgreements/GenieAgreement.tsx
```diff
@@ -8,12 +8,8 @@ export const GENIE_PERSISTENCE_KEY = 'genie-private-usage-agreement';
 export function shouldShowGenieWarning() {
     // we want to show a warning message on private cbioportal instances
     // to prevent users from adding links in manuscripts
-    const showGenieWarning = ['cbioportal-genie-private'].includes(
-        AppConfig.serverConfig.app_name!
-    );
-
     return (
-        showGenieWarning &&
+        AppConfig.serverConfig.app_is_private_portal &&
```
Luke-Sikina, line 12: Let's not change this one. The `cbioportal-genie-private` name is unique, and this code should always be specific to that server.

Luke-Sikina (commented):
I don't think I was clear enough about how I wanted this done. I left a few suggestions for turning the logic you wrote into the feature toggles I was expecting. Those changes will propagate to the the two frontend portal properties fields, the example file in the backend, and the jsp, so make sure you make the changes there too.

src/appShell/App/Container.tsx
```diff
@@ -36,10 +36,7 @@ interface IContainerProps {
 
 function isLocalDBServer() {
     try {
-        return (
-            (window as any).frontendConfig.serverConfig.app_name ===
-            'localdbe2e'
-        );
+        return (window as any).frontendConfig.serverConfig.is_localdb_server;
```
inodb, line 39: in this particular case i don't think using app_name is too bad (kind of unlikely that people will use the same name). This parameter is maybe cleaner but in that case we also want to set `is_localdb_server` when we are running the localdb tests. It seems like more work than it's worth so maybe ok to leave as is

src/appShell/App/PortalFooter.tsx
```diff
@@ -156,8 +156,7 @@ export default class PortalFooter extends React.Component<
                                     </If>
                                     <If
                                         condition={
-                                            AppConfig.serverConfig.app_name ===
-                                            'public-portal'
+                                            AppConfig.serverConfig.show_twitter
```
inodb, line 159: maybe `skin.right_nav.show_twitter`? See: https://docs.cbioportal.org/2.3-customization/portal.properties-reference#hide-sections-in-the-right-navigation-bar

src/appShell/App/PortalFooter.tsx
```diff
@@ -173,10 +172,7 @@ export default class PortalFooter extends React.Component<
                             </div>
                         </If>
                         <If
-                            condition={
-                                AppConfig.serverConfig.app_name ===
-                                'public-portal'
-                            }
+                            condition={AppConfig.serverConfig.show_dev_channels}
```
inodb, line 175: maybe good to namespace this property a bit more, e.g. `skin.footer_show_dev`

src/appShell/App/usageAgreements/StudyAgreement.tsx
```diff
@@ -11,13 +11,9 @@ export function shouldShowStudyViewWarning() {
...
     // we don't want to show this in MSK CIS setting (iframe)
-    const showStudyViewWarning =
-        ['triage-portal', 'mskcc-portal'].includes(
-            AppConfig.serverConfig.app_name!
-        ) && !getBrowserWindow().isMSKCIS;
-
     return (
-        showStudyViewWarning &&
+        AppConfig.serverConfig.is_private_portal &&
+        !getBrowserWindow().isMSKCIS &&
```
inodb, line 16: Maybe ok leave as it was, `is_private_portal` is pretty similar to using `app_name === mskcc-portal`. If we would want to add a prop for something that describes this it would maybe something like `show_mskcc_study_agreement_banner`. Just don't really see a use case for anyone outside of MSKCC and we would have to change the prop again if we'd generalize it, so maybe better to leave as it was

src/shared/lib/StoreUtils.ts
```diff
@@ -231,7 +231,7 @@ export async function fetchAllReferenceGenomeGenes(
     genomeName: string,
     client: CBioPortalAPI = defaultClient
 ) {
-    if (AppConfig.serverConfig.app_name === 'public-portal') {
+    if (AppConfig.serverConfig.show_reference_genome_genes) {
```
inodb, line 234: this is some temp performance feature, maybe `feature_flag.use_static_reference_genome_genes`?

src/pages/staticPages/datasetView/DatasetPage.tsx
```diff
@@ -21,7 +21,7 @@ export class DatasetPageStore {
     });
 
     readonly downloadList = remoteData(() => {
-        if (AppConfig.serverConfig.app_name === 'public-portal') {
+        if (AppConfig.serverConfig.show_study_download_list) {
```
inodb, line 24: this one is maybe ok to leave as it was as well, not sure what exactly to call this prop. It's like some way to only list studies that are also on datahub (https://github.com/cBioPortal/datahub/tree/master/public). Not really something anyone else would use

inodb (commented):
Thanks so much! This was helpful to see what properties are still not really described

I think it might make sense to only add a handful of these instead of all for a variety of reasons (see inline comments)

Thanks!

src/shared/lib/StoreUtils.ts
Rajat-Sirohi: It looks like this line was replaced in a new commit since I last updated it. I will just remove my flag altogether.

inodb (approved):
Great - thank you 🙏 !

PR #3778, opened 2021-06-02
Tooltip should not pop up on top left first

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8652

### DIFF
--- a/packages/react-variant-view/src/component/variantView/Variant.scss
+++ b/packages/react-variant-view/src/component/variantView/Variant.scss
@@ -61,7 +61,7 @@ h5 {
 }
 
 .rc-tooltip {
-    opacity: 1 !important;
+    opacity: 1;
 }
 
 .page-section {


alisman (approved):

onursumer (changes requested):
@leexgh I don't think this fixes cBioPortal/cbioportal#8652 entirely. We should not have global css references in this package. We should aim to make Variant.css a module if possible. Currently we have styling for `body`, `h1`, `h2`, etc. These ones are especially problematic:

https://github.com/cBioPortal/cbioportal-frontend/blob/3cdbcc1a90c6464bb51f76ddfbb863173de7d924/packages/react-variant-view/src/component/variantView/Variant.scss#L49-L61

Ideally, we should identify the ones we really need for the variant view and remove the rest.

onursumer:
Superseded by #3780

PR #3779, opened 2021-06-02
Fix url for Cancer Digital Slide Archive iframe

alisman:
CDSA changed api.  Need to pass patientid in hash

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -311,8 +311,15 @@ export function getG2SApiUrl() {
 export function getDigitalSlideArchiveMetaUrl(patientId: string) {
     return AppConfig.serverConfig.digital_slide_archive_meta_url + patientId;
 }
-export function getDigitalSlideArchiveIFrameUrl(patientId: string) {
-    return AppConfig.serverConfig.digital_slide_archive_iframe_url + patientId;
+export function getDigitalSlideArchiveIFrameUrl(
+    patientId: string,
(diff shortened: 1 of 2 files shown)


src/shared/api/urls.ts
```diff
@@ -311,8 +311,15 @@ export function getG2SApiUrl() {
...
+export function getDigitalSlideArchiveIFrameUrl(
+    patientId: string,
+    studyId: string
+) {
+    const cancertype = studyId.split('_')[0];
+    return (
+        AppConfig.serverConfig.digital_slide_archive_iframe_url +
+        `#!/CDSA/${cancertype}/${patientId}`
+    );
 }
```
inodb, line 323: this only works for TCGA studies, kinda forget where we are checking that but i tested a non TCGA study and it correctly hides that case.

There might be some for which there is a mismatch, I found this one:

https://deploy-preview-3779--cbioportalfrontend.netlify.app/patient/tissueImage?studyId=coadread_tcga_pan_can_atlas_2018&caseId=TCGA-3L-AA1B

coadread -> coad

inodb (approved):
Thanks for this quick fix! I think it's better than the current state, so good to go imo

Would be nice to double check that it works for all tcga and tcga pancan studies. I only found coadread so far to not be not working

inodb:
I checked for all pancan studies I found another one, so:

- coadread
- laml

PR #3781, opened 2021-06-03
Query for mutations by molecular profile id instaead of sample ids

adamabeshouse:


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1307,9 +1307,10 @@ export class ResultsViewPageStore {
     readonly molecularData = remoteData<NumericGeneMolecularData[]>({
         await: () => [this.sampleKeyToSample, this.molecularData_preload],
         invoke: () => {
+            const sampleKeys = this.sampleKeyToSample.result!;
```
adamabeshouse, line 1310: I realized that since this is potentially called many times it would be better not to repeatedly access it

alisman: good catch

alisman (approved):

PR #3783, opened 2021-06-04
Fix group comparison download

dippindots:
Fix cBioPortal/cbioportal#8485

`logRatio` can be undefined, return `-` when it's undefined.
- logRatio Undefined case
image

**Testing:**
1. Go to https://genie.cbioportal.org/comparison/alterations?sessionId=60664081e4b0242bd5d46323
2. Set `localStorage.netlify ="deploy-preview-3783--cbioportalfrontend"` in browser console.
3. Click download button
image

### DIFF
--- a/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
+++ b/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
@@ -612,7 +612,7 @@ export function getAlterationEnrichmentColumns(
             ),
             sortBy: (d: AlterationEnrichmentRow) => Number(d.logRatio),
             download: (d: AlterationEnrichmentRow) =>
-                formatLogOddsRatio(d.logRatio!),
+                d.logRatio ? formatLogOddsRatio(d.logRatio) : '-',
         });
 
         enrichedGroupColum.tooltip = (


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -612,7 +612,7 @@ export function getAlterationEnrichmentColumns(
             ),
             sortBy: (d: AlterationEnrichmentRow) => Number(d.logRatio),
             download: (d: AlterationEnrichmentRow) =>
-                formatLogOddsRatio(d.logRatio!),
+                d.logRatio ? formatLogOddsRatio(d.logRatio) : '-',
```
alisman, line 615: @dippindots is the "-" what we use elsewhere for empty?

alisman (approved):

src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
dippindots: @alisman Not sure if we use `-` globally. In this case, we use `-` in render method for empty: https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/resultsView/enrichments/EnrichmentsUtil.tsx#L605

PR #3784, opened 2021-06-04
Use localdb cache (localforage) for caching ref genome data

alisman:
Even when we pull this from CDN it still often takes seconds to download.

### DIFF
--- a/src/config/serverConfigDefaults.ts
+++ b/src/config/serverConfigDefaults.ts
@@ -175,6 +175,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     saml_logout_local: false,
     patient_view_use_legacy_timeline: false,
     enable_request_body_gzip_compression: false,
+
+    referenceGenomeVersion: 'hg19',
 };
 
 export default ServerConfigDefaults;
--- a/src/shared/lib/StoreUtils.ts
+++ b/src/shared/lib/StoreUtils.ts
@@ -1,5 +1,6 @@
 import * as _ from 'lodash';
 import $ from 'jquery';
+import localForage from 'localforage';
 import {
     fetchVariantAnnotationsByMutation as fetchDefaultVariantAnnotationsByMutation,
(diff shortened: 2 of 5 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -232,19 +233,38 @@ export async function fetchAllReferenceGenomeGenes(
...
-        return $.ajax({
-            url: getFrontendAssetUrl('reactapp/reference_genome_hg19.json'),
-            dataType: 'json',
-        });
-    }
-    if (genomeName) {
-        return await internalClient.getAllReferenceGenomeGenesUsingGET({
-            genomeName: genomeName,
-        });
+    const doCaching = window.location.hostname.includes('.cbioportal.org');
```
alisman, line 236: shrug emoji

src/config/serverConfigDefaults.ts
```diff
@@ -175,6 +175,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     saml_logout_local: false,
     patient_view_use_legacy_timeline: false,
     enable_request_body_gzip_compression: false,
+
+    referenceGenomeVersion: 'hg19',
```
inodb, line 179: seems kinda weird to have this property, since one portal can have multiple reference genomes

inodb:
@alisman - I don't fully follow why we need to store this in localStorage? Wouldn't built in browser caching solve this?

alisman:
@inodb i think netlify blows away the browser cache every time we deploy.    unless it's smart enough to actually diff files.  if i'm right there, then it basically means people load this data about once a day.  sometimes, in my experiemnts, even from cdn, it takes a long time to load.

alisman:
@inodb https://www.netlify.com/blog/2017/02/23/better-living-through-caching/

I'm pretty sure the etag is specific to release, not to each inidivuual file

alisman:
@inodb a little more research makes me think maybe they do actually change EACH file's etag ONLY when they detect a change to it.

PR #3786, opened 2021-06-04
make legend selector working as a checkbox

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8572
Jun-04-2021 14-29-00

### DIFF
--- a/src/shared/components/mutationMapper/MutationMapper.tsx
+++ b/src/shared/components/mutationMapper/MutationMapper.tsx
@@ -90,6 +90,8 @@ export default class MutationMapper<
         makeObservable(this);
     }
 
+    private driverVusVusSelectorComponent: DriverAnnotationProteinImpactTypeBadgeSelector | null = null;
+
     protected legendColorCodes = (
         <div style={{ maxWidth: 700, marginTop: 5 }}>
             <strong style={{ color: '#2153AA' }}>Color Codes</strong>
@@ -346,6 +348,7 @@ export default class MutationMapper<
                         }}
                     >
                         <DriverAnnotationProteinImpactTypeBadgeSelector
+                            ref={this.driverSelectorComponentRef}
                             filter={this.proteinImpactTypeFilter}
                             counts={this.mutationCountsByProteinImpactType}
                             onSelect={this.onProteinImpactTypeSelect}
@@ -534,7 +537,10 @@ export default class MutationMapper<
 
         return (
             <FilterResetPanel
-                resetFilters={() => dataStore.resetFilters()}
+                resetFilters={() => {
+                    dataStore.resetFilters();
+                    this.driverVusVusSelectorComponent?.reset();
+                }}
                 filterInfo={`Showing ${dataStore.tableData.length} of ${dataStore.allData.length} mutations.`}
                 className={classnames(
                     'alert',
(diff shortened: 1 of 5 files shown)


src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -90,6 +90,8 @@ export default class MutationMapper<
         makeObservable(this);
     }
 
+    private driverVusVusSelectorComponent: DriverAnnotationProteinImpactTypeBadgeSelector | null = null;
```
onursumer, line 93: I guess there is a typo here:

```suggestion
    private driverVsVusSelectorComponent: DriverAnnotationProteinImpactTypeBadgeSelector | null = null;
```

But actually, I would prefer to handle driver vs vus logic in the filter applier and derive everything from the filter itself. We wouldn't need a reference like this in that case.

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -534,7 +537,10 @@ export default class MutationMapper<
 
         return (
             <FilterResetPanel
-                resetFilters={() => dataStore.resetFilters()}
+                resetFilters={() => {
+                    dataStore.resetFilters();
+                    this.driverVusVusSelectorComponent?.reset();
```
onursumer, line 542: Ideally we should override the default filter applier for mutation type and handle driver vs vus logic in the filter applier function, too. We can then derive everything from the filter itself. That would eliminate the need for a custom reset function.

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -534,7 +539,9 @@ export default class MutationMapper<
 
         return (
             <FilterResetPanel
-                resetFilters={() => dataStore.resetFilters()}
+                resetFilters={() => {
+                    dataStore.resetFilters();
+                }}
```
onursumer, line 544: Let's revert this to a single line?

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -350,6 +350,11 @@ export default class MutationMapper<
                             counts={this.mutationCountsByProteinImpactType}
                             onSelect={this.onProteinImpactTypeSelect}
                             onClickSettingMenu={this.props.onClickSettingMenu}
+                            annotatedProteinImpactTypeFilter={this.store.dataStore.dataFilters.find(
+                                filter =>
+                                    filter.type ===
+                                    ANNOTATED_PROTEIN_IMPACT_FILTER_TYPE
+                            )}
```
onursumer, line 357: Might be better to add a `@computed get annotatedProteinImpactTypeFilter` for this.

onursumer (commented):
Thank you! Just added a couple of minor comments.

onursumer:
The build is failing on circleci. To fix it we need to import this function from `cbioportal-utils` instead of `react-mutation-mapper`
https://github.com/cBioPortal/cbioportal-frontend/blob/3ac14994735f3a1429a87b5b32e36debb578517e/src/pages/patientView/structuralVariant/column/AnnotationColumnFormatter.tsx#L10

onursumer (approved):
🙌

onursumer:
Failing e2e tests are not related, merging.

PR #3787, opened 2021-06-04
force browser caching on reference genome data

alisman:
For browser to use caching on get request for genome data.

### DIFF
--- a/src/shared/lib/StoreUtils.ts
+++ b/src/shared/lib/StoreUtils.ts
@@ -232,11 +232,11 @@ export async function fetchAllReferenceGenomeGenes(
     genomeName: string,
     client: CBioPortalAPI = defaultClient
 ) {
-    if (AppConfig.serverConfig.app_name === 'public-portal') {
-        // this is temporary
+    if (/\.cbioportal\.org|\.mskcc\.org/.test(window.location.hostname)) {
         return $.ajax({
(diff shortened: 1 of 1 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -232,11 +232,12 @@ export async function fetchAllReferenceGenomeGenes(
     genomeName: string,
     client: CBioPortalAPI = defaultClient
 ) {
-    if (AppConfig.serverConfig.app_name === 'public-portal') {
+    if (window.location.hostname.includes('.cbioportal.org')) {
```
inodb, line 235: should we also include `.mskcc.org`?

src/shared/lib/StoreUtils.ts
```diff
@@ -232,11 +232,12 @@ export async function fetchAllReferenceGenomeGenes(
...
         // this is temporary
         return $.ajax({
             url: getFrontendAssetUrl('reactapp/reference_genome_hg19.json'),
             dataType: 'json',
+            cache: true,
```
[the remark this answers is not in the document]
inodb: gotcha - i guess most of the time all our servers are pointed to netlify anyway so that works

inodb (approved):

src/shared/lib/StoreUtils.ts
alisman: yes

PR #3791, opened 2021-06-07
Use localdb cache (localforage) for caching ref genome data

alisman:


src/shared/lib/StoreUtils.ts
```diff
@@ -232,19 +233,38 @@ export async function fetchAllReferenceGenomeGenes(
...
-            url: getFrontendAssetUrl('reactapp/reference_genome_hg19.json'),
-            dataType: 'json',
-            cache: true,
-        });
-    }
-    if (genomeName) {
-        return await internalClient.getAllReferenceGenomeGenesUsingGET({
-            genomeName: genomeName,
-        });
+    const doCaching = window.location.hostname.includes('.cbioportal.org');
```
inodb, line 236: @alisman prolly we want this for msk portal as well

src/config/serverConfigDefaults.ts
```diff
@@ -175,6 +175,8 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     saml_logout_local: false,
     patient_view_use_legacy_timeline: false,
     enable_request_body_gzip_compression: false,
+
+    referenceGenomeVersion: 'hg19',
```
inodb, line 179: @alisman still seems like a strange property since one portal can support multiple reference genomes - can we name this slightly differently? Like maybe more indicative that it's a cache key

PR #3793, opened 2021-06-09
Add Clinical Attributes to Mutations Tab in Results View

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8657 + first two of cBioPortal/cbioportal#8716

# Changes:
- Add clinical attributes to columns selection on Mutations Tab in Results View (loaded lazily)
- Replace react-bootstrap drop-down menu with custom one from `pages/studyView/.../AddChartByType`
  (only replaces drop-down menu for Mutations Tab in Results View, not all Mutation Tables)
- Minor fixes/refactoring on 'Deselect all' button in `FixedHeaderTable.tsx`

Currently, neither the group comparison nor custom charts clinical attributes are included, since they need individualized methods of fetching the clinical data and testing them proves difficult without a login. This should be addressed in a later PR.

# Old drop-down menu:
old_dropdown

# New drop-down menu:
new_mutations_tab
new_clinical_tab


src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -62,8 +74,30 @@ export default class ResultsViewMutationTable extends MutationTable<
             MutationTableColumnType.DBSNP,
             MutationTableColumnType.SIGNAL,
         ],
+        columnVisibilityProps: {},
     };
 
+    componentDidMount() {
+        this.props.columnVisibilityProps!.customDropdown = (
```
onursumer, line 81: Is it possible to do this in the render cycle instead of `componentDidMount`? So, somewhere something like this: 

```JavaScript
columnVisibilityProps={
    ...this.props.columnVisibilityProps,
    customDropdown: ( ... )
}
```

onursumer (commented):
Thanks for the refactoring! Added a minor comment about the custom dropdown component.

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
Rajat-Sirohi: I'm unable to change `columnVisibilityProps` since props are readonly. I've moved this to the constructor instead of `componentDidMount`. I don't think I can/would want to put it in the render cycle since it only passes the dropdown component once and then the rendering is handled in that component itself. Please let me know if I'm misunderstanding your request.

onursumer: Defining and passing all properties of the Column Visibility component at a single place would be ideal but I guess it is tricky to do it that way with the current design. I think it is okay to modify the props object in the constructor for now.

src/pages/resultsView/mutation/AddColumns.tsx
```diff
@@ -0,0 +1,265 @@
...
+                key: col.id,
+                label: col.name,
+                selected: col.visible,
+            }));
+    }
+
+    @computed get mutationsTabContent() {
+        return (
+            <div>
+                {this.mutationsOptions && (
```
Luke-Sikina, line 114: Do we really want to add the tab if `this.mutationsOptions` is `[]`?

src/pages/resultsView/mutation/AddColumns.tsx
```diff
@@ -0,0 +1,265 @@
...
+                key: col.id,
+                label: col.name,
+                selected: col.visible,
+            }));
+    }
+
+    @computed get clinicalTabContent() {
+        return (
+            <div>
+                {this.props.columnVisibility && this.clinicalOptions && (
```
Luke-Sikina, line 149: Same question here

src/pages/resultsView/mutation/AddColumns.tsx
```diff
@@ -0,0 +1,265 @@
...
+                </span>
+            </div>
+        );
+    }
+
+    private getTextPixel(text: string | undefined, fontSize: string) {
+        // This is a very specified function to calculate the text length in Add Tracks dropdown
+        const FRONT_FAMILY = 'Helvetica Neue';
+        return Math.floor(
+            getTextWidth(text ? text : '', FRONT_FAMILY, fontSize)
```
Luke-Sikina, line 200:
```suggestion
        return Math.floor(
            getTextWidth(text ? text : '', 'Helvetica Neue', fontSize)
```

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -98,7 +162,9 @@ export default class ResultsViewMutationTable extends MutationTable<
         // order columns
         this._columns[MutationTableColumnType.STUDY].order = 0;
         this._columns[MutationTableColumnType.SAMPLE_ID].order = 10;
-        this._columns[MutationTableColumnType.CANCER_TYPE].order = 15;
+        if ('CANCER_TYPE' in this._columns)
+            this._columns['CANCER_TYPE'].order = 14;
```
Luke-Sikina, line 166: {

src/shared/cache/ClinicalAttributeCache.ts
```diff
@@ -0,0 +1,82 @@
+import client from '../api/cbioportalClientInstance';
```
Luke-Sikina, line 1: This would be a big change, so don't make it until @alisman comments on this too. 

I think you can simplify this cache a lot. Keep in mind, every request you make on the frontend is already cached until you refresh the page. When the same request is made, the cached version is used. If you're curious how this works, you can look at `apiClientCache` and `config.ts`.

With this in mind, I think what I would do is make a method that takes in a request for a bunch of different clinical data types, splits that request into 1 request per clinical data type, makes those requests and then merges the result. This way each clinical data request is cached on its own by the apiClientCache.

We can talk about how to design this on Monday if Aaron agrees it's a good idea.

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -1157,10 +1168,11 @@ export default class MutationTable<
         };
     }
 
-    @computed protected get orderedColumns(): MutationTableColumnType[] {
-        const columns = (this.props.columns || []) as MutationTableColumnType[];
+    @computed protected get orderedColumns(): MutationTableColumnTypes[] {
+        const columns = (this.props.columns ||
+            []) as MutationTableColumnTypes[];
```
Luke-Sikina, line 1173: Don't think you need this cast.
```suggestion
        const columns = (this.props.columns || []);
```

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -364,26 +364,28 @@ export default class FixedHeaderTable<T> extends React.Component<
         const showDeselectAll = !noneSelected && this.props.removeAll;
 
         return (
-            <>
+            <div style={{ overflow: 'auto', width: '50%' }}>
```
alisman, line 367: what's reason for this 50%?    i would like to see the input on right side.  you can achieve this with flexbox styles

src/pages/studyView/table/FixedHeaderTable.tsx
```diff
@@ -364,26 +364,28 @@ export default class FixedHeaderTable<T> extends React.Component<
...
                         {selectAllContent!}
                     </button>
                 )}
                 {showDeselectAll && (
                     <button
                         className="btn btn-default btn-xs"
-                        data-test={'fixed-header-table-remove-all'}
+                        data-test="fixed-header-table-remove-all"
                         onClick={this.onRemoveAll}
+                        style={{ float: 'left', maxWidth: '85px' }}
```
alisman, line 383: we should use flexbox instead of float

src/shared/cache/ClinicalAttributeCache.ts
alisman: @onursumer can you weight in?   i can't totally recall how all this works.  if we are allowing sort by these column fields, would we not need to fetch everything up front?

src/shared/cache/ClinicalAttributeCache.ts
```diff
@@ -0,0 +1,82 @@
...
+            attributeIds: sampleQueries.map(
+                q => q.clinicalAttribute.clinicalAttributeId
+            ),
+            identifiers: sampleQueries.map(q => ({
+                entityId: q.entityId,
+                studyId: q.studyId,
+            })),
+        };
+
+        let patientData = Promise.resolve(new Array<ClinicalData>());
```
alisman, line 52: you should use async/await (fiunction is already designated async).   that way you could do 
```
const patientData = [];

if (whatever) {
   patientData = await client.fetch()
}

```

src/shared/cache/ClinicalAttributeCache.ts
```diff
@@ -0,0 +1,82 @@
...
+            });
+        }
+        if (sampleQueries.length > 0) {
+            sampleData = client.fetchClinicalDataUsingPOST({
+                clinicalDataType: 'SAMPLE',
+                clinicalDataMultiStudyFilter: sampleFilter,
+            });
+        }
+        return Promise.all([patientData, sampleData]).then(both =>
+            both[0].concat(both[1])
```
alisman, line 67: i think this is flatten?

src/shared/cache/ClinicalAttributeCache.ts
```diff
@@ -0,0 +1,82 @@
...
+        );
+    } else {
+        return Promise.resolve([]);
```
alisman, line 70: synice function is asyc, you can just return [].   no promise resolution necessary

src/shared/cache/ClinicalAttributeCache.ts
```diff
@@ -0,0 +1,82 @@
...
+            ),
+            identifiers: sampleQueries.map(q => ({
+                entityId: q.entityId,
+                studyId: q.studyId,
+            })),
+        };
+
+        let patientData = Promise.resolve(new Array<ClinicalData>());
+        let sampleData = Promise.resolve(new Array<ClinicalData>());
+        if (patientQueries.length > 0) {
```
alisman, line 54: what i would just do is 
```
const promiseArr = [];

if (patientQueries.length) {
   promiseArr.push(await client.fetch)
}
... again for sampleData

return Promise.all(promiseArr).then(d=>_.flatten(d))
```

src/shared/cache/ClinicalAttributeCache.ts
onursumer: As far as I remember, we support delaying data fetch until the user decides to sort by a specific column. If we really want to lazy load clinical attribute columns by page, then we need to have this cache. 

Also we need to make sure that download functionality is working as expected for the clinical data columns. Currently data seems to be empty in the download file for not yet loaded data. I don't remember exactly where/how but we need to update some download data fetcher to make sure that all the data is fetched before starting the data download.

inodb:
@Rajat-Sirohi Thanks for implementing this! Is it correct that the filtering with the search field doesn't currently work?

E.g.:

Screen Shot 2021-06-21 at 10 27 33 AM

Rajat-Sirohi:
@inodb Currently, since clinical attribute datatypes are not always strings, I have not implemented any filter for them. If you like, I can attach a filter for those columns with string datatypes?

inodb:
@Rajat-Sirohi if it's easy to do it for all string types it would be a nice enhancement

inodb:
@Rajat-Sirohi i know you'll be adding filtering per column as well in the future, which would address it as well. So only if it's easy to add (CC @Luke-Sikina thoughts?)

src/shared/cache/ClinicalAttributeCache.ts
Luke-Sikina: Aaron and I chatted and came to the conclusion that @Rajat-Sirohi's current solution is the best way forward. Resolving this issue.

Luke-Sikina:
> @Rajat-Sirohi i know you'll be adding filtering per column as well in the future, which would address it as well. So only if it's easy to add (CC @Luke-Sikina thoughts?)

I think Nikki mentioned that he wanted to keep the global filtering, so we should definitely make it work for the string typed columns

onursumer:
> @inodb Currently, since clinical attribute datatypes are not always strings, I have not implemented any filter for them. If you like, I can attach a filter for those columns with string datatypes?

We need to make sure that search is functional for not-yet-loaded clinical data. I can't remember the current behavior for lazy loaded columns, but it may not work as desired by default.

Rajat-Sirohi:
@onursumer Thanks for the heads-up! Luckily, it looks like filtering via search automatically traverses all the pages in the table, which activates the respective fetch calls for the currently visible lazy loaded columns. So no need to change anything.

src/pages/studyView/addChartButton/addChartByType/AddChartByType.tsx
```diff
@@ -114,7 +115,7 @@ export default class AddChartByType extends React.Component<
     }
 
     @computed get columns() {
-        const columns: Column<AddChartOption>[] = [
+        let columns: Column<AddChartOption>[] = [
```
onursumer, line 118: I don't think we are assigning something else to `columns`, let's keep it `const`.

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -179,6 +179,7 @@ export enum MutationTableColumnType {
     GENE_PANEL,
     SIGNAL,
 }
+type MutationTableColumnTypes = MutationTableColumnType | string;
```
onursumer, line 182: Having both `MutationTableColumnType` and `MutationTableColumnTypes` is a bit confusing. Maybe call the new type `ExtendedMutationTableColumnType`?

onursumer (commented):
Looks good, thanks! Just added a few more minor comments.

src/pages/resultsView/mutation/AddColumns.tsx
```diff
@@ -0,0 +1,263 @@
...
+import { observer } from 'mobx-react';
+import { action, computed, makeObservable, observable } from 'mobx';
+import { Checkbox } from 'react-bootstrap';
+import { remoteData, getTextWidth } from 'cbioportal-frontend-commons';
+import { ClinicalAttribute } from 'cbioportal-ts-api-client';
+import { IColumnVisibilityControlsProps } from 'shared/components/columnVisibilityControls/ColumnVisibilityControls';
+import { MSKTab, MSKTabs } from 'shared/components/MSKTabs/MSKTabs';
+import CustomDropdown from 'shared/components/oncoprint/controls/CustomDropdown';
+import AddChartByType from '../../studyView/addChartButton/addChartByType/AddChartByType';
+import { ChartDataCountSet } from '../../studyView/StudyViewUtils';
```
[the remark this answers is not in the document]
Rajat-Sirohi: When attempting to move `AddChartByType` to shared components, it quickly became apparent that this would be a recursive nightmare, since it itself imports things from `studyView`, which in turn import other things from `studyView` and so on. I think it should be doable, but not without major refactoring. Given that `oncoprint` in `resultsView` already imports the same things from `studyView`, I agree with postponing this to a later PR.

inodb:
Failing tests seem unrelated: https://app.circleci.com/pipelines/github/cBioPortal/cbioportal-frontend/5099/workflows/e5e694fb-7b7b-4b0b-bacd-6f2b28eb964a/jobs/103269/artifacts. Somehow the heatmaps tab started disappearing, so that's the reason why quite a few screenshots are failing

inodb:
Thanks so much @Rajat-Sirohi ! Really nice work 🥇 !

PR #3806, opened 2021-06-16
Genomic evolution fixes - VAF Chart, Group By issues

dianab0:
If samples are grouped by a clinical value, when adding data points for samples without mutations there must be a check that they are in the same group as the other samples.

For example on https://www.cbioportal.org/patient/genomicEvolution?studyId=msk_impact_2017&caseId=P-0000377

Before this fix, the highlighted mutation below contains only samples 2 and 4, but sample 3 was also added even if it is in a different group
Screenshot from 2021-06-16 11-33-20

After this fix, sample 3 is no longer added because it is part of a different group
Screenshot from 2021-06-16 11-37-48

### DIFF
--- a/src/pages/patientView/timeline2/VAFChartUtils.ts
+++ b/src/pages/patientView/timeline2/VAFChartUtils.ts
@@ -71,6 +71,11 @@ export function computeRenderData(
         let thisLineData: Partial<IPoint>[] = [];
         // keep track of which samples have mutations
         const samplesWithData: { [uniqueSampleKey: string]: boolean } = {};
+        const samplesWithDataGroup =
+            groupByOption && groupByOption != GROUP_BY_NONE
+                ? sampleIdToClinicalValue[mergedMutation[0].sampleId]
+                : undefined;
+
         for (const mutation of mergedMutation) {
             const sampleKey = mutation.uniqueSampleKey;
             const sampleId = mutation.sampleId;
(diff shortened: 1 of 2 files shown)


dianab0:
Added another commit for another Group By issue of the VAF Chart:

When using Group By, in some cases there are samples that are not assigned to a group and the page would crash.
Example: Grouping by "Tumor Purity" or "Metastatic Site" in https://www.cbioportal.org/patient/genomicEvolution?studyId=msk_impact_2017&caseId=P-0001104
    I fixed this, so the samples with undefined group are shown on a different row without group label in the left and will be drawn
    on the chart using the with black color.
Screenshot from 2021-06-17 11-39-17

src/pages/patientView/timeline2/VAFChartUtils.ts
```diff
@@ -118,29 +123,38 @@ export function computeRenderData(
...
             if (!(sample.uniqueSampleKey in samplesWithData)) {
+                // check if it is in the same group as samplesWithData
                 if (
-                    !isSampleProfiled(
-                        sample.uniqueSampleKey,
-                        mutationProfileId,
-                        mutation.gene.hugoGeneSymbol,
-                        coverageInformation
-                    )
+                    groupByOption == undefined ||
```
alisman, line 128: in general better to use === (will not do type coercion)

PR #3807, opened 2021-06-16
Adapt frontend to new driver-filtering-enabled api

pvannierop:
Ensure backward compatibility with new driver-filtering-enabled api.  

See backed PR for detail and specification of feature
https://github.com/cBioPortal/cbioportal/pull/8735


inodb:
@pvannierop Thanks! Works for `lgg_ucsf_2014` now, but I still see an issue with https://master.cbioportal.org/study/summary?id=msk_impact_2017:

Screen Shot 2021-06-16 at 9 53 05 AM

pvannierop:
@inodb The problem here is that this study has _NA_ value for mutation status in the MAF file. The driver-filtering SQL code did not take that into account. I will submit a PR with the fix as soon as I know how _NA_ values should be classified when selecting _somatic_, _germline_ or _unknown_ from the driver-filtering menu.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2575,13 +2581,21 @@ export class StudyViewPageStore {
         geneFilters = _.reduce(
             geneFilters,
             (acc, next) => {
-                const newGroup = next.filter(oql => oql !== toBeRemoved);
+                const [
+                    hugoGeneSymbol,
+                    alteration,
+                ]: string[] = toBeRemoved.split(':');
```
alisman, line 2587: @pvannierop could we not run this through the parser you wrote?

pvannierop:
Closed because I needed to open a draft PR for local e2e-testing against custom backend.

pvannierop:
RT[](url)

PR #3819, opened 2021-06-25
Survival charts don't show p-value whenever any group has less than 10 cases

dippindots:
Fix cBioPortal/cbioportal#8719


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -341,24 +341,33 @@ export default class SurvivalChart
         return data;
     }
 
+    private get showPValueText() {
+        // 1. p value is not null or undefined
+        // 2. every group has at least 10 cases
+        return (
+            !_.isNil(this.props.pValue) &&
```
dippindots, line 348: https://lodash.com/docs/4.17.15#isNil

_.isNil(value): Checks if value is null or undefined.

inodb (approved):
Great, thank you!!

jjgao (changes requested):
We should add a message about why p-value is not shown.

inodb:
@dippindots @jjgao maybe we can show the text:

"p-value not shown if <10 samples"

dippindots:
Thanks @jjgao @inodb for review this! Any suggestions on how to show this message? On our example link, I already see 3 messages on the top...

inodb:
@dippindots thanks so much! Just realized the samples typo on my end, my bad. Maybe instead of

```
Logrank Test P-Value: N/A (<10 samples)
```

We can do 

```
Logrank Test P-Value: N/A (<10 cases)
```

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -341,26 +341,36 @@ export default class SurvivalChart
...
+    }
+
+    private get pValue() {
+        // show NA if any group has least than 10 cases
+        const showNA = _.some(
+            this.props.analysisGroups,
+            group => this.props.sortedGroupedSurvivals[group.value].length < 10
+        );
+        if (showNA) {
+            return 'N/A (<10 samples)';
```
inodb, line 356:
```suggestion
            return 'N/A (<10 cases)';
```

inodb:
Maybe we should write this out a bit more - 

Be careful when interpreting this p-value because group x has less than y samples

dippindots:
close this until we have better understanding of how to fix this.

PR #3826, opened 2021-06-30
Add tooltip for annotation column

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/8305
Part of: https://github.com/cBioPortal/cbioportal/issues/8711

image

Test: query any study and check mutations tab, patients view and mutation mapper tool
For example:
https://deploy-preview-3826--cbioportalfrontend.netlify.app/results/mutations?cancer_study_list=msk_impact_2017&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cfusion%2Ccna&case_set_id=msk_impact_2017_cnaseq&gene_list=BRAF&geneset_list=%20&tab_index=tab_visualize&Action=Submit


packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -231,6 +231,12 @@ export function getAnnotationData(
                 oncoKbIndicator,
                 oncoKbAvailableDataTypes,
             };
+        } else if (oncoKbData && oncoKbData.isPending) {
```
leexgh, line 234: We need to add this because we don't catch pending status for `oncoKbData` correctly. If `oncoKbGeneExist` is false then it directly goes to 
```
else {
    value = {
        ...value,
        oncoKbStatus: 'complete',
        oncoKbIndicator: undefined,
    };
}
```
But at this moment `oncoKbData` could still be pending

inodb (commented):
Thanks for putting this together! Great work!

There are some more show/enable properties for each resource that would be good to handle

Try e.g.
```
localStorage.frontendConfig = '{"serverConfig":{"show_civic":false}}'
```

It will still show the civic icon in the header (prolly good to not show in the header if we've disabled the header at the app configuration level):
Screen Shot 2021-07-02 at 12 45 15 PM

Similarly for other resources it would be good to not show them in the header if they are disabled (oncokb, hotspots, civic, mycancergenome)

src/shared/components/mutationTable/column/annotation/AnnotationHeader.tsx
```diff
@@ -0,0 +1,466 @@
...
+    [OncokbDiagnosticLevelIconEnum.LEVEL_DX2]:
+        'FDA and/or professional guideline-recognized biomarker that supports diagnosis in this indication',
+    [OncokbDiagnosticLevelIconEnum.LEVEL_DX3]:
+        'Biomarker that may assist disease diagnosis in this indication based on clinical evidence',
+    [OncokbPrognosticLevelIconEnum.LEVEL_PX1]:
+        'FDA and/or professional guideline-recognized biomarker prognostic in this indication based on well-powered studie(s)',
+    [OncokbPrognosticLevelIconEnum.LEVEL_PX2]:
+        'FDA and/or professional guideline-recognized biomarker prognostic in this indication based on a single or multiple small studies',
+    [OncokbPrognosticLevelIconEnum.LEVEL_PX3]:
+        'Biomarker is prognostic in this indication based on clinical evidence in well-powered studies',
```
zhx828, line 83: In react-mutation-mapper, we have something similar to this. @onursumer I also realize there are good amount of content here can be shared between RMM and portal. What do you suggest in terms of how to resolve the dependency?

src/pages/patientView/structuralVariant/PatientViewStructuralVariantTable.tsx
```diff
@@ -36,8 +37,33 @@ export default class PatientViewStructuralVariantTable extends React.Component<
...
+            ) {
+                this.oncokbWidth =
+                    Number(
+                        document
+                            .getElementById('sv-annotation')
+                            ?.getElementsByClassName('oncokb-content')[0]
+                            ?.clientWidth
+                    ) || 22;
+                clearInterval(this.oncokbInterval);
+            }
```
onursumer, line 61: should we extract this to a utility function? this is very similar to the one for `copy-number-annotation`

src/shared/components/mutationTable/column/annotation/AnnotationHeader.tsx
onursumer: We should probably get these from `react-mutation-mapper` if possible.

onursumer (commented):
Looks good. Just a few minor suggestions.

src/pages/patientView/structuralVariant/PatientViewStructuralVariantTable.tsx
leexgh: @onursumer It could work but would it introduce some other code? Because we need to have `clearInterval(this.oncokbInterval);`, so for example if we move them to a util function, then we will have something like this:
```
this.oncokbInterval = setInterval(() => {
            someUtilFunction(this.setOncokbWidth, this.clearIntervalFn);
        }, 500);
```
and we will have to add this wherever that util function is called:
```
    @action.bound
    private clearIntervalFn() {
        clearInterval(this.oncokbInterval);
    }

    @action.bound
    private setOncokbWidth(width: number) {
        this.oncokbWidth = width;
    }
```

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -1,5 +1,11 @@
 import * as React from 'react';
-import { computed, makeObservable } from 'mobx';
+import {
+    autorun,
```
onursumer, line 3: is this used anywhere?

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -1,5 +1,11 @@
...
+    computed,
+    IReactionDisposer,
```
onursumer, line 5: I guess this is also not used

src/pages/patientView/structuralVariant/PatientViewStructuralVariantTable.tsx
onursumer: Let's not worry about this then for now.

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -1,6 +1,12 @@
 import * as React from 'react';
 import { observer } from 'mobx-react';
-import { observable, computed, makeObservable } from 'mobx';
+import {
+    observable,
+    computed,
+    makeObservable,
+    IReactionDisposer,
+    autorun,
```
onursumer, line 8: these are not used, right?

src/pages/patientView/mutation/PatientViewMutationTable.tsx
leexgh: @onursumer yes, good catch!

packages/react-mutation-mapper/src/index.tsx
```diff
@@ -170,3 +171,4 @@ export { default as DefaultMutationMapperDataStore } from './store/DefaultMutati
 export { default as DefaultMutationMapperFilterApplier } from './store/DefaultMutationMapperFilterApplier';
 export { default as DefaultMutationMapperStore } from './store/DefaultMutationMapperStore';
 export * from './component/oncokb/constants';
+export * from './util/OncoKbUtils';
```
onursumer, line 174: Instead of introducing a new export statement we should replace the lines [162-165](https://github.com/cBioPortal/cbioportal-frontend/pull/3826/files#diff-b1ffa3f6b2299ce3783b21c0dd11d9e23ba2c254568da3400726ec2033064bbeR162-R165):

```JavaScript
export {
    defaultOncoKbFilter,
    groupOncoKbIndicatorDataByMutations,
} from './util/OncoKbUtils';
```

leexgh: Good catch!

onursumer (approved):
Thanks!

PR #3831, opened 2021-07-02
Add column-based filters for mutation tables

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8713 and `2.2 Filtering per column` in cBioPortal/cbioportal#8711

Changes:
- Added two types of data filters, `NumericalFilter` and `CategoricalFilter`, which are attached to columns in the results view mutation table
- Numerical filters consist of a range slider with two handles and input boxes
- Categorical filters consist of search and multi-select functionality
- Multiple column filters can be active simultaneously, and they work in conjunction with the search filter in the top-right (i.e. only those rows which match ALL active filters will be shown)
- Hover cursor over column in order to show filter icon (if it's applicable for that column)

numerical_filter
categorical_filter


src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -113,6 +117,14 @@ export default class PatientViewMutationTable extends MutationTable<
                 ),
             download: (d: Mutation[]) =>
                 AlleleFreqColumnFormatter.getFrequency(d),
+            numericalFilter: (d: Mutation[], config: NumericalFilterConfig) =>
```
onursumer, line 120: I don't think we should couple a specific filter logic with the column definition. Are we going to add a `categoricalFilter` as well? I would instead define new `DataFilter`s (called `NumericalDataFilter`, `CategoricalDataFilter`, etc.) and handle all the logic inside a filter applier (say `NumericalDataFilterApplier`, `CategoricalFilterApplier`, etc.). By this way we wouldn't couple any filter specific logic with the column definition and also keep all data filters in one place. That would also simplify the cases where multiple data filters applied at once.

Here is an example implementation where we add a custom filter for the results view mutation mapper store: https://github.com/cBioPortal/cbioportal-frontend/blob/c0be78ff9eced5400308c82cd69cf142cf227bad/src/pages/resultsView/ResultsViewPageStore.ts#L3458

One problem with this approach is that we wouldn't be supporting generic numerical filters by default for the `LazyMobXTable`, but maybe we shouldn't worry about that for now.

src/shared/lib/MutationUtils.tsx
```diff
@@ -565,3 +571,66 @@ export function createAnnotatedProteinImpactTypeFilter(
...
+export function createCategoricalFilter(getData: (d: Mutation) => string) {
+    const filter = (filter: CategoricalFilter, mutation: Mutation) => {
+        const value = getData(mutation);
+        if (filter.values.length < 2 || value === '') {
+            return false;
+        }
+
+        const filterType = filter.values[0] as string;
+        const filterString = filter.values[1] as string;
+        const selections = filter.values[2] as Set<string>;
```
onursumer, line 594: This is not ideal and a bit confusing. We should probably come up with a better solution. Ideally `filterType` should be assigned to `filter.type` not to `filter.values[0]`. Filter values should just be the actual values. Filter value of a categorical filter should probably be defined like this (instead of an array of different types) to make things more clear:

```
{
    filterString: string;
    selections: Set<String>;
}
```

This way we can also pass multiple values if we want/need. But with the current design we can only pass 3 values and they actually correspond to a single value.

src/shared/lib/MutationUtils.tsx
```diff
@@ -565,3 +571,66 @@ export function createAnnotatedProteinImpactTypeFilter(
...
+    return filter;
+}
+
+export function matchCategoricalFilterSearch(
+    value: string,
+    filterType: string,
+    filterString: string
+) {
+    switch (filterType) {
+        case 'contains':
```
onursumer, line 614: I would define a separate function for each `DataFilter.type` instead of a lengthy switch case.

src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
```diff
@@ -66,7 +66,7 @@ class FunctionalImpactColumnTooltip extends React.Component<
                     <thead>
                         <tr>
                             <th>Legend</th>
-                            {SHOW_MUTATION_ASSESSOR && (
+                            {shouldShowMutationAssessor() && (
```
onursumer, line 69: I wouldn't call this function multiple times. How about defining a constant before the return statement and using it everywhere else?

```JavaScript
    const showMutationAssessor = shouldShowMutationAssessor();
```
```suggestion
                            {showMutationAssessor && (
```

src/shared/lib/MutationUtils.tsx
```diff
@@ -565,3 +571,66 @@ export function createAnnotatedProteinImpactTypeFilter(
     };
     return filter;
 }
+
+export function createNumericalFilter(getData: (d: Mutation) => number | null) {
+    const filter = (filter: NumericalFilter, mutation: Mutation) => {
+        const value = getData(mutation);
+        const lowerBound = filter.values[0];
+        const upperBound = filter.values[1];
```
onursumer, line 579: This actually a single value. We should define our `NumericalFilter` value something like this:

```JavaScript
{
    lowerBound: number;
    upperBound: number;
}
```

env/custom.sh
```diff
@@ -1,2 +1,2 @@
-export CBIOPORTAL_URL="http://localhost:8080"
+export BRANCH_ENV="http://localhost:8080"
```
onursumer, line 1: not sure if we should touch this

onursumer (commented):
Thanks for changing the design to use `DataFilters`! Added some suggestions to further improve the clarity of the filters.

src/shared/lib/MutationUtils.tsx
Rajat-Sirohi: Are we sure about this? This would require introducing 9 filter appliers for each categorical `MutationTableColumnType` (of which there are well over 10). Additionally, adding/removing/renaming any categorical filter types would require adding/removing/renaming that many more filter appliers for each column. I would also need to make some adjustments to the mutation mapper code in order to track the categorical filter type for each column, which seems confusing in my opinion, since I don't do that for numerical filters.

Why not just keep the categorical filter type as a pseudo-value called `filterCondition` as part of the `CategoricalFilterValue` object? I see it as just another data filter configuration option, like `filterString` or `lowerBound` and `upperBound` for the numerical filter. The alternative just seems too bloated and not extensible.

src/shared/components/doubleHandleSlider/DoubleHandleSlider.tsx
```diff
@@ -0,0 +1,305 @@
...
+    callbackUpperValue: (upperValue: number) => void;
+}
+
+export interface IDoubleHandleSliderState {
+    lowerValue: number;
+    upperValue: number;
+}
+
+const WIDTH = '135px';
+export default class DoubleHandleSlider extends React.Component<
```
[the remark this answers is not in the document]
Rajat-Sirohi: The two slider components which are already installed are `react-bootstrap-slider` and `react-rangeslider`. As far as I can tell, the former DOES support multiple values whereas the latter DOES NOT. Unfortunately, I had tried to get `react-bootstrap-slider` actually imported and visible in the past without luck, and all the existing code uses `react-rangeslider`. This is why I ended up creating my own component, since I didn't want to add anymore package dependencies.

Additionally, most of the code in my file is dedicated to syncing the slider with the two input boxes and having everything be the right size. Importing a double-slider wouldn't actually save that much code space.

src/shared/lib/MutationUtils.tsx
onursumer: I see. This one is a bit tricky. I guess we can think of it as `filterCondition` and pair it with the filter value object. I would still prefer a simple filter applier function doing a single thing, but I think I can live with this for now 🙂

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -103,6 +103,8 @@ function getDefaultTumorType(): string {
     return 'Unknown';
 }
 
+const memoized: Map<string, IAnnotation> = new Map();
```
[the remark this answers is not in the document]
Rajat-Sirohi: Unfortunately, the compiler complains when I add a `@memoize` decorator, probably because the method isn't simple like those in `CosmicColumnFormatter.tsx`

Luke-Sikina (commented):
Few small code changes, + 2 more commends:
- The regex filter shouldn't be case sensitive.
- The hide NA values check box isn't aligned with the text, that should be fixed. 
Screenshot from 2021-07-15 15-28-07

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -82,7 +95,25 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
...
+                mutationMapperStoreConfig['filterAppliersOverride']![
+                    MutationTableColumnType.DBSNP
+                ] = createCategoricalFilter((d: Mutation) =>
+                    DbsnpColumnFormatter.download(
+                        [d],
+                        this.indexedMyVariantInfoAnnotations
+                    )
+                );
+                return mutationMapperStoreConfig;
+            })(),
```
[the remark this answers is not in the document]
Rajat-Sirohi: Unfortunately, I need to access `this.indexedMyVariantInfoAnnotations` so I can't access it from a `static` method; nor can I pass it to `mutationMapperStoreConfig` from `ResultsViewPageStore` since it's a property of `DefaultMutationMapperStore`.

Therefore I have to either leave my function as is or add those two filter appliers to `@computed get numericalFilterColumns` and `@computed get categoricalFilterColumns` respectively, which is prettier but a bit less efficient.

jjgao:
Great job! Thanks, @Rajat-Sirohi 

@Rajat-Sirohi @Luke-Sikina  can we add one additional feature either part of this or or a separate one? It would be really useful to add a link to the study view for the filtered samples.

image

jjgao:
Issue with Exon column in [this query](https://deploy-preview-3831--cbioportalfrontend.netlify.app/results/mutations?cancer_study_list=ov_tcga_pub&cancer_study_id=ov_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ov_tcga_pub_mutations&Z_SCORE_THRESHOLD=2.0&case_set_id=ov_tcga_pub_3way_complete&gene_list=BRCA1+BRCA2&gene_set_choice=user-defined-list)

image

Rajat-Sirohi:
Thanks for identifying that bug, @jjgao ! It was a problem with computing the min/max for the column, where the Exon column was returning a fraction, not a valid number. I've now fixed it and will update code in the next push.

inodb:
@Rajat-Sirohi i noticed one tiny bug where the filter icon won't show up. If I go [here](https://deploy-preview-3831--cbioportalfrontend.netlify.app/results/mutations?tab_index=tab_visualize&Action=Submit&session_id=60f9c65ce4b0f0abac8bbcdf&plots_horz_selection=%7B%7D&plots_vert_selection=%7B%7D&plots_coloring_selection=%7B%7D), mouse over the protein change column header, click on the fliter icon and search for 1000, then click anywhere on the page. The blue filter icon doesn't show up:

Screen Shot 2021-07-22 at 4 45 08 PM

Once I mouse over again, it does show up. Doesn't really have to be a blocker for merging tho, but curious if there's an easy solution

Rajat-Sirohi:
@inodb The first issue happens because the input box is over a different column, so the filter icon isn't shown until that column is moused over again. I'll definitely see what I can do, but I don't think there's an easy fix.

To your second point, I just check the horizontal coordinates of the cursor and see if it's inside a column, and if so then I show the filter icon. I don't check vertical coordinates because I was told to show it if the mouse is inside any part of the column, not just the header. Unfortunately I also end up showing it if the mouse is above the column due to an issue where the vertical position of the column header sometimes changes. I don't think there's an easy fix for that either.

Rajat-Sirohi:
@inodb I've now fixed that first issue so that the filter icon is visible so long as the cursor is either over the column OR the filter menu is open. Thanks for pointing it out.

I think I will leave the other part as is, where the icon is shown regardless of the vertical position of the cursor. This might seem unnecessary, especially when the cursor is above the column header, but I think it's better to ensure that the user knows that filtering is an option rather than risk them not even noticing the icon if were to make its visibility so restrictive, e.g., if it were only shown when hovering over the column header.

PR #3836, opened 2021-07-09
Fix hgvsc column blank

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/517
HGVSc is empty at first because it doesn't have `@observer`

Test: https://www.cbioportal.org/patient?studyId=lgg_ucsf_2014&caseId=P04


leexgh:
Duplication of https://github.com/cBioPortal/cbioportal-frontend/pull/3835

PR #3839, opened 2021-07-12
use function to return mutation assessor configuration setting at runtime

alisman:
We should never access configuration at "parse" time because configuration may not yet have been loaded into page.  We should always check configuration at run time


leexgh (approved):
Thanks!

PR #3840, opened 2021-07-13
Added showNA property to generic assay charts

dianab0:
ShowNAs is available for generic assay charts but not working.

Replication steps:
- Go to ccle_broad_2019 study
- Select Chart -> Treatment Response -> Pick a treatment and Add Chart
- Then go to the displayed histogram, hover the top-right corner, and pick "Show NAs"
- It is not working and there will be the following error in the console:
ShowNAError

ShowNA_Error

The fix can be checked here: https://cbioportal.thehyve.net/study/summary?id=ccle_broad_2019 using account testuser - P@ssword.

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -352,22 +352,54 @@ export class StudyViewPageStore {
     @action
     updateNAValue = (uniqueKey: string): void => {
         let newValue = this.isShowNAChecked(uniqueKey);
-        let newFilter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey))!;
-        newFilter.showNA = newValue;
-        this._clinicalDataBinFilterSet.set(uniqueKey, newFilter);
+        if (this.isGenericAssayChart(uniqueKey)) {
+            let newFilter = _.clone(
+                this._genericAssayDataBinFilterSet.get(uniqueKey)
+            )!;
+            newFilter.showNA = newValue;
+            this._genericAssayDataBinFilterSet.set(uniqueKey, newFilter);
+        } else {
+            let newFilter = _.clone(
+                this._clinicalDataBinFilterSet.get(uniqueKey)
+            )!;
+            newFilter.showNA = newValue;
+            this._clinicalDataBinFilterSet.set(uniqueKey, newFilter);
+        }
     };
 
     @action
     toggleNAValue = (uniqueKey: string): void => {
         this.updateNAValue(uniqueKey);
-        let newFilter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey))!;
+
+        let newFilter = this.isGenericAssayChart(uniqueKey)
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -352,22 +352,49 @@ export class StudyViewPageStore {
...
     };
 
     public isShowNAChecked = (uniqueKey: string): boolean => {
-        let filter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey));
+        let filter;
+        if (this.isGenericAssayChart(uniqueKey))
+            filter = _.clone(
+                this._genericAssayDataBinFilterSet.get(uniqueKey)
+            )!;
+        else filter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey));
```
dippindots, line 396: Personally I prefer a full brace style, but I am ok with this also.

dippindots (approved):
Hi @dianab0, thanks for fixing this issue! It looks good to me.

src/pages/studyView/StudyViewPageStore.ts
alisman: @dianab0 lets just put in the full braces.  want to keep style consistency

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -352,22 +352,49 @@ export class StudyViewPageStore {
...
     };
 
     @action
     toggleNAValue = (uniqueKey: string): void => {
         this.updateNAValue(uniqueKey);
-        let newFilter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey))!;
-        let showNA = newFilter.showNA;
-        newFilter.showNA = !showNA;
-        this._clinicalDataBinFilterSet.set(uniqueKey, newFilter);
+        if (this.isGenericAssayChart(uniqueKey)) {
```
alisman, line 373: seems this code is quite redundant.   how about:
```
targetWhateverSet = this.isGenericAssChart(uniqueKey);
// now proceed with cloning and setting
```

dianab0: I tried this but I get an error when trying to set newFilter to targetWhateverSet (targetWhateverSet.set(uniqueKey, newFilter)) as newFilter can have two types GenericAssayDataBinFilter or ClinicalDataBinFilter.
However, I modified a bit the method to be less redundant.

dianab0:
Thank you for review! I updated the code under the same commit.

alisman:
@adamabeshouse does this duplicate the work you did?

alisman:
@adamabeshouse can you decide we decide which to merge (if i'm right that it's redundant)

adamabeshouse:
@alisman ah, I didn't realize this was a problem for generic assay charts too. this does not duplicate but it would dovetail well with my work. lets merge this and then Ill combine it in my PR

PR #3856, opened 2021-07-19
Only show "ERROR" on failed mutations instead of all mutations

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/8739
If one of the mutations can't be annotated, only this mutation should show "ERROR", instead of all mutations in this query.

Reason:
For example:
```
{
    "variant": "4:g.55133624_55133897del",
    "originalVariantQuery": "4,55133624,55133897,CATGGTACATTCCGCTTTCTAAAATGTCAGTTGTCCATGCTGCTCGGGATCCATATGTGGTAATCATTATTTAATGGAAACTCTTCCCTGTACAGAGAAAGGTTTCATTGAAATCAAACCCACCTTCAGCCAGTTGGAAGCTGTCAACCTGCATGAAGTCAAACATTTTGTTGTAGAGGTGCGGGCCTACCCACCTCCCAGGATATCCTGGCTGAAAAACAATCTGACTCTGATTGAAAATCTCACTGAGATCAC,-",
    "successfully_annotated": false
  },
```
This mutation can't be annotated. We use `genomicLocation` to map annotations back to mutations, when annotation fails it doesn't have `genomicLocation`. In this case we could use `originalVariantQuery` which will be always available.


onursumer (approved):
Thanks!

PR #3858, opened 2021-07-19
study view: open a comparison page from structural variants table

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/7991


Luke-Sikina (approved):
Looks good to me. The only thing I saw that didn't see amazing was the whole fetching of entrez gene IDs. I'm surprised that information isn't already in the StudyViewPageStore. That said, I see this pattern is already used quite a bit, so I'm not going to worry about it too much- it clearly works.

adamabeshouse:
@Luke-Sikina thats a good point. let me take a look at refactoring this a bit so we can avoid that extra call

PR #3861, opened 2021-07-20
Update gene symbols to match targeted panels

rmadupuri:
- Validator is updated to identify genes outside the sequenced panel - https://github.com/cBioPortal/cbioportal/pull/8769. It now reports an error for off-panel genes. 
- Updated the symbols in the test study `genepanel_test_study` to match the sequenced panel and to pass the validation.


dippindots (commented):
To test new file against correct backend version, you can add this `export BACKEND=cbioportal:validate_off_panel_genes` in this file: https://github.com/cBioPortal/cbioportal-frontend/blob/master/env/custom.sh

This should be able to use your new changes in backend for testing, the full documentation is here: https://github.com/cBioPortal/cbioportal-frontend#cbioportal-backend-version

PR #3885, opened 2021-07-29
SV: Make cancer study chart slices color consistent when applying filter

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8524


zhx828 (approved):
Looks good

PR #3887, opened 2021-07-29
Study View: Fix exporting cancer study table

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8471

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -6837,72 +6837,90 @@ export class StudyViewPageStore
 
     public getCustomChartDownloadData(chartMeta: ChartMeta): Promise<string> {
         return new Promise<string>(resolve => {
-            if (
-                chartMeta &&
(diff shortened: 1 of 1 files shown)


zhx828 (approved):
💯

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6880,72 +6880,90 @@ export class StudyViewPageStore
 
     public getCustomChartDownloadData(chartMeta: ChartMeta): Promise<string> {
```
alisman, line 6881: I'm curious why this stuff takes place in "custom chart download data"?   and all the forking in here is kinda unpleasant.  like, this should be in configuration no?  should we just bite the bullet and do the refactor?

kalletlak: Cancer Study chart is also treated as custom chart since it is not a clinical attribute. Yes, this needs a refactoring. May be we can merge this for now and do refactoring in next cycle - as it takes much time.

PR #3890, opened 2021-08-01
Add link to filtered study view + Fix URL generation

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8802

### Changes:
- Modify the filter reset panel in the results view "Mutations" tab to include a link to a filtered study view page (according to the mutation table filters)
- Fix issue with local development where clicking hyperlinks (e.g. "Sample ID" link to patient view from results view mutation table cell) requires the user to manually modify the generated link. The problem was that `AppConfig.baseUrl` is undefined outside of the production portal.

filterResetPanel_0
filterResetPanel_1

### DIFF
--- a/src/shared/api/urls.ts
+++ b/src/shared/api/urls.ts
@@ -59,9 +59,11 @@ export function buildCBioPortalPageUrl(
         typeof pathnameOrParams === 'string'
             ? { pathname: pathnameOrParams, query, hash }
             : pathnameOrParams;
+
+    // AppConfig.frontendUrl format is '//url/', hence the specified slice to get just 'url'
     return URL.format({
         protocol: window.location.protocol,
-        host: AppConfig.baseUrl,
+        host: AppConfig.baseUrl || AppConfig.frontendUrl?.slice(2, -1),
         ...params,
     });
 }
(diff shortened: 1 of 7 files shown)


src/shared/components/mutationMapper/MutationMapperDataStore.ts
```diff
@@ -196,6 +196,19 @@ export default class MutationMapperDataStore
         return countDuplicateMutations(this.tableDataGroupedByPatients);
     }
 
+    @computed
+    get tableDataSamples() {
+        return _.uniqBy(_.flatten(this.tableData), 'sampleId').map(m => ({
```
onursumer, line 201: property shorthands are not recommended because they are not type safe. this is better:

```suggestion
        return _.uniqBy(_.flatten(this.tableData), m => m.sampleId).map(m => ({
```

src/shared/api/urls.ts
```diff
@@ -61,7 +61,7 @@ export function buildCBioPortalPageUrl(
             : pathnameOrParams;
     return URL.format({
         protocol: window.location.protocol,
-        host: AppConfig.baseUrl,
+        host: AppConfig.baseUrl || AppConfig.frontendUrl?.slice(2, -1),
```
onursumer, line 64: should we add a comment here? it's not immediately clear to me why we are slicing the frontend url.

onursumer (commented):
Looks good. Just a few minor comments/suggestions.

onursumer:
We have this minor issue when the number of patients/samples is 1:

1_samples_patients

inodb (approved):
Nice ✨ !

PR #3891, opened 2021-08-01
fix hover detection bug

Rajat-Sirohi:
Fix cBioPortal/cbioportal#8803

It turns out that I had initially misdiagnosed the issue (#3880), but I've now understood the problem and have properly addressed it. The issue was not related to any difference between `window.onmousemove` and `document.onmousemove`. In fact, I had simply not noticed that I was manually resetting `window.onmousemove` to `null` when the table unmounted. The real issue required me to distinguish between a new table being created due to, for example, a tooltip window (in which case I call it a "child table") and a new table being created due to switching between genes.


inodb (approved):
Thanks for pro-actively fixing this!

PR #3893, opened 2021-08-02
Change error message when study doesn't exist or user doesn't have ac…

dianab0:
Changed error page when study doesn't exist or user doesn't have access according to issue https://github.com/cBioPortal/cbioportal/issues/8500

Currently there is code that marks the study as unknown or unauthorized but only the pageStatusMessages is updated and it continues with fetching data about the study until it finally fails at api/treatment/display-patient. So the error page "Oops. There was an error retrieving data" refers to failed api/treatment/display-patient.

I have changed this by setting a flag when all queried studies are unknown/unauthorized and displaying the error message "The study/studies you are trying to access do not exist, or you do not have access." When only some study ids are unknown/unauthorized then the page is displayed with the valid studies and a status message on the header about the unknown/unauthorized ones.

Also the status message was displayed twice (in StudyViewPage and StudyPageHeader) so I removed the one from StudyViewPage.

Error page before fix:

Screenshot from 2021-08-06 11-34-16

Status messages before fix with physical studies and virtual studies:

Screenshot from 2021-08-06 11-33-17

Screenshot from 2021-08-06 11-33-32

Error page after fix:

Screenshot from 2021-08-06 11-34-08

Status messages after fix:

Screenshot from 2021-08-06 11-33-51

### DIFF
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -137,6 +137,27 @@ export default class Container extends React.Component<IContainerProps, {}> {
                         </div>
                     </Else>
                 </If>
+                <If condition={this.appStore.unknownOrUnauthorizedStudyIds}>
+                    <Then>
+                        <div className="contentWrapper">
+                            <ErrorScreen
+                                title={
+                                    'The study/studies you are trying to access do not exist, or you do not have access.'
+                                }
+                                body={
+                                    <a href={buildCBioPortalPageUrl('/')}>
+                                        Return to homepage
+                                    </a>
+                                }
+                            />
+                        </div>
+                    </Then>
+                    <Else>
+                        <div className="contentWrapper">
+                            {makeRoutes(this.routingStore)}
+                        </div>
+                    </Else>
+                </If>
             </div>
         );
     }
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -4337,12 +4337,19 @@ export class StudyViewPageStore
         onError: () => {},
         onResult: unknownIds => {
             if (unknownIds.length > 0) {
-                this.pageStatusMessages['unknownIds'] = {
-                    status: 'danger',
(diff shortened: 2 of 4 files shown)


Sjoerd-van-Hagen:
@mandawilson fix for the issue you reported.

@alisman please review.

Sjoerd-van-Hagen:
Ok, this is not ready for review.

Sjoerd-van-Hagen:
@alisman this one is ready for review now

src/appShell/App/Container.tsx
```diff
@@ -137,6 +137,27 @@ export default class Container extends React.Component<IContainerProps, {}> {
                         </div>
                     </Else>
                 </If>
+                <If condition={this.appStore.unknownOrUnauthorizedStudyIds}>
```
alisman, line 140: we shouldn't need to do this in the container.   the study view page itself can return an <ErrorScreen> component under this condition.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4337,12 +4337,19 @@ export class StudyViewPageStore
         onError: () => {},
         onResult: unknownIds => {
             if (unknownIds.length > 0) {
-                this.pageStatusMessages['unknownIds'] = {
-                    status: 'danger',
-                    message: `Unknown/Unauthorized ${
-                        unknownIds.length > 1 ? 'studies' : 'study'
-                    } ${unknownIds.join(', ')}`,
-                };
+                // if all studies are unknown, set flag for error page message
```
alisman, line 4340: i think this whole onResult handler is unnecessary.  study page should just be able to check the promise itself and if the the unknownOrUnauthorizedStudyIds is true, then it can return an <ErrorPage> (or embed an alert bar if there are some legit study idds)

alisman:
@Sjoerd-van-Hagen @dianab0 i wanted this to be accomplished in different way so i made this PR:
https://github.com/cBioPortal/cbioportal-frontend/pull/3914/files

PR #3898, opened 2021-08-03
Fix bug in NA toggling for gene-specific charts in study view

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8772

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -31,7 +31,10 @@ import {
     DataFilterValue,
     DensityPlotBin,
     GeneFilter,
+    GeneFilterQuery,
     GenePanel,
+    GenericAssayDataFilter,
+    GenericAssayMeta,
     GenomicDataBin,
     GenomicDataBinFilter,
     GenomicDataFilter,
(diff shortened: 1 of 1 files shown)


zhx828 (approved):
👍🏼

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -378,24 +375,45 @@ export class StudyViewPageStore
...
-        let filter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey));
+    public isShowNAChecked = (
+        uniqueKey: string,
+        dataType: ChartMetaDataTypeEnum
+    ): boolean => {
+        const filterSet =
+            dataType === ChartMetaDataTypeEnum.GENE_SPECIFIC
+                ? this._genomicDataBinFilterSet
+                : this._clinicalDataBinFilterSet;
+        let filter = _.clone(filterSet.get(uniqueKey));
```
alisman, line 416: why do we need to clone this here given that we are just returning a boolean?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -378,24 +375,45 @@ export class StudyViewPageStore
...
-    updateNAValue = (uniqueKey: string): void => {
-        let newValue = this.isShowNAChecked(uniqueKey);
-        let newFilter = _.clone(this._clinicalDataBinFilterSet.get(uniqueKey))!;
+    updateNAValue = (
+        uniqueKey: string,
+        dataType: ChartMetaDataTypeEnum
+    ): void => {
+        let newValue = this.isShowNAChecked(uniqueKey, dataType);
+        const filterSet =
+            dataType === ChartMetaDataTypeEnum.GENE_SPECIFIC
```
alisman, line 384: for some reason i hate that we have to do this check and then repeat it in some many places.  like, it feels to that this kind of function shouldn't need knowledge of the type.  it should instead reside in the chart obj itself?

alisman (approved):

PR #3902, opened 2021-08-04
Improve specificity of URLWrapper hash props

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8801
Credit to @kalletlak for finding the problem

This PR decouples two unrelated concepts in URLWrapper, which until now had been unified:
(1) Props that should be placed in the session service, either because they are fundamental to the query or because their long length may exceed browser URL limits

(2) Props that should be used to determine the URLWrapper "hash" which is used in the ResultsViewPage to trigger a page remount.

### DIFF
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -20,7 +20,6 @@ import IComparisonURLWrapper from 'pages/groupComparison/IComparisonURLWrapper';
 import _ from 'lodash';
 import { MapValues } from 'shared/lib/TypeScriptUtils';
 import { GroupComparisonTab } from 'pages/groupComparison/GroupComparisonTabs';
-import hashString from 'shared/lib/hashString';
 
 export type PlotsSelectionParam = {
     selectedGeneOption?: string;
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -118,112 +118,150 @@ export type ResultsViewURLQuery = {
     plots_coloring_selection: PlotsColoringParam;
 };
 
-const propertiesMap: PropertiesMap<ResultsViewURLQuery> = {
-    // NON session props here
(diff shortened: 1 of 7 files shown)


kalletlak (approved):

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -117,87 +117,151 @@ export type ResultsViewURLQuery = {
...
+        },
+        genetic_profile_ids_GENERIC_ASSAY: {
+            isSessionProp: true,
+        },
+        genetic_profile_ids: { isSessionProp: true },
+        comparison_createdGroupsSessionId: {
+            isSessionProp: true,
+        },
+    } as PropertiesMap<ResultsViewURLQuery>,
+    (propertySpec, propertyName) => {
```
alisman, line 258: curious why we do this merge instead of just adding property to the objects listed in propertiesMap?

src/pages/resultsView/comparison/ResultsViewComparisonStore.ts
```diff
@@ -171,6 +171,7 @@ export default class ResultsViewComparisonStore extends ComparisonStore {
     protected async saveAndGoToSession(newSession: Session) {
         const { id } = await comparisonClient.addComparisonSession(newSession);
         this.urlWrapper.updateURL({ comparison_createdGroupsSessionId: id });
+        this.newSessionPending = false;
```
alisman, line 174: what's reason for this?

adamabeshouse: newSessionPending is an observable on the comparison store which is set whenever an action is initiated that involves creating a new comparison session (such as creating or deleting a group). its used for displaying a loading indicator. at this point, at the end of `saveAndGoToSession`, the new session is surely no longer pending, so it is set to false. this fixes an issue where the loading indicator was never disappearing - we didn't see that issue before because the whole page remounted (which it needn't do) whenever a new group was added or deleted from the comparison tab.

PR #3903, opened 2021-08-05
fix mangled unicode sort arrow

alisman:
Fix cBioPortal/cbioportal#8808

Sort arrow was using unicode character which got mangled by jitpack.  As easy fix, use fontawesome icons `fa-sort-asc` and `fa-sort-desc` instead.

Also centers the sort arrow icon in `FixedHeaderTable`, which is used in Study View.


Rajat-Sirohi (approved):
Overall, looks good. I left a few suggested code changes. The last thing which we might want to address is that the sort indicator icon sometimes looks out of place, specifically on the "Annotation" column or any multiline header, such as "# Mut in Sample":
annotation_sort_icon
multiline_sort_icon

But this is the case on the public portal anyways, so it can be addressed in a separate PR if at all.

Rajat-Sirohi (changes requested):
Sorry, meant to put requested changes, not approved status

Rajat-Sirohi (approved):
Looks good to me!

PR #3906, opened 2021-08-06
Fix bug relating to molecularData preload with massive data

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8817

Recently, we introduced a caching optimization where we fetch all data for queried genes and profiles, then filter through for the data for selected samples. This works most of the time, except when the study is massive. In that case, it seems that the query somehow fails, probably due to the sheer size of the preloaded data. In the failing case it is around 1.7 million objects.

This PR implements code that first checks if the data size is too large to preload. This still leverages caching. If so, it will just load data for the queried samples, making the necessary sacrifice of cacheability in this case.


inodb (approved):
Thanks for the quick fix!

PR #3910, opened 2021-08-06
Chunk calls for molecularData_preload

adamabeshouse:
Fix https://github.com/cBioPortal/cbioportal/issues/8817 in a different way


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1367,49 +1331,69 @@ export class ResultsViewPageStore
...
+                    // Recursively subdivide data requests until theyre a manageable size
+                    const molecularDataMultipleStudyFilter = {
+                        entrezGeneIds,
+                        molecularProfileIds,
+                    } as MolecularDataMultipleStudyFilter;
+
+                    // First check the data size. If it's too big, there's going to be server errors,
+                    //  so we can't prefetch the data for all samples
+                    const dataSize = await client
+                        .fetchMolecularDataInMultipleMolecularProfilesUsingPOSTWithHttpInfo(
```
alisman, line 1350: this assumes that using PROJECTION_META consumes less resources on backend so won't break down if query is too big?   i guess the proof is in the pudding (doesn't crash) but I wouldn't make this assumption.   why don't we just always chunk?  if we just always do say 3 gene chunks, seems safe to me.

adamabeshouse: okay, I guess that's probably fine

PR #3911, opened 2021-08-09
Do not merge

alisman:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3510,7 +3513,7 @@ export class StudyViewPageStore
                 const uniqueKey = item.attributeId;
                 if (this.isNewlyAdded(uniqueKey)) {
                     this.showAsPieChart(uniqueKey, item.counts.length);
-                    this.newlyAddedCharts.remove(uniqueKey);
+                    //this.newlyAddedCharts.remove(uniqueKey);
```
alisman, line 3516: commenting out these removals (there are 4) seems to fix the problem, except of course that newlyAddedCharts doesn't get cleared.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3491,6 +3491,8 @@ export class StudyViewPageStore
         await: () => [this.selectedSamples],
         invoke: () => {
             //only invoke if there are filtered samples
+
+            console.log('this gets invoked twice. why?');
```
alisman, line 3495: comment out the the newlyAddedCharts.remove() prevents this invoke from firing twice.  the second time it fires, it reaches the ```Promise.resolve([])``` which i believe is why the data disappears from the chart.

src/pages/studyView/StudyViewPageStore.ts
alisman, line 3516: commenting this out seems to fix problem.  i think that mutating this observable state here causes a promise to reinvoke and resolve with empty [].  leading to blank chart

alisman: Perhaps newlyAddedCharts doesn't really need to be observable?

PR #3912, opened 2021-08-09
SV: ignore alterationFilter while checking initial filter state

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8805


adamabeshouse (approved):
LGTM 🚀

dippindots (approved):
👍

inodb (commented):
@kalletlak Thanks for fixing this! Just curious - does this fully address the bug? Is it correct that the issue might still show up if one were to share a link with a driver annotation?

kalletlak:
Currently driver annotation are not yet supported in URL. It might need some updates when it is enabled

PR #3922, opened 2021-08-13
SV: Fix group refreshing issue on checkbox selection

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/8616

Problem: same object (`_selectedComparisonGroups`) is used for both creating and selecting groups. This is causing list to re-render whenever selection is made.

Solution: use different object to track created groups and selected groups

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -605,6 +605,10 @@ export class StudyViewPageStore
         ComparisonGroupId,
         boolean
     >({}, { deep: false });
+    private createdGroups = observable.map<ComparisonGroupId, boolean>(
+        {},
+        { deep: false }
+    );
     private _selectedComparisonGroupsWarningSigns = observable.map<
         string,
         boolean
@@ -618,6 +622,10 @@ export class StudyViewPageStore
         {},
         { deep: false }
     );
+
+    @action public trackNewlyCreatedGroup(groupId: string): void {
+        this.createdGroups.set(groupId, true);
+    }
 
     @action public setComparisonGroupSelected(
         groupId: string,
@@ -711,6 +719,7 @@ export class StudyViewPageStore
                 }
                 // delete it even from the shared group set
                 delete this.sharedGroupSet[groupId];
+                this.createdGroups.delete(groupId);
 
                 this._selectedComparisonGroups.delete(groupId);
             }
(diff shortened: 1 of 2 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -619,6 +623,10 @@ export class StudyViewPageStore
         { deep: false }
     );
 
+    @action public persistNewlyCreatedGroup(groupId: string): void {
```
adamabeshouse, line 626: I'm not sure I understand the word "persist" here, it doesn't look like its persisting it, just keeping track of it as created

kalletlak: Yes, I'll rename it `trackNewlyCreatedGroup`

adamabeshouse (approved):

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -939,29 +948,26 @@ export class StudyViewPageStore
...
-                        group =>
-                            !_.some(
-                                group.data.studies,
-                                study => studyIdsSet[study.id] === undefined
-                            )
-                    )
-                    .forEach(group =>
+                newGroups.forEach(group => {
+                    if (
+                        !_.some(
```
alisman, line 966: this would be more clear written:
```
_.every(studies, (s)=>s.id in studyIdsSet)
```
Also can you put a comment explaining why we would might sometimes get a group with studies that are not in the current queriedPhysicalStudies?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -939,29 +948,26 @@ export class StudyViewPageStore
...
+                this.createdGroups.keys()
             ).filter(groupId => groupIdSet[groupId] === undefined);
 
             if (missingGroupIds.length > 0) {
-                const promises = [];
-                for (const groupId of missingGroupIds) {
-                    promises.push(comparisonClient.getGroup(groupId));
-                }
+                const promises = missingGroupIds.map(groupId =>
+                    comparisonClient.getGroup(groupId)
```
alisman, line 957: comment getGroup will make a new group if it doesn't exist?

kalletlak: It doesn't make new group, if gets group by id if it is present.

alisman:
@kalletlak i'm curious why tracking newly created groups is important in fixing a bug that doesn't seem to involve new group creation (appears when selecting a group)?    maybe put explaination in main comment

PR #3924, opened 2021-08-13
Do not filter out structural variant for annotation when one of the genes is missing

zhx828:
In this case, it would be an intragenic event. This fixes https://github.com/cBioPortal/cbioportal/issues/8600

Signed-off-by: Hongxin <5400599+zhx828@users.noreply.github.com>

### DIFF
--- a/end-to-end-test/remote/specs/core/plots.screenshot.spec.js
+++ b/end-to-end-test/remote/specs/core/plots.screenshot.spec.js
@@ -162,17 +162,15 @@ describe('plots tab screenshot tests', function() {
     });
     it('plots tab clinical vs clinical boxplot', function() {
         browser.execute(function() {
-            resultsViewPlotsTab.onHorizontalAxisDataTypeSelect({
+            resultsViewPlotsTab.onVerticalAxisDataTypeSelect({
                 value: 'clinical_attribute',
             });
         });
         browser.execute(function() {
-            resultsViewPlotsTab.onHorizontalAxisDataSourceSelect({
+            resultsViewPlotsTab.onVerticalAxisDataSourceSelect({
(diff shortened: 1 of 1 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -846,14 +846,20 @@ export async function fetchStructuralVariantOncoKbData(
...
-            structuralVariantData.result,
-            d =>
+        const alterationsToQuery = _.filter(structuralVariantData.result, d => {
+            // We assume when one of the gene partners is missing, it should be intragenic deletion.
+            const isIntragenic =
+                d.variantClass &&
+                (d.site1EntrezGeneId === d.site2EntrezGeneId ||
+                    d.site1EntrezGeneId === undefined ||
+                    d.site2EntrezGeneId === undefined);
+            const otherValidQuery =
```
alisman, line 856: this name could be improved.  like, what about, ```isAnnotated``` or ```atLeastOneAnnotated```

i also think using the ```in``` operator would be clearer than ```!!```

And this validation filter should really be have a unit test.

zhx828: All valid points. Before making any changes, I do want to get some feedback from @kalletlak. I do believe intragenic events should be properly marked with Deletion as VariantClass. But it looks like it's not always the case. So I do not check the variant class anymore. As long as one of the genes is missing, I consider it intragenic.

zhx828:
Got feedback from Karthik. I'm going to update the backend logic so that they can all reside on the same place. The frontend should not have too much custom logic.

zhx828:
Accidentally force pushed the local changes and resulted the pr closed. The one pr is https://github.com/cBioPortal/cbioportal-frontend/pull/3938

PR #3926, opened 2021-08-16
Fix bug where oncoprint profiled-in tracks could not all be deleted

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8833

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -876,25 +876,25 @@ export class ResultsViewPageStore
     });
 
     @computed.struct get comparisonGroupsReferencedInURL() {
-        const clinicalTracksParam = this.urlWrapper.query.clinicallist;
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -876,25 +876,19 @@ export class ResultsViewPageStore
...
-                    convertComparisonGroupClinicalAttribute(
-                        clinicalAttributeId,
-                        false
-                    )
-                ); // convert track ids to group ids
-            return groupIds;
-        } else {
-            return [];
-        }
+        const groupIds = this.urlWrapper.oncoprintSelectedClinicalTracks
```
alisman, line 879: dang we should add comment explaining why groups are in clinical track list.

alisman (approved):

tmazor (approved):
looks good!

PR #3927, opened 2021-08-16
Study view chart button fixes

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8785

image

### DIFF
--- a/src/pages/studyView/charts/ChartContainer.tsx
+++ b/src/pages/studyView/charts/ChartContainer.tsx
@@ -556,6 +556,11 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
                             onChangeSelectedRows={
                                 this.handlers.onChangeSelectedRows
                             }
+                            extraButtons={
+                                this.comparisonButtonForTables && [
+                                    this.comparisonButtonForTables,
+                                ]
+                            }
                             selectedRowsKeys={this.selectedRowsKeys}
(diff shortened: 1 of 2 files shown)


src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -365,7 +365,15 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
                             className={classnames('fa fa-fw')}
                             style={{ marginRight: 4 }}
                         />
-                        Compare
+                        <span
```
alisman, line 368: @adamabeshouse i think better to solve this by using display:flex btn-group div.  This seems to prevent the wrapping in my experiment

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -556,6 +556,11 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
                             onChangeSelectedRows={
                                 this.handlers.onChangeSelectedRows
                             }
+                            extraButtons={
```
alisman, line 559: @adamabeshouse i think this is defunct, no?

alisman: i mean why is this change necessary?  i thought this was just a layout issue

alisman (approved):

PR #3928, opened 2021-08-17
Fix alteration menu localdb e2e tests

alisman:


end-to-end-test/local/specs/core/comparison-alterations-tab.spec.js
```diff
@@ -43,7 +43,7 @@ if (useExternalFrontend) {
             $('[data-test=LazyMobXTable]').waitForDisplayed();
             var rows = $$('[data-test=LazyMobXTable] tbody tr');
             assert.strictEqual(rows.length, 8, 'table has 8 rows');
-            openAlterationTypeSelectionMenu();
```
alisman, line 46: menu is already open from before block.

adamabeshouse: I have found that relying on this is a recipe for flaky tests, thats why I made `setDropdownOpen` which is more robust

I suggest, instead of deleting this line, update the method:

```
function openAlterationTypeSelectionMenu() {
    setDropdownOpen(
        true,
        '[data-test=AlterationEnrichmentTypeSelectorButton]',
        '[data-test=AlterationTypeSelectorMenu],
    );
} 
```

PR #3930, opened 2021-08-17
Add gene panel charts

Rajat-Sirohi:
Fix cBioPortal/cbioportal#6762

## Progress Notes
So far I've added the gene panel charts under the 'Genomic' tab + icons (with tooltips) next to each Data Profile. Filtering functionality has also been completed for the pie charts. What remains to be done is:
- [ ] Fix the group comparison options on the gene panel charts. I've added some dummy logic to `StudyViewPageStore:createNumberAttributeComparisonSession` which calls `StudyViewUtils:getGenePanelDataAsClinicalData` but I'm not entirely sure what that latter function should be returning. I've tried to model it after `getGenomicDataAsClinicalData`
- [ ] Address issue where gene panel charts seem to be reloading too many times
- [ ] Address this comment potentially following my suggestion here

## Changes
- changed 'Genomic Profile Sample Counts' to 'Data Profiles'
- added icon next to each data profile, if applicable, which toggles the visibility of its associated gene panel chart
- icon is black by default and turns blue when its associated gene panel chart becomes visible
- gene panel charts filter normally like pie charts
- one chart for each type of data profile (e.g., fusion, mutation, cna) with corresponding pie chart of samples, each of which is classified into one of three categories: 1) not being profiled by the specified molecular profile 2) not having any associated gene panels 3) associated gene panel

## Screenshots
add_charts_dropdown
charts
full


src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -190,25 +193,35 @@ export class MultiSelectionTable extends React.Component<
...
+                                        }
+                                        style={{
+                                            cursor: 'pointer',
+                                            marginLeft: 10,
+                                            color: this.props.visibleChartIds.includes(
+                                                getGenePanelChartUniqueKey(
+                                                    data.uniqueKey
+                                                )
+                                            )
+                                                ? '#0000ff'
```
zhx828, line 220: Should the color coming from the Colors.ts?

zhx828 (approved):
Looks good to me, just a minor comment.

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -87,6 +88,8 @@ export type MultiSelectionTableProps = {
     genePanelCache: MobxPromiseCache<{ genePanelId: string }, GenePanel>;
     filterByCancerGenes: boolean;
     onChangeCancerGeneFilter: (filtered: boolean) => void;
+    onClickPieChartIcon?: (profileId: string) => void;
```
alisman, line 91: onClickAddChartIcon?   i just don't see reason to specify the chart type

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -190,25 +193,35 @@ export class MultiSelectionTable extends React.Component<
...
                 render: (data: MultiSelectionTableRow) => {
                     return (
                         <div className={styles.labelContent}>
-                            <EllipsisTextTooltip
-                                text={data.label}
-                            ></EllipsisTextTooltip>
+                            <EllipsisTextTooltip text={data.label} />
+                            {this.props.onClickPieChartIcon &&
+                                this.props.visibleChartIds && (
+                                    <i
```
alisman, line 205: i kinda think this icon needs a tooltip to explain what it does (add chart)

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -190,25 +193,35 @@ export class MultiSelectionTable extends React.Component<
...
+                                        className="fa fa-pie-chart"
+                                        onClick={() =>
```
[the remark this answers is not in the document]
Rajat-Sirohi: How about this?
```suggestion
                                        onClick={this.onClickAddChartIcon(data)}
```
and
```
onClickAddChartIcon = (data: MultiSelectionTableRow) => () => {
    this.props.onClickAddChartIcon!(data.uniqueKey);
}
```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -410,6 +413,28 @@ export class StudyViewPageStore
         );
     }
 
+    private setCustomCharts(chartMeta: ChartMetaWithDimensionAndChartType) {
+        const uniqueKey = chartMeta.uniqueKey;
+        if (!this.chartsType.has(uniqueKey)) {
```
alisman, line 418: maybe put a comment here explaining this validation

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -529,30 +554,46 @@ export class StudyViewPageStore
             )
         );
 
+        // Add special gene panel charts for each molecular profile
+        this.reactionDisposers.push(
+            reaction(
+                () => this.molecularProfileNameSet.isComplete,
```
alisman, line 560: i wonder if you could just observe ```this.molecularProfileNameSet.result ```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -529,30 +554,46 @@ export class StudyViewPageStore
...
+                                dataType: ChartMetaDataTypeEnum.GENOMIC,
+                                patientAttribute: false,
+                                dimension: {
+                                    w: 1,
+                                    h: 1,
+                                },
+                                priority: 999,
+                                renderWhenDataChange: false,
+                            };
+                            this.setCustomCharts(chartMeta);
```
alisman, line 586: i assume that if it tries to add an already existing chart type, it will just lay off?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2535,6 +2576,22 @@ export class StudyViewPageStore
                         chartUniqueKey,
                         clinicalDataFilter
                     );
+                } else if (isGenePanelChart(chartUniqueKey)) {
+                    const filteredSampleIdentifiers = getFilteredSampleIdentifiers(
+                        this.samples.result.filter(sample =>
+                            values.includes(
```
alisman, line 2582: i wonder if we could improve the name ```values```.   i guess it's genePanelIds?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3795,13 +3863,120 @@ export class StudyViewPageStore
         );
     }
 
+    @observable private sampleIdToGenePanelId: {
```
alisman, line 3866: you might want to explicitly initlaize this  with an observable map.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3795,13 +3863,120 @@ export class StudyViewPageStore
...
+            );
+            const molecularProfileNameSet = this.molecularProfileNameSet
+                .result!;
+            const genePanelData: {
+                [uniqueKey: string]: MobxPromise<ClinicalDataCountSummary[]>;
+            } = {};
+
+            for (const profileId of Object.keys(molecularProfileNameSet)) {
+                const uniqueKey = getGenePanelChartUniqueKey(profileId);
+                genePanelData[uniqueKey] = remoteData<
```
alisman, line 3891: probably not a good idea to return remoteData inside a remoteData

alisman: i guess we're doing this down below.  hmmm

Rajat-Sirohi: Technically `genePanelData` (the outer remoteData) isn't really remote data. It just needs to be a promise since I'm waiting for some other promises to complete

alisman: i guess it allows us to track the individual loading state of the nested promises.  do we actually use that ability?

Rajat-Sirohi: If you're talking about the nested remote data, then I need each return value to be its own `MobxPromise` which is why I use remote data. See line 3977 in `StudyViewPageStore`: `return this.genePanelData.result[uniqueKey]`. So I access and return each promise individually

src/pages/studyView/StudyViewPageStore.ts
Rajat-Sirohi: `values` is a generic parameter for any type of chart. For my gene panel charts, it happens to be genePanelIds, but it can be any array of strings, so I think it's good to keep the generic name

src/pages/studyView/StudyViewPageStore.ts
Rajat-Sirohi: If the chart specified by `chartMeta` already exists in `chartTypes`, then this function will override that as well as whatever is stored in `this._customCharts` for the given key. I didn't write the function (I just extracted the code into its own function), so I don't think I should mess with it

src/pages/studyView/table/MultiSelectionTable.tsx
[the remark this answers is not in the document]
adamabeshouse: @Rajat-Sirohi I think that would work as long as `this.onClickAddChartIcon` is memoized. otherwise its the same problem.

Another way to handle this would be to wrap the render body in a component that takes the callback and the data as props, and on the inside it creates its own bound method. Then you would have to add an appropriate `key` for the row, I think

Luke-Sikina (approved):

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -136,6 +142,11 @@ export class MultiSelectionTable extends React.Component<
         this.sortBy = this.props.defaultSortBy;
     }
 
+    @memoize
```
adamabeshouse, line 145: nice, just want to confirm the memoization works since the argument is an object. is there a need to pass a memoization key?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3795,13 +3876,119 @@ export class StudyViewPageStore
         );
     }
 
+    private sampleIdToGenePanelId = observable.map<string, string>();
+
+    readonly genePanelData = remoteData<{
```
kalletlak, line 3881: All this logic should be done backend. Need to create new study-view API to support through backend. Also need to support it in StudyViewFilter object

PR #3932, opened 2021-08-18
RFC 59 - show studies for which a user is not authorized

dianab0:
Rfc 59 (https://docs.google.com/document/d/1NkNhBG1sk0ZvCOwk3XtiuRT9a_tKkFCgXC9I3XT-Ye4/edit#heading=h.6c2l76kgw1ta)
Physical studies that are not authorized are grayed out, have a gray info icon with no tooltip and a lock icon at then end. This behavior can be enabled/disabled from portal.properties (by default false). Also the global message that appears when hover over lock icon can be configured in portal.properties as well.

Screenshot from 2021-08-17 12-24-03

Backend PR is https://github.com/cBioPortal/cbioportal/pull/8847

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -4120,7 +4120,14 @@ export class StudyViewPageStore
             return _.reduce(
                 this.studyIds,
                 (acc: CancerStudy[], next) => {
-                    if (everyStudyIdToStudy[next]) {
+                    if (
+                        everyStudyIdToStudy[next] &&
+                        (!AppConfig.serverConfig
+                            .skin_show_unauthorized_studies ||
+                            (AppConfig.serverConfig
+                                .skin_show_unauthorized_studies &&
+                                everyStudyIdToStudy[next].isAuthorized == true))
+                    ) {
                         acc.push(everyStudyIdToStudy[next]);
                     }
                     return acc;
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -237,6 +237,7 @@ import {
     CopyNumberEnrichmentEventType,
     MutationEnrichmentEventType,
 } from 'shared/lib/comparison/ComparisonStoreUtils';
+import { isQueriedStudyAuthorized } from 'shared/components/lazyMobXTable/utils';
 
 type ChartUniqueKey = string;
 type ResourceId = string;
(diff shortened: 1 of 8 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4120,7 +4120,14 @@ export class StudyViewPageStore
             return _.reduce(
                 this.studyIds,
                 (acc: CancerStudy[], next) => {
-                    if (everyStudyIdToStudy[next]) {
+                    if (
```
alisman, line 4123: lets put this condition in a helper function

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -4120,7 +4120,14 @@ export class StudyViewPageStore
...
+                        everyStudyIdToStudy[next] &&
+                        (!AppConfig.serverConfig
+                            .skin_show_unauthorized_studies ||
+                            (AppConfig.serverConfig
+                                .skin_show_unauthorized_studies &&
+                                everyStudyIdToStudy[next].isAuthorized == true))
```
alisman, line 4129: use ===

alisman: == in javascript means type coercion.  it's just a good practice to use === as a rule

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -367,27 +372,37 @@ export default class StudyList extends QueryStoreComponent<
...
-                                    isVirtualStudy={this.store.isVirtualStudy(
-                                        study.studyId
-                                    )}
-                                    mouseEnterDelay={0}
-                                    placement="top"
-                                >
-                                    <a>{content}</a>
-                                </StudyTagsTooltip>
-                            );
+                            if (
```
alisman, line 375: again, lets wrap in a helper function.

dianab0:
Thank you for review @alisman. I updated the code. I added the helper function in both StudyViewPageStore.ts and StudyList.tsx. However in the StudyList we may remove it the future if we'll implement, as per RFC, the retrieval of info icon data for unauthorized studies too.

dianab0:
@alisman sorry, I forgot to push the commit containing the updates after review. It is pushed now.

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -66,6 +66,14 @@ export default class StudyList extends QueryStoreComponent<
         if (this.props.showSelectedStudiesOnly) this._view = this.view;
     }
 
+    isQueriedStudyAuthorized = (study: CancerStudy): boolean => {
```
alisman, line 69: can we put this in utils so we don't define it twice?

PR #3938, opened 2021-08-19
Do not filter out structural variant when one of the genes is missing

zhx828:
This fixes https://github.com/cBioPortal/cbioportal/issues/8600


kalletlak (approved):

packages/cbioportal-utils/src/oncokb/OncoKbUtils.ts
```diff
@@ -250,14 +250,18 @@ export enum StructuralVariantType {
 
 export function generateAnnotateStructuralVariantQueryFromGenes(
     site1EntrezGeneId: number,
-    site2EntrezGeneId: number,
+    site2EntrezGeneId: number | undefined,
```
alisman, line 253: is it possible that site1 could be missing?  or by definition, no?  sorry if this is stupid question

packages/cbioportal-utils/src/oncokb/OncoKbUtils.ts
```diff
@@ -250,14 +250,18 @@ export enum StructuralVariantType {
...
     tumorType: string | null,
     structuralVariantType: keyof typeof StructuralVariantType,
     evidenceTypes?: EvidenceType[]
 ): AnnotateStructuralVariantQuery {
     // For most of the SV in the portal, we can assume they are fusion event. The assumption is based on that user generally will not import none fusion event in database.
     let isIntragenic = false;
-    if (site1EntrezGeneId === site2EntrezGeneId || site2EntrezGeneId === null) {
+    if (
```
alisman, line 260: how about 
```
const isIntragenic = ternary
```
also please add a comment what intragenic means.

PR #3942, opened 2021-08-23
Treatment groups

Luke-Sikina:
Fix cBioPortal/cbioportal#8834

- Adds 2 new charts to the study view for treatment groups:
  - sample treatment groups
  - patient treatment groups
- Adds matching filters to study view filter applier

Here are some pictures:
Screenshot from 2021-09-16 16-21-57
Screenshot from 2021-09-16 16-22-15
Screenshot from 2021-09-16 16-23-51


alisman:
@cBioPortal/product  @Luke-Sikina  I find the chart title "Sample Treatments" to be confusing.   If I understand correctly, the main concept of this filtering is temporal:  when was sample taken relative to treatment admin.   It would be good if the title conveyed that.  "Treatment Timeline" ?     
Having "Sample treatment" right next to "Patient treatment" creates a false dichotomy.   Maybe the the latter could be "Treatment received"?    
Lastly, i think if the items (pre/post) were grouped by treatment instead of ordered by size, that might be less confusing

src/pages/studyView/StudyViewComparisonUtils.ts
```diff
@@ -181,6 +181,8 @@ export function getComparisonParamsForTable(
     selectedRowsKeys: string[],
     chartType: ChartType
 ) {
+    console.log('ey');
```
alisman, line 184: console!

PR #3950, opened 2021-08-26
Fix the allele frequency issue in multi-sample patient view

zhx828:
This is to fix an issue introduced by https://github.com/cBioPortal/cbioportal-frontend/pull/3931

The allele frequency is not shown when the mutation only exists for one sample in the multi-sample patient view.


zhx828:

adamabeshouse (approved):
perfect, thanks!!

PR #3951, opened 2021-08-30
Upgrade to webpack 5

onursumer:


end-to-end-test/shared/specUtils.js
```diff
@@ -184,6 +184,16 @@ function goToUrlAndSetLocalStorage(url, authenticated = false) {
     if (!useExternalFrontend) {
         browser.url(url);
         console.log('Connecting to: ' + url);
+    } else if (useNetlifyDeployPreview) {
+        browser.url(url);
+        browser.execute(
+            function(config) {
+                this.localStorage.setItem('netlify', config.netlify);
```
onursumer, line 191: @alisman do you think it is worth adding support for a url parameter like `?netlify=...` in the backend instead of setting it in the browser?

PR #3953, opened 2021-08-30
Refactor signal component to fit expandable design

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/524


onursumer (approved):
LGTM!

PR #3954, opened 2021-09-01
Make signal table’s pagination controls more compact

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/524

Make the signal table’s pagination controls more compact. To be consistent with the other 2 tables.
image
vs
image


onursumer (approved):
👍

PR #3955, opened 2021-09-01
Include a tip when yarn start throws dependency error

zhx828:
If the project hasn't been built before, the webpack may have issues identifying the packages managed by lerna(monorepo). You want to do a `yarn build` first before starting the project.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ yarn run start
 Example pages:
  - http://localhost:3000/
  - http://localhost:3000/patient?studyId=lgg_ucsf_2014&caseId=P04
+> **Tip:** If you see dependency errors, especially the error that the script cannot identify the packages managed by lerna(monorepo), you could do a `yarn buildModules` first before starting the project.
 
 To run unit/integration tests
 ```


README.md
```diff
@@ -45,10 +45,10 @@ export BRANCH_ENV=master # or rc if branching from rc
 # export any custom external API URLs by editing env/custom.sh
 yarn run start
 ```
-
 Example pages:
  - http://localhost:3000/
  - http://localhost:3000/patient?studyId=lgg_ucsf_2014&caseId=P04
+> **Tip:** If you see dependency errors, especially the error that the script cannot identifying the packages managed by lerna(monorepo), you could do a `yarn build` first before starting the project.
```
onursumer, line 51: Running `yarn buildModules` alone may fix the issue and it is faster than `yarn build`, so maybe a tip like this is better?

```suggestion
> **Tip:** If you see dependency errors, especially the error that the script cannot identify the packages managed by lerna(monorepo), you could do a `yarn build` or `yarn buildModules` first before starting the project.
```

zhx828: @onursumer let's just include the `yarn buildModules` then, updated.

onursumer (approved):

PR #3957, opened 2021-09-02
Find the sample frequency in the list

zhx828:
The frequency list is populated based on all samples. When only one sample is selected, we need to find the sample index in the sample list.


adamabeshouse (approved):

zhx828:

PR #3960, opened 2021-09-10
Annotation tracks download

calvinlu3:
Fix cBioPortal/cbioportal#6044

* Removed .svg in filename. (ex. 'BRCA1_lollipop.svg.png' -> 'BRCA1_lollipop.png')
* Created SVG that contains the lollipop plot and selected tracks.
* Individual sub tracks will always show [[Issue from standup](https://github.com/cBioPortal/cbioportal/issues/6044#issuecomment-915476830)]

Screenshot examples:
All tracks with PTM track expanded
BRCA_page|BRCA1_lollipop 
All tracks with PTM track collapsed
Screenshot 2021-09-10 161821 BRCA1_lollipop (1)

### DIFF
--- a/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
+++ b/packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
@@ -50,6 +50,33 @@ import DomainTooltip from '../lollipopPlot/DomainTooltip';
 
 const DEFAULT_PROTEIN_LENGTH = 10;
 
+const SVG_NS = 'http://www.w3.org/2000/svg';
+function createTextTag(
+    textNode?: Text,
+    style: string = '',
+    x: number = 0,
+    y: number = 0
+) {
+    const text = (document.createElementNS(
+        SVG_NS,
+        'text'
+    ) as unknown) as SVGTextElement;
+    $(text).attr({ style, x, y });
+    if (textNode) {
+        text.append(textNode);
+    }
+    return text;
+}
+
+function createGroupTag(transform?: string) {
+    const group = (document.createElementNS(
+        SVG_NS,
+        'g'
+    ) as unknown) as SVGGElement;
+    group.setAttribute('transform', transform || '');
+    return group;
+}
+
 export type LollipopMutationPlotProps<T extends Mutation> = {
     store: MutationMapperStore<T>;
     controlsConfig?: LollipopPlotControlsConfig;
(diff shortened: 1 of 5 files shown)


packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -556,10 +556,76 @@ export default class LollipopMutationPlot<
...
             '.lollipop-svgnode'
         )[0] as any;
-        return svg;
+
+        const SVG_NS = 'http://www.w3.org/2000/svg';
+        const TEXT_STYLE = 'font-size: 12px; font-family:Arial;';
+        const TRACK_INDENT = 16;
+        const SUBTRACK_INDENT = TRACK_INDENT + 12;
+
+        function createTextTag(textNode?: Text, x: number = 0, y: number = 0) {
```
alisman, line 568: i would avoid the practice of defining functions inside of other functions/methods.   if you wanted to write a test for this for example, you could not do so.   just put it up at the top of this file or in a helper file which is associated with this file.

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -556,10 +556,76 @@ export default class LollipopMutationPlot<
...
+
+        function createGroupTag(transform?: string) {
+            const group = document.createElementNS(SVG_NS, 'g') as SVGGElement;
+            group.setAttribute('transform', transform || '');
+            return group;
+        }
+
+        var outerSvg = document.createElementNS(SVG_NS, 'svg') as SVGElement;
+        outerSvg.append(lollipopSvg.cloneNode(true));
+        this.divContainer.appendChild(outerSvg);
```
alisman, line 588: you should put a comment here explaining that you are putting it into DOM only in order to measure it.  you should also make an attempt to hide it so that there's no flash for the user.  perhaps by wrapping in a div that has visibility:hidden.    You might also consider putting it into the document.body rather than inside the div of this class.

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
alisman: again, this would allow you to export these functions.    also, defining function this way inhibits performance.  it's totally negligible here but best to avoid in general

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -556,10 +556,76 @@ export default class LollipopMutationPlot<
...
+        }
+
+        function createGroupTag(transform?: string) {
+            const group = document.createElementNS(SVG_NS, 'g') as SVGGElement;
+            group.setAttribute('transform', transform || '');
+            return group;
+        }
+
+        var outerSvg = document.createElementNS(SVG_NS, 'svg') as SVGElement;
+        outerSvg.append(lollipopSvg.cloneNode(true));
```
alisman, line 587: put the clone operation outside of the append to so you can write a comment like ```clone the chart for use in download```

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlot.tsx
```diff
@@ -560,33 +587,18 @@ export default class LollipopMutationPlot<
...
-        }
-
         var outerSvg = document.createElementNS(SVG_NS, 'svg') as SVGElement;
-        outerSvg.append(lollipopSvg.cloneNode(true));
-        this.divContainer.appendChild(outerSvg);
 
+        // Make copy of chart to be used in download
+        var chartClone = lollipopSvg.cloneNode(true);
+        outerSvg.append(chartClone);
+        document.body.appendChild(outerSvg);
```
alisman, line 599: you should put a comment here explaining why you are inserting it into DOM ("in order to measure it"). also, wrap it in a div <div class="invisible"></div> so that user doesn't see it

PR #3963, opened 2021-09-15
Fix bug with generic assay profile selection in comparison view

pvannierop:
# Problem
When selecting a new molecular profile in a Generic Assay Tab the frontend breaks when there are multiple Generic Assay types (white screen).

# Solution
This PR will fix the update of the `selectedGenericAssayEnrichmentProfileMapGroupedByGenericAssayType` data structure so that this problem does not longer occur.

# Remark
This problem cannot be reproduced on the public portal because a dataset with multiple Generic Assay types does not exist. I confirmed this problem and fix on a confidential data set. I cannot share movies in support of this PR.


adamabeshouse (approved):

dippindots (commented):
Hi @pvannierop, thanks for fixing this issue!
The tests failed at `install` step, it looks like there is a typescript error, could you try to look at it? Besides that, this looks good to me!

pvannierop:
@adamabeshouse I fixed the compilation error. Can we merge?

PR #3965, opened 2021-09-17
Only show entities associated with selected profile

dippindots:
Fix cBioPortal/cbioportal#8520

If there are more than one profiles falls into one `Generic Assay Type`, in this case, one tab has more than one profiles.
When you switch between profiles, entities should be changed based on your selection. Previously, we just show all entities in all profiles.

image

image


src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -99,12 +113,34 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
         super(props, context);
         makeObservable(this);
         this.tabsDivRef = React.createRef<HTMLDivElement>();
+
+        // Using the first profile to initialze genericAssayProfileOptionsByType
+        this.genericAssayEntityOptionsReaction = when(
```
alisman, line 118: interesting. haven't seen when used before

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -99,12 +113,34 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
         super(props, context);
         makeObservable(this);
         this.tabsDivRef = React.createRef<HTMLDivElement>();
+
+        // Using the first profile to initialze genericAssayProfileOptionsByType
```
alisman, line 117: just want to make sure, is it the first profile or entity?

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -433,13 +480,31 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
             (options, type) => {
+                const molecularProfileIdsInType = options.find(
+                    option =>
+                        option.value ===
+                        this.selectedGenericAssayProfileIdByType.get(type)
+                )!.profileIds;
+                const entities = _.reduce(
+                    molecularProfileIdsInType,
+                    (set, profileId) => {
+                        set = new Set([
```
alisman, line 491: this seems odd to create a new Set every iteration.  you cannot add them to existing set?

src/pages/studyView/addChartButton/AddChartButton.tsx
dippindots: This should be the first profile, we want to generate entity options based on a profile selection, so use the first profile as the default.

PR #3966, opened 2021-09-17
Fix germline rendering issue in oncoprint and add test

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8906


inodb (approved):
Sweet - that fixes it!

Once the screenshot is added we can merge it 👍

PR #3974, opened 2021-09-27
Only show entities that associated with selected profile

dippindots:
Fix cBioPortal/cbioportal#8520

If there are more than one profiles falls into one `Generic Assay Type`, in this case, one tab has more than one profiles.
When you switch between profiles, entities should be changed based on your selection. Previously, we just show all entities in all profiles.

image

image

### DIFF
--- a/end-to-end-test/local/specs/core/studyview.screenshot.spec.js
+++ b/end-to-end-test/local/specs/core/studyview.screenshot.spec.js
@@ -38,6 +38,7 @@ describe('study view generic assay categorical/binary features', function() {
         $('div[data-test="GenericAssaySelection"] input').setValue(
             'mutational_signature_category_10'
         );
+        $('div=Select all filtered options (1)').waitForExist();
         $('div=Select all filtered options (1)').click();
         // close the dropdown
         var indicators = $$('div[class$="indicatorContainer"]');
--- a/src/pages/studyView/addChartButton/AddChartButton.tsx
+++ b/src/pages/studyView/addChartButton/AddChartButton.tsx
@@ -1,5 +1,12 @@
 import * as React from 'react';
-import { action, computed, makeObservable, observable } from 'mobx';
+import {
+    action,
+    computed,
+    IReactionDisposer,
+    makeObservable,
+    observable,
+    when,
+} from 'mobx';
 import { observer } from 'mobx-react';
 import 'react-mfb/mfb.css';
 import {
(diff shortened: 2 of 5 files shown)


dippindots:
Last pull request has been closed because the base branch has been merged and deleted, this pr will address comments from https://github.com/cBioPortal/cbioportal-frontend/pull/3965

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -433,13 +480,27 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+                    option =>
+                        option.value ===
+                        this.selectedGenericAssayProfileIdByType.get(type)
+                )!.profileIds;
+                const entities = _.reduce(
+                    molecularProfileIdsInType,
+                    (set, profileId) => {
+                        this.props.store.genericAssayEntitiesGroupedByProfileId.result![
+                            profileId
+                        ].forEach(set.add, set);
```
alisman, line 493: never saw this syntax before.  neat.

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -433,13 +480,27 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
         tabs = _.map(
             this.props.store.genericAssayProfileOptionsByType.result,
             (options, type) => {
+                const molecularProfileIdsInType = options.find(
```
alisman, line 483: this code could really use some commenting

src/pages/studyView/addChartButton/AddChartButton.tsx
```diff
@@ -433,13 +480,27 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
...
+                    option =>
+                        option.value ===
+                        this.selectedGenericAssayProfileIdByType.get(type)
+                )!.profileIds;
```
alisman, line 487: i take it we can be sure we will find one because the list was made using ...

src/pages/studyView/addChartButton/AddChartButton.tsx
dippindots: Sure, I will add some comments here

src/pages/studyView/addChartButton/AddChartButton.tsx
dippindots: Yes, because this.selectedGenericAssayProfileIdByType is generated by using `this.props.store.genericAssayProfileOptionsByType.result`, so we can be sure we can always find one.

I will add some comments at here too.

end-to-end-test/local/specs/core/studyview.screenshot.spec.js
```diff
@@ -36,7 +36,7 @@ describe('study view generic assay categorical/binary features', function() {
         // and select a option
         $('div[data-test="GenericAssaySelection"]').waitForExist();
         $('div[data-test="GenericAssaySelection"] input').setValue(
-            'mutational_signature_category_10'
```
dippindots, line 39: I need to update this line since we use the first profile as default option, and the first profile is `pvalue profile`, and we cannot find `mutational_signature_category_10` anymore when we selected `pvalue profile`

PR #3992, opened 2021-10-05
Make numerical interval filter breadcrumbs in study view editable

adamabeshouse:
editable-breadcrumbs


alisman:
@adamabeshouse i made a tweak to the style of the EditableSpan on the study page and force pushed (after rebase).  I noticed one other thing:  when you submit a change to the range, the old number briefly flashes inside the text input.  obviously this is some state management issue.  but it's noticeable enough that we should try to fix it.

src/pages/studyView/UserSelections.tsx
```diff
@@ -451,6 +493,21 @@ export default class UserSelections extends React.Component<
                               )
                             : this.renderDataBinFilter(
                                   clinicalDataFilter.values,
+                                  async (
```
alisman, line 496: why is it async?

packages/cbioportal-frontend-commons/src/components/editableSpan/EditableSpan.tsx
```diff
@@ -46,12 +49,12 @@ export default class EditableSpan extends React.Component<
         }
 
         if (this.props.numericOnly) {
-            const regex = /^\d$/;
+            const regex = /^[\d.-]$/;
```
alisman, line 52: what's that dash in the regex?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1392,37 +1415,130 @@ export function intervalFiltersDisplayValue(values: DataFilterValue[]) {
             ? numericals[numericals.length - 1].end
             : undefined;
 
-    let displayValues: string[] = [];
+    let displayValues: any[] = [];
 
     if (numericals.length > 0) {
+        const startValue = formatValue(start!);
```
alisman, line 1421: it's odd that you use the ! even though above (1412) they are allowed to be undefined?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -1392,37 +1415,130 @@ export function intervalFiltersDisplayValue(values: DataFilterValue[]) {
...
-                `${formatValue(start)} < x ≤ ${formatValue(end)}`
-            );
+            intervalDisplayValues.push(startText);
+            intervalDisplayValues.push(` < x ≤ `);
+            intervalDisplayValues.push(endText);
         }
+
+        displayValues.push(
+            stringOutput
+                ? intervalDisplayValues.join('')
```
alisman, line 1519: it seems odd that you either push a string or an array of strings.  is there some flattening later?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3354,6 +3470,66 @@ export function getUserGroupColor(
         : undefined;
 }
 
+export async function updateCustomIntervalFilter(
```
alisman, line 3473: i think this needs a high level comment on what exactly this does.

alisman: it also could really use some unit testing : )

packages/cbioportal-frontend-commons/src/components/editableSpan/EditableSpan.tsx
adamabeshouse: thats a negative sign

src/pages/studyView/StudyViewUtils.tsx
adamabeshouse: the array would actually be of jsx elements (e.g. startText which is actually an EditableSpan), so its a different use case

src/pages/studyView/UserSelections.tsx
adamabeshouse: oh Im sure thats from before I refactored pulling the actual asynchronous code out into `updateCustomIntervalFilter`

src/pages/studyView/StudyViewUtils.tsx
adamabeshouse: the case when it is undefined is when the selected filters don't include numerical. e.g. in some bar charts you can have NA, a categorical filter in an otherwise numerical attribute. if you are only filtering for NA then you wouldn't have any numerical.

src/pages/studyView/StudyViewUtils.tsx
adamabeshouse: yes, both are good ideas.

src/pages/studyView/StudyViewUtils.tsx
alisman: what i meant is, i take the ! to mean you are telling typescript that it is NOT undefined.  but it can be undefined, which would suggest that formatValue can handle being passed undefined, in which case why the !      
sorry if i misunderstand

src/pages/studyView/StudyViewUtils.tsx
alisman: i see.  so like a component, you can pass it children that are either string or array of things

src/pages/studyView/StudyViewUtils.tsx
adamabeshouse: ah I see. yes so formatValue can handle an undefined but if you look at the declaration it has multiple type signatures so it knows that a defined number input leads to a string. so you can look at `const startValue = formatValue(start!)` as equivalent to `const startValue = formatValue(start) as string;`

PR #3998, opened 2021-10-12
Embed cancer gene list in project until caching behavior improved

alisman:


alisman:
@zhx828 @inodb are we ok with this if i make an issue to update it once oncokb on aws?

inodb:
@alisman Can we make this opt-in thru some property? Note that this list might be different for internal instance of oncokb vs the public one. So maybe also double check if that is the case. And if it is indeed different, if it is ok to make this data public.

PR #4003, opened 2021-10-15
Dont wait for selectedSamples to make several API calls in study view

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8979

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -157,7 +157,7 @@ import {
     getStudyDownloadListUrl,
     redirectToComparisonPage,
 } from '../../shared/api/urls';
-import onMobxPromise from '../../shared/lib/onMobxPromise';
+import onMobxPromise, { toPromise } from '../../shared/lib/onMobxPromise';
 import request from 'superagent';
 import { trackStudyViewFilterEvent } from '../../shared/lib/tracking';
 import comparisonClient from '../../shared/api/comparisonGroupClientInstance';
@@ -3659,13 +3659,8 @@ export class StudyViewPageStore
     }
 
     readonly unfilteredClinicalDataCount = remoteData<ClinicalDataCountItem[]>({
-        await: () => [this.selectedSamples],
         invoke: () => {
-            //only invoke if there are filtered samples
-            if (
-                this.hasFilteredSamples &&
-                !_.isEmpty(this.unfilteredAttrsForNonNumerical)
-            ) {
+            if (!_.isEmpty(this.unfilteredAttrsForNonNumerical)) {
                 return internalClient.fetchClinicalDataCountsUsingPOST({
                     clinicalDataCountFilter: {
                         attributes: this.unfilteredAttrsForNonNumerical,
@@ -3718,13 +3713,8 @@ export class StudyViewPageStore
     readonly newlyAddedUnfilteredClinicalDataCount = remoteData<
         ClinicalDataCountItem[]
     >({
-        await: () => [this.selectedSamples],
         invoke: () => {
-            //only invoke if there are filtered samples
-            if (
-                this.hasFilteredSamples &&
-                !_.isEmpty(this.newlyAddedUnfilteredAttrsForNonNumerical)
-            ) {
+            if (!_.isEmpty(this.newlyAddedUnfilteredAttrsForNonNumerical)) {
                 return internalClient.fetchClinicalDataCountsUsingPOST({
                     clinicalDataCountFilter: {
                         attributes: this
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3860,10 +3849,6 @@ export class StudyViewPageStore
                             this._clinicalDataFilterSet.has(uniqueKey) ||
                             this.isInitialFilterState
                         ) {
-                            // return empty if there are no sample identifiers or study ids in the filter
-                            if (!this.hasFilteredSamples) {
```
adamabeshouse, line 3864: @alisman this is an example of what we were talking about - hasFilteredSamples references selectedSamples, so selectedSamples was being awaited in order to make sure hasFilteredSamples was ready. too opaque

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8008,47 +7981,42 @@ export class StudyViewPageStore
...
-                return [];
-            }
-            const counts = await internalClient.fetchCaseListCountsUsingPOST({
-                studyViewFilter: this.filters,
-            });
+            const [counts, selectedSamples] = await Promise.all([
+                internalClient.fetchCaseListCountsUsingPOST({
+                    studyViewFilter: this.filters,
+                }),
+                toPromise(this.selectedSamples),
```
adamabeshouse, line 8011: here we're taking selectedSamples out of await, and awaiting it in the body of invoke

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8008,47 +7981,42 @@ export class StudyViewPageStore
...
+            ]);
 
             return counts.map(caseListOption => {
                 return {
                     uniqueKey: caseListOption.value,
                     label: caseListOption.label,
                     numberOfAlteredCases: caseListOption.count,
-                    numberOfProfiledCases: this.selectedSamples.result.length,
+                    numberOfProfiledCases: selectedSamples.length,
```
alisman, line 7998: is it odd that we assume all seelctedSamples are profiled here?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8008,47 +7981,42 @@ export class StudyViewPageStore
...
-            }
-            const counts = await internalClient.fetchMolecularProfileSampleCountsUsingPOST(
-                {
+            const [counts, selectedSamples] = await Promise.all([
+                internalClient.fetchMolecularProfileSampleCountsUsingPOST({
                     studyViewFilter: this.filters,
-                }
-            );
+                }),
+                toPromise(this.selectedSamples),
```
alisman, line 7990: this is sort of the heart of this PR here, right?  We need to use all so that we can concurrently load selectedSamples

alisman: what's kinda weird though is that selectedSamples itself depends on filters.  maybe it's not weird.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8938,12 +8905,9 @@ export class StudyViewPageStore
     // a row represents a list of patients that either have or have not recieved
     // a specific treatment
     public readonly sampleTreatments = remoteData({
-        await: () => [this.selectedSamples, this.displaySampleTreatments],
+        await: () => [this.displaySampleTreatments],
```
alisman, line 8908: i think displaySampleTreatments is a boolean?  maybe we should rename it to shouldDisplaySampleTreatments?

alisman: and while we're at it shouldDisplayPatientTreatments?

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: in this case, yes - in other cases, we can just remove the selectedSamples dependency completely

src/pages/studyView/StudyViewPageStore.ts
alisman: but why did we want to short circuit in case of not having filtered samples?  was that an optimization?

alisman: i realize that you didn't create that code but maybe we should understand exactly why that was

adamabeshouse: as far as I understand, it is just to not make unnecessary calls to the server

src/pages/studyView/StudyViewPageStore.ts
adamabeshouse: added comment

PR #4004, opened 2021-10-18
Add uniprot topology track

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/547

### Goal
Add extracellular info from uniprot. For example their topology track shows cytoplasmic vs extracellular in the description of the tooltip. We group them by topological domains and adding as individual tracks.

### Test
https://deploy-preview-4004--cbioportalfrontend.netlify.app/results/mutations?cancer_study_list=metastatic_solid_tumors_mich_2017&tab_index=tab_visualize&case_set_id=metastatic_solid_tumors_mich_2017_all&Action=Submit&gene_list=SCN5A%2520EGFR&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cfusion&geneset_list=%20

Click `Add annotation tracks` dropdown, select `Uniprot Topology`. 
`SCN5A` is an example with complex domains, switch to `EGFR` to see a less complex track.

#### Expand
Tracks are expanded by default, click track name `Topology` to collapse or expand tracks.
image

#### Collapse
image

#### Track Tooltip
image

#### Track description tooltip
image

### DIFF
--- a/packages/react-mutation-mapper/src/component/track/TrackItem.tsx
+++ b/packages/react-mutation-mapper/src/component/track/TrackItem.tsx
@@ -27,13 +27,11 @@ export type TrackItemSpec = {
     label?: string;
     color?: string;
     tooltip?: JSX.Element;
+    height?: number;
 };
 
 @observer
-export default class TrackItem extends React.Component<
-    TrackItemProps,
-    TrackItemSpec
-> {
+export default class TrackItem extends React.Component<TrackItemProps> {
     constructor(props: any) {
         super(props);
         makeObservable(this);
@@ -85,6 +83,7 @@ export default class TrackItem extends React.Component<
                 <TrackRect
                     isHovered={this.isHovered}
                     width={this.props.dim1}
+                    height={this.props.dim2}
                     {...this.props}
                 />
             );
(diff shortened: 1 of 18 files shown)


packages/react-mutation-mapper/src/component/track/TrackItem.tsx
```diff
@@ -27,6 +27,7 @@ export type TrackItemSpec = {
     label?: string;
     color?: string;
     tooltip?: JSX.Element;
+    rectHeight?: number; // set rectangle height (only works if itemType is RECTANGLE)
```
onursumer, line 30: Can we try using `dim2` instead of introducing a new height prop?

https://github.com/cBioPortal/cbioportal-frontend/blob/3675a07a438ad31562b5945e22f71d1be8e7de55/packages/react-mutation-mapper/src/component/track/TrackItem.tsx#L11

Looks like it is hardcoded to `22` in `Track.tsx`. So maybe we can set its default value to `22`, and set it to `15` for the new uniprot track?

https://github.com/cBioPortal/cbioportal-frontend/blob/639003804a8d68bea8f2fd1a0e53e8a28dff7a73/packages/react-mutation-mapper/src/component/track/Track.tsx#L220

onursumer (changes requested):

onursumer (approved):

onursumer:
failing tests are not related, merging

PR #4018, opened 2021-11-01
Fix news markdown images

inodb:
Do not escape markdown. The HTML Parser was removed from react markdown b/c it was buggy, instead one now needs to use some plugins to allow for it: https://github.com/remarkjs/react-markdown/blob/main/changelog.md#remove-buggy-html-in-markdown-parser.

~This didn't work: https://stackoverflow.com/a/68722053~


onursumer (approved):
🙌

PR #4028, opened 2021-11-03
Add new how-to video to tutorials page

alisman:


inodb (approved):
Looks good! Should we add a section to the body of the page as well similar to the other webinar videos?

alisman:
@inodb this from Tali:

"i think just add it to the list at the top of the page under a new section called ‘How to videos’ (or something like that). since the plan is to do a bunch of these, i think we agreed to not embed the video on the page, but just list it and have a link out"

PR #4031, opened 2021-11-03
Fix nightly build

onursumer:
Fix cBioPortal/cbioportal#9028

- We forgot to add the netlify step for `nightly` job, this PR fixes that.
- Add a fallback to `master` build if we cannot find a valid netlify build for a given commit hash on master branch.

## Checks

### DIFF
--- a/scripts/find_netlify_deploy_id.js
+++ b/scripts/find_netlify_deploy_id.js
@@ -16,15 +16,25 @@ async function findDeployId(search) {
 async function main() {
     const args = process.argv;
     const search = args.length > 2 ? args[2] : undefined;
-    const deployId = search ? await findDeployId(search) : undefined;
+    const branch = args.length > 3 ? args[3] : undefined;
+    let deployId = search ? await findDeployId(search) : undefined;
 
     if (!deployId) {
-        // exit with a non-zero value to indicate error
-        process.exit(1);
+        if (isMasterBranch(branch)) {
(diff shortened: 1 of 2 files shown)


inodb (approved):
Makes sense

scripts/find_netlify_deploy_id.js
```diff
@@ -16,15 +16,25 @@ async function findDeployId(search) {
...
     const search = args.length > 2 ? args[2] : undefined;
-    const deployId = search ? await findDeployId(search) : undefined;
+    const branch = args.length > 3 ? args[3] : undefined;
+    let deployId = search ? await findDeployId(search) : undefined;
 
     if (!deployId) {
-        // exit with a non-zero value to indicate error
-        process.exit(1);
+        if (isMasterBranch(branch)) {
+            // we are on master branch, we can use the master build if no other build is available
```
inodb, line 24: @onursumer any clue in what situation there would be no build id? Is it possible that e.g. circleci starts before netlify has a build going?

onursumer: We wait for the netlify build in a previous step, so at this point it should be ready or error or not found.

I am not sure how things exactly work for the master branch. My guess is that it might be possible for a certain commit hash not to have a netlify instance because the newer merge to master branch may override it.

alisman (approved):

PR #4033, opened 2021-11-04
Fix mobx error

dippindots:
makeObservable called but no observable variables will break the component.


inodb (approved):
Wasn't causing an issue for me for some reason, but looks good to fix anyway

inodb:
Not sure how to label this exactly, it's mostly something that shows up while developing rather than a production bug. Maybe "cleanup" label would be good

PR #4034, opened 2021-11-05
Fix uniprot PTM crash when centain fields are missing in uniprot data

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9033

From Uniprot api doc, there are some option value in the model:
https://www.ebi.ac.uk/proteins/api/doc/#!/features/getByAccession
image
But based on my testing, `evidences` could be missing too, for example: https://www.ebi.ac.uk/proteins/api/features/P06276?categories=PTM
image

So we change `description`, `molecule` and `evidences` in UniprotFeature to optional, as well as `description` and `evidence` in UniprotTopology to optional.

### DIFF
--- a/packages/cbioportal-utils/src/ptm/PtmUtils.spec.ts
+++ b/packages/cbioportal-utils/src/ptm/PtmUtils.spec.ts
@@ -5,6 +5,8 @@ import {
     compareByPtmTypePriority,
     convertDbPtmToPtm,
     convertUniprotFeatureToPtm,
+    getPtmTypeFromUniprotFeature,
+    getPubmedIdsFromUniprotFeature,
     groupPtmDataByPosition,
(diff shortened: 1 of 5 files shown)


packages/cbioportal-utils/src/ptm/PtmUtils.spec.ts
```diff
@@ -341,4 +352,19 @@ describe('PtmUtils', () => {
             );
         });
     });
+    describe('partial Uniprot data', () => {
+        it('test data with partial of fields', () => {
```
onursumer, line 356: Very minor but might be better to have two separate `it`s. One named `uses default ptm type when description is empty`, and the other named `returns no pubmed ids when evidence is empty` or something like that.

onursumer (approved):
Looks good!

PR #4039, opened 2021-11-09
Fix missing co-expression results for miRNAs and phosphoproteins

onursumer:
Fix cBioPortal/cbioportal#8988

## Checks

### DIFF
--- a/src/pages/resultsView/coExpression/CoExpressionTab.tsx
+++ b/src/pages/resultsView/coExpression/CoExpressionTab.tsx
@@ -263,6 +263,7 @@ export default class CoExpressionTab extends React.Component<
     private coExpressionCache: CoExpressionCache = new CoExpressionCache(
         q => ({
             await: () => [
+                this.props.store.entrezGeneIdToGeneAll,
                 this.props.store.entrezGeneIdToReferenceGene,
                 this.props.store.molecularProfileIdToProfiledFilteredSamples,
             ],
@@ -355,6 +356,9 @@ export default class CoExpressionTab extends React.Component<
                         this.props.store.entrezGeneIdToReferenceGene.result ||
                         {};
 
+                    const entrezGeneIdToGeneMap =
+                        this.props.store.entrezGeneIdToGeneAll.result || {};
+
                     return Promise.resolve(
                         _.reduce(
                             data,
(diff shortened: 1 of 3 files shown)


src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -382,6 +386,16 @@ export default class CoExpressionTab extends React.Component<
...
+                                } else if (
+                                    entrezGeneIdToGene[
+                                        datum.geneticEntityId
+                                    ] !== undefined
+                                ) {
+                                    geneticEntityName =
+                                        entrezGeneIdToGene[
+                                            datum.geneticEntityId
+                                        ].hugoGeneSymbol;
+                                    cytoband = '';
```
dippindots, line 398: Is it ok to assign empty string to `cytoband`?
```
cytoband = '';
```

dippindots (commented):
@onursumer This pr looks good to me, I just have one question regarding to the `cytoband`.

And I'd like to test this fix manually, could you tell me how can I test and which part should I look at? Thanks!

src/pages/resultsView/coExpression/CoExpressionTab.tsx
onursumer: Good question. It looks empty in the tutorial video https://www.youtube.com/watch?v=YUxVv6pkxD4&t=2240s, but that may not always be the case.

src/pages/resultsView/coExpression/CoExpressionTab.tsx
```diff
@@ -355,6 +356,9 @@ export default class CoExpressionTab extends React.Component<
                         this.props.store.entrezGeneIdToReferenceGene.result ||
                         {};
 
+                    const entrezGeneIdToGene =
```
alisman, line 359: entrezGeneIdToGeneMap?

onursumer: Yeah, makes more sense. We already have another variable called `entrezGeneIdToReferenceGene`. We should probably rename that one as well.

inodb (commented):
Seems good to me but would be nice to confirm performance is similar

onursumer:
Failing e2e tests don't seem related, merging.

PR #4054, opened 2021-11-12
Generic Assay dropdown: don't show NA when name or description missing (use id instead)

dippindots:
Fix cBioPortal/cbioportal#9041

- Fix Generic Assay options show NA when name or description missing
- Add related unit tests

### DIFF
--- a/src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.spec.ts
+++ b/src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.spec.ts
@@ -88,6 +88,61 @@ describe('GenericAssayCommonUtils', () => {
             );
         });
 
+        it('Shows entity_stable_id and description when name is missing in properties', () => {
+            const genericAssayEntity: GenericAssayMeta = {
+                stableId: 'id_1',
+                entityType: 'GENERIC_ASSAY',
+                genericEntityMetaProperties: {
+                    DESCRIPTION: 'desc_1',
+                },
+            };
+
+            const expect = {
+                value: 'id_1',
+                label: 'id_1: desc_1',
+            };
+            assert.deepEqual(
+                makeGenericAssayOption(genericAssayEntity),
+                expect
+            );
+        });
+
+        it('Shows entity_stable_id and name when description is missing in properties', () => {
+            const genericAssayEntity: GenericAssayMeta = {
+                stableId: 'id_1',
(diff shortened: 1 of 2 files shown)


src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.spec.ts
```diff
@@ -88,6 +88,61 @@ describe('GenericAssayCommonUtils', () => {
             );
         });
 
+        it('Shows entity_stable_id and description when name is missing in properties', () => {
```
alisman, line 91: these tests seem to assert something different from what the name of the PR implies.  just curious how it works.

dippindots: @alisman Maybe rename title to `Generic Assay options should not show NA when name or description missing from genericEntityMetaProperties`

dippindots: This test is for `makeGenericAssayOption` function. These three tests will assert if `NAME` and `DESCRIPTION` property missing from `genericEntityMetaProperties`, will the label still correct.

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -195,11 +195,13 @@ export function makeGenericAssayOption(
     // indentical to the entity_stable_id.
     const name = getGenericAssayMetaPropertyOrDefault(
         meta,
-        COMMON_GENERIC_ASSAY_PROPERTY.NAME
+        COMMON_GENERIC_ASSAY_PROPERTY.NAME,
+        meta.stableId
     );
```
dippindots, line 200: The reason why it was showing `NA`:
getGenericAssayMetaPropertyOrDefault will return `NA` when an genericAssayMetaProperty is missing.
```
export function getGenericAssayMetaPropertyOrDefault(
    meta: GenericAssayMeta,
    property: string,
    defaultValue: string = NOT_APPLICABLE_VALUE
): string {
    if (property in meta.genericEntityMetaProperties) {
        return (meta.genericEntityMetaProperties as {
            [property: string]: string;
        })[property];
    } else {
        return defaultValue;
    }
}
```

alisman (approved):

PR #4068, opened 2021-12-01
X vs Y scatter charts now give the correct data download

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/9035

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -87,6 +87,7 @@ import {
     geneFilterQueryFromOql,
     geneFilterQueryToOql,
     generateScatterPlotDownloadData,
+    generateXVsYScatterPlotDownloadData,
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -254,6 +256,7 @@ import {
     StudyPageSettings,
     VirtualStudy,
 } from 'shared/api/session-service/sessionServiceModels';
+import client from 'shared/api/cbioportalClientInstance';
```
dippindots, line 259: Looks like we don't need to import `client` at here

dippindots (approved):
Looks good to me, I also tested this changes in browser. Chart data download is working as expected. Only one very minor comment, thanks!

tmazor (approved):
Looks great! And you fixed the file names so they aren't just txt.txt and svg.svg :)

PR #4069, opened 2021-12-01
Add NDEx Pathways Integration 

inodb:
Adds NDEx pathways on pathways tab in the results view page. Related to https://github.com/cBioPortal/cbioportal/issues/9085.

Test URL: https://bit.ly/3vddazA

Screen Shot 2022-04-26 at 8 54 06 AM

TODO:

- [x] add iframe with iquery page that searches for queried genes rather than just an iframe of the homepage
- [x] Do we want to combine pathwaymapper and NDEx as single tab, but using pills to navigate between them? So you would have one pill for pathwaymapper and one for NDEx inside of a pathways tab.
- [x] Explain difference between NDEx and PathwayMapper in UI

We decieded not to include other alteration frequencies yet as part of this release:

- current alteration frequency does not reflect queries study and subset alteration frequency. We prolly need to pass that over. Sample IDs, plus study ids. Or we pass over a session id.  Addressing later in: https://github.com/cBioPortal/cbioportal/issues/9305
    - decide interface (JSON format).
           ```
            "geneAnnotationServices": {
                             "mutation": "https://iquery-cbio.dev.ucsd.edu/integratedsearch/v1/mutationfrequency",
                             "alteration": "http://localhost/"
            }
           ``` 

Follow up with tests: https://github.com/cBioPortal/cbioportal/issues/9510

Let's use squash on merge button


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1585,6 +1585,50 @@ export class ResultsViewPageStore
...
+    readonly remoteNdexUrl = remoteData<string>({
+        await: () => [this.studyIds, this.genes],
+        invoke: async () => {
+            var result = '';
+
+            if (this.studyIds.result!.length > 0 && this.genes.result!.length > 0) {
+                const postData = {
+                    // this might be necessary at some point
+                    // studyid: this.studyIds.result![0],
+                    geneList: this.genes.result!.map(g => g.hugoGeneSymbol),
```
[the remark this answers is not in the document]
inodb: went for `alterationData:`

which contains something like:

```
altered: 1916
gene: "PIK3CA"
percentAltered: "17%"
sequenced: 10950
```

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1585,6 +1587,52 @@ export class ResultsViewPageStore
...
+            var result = '';
+            const alterationData = this.alterationFrequencyDataForQueryGenes;
+
+            if (this.studyIds.result!.length > 0 && this.genes.result!.length > 0 && this.oqlFilteredCaseAggregatedDataByUnflattenedOQLLine.result!.length > 0) {
+                const postData = {
+                    // this might be necessary at some point
+                    // studyid: this.studyIds.result![0],
+                    geneList: this.genes.result!.map(g => g.hugoGeneSymbol),
+                    sourceList:["enrichment"],
+                    alterationData,
```
inodb, line 1604:
```
altered: 1916
gene: "PIK3CA"
percentAltered: "17%"
sequenced: 10950
```

src/config/serverConfigDefaults.ts
```diff
@@ -47,6 +47,7 @@ const ServerConfigDefaults: Partial<IServerConfig> = {
     show_mutation_mapper_tool_grch38: true,
     show_transcript_dropdown: false,
     show_signal: true,
+    show_ndex: true,
```
alisman, line 50: @inodb should this really default to true?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -1585,6 +1587,67 @@ export class ResultsViewPageStore
...
+                        )
+                        .send(postData)
+                        .set('Accept', 'application/json')
+                        .timeout(30000)) as any;
+                } catch (err) {
+                    // Just eat the exception. Result will be empty string.
+                }
+
+                if (
+                    urlResponse &&
```
alisman, line 1638: FYI, you can now do ```urlResponse?.body?.webUrl?.startWith('...')``` and it will evaluate false if anything is undefined

alisman (approved):

src/config/serverConfigDefaults.ts
inodb: thanks for checking - yeah i think it makes sense to have this be the default for all installations

PR #4070, opened 2021-12-02
Show namespace columns in MAF file in Mutation Table component

pvannierop:
# Description

This PR will show custom columns in the MAF file imported with the namespace import mechanism in Mutation Table components of Results View, Patient View and Standalone Mutation Mapper.

# Comment
- Passing of e2e-localdb tests depends on acceptance of PR https://github.com/cBioPortal/cbioportal/pull/9086.

# Video
Peek 2021-12-02 13-35

### DIFF
--- a/src/shared/components/mutationTable/MutationTable.tsx
+++ b/src/shared/components/mutationTable/MutationTable.tsx
@@ -113,6 +113,7 @@ export interface IMutationTableProps {
     mrnaExprRankMolecularProfileId?: string;
     discreteCNAMolecularProfileId?: string;
     columns?: ExtendedMutationTableColumnType[];
+    namespaceColumns?: NamespaceColumnConfig;
     data?: Mutation[][];
     dataStore?: ILazyMobXTableApplicationDataStore<Mutation[]>;
     downloadDataFetcher?: ILazyMobXTableApplicationLazyDownloadDataFetcher;
@@ -198,11 +199,21 @@ export enum MutationTableColumnType {
     GENE_PANEL = 'Gene panel',
     SIGNAL = 'SIGNAL',
 }
-type ExtendedMutationTableColumnType = MutationTableColumnType | string;
+export type ExtendedMutationTableColumnType = MutationTableColumnType | string;
 
-type MutationTableColumn = Column<Mutation[]> & {
+export type MutationTableColumn = Column<Mutation[]> & {
     order?: number;
     shouldExclude?: () => boolean;
+};
+
+// Namespace columns are custom columns that can be added to the MAF file.
+// They are imported via the namespace configuration during data import.
+// See: https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#adding-mutation-annotation-columns-through-namespaces
+// The MutationTable component will render these columns dynamically.
+export type NamespaceColumnConfig = {
+    [namespaceName: string]: {
+        [namespaceColumnName: string]: 'string' | 'number';
+    };
 };
 
 export class MutationTableComponent extends LazyMobXTable<Mutation[]> {}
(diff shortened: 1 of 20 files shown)


src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -111,6 +111,7 @@ export interface IMutationTableProps {
     mrnaExprRankMolecularProfileId?: string;
     discreteCNAMolecularProfileId?: string;
     columns?: ExtendedMutationTableColumnType[];
+    namespaceColumns?: NamespaceColumnConfig;
```
adamabeshouse, line 114: why does it make sense for this to be a separate prop, and not just passed in like other columns?

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -186,13 +187,19 @@ export enum MutationTableColumnType {
...
+export type ExtendedMutationTableColumnType = MutationTableColumnType | string;
 
-type MutationTableColumn = Column<Mutation[]> & {
+export type MutationTableColumn = Column<Mutation[]> & {
     order?: number;
     shouldExclude?: () => boolean;
 };
 
+export type NamespaceColumnConfig = {
+    [namespaceName: string]: {
```
adamabeshouse, line 198: I'm not sure if this is the right place for it, but a comment explaining what a namespace column is would be good

src/shared/components/mutationTable/column/GenericCategoricalColumnFormatter.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
+import { Mutation } from 'cbioportal-ts-api-client';
+import _ from 'lodash';
+
+/**
+ * @author Pim van Nierop
+ */
+export default class GenericCategoricalColumnFormatter {
```
adamabeshouse, line 8: I'm not sure if this name is apt since it seems to use the namespace terminology

src/shared/components/mutationTable/column/GenericNumericColumnFormatter.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+import { Mutation } from 'cbioportal-ts-api-client';
+import _ from 'lodash';
+import generalStyles from 'shared/components/mutationTable/column/styles.module.scss';
+
+/**
+ * @author Pim van Nierop
+ */
+export default class GenericNumericColumnFormatter {
```
adamabeshouse, line 9: same thing - generic name but the implementation concerns namespaces

adamabeshouse:
@pvannierop I think it would help me a lot to understand this more - could you please explain what namespaces are?

src/shared/components/mutationTable/MutationTable.tsx
pvannierop: The interplay between all MutationMapperStore and MutationMapperTable classes for column definition is rather complicated when looking at the code. I have looked at possibilities to pass the namespace columns as names in the `columns` prop to MutationTable, but this does not work. The fundamental problem is that in order to dynamically create the custom columns the information on type (`string` or `number`) need to be passed along with the column name. This information is used f.i in ResultsViewMutationTable (see comments named `// generate namespace columns`). The new type NamespaceColumnConfig contains this type information.

src/shared/components/mutationTable/MutationTable.tsx
pvannierop: I added come comment lines to explain this feature to frontend-devs.

src/shared/components/mutationTable/column/GenericCategoricalColumnFormatter.tsx
pvannierop: Ah you are right. Names update to `NumericNamspaceColumnFormatter` and `CategoricalNamspaceColumnFormatter`.

src/shared/components/mutationTable/column/GenericNumericColumnFormatter.tsx
pvannierop: Updated. See comment above.

pvannierop:
@adamabeshouse I added this comment to the code. Does this provide enough information:

```
// Namespace columns are custom columns that can be added to the MAF file.
// They are imported via the namespace configuration during data import.
// See: https://docs.cbioportal.org/5.1-data-loading/data-loading/file-formats#adding-mutation-annotation-columns-through-namespaces
// The MutationTable component will render these columns dynamically.
```

src/shared/components/mutationTable/MutationTable.tsx
adamabeshouse: great, thank you.

src/shared/components/mutationTable/MutationTable.tsx
adamabeshouse: Understood, thank you.

adamabeshouse (approved):
LGTM 🎈

pvannierop:
relaced by #4159

PR #4071, opened 2021-12-02
Fix screenshot after cna data introduced

alisman:


alisman:
@inodb you think it makes sense that cna data was introduced all of a sudden?

PR #4076, opened 2021-12-06
Timeline: allow user to mouse into the tooltip

alisman:
This PR allows data managers to add links to timeline event tooltips by adding attributes to event data.   They can do so by using markdown in the text value of the attribute.

This event

```{ key:"SLIDES", value: '[Slide Viewer](http://www.example.com)' }```

produces the following output:

image

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
@@ -30,6 +30,7 @@ import {
     renderShape,
     SHAPE_ATTRIBUTE_KEY,
 } from './renderHelpers';
+import ReactMarkdown from 'react-markdown';
 
 export interface ITimelineTrackProps {
     trackData: TimelineTrackSpecification;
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx [hunk 1]
```diff
@@ -350,20 +357,23 @@ const TimelineItemWithTooltip: React.FunctionComponent<{
                     });
 
                     setTooltipUid(uid);
+
+                    store.setHoveredTooltipUid(uid);
```
alisman, line 361: this means tooltip will only show up on initial entry and won't adjust as user continues to move mouse.  this is necessary in order to allow user to mouse INTO a tooltip

packages/cbioportal-clinical-timeline/src/TimelineTrack.tsx
```diff
@@ -350,20 +357,23 @@ const TimelineItemWithTooltip: React.FunctionComponent<{
...
-                });
             }}
             onMouseLeave={e => {
-                const uid = syncTooltipUid();
-
-                if (uid && !store.isTooltipPinned(uid)) {
-                    store.removeTooltip(uid);
-                    setTooltipUid(null);
-                }
+                setTimeout(() => {
```
alisman, line 369: by using a timeout here, we can PIN the tooltip on entry into the tooltip. that way, when this executes, if the user is in the tooltip itself, the if statement at 371 will evaluate false (and not remove the tooltip)

adamabeshouse: would be good to put this into a comment

adamabeshouse (approved):

PR #4077, opened 2021-12-07
Deactivate checkbox for unauthorized studies (when skin.home_page.sho…

dianab0:
The unauthorized studies can be displayed in the study list when  when skin.home_page.show_unauthorized_studies is set to true in portal.properties.

This PR is deactivating the checkbox of these studies.
Screenshot from 2021-12-07 11-00-48

### DIFF
--- a/src/shared/components/query/StudyListLogic.ts
+++ b/src/shared/components/query/StudyListLogic.ts
@@ -16,7 +16,7 @@ import {
     SearchResult,
 } from '../../lib/textQueryUtils';
 import { cached } from 'mobxpromise';
-import { ServerConfigHelpers } from '../../../config/config';
+import { getServerConfig, ServerConfigHelpers } from '../../../config/config';
 import memoize from 'memoize-weak-decorator';
 
 export const PAN_CAN_SIGNATURE = 'pan_can_atlas';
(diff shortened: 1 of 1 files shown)


src/shared/components/query/StudyListLogic.ts
```diff
@@ -299,6 +299,13 @@ export class FilteredCancerTreeView {
             if (this.store.isDeletedVirtualStudy(study.studyId)) {
                 return true;
             }
+            if (
+                getServerConfig().skin_home_page_show_unauthorized_studies &&
```
alisman, line 303: i think if readPermission is false we can ALWAYS deactivate?  i.e. no need to check the configuration flag?   also how come we check for studyId (just curious)?

dianab0: Correct. I modified it to only check if the readPermission is false.
Regarding studyId, I copied the entire condition code from the StudyList. I don't remember the reason I put it there and why I was getting error otherwise, but I left it here as well just as double checking. I removed it and tested it.

src/shared/components/query/StudyListLogic.ts
```diff
@@ -16,7 +16,7 @@ import {
     SearchResult,
 } from '../../lib/textQueryUtils';
 import { cached } from 'mobxpromise';
-import { ServerConfigHelpers } from '../../../config/config';
+import { getServerConfig, ServerConfigHelpers } from '../../../config/config';
```
alisman, line 19: maybe not used anymore?

alisman (approved):

src/shared/components/query/StudyListLogic.ts
dianab0: done. thank you

alisman (approved):

PR #4084, opened 2021-12-09
Refactor reference genome id evaluation

pvannierop:
This PR moves logic for translation of reference genome identifier to the `referenceGenomeUtils` class, as well as adds a comment that explains this strategy in the `tracksHelper.ts` class.

### DIFF
--- a/src/pages/patientView/genomicOverview/tracksHelper.ts
+++ b/src/pages/patientView/genomicOverview/tracksHelper.ts
@@ -2,9 +2,9 @@ import Raphael from 'webpack-raphael';
 import $ from 'jquery';
 import _ from 'lodash';
 import { Mutation } from 'cbioportal-ts-api-client';
-import { DEFAULT_GENOME_BUILD } from 'pages/patientView/genomicOverview/Tracks';
 import { default as chromosomeSizes } from './chromosomeSizes.json';
 import { IIconData } from './GenomicOverviewUtils.js';
+import { GENOME_ID_TO_GENOME_BUILD } from 'shared/lib/referenceGenomeUtils';
 
 export function GenomicOverviewConfig(
     nRows: any,
@@ -77,17 +77,6 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
     return ends;
 }
 
-/**
- * storing chromesome length info
- */
-export const genomeBuilds: Map<string, string> = new Map([
-    ['hg19', 'GRCh37'],
-    ['37', 'GRCh37'],
-    ['hg38', 'GRCh38'],
-    ['38', 'GRCh38'],
-    ['mm10', 'GRCm38'],
-]);
-
 export type ChromosomeSizes = {
     genomeBuild: string;
     chromosomeSize: number[];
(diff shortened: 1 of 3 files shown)


src/shared/lib/referenceGenomeUtils.tsx
```diff
@@ -18,6 +18,17 @@ export const REFERENCE_GENOME = {
     },
 };
 
+export const GENOME_ID_TO_GENOME_BUILD = {
+    'GRCh37': 'GRCh37',
+    '37': 'GRCh37',
+    'hg19': 'GRCh37',
+    'GRCh38': 'GRCh38',
+    '38': 'hg38',
```
SRodenburg, line 26: There is a mistake here

```suggestion
    '38': 'GRCh38',
```

src/pages/patientView/genomicOverview/tracksHelper.ts
```diff
@@ -77,39 +77,25 @@ function getChmEndsPerc(chms: Array<any>, total: any) {
...
-        return map;
-    },
-    {}
-);
+} = _(chromosomeSizes)
+        .keyBy(entry => entry.genomeBuild)
+        .mapValues(entry => entry.chromosomeSize)
+        .value();
 
 export function getChmInfo(genomeBuild: string) {
```
SRodenburg, line 92: I would be in favor of refactoring this function name to `getGenomeCoordinates`  or something

SRodenburg (changes requested):

src/shared/lib/referenceGenomeUtils.tsx
pvannierop: Fixed. Thanks for that.

src/pages/patientView/genomicOverview/tracksHelper.ts
pvannierop: It is not a function, but a data structure (a map). The data that the keys are pointing to are number that each represent a chromosome size. So, I think the name is appropriate. Why would you like a name change?

SRodenburg: I was indicating `getChmInfo` ;)

pvannierop: Ah ok, this function build a whole collection of functions that translate chromosomal location to positions on a 2d canvas for drawing the tracks. I agree that the name `getChmInfo` can be improved. I also think that `getGenomeCoordinates` does not cover the content correctly. What do you think of `getChromosomeCoordinateTranslation`?

SRodenburg: Why does it not cover the content? What the methods does it calculate the relative coordinates of the points, based on the chromosome length and the start/end position. What about `getRelativeCoordinates`? I am sure you can work from here. Please get this merged.

SRodenburg (approved):

SRodenburg:
@pvannierop can you check the failing tests? We need this merged.

src/pages/patientView/genomicOverview/tracksHelper.ts
pvannierop: Update to getRelativeCoordinates

alisman (approved):

PR #4088, opened 2021-12-10
Handle case where study has more than one profile for a given alterat…

alisman:
When we populate query store from URL, we are validating selected profiles against a list of valid profiles in selected studies.  We were not handling the case where a study could have multiple profiles for a given alteration type and so were not properly restoring state from url


alisman:
message to self: there's still a problem where you cannot turn OFF mrna expression.  it just defaults to on if unselected

adamabeshouse (approved):

PR #4091, opened 2021-12-14
Fix import of structural variants from query into oncoprinter

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/9068

image

it's not perfect, as I still need to implement oncokb annotation of structural variants in oncoprinter. But that is a bit more work because it requires additional user input of the site1 and site2 genes.


dippindots (changes requested):
@adamabeshouse The unit tests are failing, it looks like there are some tests need to be updated, besides that, this looks good to me.

src/pages/staticPages/tools/oncoprinter/OncoprinterImportUtils.ts
```diff
@@ -142,7 +143,8 @@ function getOncoprinterGeneticInputLine(parsed: OncoprinterGeneticInputLine) {
                 line.push('PROT');
                 break;
             case AlterationTypeConstants.STRUCTURAL_VARIANT:
-                line.push('STRUCTURAL VARIANT');
+                line.push(parsed.eventInfo || 'structural_variant');
```
alisman, line 146: maybe comment why it would be missing eventInfo?

alisman (approved):

src/pages/staticPages/tools/oncoprinter/OncoprinterImportUtils.ts
adamabeshouse: I just wanted to do this as a catch in case there's missing data. would eventInfo always exist for structural variants?

PR #4094, opened 2021-12-17
Fixed the rendering of html inside of the unauthorized message

dianab0:
Minor fix for listing unauthorized studies feature.

1. Adding a link to the global message (like skin.home_page.unauthorized_studies_global_message=`<a href="mailto:{$.Owner.email};subject='Access Request'">Request Access</a>`) would not be rendered by windows browser due to additional `` added in the code.

2. In the browser console there is an error about DefaultTooltip not having a key property. I commented this out
Screenshot from 2021-12-17 15-44-33
.

### DIFF
--- a/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
+++ b/src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
@@ -153,8 +153,7 @@ export default class StudyTagsTooltip extends React.Component<
 > {
     renderTooltip() {
         return (
-            <DefaultTooltip
-                //key={this.props.key}
+            <DefaultTooltip 
                 mouseEnterDelay={this.props.mouseEnterDelay}
                 placement={this.props.placement}
                 overlay={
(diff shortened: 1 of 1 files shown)


src/shared/components/studyTagsTooltip/StudyTagsTooltip.tsx
```diff
@@ -154,7 +154,7 @@ export default class StudyTagsTooltip extends React.Component<
     renderTooltip() {
         return (
             <DefaultTooltip
-                key={this.props.key}
+                //key={this.props.key}
```
alisman, line 157: what's this about?

dianab0: Hi @alisman. I saw an error in the browser console saying that key is not a prop. I checked and indeed, key is not a property of DefaultTooltip. Not sure why it is there so I only commented it. The error no longer appears and the tooltips work fine. You can see a screenshot in the PR description.

PR #4101, opened 2021-12-22
appBootstrapper and routes .jsx->.tsx

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/9055

### DIFF
--- a/src/appBootstrapper.tsx
+++ b/src/appBootstrapper.tsx
@@ -13,7 +13,6 @@ import {
     initializeAppStore,
     initializeLoadConfiguration,
     initializeServerConfiguration,
-    setConfigDefaults,
     setServerConfig,
 } from './config/config';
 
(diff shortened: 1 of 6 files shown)


src/appBootstrapper.tsx
```diff
@@ -74,7 +91,7 @@ if (browser) {
 if (getBrowserWindow().navigator.webdriver) {
     $(document).ready(() => {
         $('body').addClass('e2etest');
-        window.e2etest = true;
+        ((window as any) as ICBioWindow).e2etest = true;
```
alisman, line 94: I think we should switch to always use getBrowserWindow() and then make that return ICbioWindow which should inherit from browser window.  Of course, that may cause a host of other ts errors where we use getBrowserWindow() in the app.  But it would be an improvement.

adamabeshouse: okay thats a good idea

adamabeshouse: so getBrowserWindow is in commons, and a lot of things in ICBioWindow are not (e.g. ExtendedRouterStore, AppStore, IServerConfig). So refactoring this would mean bringing a lot of things into commons that aren't already. What do you think?

alisman: i see.  yeah, that doesn't sound like a good idea.  maybe then just in this file make a variable browserWindow with this type and just refer to that?    i just think we should take the opportunity to make it cleaner since that's what this PR is all about

adamabeshouse: sure, that works

src/routes.tsx
```diff
@@ -443,7 +435,7 @@ export const makeRoutes = routing => {
                     component={ScrollToTop(
                         LocationValidationWrapper(
                             GroupComparisonPage,
-                            comparisonTabParamValidator
+                            comparisonTabParamValidator()
```
alisman, line 438: was this just not working before?

adamabeshouse: I guess so?

PR #4104, opened 2021-12-28
Add back github friendly markdown support to static markdown component

alisman:
Fixes https://github.com/cbioportal/cbioportal/issues/9158
Support for GFM was lost when we upgraded react-markdown.  Latest version requires plugin


inodb (approved):
Works like a charm - thanks!

PR #4106, opened 2021-12-30
Conditionally fetch mutation assessor download data

onursumer:
Related to cBioPortal/cbioportal#9104

We still need to update the `show.genomenexus.annotation_sources` property in the backend to make this change effective.

## Checks

### DIFF
--- a/src/shared/lib/MutationTableDownloadDataFetcher.ts
+++ b/src/shared/lib/MutationTableDownloadDataFetcher.ts
@@ -27,6 +27,7 @@ import {
     default as GenomeNexusMutationAssessorCache,
     defaultGNFetch as fetchGenomeNexusMutationAssessorData,
 } from 'shared/cache/GenomeNexusMutationAssessorCache';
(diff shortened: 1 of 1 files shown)


leexgh (approved):
Thanks! Not sure should we also disable it on mutation table to be consistent?

leexgh:
I've also disabled mutation assessor on genome nexus too, we should find a good way to handle this unstableness

onursumer:
> Thanks! Not sure should we also disable it on mutation table to be consistent?

@leexgh Seems like we already conditionally show/hide it in the mutation table using the same config prop: https://github.com/cBioPortal/cbioportal-frontend/blob/6211c300dc0210ee347532b46288aef675ee773a/src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx#L63. It should be disabled when we set the corresponding prop, right?

leexgh:
@onursumer Oh I forgot we already have this, then it's fine I guess. Do we want to have another pr to disable mutation assessor?

PR #4107, opened 2022-01-03
Fix handling of Spearman correlation (et al) with NaN value from server

alisman:
Related to https://github.com/cBioPortal/cbioportal/issues/9148

image

Issue comes from this service when you submit only one bin. 

https://www.cbioportal.org/api/clinical-data-density-plot/fetch?xAxisAttributeId=FRACTION_GENOME_ALTERED&xAxisBinCount=44&xAxisEnd=1&xAxisLogScale=false&xAxisStart=0&yAxisAttributeId=MUTATION_COUNT&yAxisBinCount=38&yAxisLogScale=false&yAxisStart=0

Service returns NaN for the correlation values which javascript then barfs on.   This PR will just print NaN instead of white screen so it's improvement.  

There seems to be a problem with selection mechanism wherein if you choose less than four bars, it considers it just a single bar.  Questions are:
A) should service return NaN?
B) should selection UI allow selection of a single bar?  When you drag from within a bar, how should that affect selection?

### DIFF
--- a/src/pages/studyView/charts/scatterPlot/StudyViewDensityScatterPlot.tsx
+++ b/src/pages/studyView/charts/scatterPlot/StudyViewDensityScatterPlot.tsx
@@ -493,6 +493,14 @@ export default class StudyViewDensityScatterPlot
                 </text>,
             ];
             let correlationLabels;
+
+            const pearson = _.isNumber(this.props.pearsonCorr)
+                ? this.props.pearsonCorr.toFixed(4)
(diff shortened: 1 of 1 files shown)


inodb (approved):
Thanks!

inodb:
@alisman My thoughts on the questions:

A) I think it makes sense to return NA when no correlation can be computed with given input
B) Selection of single bar makes sense to me. When starting from within a bar, it should select the whole bar

PR #4109, opened 2022-01-03
Add e2e test for freqency calculation in study view on study with cal…

alisman:
Adds test for https://github.com/cBioPortal/cbioportal/issues/9163

A recent backend regression caused a problem with calculation of mutation frequency in study view for studies with called but unprofiled mutations.  This e2e test catches this condition on public impact study.


alisman:
@inodb you think this is sufficient?

inodb:
@alisman Thanks! For this study it only showed up in the SV table, see: https://master.cbioportal.org/study/summary?id=egc_mskcc_2020. If you can change it to take a screenshot from that table then yes it would be perfect 👍

inodb:
@alisman ah actually never mind I was looking at the wrong study. You're right `msk_impact_2017` has it in the mutations table as well. Great - thanks!

inodb (approved):
Nice one! Thanks for the quick turnaround!

PR #4110, opened 2022-01-04
rewrite e2e xml result to eliminate duplicates reports due to retries

alisman:
Duplicate test reports due to retries were complicating result interpretation in CircleCI.  This eliminates duplicates, using only the final entry, which will either be a failure or pass.

### DIFF
--- a/end-to-end-test/shared/edit-junit.js
+++ b/end-to-end-test/shared/edit-junit.js
@@ -3,18 +3,6 @@ var fs = require('fs'),
 var _ = require('lodash');
 
 function transformJUNITFiles(dir) {
-    // fs.readdir(dir, function(err, files) {
-    //     if (err) {
-    //         console.log(err);
-    //         return;
-    //     }
-    //     files
-    //         .filter(s => /results.*.xml$/.test(s))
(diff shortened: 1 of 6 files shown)


end-to-end-test/shared/edit-junit.js
```diff
@@ -0,0 +1,82 @@
+var fs = require('fs'),
+    xml2js = require('xml2js');
+var _ = require('lodash');
+
+function transformJUNITFiles(dir) {
+    console.log('AARON', dir);
```
inodb, line 6: log

inodb (approved):
LGTM -

As a side note: it seems kinda like this is a bug in the reporter thing of wdio itself? Or am I misunderstanding?

alisman:
@inodb i don't think it's a bug.   the wdio reporter just adds a node whenever a test finishes and probably shouldn't worry about anything else.  the duplicates actually represent valuable information (that the test was retried).  the problem is more that CircleCI interface is too dumb to interpret the duplicates as retries.  but i don't really blame them for that either.  they are just printing xunit and i don't think xunit has any representation of "retry".   shrug emoji

PR #4117, opened 2022-01-11
Remove root node fetch in placeArrowBottomLeft

zhx828:
Cannot locate the `this` reference which causes page goes blank. The targetEl is not used anywhere.

Not sure which recent change causes this issue. Maybe a better solution exists with more knowledge from @onursumer @alisman 

Jan-11-2022 13-41-37


zhx828:
I only put it as draft PR waiting the confirmation on the fix. Happy to convert to regular once getting the team's approval.

onursumer:
It's not clear to me why we have an unused `targetEl`. Maybe we were using it previously but not anymore? If removing it fixes the issue I would say let's do it.

onursumer (approved):
It should be okay to remove an unused variable

inodb:
Duplicate of https://github.com/cBioPortal/cbioportal-frontend/pull/4118

PR #4119, opened 2022-01-11
study view - Always show available generic assay add chart tabs regardless of filtering

adamabeshouse:
Fixes https://github.com/cBioPortal/cbioportal/issues/8969

The study view would crash when, in a study view with multiple studies, some with generic assay data, all the samples with generic assay data were filtered out. This was related to code in the UI for adding generic assay charts, which would hide/show depending on whether the currently filtered samples had generic assay data. This PR does away with the problem by no longer hiding the generic assay add chart UI regardless of the filtered sample state.

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -5019,7 +5019,6 @@ export class StudyViewPageStore
                                     ),
                                 };
                             })
-                            .filter(record => record.count > 0)
                             .value();
                     })
                     .value()
--- a/src/pages/studyView/addChartButton/AddChartButton.tsx
+++ b/src/pages/studyView/addChartButton/AddChartButton.tsx
@@ -518,11 +518,12 @@ class AddChartTabs extends React.Component<IAddChartTabsProps, {}> {
                 // And one tab can only has one selected profile at a time
                 // selectedGenericAssayProfileIdByType been initialzed at the begining
                 // so we know we can always find a selected profile for each Generic Assay type
-                const molecularProfileIdsInType = options.find(
-                    option =>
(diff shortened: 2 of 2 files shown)


onursumer (approved):
LGTM!

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5019,7 +5019,6 @@ export class StudyViewPageStore
                                     ),
                                 };
                             })
-                            .filter(record => record.count > 0)
```
alisman, line 5022: will we now show options even when there are no matching samples?

alisman:
@adamabeshouse looks good.   could you put an explaination of fix in the PR description?   i.e. what the problem was and how this fixes it

src/pages/studyView/StudyViewPageStore.ts
alisman: @dippindots this filtering must have a purpose (don't show options what have no samples?)

adamabeshouse: @alisman the way it looks now is the options actually say "0 samples" in the UI. so this just makes the UI still appear even if there are no profiles with profiled, selected samples. the main word there is selected - so there could be profiled samples in the study, but not in the current study view filter. I think this is the most consistent way to handle it since we're not hiding any other add chart UI based on the study view filter state

dippindots: @alisman Agreed with Adam, based on the discussion yesterday during the meeting, I think we decided to show all profiles regardless of it has samples or not, so even if a profile has `0 samples`, we will still show that option.

PR #4120, opened 2022-01-12
Putative Passenger tooltip icon

dianab0:
There is a flash icon for the Putative Driver in the Oncoprint tooltip, but there is no icon for the Putative Passenger. I have added a gray circle icon for it as in the gif below.

Putative-Passenger


alisman:
@dianab0 you could probably accomplish this with svg in about 1% of the file size.  here's a start:
```
<svg xmlns="http://www.w3.org/2000/svg" version="1.1">
   <circle cx="100" cy="50" r="40" stroke="black" stroke-width="2" fill="none" />
</svg> 
```

dianab0:
Rebased to the new master and changed the img icon to svg circle as below:

Screenshot from 2022-01-17 15-06-42

src/shared/components/oncoprint/TooltipUtils.ts
```diff
@@ -585,6 +585,15 @@ export function makeGeneticTrackTooltip(
                             `<img src="${customDriverImg}" title="${driver_filter}: ${driver_filter_annotation}" alt="driver filter" style="height:11px; width:11px;margin-left:3px"/>`
                         );
                     }
+                    //If we have data for the binary custom driver annotations, append an icon to the tooltip with the annotation information
+                    if (driver_filter && driver_filter === PUTATIVE_PASSENGER) {
```
pvannierop, line 589: Small thing. This should be an `else if` since driver (previous condition) and passenger are mutually exclusive.

pvannierop (changes requested):
One small correction suggested.

src/shared/components/oncoprint/TooltipUtils.ts
dianab0: Done

PR #4123, opened 2022-01-12
Fix detection of rna_seq molecular profile

alisman:
We are submitting an rppa profile to an incompatible service.   We should not fallback to an rppa profile for the default mrna expression column of the patient view

Fixes https://github.com/cbioportal/cbioportal/issues/9140

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -1310,7 +1310,11 @@ export class PatientViewPageStore {
         {
             await: () => [this.molecularProfilesInStudy],
             invoke: async () =>
-                findMrnaRankMolecularProfileId(this.molecularProfilesInStudy),
+                findMrnaRankMolecularProfileId(
+                    this.molecularProfilesInStudy.result!.map(
+                        p => p.molecularProfileId
+                    )
+                ),
         },
         null
     );
--- a/src/shared/lib/StoreUtils.ts
+++ b/src/shared/lib/StoreUtils.ts
@@ -1109,33 +1109,16 @@ export function findUncalledMutationMolecularProfileId(
 }
 
 export function findMrnaRankMolecularProfileId(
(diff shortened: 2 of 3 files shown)


src/shared/lib/StoreUtils.ts
```diff
@@ -1109,33 +1109,16 @@ export function findUncalledMutationMolecularProfileId(
...
-    }
-
-    const regex1 = /^.+rna_seq.*_zscores$/; // We prefer profiles that look like this
-    const regex2 = /^.*_zscores$/; // If none of the above are available, we'll look for ones like this
-    const preferredProfile:
-        | MolecularProfile
-        | undefined = molecularProfilesInStudy.result.find(
-        (gp: MolecularProfile) =>
-            regex1.test(gp.molecularProfileId.toLowerCase())
+    const regex1 = /^.+rna_seq.*_zscores$/i; // We prefer profiles that look like this
```
alisman, line 1114: @inodb this will still allow non mrna profiles when the column header implies mrna.  However, if  i change it to mrna_seq (is that valid?) then we no longer match most of the profiles currently used.  So we'd have to change the matching to make it consistent.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -1310,7 +1310,11 @@ export class PatientViewPageStore {
         {
             await: () => [this.molecularProfilesInStudy],
             invoke: async () =>
-                findMrnaRankMolecularProfileId(this.molecularProfilesInStudy),
+                findMrnaRankMolecularProfileId(
```
alisman, line 1313: i just changed this function to accept a list of ids instead of whole promise, which was silly and complicated testing

src/shared/lib/StoreUtils.ts
inodb: This looks good to me

inodb (approved):
Thanks for fixing! LGTM!

PR #4102, opened 2021-12-28
Custom bins feature (DRAFT)

dianab0:
Extended custom bins menu according to rfc ... (TODO)
    
Added Quartiles, Median split and Generate bins options, where Generate bins allows user to define bin size and min value.

PR #4127, opened 2022-01-17
Custom bins feature

dianab0:
Extended custom bins menu according to rfc https://docs.google.com/document/d/10za8fgTsXInjUCJltWzAW-HHRJnpLa3IAT17W8M5a9E/edit

Added Quartiles, Median split and Generate bins options, where Generate bins allows user to define bin size and min value.

Backend PR: https://github.com/cBioPortal/cbioportal/pull/9198


dianab0:
Same as https://github.com/cBioPortal/cbioportal-frontend/pull/4102 but from my own fork.

PR #4130, opened 2022-01-19
Update signal component to support both somatic and germline

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9222

### DIFF
--- a/packages/react-mutation-mapper/src/component/column/Signal.tsx
+++ b/packages/react-mutation-mapper/src/component/column/Signal.tsx
@@ -25,6 +25,7 @@ type SignalProps = {
     indexedVariantAnnotations?: RemoteData<
         { [genomicLocation: string]: VariantAnnotation } | undefined
     >;
+    mutationType?: Pathogenicity;
 };
 
 type SignalValueProps = SignalProps & {
@@ -36,7 +37,7 @@ export function getSignalData(
     indexedVariantAnnotations?: RemoteData<
         { [genomicLocation: string]: VariantAnnotation } | undefined
     >,
-    mutationType?: typeof Pathogenicity[keyof typeof Pathogenicity] // mutationType could be "germline", "somatic", or "undefined" which means both
+    mutationType?: Pathogenicity // mutationType could be "germline", "somatic", or "undefined" which means both
 ): IExtendedSignalMutation[] {
     let signalData: IExtendedSignalMutation[] = [];
     const variantAnnotation = indexedVariantAnnotations
@@ -48,8 +49,7 @@ export function getSignalData(
         variantAnnotation.signalAnnotation.annotation &&
         variantAnnotation.signalAnnotation.annotation.length > 0
     ) {
-        console.log(variantAnnotation.signalAnnotation.annotation);
-        // if mutation is somatic OR germline, get annotation depands on mutationType
+        // if mutation is somatic OR germline, get annotation depending on mutationType
         if (variantAnnotation.signalAnnotation.annotation.length === 1) {
             if (
                 (mutationType === Pathogenicity.GERMLINE &&
@@ -66,7 +66,7 @@ export function getSignalData(
                 ]);
             }
         }
-        // if mutation is both somatic AND germline, get annotation for both or depands on mutationType
+        // if mutation is both somatic AND germline, get annotation for both or get one of them depending on mutationType
         else {
             // if mutationType is undefined, get annotation for both somatic and germline
             if (mutationType === undefined) {
@@ -74,21 +74,24 @@ export function getSignalData(
                     variantAnnotation.signalAnnotation.annotation
                 );
             }
-            variantAnnotation.signalAnnotation.annotation.forEach(
-                annotation => {
-                    if (
-                        mutationType === Pathogenicity.GERMLINE &&
-                        annotation.mutationStatus.includes('germline')
-                    ) {
-                        signalData = extendMutations([annotation]);
-                    } else if (
-                        mutationType === Pathogenicity.SOMATIC &&
-                        annotation.mutationStatus.includes('somatic')
-                    ) {
-                        signalData = extendMutations([annotation]);
+            // if mutationType is defined, get annotation depending on mutationType
+            else {
+                variantAnnotation.signalAnnotation.annotation.forEach(
+                    annotation => {
+                        if (
+                            mutationType === Pathogenicity.GERMLINE &&
+                            annotation.mutationStatus.includes('germline')
+                        ) {
+                            signalData = extendMutations([annotation]);
+                        } else if (
+                            mutationType === Pathogenicity.SOMATIC &&
+                            annotation.mutationStatus.includes('somatic')
+                        ) {
+                            signalData = extendMutations([annotation]);
+                        }
                     }
-                }
-            );
+                );
+            }
         }
     }
     return signalData;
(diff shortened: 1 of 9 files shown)


packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,48 @@ export function getSignalData(
         variantAnnotation.signalAnnotation.annotation &&
         variantAnnotation.signalAnnotation.annotation.length > 0
     ) {
-        variantAnnotation.signalAnnotation.annotation.forEach(annotation => {
-            // only have one germline annotation
-            if (annotation.mutationStatus.includes('germline')) {
-                signalData = extendMutations([annotation])[0];
+        // if mutation is somatic OR germline, get annotation depands on mutationType
```
onursumer, line 52:
```suggestion
        // if mutation is somatic OR germline, get annotation depending on mutationType
```

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,48 @@ export function getSignalData(
...
+                    variantAnnotation.signalAnnotation.annotation
+                );
+            } else {
+                variantAnnotation.signalAnnotation.annotation.forEach(
+                    annotation => {
+                        if (
+                            mutationType === Pathogenicity.GERMLINE &&
+                            annotation.mutationStatus.includes('germline')
+                        ) {
+                            signalData = extendMutations([annotation]);
```
onursumer, line 83: Do we really want to overwrite the existing `signalData` in a for loop? Or should we do `signalData.push(extendMutations([annotation]))` instead?

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,48 @@ export function getSignalData(
...
+                        } else if (
+                            mutationType === Pathogenicity.SOMATIC &&
+                            annotation.mutationStatus.includes('somatic')
+                        ) {
+                            signalData = extendMutations([annotation]);
```
onursumer, line 88: same here. should we push instead of overwrite?

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -64,29 +103,45 @@ export function signalSortMethod(
 export function getSortValue(
     signalData: IExtendedSignalMutation
 ): number | null {
-    return signalData.germlineFrequency || null;
+    if (signalData) {
+        return signalData.germlineFrequency || null;
+    }
+    return null;
```
onursumer, line 109: Not a big deal but maybe do this instead:

```suggestion
    return signalData ? signalData.germlineFrequency || null: null;
```

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,48 @@ export function getSignalData(
...
+            ) {
+                signalData = extendMutations([
+                    variantAnnotation.signalAnnotation.annotation[0],
+                ]);
             }
-        });
+        }
+        // if mutation is both somatic AND germline, get annotation for both or depands on mutationType
+        else {
+            // if mutationType is undefined, get annotation for both somatic and germline
```
onursumer, line 71: are we getting annotation for both somatic and germline here? or actually in the `else` block?

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,48 @@ export function getSignalData(
...
+                        'somatic'
+                    ))
+            ) {
+                signalData = extendMutations([
+                    variantAnnotation.signalAnnotation.annotation[0],
+                ]);
             }
-        });
+        }
+        // if mutation is both somatic AND germline, get annotation for both or depands on mutationType
```
onursumer, line 69:
```suggestion
        // if mutation is both somatic AND germline, get annotation for both depending on mutationType
```

onursumer (changes requested):
Looks good in general, thanks! Added a few requests and questions regarding the changes in `getSignalData` function.

packages/react-mutation-mapper/src/component/column/Signal.tsx
leexgh: @onursumer Good question, `extendMutations` actrullay could only be called once , probably no need to do `signalData.push(extendMutations([annotation]))`.
The loop iterates every `annotation` from `variantAnnotation.signalAnnotation.annotation`, but there are only 3 different cases that `variantAnnotation.signalAnnotation.annotation` could be:
1. only somatic annotation, variantAnnotation.signalAnnotation.annotation.length() = 1
2. only germline annotation, ariantAnnotation.signalAnnotation.annotation.length() = 1
3. both somatic and germline annotation, ariantAnnotation.signalAnnotation.annotation.length() = 2
We go into this loop only if the data is case 3 and `mutationType` is defined. And `extendMutations` would only be called once, which is when annotation object type matches `mutationType` that we defined

packages/react-mutation-mapper/src/component/column/Signal.tsx
leexgh: @onursumer If no `mutationType` is defined then get both, if `mutationType` is defined then we get one of them based on `mutationType`. 
Actually we don't have a use case for getting both germline and somatic for now, because in signal column of mutations table we only show germline frequencies, but in the future we probably want to change it to show both somatic and germline, or add a new column to show somatic, so I just leave it a possibility to be able to get both

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,50 @@ export function getSignalData(
         variantAnnotation.signalAnnotation.annotation &&
         variantAnnotation.signalAnnotation.annotation.length > 0
     ) {
-        variantAnnotation.signalAnnotation.annotation.forEach(annotation => {
-            // only have one germline annotation
-            if (annotation.mutationStatus.includes('germline')) {
-                signalData = extendMutations([annotation])[0];
+        // if mutation is somatic OR germline, get annotation depanding on mutationType
```
onursumer, line 52: almost there :)

```suggestion
        // if mutation is somatic OR germline, get annotation depending on mutationType
```

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,50 @@ export function getSignalData(
...
+                    variantAnnotation.signalAnnotation.annotation[0].mutationStatus.includes(
+                        'somatic'
+                    ))
+            ) {
+                signalData = extendMutations([
+                    variantAnnotation.signalAnnotation.annotation[0],
+                ]);
+            }
+        }
+        // if mutation is both somatic AND germline, get annotation for both or depanding on mutationType
```
onursumer, line 69:
```suggestion
        // if mutation is both somatic AND germline, get annotation for both or get one of them depending on mutationType
```

packages/react-mutation-mapper/src/component/column/Signal.tsx
```diff
@@ -46,12 +49,50 @@ export function getSignalData(
...
+        else {
+            // if mutationType is undefined, get annotation for both somatic and germline
+            if (mutationType === undefined) {
+                signalData = extendMutations(
+                    variantAnnotation.signalAnnotation.annotation
+                );
+            }
+            // if mutationType is defined, get annotation depanding on mutationType
```
onursumer, line 77:
```suggestion
            // if mutationType is defined, get annotation depending on mutationType
```

packages/react-mutation-mapper/src/component/column/Signal.tsx
leexgh: Oops 😅

PR #4131, opened 2022-01-19
Format exon start and end location

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/8935

Screen Shot 2022-01-25 at 1 28 18 PM

### DIFF
--- a/packages/cbioportal-utils/src/exon/ExonUtils.ts
+++ b/packages/cbioportal-utils/src/exon/ExonUtils.ts
@@ -61,13 +61,67 @@ export function extractExonInformation(
     return exonInfo;
 }
 
-export function formatExonLocation(exonLocation: number) {
+// Generate exon location description by exon location number.
+// Description should follow this format: "Nucleotide xx of amino acid xx".
+// Also need to make it clear which location is inclusive for start position and last end position
+export function formatExonLocation(exonLocation: number, index?: number) {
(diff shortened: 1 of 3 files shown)


packages/cbioportal-utils/src/exon/ExonUtils.ts
```diff
@@ -61,13 +61,63 @@ export function extractExonInformation(
     return exonInfo;
 }
 
-export function formatExonLocation(exonLocation: number) {
+export function formatExonLocation(exonLocation: number, index?: number) {
```
onursumer, line 64: Should we add more comments for this function to explain what we are doing and why we are doing it? Also maybe add some tests to cover each case in this function?

packages/cbioportal-utils/src/exon/ExonUtils.ts
```diff
@@ -61,13 +61,63 @@ export function extractExonInformation(
...
+        } else {
+            return (
+                'Nucleotide 2 of amino acid ' +
+                (Math.round(exonLocation) + 1).toString()
+            );
+        }
+    }
+}
+
+export function formatExonLength(exonLocation: number) {
```
onursumer, line 108: same here. a few more comments and tests would be nice.

onursumer (commented):
Looks good! Added a few suggestions

onursumer (approved):
Thanks!

PR #4138, opened 2022-01-21
Fix issue for unauthorized studies global message

dianab0:
This PR fixes two problems:
- The SELECT ALL button selected unauthorized studies
- The message for unauthorized studies would not be rendered when there were no study tag placeholders.  

Plus, it allows the use of {$.studyId} placeholder in the global message of unauthorized studies.

### DIFF
--- a/src/shared/components/query/QueryContainer.tsx
+++ b/src/shared/components/query/QueryContainer.tsx
@@ -24,6 +24,7 @@ import { StudySelectorStats } from 'shared/components/query/StudySelectorStats';
 import $ from 'jquery';
 import { serializeEvent } from 'shared/lib/tracking';
 import { ModifyQueryParams } from 'pages/resultsView/ResultsViewPageStore';
+import { getServerConfig } from 'config/config';
 
 interface QueryContainerProps {
     store: QueryStore;
@@ -202,11 +203,13 @@ export default class QueryContainer extends React.Component<
                         </div>
                     )}
 
-                {this.store.unknownStudyIds.isComplete && (
-                    <UnknownStudiesWarning
-                        ids={this.store.unknownStudyIds.result}
-                    />
-                )}
+                {this.store.unknownStudyIds.isComplete &&
+                    !!getServerConfig()
+                        .skin_home_page_show_unauthorized_studies === false && (
+                        <UnknownStudiesWarning
+                            ids={this.store.unknownStudyIds.result}
+                        />
+                    )}
 
                 {this.store.forDownloadTab && (
                     <div className={'alert alert-danger'} role="alert">
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -65,6 +65,7 @@ import {
 import { isMixedReferenceGenome } from 'shared/lib/referenceGenomeUtils';
 import { getSuffixOfMolecularProfile } from 'shared/lib/molecularProfileUtils';
 import { VirtualStudy } from 'shared/api/session-service/sessionServiceModels';
+import { isQueriedStudyAuthorized } from 'pages/studyView/StudyViewUtils';
 
 // interface for communicating
 export type CancerStudyQueryUrlParams = {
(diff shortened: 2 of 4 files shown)


src/shared/lib/JsonPathUtils.ts
```diff
@@ -5,21 +5,28 @@ export function hasJsonPathPlaceholders(message: String) {
...
         let jp = require('jsonpath');
         placeholders.forEach(placeholder => {
-            let placeholderReplaceValue = jp.query(
-                studyMetadata,
-                placeholder.replace(/["'{}]/g, '')
-            );
-            if (placeholderReplaceValue.length === 0)
+            let placeholderReplaceValue = '';
+            if (placeholder === '{$.studyId}') {
+                placeholderReplaceValue = studyId.toString();
```
pvannierop, line 18: StudyId is a string already. Why use `toString()`?

pvannierop (changes requested):
One or two sentences should be added to the documentation that this is possible.

src/shared/lib/JsonPathUtils.ts
dianab0: Fixed

dianab0:
Docs update PR: https://github.com/cBioPortal/cbioportal/pull/9272

dianab0:
Added fix for issue https://github.com/cBioPortal/cbioportal/issues/9255

pvannierop (approved):

src/shared/components/query/QueryContainer.tsx
```diff
@@ -202,11 +203,13 @@ export default class QueryContainer extends React.Component<
...
                     )}
 
-                {this.store.unknownStudyIds.isComplete && (
-                    <UnknownStudiesWarning
-                        ids={this.store.unknownStudyIds.result}
-                    />
-                )}
+                {this.store.unknownStudyIds.isComplete &&
+                    getServerConfig()
+                        .skin_home_page_show_unauthorized_studies === false && (
```
pvannierop, line 208:
```suggestion
                    !!getServerConfig()
                        .skin_home_page_show_unauthorized_studies && (
```

pvannierop: This also takes into account that the value is not defined.

pvannierop (changes requested):

src/shared/lib/JsonPathUtils.ts
```diff
@@ -1,25 +1,32 @@
...
     let placeholdersReplaced: boolean = true;
     if (placeholders !== null) {
         let jp = require('jsonpath');
         placeholders.forEach(placeholder => {
-            let placeholderReplaceValue = jp.query(
-                studyMetadata,
-                placeholder.replace(/["'{}]/g, '')
-            );
-            if (placeholderReplaceValue.length === 0)
+            let placeholderReplaceValue = '';
```
alisman, line 16: we should probably put the require() from jsonpath at the top of the file.  not sure why it ended up down here.

src/shared/components/query/QueryContainer.tsx
dianab0: Done

src/shared/lib/JsonPathUtils.ts
dianab0: done

src/shared/components/query/QueryStore.ts
```diff
@@ -825,7 +826,9 @@ export class QueryStore {
             let result: { [studyId: string]: string[] } = {};
 
             _.each(this.physicalStudiesSet.result, (study, studyId) => {
-                result[studyId] = [studyId];
+                if (isQueriedStudyAuthorized(study)) {
+                    result[studyId] = [studyId];
+                }
```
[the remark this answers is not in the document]
dianab0: I didn't add it because the unauthorized studies feature shouldn't apply to the virtual studies. Also in the StudyViewPageStore, I see that the isQueriedStudyAuthorized is added only to the physical studies.

dianab0:
Rebased and pushed everything (including fixes for review comments) under one single commit

src/shared/components/query/QueryStore.ts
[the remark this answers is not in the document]
pvannierop: @alisman 
>it seems at least possible that an unauthorized study could make it's way into a virtual study from a user that has privelages to it.

This is already the case when not running with the unauthorized studies feature. Or am I missing something in your feedback?

PR #4149, opened 2022-01-26
Remove observability from MutationTable columns collection to fix tooltip issue

alisman:
This is pushing observability onto React components in columns, which causes obscure issue in tooltip.

Fixes https://github.com/cbioportal/cbioportal/issues/9247

### DIFF
--- a/src/shared/components/mutationTable/MutationTable.tsx
+++ b/src/shared/components/mutationTable/MutationTable.tsx
@@ -1,6 +1,12 @@
 import * as React from 'react';
 import { observer } from 'mobx-react';
-import { action, observable, computed, makeObservable } from 'mobx';
+import {
+    action,
+    observable,
+    computed,
+    makeObservable,
+    makeAutoObservable,
+} from 'mobx';
 import _ from 'lodash';
 import {
     default as LazyMobXTable,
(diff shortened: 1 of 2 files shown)


src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -257,7 +257,7 @@ const ANNOTATION_ELEMENT_ID = 'mutation-annotation';
 export default class MutationTable<
     P extends IMutationTableProps
 > extends React.Component<P, {}> {
-    @observable protected _columns: Record<
+    @observable.ref protected _columns: Record<
```
alisman, line 260: @onursumer this could screw things up if we are actually mutating columns somewhere.   we are toggling the visible boolean but it seems that toggling visibility still works

onursumer:
@alisman Removing observability seems to break the column filters.

Before:
column_filter_before


After:
column_filter_after

src/shared/components/mutationTable/MutationTable.tsx
onursumer: Looks like column filters are not working anymore though 😞

alisman: @onursumer good find. thanks.  will try to fix

onursumer (approved):
Looks good

PR #4153, opened 2022-01-27
Correct use of case list by gene set hierarchy selector component

pvannierop:
# Problem
The selected samples determine the outcome of the score-based geneset selection in the hierarchy selection component. In the present state the component receives the _defaultSelectedSampleList_ to represent the sample collection. This behavior is incorrect since the gene set selection should depend on the sample list that is currently selected on Query Page. In addition, the dependence of geneset score-based selection on sample set is not made clear to the user.

# Solution
This PR will:
- Correct GenesetHierarchySelector to use the currently selected sample list.
- Add a message to the GenesetHierarchySelector to inform the user of the dependence of geneset score-based selection on sample set (see image below)

image


pvannierop:
replaced by #4158

PR #4162, opened 2022-02-02
API client cleanup

onursumer:
No monkeys have been harmed in the making of this PR.

See also https://www.youtube.com/watch?v=I5yOTJ40qew

### DIFF
--- a/packages/cbioportal-utils/src/api/apiUtils.ts
+++ b/packages/cbioportal-utils/src/api/apiUtils.ts
@@ -1,4 +1,10 @@
+import { Buffer } from 'buffer';
 import _ from 'lodash';
+import * as request from 'superagent';
+
+const BASE64_ENCODING = 'base64';
+const UTF8_ENCODING = 'utf8';
+
 export async function chunkCalls<ParamData, ResponseData>(
     callback: (chunk: ParamData[]) => Promise<ResponseData[]>,
     paramData: ParamData[],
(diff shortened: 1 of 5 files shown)


src/shared/api/urls.ts
```diff
@@ -208,7 +208,8 @@ export function getProxyUrlIfNecessary(url: any) {
 }
 
 export function getOncoKbApiUrl() {
-    return buildCBioPortalAPIUrl(`proxy/oncokb`);
+    // return buildCBioPortalAPIUrl(`proxy/oncokb`);
+    return buildCBioPortalAPIUrl(`proxy/monkeykb`);
```
onursumer, line 212: This is the tentative name of the new endpoint. We can change this one to a non human readable name as well.

src/config/config.ts
```diff
@@ -167,6 +167,7 @@ export function initializeAPIClients() {
     cachePostMethods(GenomeNexusAPI, [], /POST$/);
     cachePostMethods(GenomeNexusAPIInternal, [], /POST$/);
     cachePostMethods(OncoKbAPI);
+    maskApiRequests(OncoKbAPI, getOncoKbApiUrl());
```
onursumer, line 170: should we add the warning/notice to the developer here?

packages/cbioportal-utils/src/api/apiUtils.ts
```diff
@@ -8,3 +10,69 @@ export async function chunkCalls<ParamData, ResponseData>(
...
     return _.flatten(await Promise.all(proms));
 }
+
+function maskObject(obj: any) {
+    return Buffer.from(_.isString(obj) ? obj : JSON.stringify(obj)).toString(
+        'base64'
+    );
+}
+
+function unmaskObject(base64String: string) {
```
alisman, line 20: unmonkify

src/shared/api/urls.ts
alisman: proxy/nothingToSeeHere

src/config/config.ts
alisman: i think it should actually be where we CALL the api because that's where someone combing through the code would likely wind up.

packages/cbioportal-utils/src/api/apiUtils.ts
```diff
@@ -8,3 +10,69 @@ export async function chunkCalls<ParamData, ResponseData>(
     const proms = chunks.map(callback);
     return _.flatten(await Promise.all(proms));
 }
+
+function monkify(obj: any) {
+    return Buffer.from(_.isString(obj) ? obj : JSON.stringify(obj)).toString(
+        'base64'
```
zhx828, line 16: Super minor, but you might want to use a constant for the encoding?

packages/cbioportal-utils/src/api/apiUtils.ts
```diff
@@ -8,3 +10,69 @@ export async function chunkCalls<ParamData, ResponseData>(
...
+
+export function maskApiRequests(apiClient: any, domain: string) {
+    const oldRequestFunc = apiClient.prototype.request;
+
+    apiClient.prototype.request = (
+        method: string,
+        url: string,
+        body: any,
+        headers: any,
+        queryParameters: any,
```
zhx828, line 32: Should we enforce this at least to be array|object?

zhx828 (commented):
🙊

src/config/config.ts
onursumer: We call the oncokb API client functions in the packages (mutation mapper, variant view, etc.) as well. It may not be a good idea to add a warning/notice in those packages. Because ideally those standalone packages should work with the public oncokb API endpoints without crashing. We should revisit the packages (in a follow up PR) to make sure that they still work with the public endpoints to a certain extend.

On the other hand, we still want to warn others not to configure their client to consume the new portal backend endpoint. They should either use the public endpoints only or have a proper configuration with their own API tokens if they need the private data. I guess we can add a warning message where we call the api functions under the main `src` files. We just need to be clear that we don't want them to configure the client to consume the portal endpoint.

packages/cbioportal-utils/src/api/apiUtils.ts
onursumer: Ideally yes. But this is inline with the function that we override:

https://github.com/cBioPortal/cbioportal-frontend/blob/fa86ca51b2a535aad89c2428270e35f1068e75f6/packages/oncokb-ts-api-client/src/generated/OncoKbAPI.ts#L375

alisman (approved):

onursumer:
Looks like failing screenshot tests for standalone oncoprinter and mutation mapper is valid. Need to investigate what's going wrong.

src/shared/api/urls.ts
```diff
@@ -208,7 +208,7 @@ export function getProxyUrlIfNecessary(url: any) {
 }
 
 export function getOncoKbApiUrl() {
-    return buildCBioPortalAPIUrl(`proxy/oncokb`);
+    return buildCBioPortalAPIUrl(`proxy/nothingToSeeHere`);
```
alisman, line 211: @onursumer i think the idea here was that we would also mask the path.   but if that's problematic we could change this to something which doesn't attract attention.   this shows up in network tab so, kind of defeats the purpose
image

onursumer: We can do `proxy/x3345QR` but we need a new backend PR for that. Do you think we should also get rid of `proxy`?

alisman: sure, if it's all the same, why not get rid of that too.

alisman: and yes, we should do new backend PR

onursumer: @alisman Looks like it is not straightforward to get rid of proxy. We need to create a new endpoint class, and also move shared private methods to a utility class or something like that. Let's keep proxy for now?

onursumer:
We should deploy the changes introduces by cBioPortal/cbioportal#9362 before merging this PR

PR #4159, opened 2022-01-31
Show namespace columns in MAF file in Mutation Table component

pvannierop:
Replaces #4070 from different repo.

# Description

This PR will show custom columns in the MAF file imported with the namespace import mechanism in Mutation Table components of Results View, Patient View and Standalone Mutation Mapper.

# Comment
- Passing of e2e-localdb tests depends on acceptance of PR https://github.com/cBioPortal/cbioportal/pull/9086.

# Video
Peek 2021-12-02 13-35

PR #4163, opened 2022-02-02
Show namespace columns in MAF file in Mutation Table component

pvannierop:
This is a redo from PR #4159. There were breaking bugs in this and it was reverted. Here regression tests are added for mutation table filtering.

# Description

This PR will show custom columns in the MAF file imported with the namespace import mechanism in Mutation Table components of Results View, Patient View and Standalone Mutation Mapper.

# Comment
- Passing of e2e-localdb tests depends on acceptance of PR https://github.com/cBioPortal/cbioportal/pull/9086.

# Video
Peek 2021-12-02 13-35

# Tests
- e2e-localdb tests are included for namespace feature.
- e2e-remote test is included for filter functionality for mutation tables.

### DIFF
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -155,4 +155,5 @@ export interface IServerConfig {
     referenceGenomeVersion: string;
     skin_home_page_show_unauthorized_studies: boolean;
     skin_home_page_unauthorized_studies_global_message: string;
+    skin_mutation_table_namespace_column_show_by_default: boolean;
 }
--- a/src/pages/patientView/mutation/PatientViewMutationTable.tsx
+++ b/src/pages/patientView/mutation/PatientViewMutationTable.tsx
@@ -396,13 +396,6 @@ export default class PatientViewMutationTable extends MutationTable<
         _.forIn(
             namespaceColumns,
             (column: MutationTableColumn, columnName: string) => {
-                if (
-                    this.props.columns &&
-                    !this.props.columns.includes(columnName)
-                ) {
-                    this.props.columns.push(columnName);
-                }
-
                 this._columns[columnName] = column;
             }
         );
--- a/src/pages/patientView/mutation/PatientViewMutationTable.tsx
+++ b/src/pages/patientView/mutation/PatientViewMutationTable.tsx
@@ -3,6 +3,7 @@ import { computed, makeObservable } from 'mobx';
 import {
     default as MutationTable,
     IMutationTableProps,
+    MutationTableColumn,
     MutationTableColumnType,
 } from 'shared/components/mutationTable/MutationTable';
 import SampleManager from '../SampleManager';
@@ -21,6 +22,8 @@ import { getDefaultClonalColumnDefinition } from 'shared/components/mutationTabl
 import { getDefaultExpectedAltCopiesColumnDefinition } from 'shared/components/mutationTable/column/expectedAltCopies/ExpectedAltCopiesColumnFormatter';
 import { ASCNAttributes } from 'shared/enums/ASCNEnums';
 import AnnotationHeader from 'shared/components/mutationTable/column/annotation/AnnotationHeader';
+import _ from 'lodash';
+import { createNamespaceColumns } from 'shared/components/mutationTable/MutationTableUtils';
 
 export interface IPatientViewMutationTableProps extends IMutationTableProps {
     sampleManager: SampleManager | null;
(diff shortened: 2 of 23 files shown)


src/config/IAppConfig.ts
```diff
@@ -155,4 +155,5 @@ export interface IServerConfig {
     referenceGenomeVersion: string;
     skin_home_page_show_unauthorized_studies: boolean;
     skin_home_page_unauthorized_studies_global_message: string;
+    skin_mutation_table_namespace_column_show_by_default: boolean;
```
alisman, line 158: maybe we make a string of default column IDs?  We can follow up with that but this could get silly.  I think this would actually be pretty useful.

src/pages/patientView/mutation/PatientViewMutationTable.tsx
```diff
@@ -385,6 +388,24 @@ export default class PatientViewMutationTable extends MutationTable<
...
+            this.props.namespaceColumns
+        );
+        _.forIn(
+            namespaceColumns,
+            (column: MutationTableColumn, columnName: string) => {
+                if (
+                    this.props.columns &&
+                    !this.props.columns.includes(columnName)
+                ) {
+                    this.props.columns.push(columnName);
```
alisman, line 403: looks like we are mutating a prop here? that's antipattern.  lets discuss.

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -307,6 +336,33 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
...
+                            namespaceName,
+                            namespaceColumnName
+                        );
+                        columnIds.add(columnName);
+                        this.mutationMapperStoreConfig[
+                            'filterAppliersOverride'
+                        ]![
+                            columnName
+                        ] = createCategoricalFilter((d: Mutation) =>
+                            CategoricalNamespaceColumnFormatter.download(
```
alisman, line 355: maybe a comment explaining why we are using download?   i mean, the problem is perhaps that the "download" is too specific

src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
```diff
@@ -173,6 +176,24 @@ export default class ResultsViewMutationTable extends MutationTable<
             };
         }
 
+        // generate namespace columns
+        const namespaceColumns = createNamespaceColumns(
```
alisman, line 180: again it seems odd that we accept a prop of namespace columns only the mutate the columns prop.  seems like we should be doing this where the columns prop is prepared?

src/shared/components/mutationMapper/MutationMapperUtils.ts
```diff
@@ -32,3 +34,39 @@ export function createVariantAnnotationsByMutationFetcher(
...
 }
+
+export function buildNamespaceColumnConfig(
+    mutations: Mutation[]
+): NamespaceColumnConfig {
+    if (!mutations) {
+        return {};
+    }
+    const namespaceConfig: NamespaceColumnConfig = {};
+    const nameSpaces = _.flatMap(mutations, m => _.keys(m.namespaceColumns));
```
alisman, line 45: it seems strange here that something of type Mutation[] would have something called "namespaceColumns" on it

pvannierop: I do not see why that would be strange. Maybe `namespaceValues` would be better. But would we go through the motions in the frontend and backend to update this?

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
pvannierop: Download here means the value formatted for download. I just copied this from other implementations.

src/pages/patientView/mutation/PatientViewMutationTable.tsx
pvannierop: Fixed.

src/config/IAppConfig.ts
pvannierop: We can. Good idea actually. Lets do this in a subsequent PR.

src/pages/resultsView/mutation/ResultsViewMutationMapperStore.ts
```diff
@@ -222,6 +226,31 @@ export default class ResultsViewMutationMapperStore extends MutationMapperStore
             }
         });
 
+        // Add numerical namespace column definitions.
```
alisman, line 229: not necessary for this PR but this great example of something that should be in utility function and not bloating store

src/shared/components/mutationMapper/MutationMapperUtils.ts
alisman: Again not necessary to address in this PR but what I object to is A) having something called "column" on Mutation type.  And "namespace".    Shouldn't this be more generic?   Like, customValues or something.  The fact that they belong to namespaces seems like a property of the individual values themselves.   Again, I know this is very late to the game and I am not suggesting we make a change, but worth bringing up.  Adding @inodb here.  Is this THE one and only way we allow people to extend the Mutation data?

alisman (approved):

PR #4167, opened 2022-02-03
Add async hide/show check for custom tabs

alisman:
Introduce a way to configure custom tabs such that an async check can be run to see whether or not the tab should show for a given view of the page.  E.g. we should rerun this check when the query changes on the result page or when the patient/sample change on patient page.

The hideAsync property is a function which MUST return a promise.  When the promise resolves with a boolean value, we will either show or hide the tab button.   Note that if the user is already routed to the tab, we should always show the button even if the hydeAsync function returns false.   In that case we should render some N/A placeholder for the tab content.
```
 {
     title: 'Async Tab 2',
     id: 'customTab2',
     location: 'RESULTS_PAGE',
     hideAsync: `()=>{
                return new Promise((resolve)=>{
                     setTimeout(()=>{
                         resolve(true);
                     }, 2000);
                });
     }`,
 },

```


my-index.ejs
```diff
@@ -14,21 +14,87 @@
...
-            serverConfig:{}
+
+//             serverConfig:{
+//
+// //             custom_tabs:[
+// //                             {
+// //                                 "title": "Custom Tab",
+// //                                 "location": "RESULTS_PAGE",
+// //                                 "mountCallbackName": "renderCustomTab1",
+// //                                 "hide":false,
```
pvannierop, line 25: Should this file be committed?

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -106,6 +110,13 @@ export default class GroupComparisonPage extends React.Component<
         this.urlWrapper.destroy();
     }
 
+    @computed get customTabs() {
+        return prepareCustomTabConfigurations(
+            getServerConfig().custom_tabs,
+            'COMPARISON_PAGE'
```
pvannierop, line 116: So, you extended the custom tabs to Comparison Page? Nice would be good to make that clean in docs and release notes.

src/pages/patientView/PatientViewPageTabs.tsx
```diff
@@ -0,0 +1,779 @@
...
+    //                     key={getPatientViewResourceTabId(
+    //                         'customTab' + i
+    //                     )}
+    //                     id={getPatientViewResourceTabId(
+    //                         'customTab' + i
+    //                     )}
+    //                     unmountOnHide={
+    //                         tab.unmountOnHide ===
+    //                         true
+    //                     }
```
pvannierop, line 766: Remove commented lines?

src/routes.tsx
```diff
@@ -263,15 +263,15 @@ function comparisonTabParamValidator() {
...
         `patient\/${PatientViewResourceTabPrefix}.+`
     );
     const studyViewRegex = new RegExp(`study\/${StudyViewResourceTabPrefix}.+`);
+    const customTabRegex = /.+\/customTab\d/;
     return (
-        location.pathname.match(resultsRegex) !== null ||
         patientViewRegex.test(location.pathname) ||
-        studyViewRegex.test(location.pathname)
+        studyViewRegex.test(location.pathname) ||
+        customTabRegex.test(location.pathname)
```
pvannierop, line 274: Just asking, but should there not be a comparisonView test as well?

end-to-end-test/remote/specs/core/customTabs.spec.js
```diff
@@ -0,0 +1,397 @@
...
+            $('div=tab for patient TCGA-AP-A053').isDisplayed(),
+            true,
+            'We first patient'
+        );
+    });
+});
+
+runTests('ResultsView', resultsUrl, 'RESULTS_PAGE');
+
+runTests('PatientView', patientUrl, 'PATIENT_PAGE');
```
pvannierop, line 397: I noticed that Comparison View is also supported not. Should we not include a test for it?

pvannierop (commented):
Wow, this was quite the project! Looks good. I have some remarks and questions.

src/routes.tsx
alisman: i'll make a ticket for it and do it separately

pvannierop (approved):
Nice work, and thanks again for doing this.

PR #4169, opened 2022-02-06
Gene Panel Charts and Filters

Luke-Sikina:


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -420,6 +427,29 @@ export class StudyViewPageStore
         );
     }
 
+    private setCustomChart(chartMeta: ChartMetaWithDimensionAndChartType) {
+        const uniqueKey = chartMeta.uniqueKey;
+        // initialize chartsType.get(uniqueKey) if not already set
+        if (!this.chartsType.has(uniqueKey)) {
+            this.chartsType.set(uniqueKey, chartMeta.chartType);
```
alisman, line 434: i know this is legacy but should probably be this.chartTypes or something even more descriptive.  if it can be easily improved lets do it

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -420,6 +427,29 @@ export class StudyViewPageStore
...
+        }
+        const chartType = this.chartsType.get(uniqueKey);
```
alisman, line 436: picky but would be clearer if we should obtaine this from passed chartMeta, no?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -420,6 +427,29 @@ export class StudyViewPageStore
         );
     }
 
+    private setCustomChart(chartMeta: ChartMetaWithDimensionAndChartType) {
```
alisman, line 430: should be @action since it mutates state (and multiple times in fact)

alisman: should we call this addCustomChart?  set is a little fuzzy

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -539,6 +569,32 @@ export class StudyViewPageStore
...
+                                displayName: 'Gene Panels: ' + profileName,
+                                description: '',
+                                chartType: ChartTypeEnum.PIE_CHART,
+                                dataType: ChartMetaDataTypeEnum.GENOMIC,
+                                patientAttribute: false,
+                                dimension: { w: 1, h: 1 },
+                                priority: 999,
+                                renderWhenDataChange: false,
+                            };
+                            this.setCustomChart(chartMeta);
```
alisman, line 591: since we are calling this inside a loop and it mutates state, we need to use mobx "action" to update state, so that we don't get multiple update cycles.  easiest thing to do would be wrap in mobx runInAction.  alternative: allow setCustomChart to accept multiple chartMeta.  This is a little nicer because it's only one level of "action"

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2691,6 +2757,41 @@ export class StudyViewPageStore
...
+            .map(mp => mp.molecularProfileId)
+            .filter(mp => mp.endsWith(suffix));
+        const genePanelFiltersForProfiles = values.map(panel => {
+            return {
+                genePanel: panel,
+                molecularProfileIds,
+            };
+        });
+
+        const otherGenePanelFilters = this.genePanelFilters.filter(
```
alisman, line 2774: this seems strange.  ```since molecularProfileIds``` is made locally here, i don't think it will EVER be referentially equal to an existing array within genePanelFilters.

alisman: i.e. even if it has the same items, it won't evaluate equal ```[1,2,3]===[1,2,3]```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3066,6 +3169,8 @@ export class StudyViewPageStore
     @action
     changeChartVisibility(uniqueKey: string, visible: boolean): void {
         if (visible) {
+            console.log('setting ' + uniqueKey);
```
alisman, line 3172: console

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3537,6 +3652,21 @@ export class StudyViewPageStore
...
+        return this.genePanelFilters
+            .filter((f: GenePanelFilter) => {
+                return f.molecularProfileIds
+                    .map(this.extractSuffix)
+                    .includes(extractGenePanelSuffix(key));
+            })
+            .map((f: GenePanelFilter) => f.genePanel);
+    }
+
+    private extractSuffix(molecularProfileId: string): string {
```
alisman, line 3665: extractSuffixFromMolecularProfileId.   also there has to be some utility file where we could put this.  it shouldn't be a method as it doesn't reference instance

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8233,6 +8407,7 @@ export class StudyViewPageStore
     });
 
     readonly initialMolecularProfileSampleCounts = remoteData({
+        await: () => [this.samples, this.molecularProfiles],
```
alisman, line 8410: curious why this becomes necessary.  we need to be careful about adding load dependencies as they can affect performance of the study view pretty drastically

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -539,6 +569,32 @@ export class StudyViewPageStore
             )
         );
 
+        this.reactionDisposers.push(
```
alisman, line 572: this could arguably be done in onresult handler of the molecularProfileNameSet promise.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3998,6 +4128,48 @@ export class StudyViewPageStore
...
+        await: () => [
+            this.selectedSamples,
+            this.molecularProfiles,
+            this.molecularProfileNameSet,
+        ],
+        invoke: async () => {
+            type Response = {
+                [suffix: string]: { value: string; count: number }[];
+            };
+            // This is a partial StudyViewFilter. We any type it because the type declared
```
alisman, line 4141: there is a Ts helper which allow us to take a type and make all (or some) of its members optional

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3998,6 +4128,48 @@ export class StudyViewPageStore
...
+            // We need the unselected gene panel count data as well as the selected.
+            delete studyViewFilter['genePanelFilters'];
+
+            const counts: Response = await internalClient.fetchGenePanelCountsUsingPOST(
+                {
+                    studyViewFilter,
+                }
+            );
+
+            return Object.keys(counts).reduce((col, p) => {
```
alisman, line 4154: lodash reduce applied to counts will do the same

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8912,6 +9087,45 @@ export class StudyViewPageStore
...
+    public removeGenePanelFilter(filterToRemove: GenePanelFilter) {
+        this.genePanelFilters = this.genePanelFilters.filter(f => {
+            return (
+                f.genePanel != filterToRemove.genePanel &&
+                f.molecularProfileIds != filterToRemove.molecularProfileIds
+            );
+        });
+    }
+
+    private toGenePanelFilter(
```
alisman, line 9113: doesn't appear to use chartUniqueKey param.   could probably just a function rather than hiding its reliance on molecularProfiles (and it's being resolved).   i.e. should accept MolecularProfiles[] as argument

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3192,6 +3207,80 @@ export async function getGenericAssayDataAsClinicalData(
     });
 }
 
+export async function getGenePanelDataAsClinicalData(
```
alisman, line 3210: this is a beautiful function

src/pages/studyView/UserSelections.tsx
```diff
@@ -600,6 +608,66 @@ export default class UserSelections extends React.Component<
...
+        const groupedPanels = filters
+            .map(f => {
+                const profile = f.molecularProfileIds[0].split('_');
+                return {
+                    suffix: profile[profile.length - 1],
+                    filter: f,
+                };
+            })
+            .reduce((prev, cur) => {
+                if (prev.has(cur.suffix)) {
```
alisman, line 621: ```prev[cur.suffix] = [cur.filter, ...prev[cur.suffice]  ]```

src/pages/studyView/UserSelections.tsx
```diff
@@ -600,6 +608,66 @@ export default class UserSelections extends React.Component<
...
+                        }
+                    />
+                );
+            });
+            const label = (
+                <span className={styles.filterClinicalAttrName}>
+                    {'Gene Panels: ' + entry[0] + ''}
+                </span>
+            );
+            components.unshift(label);
```
alisman, line 651: don't love unshift here.  maybe start with the label and push to the array in forEach above

alisman:
@Luke-Sikina if you go to master.cbioportal.org and turn on PR 4169 (this PR).  You will see the gene panel charts show up in the add chart UI but when added, result in empty charts:
image

alisman:
closing for now

PR #4170, opened 2022-02-07
Genome Nexus api clients sync

leexgh:
Update genome nexus api clients

### DIFF
--- a/packages/genome-nexus-ts-api-client/src/generated/GenomeNexusAPIInternal-docs.json
+++ b/packages/genome-nexus-ts-api-client/src/generated/GenomeNexusAPIInternal-docs.json
@@ -35,6 +35,10 @@
       "description": "Annotation Summary Controller"
     },
     {
+      "name": "curious-cases-controller",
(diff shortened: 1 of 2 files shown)


packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
```diff
@@ -1767,182 +1727,6 @@ export default class CBioPortalAPIInternal {
                 return response.body;
             });
         };
-    fetchClinicalDataViolinPlotsUsingPOSTURL(parameters: {
```
adamabeshouse, line 1770: taking this change out is going to cause type errors because this endpoint is being used. can you just do the genome nexus updates? the backend will be updated soon making these in sync

adamabeshouse (changes requested):

adamabeshouse (approved):

Luke-Sikina:
merge pls

onursumer (approved):

PR #4172, opened 2022-02-08
Fix Select All button logic for unauthorized studies

pvannierop:
This PR fixes the Select All button on Query page selecting unauthorized studies when this feature is activated.

### DIFF
--- a/src/shared/components/query/StudyListLogic.ts
+++ b/src/shared/components/query/StudyListLogic.ts
@@ -2,8 +2,8 @@ import _ from 'lodash';
 import {
     CancerTreeNode,
     CancerTypeWithVisibility,
+    NodeMetadata,
 } from './CancerStudyTreeData';
-import { NodeMetadata } from './CancerStudyTreeData';
 import {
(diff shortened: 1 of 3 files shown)


src/shared/components/query/StudyListLogic.ts
```diff
@@ -2,8 +2,8 @@ import _ from 'lodash';
 import {
     CancerTreeNode,
     CancerTypeWithVisibility,
+    NodeMetadata,
```
alisman, line 5: seems like not used

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -121,7 +122,7 @@ export default class StudyList extends QueryStoreComponent<
                         {indentArrow}
 
                         <span>{cancerType.name}</span>
-                        {!!!this.store.forDownloadTab && (
+                        {!this.store.forDownloadTab && (
```
alisman, line 125: for what it's worth i think we got rid of download tab.  no action required.

alisman (approved):

PR #4175, opened 2022-02-10
Fix hrefs on MSKTabs (for opening directly in new tab)

alisman:
Right clicking and "open in new tab" was not working on any pages

fixes https://github.com/cbioportal/cbioportal/issues/9289


inodb (approved):
More complicated than I thought lol - thanks for fixing @ritikakundra will be happy :)!

PR #4179, opened 2022-02-14
Prevent auto publish from failing when yarn.lock is different

onursumer:
Fix cBioPortal/cbioportal#9303

## Checks

### DIFF
--- a/scripts/update_package_version.sh
+++ b/scripts/update_package_version.sh
@@ -15,6 +15,10 @@ if [[ "$GITHUB_RUN_ID" ]]; then
   git config --local user.name "Lerna Publish Bot"
   git config --global github.token $GITHUB_TOKEN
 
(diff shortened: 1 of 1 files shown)


inodb:
@onursumer wait - doesn't this seem like a legitimate error. That is, we should fix the lock file? I guess somehow it's different on linux vs osx or something?

onursumer:
> @onursumer wait - doesn't this seem like a legitimate error. That is, we should fix the lock file? I guess somehow it's different on linux vs osx or something?

This happens because of https://github.com/cBioPortal/cbioportal-frontend/commit/9236b6f671ac4b8d0d8fcf0d9f7f5678ce564c8a. Didn't we actually want to get rid of the `yarn.lock` check?

onursumer:
Closing this in favor of cBioPortal/cbioportal-publisher#5

PR #4188, opened 2022-02-23
Replace legacy tracks with IGV

onursumer:
Fix cBioPortal/cbioportal#9311

Need to fix these before merging:

- [x] Remove the legacy tracks completely (currently for comparison both legacy tracks and IGV are rendered)
- [x] Update screenshots
- [x] Mutation colors are random each time the IGV is rendered. Should match the colors in the table.
- [x] ~~Should point to https://github.com/cBioPortal/igv.js fork instead of https://github.com/onursumer/igv.js~~ Use npm package directly

## Checks


src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -133,40 +434,28 @@ export default class GenomicOverview extends React.Component<
...
     }
 
-    private getTracksWidth(): number {
+    private get tracksWidth(): number {
         return (
             this.props.containerWidth - (this.shouldShowVAFPlot() ? 140 : 40)
         );
     }
 
-    private computeMutationFrequencyBySample(
```
onursumer, line 142: moved to `GenomicOverviewUtils`

src/pages/patientView/genomicOverview/GenomicOverviewUtils.spec.tsx
```diff
@@ -198,4 +200,236 @@ describe('GenomicOverviewUtils', () => {
             );
         });
     });
+
+    describe('computeMutationFrequencyBySample()', () => {
```
onursumer, line 204: tests moved from `GenomicOverview.spec.tsx`

src/pages/patientView/genomicOverview/GenomicOverviewUtils.ts
```diff
@@ -92,3 +95,32 @@ export function sampleIdToIconData(
 
     return lookupTable;
 }
+
+export function computeMutationFrequencyBySample(
```
onursumer, line 99: moved from `GenomicOverview.tsx`

src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -118,6 +385,12 @@ export default class GenomicOverview extends React.Component<
         );
     }
 
+    componentWillReceiveProps(nextProps: Readonly<IGenomicOverviewProps>) {
```
alisman, line 388: this is so it doesn't render if locus hasn't actually changed?

onursumer: Right. This is to prevent resetting zoom state on window resize. Hopefully this didn't introduce some unknown side effects.

alisman: We should put a comment on this to explain

inodb (commented):
Looks awesome! I think the main thing I noticed is:

- There's a short time where the component just shows up as blank. I'm guessing this is in the IGV component itself so might be hard to fix. It happens both when loading the page the first time and when clicking on a mutation
- There's a lot of whitespace: (1) between the first track and the chromosome track and (2) the chromosome track has a lot. Old representation was better about space use. Not sure if we can tweak this much more

onursumer:
> Looks awesome! I think the main thing I noticed is:
> 
> * There's a short time where the component just shows up as blank. I'm guessing this is in the IGV component itself so might be hard to fix. It happens both when loading the page the first time and when clicking on a mutation

Yeah, this might be a little tricky to fix. We may ask IGV team to look into this one.

> * There's a lot of whitespace: (1) between the first track and the chromosome track and (2) the chromosome track has a lot. Old representation was better about space use. Not sure if we can tweak this much more

We may still be able to reduce whitespace further. I will look into it.

onursumer:
Looks like sample mutation data do not always get sorted properly. e2e tests actually caught this problem. Here is an example where sample mutation data don't match the sample order: https://deploy-preview-4188--cbioportalfrontend.netlify.app/patient?caseId=P04&studyId=lgg_ucsf_2014.

This needs further debugging. The problem might be in the `compareFn` defined here: https://github.com/cBioPortal/cbioportal-frontend/blob/8198ddfc55b48abc4b96e2f920a74d419160273f/src/pages/patientView/genomicOverview/GenomicOverview.tsx#L167

We should not merge the PR before fixing the sorting issue.

onursumer:
The sort function seems working fine. Looks like IGV does not always keep the sample order intact. Need to investigate further to see if it is possible to force the sample order in IGV.

onursumer:
Looks like this is a know igv.js issue. I have left a comment there as well. Let's see if we can get a quick fix. https://github.com/igvteam/igv.js/issues/1480#issuecomment-1113689007

onursumer:
Sorting issue should be fixed now.

PR #4191, opened 2022-02-28
Don't show legacy tooltip (qtip) when hovering VAF plot on patient page

alisman:
Tooltip seems superfluous given that the actual feature (also in a tooltip) appears on hover over the thumbnail


inodb (approved):
LGTM! We no longer need a tooltip for the sample label

PR #4193, opened 2022-03-01
Treatment Target Frontend

Luke-Sikina:
Add new charts on study view when `AGENT_TARGET` is defined in timeline data:
Screen Shot 2022-03-17 at 3 43 05 PM


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6316,6 +6375,27 @@ export class StudyViewPageStore
             );
         }
 
+        if (this.shouldDisplaySampleTreatmentTarget.result) {
```
alisman, line 6378: is it weird that you test just for the sample (shouldDisplaySampleTreatmentTarget) but then change viseability for both sample and patient?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -965,7 +967,11 @@ export function isFiltered(
             (!filter.patientTreatmentGroupFilters ||
                 _.isEmpty(filter.patientTreatmentGroupFilters.filters)) &&
             (!filter.sampleTreatmentGroupFilters ||
-                _.isEmpty(filter.sampleTreatmentGroupFilters.filters)))
+                _.isEmpty(filter.sampleTreatmentGroupFilters.filters)) &&
```
alisman, line 970: there's a few places in this PR where code could benefit from the new optional chaining feature of js https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

alisman:
@Luke-Sikina maybe we use this PR to try ideas to reduce bload of studyview store.  let discuss

PR #4196, opened 2022-03-03
Use variant class as well when generating the OncoKB query id

onursumer:
Fix cBioPortal/cbioportal#9309

This is a workaround. We may still end up with duplicate ids for other cases when the site 2 gene id is missing.

Archer_fusions_annotated


zhx828 (approved):
LGTM

PR #4197, opened 2022-03-03
Adapt oql logic to handle structural variant alterations (associated with two genes)

alisman:


src/shared/lib/oql/AccessorsForOqlFilter.ts
```diff
@@ -149,12 +149,25 @@ export default class AccessorsForOqlFilter
...
+    //     return (d as any).gene.hugoGeneSymbol;
+    // }
+
     public gene(d: Datum) {
         if (
             this.molecularAlterationType(d.molecularProfileId) ===
             AlterationTypeConstants.STRUCTURAL_VARIANT
         ) {
-            return (d as StructuralVariant).site1HugoSymbol;
+            return [
```
alisman, line 167: @adamabeshouse it's odd that TS allows me to return an array here, no?  I would expect that IAccessorsForOqlFilter interface to enforce string return type.

adamabeshouse: yeah, I have no idea why that is.

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -376,7 +376,8 @@ function isDatumWantedByOQL<T>(
     accessors: IAccessorsForOqlFilter<T>
 ): boolean {
     //  Out: Boolean, whether datum is wanted by this OQL query
-    const gene = accessors.gene(datum).toUpperCase();
+
+    const gene = accessors.gene(datum);
```
adamabeshouse, line 380: not to upper case?

src/shared/lib/oql/AccessorsForOqlFilter.ts
```diff
@@ -149,12 +149,25 @@ export default class AccessorsForOqlFilter
...
+                (d as StructuralVariant).site1HugoSymbol,
+                (d as StructuralVariant).site2HugoSymbol,
+            ];
         }
         return (d as any).gene.hugoGeneSymbol;
```
adamabeshouse, line 172: should probably return this as an array now and change the interface

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -394,15 +395,28 @@ function isDatumWantedByOQL<T>(
 function isDatumWantedByOQLLine<T>(
     query_line: SingleGeneQuery,
     datum: T,
-    datum_gene: string, // lower case gene in datum - passed instead of reaccessed as an optimization
+    datum_gene: string | string[], // lower case gene in datum - passed instead of reaccessed as an optimization
```
adamabeshouse, line 398: lets just standardize and always use string[], less branches

adamabeshouse (commented):
I would be extremely surprised if there aren't other parts of the code to touch, considering that I wrote so much oncoprint, oql, and possibly other code that assumes each gene has its own object for an alteration. Lets definitely keep a close eye on the tests for this one.

src/shared/lib/oql/AccessorsForOqlFilter.ts
alisman: I don't want to do that because using array is not as performant and most of the time it's just a string

Luke-Sikina:
Ready for review, need to inform product team. Maybe dedicate steering call to testing?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -284,6 +286,7 @@ import {
 } from 'shared/api/session-service/sessionServiceModels';
 import { ICBioData } from 'pathway-mapper';
 import { getAlterationData } from 'shared/components/oncoprint/OncoprintUtils';
+import { getHugoGeneSymbols } from 'pages/studyView/StudyViewComparisonUtils';
```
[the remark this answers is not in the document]
alisman: turns out wasn't used!!!

onursumer:
@alisman Not sure if this is an issue on cBioPortal side or OncoKB side, but looks like there is sometimes annotation discrepancy between results view and patient view.

For example go to https://beta.cbioportal.org/results/mutations?cancer_study_list=msk_impact_2017&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cstructural_variants%2Ccna&case_set_id=msk_impact_2017_cnaseq&gene_list=PTEN%2520EGFR&geneset_list=%20&tab_index=tab_visualize&Action=Submit&mutations_gene=PTEN and search for `P-0006378-T01-IM5`. You will see that the fusion is annotated as `oncogenic`.

results_view_likely_oncogenic_fusion

However, the corresponding structural variant is not annotated the same way on the patient view page. See https://beta.cbioportal.org/patient?sampleId=P-0006378-T01-IM5&studyId=msk_impact_2017

patient_view_sv_table

src/routes.tsx
```diff
@@ -209,6 +209,18 @@ function ResultsViewQueryParamsAdjuster(oldParams: ResultsViewURLQuery) {
...
     }
+
+    // we used to call the structural variant alteration class "fusions"
+    // we have generalized it to structural variants
+    const profileRegex = /fusion/;
+    if (profileRegex.test(oldParams.profileFilter)) {
+        newParams.profileFilter = oldParams.profileFilter.replace(
+            profileRegex,
+            'structural_variants'
+        );
```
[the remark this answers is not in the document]
onursumer: oh, didn't know that this was for backward compatibility. now it makes more sense :)

onursumer:
@alisman issue seems to be fixed. thanks!

PR #4201, opened 2022-03-08
Filter na based on customized na category

leexgh:
Fix https://github.com/cBioPortal/cbioportal/issues/9372


src/pages/groupComparison/ClinicalDataUtils.ts
```diff
@@ -0,0 +1,7 @@
+import { getServerConfig } from 'config/config';
+
+export function getComparisonCategoricalNaValue(): string[] {
+    const rawString = getServerConfig().comparison_categorical_na_value;
```
inodb, line 4: is it ever possible this value is null or is the default in the server configuration empty string?

inodb (approved):
It looks good - one minor thing is that the `comparison_categorical_na_value` might be better called `comparison_categorical_na_values`, since it can be multiple

leexgh:
A bunch of tests failed but not related

PR #4205, opened 2022-03-15
Fix tooltip overflow

Nesopie:
Fix cBioPortal/cbioportal#8626 
- Fixed tooltip text from overflowing from the box
Before: 
Screenshot 2022-03-16 at 12 56 50 AM
After: 
Screenshot 2022-03-16 at 12 57 19 AM


package.json
```diff
@@ -354,6 +353,7 @@
     "jest-canvas-mock": "^2.3.1",
     "jest-environment-jsdom": "^27.2.0",
     "json": "^9.0.6",
+    "lerna": "^3.19.0",
```
alisman, line 356: lets try to avoid update to package.json

packages/cbioportal-frontend-commons/src/components/scatterPlot/styles.scss
```diff
@@ -4,7 +4,7 @@
         display: block !important;
         max-width: 600px !important;
         div:last-child {
-            white-space: nowrap;
+            // white-space: nowrap;
```
alisman, line 7: @Nesopie thanks so much for this PR.  I think you're right that a good approach is to allow wrapping inside tooltip, however, I'm pretty sure we must have forbidden it for a reason.   Perhaps see if you can use git blame to find the commit in which the this line was introduced.  We might get lucky and figure out why it was this way in the first place.

Nesopie: Yeah, I thought about the other stuff breaking. The tooltips inside the scatter points did not have any effect on them which is why I though it was a good idea to make a PR. I'll look into it further

alisman:
@Nesopie how come you closed?

Nesopie:
@alisman because it was a mess - with the delete package.json and all those files. I'll reopen it once I have further investigated this issue

packages/cbioportal-frontend-commons/src/components/scatterPlot/styles.scss
Nesopie: @alisman The whole file was introduced in a single commit - not just the single line. Is it safe to assume that this was not done to fix some bug?

alisman: @Nesopie i see.    not sure.   i wonder if you can add a property to the component "mutlLine" or "wrapText" which would allow us to do this in a targetted way?

alisman: I mean add a property to the <DefaultTooltip> component.

Nesopie: Sure! I'll try doing that.

Nesopie: @alisman It looks like the group selector buttons use the react-bootstrap Popover component instead of the DefaultTooltip component. Adding an inline style whiteSpace: normal in the popover component for the group selector buttons seems to fix it. Can I proceed with this?

PR #4212, opened 2022-03-24
Fix virtual study descriptions for download page buttons

alisman:
Fix [#9381](https://github.com/cbioportal/cbioportal/issues/9381). The description field for virtual studies made from the download page had an erroneous count


adamabeshouse (approved):
looks fine to me

PR #4214, opened 2022-03-24
Fix download when number of samples and gene panel id differs

leexgh:


inodb (approved):
LGTM - @pvannierop is this fine?

pvannierop:
@leexgh @inodb I do not quite understand this fix. The purpouse of the `Gene panel` column is to show what gene panel was used to analyze a respective sample. This information is independent of whether a gene was profiled (can show `mutated` or `not mutated`) or was `not profiled`. My intuition is to correct the logic used for downloading, rather than changing the information shown in the columns to fix downloading. I propose additional discussion. Perhaps I do not understand the problem correctly.

leexgh:
It's not just changing the information shown in the columns, it's also our logic for gene panel column. There are 2 ids in Samples column, 3 ids in Gene panel column. Our logic for download is: the number of data in these columns must be the same. 
The reasons I prefer not fixing the download itself are: 
1. the download file is like a MAF file, each row represents a sample. For example:
image
But if we include samples with not profiled gene panel id, that row would have multiple empty cells due to no information (e.g. gene panel, Allele Freq, Variant Reads, Ref Reads, Variant reads(normal), Ref reads(Normal))
image

2. Logic for `Samples` and `Gene Panel` is conflicting, so that causes only 2 ids in Samples column but 3 ids in Gene panel column. 
3. It's safe to remove not profiled gene panel id because when you hover over the not profiled sample(`-` in middle of samples column), you can still find Gene Panel info in tooltip.
image

pvannierop:
@leexgh 
>It's safe to remove not profiled gene panel id because when you hover over the not profiled sample(- in middle of samples column), you can still find Gene Panel info in tooltip.

It may be safe, but I still consider not showing the panel id in the `Gene panel` column a hack. Even when the gene was not profiled for a sample the sample was analyzed with the gene panel and the panel should show in the the `Gene panel` column. What you are doing it solve an issue by conforming the information that is displayed to the user in a misleading way. For this reason alone I think you are not adequately addressing the real problem.

leexgh:
@pvannierop Thanks for your reply. I don't quite get why "What you are doing it solve an issue by conforming the information that is displayed to the user in a misleading way." In my opinion the real problem is we have logic conflicting,  
my fix is correcting the existing bug in our code. 

`Samples` column logic (see the last column in the third and fourth row)
| case | is profiled | is mutated | samples column result |
| --- | --- | --- | --- | 
| 1 | yes | no | hide |
| 2 | yes | yes | show sample |
| 3 | no | no | **not profiled** |
| 4 | no | yes | **not profiled** |

`Gene panel` column logic correction (see the last column in the third and fourth row)
Before the arrow is previous logic, and after the arrow is corrected logic that matches the logic we are using in `Samples` column.
|case|is profiled|is mutated|gene panel column result|
|--|--|--|--|
|1|yes|no|hide|
|2|yes|yes|show gene panel id|
|3|no|no|**profiled --> _not profiled_**|
|4|no|yes|**profiled --> _not profiled_**|

If you have better solutions, could you give me some suggestions?

pvannierop:
@leexgh @inodb I am starting to lose track of this conversation. Please correct me if I am wrong, but I am under the impression that you intend to remove the entry from the `Gene panel` column when the gene for a respective sample was `not profiled` for said sample. Is this correct? If so, I do indeed object.

Perhaps I am missing something here, but, again, to me at this point it appears you are trying to solve a problem in a suboptimal way. Would it be a good idea to schedule a short call so that we can further elaborate on this?

leexgh:
We decide to keep gene panel column as is, only remove unprofiled gene panel ids for download.

inodb (approved):
Looks innocuous to me

pvannierop (changes requested):
Yes, I think this approach is better than the previous. However, I would update the PR text with an accurate description for your fix. When someone will revisit the PR text, this may be a cause for confusion.

PR #4220, opened 2022-04-01
Don't call mutation assessor for download when it's disabled

leexgh:
Fix: https://github.com/genome-nexus/genome-nexus/issues/600
Mutation table download shouldn't call mutation assessor when it's disabled.
Test: 
`localStorage.frontendConfig = '{"serverConfig":{"show_genomenexus_annotation_sources":"mutation_assessor"}}'`
vs
`localStorage.frontendConfig = '{"serverConfig":{"show_genomenexus_annotation_sources":""}}'`
Download a mutation table and try with and without mutation_assessor. The download file should have`MutationAssessor` in `Functional Impact` column only when it's enabled

### DIFF
--- a/src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
+++ b/src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
@@ -455,10 +455,12 @@ export default class FunctionalImpactColumnFormatter {
             data,
             siftPolyphenCache
         );
-        const mutationAssessorCacheData = FunctionalImpactColumnFormatter.getDataFromCache(
-            data,
-            mutationAssessorCache
-        );
+        const mutationAssessorCacheData = shouldShowMutationAssessor()
+            ? FunctionalImpactColumnFormatter.getDataFromCache(
+                  data,
+                  mutationAssessorCache
+              )
+            : null;
 
         const siftData = siftPolyphenCacheData
             ? this.getSiftData(siftPolyphenCacheData.data)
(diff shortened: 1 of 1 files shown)


src/shared/components/mutationTable/column/FunctionalImpactColumnFormatter.tsx
```diff
@@ -60,14 +60,13 @@ class FunctionalImpactColumnTooltip extends React.Component<
     }
 
     legend() {
-        const showMutationAssessor = shouldShowMutationAssessor();
```
onursumer, line 63: Is there a specific reason why we want to call the `shouldShowMutationAssessor` function multiple times instead of once?

leexgh: Because sometimes serverConfig is not ready when it runs `const showMutationAssessor = shouldShowMutationAssessor();`, which would be false and doesn't refresh. But good point, better to have a flag const rather than calling the function itself.

src/shared/lib/genomeNexusAnnotationSourcesUtils.ts
```diff
@@ -5,3 +5,5 @@ export function shouldShowMutationAssessor() {
         getServerConfig().show_genomenexus_annotation_sources
     );
 }
+
+export const SHOW_MUTATION_ASSESSOR = shouldShowMutationAssessor();
```
onursumer, line 9: Here, we may have the same problem of `serverConfig` not being ready yet, right? Is there a way to make sure that we call the function after `serverConfig` is ready?

onursumer:
@leexgh Not critical at this point but it doesn't seem to call mutation assessor anymore even if I do `localStorage.frontendConfig = '{"serverConfig":{"show_genomenexus_annotation_sources":"mutation_assessor"}}'`. Are you able to make it call mutation assessor for the latest netlify preview link?

leexgh:
@onursumer Actually I found it's a bit tricky to wait for `serverConfig`, because default serverConfig is empty, waiting for the correct config from server, then overrides itself. Seems const and static values are created before `serverConfig` overrides. So probably better to call serverConfig inside component

onursumer (approved):
LGTM!

onursumer:
Failing tests don't seem related, merging.

PR #4226, opened 2022-04-11
An else if statement is added in the function toConditionalPrecision …

aswin2108:
Fix cBioPortal/cbioportal#9192

## Screenshots
**Before:** 
image

**After**
image

Currently, all the tests are passing in the local repo.


src/shared/lib/NumberUtils.ts
```diff
@@ -5,6 +5,8 @@ export function toConditionalPrecision(
 ): string {
     if (0.000001 <= number && number < threshold) {
         return number.toExponential(precision);
+    } else if (number === 0) {
+        return '< 10^-10';
```
onursumer, line 9: Not sure if this is the right place to add this condition. `toConditionalPrecision` seems like a general purpose utility function used in multiple places, not only in the comparison page. I would add this check directly in the comparison page to make sure that we don't accidentally break other components.

dippindots (approved):
@aswin2108 This change looks good to me. Thanks!

onursumer (changes requested):

src/shared/lib/NumberUtils.ts
onursumer: @aswin2108 Did you try using the table value directly? Looks like we have an already working implementation there. Why not just do the same thing here? https://github.com/cBioPortal/cbioportal-frontend/blob/099e4e9f3aba66bf83b64167d81d7be29b002d0a/src/pages/resultsView/survival/SurvivalPrefixTable.tsx#L222

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -365,6 +365,8 @@ export default class SurvivalChart
             showNA
         ) {
             return 'N/A (<10 cases in a group)';
+        } else if (this.props.pValue === 0) {
```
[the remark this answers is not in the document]
onursumer: @aswin2108 Did you try using the table value directly? Looks like we have an already working implementation there. Why not just do the same thing here? https://github.com/cBioPortal/cbioportal-frontend/blob/099e4e9f3aba66bf83b64167d81d7be29b002d0a/src/pages/resultsView/survival/SurvivalPrefixTable.tsx#L222

alisman:
@aswin2108 were you able to respond to @onursumer's comments?

aswin2108:
> @aswin2108 were you able to respond to @onursumer's comments?

Actually, my first approach was the same as he said but since it was not working I changed it and pushed the pr. Then after receiving the comments from @onursumer, I tried several times to rectify this error, I had a long conversation with him on this topic. When I approached it through `? toConditionalPrecisionWithMinimum(d.pValue, 3, 0.01, -10)`  `{Object,object}` is getting displayed. I tried different ways but the same outcome. In the middle of it I noticed a mistake within this, which I had already posted in https://github.com/cBioPortal/cbioportal/issues/9192 as per @onursumer's directions, pls look into that also. Sorry for the late reply!

aswin2108:
@alisman is it happening because it is inside of an SVG component? And in your knowledge are there any utility functions currently existing which could be used to get the information needed in the form of a string and then using that function in this case so that it displays it correctly!

PR #4227, opened 2022-04-11
Remove cbioportal and cosmic annotation threshholds in oncoprint controls

alisman:


dippindots:
@alisman I remember I reviewed and merged [this pr](https://github.com/cBioPortal/cbioportal/pull/9220) which is related 
to cbioportal and cosmic control on February, are we sure we should remove these? I think it's still a feature we need, right?

PR #4152, opened 2022-01-27
New oncoprint property to configure tracks that are shown by default (frontend)

dianab0:
Added a new portal property named oncoprint.clinical_tracks.show_by_default that points to a JSON file (on the classpath) that configures tracks that should be visible by default. 

Needs backend functionality of https://github.com/cBioPortal/cbioportal/pull/9709

Example configuration file `oncoprint.clinical_tracks.show_by_default=classpath:/oncoprint-default-tracks.json`:
The configuration JSON contains an array of objects with stable ids of clinical variables, and optionally their (nested) sort order and/or gap on option. Example:

```
[
  {
    "stableId": "CANCER_TYPE",
     "sortOrder": "ASC",
     "gapOn": true
  },
  {
    "stableId": "MUTATION_COUNT",
    "sortOrder": "DESC",
    "gapOn": false
  },
  {
    "stableId": "AGE",
    "sortOrder": null,
    "gapOn": null
  }
]
```

When the backend exposes this new property, the cbioportal-frontend should render the tracks when initializing. For example, on the configuration json above:

Oncoprint will show the CANCER_TYPE, MUTATION_COUNT and AGE tracks when first showing, and will sort the samples on oncoprint first on CANCER_TYPE (top track) in ascending order and subsequently on MUTATION_COUNT in descending order (middle track). The AGE track will be shown as the bottom track (not sorted).

Oncoprint will ignore any clinical variable that is configured in the configuration JSON that is not available in the study that is analyzed at that moment in Results View.

### Update
- Initialization of clinical tracks has been moved to [ResultsViewOncoprint](https://github.com/cBioPortal/cbioportal-frontend/pull/4152/files#diff-74a4f3345d8f0926e6fe1c19df8dcfc2743ced608619c4d13700d86ade34693fR618)   
- When clinical tracks are initialized, added or deleted, the url query param `clinicallist` is updated with an url encoded json configuration object (as in example shown above).
- When gaps are added or removed from clinical tracks or when the tracks are sorted or 'unsorted', the url query param is updated
- Added e2e tests for initialization and configuration of clinical tracks (incl url query param changes) to [oncoprint.screenshot.spec.js](https://github.com/cBioPortal/cbioportal-frontend/pull/4152/files#diff-2e7d5c747b002ab15afa7e302402b5337b7b4d6d7e667aca985fd2fe1a48d3acR78).

PR #4230, opened 2022-04-13
Oncoprint config user session

BasLee:
# Store Oncoprint track config in user session

_Uses `new-oncoplot-property` to show changes: this branch will be removed when  [PR #4152](https://github.com/cBioPortal/cbioportal-frontend/pull/4152) has been merged._

- When a user opens an oncoprint, clinical tracks will be restored from user session configuration.
- When user makes changed to clinical tracks, the 'save-to-session'-button becomes active.
- When a user clicks the oncoprint save-to-session-button, clinical track configuration of the current oncoprint will be stored to the user session.

Needs backend functionality of https://github.com/cBioPortal/cbioportal/pull/9450

Screenshot from 2022-04-21 11-37-38

### Changes
- Add save-to-session-button 'Save Tracks' to Tracks menu
- Add `PageUserSession` to `ResultsViewPageStore` to manage changes in the user session. PageUserSession bundles all user session state management for the ResultViewPageStore, and prevents adding any additional complexity to an already quite complex store. From an architectural point of view this differs from the `StudyViewPageStore` which handles all user session management itself. However the PageUserSession has been kept as generic as possible: we can use the same mechanism for the StudyViewPageStore in a followup PR.

- Add e2e-tests to `oncoprint.screenshot.spec.js` Oncoprint can now be configured in three ways: using the default settings, user session and url query params. When should we use which method?
```
- if no session service:
  -> use url; or init with default
- if session service enabled:
  -> load page:
     - if url is default*:
       -> initialize with session
     - if url is non-default:
       -> initialize with url (disregard user session)
       -> enable save-session button
  -> change clinicallist:
     -> save to url
     -> enable save-session button
  -> change id:
     -> fetch session at id
        - if session is none:
          -> init with default*
        - if session is non-default:
          -> save to url
          -> initialize with clinicallist
  -> click session-save button:
     -> save clinicallist to session
     -> disable save-session button

*default: as configured in oncoprint_clinical_tracks_show_by_default
```


BasLee:
Updates after community call 2022-04-20.
To clarify the function of the button the following changes were made:
- Move the button to the 'add tracks' menu
- Rename the 'add tracks' menu to 'tracks'
- Add text 'save tracks' to button and change cloud to pin icon

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -261,9 +261,15 @@ export default class ResultsViewOncoprint extends React.Component<
...
-            this.urlWrapper.oncoprintSelectedClinicalTracks
-        );
+        const clinicallist = this.props.store.pageUserSession.userSettings
+            ?.clinicallist;
+        if (clinicallist) {
+            const map = {} as ClinicalTrackConfigMap;
+            clinicallist.forEach(cl => (map[cl.stableId] = cl));
+            return map;
+        }
+        const clinicalTracks =
```
alisman, line 271: why are we still looking at urlWrapper here if we're using session as source of truth

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -618,20 +624,48 @@ export default class ResultsViewOncoprint extends React.Component<
...
+        } else if (!urlConfig && sessionConfig) {
+            // do not update deprecated clinicallist url param
+        } else if (urlConfig) {
+            this.setSessionClinicalTracks(
+                this.urlWrapper.oncoprintSelectedClinicalTracks!
+            );
         }
     }
 
     private initializeClinicalTracksFromServerConfig(): void {
```
alisman, line 651: put a comment explaining that this applies the default set in instance configuration for clinical tracks

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -87,6 +87,7 @@ export interface IOncoprintControlsState
     onlyShowClinicalLegendForAlteredCases?: boolean;
     showOqlInLabels?: boolean;
     showMinimap: boolean;
+    isSaveButtonEnabled: boolean;
```
alisman, line 90: is this too general a name?

src/shared/userSession/PageUserSession.ts
```diff
@@ -0,0 +1,148 @@
...
+        return this._userSettings;
+    }
+
+    @observable
+    public set userSettings(userSettings: T | undefined) {
+        this._userSettings = userSettings;
+    }
+
+    @computed
+    public get canSaveSession() {
```
alisman, line 80: userSessionEnabled

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -317,67 +325,37 @@ export default class ResultsViewURLWrapper
...
-     * @param clinicalTracks
-     */
-    public convertClinicalTracksToUrlParam(
-        clinicalTracks: ClinicalTrackConfig[]
-    ) {
-        if (clinicalTracks.length) {
-            return { clinicallist: JSON.stringify(clinicalTracks) };
-        } else {
-            return { clinicallist: ALL_TRACKS_DELETED };
+    private extractUserSettings(query: { [key: string]: any }) {
```
alisman, line 328: could we do this outside of urlwrapper?

alisman: so that urlWrapper is not aware at all of page settings information

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -566,30 +583,65 @@ export class ResultsViewPageStore
...
+                () => {
+                    this.driverAnnotationSettings = buildDriverAnnotationSettings(
+                        () => store.didHotspotFailInOncoprint
+                    );
+                },
+                { fireImmediately: true }
+            )
+        );
+
+        this.reactionDisposers.push(
```
alisman, line 609: check to see if you actually get a whole new store every time you requery (change study ids).  if that's the

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -566,30 +583,65 @@ export class ResultsViewPageStore
...
+                () => {
+                    this.pageUserSession.id = {
+                        page: PageType.RESULTS_VIEW,
+                        origin: this.cancerStudyIds,
+                    };
+                }
+            )
+        );
+
+        this.reactionDisposers.push(
```
alisman, line 621: if we aren't tracking clinical tracks in url, why do we need this?

alisman:
1. Is it ok that we lose clinical track url state for logged out users because we no longer update url
2. Same question for groups. Why are groups used in results view page

alisman:
1. Is it ok that we lose clinical track url state for logged out users because we no longer update url
2. Same question for groups. Why are groups used in results view page

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
BasLee: Fixed in [240ae85](https://github.com/cBioPortal/cbioportal-frontend/compare/b0bd2c653119b4363c45d9c7f7acd584ea0d6436..240ae85eba446230bffd086012e309356e71f413)

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
BasLee: Fixed in [9065f7e](https://github.com/cBioPortal/cbioportal-frontend/compare/240ae85eba446230bffd086012e309356e71f413..9065f7ea67dda257aad80c7a2b09cc88272492d9)

src/shared/components/oncoprint/controls/OncoprintControls.tsx
BasLee: Fixed in [9065f7e](https://github.com/cBioPortal/cbioportal-frontend/compare/240ae85eba446230bffd086012e309356e71f413..9065f7ea67dda257aad80c7a2b09cc88272492d9)

src/shared/userSession/PageUserSession.ts
BasLee: Fixed in [9065f7e](https://github.com/cBioPortal/cbioportal-frontend/compare/240ae85eba446230bffd086012e309356e71f413..9065f7ea67dda257aad80c7a2b09cc88272492d9)

src/pages/resultsView/ResultsViewURLWrapper.ts
BasLee: @alisman: After looking in the code, I would prefer to keep it in the url wrapper. 
These user settings are stored and retrieved from the url, just like the legacy `clinicallist` query param. See also `oncoprintSelectedClinicalTracks`. It is all url related functionality, so moving it to the `ResultViewPageStore` or `ResultViewOncoprint` doesn't feel logical to me (and whould further bloat those two already quite large files).

src/pages/resultsView/ResultsViewPageStore.ts
BasLee: Solved: Does not re-instantiate resultsViewPageStore when we update cancer studies, so this is actually needed

src/pages/resultsView/ResultsViewPageStore.ts
BasLee: Fixed in [d44a497](https://github.com/cBioPortal/cbioportal-frontend/compare/9065f7ea67dda257aad80c7a2b09cc88272492d9..d44a497fd7eb858dfe038ae4f6050daad90b67d9)

BasLee:
> 1. Is it ok that we lose clinical track url state for logged out users because we no longer update url
> 2. Same question for groups. Why are groups used in results view page

@alisman: 
1. this is ok, as discussed in [slack](https://cbioportal.slack.com/archives/C359NNB41/p1660921939884819)
2. groups were already used in results view page, not an issue of this PR

PR #4234, opened 2022-04-21
Add p and q values to clinical plots when comparing groups

BasLee:
When comparing groups, the p and q values are available in the clinical table but these values were not shown in the plots. P and q values are already shown in survival, methylation and phosphosite plots. Now the p and q values are also shown in the bar charts of the clinical tab.


pvannierop (changes requested):
Hi Bas, there are some code format changes that you should revert I guess.

BasLee:
> there are some code format changes that you should revert I guess.
solved in f1a17b10d4418211a5632b355bd2b72cac81e75d

src/shared/lib/FormatUtils.tsx
```diff
@@ -103,20 +103,22 @@ export function toConditionalPrecisionWithMinimum(
...
             </span>
         );
     } else {
-        return toConditionalPrecision(
+        let transformed = toConditionalPrecision(
             positiveNumber,
             precision,
             precisionThreshold
         );
+        return <>{prependEqualSign ? ' = ' : '' }{transformed}</>;
```
alisman, line 121: seems like the equal sign belongs outside this function.  what is it for?

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -917,3 +929,39 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+                <div className="p-value-label">{pFormatted}</div>
+                <div className="q-value-label">{qFormatted}</div>
+            </g>
+        </foreignObject>
+    );
+
+    function formatLabel(name: string, value: number | null) {
+        if (value != null) {
+            return <>
+                {name} {toConditionalPrecisionWithMinimum(value, 3, 0.01, -10, true)}
```
alisman, line 961: see comment below.  why wouldn't you put equal sign logic here?

alisman:
Couple functional problems here:
- [ ] The chart download functionality is broken by this PR, probably as result of foreignobject?  
- [ ] The state of the p & q setting doesn't survive change of group visibility (in the way that e.g. Show NA does)

src/shared/lib/FormatUtils.tsx
BasLee: `toConditionalPrecisionWithMinimum` sometimes returns jsx, which makes checking if it starts with an `<` a bit dirty.
For example:
- when `<10^-10` no equal sign has to be added in `p < 10^-10`
- when `1.23e-4` you want to add an equal sign in `p = 1.23^4`
So I added the equal sign functionality to `toConditionalPrecisionWithMinimum`, to avoid having to parse the jsx

But maybe we can also simplify it, by rendering in the plot `p: < 10^-10` and `p: 1.23^4`? In that case no equal sign has to be added at all.

src/shared/components/plots/MultipleCategoryBarPlot.tsx
BasLee: See https://github.com/cBioPortal/cbioportal-frontend/pull/4234#discussion_r871079157

BasLee:
>  The chart download functionality is broken by this PR, probably as result of foreignobject?
Solved in [49328fa](https://github.com/cBioPortal/cbioportal-frontend/pull/4234/commits/49328fae81db0663e61bcc73daf13a8923bf0309)

BasLee:
> The state of the p & q setting doesn't survive change of group visibility (in the way that e.g. Show NA does)

The state of Show NA doesn't survive change of group visibility either, the default is different (true instead of false).
(I toggled the groups by clicking group labels at the top)

src/shared/lib/FormatUtils.tsx
alisman: i see.  no i think we should keep equal sign.  guess there's no good solution here so this is fine.

BasLee:
Run prettier and squashed to 1 commit in [f3dfa85](https://github.com/cBioPortal/cbioportal-frontend/pull/4234/commits/f3dfa855962a7244d11c184e658d0647707feb0e)

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -917,3 +929,47 @@ export default class MultipleCategoryBarPlot extends React.Component<
         );
     }
 }
+
+type QPLabelProps = {
```
dippindots, line 933: Could we change this to `PQLabelProps`, just wanted to make this same as the name you give for `PQValueLabel`

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -917,3 +929,47 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+    x: number;
+    y: number;
+    pValue: number | null;
+    qValue: number | null;
+};
+
+const PQValueLabel = (props: QPLabelProps) => {
```
dippindots, line 940: I think it's better to give it a type like this, although it won't impact the performance, but it gives us a little bit more control on return type (e.g. this will force you to return a component instead of a string).
```
const PQValueLabel: React.FunctionComponent<PQLabelProps> = props => {
    // COMPONENT CONTENT
}

dippindots (commented):
@BasLee Thanks for making this enhancement! Overall it looks good to me, I just have some minor comments.

src/shared/components/plots/MultipleCategoryBarPlot.tsx
BasLee: Solved in [846942a](https://github.com/cBioPortal/cbioportal-frontend/pull/4234/commits/846942aa16d04fc787ad358d5004dfd3e43c7828)

src/shared/components/plots/MultipleCategoryBarPlot.tsx
BasLee: Solved in [846942a](https://github.com/cBioPortal/cbioportal-frontend/pull/4234/commits/846942aa16d04fc787ad358d5004dfd3e43c7828)

PR #4246, opened 2022-05-09
Add feature instruction component and apply to patient view page

alisman:
image


src/pages/patientView/PatientViewPageTabs.tsx
```diff
@@ -175,7 +177,10 @@ export function tabs(
                     pageComponent.patientViewPageStore.cnaSegments.result
                         .length > 0) &&
                 sampleManager && (
-                    <div>
+                    <FeatureInstruction
+                        content={'Click location for detail'}
```
onursumer, line 181: Is it clear what `location` means? Should we say `location on a feature track`? `Feature track` may also be confusing though. Maybe location is okay as is.

onursumer (approved):
LGTM

PR #4259, opened 2022-05-18
Implement left truncation adjustment to survival analysis

dippindots:


src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,66 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+                atRisk.push(atRiskAccumulator);
+                atRiskAccumulator -= 1;
+
+                return atRisk;
+            },
+            atRisk
+        );
+    } else {
+        // Calculate the number of patients at risk for data without an entry time
+        for (let i = sortedPatientSurvivals.length; i > 0; i--) {
```
alisman, line 135: just do 
```
// we need this in descending order, so reverse
return sortedPatientSurvivals.map((el,i)=>i).reverse()
```

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,66 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+        // Calculate number of patients at risk for data with entry time (left truncation)
+        let entryIndex = 0;
+        let atRiskAccumulator = 0;
+        _.reduce(
+            sortedPatientSurvivals,
+            (atRisk, patientSurvival) => {
+                const exitTime = patientSurvival.months;
+                // This step is to count the total number of subjects remaining in the study at this specific exit time
+                // fast forward through entries until we get past exit time
+                while (
```
alisman, line 116: see how the use of the while and multiple incrementors confuses things.  i think you are just trying to get the count of entryTimes members which are less than exitTime and add that count to the accuulator  
```
atRiskAccumulator += entryTimes.filter(t=>t>extitTime).length
```

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,66 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+                    entryIndex < entryTimes.length &&
+                    entryTimes[entryIndex] < exitTime
+                ) {
+                    // add 1 for each entry
+                    atRiskAccumulator += 1;
+                    entryIndex += 1;
+                }
+                // Now we found the number of risk this specific exit time, push to atRisk as we move forward to the next exit time,
+                // and subtract one from atRiskAccumulator because we need to subject this subject from count while we move to the next subject
```
alisman, line 125: subtract this subject?

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,66 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+    const atRisk: number[] = [];
+
+    const entryTimes = _.sortBy(sortedPatientSurvivals.map(s => s.entryMonths));
+
+    // When we see one entry time is not zero, we consider the data is left truncated
+    if (_.some(entryTimes, entryTime => entryTime !== 0)) {
+        // Calculate number of patients at risk for data with entry time (left truncation)
+        let entryIndex = 0;
+        let atRiskAccumulator = 0;
+        _.reduce(
```
alisman, line 110: the point of using reduce is that it returns the accumulator when it's finished.  so you always want
```const result = _.reduce(...)

File: src/shared/lib/comparison/ComparisonStore.ts [hunk 5]
```diff
@@ -1818,13 +1819,47 @@ export default abstract class ComparisonStore
...
 
+    readonly survivalSequencedMonths = remoteData<
+        { [uniquePatientKey: string]: number } | undefined
+    >({
+        await: () => [this.studies],
+        invoke: async () => {
+            const studyIds = this.studies.result!.map(s => s.studyId);
+            if (studyIds.length === 1 && studyIds[0] === 'nsclc_genie_bpc') {
+                const allTimelineData = _.flatten(
+                    await Promise.all(
```
alisman, line 1830: i would not nest await like this.     first assign the await to something and then flatten that.  it's too hard to see the asynchrouns thing

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1818,13 +1819,47 @@ export default abstract class ComparisonStore
...
+                        studyIds.map(studyId =>
+                            internalClient.getAllClinicalEventsInStudyUsingGET({
+                                studyId,
+                            })
+                        )
+                    )
+                );
+                const sequencingData: ClinicalEvent[] = allTimelineData.filter(
```
alisman, line 1838: you can flatten it right here.

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1818,13 +1819,47 @@ export default abstract class ComparisonStore
...
+                    d => d.eventType === 'Sequencing'
+                );
+                return sequencingData.reduce(
```
alisman, line 1841: it's not great that you are fetching the clinican events and then throwing away all the data except that which you need.  this makes it impossible to resuse that clincail events for other purposes.   instead, have one promise that just load the event and then do the processing of the data in your business logic.

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: Updated.

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: This is true, thanks for the suggestion, I modified this based on the suggestion. One thing to consider is that using index can ignore repeating search (O(n) time complexity), and if we filter every time, it will be O(n^2). So we can keep this in mind, and if we experiencing performance issue later, I can change to use index approach if needed.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,66 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+            sortedPatientSurvivals,
+            (atRisk, patientSurvival) => {
+                const exitTime = patientSurvival.months;
+                // This step is to count the total number of subjects remaining in the study at this specific exit time
+                // fast forward through entries until we get past exit time
+                while (
+                    entryIndex < entryTimes.length &&
+                    entryTimes[entryIndex] < exitTime
```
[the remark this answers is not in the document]
dippindots: removed entryIndex

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,50 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+    // Then sort by status in desc order (which means if status are '1's, then go to the front, '0's are right after '1's in the same time stamp)
+    sortedPatientSurvivals: Pick<PatientSurvival, 'entryMonths' | 'months'>[]
+): number[] {
+    // returns an array in the same order as patientSurvivals
+    const entryTimes = _.sortBy(sortedPatientSurvivals.map(s => s.entryMonths));
+
+    // When we see one entry time is not zero, we consider the data is left truncated
+    if (_.some(entryTimes, entryTime => entryTime !== 0)) {
+        // Calculate number of patients at risk for data with entry time (left truncation)
+        return _.reduce(
```
alisman, line 106: i think this is actually a mapping operation because you are always pushing to the aggregator for every iteration.  in that case, the output array will always have the same number of elements as the input array.  reduce should be used only if there is a conditional or if the aggregator has a different type than the input (e.g. array -> key/value or summation).

alisman: apologies if i suggested before using reduce.  but the semantics of these methods really helps a reader understand what is going on.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +85,43 @@ export const survivalClinicalDataNullValueSet = new Set([
...
+    // Patient survivals need to be sorted by months and status
+    // First sort by month in asc order (smaller number to the front)
+    // Then sort by status in desc order (which means if status are '1's, then go to the front, '0's are right after '1's in the same time stamp)
+    sortedPatientSurvivals: Pick<PatientSurvival, 'entryMonths' | 'months'>[]
+): number[] {
+    // returns an array in the same order as patientSurvivals
+    const entryTimes = _.sortBy(sortedPatientSurvivals.map(s => s.entryMonths));
+
+    // When we see one entry time is not zero, we consider the data is left truncated
+    if (_.some(entryTimes, entryTime => entryTime !== 0)) {
```
alisman, line 104: _.some(sortedPatientSurvivals,p=p.entryMonths !==0)

src/pages/groupComparison/Survival.tsx
```diff
@@ -211,7 +211,11 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
                     survivalsByPrefixByAnalysisGroup,
                     survivalsByAnalysisGroup =>
                         _.mapValues(survivalsByAnalysisGroup, survivals =>
-                            survivals.sort((a, b) => a.months - b.months)
+                            _.orderBy(
+                                survivals,
+                                ['months', 'status'],
```
alisman, line 216: if at all possible to avoid using text to represent the property names since i don't think it's typesafe.  we'd have to look at signature of orderBy to see if it accepts callback instead of his shorthand

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -26,6 +27,10 @@ export function getPatientSurvivals(
                     const monthsClinicalData = clinicalData.find(
                         c => c.clinicalAttributeId === monthsAttributeId
                     );
+                    const entryMonths =
+                        (entryMonthsByUniquePatientKey &&
```
alisman, line 31: you can now do ```entryMonthsByUniquePatientKey?[uniquePatientKey] || 0```

src/pages/studyView/charts/survival/StudyViewSurvivalUtils.ts
```diff
@@ -31,7 +31,7 @@ export function makeSurvivalChartData(
     );
 
     const sortedGroupedSurvivals = _.mapValues(groupedSurvivals, survivals =>
-        survivals.sort((a, b) => a.months - b.months)
+        _.orderBy(survivals, ['months', 'status'], ['asc', 'desc'])
```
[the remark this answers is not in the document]
alisman: see about strings as property names.  i don't think that typescript is good enough to check

PR #4263, opened 2022-05-19
Fix request body compression

Luke-Sikina:
- Url params are now omitted from the url passed to the api client
- This was breaking some startsWith logic for requests that we compress
based on both url params and the uri
- To fix this, we now match against both the url string and the param
object


pvannierop (approved):
Thanks for solving this!

inodb (approved):
LGTM!

There are some broken localdb tests but it doesn't seem to be specific to this PR? Bit tricky to verify with junit test reports broken

PR #4269, opened 2022-05-20
Fix jest junit reporting and unit test circular reference

alisman:
There were a number of problems with unit test
1. We were not outputting junit reports for circleci to interpret
2. Running two jest jobs in one command seemed problematic
3. There was a circular dependency which seemed to be causing jest to crash/tests to fail


Luke-Sikina (approved):
Approved, pending passing tests

PR #4273, opened 2022-05-26
Update unit test yarn scripts and add grepping ability

alisman:
We now have two different unit testing jobs and dev scripts needed to reflect that.  README as well.


packages/cbioportal-frontend-commons/package.json
```diff
@@ -27,7 +27,7 @@
     "tcm": "tcm -p src/**/*.module.scss",
     "tcm:watch": "yarn run tcm --watch",
     "prepare": "yarn run build",
-    "test": "cross-env jest --env=jsdom --runInBand --ci --reporters=default --reporters=jest-junit",
+    "test": "cross-env jest $GREP --env=jsdom --runInBand --ci --reporters=default --reporters=jest-junit --passWithNoTests",
```
alisman, line 30: these pass with no tests needed to be added in order to support GREP because there will be no matches there.

packages/react-variant-view/package.json
```diff
@@ -27,7 +27,7 @@
     "tcm": "tcm -p src/**/*.module.scss",
     "tcm:watch": "yarn run tcm --watch",
     "prepare": "yarn run build",
-    "test": "cross-env CI=1 jest --env=jsdom --runInBand --passWithNoTests",
+    "test": "cross-env CI=1 jest $GREP --env=jsdom --runInBand --passWithNoTests",
```
onursumer, line 30: Why don't we use the same script here? Is it because this package has no test cases? 
```suggestion
    "test": "cross-env jest $GREP --env=jsdom --runInBand --ci --reporters=default --reporters=jest-junit --passWithNoTests",
```

package.json
```diff
@@ -48,9 +48,10 @@
     "serveDistLocalDb": "./end-to-end-test/local/runtime-config/serve_dist.sh",
     "testTime": "start_time=`date +%s` && karma start karma.conf.js && echo run time is $(expr `date +%s` - $start_time) s",
     "test": "yarn run testMain",
-    "testMain": "jest TableHeaderControls.spec.tsx --env=jsdom --runInBand --ci --reporters=default --reporters=jest-junit",
-    "testModulesCI": "JEST_JUNIT_UNIQUE_OUTPUT_NAME=true JEST_JUNIT_OUTPUT_DIR=/tmp/junit/module/ JEST_JUNIT_OUTPUT_NAME=unit-test-module yarn run testModules",
-    "testModules": "lerna run test --ignore=cbioportal-frontend --stream",
+    "check-env": "echo ",
+    "testMain": "jest $GREP --env=jsdom --runInBand --ci --reporters=default --reporters=jest-junit",
+    "testPackagesCI": "JEST_JUNIT_UNIQUE_OUTPUT_NAME=true JEST_JUNIT_OUTPUT_DIR=/tmp/junit/module/ JEST_JUNIT_OUTPUT_NAME=unit-test-module yarn run testPackages",
+    "testPackages": "lerna run test --ignore=cbioportal-frontend --stream",
```
onursumer, line 54: Hmm. We should probably rename `buildModules` and `publishModules` as well for consistency.  In that case we will also need to update https://github.com/cBioPortal/cbioportal-publisher/blob/master/.github/workflows/cbioportal-frontend-publish.yml#L27-L28.

packages/react-variant-view/package.json
alisman: oh, yeah, i didn't bother putting in packages that didn't have tests.  but i guess this one COUNT have tests.  so that's a good point.

onursumer (approved):
LGTM! This PR now successfully catches the failing unit tests :tada:

PR #4280, opened 2022-05-31
Don't fetch ndex status unless Pathways tab is active

alisman:
We need to wait until user navigates to the pathways tab before we call ndex service

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -33,7 +33,6 @@ import Helmet from 'react-helmet';
 import {
     parseConfigDisabledTabs,
     ResultsViewTab,
-    ResultsViewPathwaysSubTab,
 } from './ResultsViewPageHelpers';
 import {
     buildResultsViewPageTitle,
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -410,123 +408,10 @@ export default class ResultsViewPage extends React.Component<
...
-                                                                WikiPathways
-                                                            </a>
-                                                        </li>
-                                                    </ul>
-                                                </div>
-                                            }
-                                            hide={
-                                                !getServerConfig().show_ndex ||
-                                                !this.resultsViewPageStore
-                                                    .remoteNdexUrl.isComplete ||
```
alisman, line 518: problem was that this actually gets executed even though the component which is consuming it won't render.  i.e. if you look at code around the tab, you will see that there is no condition which prevents this from getting executed.  you could put a condition activeTab=pathways, and that would also solve this problem, but really, the fetch for ndex status should be nested inside the pathways component so that it the code won't be reached unless the pathways tab actually renders.

inodb: Gotcha -thank you! That makes sense

PR #4283, opened 2022-06-03
Impl. size-dependent request body compression

pvannierop:
# Problem1
When loading studies with large number of samples there are a large number of endpoints that try to send extended lists of sample identifiers to the backend. In uncompressed for this can take minutes for ~100k samples. This bottlenet can be removed by compressing the request body before sending. At present this is implemented for specific endpoints. I found out that certain (new?) endpoints are not listed for compression and therefore lock the frontend in an inresponsive state. Therefore, the mechanism of selecting specific endpoints for compression does not effectively solve the problem of large request bodies.

# Problem 2
The current request body compression is only applied to endpoints of the public API. Certain endpoints of the internal API are very slow.

# Solution
I propose to evaluate any body of POST requests for size and compress the body when the number of characters in the JSON string exceeds a certain threshold (now arbitrarily set at `500` characters). The compression mechanism covers both public and internal API's.


src/config/config.ts
```diff
@@ -258,14 +233,15 @@ function compressRequestBodies(
...
-            method === 'POST' &&
-            urlsToCompress.filter(pair =>
-                pairMatchesPath(pair, url, queryParameters)
-            ).length > 0
-        ) {
-            headers['Content-Encoding'] = 'gzip';
-            body = pako.gzip(JSON.stringify(body)).buffer;
+        if (method === 'POST') {
+            var bodyString = JSON.stringify(body);
+            if (bodyString.length > REQ_BODY_SIZE_CHAR_LIMIT) {
```
Luke-Sikina, line 238: How expensive would it be to do `new Blob([bodyString]).size`? Using char count to estimate size is a bit unreliable.

alisman: @Luke-Sikina it's just a heuristic right?  and why is it innacurate?

pvannierop: @alisman I also reasoned that a precise estimate is not needed here. We just need the least resource-intensive estimate of request size.

alisman (approved):

PR #4286, opened 2022-06-07
Improve local db screenshot

dippindots:
This will remove the background image and make sure this test only tests what it concerns.
image


inodb (approved):
Group approval 🙌

PR #4287, opened 2022-06-07
Optimize namespace column configuration for mutation tables

pvannierop:
# Problem
For namespace columns there is a data-dependent evaluation of the data type per column (_numeric_ or _categorical_). The current implementation was very slow for ~100k samples/mutations.

# Solution
The algorithm was updated to handle large number of mutations.

### DIFF
--- a/src/shared/components/mutationMapper/MutationMapperUtils.ts
+++ b/src/shared/components/mutationMapper/MutationMapperUtils.ts
@@ -41,19 +41,26 @@ export function buildNamespaceColumnConfig(
     if (!mutations) {
         return {};
     }
-    const namespaceConfig: NamespaceColumnConfig = {};
-    const nameSpaces = _.flatMap(mutations, m => _.keys(m.namespaceColumns));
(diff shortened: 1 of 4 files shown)


src/shared/components/mutationMapper/MutationMapperUtils.ts
```diff
@@ -41,19 +41,26 @@ export function buildNamespaceColumnConfig(
...
+    mutations.forEach(m => {
+        _.forIn(m.namespaceColumns, (columns, namespace) => {
+            _.forIn(columns, (value, columnName) => {
+                if (columnTypes[namespace] === undefined) {
+                    columnTypes[namespace] = {};
+                }
+                if (columnTypes[namespace][columnName] === undefined) {
+                    columnTypes[namespace][columnName] = 'number';
+                }
+                if (
```
alisman, line 54: could probably use a comment about how this would come about

alisman (approved):

PR #4290, opened 2022-06-10
OncoKB Updates

zhx828:
- Update oncokb-styles to use latest OncoKB logo
- Remove OncoKB images from portal repo and use the ones in oncokb-styles
- Add trademark to most of OncoKB text
- Remove OncoKB Predicted Oncogenic concept


onursumer (approved):
🙌

leexgh (approved):
LGTM! Would it be good to add trademark in download file as well?

PR #4294, opened 2022-06-10
Add numeral into the portal commons package dependency

zhx828:
The package is using the dependency from the root which will run into issue when other projects, like GenomeNexus, importing commons separately.


packages/cbioportal-frontend-commons/package.json
```diff
@@ -37,13 +37,15 @@
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
+    "@types/numeral": "0.0.22",
```
onursumer, line 40: let's remove this one. this is not needed by external projects.

onursumer (changes requested):
Good catch!

Let's remove the @types dependency before merging though. It's not needed by external projects.

packages/cbioportal-frontend-commons/package.json
zhx828: Done

PR #4295, opened 2022-06-10
Fix download controls localdb tests

onursumer:
Fix cBioPortal/cbioportal#9608

- make sure that generic assay tab names are always sorted alphabetically.
- replace `describe.omit` with `describe.skip` since `describe.omit` is not a function
- implement a missing `studyViewChartHoverHamburgerIcon` function

### DIFF
--- a/src/pages/resultsView/comparison/ComparisonTab.tsx
+++ b/src/pages/resultsView/comparison/ComparisonTab.tsx
@@ -35,7 +35,7 @@ import GenericAssayEnrichments from 'pages/groupComparison/GenericAssayEnrichmen
 import styles from 'pages/resultsView/comparison/styles.module.scss';
 import { getServerConfig } from 'config/config';
 import { AlterationFilterMenuSection } from 'pages/groupComparison/GroupComparisonUtils';
-import { deriveDisplayTextFromGenericAssayType } from 'shared/lib/GenericAssayUtils/GenericAssayCommonUtils';
+import { getSortedGenericAssayTabSpecs } from 'shared/lib/GenericAssayUtils/GenericAssayCommonUtils';
 
 export interface IComparisonTabProps {
     urlWrapper: ResultsViewURLWrapper;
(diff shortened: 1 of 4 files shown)


src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
```diff
@@ -358,3 +358,21 @@ export function deriveDisplayTextFromGenericAssayType(
     }
     return derivedDisplayText;
 }
+
+export function getSortedGenericAssayTabInfo(
+    genericAssayEnrichmentProfilesGroupedByGenericAssayType: {
+        [key: string]: MolecularProfile[];
+    } = {}
+): { genericAssayType: string; linkText: string }[] {
+    return _.sortBy<{ genericAssayType: string; linkText: string }>(
```
alisman, line 367: nested lodash is super hard to read. can we use _.chain().keys().map().value() ?

src/pages/resultsView/comparison/ComparisonTab.tsx
```diff
@@ -264,19 +264,17 @@ export default class ComparisonTab extends React.Component<
...
+                        ).map(helper => {
                             return (
                                 <MSKTab
                                     id={`${
                                         ResultsViewComparisonSubTab.GENERIC_ASSAY_PREFIX
-                                    }_${genericAssayType.toLowerCase()}`}
-                                    linkText={deriveDisplayTextFromGenericAssayType(
-                                        genericAssayType
-                                    )}
+                                    }_${helper.genericAssayType.toLowerCase()}`}
```
alisman, line 276: the name "helper" doesn't really help me understand what it is.

onursumer: good point, I'll try to find a better name.

onursumer: renamed as `genericAssayTabSpecs`

src/shared/lib/GenericAssayUtils/GenericAssayCommonUtils.ts
onursumer: separated this into multiple operations to make it easier to read

PR #4297, opened 2022-06-14
Fix checkbox controls above clinical attribute plots

BasLee:
Heatmap and table visualisations in the clinical tab of group comparison show p+q checkboxes that do nothing: heatmap does not support p+q values, and its is irrelevant in the table display. 

This PR:
- hides p+q checkbox when heatmap or table;
- cleans up bloated if statements around clinical plot checkboxes.


src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -246,10 +246,41 @@ export default class ClinicalData extends React.Component<
         }
     );
 
-    @computed get showLogScaleControls() {
+    @computed get isNumericalPlot() {
         return isNumerical(this.highlightedRow!.clinicalAttribute.datatype);
     }
 
+    @computed get showLogScaleControls() {
```
alisman, line 253: a little picky, but probably a better way to do this would be to have a single visualizationMode computed that returned a type.   That way you consolidate all these conditions into one place.   And if you ever need to add a mode, it's simple.

BasLee: What type or mode would visualisationMode return? table, heatmap, etc?
That would mean that we move the if-checks back into the jsx, because for each checkbox a different set of modes are relevant.

alisman (approved):

PR #4298, opened 2022-06-14
Set max for number of categories in Plots tab

pvannierop:
# Problem
When there there are a large number of samples and there is a clinical variable that is unique to each sample (e.g. sample identifier), the plots tab will lockup when rendering a Categorical plot. The cause of this is that there are too many elements that need to be rendered.

# Solution
In this PR we limit the number of categories for a categorical plot to _150_. This level was chosen based on the ability to render this number of elements. A warning message is shown to inform the user to select another variable:

image

### DIFF
--- a/src/pages/resultsView/plots/PlotsTab.tsx
+++ b/src/pages/resultsView/plots/PlotsTab.tsx
@@ -4838,7 +4838,7 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                             className={'alert alert-info'}
                         >
                             There are too many unique categories in the data
-                            selected for horizontal or vertical axis. The
+                            selected for the horizontal or vertical axis. The
                             current threshold is set at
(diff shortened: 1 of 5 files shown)


src/pages/resultsView/plots/PlotsTab.tsx
```diff
@@ -4816,6 +4831,21 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
...
+                    return (
+                        <div
+                            data-test="PlotsTabNoDataDiv"
+                            className={'alert alert-info'}
+                        >
+                            There are too many unique categories in the data
+                            selected for horizontal or vertical axis. The
+                            current threshold is set at
+                            {DISCRETE_CATEGORY_LIMIT} categories to prevent the
+                            browser from freezing. Please update you selection
```
alisman, line 4844: type: your selection

alisman (approved):

PR #4300, opened 2022-06-15
Add missing info for setting up dev env

inodb:


README.md
```diff
@@ -37,6 +37,12 @@ To build DLLs in common-dist folder (must be done prior to start of dev server)
 yarn run buildDLL:dev
 ```
 
+To build all packages the main project depends on (must be done prior to start of dev server):
+
+```
+yarn run buildModules
```
alisman, line 43: i don't actually think this is necessary.  it gets run as part of ```yarn run start```

onursumer: In less powerful systems `yarn run start` may fail to concurrently build all the modules under `packages`. So a combination of `yarn run buildModules` and `yarn run watch` is less error-prone.

PR #4301, opened 2022-06-16
fix fsevents error

nr23730:
Fix build error related to fsevents on macOS as discussed on Slack. 
@alisman


alisman:
Thanks [nr23730](https://github.com/nr23730)

PR #4303, opened 2022-06-16
Fix 3d hotspots svg font family

onursumer:
Fix cBioPortal/cbioportal#9620


onursumer:
@zhx828 apparently the new `@rollup/plugin-image` cannot handle malformed svg files. Also, looks like we have 2 copies of this image. I already fixed the other one in my previous PR (https://github.com/onursumer/cbioportal-frontend/blob/master/src/shared/components/annotation/images/3d-hotspots.svg?short_path=8b398df#L12), but missed this one.

PR #4304, opened 2022-06-16
react-mutation-mapper updates

zhx828:
- Do not use Unknown as default cancer type 
  - The unknown will be interpreted as cancer type in OncoKB which will return inaccurate annotation
  - When cancer type is not available, OncoKB will return all treatments for the mutation
- Show level associated cancer type instead of query cancer type in the OncoKBSummaryTable
- Create OncoKB LevelIcon and OncogenicIcon to reuse

### DIFF
--- a/packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
+++ b/packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
@@ -9,6 +9,7 @@ import { ICache } from '../../model/SimpleCache';
 import {
     annotationIconClassNames,
     calcHighestIndicatorLevel,
+    normalizeLevel,
 } from '../../util/OncoKbUtils';
 import OncoKbCardLevelsOfEvidenceDropdown from './OncoKbCardLevelsOfEvidenceDropdown';
 import OncoKBSuggestAnnotationLinkout from './OncoKBSuggestAnnotationLinkout';
@@ -19,6 +20,8 @@ import { ImplicationContent } from './oncokbCard/ImplicationContent';
 
 import tabsStyles from './tabs.module.scss';
 import mainStyles from './main.module.scss';
+import OncogenicIcon from './icon/OncogenicIcon';
+import LevelIcon from './icon/LevelIcon';
 
 const OncoKbMedicalDisclaimer = (
     <p className={mainStyles.disclaimer}>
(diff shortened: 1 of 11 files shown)


packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
```diff
@@ -76,13 +79,21 @@ const TabTitle: React.FunctionComponent<{
...
-            className={annotationIconClassNames(
-                props.type,
-                calcHighestIndicatorLevel(props.type, props.indicator),
-                props.indicator
-            )}
-        />
+        props.type === OncoKbCardDataType.BIOLOGICAL ? (
+            <OncogenicIcon
+                oncogenicity={props.indicator?.oncogenic || ''}
+                showDescription
```
onursumer, line 85: I know it is minor but a prop without a value always bothers me
```suggestion
                showDescription={true}
```

onursumer (approved):
Looks good!

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
zhx828: Is this the convention in portal? We don't specify true in oncokb anymore

onursumer: I don't think we have a strict rule about this, but I think the majority of the boolean props in portal codebase have the explicit `={true}`.

packages/react-mutation-mapper/src/component/oncokb/OncoKbCardBody.tsx
```diff
@@ -76,13 +79,21 @@ const TabTitle: React.FunctionComponent<{
...
     const title = DATA_TYPE_TO_TITLE[props.type];
     const icon = props.displayHighestLevelInTabTitle ? (
-        <i
-            className={annotationIconClassNames(
-                props.type,
-                calcHighestIndicatorLevel(props.type, props.indicator),
-                props.indicator
-            )}
-        />
+        props.type === OncoKbCardDataType.BIOLOGICAL ? (
```
alisman, line 82: in general lets avoid nested ternary.  just too difficult to read

alisman (approved):

alisman:
@zhx828 i want to wait till after 5.0.0 release to merge this.  can't take any more conflicts and we need to meet tuesday deadline.

Luke-Sikina:
Rebase, pass tests and let's merge this? @alisman @onursumer

PR #4305, opened 2022-06-21
Filter studies by reference genome

BasLee:
Allow filtering by reference genome when searching for studies on the landing page. The current search filter setup can easily be extended with additional study fields.
Implementation of RFC 63. _Work in progress._ 

- Add a search filter that allows filtering on one or more reference genomes (e.g. `reference-genome:hg19,hg38`)
- Add a graphical interface to 'toggle' the reference genomes that are available in the loaded studies:

image

## Changes
- Extended study search with a new filter syntax (i.e. `<prefix>:<value>`) in which value can be a comma separated list
- Create `SearchClause` and `Phrase` interfaces to simplify search query modifications
- Replace external autocomplete package with a minimal bootstrap `StudySearch` search box component
- Add dropdown form to StudySearch box that renders filters as defined in QueryParser
- Add `QueryParser` that contains a extendable list of search filter (see example below)
  - search prefix (e.g. `reference-genome`)
  - relevant study fields (e.g. `study.referenceGenome`)
  - rendering in drop down form (e.g. as a checkbox)

Example search filter:
```
{
  phrasePrefix: 'reference-genome',
  nodeFields: ['referenceGenome'],
  form: {
    input: FilterCheckbox,
    options: [...referenceGenomes],
    label: 'Reference genome',
  },
}
```

## Checks
- Unit tests:
  - Phrase.spec.tsx
  - SearchClause.spec.tsx
  - CheckboxFilterField.spec.tsx
  - QueryParser.spec.ts
  - textQueryUtils.spec.ts 
- End to end tests:
  - TODO
- Manual: 
  - open index page;
  - start typing in search box;
  - or open dropdown filter form


src/shared/components/query/QueryStore.ts
```diff
@@ -2275,6 +2294,13 @@ export class QueryStore {
     }
 }
 
+export enum Focus {
```
alisman, line 2297: this is too generic i think.  i would just keep safe and limit it to this feature

src/shared/components/query/QueryStore.ts
```diff
@@ -1430,6 +1441,14 @@ export class QueryStore {
             .filter(_.identity);
     }
 
+    @computed get referenceGenomes(): Set<string> {
+        const studies = Array.from(this.treeData.map_node_meta.keys());
+        const referenceGenomes = studies
+            .map(n => (n as CancerStudy).referenceGenome)
+            .filter(n => !!n);
+        return new Set(referenceGenomes);
```
alisman, line 1449: how come you need a set here?

src/shared/components/query/StudySearch.tsx
```diff
@@ -0,0 +1,105 @@
...
+export type AutosuggestStudySearchProps = {
+    parser: QueryParser;
+    query: SearchClause[];
+    onSearch: (query: SearchClause[]) => void;
+};
+
+export const StudySearch: FunctionComponent<AutosuggestStudySearchProps> = function(
+    props
+) {
+    const [isMenuOpen, setMenuOpen] = useState(false);
```
alisman, line 26: for consistency, i would like to use mobx for handling all state, local or global.  look up useLocalObservable for the analog to useState hook.  it's already used a bit in our codebase.

src/shared/components/query/CancerStudySelector.tsx
```diff
@@ -237,74 +235,22 @@ export default class CancerStudySelector extends React.Component<
...
-                                            );
-                                        }}
-                                        onFocus={(value: string) => {
-                                            if (value.length === 0) {
-                                                setTimeout(() => {
-                                                    this.autosuggest.setState({
-                                                        open: true,
-                                                    });
-                                                }, 400);
+                                    {this.store.parser && (
```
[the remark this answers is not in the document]
BasLee: Renamed in [2bdadc7](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/2bdadc7240069b986b9da68e0539c1be485f3136)

src/shared/components/query/QueryStore.ts
```diff
@@ -153,6 +155,11 @@ export class QueryStore {
                 this.studiesHaveChangedSinceInitialization = true;
             }
         );
+
+        reaction(
```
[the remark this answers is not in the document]
BasLee: Solved in [c68a1c9](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/c68a1c95ab39056ff575f85b48627b15f151f203)

src/shared/components/query/QueryStore.ts
```diff
@@ -2144,7 +2163,7 @@ export class QueryStore {
 
     @action setSearchText(searchText: string) {
         this.clearSelectedCancerType();
-        this.searchText = searchText;
+        this.query = this.parser.parseSearchQuery(searchText);
```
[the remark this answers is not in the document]
BasLee: Renamed in [660162d](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/660162d865164a19980eb2a9c167b8473a913414)

src/shared/components/query/QueryStore.ts
```diff
@@ -254,7 +261,11 @@ export class QueryStore {
 
     @observable transposeDataMatrix = false;
 
-    @observable searchText: string = '';
+    @observable query: SearchClause[] = [];
```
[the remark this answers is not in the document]
BasLee: Solved in [0ba7e2f](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/0ba7e2fc96b5669ff9e7cc63f448d07bc0b17254)

src/shared/components/query/QueryStore.ts
```diff
@@ -1430,6 +1441,14 @@ export class QueryStore {
             .filter(_.identity);
     }
 
+    @computed get referenceGenomes(): Set<string> {
+        const studies = Array.from(this.treeData.map_node_meta.keys());
+        const referenceGenomes = studies
+            .map(n => (n as CancerStudy).referenceGenome)
```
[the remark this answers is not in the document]
BasLee: n is of type `CancerTreeNode`, which can be a `CancerTypeWithVisibility` and a `CancerStudy`

src/shared/components/query/StudySearch.tsx
BasLee: Check, fixed in [d8627b7](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/d8627b7399cf5c4e0a6e32decfb6d7d1e2a15ace)

src/shared/components/query/StudySearch.tsx
```diff
@@ -0,0 +1,105 @@
...
+
+    return (
+        <div className={`autosuggest dropdown ${isMenuOpen ? 'open' : ''}`}>
```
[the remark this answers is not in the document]
BasLee: Fixed in [14e49cc](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/14e49cc8f5f616dced35db6568835980494086ac)

src/shared/components/query/StudySearch.tsx
```diff
@@ -0,0 +1,114 @@
...
+                <StudySearchControls
+                    query={props.query}
+                    filterConfig={props.parser.searchFilters}
+                    onChange={handleQueryUpdate}
+                    parser={props.parser}
+                />
+            </div>
+        );
+
+        function handleQueryTyping(update: string) {
```
alisman, line 58: interesting that you put these functions down below the return.  i think it works because of hoisting, but it's a bit confusing to me to read.  in any case, these probably should be outside of the function component or you can use the useCallback hook to make sure that these functions are not unnecessarily redeclared every time a component renders.   this is sort of the curse of funtional components.  that we no longer have "methods"

alisman: i would be in favor of moving them outside the function and then passing them the props they need (or even props)

src/shared/components/query/StudySearch.tsx
```diff
@@ -0,0 +1,114 @@
...
+        </span>
+    );
+};
+
+const ClearSearchButton: FunctionComponent<{
+    onClick: () => void;
+    show: boolean;
+}> = props => {
+    return (
+        <span
```
alisman, line 97: you should make this a <div>

alisman: DIV

alisman:
@BasLee did some manual testing of the feature.  Seems good but for a couple things:

1. In the production version, the example query dropdown opens on focus and I think that's important behavior b/c it informs the user.  I know this was a specific request of the original product designers.  The behavior is a little nuanced.  I think it only opens on focus IF the field is empty.

2. Escape key closes the menu.  Probably not crucial, but nice.

3. Can you put a colon after the "Example queries" title?  "Example queries:"
4. Can you nix the line beneath the example queries when there are no controls below it. 
image

BasLee:
> @BasLee did some manual testing of the feature. Seems good but for a couple things:
> [..]
@alisman, @pvannierop : all these things were solved in [889494c](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/889494c510627e95bcd65f7143df6db6191a7997), [231e9ed](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/231e9ed83239096ca3443589c6f24a4bd502c5ad) and [471a8ab](https://github.com/cBioPortal/cbioportal-frontend/pull/4305/commits/471a8ab6a0f2358c9384a7867f9afe804d7e6e7d) I guess?

src/shared/components/query/StudySearch.tsx
BasLee: Solved in d796045b2

src/shared/components/query/StudySearch.tsx
BasLee: Solved in 9b91c7578

src/shared/components/query/StudySearch.tsx
```diff
@@ -0,0 +1,114 @@
...
+                updatedQuery = addClauses(update.toAdd, updatedQuery);
+            }
+            props.onSearch(updatedQuery);
+        }
+    }
+);
+
+const SearchMenuToggle: FunctionComponent<{ onClick: () => void }> = props => {
+    return (
+        <span className="input-group-btn">
```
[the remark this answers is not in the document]
BasLee: Solved in 9b91c7578

inodb (approved):
Thanks @BasLee !

I like the idea of adding more filtering of studies directly from the query page. The current user flow is to go to the datasets page and filter studies there. E..g. one can find studies there having data type x

Some comments:

- It looks like this change introduces an extra separator at the bottom with this change. Can we remove the bottom separator here if there are no filters available:
    
    Screen Shot 2022-09-08 at 9 52 31 AM

- Do we want to keep a text presentation of the filter (`reference-genome:`? The current search bar is very small and one can imagine all kinds of filters (e.g. "has timeline data", "has mRNA data"). It seems like a text representation of that can get kinda complex quickly. Maybe better to remove?
- It would be good to think about how we can combine these study filters + study view filter objects
- A nice to have: maybe a separate PR that adds an optional reference genome column to the data sets page?

Overall I like the functionality of being able to filter by reference genome and this is a nice addition to be able to do so. To scale this to more study filters it will need some design work. Would be good to improve the design of how the user flow between query / quick search / datasets page is meant to work at some point

tmazor:
Nice addition @BasLee !

I can't test the actual functionality in the public portal, but as @inodb suggested, the separator should not be present if there are no reference genome options.

In fact, I just agree with everything @inodb said :)

In terms of keeping 'reference-genome:' in the search bar - it's a really good point it takes up a lot of space that doesn't really exist. @inodb were you proposing to just show 'hg19' rather than 'reference-genome:hg19'? I worry that won't scale well - imagine a filter for 'has-mRNA:True' - just showing 'True' wouldn't be informative. I could go both ways at this point, but maybe it's ok for now, and we just need to think about a more general solution as we expand the feature beyond reference genome.

BasLee:
@inodb and @tmazor Thank you for reviewing!
I think we already addressed the horizontal bar issue: will look into it.
The search bar now shows a filter name and value, a 'gmail-like' syntax to allow filtering on specific fields, as described in the RFC and suggested by JJ and Sjoerd. The search bar might be a bit small for larger queries: as the number of filter options increases, we might want to give the search bar a more prominent place on the index page?

inodb:
Thanks so much @BasLee! I like the wider search bar!

One more bug I found:

- [ ] Something broken with example filtering. Production:
    Screen Shot 2022-09-14 at 10 34 10 AM
   This PR keeps showing same examples:
    Screen Shot 2022-09-14 at 10 34 03 AM

> @tmazor: In terms of keeping 'reference-genome:' in the search bar - it's a really good point it takes up a lot of space that doesn't really exist. @inodb were you proposing to just show 'hg19' rather than 'reference-genome:hg19'? I worry that won't scale well - imagine a filter for 'has-mRNA:True' - just showing 'True' wouldn't be informative. I could go both ways at this point, but maybe it's ok for now, and we just need to think about a more general solution as we expand the feature beyond reference genome.

My thought was to remove it entirely (so no `hg19` nor `reference-genome:` in the text field), only show the checkbox status in the dropdown, but I'm happy with the bigger search box as a solution for now. Thanks @BasLee ! I think the study text filter query language should eventually work in harmony with OQL

inodb:
@BasLee one more question: is it really necessary to be able to select both hg19 + hg38? If not it might be cleaner to just list in the examples:

```
reference-genome:hg19
reference-genome:hg38
```
then it would also filter nicely when type `reference-...`

BasLee:
> @BasLee one more question: is there a use case to be able to select both hg19 + hg38? If not it might be cleaner to just list in the examples:
> 
> ```
> reference-genome:hg19
> reference-genome:hg38
> ```
> 
> then it would also filter nicely when typing `reference-...`

I think you want to show all studies by default? So that means selecting both hg19 and hg38 (when present)

BasLee:
> One more bug I found:
@inodb: fixed in 9db3773

inodb:
Thanks @BasLee ! Looks like there's still an issue vs production for queries not part of example queries:

PR:
Screen Shot 2022-09-16 at 10 44 03 AM

Production:
Screen Shot 2022-09-16 at 10 44 49 AM

I think these should look identical

BasLee:
> Looks like there's still an issue vs production for queries not part of example queries [..]

@inodb a bit late, reading this just now... But yes, lets discuss what is should look like, with a follow up PR!

PR #4306, opened 2022-06-22
Display a detailed error message for invalid molecular profiles

onursumer:
Fix cBioPortal/cbioportal#9629

invalid_molecular_profile_ids

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -85,6 +85,7 @@ import {
     excludeFiltersForAttribute,
     FGA_PLOT_DOMAIN,
     FGA_VS_MUTATION_COUNT_KEY,
+    findInvalidMolecularProfileIds,
     geneFilterQueryFromOql,
     geneFilterQueryToOql,
     generateScatterPlotDownloadData,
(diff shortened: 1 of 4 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7090,13 +7091,27 @@ export class StudyViewPageStore
...
+                // here we are validating only the molecular profile ids,
+                // but ideally we should validate the entire filters object
+                const invalidMolecularProfiles = findInvalidMolecularProfileIds(
+                    this.filters,
+                    this.molecularProfiles.result
+                );
+
+                if (invalidMolecularProfiles.length > 0) {
+                    return Promise.reject({
+                        detailedErrorMessage: `Invalid molecular profile id(s): ${invalidMolecularProfiles.join(
```
onursumer, line 7110: Not sure if this is the best solution. `detailedErrorMessage` is a custom field. We may want to have a well defined `Error` object for consistency and always use the same fields everywere.

src/shared/components/progressIndicator/ProgressIndicator.tsx
```diff
@@ -129,13 +142,19 @@ export default class ProgressIndicator extends React.Component<
...
+            const detailedErrorMessages = getDetailedErrorMessages(this.items);
             return (
                 <div className={styles.container}>
                     <div className={styles['items-container']}>
                         {this.items.map(this.makeItem)}
                     </div>
                     {_.some(this.items, i => getItemStatus(i) === 'error') && (
-                        <ErrorMessage />
+                        <ErrorMessage
+                            message={detailedErrorMessages.join('\n')}
```
onursumer, line 153: Joining with `\n` is practically the same as joining with any white space. If we really need to format multiple lines of messages I think we need to pass a component instead of a simple string. Since encountering multiple detailed error messages is a low chance event, maybe it's okay to keep it like this for now?

alisman:
This is better.  But I think we should try to just kind of abort the whole page when this happens.   We've struggled with how to this.  Basically we need to dispatch an error to global error handler.  Let me play with it a bit.

PR #4314, opened 2022-07-05
When we detect an invalid profile id in study view filter, show a site error

alisman:
image


onursumer:
@alisman is this based on a wrong branch? There are 35 commits.

src/shared/components/errorScreen/ErrorScreen.tsx
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import _ from 'lodash';
```
onursumer, line 2: unused import?

onursumer (approved):
looks good!

PR #4324, opened 2022-07-11
Left truncation adjustment for KM plot

dippindots:
Please using links:
- https://genie-private.cbioportal.org/study/summary?id=heme_onc_nsclc_genie_bpc&featureFlags=LEFT_TRUNCATION_ADJUSTMENT
- https://genie-private.cbioportal.org/results/comparison?cancer_study_list=heme_onc_nsclc_genie_bpc&tab_index=tab_visualize&case_set_id=heme_onc_nsclc_genie_bpc_all&Action=Submit&gene_list=TP53&comparison_subtab=survival&featureFlags=LEFT_TRUNCATION_ADJUSTMENT
to test this feature.

What is in this pr:
- Left truncation adjustment algorithm for KM plot
- Update unit tests
- Update tooltip for KM plot
- Add number of risk in the tooltip when hover over the data point in the KM plot
- Show how many patients get excluded when applying left truncation algorithm


inodb:
@dippindots really nice work!

For study view:

- [ ] I like the banner re adjustment for left truncation! Wondering though if we should make it a bit smaller. Like maybe just put "Adjust for left truncation (xx patients excluded) [info icon]" and then when you mouse over we can put the rest of the text. This would also help in getting the survival chart to look square. Usually these plots are square rather than rectangular so it would be good if we can figure out how to accomplish that. Maybe ok also if the background is just white instead of yellow, it is very eye drawing at the moment
- [ ] Maybe we can change the text to something like: "Patients enter a study when they are profiled, which can be months or even years after the initial diagnosis. To mitigate the effects of left truncation one can enable the risk-set adjustment method as described in Brown et al (2022). This involves adjusting patients at risk at time *t* to date of sequencing rather than date of diagnosis (xx patients are excluded because they passed away before their biopsies were sequenced)". I know it's a lot of text so need to figure out how we can display it. Alternatively we can link to the docs site or something with a more elaborate explanation of the method

Results/Comparison View View:

- [ ] Can we add an enable/disable checkbox here as well?

brookemastrog:
@dippindots Excellent work!  This is amazing.  

My only comment is that we should probably adjust this warning: "Kaplan-Meier estimates do not account for the lead time bias introduced by the inclusion criteria for the GENIE BPC Project."  as we can adjust for some curves at this point.  Some of these warnings also come from the data files themselves, so we need to figure out a way to correctly display these warnings.

In the space where the left truncation adjustment is described, it would be convenient/ consistent to have a warning that the curve is not corrected.  In both the smaller plots on the summary page and in the comparison view.

Alternatively, we could change the warning to something more general such as: "Not all Kaplan-Meier estimates account for the lead time bias introduced by the inclusion criteria for the GENIE BPC Project."

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -405,6 +410,22 @@ export class StudySummaryTab extends React.Component<
...
+                    this.store.isLeftTruncationAvailable.result &&
+                    chartMeta.uniqueKey === 'OS_SURVIVAL'
+                ) {
+                    props.isLeftTruncationAvailable = true;
+                    props.patientSurvivalsWithoutLeftTruncation = this.store.survivalPlotData.result.find(
+                        survivalType => survivalType.id === 'OS_SURVIVAL'
+                    )?.survivalDataWithoutLeftTruncation;
+                    if (props.survivalPlotLeftTruncationChecked) {
+                        props.title = `(Adjusted) ${props.title}`;
+                    }
```
inodb, line 427:
```suggestion
@dippindots i think we can remove the (Adjusted) text from the title now that you have made a more elegant solution
```

inodb:
@dippindots for this tooltip:
Screen Shot 2022-08-18 at 11 23 22 AM

It might be good to:

- [ ] Remove `Adjusted` from the title now that you have a more elegant solution
- [ ] Revert the text in the tooltip of the info icon in the header back to what it was. With one minor change. That is: if no left truncation data is available set the text to:`Overall survival in months from first index diagnosis. Kaplan-Meier estimates do not account for the lead time bias introduced by the inclusion criteria for the GENIE BPC Project.`. If it is available remove the last sentence: `Overall survival in months from first index diagnosis.` (since we **are** accounting for it now)

Does that make sense? Thanks!

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -431,4 +432,11 @@ export default class GroupComparisonStore extends ComparisonStore {
                 this.customDriverAnnotationReport.result!.tiers.length > 0)
         );
     }
+
+    // override parent method
+    protected get enableLeftTruncation() {
```
alisman, line 437: enableLeftTruncationForSuvivalData?

src/pages/groupComparison/Survival.tsx
```diff
@@ -659,6 +663,46 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
                                     showCurveInTooltip={true}
+                                    isLeftTruncationAvailable={
+                                        // Currently, left truncation is only appliable for Overall Survival data
+                                        key === 'OS'
+                                            ? this.props.store
+                                                  .isLeftTruncationAvailable
+                                                  .result
+                                            : false
+                                    }
+                                    showLeftTruncationCheckbox={
```
alisman, line 674: seems redundant with the above prop?

src/pages/groupComparison/Survival.tsx
```diff
@@ -659,6 +663,46 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+                                            : false
+                                    }
+                                    isLeftTruncationChecked={
+                                        this.props.store.adjustForLeftTruncation
+                                    }
+                                    patientsCountWithLeftTruncation={
+                                        this.props.store.patientSurvivals
+                                            .result![key].length
+                                    }
+                                    showLeftTruncationCheckboxInWarningTooltip={
```
alisman, line 689: again seems redundant no?

src/pages/resultsView/comparison/ResultsViewComparisonStore.ts
```diff
@@ -182,4 +183,11 @@ export default class ResultsViewComparisonStore extends ComparisonStore {
     public get studies() {
         return this.resultsViewStore.studies;
     }
+
+    // override parent method
+    protected get enableLeftTruncation() {
```
alisman, line 188: forSurvivalData?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8115,11 +8148,76 @@ export class StudyViewPageStore
...
+
     readonly survivalPlotData = remoteData<SurvivalType[]>({
         await: () => [
             this.survivalData,
             this.selectedPatientKeys,
             this.survivalPlots,
+            this.survivalEntryMonths,
         ],
         invoke: async () => {
             return this.survivalPlots.result.map(obj => {
```
alisman, line 8223: i can see that this isn't new but it's a little wrong that we are mutating survivalPlots here, tacking on the survival data.   it's suprising to me that typescript allows this.   in other words, that the survivalPlot object will accept  survivalDataWithoutLeftTruncation as a property.   better to just make consumers look it up using plot id?

alisman:
Style tweaks. Lets remove alert styling.  For one thing this isn't really semantically an "alert".   Just a div with some padding and no border will be good. 
image

alisman:
Tweaks for comparison tab.  Lets move the the checkbox and label outside of the dotted chart container like so.  Again, remove the alert class.  In general, we shouldn't be stuffing controls into the charts.  Sometimes we do so to save space but I don't think it's a good habit. 

image

src/pages/studyView/StudyViewPageStore.ts
dippindots: I fixed it by not changing the property in place, is this looks good? https://github.com/cBioPortal/cbioportal-frontend/pull/4324/commits/313fbab0a939a6f75e4e0897e61a1098e2aa546a

src/pages/resultsView/comparison/ResultsViewComparisonStore.ts
```diff
@@ -18,6 +18,7 @@ import ifNotDefined from '../../../shared/lib/ifNotDefined';
 import comparisonClient from '../../../shared/api/comparisonGroupClientInstance';
 import { ComparisonGroup } from '../../groupComparison/GroupComparisonUtils';
 import { ComparisonSession } from 'shared/api/session-service/sessionServiceModels';
+import { LEFT_TRUNCATION_ADJUSTMENT_FLAG } from '../survival/SurvivalUtil';
```
alisman, line 21: lets keep these feature flag constants in /shared/featureFlags.ts

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8115,22 +8148,75 @@ export class StudyViewPageStore
...
+    >({
+        invoke: async () => {
+            const studyIds = this.studyIds;
+            // Please note:
+            // The left truncation adjustment is only available for one study: heme_onc_nsclc_genie_bpc at this time
+            // clinical attributeId still need to be decided in the future
+            if (
+                studyIds.length === 1 &&
+                studyIds[0] === 'heme_onc_nsclc_genie_bpc' &&
+                this.enableLeftTruncationForSuvivalData
```
alisman, line 8162: typo in that name "Suvival" also, if this is a boolean, lets name it isLeftTruncationForSurvivalDataEnabled.    "enable" is very making me think this is an action and not a piece of state.

src/pages/studyView/StudyViewPageStore.ts
alisman: i don't think this is great either.    it doesn't really make sense to collapse these two types in this way.    maybe just key the survivalPlotData by id and then look it up when we need it?

src/pages/resultsView/comparison/ResultsViewComparisonStore.ts
dippindots: Comments addressed in https://github.com/cBioPortal/cbioportal-frontend/pull/4324/commits/1359463a4b53e54a0530bdb509f659ebb929368b

src/pages/studyView/StudyViewPageStore.ts
dippindots: Comments addressed in https://github.com/cBioPortal/cbioportal-frontend/pull/4324/commits/1359463a4b53e54a0530bdb509f659ebb929368b

src/pages/studyView/StudyViewPageStore.ts
dippindots: changed to a search by id approach in commit : https://github.com/cBioPortal/cbioportal-frontend/pull/4324/commits/dbdd28d6091e14b3ed8c0712192aef608f512737

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -36,14 +39,16 @@ export function getPatientSurvivals(
                             monthsClinicalData.value
                         ) &&
                         !Number.isNaN(Number(monthsClinicalData.value)) &&
-                        Number(monthsClinicalData.value) >= 0
+                        Number(monthsClinicalData.value) >= 0 &&
```
alisman, line 42: lets briefly discuss this condition.  at very least it needs a comment to explain that if we have no entryMonth data then we assume entryMonth = 0

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +91,54 @@ export const survivalClinicalDataNullValueSet = new Set([
     'na',
 ]);
 
+export function sortPatientSurvivals(patientSurvivals: PatientSurvival[]) {
+    // First sort by month in asc order (smaller number to the front)
+    // Then sort by status in desc order (which means if status are '1's, then go to the front, '0's are right after '1's in the same time stamp)
+    return _.orderBy(patientSurvivals, [s => s.months, s => !s.status]);
```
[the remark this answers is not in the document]
dippindots: @alisman I used the way you suggested before and changed it to this because I think it's neat and still easy to understand. I can change it back to `_.orderBy(patientSurvivals, [s => s.months, s => s.status],['asc','desc']);` if you think it can lead to better understanding.

src/pages/resultsView/SurvivalStoreHelper.ts
dippindots: Yes, I agree. I will add a comment here

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -85,8 +91,54 @@ export const survivalClinicalDataNullValueSet = new Set([
     'na',
 ]);
 
+export function sortPatientSurvivals(patientSurvivals: PatientSurvival[]) {
+    // First sort by month in asc order (smaller number to the front)
+    // Then sort by status in desc order (which means if status are '1's, then go to the front, '0's are right after '1's in the same time stamp)
```
alisman, line 96: fix this comment to show that status is a bolean and not numeric/binary

src/pages/groupComparison/styles.module.scss
```diff
@@ -166,3 +166,8 @@
         font-size: 12px !important;
     }
 }
+
+.NoPaddingLeftTruncationCheckbox {
```
alisman, line 170: make this camelcase (noPadding)

src/pages/resultsView/SurvivalStoreHelper.spec.ts
```diff
@@ -129,11 +129,26 @@ describe('SurvivalStoreHelper', () => {
...
                         studyId: 'study_1',
                         months: 0,
+                        entryMonths: 0,
                         status: true,
                     },
                 ]
             );
         });
+
+        it('returns correct result for example data with entry data', () => {
```
alisman, line 139: returns empty (censored?) because XYZ

src/pages/resultsView/SurvivalStoreHelper.ts
```diff
@@ -26,6 +27,8 @@ export function getPatientSurvivals(
                     const monthsClinicalData = clinicalData.find(
                         c => c.clinicalAttributeId === monthsAttributeId
                     );
+                    const entryMonths =
```
alisman, line 30: out a comment here that this establishes a default entryMonths

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -574,3 +638,107 @@ export function createSurvivalAttributeIdsDict(
...
+                            style={{
+                                marginTop: -3,
+                                paddingRight: 10,
+                            }}
+                        >
+                            Adjust for left truncation
+                            {props.isLeftTruncationChecked && (
+                                <>
+                                    {' ('}
+                                    <b>
```
alisman, line 677: <strong>

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -574,3 +638,107 @@ export function createSurvivalAttributeIdsDict(
...
+export type LeftTruncationCheckboxProps = {
+    className: string;
+    onToggleSurvivalPlotLeftTruncation?: () => void;
+    isLeftTruncationChecked?: boolean;
+    patientSurvivalsWithoutLeftTruncation?: PatientSurvival[];
+    patientToAnalysisGroups: { [uniquePatientKey: string]: string[] };
+    sortedGroupedSurvivals: { [group: string]: PatientSurvival[] };
+};
+
+export const LeftTruncationCheckbox: React.FunctionComponent<LeftTruncationCheckboxProps> = props => {
```
alisman, line 660: make this it's own file and move to shared

src/pages/resultsView/survival/styles.module.scss
```diff
@@ -38,3 +38,9 @@
     padding: 0 5px 0 5px;
     display: inline-block;
 }
+
+.PaddingLeftTruncationCheckbox {
```
alisman, line 42: camel case

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1815,13 +1817,96 @@ export default abstract class ComparisonStore
...
+        { [uniquePatientKey: string]: number } | undefined
+    >({
+        await: () => [this.studies],
+        invoke: async () => {
+            const studyIds = this.studies.result!.map(s => s.studyId);
+            // Please note:
+            // The left truncation adjustment is only available for one study: heme_onc_nsclc_genie_bpc at this time
+            // clinical attributeId still need to be decided in the future
+            if (
+                studyIds.length === 1 &&
```
alisman, line 1835: hide these checks inside the method isLeftTruncationFor...

dippindots:
@alisman Thanks for reviewing this, here is the commit that addresses the most recent comments: https://github.com/cBioPortal/cbioportal-frontend/pull/4324/commits/b59c9000065488a4adc997a7652bcd795879eabc

PR #4332, opened 2022-07-20
Release 5.0.0

alisman:
Merge release-5.0.0 branch into master


env/custom.sh
```diff
@@ -1,2 +1,3 @@
 #export CBIOPORTAL_URL="http://localhost:8080"
 #export GENOME_NEXUS_URL="https://www.genomenexus.org"
+export BACKEND="onursumer:fix-cases-without-panel-data"
```
alisman, line 3: you should build against this branch here. it has a fix that hasn't yet been merged to release-5.0.0

end-to-end-test/local/specs/annotation-filter-menu.spec.js
```diff
@@ -245,7 +257,7 @@ describe('alteration filter menu', function() {
                 clickCheckBoxStudyView('Putative passengers');
             });
 
-            it('filters tables when unchecking when unchecking unknown oncogenicity checkbox', () => {
+            it.skip('filters tables when unchecking when unchecking unknown oncogenicity checkbox', () => {
```
alisman, line 260: @pvannierop it's the tests that are skipped here which relate to annotation filtering.  it kind of makes sense that these would be broken given that whatever script was running to annotate fusions now has to run on structural variant data?  if you need to build

pvannierop: @alisman That the numbers are not correct for the Struct Vars makes sense indeed. I do not know whether the Struct Var implementation also allows for _custom driver annotations_? This feature is what drives the oncogenicity filtering in said menu. Also I do not know how the Struct Vars in _study_es_0_ are supposed to filter according to somatic/germline status.

env/custom.sh
```diff
@@ -1,2 +1,3 @@
 #export CBIOPORTAL_URL="http://localhost:8080"
 #export GENOME_NEXUS_URL="https://www.genomenexus.org"
```
alisman, line 2: @dippindots @inodb lets remember to update this after merge

env/master.sh
```diff
@@ -1,2 +1,2 @@
-export CBIOPORTAL_URL="${CBIOPORTAL_URL:-https://master.cbioportal.org}"
```
alisman, line 1: @dippindots @inodb again, lets remember to update

PR #4333, opened 2022-07-21
Fix IGV sample names displayed next to tracks

onursumer:
Fix cBioPortal/cbioportal#9683

IGV_sample_names

### DIFF
--- a/src/pages/patientView/genomicOverview/GenomicOverview.tsx
+++ b/src/pages/patientView/genomicOverview/GenomicOverview.tsx
@@ -104,8 +104,8 @@ function getGenePanelInfoForIgvTrack(
     );
 }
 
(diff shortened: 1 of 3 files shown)


inodb (approved):
Awesome! Thanks! Not sure exactly how to reproduce the bug but looks sensical :)

onursumer:
@inodb the problem is the [explicit list of samples](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/pages/patientView/genomicOverview/GenomicOverview.tsx#L216) we pass to IGV.

We have to do this to make sure the sample order is correct in case of multiple samples. But apparently if we use sample keys instead of sample ids, then IGV thinks we also want to display keys instead of ids.

PR #4336, opened 2022-07-22
Keep table page as is on row click

onursumer:
Fix cBioPortal/cbioportal#9703

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -311,6 +311,7 @@ export class PatientViewPageStore {
 
     public internalClient: CBioPortalAPIInternal;
 
+    @observable public activeLocus: string | undefined;
     @observable public activeTabId = '';
 
     @observable private _patientId = '';
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -109,9 +109,6 @@ export default class PatientViewPage extends React.Component<
 
     public patientViewPageStore: PatientViewPageStore;
 
-    @observable
-    public activeLocus: string | undefined;
-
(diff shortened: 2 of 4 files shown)


src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -311,6 +311,7 @@ export class PatientViewPageStore {
 
     public internalClient: CBioPortalAPIInternal;
 
+    @observable public activeLocus: string | undefined;
```
onursumer, line 314: Another alternative is to create an explicit `GenomicOverviewStore` in `PatientViewPage` and move this observable there instead of keeping it in `PatientViewPageStore`.

inodb (approved):
LGTM!

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
alisman: I'm confused, it seems like you did this below, no?  i mean make a store.

onursumer: We initialize a new `DefaultGenomicOverviewStore` within the `GenomicOverview` component only if `prop.store` is not defined. However, we are passing the entire `PatientViewPageStore` instance to the component in [PatientViewPageTabs](https://github.com/cBioPortal/cbioportal-frontend/pull/4336/files#diff-7015cbed8f5eb4c9f9fa577da0b6b436038f0f6ea4609bf4a9e64a9a8624ca3fR191). 

So, the actual runtime instance is an instance of `PatientViewPageStore`, not an instance of a `DefaultGenomicOverviewStore`. Alternatively, we can actually instantiate a new `DefaultGenomicOverviewStore` in patient view page instead of passing `PatientViewPageStore` to the component. Probably, not a very important implementation detail though.

alisman (approved):

PR #4350, opened 2022-08-16
Use MSKCC isoforms by default

inodb:
We recently decided to retire use of the uniprot override source. MSKCC should be the new default set of transcripts everywhere


pieterlukasse (approved):

leexgh (approved):

pieterlukasse:
@inodb @alisman is this one scheduled for release soon?

pieterlukasse:
Can we also add `isoformOverrideSource` to this part https://github.com/cBioPortal/cbioportal/blob/3ca82f812fc34922486ef2a9d354674c9403b931/portal/src/main/webapp/config_service.jsp#L31-L143 so that it can be configured via `portal.properties`?

inodb:
Thanks for reviewing @pieterlukasse! I think we only want support the `mskcc` set of transcripts since we are deprecating the uniprot ones (at least for now), so maybe ok to not make it configurable?

PR #4354, opened 2022-08-18
Redirect from legacy static routes to new docs site

alisman:
Fixes [redirects doc site](https://github.com/cbioportal/cbioportal/issues/9753)


inodb (approved):
Looks like this works - thank you!! 🙏

PR #4358, opened 2022-08-22
Restore webAPI page for token download

alisman:
Fix https://github.com/cBioPortal/cbioportal/issues/9760


inodb (approved):
Sweet - thanks so much! Looks like that fixed the issue for MSK portal

PR #4360, opened 2022-08-26
Patient View CNA table should display selection state on row click

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9724

### DIFF
--- a/src/pages/patientView/PatientViewPage.tsx
+++ b/src/pages/patientView/PatientViewPage.tsx
@@ -29,6 +29,7 @@ import { validateParametersPatientView } from '../../shared/lib/validateParamete
 import LoadingIndicator from 'shared/components/loadingIndicator/LoadingIndicator';
 import ValidationAlert from 'shared/components/ValidationAlert';
 import PatientViewMutationsDataStore from './mutation/PatientViewMutationsDataStore';
+import PatientViewCnaDataStore from './copyNumberAlterations/PatientViewCnaDataStore';
 
 import './patient.scss';
 
@@ -107,6 +108,7 @@ export default class PatientViewPage extends React.Component<
     @observable
     public urlWrapper: PatientViewUrlWrapper;
     public patientViewMutationDataStore: PatientViewMutationsDataStore;
+    public patientViewCnaDataStore: PatientViewCnaDataStore;
 
     public patientViewPageStore: PatientViewPageStore;
 
@@ -123,6 +125,11 @@ export default class PatientViewPage extends React.Component<
 
         this.patientViewMutationDataStore = new PatientViewMutationsDataStore(
             () => this.mergedMutations,
+            this.urlWrapper
+        );
+
+        this.patientViewCnaDataStore = new PatientViewCnaDataStore(
+            () => this.mergedCnas,
             this.urlWrapper
         );
 
@@ -206,6 +213,10 @@ export default class PatientViewPage extends React.Component<
                 return !isFusion(mutationArray[0]);
             }
         );
+    }
+
+    @computed get mergedCnas() {
+        return this.patientViewPageStore.mergedDiscreteCNADataFilteredByGene;
     }
 
     componentDidMount() {
@@ -466,8 +477,14 @@ export default class PatientViewPage extends React.Component<
 
     @autobind
     onMutationTableRowClick(d: Mutation[]) {
+        // select mutation and toggle off previous selected
         if (d.length) {
-            this.patientViewMutationDataStore.toggleSelectedMutation(d[0]);
+            this.patientViewMutationDataStore.setSelectedMutations([d[0]]);
+            if (this.patientViewCnaDataStore.selectedCna.length > 0) {
+                this.patientViewCnaDataStore.toggleSelectedCna(
+                    this.patientViewCnaDataStore.selectedCna[0]
+                );
+            }
             this.handleLocusChange(d[0].gene.hugoGeneSymbol);
         }
     }
@@ -486,8 +503,33 @@ export default class PatientViewPage extends React.Component<
 
     @action.bound
     onCnaTableRowClick(d: DiscreteCopyNumberData[]) {
+        // select cna and toggle off previous selected
         if (d.length) {
+            this.patientViewCnaDataStore.setSelectedCna([d[0]]);
+            if (
+                this.patientViewMutationDataStore.selectedMutations.length > 0
+            ) {
+                this.patientViewMutationDataStore.toggleSelectedMutation(
+                    this.patientViewMutationDataStore.selectedMutations[0]
+                );
+            }
             this.handleLocusChange(d[0].gene.hugoGeneSymbol);
+        }
+    }
+
+    @action.bound
+    onResetViewClick() {
+        // toggle off selected cna/mutation row
+        if (this.patientViewCnaDataStore.selectedCna.length > 0) {
+            this.patientViewCnaDataStore.toggleSelectedCna(
+                this.patientViewCnaDataStore.selectedCna[0]
+            );
+        } else if (
+            this.patientViewMutationDataStore.selectedMutations.length > 0
+        ) {
+            this.patientViewMutationDataStore.toggleSelectedMutation(
+                this.patientViewMutationDataStore.selectedMutations[0]
+            );
         }
     }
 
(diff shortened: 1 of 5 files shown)


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -486,7 +514,24 @@ export default class PatientViewPage extends React.Component<
 
     @action.bound
     onCnaTableRowClick(d: DiscreteCopyNumberData[]) {
-        if (d.length) {
+        // if one cna, select it and toggle off other selected
+        if (d.length == 1) {
+            this.patientViewCnaDataStore.setSelectedCna(d);
+            if (this.patientViewMutationDataStore.selectedMutations[0]) {
```
alisman, line 520: just use arr.length

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -29,6 +29,7 @@ import { validateParametersPatientView } from '../../shared/lib/validateParamete
 import LoadingIndicator from 'shared/components/loadingIndicator/LoadingIndicator';
 import ValidationAlert from 'shared/components/ValidationAlert';
 import PatientViewMutationsDataStore from './mutation/PatientViewMutationsDataStore';
+import PatientViewCnaDataStore from './mutation/PatientViewCnaDataStore';
```
onursumer, line 32: CNA data store should probably not be under the `mutation` directory

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -466,8 +477,25 @@ export default class PatientViewPage extends React.Component<
 
     @autobind
     onMutationTableRowClick(d: Mutation[]) {
-        if (d.length) {
-            this.patientViewMutationDataStore.toggleSelectedMutation(d[0]);
+        // if one mutation, select it and toggle off other selected
+        if (d.length == 1) {
```
onursumer, line 481:
```suggestion
        if (d.length === 1) {
```

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -466,8 +477,25 @@ export default class PatientViewPage extends React.Component<
...
+            // this.patientViewMutationDataStore.toggleSelectedMutation(d[0]);
+            this.patientViewMutationDataStore.setSelectedMutations(d);
+            if (this.patientViewCnaDataStore.selectedCna[0]) {
+                this.patientViewCnaDataStore.toggleSelectedCna(
+                    this.patientViewCnaDataStore.selectedCna[0]
+                );
+            }
+            this.handleLocusChange(d[0].gene.hugoGeneSymbol);
```
onursumer, line 489: this `if` block is almost identical to the `else if` block below. It may worth trying to always do `this.patientViewMutationDataStore.setSelectedMutations([d[0]])` and get rid of the conditional selection. Because you are always selecting the first mutation of the array no matter what, right?

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -486,7 +514,24 @@ export default class PatientViewPage extends React.Component<
 
     @action.bound
     onCnaTableRowClick(d: DiscreteCopyNumberData[]) {
-        if (d.length) {
+        // if one cna, select it and toggle off other selected
+        if (d.length == 1) {
+            this.patientViewCnaDataStore.setSelectedCna(d);
```
onursumer, line 519: same here. can't we replace this with `this.patientViewCnaDataStore.setSelectedCna([d[0]]);`?

src/pages/patientView/PatientViewUrlWrapper.ts
```diff
@@ -21,6 +21,7 @@ export type PatientViewUrlQuery = {
         yAxisDataRangeInChart?: string;
 
         showOnlySelectedMutationsInTable?: string;
+        showOnlySelectedCnasInTable: string;
```
onursumer, line 24: not sure if we need this

src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -49,6 +51,8 @@ interface IGenomicOverviewProps {
     onSelectGenePanel?: (name: string) => void;
     disableTooltip?: boolean;
     store?: IGenomicOverviewStore;
+    mutationsStore?: PatientViewMutationsDataStore;
+    cnasStore?: PatientViewCnaDataStore;
```
onursumer, line 55: I would define a prop like `onResetView?:  () => void`, and handle the rest outside of this component. This component does not need to know anything about cna or mutation tables.

src/pages/patientView/mutation/PatientViewCnaDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+> {
+    @observable mouseOverCna: Readonly<DiscreteCopyNumberData> | null = null;
+    private selectedCnasMap = observable.map<string, DiscreteCopyNumberData>();
+
+    public get onlyShowSelectedInTable() {
+        return (
+            this.urlWrapper.query.genomicEvolutionSettings
+                .showOnlySelectedCnasInTable === 'true'
+        );
+    }
```
onursumer, line 31: not sure if we need this logic. we don't use it anywhere right now, right?

src/pages/patientView/mutation/PatientViewCnaDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+function cnaIdKey(c: DiscreteCopyNumberData) {
+    return `{ "entrezGeneId": "${c.entrezGeneId}", "hugoGeneSymbol": "${c.gene.hugoGeneSymbol}" }`;
+}
+
+type CnaIdKey = string;
+
+export default class PatientViewCnaDataStore extends SimpleGetterLazyMobXTableApplicationDataStore<
+    DiscreteCopyNumberData[]
+> {
+    @observable mouseOverCna: Readonly<DiscreteCopyNumberData> | null = null;
```
onursumer, line 23: do we need this one?

src/pages/patientView/mutation/PatientViewCnaDataStore.ts
```diff
@@ -0,0 +1,81 @@
...
+
+        this.dataHighlighter = (mergedCna: DiscreteCopyNumberData[]) => {
+            const highlightedCnas = [];
+            if (!this.onlyShowSelectedInTable) {
+                // dont put highlight on selected cnas if those are all we're showing
+                highlightedCnas.push(...this.selectedCna);
+            }
+            if (this.mouseOverCna) {
+                highlightedCnas.push(this.mouseOverCna);
+            }
```
onursumer, line 77: might be better to simply do this:

```suggestion
          highlightedCnas.push(...this.selectedCna)
```

src/pages/patientView/PatientViewPage.tsx
onursumer: or `isEmpty` function from `lodash`

onursumer (changes requested):

src/pages/patientView/copyNumberAlterations/PatientViewCnaDataStore.ts
```diff
@@ -0,0 +1,67 @@
...
+        const key = cnaIdKey(c);
+        if (this.selectedCnasMap.has(key)) {
+            this.selectedCnasMap.delete(key);
+        } else {
+            this.selectedCnasMap.set(key, c);
+        }
+    }
+
+    @action
+    public setSelectedCna(cna: Readonly<DiscreteCopyNumberData[]>) {
```
alisman, line 36: is there ever a time when there's more than one selected cna?

src/pages/patientView/PatientViewPage.tsx
alisman: also, try to always use trip equals ===
this avoids a feature of javascript called type coercion.    it's just a good practice

alisman (approved):

src/pages/patientView/genomicOverview/GenomicOverview.tsx
```diff
@@ -49,6 +49,7 @@ interface IGenomicOverviewProps {
     onSelectGenePanel?: (name: string) => void;
     disableTooltip?: boolean;
     store?: IGenomicOverviewStore;
+    onResetView: () => void;
```
onursumer, line 52: It would be better to make this optional.

```JavaScript
    onResetView?: () => void;
```

PR #4361, opened 2022-08-29
Calculate heatmap values per group

BasLee:
# Summary
This PR will improve the percentage and color intensity of heatmap cells in the group comparision clinical attributes heatmap.

# Percentage calculation
Groups in the comparison most often do not have the same number of samples. In the old heatmap the value of the buckets or cells where calculated as:

`fraction = samples in bucket / total number of samples across all groups`.

This should be: 

`fraction = samples in bucket / number of samples in respective group`.

Now we can compare the fraction (displayed as percentage) across groups. 

# Color calculation
The color intensity is now scaled to the largest value in the heatmap in order to use the full color spectrum available.

### DIFF
--- a/src/shared/components/plots/MultipleCategoryHeatmap.tsx
+++ b/src/shared/components/plots/MultipleCategoryHeatmap.tsx
@@ -7,7 +7,7 @@ import Oncoprint, {
     IHeatmapTrackSpec,
 } from 'shared/components/oncoprint/Oncoprint';
 import autobind from 'autobind-decorator';
-import { OncoprintJS, InitParams } from 'oncoprintjs';
+import { InitParams, OncoprintJS } from 'oncoprintjs';
 import _ from 'lodash';
 import { IStringAxisData } from 'pages/resultsView/plots/PlotsTabUtils';
 
@@ -38,7 +38,6 @@ export default class MultipleCategoryHeatmap extends React.Component<
 
     constructor(props: any) {
         super(props);
-        makeObservable(this);
     }
 
     @autobind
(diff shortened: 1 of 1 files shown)


src/shared/components/plots/MultipleCategoryHeatmap.tsx
```diff
@@ -141,33 +140,38 @@ function toHeatmapTracks(
...
-            .reduce((partialSum: number, d: number) => partialSum + d, 0)
-    ) as number[];
+    const groupsPercentages: number[][] = [];
+
+    groups.forEach(group => {
+        const groupSampleCount =
+            group.data
+                .map(d => d.profile_data)
+                .reduce((partialSum: number, d: number) => partialSum + d, 0) ||
+            0;
```
alisman, line 150: i don't understand this ```|| 0``` thing here

src/shared/components/plots/MultipleCategoryHeatmap.tsx
```diff
@@ -141,33 +140,38 @@ function toHeatmapTracks(
...
+        });
+    });
 
-    const maxSamples = Math.max(
-        ...([] as any).concat(
-            ...result.map(t => t.data.map(d => d.profile_data))
-        )
+    // Calculate 'heat' of heatmap cell:
+    const percentageMax = Math.max(
+        ...groupsPercentages.reduce((group, result) => result.concat(group), [])
```
alisman, line 166: i think flatten would work here?   flatten is native array function and also there's lodash flatten if necessary

src/shared/components/plots/MultipleCategoryHeatmap.tsx
```diff
@@ -141,33 +140,38 @@ function toHeatmapTracks(
...
     );
+    groups.forEach((group, gi) => {
+        group.data.forEach((d, di) => {
+            d.profile_data = groupsPercentages[gi][di] / percentageMax;
```
alisman, line 170: it seems odd that at line 154, "profile_data" is a sampleCount.   but here are making a percentage?

src/shared/components/plots/MultipleCategoryHeatmap.tsx
BasLee: in theory reduce can return null, so the `|| 0` is needed to fix typescript errors

src/shared/components/plots/MultipleCategoryHeatmap.tsx
BasLee: I had tried flatmap, but both flatMap and flat are not available with the current typescript compiler settings. But I'll try lodash instead

src/shared/components/plots/MultipleCategoryHeatmap.tsx
BasLee: the heatmap expects a value between 0 and 1, so sample count wouldn't work

src/shared/components/plots/MultipleCategoryHeatmap.tsx
BasLee: Fixed in [4e108f5](https://github.com/cBioPortal/cbioportal-frontend/compare/8800d778f277a681acfaaeccc20c4afb87206467..4e108f582b55e15de815e2fec03d53b8905004e7)

src/shared/components/plots/MultipleCategoryHeatmap.tsx
alisman: hmm. that doesn't sound right to me

alisman (approved):

inodb (approved):
LGTM! Thanks!

PR #4363, opened 2022-08-31
Resource table should display a column for resource description field

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9660


src/shared/components/resources/ResourceLink.tsx
```diff
@@ -60,15 +60,15 @@ export default class ResourceLink extends React.Component<
 
     render() {
```
alisman, line 61: Lets get rid of this component.  We need to break it it out into columns anyway so it doesn't really work anymore.  It's pointless.  We just put this out in the Tab

src/shared/components/resources/ResourceTable.tsx
```diff
@@ -0,0 +1,121 @@
...
+        this.props.openResource(resource);
+    }
+
+    render() {
+        return (
+            <table className="simple-table table table-striped table-border-top">
+                <thead>
+                    <tr>
+                        <th>
+                            <span>Resource</span>
```
alisman, line 76: Probably don't need to wrap these in spans?

src/shared/components/resources/ResourceTable.tsx
```diff
@@ -0,0 +1,121 @@
+import * as React from 'react';
```
alisman, line 1: We should convert this into functional component.   We want to avoid making new class components.  Will show you how.

src/shared/components/resources/SampleResourcesTable.tsx
```diff
@@ -67,11 +132,30 @@ export default class SampleResourcesTable extends React.Component<
...
-            <LazyMobXTable
-                columns={this.columns}
-                data={this.data}
-                showPagination={false}
-            />
+            <>
+                {this.data.map(datum => {
+                    return (
+                        <>
+                            <LazyMobXTable
```
alisman, line 139: Shouldn't this make use of the ResourcesTable component?

PR #4364, opened 2022-09-01
Utilize levelExcludedCancerTypes from OncoKB when showing cancer type name

zhx828:
This is the change to work with the upcoming OncoKB release.

### DIFF
--- a/packages/react-mutation-mapper/src/util/OncoKbUtils.ts
+++ b/packages/react-mutation-mapper/src/util/OncoKbUtils.ts
@@ -345,6 +345,19 @@ export function getTumorTypeName(tumorType?: TumorType) {
     }
 }
 
+export function getTumorTypeNameWithExclusionInfo(
+    tumorType?: TumorType,
+    excludedTumorTypes?: TumorType[]
+) {
+    let name = getTumorTypeName(tumorType);
+    if (!_.isEmpty(excludedTumorTypes)) {
+        name = `${name} (excluding ${excludedTumorTypes!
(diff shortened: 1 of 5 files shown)


packages/react-mutation-mapper/src/util/OncoKbUtils.ts
```diff
@@ -345,6 +345,19 @@ export function getTumorTypeName(tumorType?: TumorType) {
     }
 }
 
+export function getTumorTypeNameWithExclusionInfo(
+    tumorType?: TumorType,
+    excludedTumorTypes?: TumorType[]
+) {
+    let name = getTumorTypeName(tumorType);
+    if ((excludedTumorTypes || []).length > 0) {
```
onursumer, line 353: `_.isEmpty` might be easier to read here, not a big deal though.

```suggestion
    if (!_.isEmpty(excludedTumorTypes)) {
```

packages/react-mutation-mapper/src/util/OncoKbUtils.ts
```diff
@@ -345,6 +345,19 @@ export function getTumorTypeName(tumorType?: TumorType) {
...
+        name += ` (excluding ${excludedTumorTypes!
```
onursumer, line 354: string concat makes me uncomfortable :) how about this?

```suggestion
        name = `${name} (excluding ${excludedTumorTypes!
```

onursumer (commented):
Looks good! Just commented on a few minor things

onursumer (approved):
🙌

PR #4368, opened 2022-09-07
Oncoprint tracks menu: move groups and charts to top level

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9789


src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
```diff
@@ -210,30 +130,19 @@ export default class AddClinicalTracks extends React.Component<
...
                         hide={
-                            !this.options.isComplete ||
-                            this.options.result!.clinical.length === 0
+                            !this.props.options.isComplete ||
+                            this.props.options.result!.clinical.length === 0
                         }
                     >
                         {this.addClinicalTracksMenu.component}
                     </MSKTab>
                     <MSKTab
```
alisman, line 139: So, this is what I was looking for.  There is still a possibility that there will be more than one tab because of the custom charts tab.  Thus, we need to keep the menu.   I see that we are already hiding the menu when there is only one option, so it seems like no change is necessary here.    (If there was only ever a single tab, it would make sense to just get rid of the tab structure altogether).

src/pages/resultsView/oncoprint/AddTracks.tsx
```diff
@@ -458,6 +465,75 @@ export default class AddTracks extends React.Component<IAddTrackProps, {}> {
...
+                optionsGivenInSortedOrder={true}
+                frequencyHeaderTooltip="% samples in group"
+                width={this.dropdownWidth}
+            />
+        ),
+        renderPending: () => <LoadingIndicator isLoading={true} small={true} />,
+        showLastRenderWhenPending: true,
+    });
+
+    readonly options = remoteData({
```
alisman, line 517: I think this name could be improved (i realize it was existing).  It's really track options by type.    "trackOptionsByType" ?

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
```diff
@@ -17,16 +17,15 @@ import { ClinicalTrackConfig } from 'shared/components/oncoprint/Oncoprint';
...
-    ) => void;
-    clinicalTrackOptionsPromise: MobxPromise<{
+    clinicalAttributeIdToAvailableFrequencyPromise: MobxPromise<
+        ChartDataCountSet
+    >;
+    width?: number;
+    onAddAll: (keys: string[]) => void;
+    onClearAll: (keys: string[]) => void;
+    onToggleOption: (key: string) => void;
+    selectedClinicalAttributeIds: Dictionary<string>;
```
alisman, line 27: see if you can use the Typescript Record type instead of this lodash supplied type.

alisman: i mean, continue to use _.keyBy, that's correct, but see if you change the typing of this propery to Record if it will work.   Record is generic so you'll have to supply the key and value types, e.g. Record<string,string>

alisman:
@gblaih we need to fix the following, where there is now only one subtab tab visible in clinical tab, we shouldn't show the sub tabs at all.  that logic was actually working before, but something about changes seems to have broken it.

image

src/pages/resultsView/oncoprint/AddClinicalTracks.tsx
alisman: @gblaih @inodb now that I think about it, I don't think the custom charts belongs in the clinical tab submenu either.  so, really we should have 4 top level tabs and NO subtabs ever.

PR #4374, opened 2022-09-16
Patient View: Message does not indicate if sample is profiled

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9708


alisman:
@tmazor could you test this in your portal context
```localStorage.netlify=
'deploy-preview-4374--cbioportalfrontend'
```

src/pages/patientView/structuralVariant/PatientViewStructuralVariantTable.tsx
```diff
@@ -479,7 +479,21 @@ export default class PatientViewStructuralVariantTable extends React.Component<
             ) {
                 return (
                     <div className="alert alert-info" role="alert">
-                        Structural Variants are not available.
+                        Study has no Structual Variant Data.
```
alisman, line 482: data shouldn't be capiltazied

src/pages/patientView/structuralVariant/PatientViewStructuralVariantTable.tsx
```diff
@@ -479,7 +479,21 @@ export default class PatientViewStructuralVariantTable extends React.Component<
...
+                    </div>
+                );
+            } else if (
+                !this.props.store.genePanelDataByMolecularProfileIdAndSampleId
```
alisman, line 486: lets move this ugly condition to the top of the render method and assign it to a nice local variable 
```
const isProfiled = ...
```

src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx
```diff
@@ -370,7 +371,15 @@ export default class CopyNumberTableWrapper extends React.Component<
             <div>
                 {this.props.status === 'unavailable' && (
                     <div className="alert alert-info" role="alert">
-                        Copy Number Alterations are not available.
```
alisman, line 373: lowercase "data"

tmazor:
This is great! A few tweaks though:

1) Text should say 'Sample/Patient is not profiled for XXX' where XXX is Mutations or Copy Number Alterations or Structural Variants.
image
Image above is from our internal portal, but you can see a similar example in public portal here: https://www.cbioportal.org/patient?studyId=mpcproject_broad_2021&sampleId=RP-1532_PCProject_0064_BLOOD_P_v1_WGS_OnPrem
image

2) My original issue also requested updates to the Pathways tab, as it has the same underlying issue. Is that supposed to be part of this or a separate fix?

3) I found another related issue that I missed in the initial ticket, so I added a new ticket to capture it: https://github.com/cBioPortal/cbioportal/issues/9811

tmazor:
Update for clarity: only my first comment needs to be addressed in this PR. 2&3 can both be follow-ups to this.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -278,23 +278,6 @@ export default class PatientViewPage extends React.Component<
         updateOncoKbIconStyle({ mergeIcons });
     }
 
-    @computed get cnaTableStatus() {
-        if (this.patientViewPageStore.molecularProfileIdDiscrete.isComplete) {
```
alisman, line 282: it's odd that we get rid of this check ```molecularProfileIdDiscrete.result``` here but don't reintroduce it elsewhere

src/pages/patientView/structuralVariant/StructuralVariantTableWrapper.tsx
```diff
@@ -474,12 +474,27 @@ export default class PatientViewStructuralVariantTable extends React.Component<
             this.columns,
         ],
         render: () => {
+            const isProfiled = this.props.store.structuralVariantProfile.result
```
alisman, line 477: we should work on this logic here.  it's quite brittle and can be made more tolerant by using something called "optional chaining".   i see that this is following the pattern above so we can improve there too.

src/shared/lib/isSampleProfiled.ts
```diff
@@ -81,3 +82,24 @@ export function isSampleProfiledInMultiple(
...
+        [sampleId: string]: GenePanelData;
+    };
+}
+
+export function isSampleProfiledInProfile(
+    genePanelMap: IGenePanelDataByProfileIdAndSample,
+    profileId: string | undefined,
+    sampleId: string | undefined
+) {
+    return (
```
alisman, line 97: @gblaih i decided it was better to spell things out after all.  But at least just do it once place

src/pages/patientView/PatientViewPageTabs.tsx
```diff
@@ -97,6 +95,8 @@ export function tabs(
 ) {
     const tabs: JSX.Element[] = [];
 
+    console.log('scratch', pageComponent.patientViewPageStore.samples.result);
```
alisman, line 98: oops

src/pages/patientView/PatientViewPageTabs.tsx
```diff
@@ -275,7 +288,7 @@ export function tabs(
                                 }
                                 sampleIds={
                                     sampleManager
-                                        ? sampleManager.getSampleIdsInOrder()
+                                        ? sampleManager.getActiveSampleIdsInOrder()
```
alisman, line 291: @gblaih this was the big change.  It turns out the SampleManger is the "one" that knows the about the ordering of samples (based on a clinical data). However, the getSampleIdsInOrder doesn't know about which sample is "active" in the page (sample mode or patient mode in which all are active).   In fact, it DOES have the information (sampleIds collection passed at construction), so it was easy to add this.  Now, we can get the ordered list of active sample ids, which is what the tables need.

PR #4375, opened 2022-09-19
Tweak behavior of study search feature on homepage

alisman:
- Study search should NOT filter the list of example queries
- Dropdown should close on blur of textbox and on selection of sample query

### DIFF
--- a/src/shared/components/query/StudySearch.tsx
+++ b/src/shared/components/query/StudySearch.tsx
@@ -4,7 +4,7 @@ import {
     toQueryString,
 } from 'shared/lib/query/textQueryUtils';
 import * as React from 'react';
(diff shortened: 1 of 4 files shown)


onursumer:
study_search_dropdown

@alisman found a minor issue related to example queries. To reproduce:

- Click on the dropdown menu right next to the search box and make sure you see the example queries
- Enter text, say `msk`, into the search box
- Now press the mouse button on any of the example queries without actually releasing the button, i.e. keep the `mouse down` state for a while

What happens: The dropdown menu disappears without actually updating the search to the selected example.
Expected behavior: It should update the search text with the selection after releasing the mouse button.

onursumer (approved):
🙌

PR #4377, opened 2022-09-21
Study View: Gene charts compare button should show when only one gene is selected and forward user to results view comparison tab

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9804

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -8820,7 +8820,7 @@ export class StudyViewPageStore
     }
 
     @autobind
-    onSubmitQuery(): void {
+    onSubmitQuery(url: string, gene?: string): void {
         const unknownQueriedIdsMap = stringListToSet(
             this.unknownQueriedIds.result
         );
@@ -8909,14 +8909,12 @@ export class StudyViewPageStore
             }
         }
 
-        let url = '/';
-        if (!_.isEmpty(this.geneQueries)) {
-            formOps.Action = 'Submit';
-            formOps.gene_list = this.geneQueries
-                .map(query => unparseOQLQueryLine(query))
-                .join('\n');
-            url = '/results';
-        }
+        formOps.Action = 'Submit';
+        formOps.gene_list = gene
+            ? gene
+            : this.geneQueries
+                  .map(query => unparseOQLQueryLine(query))
+                  .join('\n');
         submitToPage(url, formOps, '_blank');
     }
 
--- a/src/pages/studyView/charts/ChartContainer.tsx
+++ b/src/pages/studyView/charts/ChartContainer.tsx
@@ -34,6 +34,7 @@ import {
     logScalePossible,
     MutationCountVsCnaYBinsMin,
     NumericalGroupComparisonType,
+    updateGeneQuery,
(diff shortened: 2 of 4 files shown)


src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -427,15 +427,28 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
-                onClick: () => {
-                    this.openComparisonPage(
-                        getComparisonParamsForTable(
-                            this.selectedRowsKeys,
-                            this.chartType
-                        )
-                    );
-                },
-                isDisabled: () => this.selectedRowsKeys!.length < 2,
+                onClick:
```
alisman, line 430: this should really be just single callback and in the callback we can check to see how many selected rows there are and act accordingly.   in general we want to keep the callbacks we embed in our JSX as minimal as possible and put business logic in methods or functions.  i.e. make onClick: ()=>this.handleComparisonClick
and then in this method, we check selectedRows and forward user

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -427,15 +427,28 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
+                        ? () => {
+                              this.openComparisonPage(
+                                  getComparisonParamsForTable(
+                                      this.selectedRowsKeys,
+                                      this.chartType
+                                  )
+                              );
+                          }
+                        : () => {
+                              this.props.store.onCheckGene(
```
alisman, line 441: why do we need to do this onCheckGene thing?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8909,13 +8909,22 @@ export class StudyViewPageStore
...
-        let url = '/';
-        if (!_.isEmpty(this.geneQueries)) {
-            formOps.Action = 'Submit';
-            formOps.gene_list = this.geneQueries
-                .map(query => unparseOQLQueryLine(query))
-                .join('\n');
-            url = '/results';
+        if (url) {
+            if (!_.isEmpty(this.geneQueries)) {
+                formOps.Action = 'Submit';
```
alisman, line 8914: this code is duplicated below and its presence muddies the conditional.    i would handle the case where there is no url using a default value for the url arg 
```
onSubmitQuery(url:string="/"){...}
```
then you can just check to see if there's geneQueries and update url to results

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -260,6 +261,21 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
             onDeleteChart: () => {
                 this.props.onDeleteChart(this.props.chartMeta);
             },
+            onComparisonClick: () => {
+                if (this.selectedRowsKeys!.length > 1) {
```
alisman, line 265: we should put a comment here that selectedRowKeys will ALWAYS be gene names.  and also perhaps an explanation of why we do this fork.  i.e when we have a single gene selected we. want to show comparison tab altered/not-altered of results view, whereas for multiple selected genese we want to show group comparison

PR #4381, opened 2022-09-23
Update Genome Nexus API Client

alisman:


leexgh:
This is related to https://github.com/genome-nexus/genome-nexus/pull/619. @averyniceday could you take a look?

src/shared/api/genomeNexusClientInstance.ts
```diff
@@ -2,10 +2,12 @@ import { GenomeNexusAPI } from 'genome-nexus-ts-api-client';
 
 async function checkVersion(client: GenomeNexusAPI) {
     const versionResp = await client.fetchVersionGET({});
-    if (parseInt(versionResp.version.split('.')[0]) !== 1) {
+    if (
+        parseInt(versionResp.genomeNexus.database.version.split('.')[0]) !== 1
```
leexgh, line 6:
```suggestion
        parseInt(versionResp.genomeNexus.server.version.split('.')[0]) !== 1
```
@averyniceday Maybe we should check `server` instead? Please correct me if wrong

src/shared/api/genomeNexusClientInstance.ts
```diff
@@ -2,10 +2,12 @@ import { GenomeNexusAPI } from 'genome-nexus-ts-api-client';
...
+    ) {
         console.error(
             'Expected version of Genome Nexus to be 1.x.y, but found: ' +
-                versionResp.version
+                versionResp.genomeNexus.database.version
```
leexgh, line 10:
```suggestion
                versionResp.genomeNexus.server.version
```
Probably should be `server` too

PR #4384, opened 2022-09-29
Timeline: allow custom event renderers to defer to default color getter

alisman:
Fixes https://github.com/cbioportal/cbioportal/issues/9836


end-to-end-test/remote/specs/core/timeline.spec.js
```diff
@@ -0,0 +1,52 @@
...
+        // now restore it
+        $$('.tl-timeline-wrapper .fa-caret-right')[0].click();
+
+        assert.equal($$('.tl-timeline-tracklabels > div').length, 15);
+    });
+
+    it('timeline zooms in on drag and drop', () => {
+        $('.tl-timeline-svg').waitForDisplayed();
+
+        const moo = $$('.tl-timelineviewport text').find(t =>
```
alisman, line 41: maybe moo isn't the greatest name

src/pages/patientView/timeline/timeline_helpers.tsx
```diff
@@ -125,8 +125,6 @@ export function configureTimelineToxicityColors(baseConfig: ITimelineConfig) {
             const path = e.containingTrack.uid.split('.');
             if (path[0] === 'TREATMENT' && path.length > 2) {
                 return getColor(path[2]);
-            } else {
-                return POINT_COLOR;
```
onursumer, line 129: Ideally, we should explicitly do `return undefined` here, so that it's clear that for some cases we don't return a valid value. Right now, I think the function now implicitly returns undefined for this case.

PR #4389, opened 2022-10-05
Add testing library as prod dep

alisman:


dippindots (approved):
Thanks for updating this, I introduced this package `@testing-library/react` in pr https://github.com/cBioPortal/cbioportal-frontend/pull/4383. But forgot to move this package from `devDependencies` to `dependencies`

PR #4390, opened 2022-10-06
Remove samples of overlapping patients in patient-level analysis

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9824

If choose "Exclude overlapping patients and samples", filter out overlapped samples from sample list
Results:
Exclude overlapping patients and samples:
image

Include overlapping patients and samples:
image

### DIFF
--- a/src/pages/groupComparison/ClinicalData.tsx
+++ b/src/pages/groupComparison/ClinicalData.tsx
@@ -35,8 +35,6 @@ import {
     CLINICAL_TAB_NOT_ENOUGH_GROUPS_MSG,
     ClinicalDataEnrichmentWithQ,
     getStatisticalCautionInfo,
-    getOverlappingPatients,
-    getOverlappingSamples,
 } from './GroupComparisonUtils';
 import ReactSelect from 'react-select1';
 import { MakeMobxView } from 'shared/components/MobxView';
@@ -44,16 +42,15 @@ import OverlapExclusionIndicator from './OverlapExclusionIndicator';
 import { RESERVED_CLINICAL_VALUE_COLORS } from '../../shared/lib/Colors';
 import ErrorMessage from '../../shared/components/ErrorMessage';
 import ComplexKeyMap from 'shared/lib/complexKeyDataStructures/ComplexKeyMap';
-import {
-    PatientIdentifier,
-    Sample,
-    SampleIdentifier,
-} from 'cbioportal-ts-api-client';
+import { Sample } from 'cbioportal-ts-api-client';
 import ComparisonStore, {
     OverlapStrategy,
 } from '../../shared/lib/comparison/ComparisonStore';
 import { createSurvivalAttributeIdsDict } from 'pages/resultsView/survival/SurvivalUtil';
-import { getComparisonCategoricalNaValue } from './ClinicalDataUtils';
+import {
+    filterSampleList,
+    getComparisonCategoricalNaValue,
+} from './ClinicalDataUtils';
 import {
     ClinicalNumericalDataVisualisation,
     ClinicalNumericalVisualisationType,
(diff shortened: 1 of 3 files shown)


src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -338,17 +346,54 @@ export default class ClinicalData extends React.Component<
...
+            // If include overlapped patients and samples, keep all sample in sample list
+            let filteredSamples: Sample[] = [];
+            if (this.props.store.overlapStrategy === OverlapStrategy.EXCLUDE) {
+                const overlappingPatients = getOverlappingPatients(
+                    this.props.store._activeGroupsNotOverlapRemoved.result!
+                );
+                const overlappingPatientsMap: {
+                    [studyId: string]: Set<string>;
+                } = {};
+                overlappingPatients.map(p => {
```
alisman, line 360: user _.groupBy instead

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -338,17 +346,54 @@ export default class ClinicalData extends React.Component<
...
+                    }
+                    overlappingPatientsMap[p.studyId].add(p.patientId);
+                });
+                const overlappingSamples = getOverlappingSamples(
+                    this.props.store._activeGroupsNotOverlapRemoved.result!
+                );
+                const overlappingSamplesMap: {
+                    [studyId: string]: Set<string>;
+                } = {};
+                overlappingSamples.map(s => {
```
alisman, line 372: same

src/pages/groupComparison/ClinicalData.tsx
alisman: actually, it's not totally clear to me why you have to put them into groups by study.  they should have unique ID which should allow you to filter them without regard for study?

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -338,17 +350,40 @@ export default class ClinicalData extends React.Component<
...
+                const overlappingSamplesMap = getOverlappingSamplesMap(
+                    getOverlappingSamples(
+                        this.props.store._activeGroupsNotOverlapRemoved.result!
+                    )
+                );
+                filteredSamples = this.props.store.samples.result!.filter(
+                    sample =>
+                        !overlappingPatientsMap[sample.studyId]?.has(
+                            sample.patientId
+                        ) &&
```
leexgh, line 372: @alisman 
> actually, it's not totally clear to me why you have to put them into groups by study. they should have unique ID which should allow you to filter them without regard for study?

Yes it would be easier if it has unique id, but `overlappingPatients` (this.props.store._activeGroupsNotOverlapRemoved) doesn't have `uniquePatientKey`, it only has `studyId` and `patientId`. When we compare `overlappingPatients` with `this.props.store.samples`, we have to use both  `studyId` and `patientId`, so I put them in groups by study Id and put patient Ids in value. 
Same for `overlappingSamples`.

alisman: I see.  Ok!

alisman: would it be trivial to add the unique id wherever that thing is calculated?

alisman: @leexgh i'm talking about this logic here.  like, move this seperate function in utils

leexgh: Done

alisman (approved):

alisman (approved):

inodb (commented):
Looks great! Thank you!

Maybe good to update this message as well:

> Patients (4) that overlap in the selected groups are excluded from patient-level analysis below.

to

> Patients (4) that overlap in the selected groups are excluded from patient-level analysis below and their samples (x) are excluded from the sample-level analysis.

It did seem like the samples were still included for some charts, e.g. for the `Oncotree code` barchart here https://genie-private.cbioportal.org/comparison/clinical?comparisonId=632caa9002558d441fb7ff30&overlapStrategy=Exclude. It looks like the samples numbers are the same as those for the original groups, rather than that the samples associated to the overlapping patients are excluded

inodb (approved):
Thinking about this some more - I actually think now that it makes sense to keep the samples from the overlapping patients when you are comparing sample attributes. E.g. look at this [prim vs met example](https://www.cbioportal.org/comparison/overlap?sessionId=5cf89323e4b0ab413787436c&selectedEnrichmentEventTypes=%5B%22missense%22%2C%22missense_mutation%22%2C%22missense_variant%22%2C%22indel%22%2C%22nonframeshift%22%2C%22inframe%22%2C%22inframe_del%22%2C%22inframe_deletion%22%2C%22in_frame_del%22%2C%22in_frame_deletion%22%2C%22nonframeshift_deletion%22%2C%22inframe_ins%22%2C%22inframe_insertion%22%2C%22in_frame_ins%22%2C%22in_frame_insertion%22%2C%22nonframeshift_insertion%22%2C%22truncating%22%2C%22feature_truncation%22%2C%22nonsense_mutation%22%2C%22nonsense%22%2C%22stopgain_snv%22%2C%22stop_gained%22%2C%22frameshift%22%2C%22frameshift_variant%22%2C%22frame_shift_del%22%2C%22frameshift_deletion%22%2C%22frame_shift_ins%22%2C%22frameshift_insertion%22%2C%22translation_start_site%22%2C%22initiator_codon_variant%22%2C%22start_codon_snp%22%2C%22start_codon_del%22%2C%22de_novo_start_outofframe%22%2C%22nonstop_mutation%22%2C%22stop_lost%22%2C%22splice_site%22%2C%22splice%22%2C%22splicing%22%2C%22splice_site_snp%22%2C%22splice_site_del%22%2C%22splice_site_indel%22%2C%22splice_region_variant%22%2C%22splice_region%22%2C%22silent%22%2C%22synonymous_variant%22%2C%22targeted_region%22%2C%22other%22%5D). You would want to keep those samples of patients that have both primary and met in your sample level comparisons

And so then my comment re updating the message doesn't really make sense either, since that's not how it works

PR #4392, opened 2022-10-10
Lollipop plots in group comparison page shows mutations for two or more groups

gblaih:
Implement issue https://github.com/cBioPortal/GSoC/issues/88


packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
```diff
@@ -146,6 +146,24 @@ export default class SVGAxis extends React.Component<SVGAxisProps, {}> {
                         : horizontalPadding);
                 transform = '';
             }
+            let wrappedLabel;
+            if (this.props.label.endsWith('Mutations')) {
+                wrappedLabel = (
```
alisman, line 151: what's reason for this?    we should evaluate whether there's some more generalized way to handle this.  also, make sure to put a comment on code like

packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
```diff
@@ -157,9 +175,10 @@ export default class SVGAxis extends React.Component<SVGAxisProps, {}> {
                     fill="#2E3436"
                     x={x}
                     y={y}
-                    transform={transform}
+                    transform={transform + ' translate(0, -30)'}
                 >
-                    {this.props.label}
+                    {wrappedLabel}
+                    {/* {this.props.label} */}
```
alisman, line 181: lets get rid of this

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                        ],
+                        sampleMolecularIdentifiers,
+                    } as MutationMultipleStudyFilter,
+                }
+            );
+            return mutations;
+        },
+    });
+
+    readonly countProfiledSamples = remoteData({
```
onursumer, line 353: `profiledSamplesCount` or `numberOfProfiledCases` would be better.

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+        await: () => [this.allSamples, this.coverageInformation],
+        invoke: async () => {
+            let count = 0;
+            if (this.selectedMutationMapperGene) {
+                const sampleKeys = this.allSamples.result!.map(
+                    s => s.uniqueSampleKey
+                );
+                sampleKeys.map(k => {
```
onursumer, line 361: I would do `sampleKeys.filter(...).length` to find the count

onursumer (changes requested):

onursumer:
@alisman @gblaih we want to enable this feature for the results view page `Comparison/Survival Tab` as well, right? I think it is only enabled for the standalone comparison page for now.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    @action.bound
+    private onScaleToggle(selectedScale: AxisScale) {
+        this.lollipopPlotControlsConfig.yMaxInput = undefined;
+        this.lollipopPlotControlsConfig.bottomYMaxInput = undefined;
+
+        this.showPercent = selectedScale === AxisScale.PERCENT;
+
+        if (this.props.onScaleToggle) {
```
alisman, line 61: get rid of this

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -130,6 +131,11 @@ export default class GroupComparisonPage extends React.Component<
             this.store.methylationEnrichmentProfiles,
             this.store.survivalClinicalDataExists,
             this.store.genericAssayEnrichmentProfilesGroupedByGenericAssayType,
+            this.store.mutationsByGroup,
+            this.store.availableGenes,
+            this.store.mutations,
+            // this.store.allSamples,
```
alisman, line 137: cleanup

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -213,6 +219,35 @@ export default class GroupComparisonPage extends React.Component<
...
+                        <MSKTab
+                            id={GroupComparisonTab.MUTATIONS}
+                            linkText="Mutations"
+                            anchorClassName={
+                                this.store.mutationsTabUnavailable
+                                    ? 'greyedOut'
+                                    : ''
+                            }
+                        >
+                            <Mutations store={this.store} />
```
alisman, line 232: Lets call this ComparisonMutationTab

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
         },
     });
 
+    readonly mutations = remoteData({
+        await: () => [this.samples, this.mutationEnrichmentProfiles],
+        invoke: async () => {
+            if (
+                this.samples.result!.length === 0 ||
+                this.mutationEnrichmentProfiles.result!.length === 0 ||
```
alisman, line 329: we need to verify that this is the right set of profiles. @onursumer can you comment?  why is "enrichmentProfiles"? Should there ever be more than one of these?

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                !this.selectedMutationMapperGene
```
alisman, line 330: @inodb what should the default gene by?  should we show something by default?  this is tied up with what genes we want to offer the user by default.

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                    ) {
+                        count += 1;
+                    }
+                });
+            }
+            return count;
+        },
+    });
+
+    readonly allSamples = remoteData({
```
alisman, line 379: make sure that we are properly filtering overlapping samples.  (if in fact we want exclude them like other tabs)

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+        },
+    });
+
+    readonly genePanelDataForAllProfiles = remoteData({
+        await: () => [this.allSamples, this.mutationEnrichmentProfiles],
+        invoke: async () => {
+            if (
+                this.allSamples.result!.length === 0 ||
+                this.mutationEnrichmentProfiles.result!.length === 0 ||
+                !this.selectedMutationMapperGene
```
alisman, line 404: assume we have a gene

alisman: just kill this if statement

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+        await: () => [
+            this.genePanelDataForAllProfiles,
+            this.sampleKeyToSample,
+            this.patients,
+        ],
+        invoke: () => {
+            if (this.selectedMutationMapperGene) {
+                return Promise.resolve(
+                    getCoverageInformation(
+                        this.genePanelDataForAllProfiles.result!,
```
alisman, line 433: genePanelDataForMutationProfiles

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+        await: () => [this.allSamples],
+        invoke: () => fetchPatients(this.allSamples.result!),
+        default: [],
+    });
+
+    readonly availableGenes = remoteData<Gene[]>({
+        invoke: async () => {
+            const genes = await getAllGenes();
+            return genes.sort((a, b) =>
+                a.hugoGeneSymbol.localeCompare(b.hugoGeneSymbol)
```
alisman, line 454: probably don't want to use "localeCompare" because our site isn't localized

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+            );
+        },
+    });
+
+    readonly loadOptions = (inputText: string, callback: any) => {
```
alisman, line 459: move this into the actual MutationsTab component

alisman: should be a computed

src/pages/groupComparison/GroupComparisonStore.ts
onursumer: I think it's called enrichment profiles because the data is mainly used for enrichment in the genomic alterations tab. There can be more than one of these depending on the number of studies, because we are fetching profiles for all studies https://github.com/cBioPortal/cbioportal-frontend/blob/f99379744fe3453a8dbaaed0cd8d97d3e4e362c3/src/shared/lib/comparison/ComparisonStore.ts#L2086

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+    public clearSelectedMutationMapperGene() {
+        this._selectedMutationMapperGene = '';
+    }
+
+    @autobind
+    public applySampleIdFilter(
+        filter: DataFilter<string>,
+        mutation: Mutation
+    ): boolean {
+        const group = _(this.mutationsByGroup.result!)
```
alisman, line 499: you can just use filter.values[0] as key on the mutationsByGroup result.

alisman: be careful to make sure that the key exists

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,225 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+
+    readonly mutationsByGroup = remoteData({
+        await: () => [this.mutations, this.activeGroups],
+        invoke: async () => {
+            const mutationsBySampleId = _.keyBy(
+                this.mutations.result!,
+                m => m.sampleId
+            );
+
+            const ret = this.activeGroups
```
alisman, line 520: comment is called for here.  why 2 groups?

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1055,3 +1063,20 @@ const AlterationMenuHeader: React.FunctionComponent<{
         );
     }
 );
+
+export function getSampleMolecularIdentifiers(
+    selectedSamples: Sample[],
+    profiles: MolecularProfile[]
```
alisman, line 1069: if there are multiple profiles per study, this won't work. determine if that's really the case here.  (whether there can be multiple profiles)

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+    private mutationMapperToolStore: MutationMapperToolStore;
+    private groupComparisonMutationMapper:
+        | GroupComparisonMutationMapper
+        | undefined;
+
+    constructor(props: IMutationProps) {
+        super(props);
+        this.mutationMapperToolStore = new MutationMapperToolStore(
+            this.props.mutations ||
+                _(this.props.store.mutationsByGroup.result!)
```
alisman, line 42: get rid of fallback mechanism.  should always come through mutations prop

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    .value()
+                    .map(group => ({
+                        group: group,
+                        filter: {
+                            type: 'GroupComparisonFilter',
+                            values: [group],
+                        },
+                    })),
+                filterAppliersOverride: {
+                    ['GroupComparisonFilter']: this.props.store
```
alisman, line 58: see if typescript will let you do   { GroupComparisonFilter: this.props.store }

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+                        .applySampleIdFilter,
+                },
+                countUniqueMutations: this.countUniqueMutations,
+            }
+        );
+    }
+
+    @computed
+    get showPercent() {
+        return this.groupComparisonMutationMapper
```
alisman, line 68: Parent component should not reach into child component.  Ideally child componet should accept a callback as a prop, which fires when say the percentage is changed and allows the parent to update the shared state.

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+    @autobind
+    private onMutationMapperInit(
+        mutationMapper: GroupComparisonMutationMapper
+    ) {
+        this.groupComparisonMutationMapper = mutationMapper;
+    }
+
+    public render() {
+        const selectorContent =
+            this.props.store.activeGroups.result!.length < 2 ? (
```
alisman, line 109: just change this to if {} else if {} else

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+                    <div>
+                        <Loader
+                            isLoading={
+                                this.mutationMapperToolStore
+                                    .mutationMapperStores.isPending &&
+                                this.props.store.mutations.isPending
+                            }
+                        />
+                        {mutationMapperStore &&
+                            this.props.store.mutations.isComplete &&
```
alisman, line 167: check to see if some of these should be added to loader above

src/pages/groupComparison/GroupComparisonPage.tsx
alisman: await these newly added promises (only used by new tab) inside of new tab so that other tabs don't have to wait for these things

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    @autobind
+    private onMutationMapperInit(
+        mutationMapper: GroupComparisonMutationMapper
+    ) {
+        this.groupComparisonMutationMapper = mutationMapper;
+    }
+
+    public render() {
```
alisman, line 107: we want to use MobxPromiseView to await promise specific to this tab and also provide the loader.

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+        const selectorContent =
+            this.props.store.activeGroups.result!.length < 2 ? (
+                <span>{MUTATIONS_NOT_ENOUGH_GROUPS_MSG}</span>
+            ) : this.props.store.activeGroups.result!.length > 2 ? (
+                <span>{MUTATIONS_TOO_MANY_GROUPS_MSG}</span>
+            ) : (
+                <div style={{ width: 200, paddingBottom: 10 }}>
+                    <AsyncSelect
```
alisman, line 115: make a functional component to wrap this thing

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+                            }
+                        />
+                        {mutationMapperStore &&
+                            this.props.store.mutations.isComplete &&
+                            this.props.store.activeGroups.result!.length ===
+                                2 && (
+                                <div>
+                                    <h3>
+                                        {_(
+                                            this.props.store.mutationsByGroup
```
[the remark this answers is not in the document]
alisman: _.keys().slice().join()

src/pages/groupComparison/Mutations.tsx
```diff
@@ -0,0 +1,206 @@
...
+                                            this.plotLollipopTooltipCountInfo
+                                        }
+                                    />
+                                </div>
+                            )}
+                    </div>
+                </>
+            );
+        } else {
+            return <>{selectorContent}</>;
```
alisman, line 203: doesn't need react fragment wrapper.

alisman: <> is shortchut for <React.Fragment> which is only necessary when you have multiple items you want to return together but you have don't have parent

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -89,8 +92,12 @@ export default class MutationMapperToolStore {
         undefined
     );
 
-    constructor() {
+    constructor(
+        mutationData?: Partial<MutationInput>[],
```
alisman, line 96: we don't need this anymore

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -76,6 +76,11 @@ export interface IMutationMapperProps {
     onTranscriptChange?: (transcript: string) => void;
     onClickSettingMenu?: (visible: boolean) => void;
     onOncoKbIconToggle?: (mergeIcons: boolean) => void;
+    getMutationCount?: (mutation: Partial<Mutation>) => number;
```
alisman, line 79: not used

src/pages/groupComparison/AxisScaleSwitch.tsx
```diff
@@ -0,0 +1,91 @@
...
+                onClick={onClick}
+            >
+                {scale}
+            </button>
+        );
+    }
+
+    public render() {
+        return (
+            <React.Fragment>
```
alisman, line 68: you don't need fragment since it's a single child

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,177 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                a.hugoGeneSymbol < b.hugoGeneSymbol ? -1 : 1
+            );
+        },
+    });
+
+    public setDefaultGene() {
+        this._selectedMutationMapperGene = this.genesWithMaxFrequency[0].hugoGeneSymbol;
+    }
+
+    @computed get selectedMutationMapperGene() {
```
alisman, line 436: activeMutationMapperGene so that conceptually it accomodatds both the default and user selected state

alisman: then you can have another computed "userSelectedMutationMapperGene" which returns the raw _userSelectedMutationMapperGene

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -0,0 +1,91 @@
...
+@observer
+export default class GroupComparisonMutationsTab extends React.Component<
+    IGroupComparisonMutationsTabProps,
+    {}
+> {
+    constructor(props: IGroupComparisonMutationsTabProps) {
+        super(props);
+    }
+
+    readonly selectorUI = MakeMobxView({
```
alisman, line 30: not much point to this MobxView.  might as well just put the wrapped component in tabUI

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -0,0 +1,91 @@
...
+                    </div>
+                    {this.selectorUI.component}
+                    <GroupComparisonMutationsTabPlot
+                        store={this.props.store}
+                        mutations={_(this.props.store.mutationsByGroup.result!)
+                            .values()
+                            .flatten()
+                            .value()}
+                        filters={{
+                            groupFilters: _(
```
alisman, line 61: _.(result).keys() and then you don't have to call .value() since you aren't actually chaining

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+    mutations?: Mutation[];
+    filters?: any;
+}
+
+@observer
+export default class GroupComparisonMutationsTabPlot extends React.Component<
+    IGroupComparisonMutationsTabPlotProps,
+    {}
+> {
+    @observable public showPercent = true;
```
alisman, line 34: axisMode: "percent" | "absolute"

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+    constructor(props: IGroupComparisonMutationsTabPlotProps) {
+        super(props);
+        makeObservable(this);
+    }
+
+    private mutationMapperToolStore = remoteData<MutationMapperToolStore>({
+        await: () => [this.props.store.mutationsByGroup],
```
alisman, line 41: we don't need to await here because we should be able to assume that the property is loaded

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
alisman: use enum typing as discussed

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+            : countUniqueMutations(mutations);
+    }
+
+    @action.bound
+    private onScaleToggle(selectedScale: AxisScale) {
+        this.showPercent = selectedScale === AxisScale.PERCENT;
+    }
+
+    @autobind
+    protected plotLollipopTooltipCountInfo(
```
alisman, line 66: make this a pure function above the class and pass in showPercent

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+            this.props.store.profiledSamplesCount,
+            this.mutationMapperToolStore,
+        ],
+        render: () => {
+            const mutationMapperStore = this.mutationMapperToolStore.result!.getMutationMapperStore(
+                this.props.store.selectedMutationMapperGene!.hugoGeneSymbol
+            );
+            return (
+                <div>
+                    <Loader
```
alisman, line 100: can we use the MakeMobxView renderPending or whatever it's called to show the loader?

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+                        isLoading={
+                            this.mutationMapperToolStore.result!
+                                .mutationMapperStores.isPending &&
+                            this.props.store.mutations.isPending &&
+                            this.props.store.mutationsByGroup.isPending &&
+                            this.props.store.profiledSamplesCount.isPending
+                        }
+                    />
+                    {mutationMapperStore &&
```
alisman, line 109: you can probably assume you have mutationmapperStore since you are awaiting them above

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -0,0 +1,148 @@
...
+            this.props.store.mutationsByGroup,
+            this.props.store.profiledSamplesCount,
+            this.mutationMapperToolStore,
+        ],
+        render: () => {
+            const mutationMapperStore = this.mutationMapperToolStore.result!.getMutationMapperStore(
+                this.props.store.selectedMutationMapperGene!.hugoGeneSymbol
+            );
+            return (
+                <div>
```
alisman, line 99: this div is probably unnceessary

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,177 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                        sampleMolecularIdentifiers,
+                    } as MutationMultipleStudyFilter,
+                }
+            );
+            return mutations;
+        },
+    });
+
+    readonly profiledSamplesCount = remoteData({
+        await: () => [this.allSamples, this.coverageInformation],
```
alisman, line 347: @inodb @onursumer  i don't understand why the denominator is all samples here.  that's all samples in ALL studies which are present here.

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -314,6 +321,177 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+        this._selectedMutationMapperGene = gene.hugoGeneSymbol;
+    }
+
+    @action.bound
+    public clearSelectedMutationMapperGene() {
+        this._selectedMutationMapperGene = '';
+    }
+
+    @autobind
+    public applySampleIdFilter(
```
alisman, line 454: shoudlApplySampleIdFilter

src/pages/groupComparison/LollipopGeneSelector.tsx
```diff
@@ -0,0 +1,67 @@
...
+import { observer } from 'mobx-react';
+import * as React from 'react';
+import AsyncSelect from 'react-select/async';
+import GroupComparisonStore from './GroupComparisonStore';
+
+interface ILollipopGeneSelectorProps {
+    store: GroupComparisonStore;
+}
+
+export const LollipopGeneSelector: React.FC<ILollipopGeneSelectorProps> = ({
```
alisman, line 11: give this a prop of genese

src/pages/groupComparison/LollipopGeneSelector.tsx
```diff
@@ -0,0 +1,67 @@
...
+                isClearable={true}
+                isSearchable={true}
+                defaultOptions={store.availableGenes
+                    .result!.slice(0, 200)
+                    .map(gene => ({
+                        label: gene.hugoGeneSymbol,
+                        value: gene,
+                    }))}
+                defaultValue={
+                    store.selectedMutationMapperGene
```
alisman, line 52: you won't need this because there will be a gene

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1580,6 +1606,30 @@ export default abstract class ComparisonStore
         );
     }
 
+    @computed get mutationsTabShowable() {
```
alisman, line 1609: isMutationsTabShowable

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1580,6 +1606,30 @@ export default abstract class ComparisonStore
...
+    @computed get showMutationsTab() {
+        return !!(
+            this.mutationsTabShowable ||
+            (this.activeGroups.isComplete &&
+                this.activeGroups.result!.length === 0 &&
+                this.tabHasBeenShown.get(GroupComparisonTab.MUTATIONS))
+        );
+    }
+
+    @computed get mutationsTabUnavailable() {
```
alisman, line 1625: isMutationsTabUnavailable

packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
alisman: remind me to discuss this with you again.  we have to come up with a better solution.

packages/react-mutation-mapper/src/component/lollipopPlot/LollipopPlotNoTooltip.tsx
```diff
@@ -689,11 +689,12 @@ export default class LollipopPlotNoTooltip extends React.Component<
         groupName?: string,
         symbol: string = '#'
     ) {
-        const label = groupName
+        const label = this.props.yAxisLabelFormatter
+            ? this.props.yAxisLabelFormatter(symbol, groupName)
+            : groupName
```
onursumer, line 694: We should avoid nested ternary conditionals. Let's add an if else statement here for `this.props.yAxisLabelFormatter`

src/globalStyles/global.scss
```diff
@@ -753,3 +753,7 @@ h2.divider {
         }
     }
 }
+
+#mutationMapperToolTabs .nav-pills {
+    display: none;
+}
```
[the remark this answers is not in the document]
onursumer: This seems still broken. This change removes the pills on the standalone mutation mapper page, so it only shows one gene and we cannot switch to different genes anymore. We still need tabs for multiple genes.

src/pages/groupComparison/LollipopTooltipCountInfo.tsx
```diff
@@ -0,0 +1,36 @@
...
+} from 'cbioportal-utils';
+import { AxisScale } from 'react-mutation-mapper';
+
+interface ILollipopTooltipCountInfoProps {
+    count: number;
+    mutations?: Mutation[];
+    axisMode?: AxisScale;
+}
+
+export const LollipopTooltipCountInfo: React.FC<ILollipopTooltipCountInfoProps> = ({
```
onursumer, line 17: Considering that this code is mostly taken from signaldb codebase we should also move this into `react-mutation-mapper` so that we can reuse it in signaldb to avoid duplication.

src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -51,6 +53,7 @@ import { getGenomeNexusHgvsgUrl } from 'shared/api/urls';
 import { GENOME_NEXUS_ARG_FIELD_ENUM } from 'shared/constants';
 import { getServerConfig } from 'config/config';
 import { REFERENCE_GENOME } from 'shared/lib/referenceGenomeUtils';
+import { DataFilterType } from 'react-mutation-mapper';
```
onursumer, line 56: unused?

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1060,6 +1068,24 @@ export default abstract class ComparisonStore
         },
     });
 
+    @computed get genesWithMaxFrequency(): AlterationEnrichmentRow[] {
```
onursumer, line 1071: This is taken from another incomplete PR, but I think the function name is not ideal. We are returning an array of `AlterationEnrichmentRow` and calling it `genesWithMaxFrequency`

onursumer (changes requested):

PR #4395, opened 2022-10-19
Fix hidden columns when switching transcripts

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9755

Reason:
Annotation column is excluded when go to non-canonical transcript, so there is no visibility setting for annotation column in user selection store. 
When switching to canonical transcript, the annotation column is available now, but store doesn't have an existing visibility setting for annotation column. The logic simply replaces default visibility setting with setting we have in user selection store, so the annotation column is missing in the new visibility setting, therefore it's always hidden

Fix:
Merge user selection store visibility with default visibility, so "new" columns' can also be in the list and their visibility can be updated.

### DIFF
--- a/packages/cbioportal-frontend-commons/src/lib/ColumnVisibilityResolver.ts
+++ b/packages/cbioportal-frontend-commons/src/lib/ColumnVisibilityResolver.ts
@@ -1,3 +1,5 @@
+import _ from 'lodash';
+
 export interface ISimpleColumnVisibilityDef {
     name: string;
     id?: string;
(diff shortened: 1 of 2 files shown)


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -910,10 +910,7 @@ export default class LazyMobXTable<T> extends React.Component<
     }
 
     protected updateColumnVisibility(id: string, visible: boolean) {
-        // ignore undefined columns
-        if (this.store.columnVisibility[id] !== undefined) {
```
leexgh, line 914: @onursumer Is there a reason we need to "ignore undefined columns"?

packages/cbioportal-frontend-commons/src/lib/ColumnVisibilityResolver.ts
```diff
@@ -35,8 +37,10 @@ export function resolveColumnVisibility(
 
     if (columnVisibility) {
         colVis = {
-            // if a custom columnVisibility object is provided use that one
-            ...columnVisibility,
+            // if columnVisibility object is provided (e.g. when swtiching transcripts), overwrite corresponding default visibility
+            // so if columnVisibilityByColumnDefinition contains more columns than columnVisibility
+            // the extra columns can still keep the default visibility and prevent being lost
+            ..._.merge(columnVisibilityByColumnDefinition, columnVisibility),
```
alisman, line 43: seems slight confusing because we are already in a merging process (...).  i think we could just start with ...columnVsibiilityByColumnDefinition and that would be same thing?

alisman: We should take this opportunity to improve the naming of the inputs of this function.  It is not at all clear what they actually are and thus why the precedence is what it is.  I realize this is not your work but if you now understand what they are, lets make better names and comments.  For example, is columnVisibilityOverride = userSelectedColumnVisibility ?  
If so, then "override" is not a great name.

alisman: Also, i don't get what the difference is between columnVisibilityByColumnDefinition and columnVisibility.

onursumer: `columnVisiblityByColumnDefinition` is computed by processing the `columns` definition passed to the table.
According to the comments [here](https://github.com/cBioPortal/cbioportal-frontend/blob/d96636d489d0348c3c76d26c300a09452b6dfc34/src/shared/components/lazyMobXTable/LazyMobXTable.tsx#L290-L298) `columnVisibility` always refers to `props.columnVisibility` except possibly once "Reset columns" has been clicked, and `columnVisibilityOverride` keeps the state of the latest user selection.

Not sure if we really need all of these. There might be a simpler way to handle the selection.

src/shared/components/lazyMobXTable/LazyMobXTable.tsx
onursumer: Hmm, I can't remember the exact reason but seems okay to remove this.

alisman (approved):

PR #4398, opened 2022-10-24
Add feature flag store

alisman:


src/appBootstrapper.tsx
```diff
@@ -34,6 +34,8 @@ import Container from 'appShell/App/Container';
 import request from 'superagent';
 import { IServerConfig } from 'config/IAppConfig';
 import { initializeGenericAssayServerConfig } from 'shared/lib/GenericAssayUtils/GenericAssayConfig';
+import { parseUrl } from 'query-string';
```
dippindots, line 37: We could clean some unused imports here like `observable` and `parseUrl`

src/shared/FeatureFlagStore.ts
```diff
@@ -0,0 +1,40 @@
+import { observable } from 'mobx';
+import { parseUrl } from 'query-string';
+import _ from 'lodash';
+
+const FEATURE_FLAG_ID = 'featureFlags';
```
dippindots, line 5: Maybe add another constant `FEATURE_FLAG_DELIMITER = ','`.

src/shared/FeatureFlagStore.ts
```diff
@@ -0,0 +1,40 @@
...
+            localStorage.getItem(FEATURE_FLAG_ID)?.split(',') || [];
+        let newState = _.reject(storageState, s => s === flag);
+        this.currentState = newState;
+        localStorage.setItem(FEATURE_FLAG_ID, this.currentState.join(','));
+
+        window.location.href = window.location.href.replace(
+            /featureFlags=[^&]+/,
+            ''
+        );
+    }
```
dippindots, line 39: Maybe we can add another method `contains(flag: string)` or `has(flag: string)`, because eventually, we want to know if I can enable a feature or not, so we should have a function for that purpose. What do you think?

dippindots (approved):
@alisman Thanks for doing this! Looks good to me, just have some small comments.

dippindots (approved):
Thanks for the update! Looks good to me!

PR #4402, opened 2022-10-28
Add spinner to iFrame of Resource Tabs

pvannierop:
# Problem

When loading resources from external URL in a Resource Tab there it is not communicated to the user that the page is loading.

# Solution

Use the existing IFrameLoader component for loading the resources in the Resource Tab. This shows a spinner while the page waits for data.

# Example
Peek 2022-10-28 10-27

### DIFF
--- a/src/shared/components/iframeLoader/IFrameLoader.tsx
+++ b/src/shared/components/iframeLoader/IFrameLoader.tsx
@@ -8,13 +8,18 @@ interface FrameLoaderProps {
     url: string;
     className?: string;
     iframeId?: string;
-    height?: number;
+    height?: number | string;
+    width?: number | string;
 }
 @observer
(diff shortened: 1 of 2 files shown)


src/shared/components/iframeLoader/IFrameLoader.tsx
```diff
@@ -30,7 +35,7 @@ export default class IFrameLoader extends React.Component<
     //NOTE: we need zindex to be higher than that of global loader
     render() {
         return (
-            <div style={{ position: 'relative' }}>
+            <div style={{ position: 'relative', width: this.props.width }}>
```
alisman, line 38: It's odd that we apply the height to the iframe but the width to the div.  Let's discuss how to sort this out better.

PR #4403, opened 2022-10-28
Fix bug where patient pathway lacks a CNA profile

alisman:
Pathways tab currently breaks on this page: https://www.cbioportal.org/patient?studyId=lgg_ucsf_2014&caseId=P01. This fixes that


src/pages/patientView/pathwayMapper/PatientViewPathwayMapper.tsx
```diff
@@ -137,10 +141,10 @@ export default class PatientViewPathwayMapper extends React.Component<
                                 .genePanelDataByMolecularProfileIdAndSampleId
                                 .result
                         }
-                        molecularProfiles={[
-                            this.props.store.mutationMolecularProfile.result!,
-                            this.props.store.discreteMolecularProfile.result!,
-                        ]}
+                        molecularProfiles={_.compact([
```
alisman, line 144: @onursumer i will follow up with better more explicit fix

inodb (approved):
Looks like that solves it -thanks!

onursumer (approved):
Thanks for the quick fix!

PR #4405, opened 2022-10-31
Update OncoKB Client

zhx828:
We updated the host in the swagger definition. I think the version of swagger we use here does not include host info.


onursumer (approved):
:dart:

PR #4407, opened 2022-11-02
Show reference genome in study list 

BasLee:
To assist the user in selecting compatible studies, the study list now shows a ~badge~ text of the used reference genome. It is only shown when a cbioportal instance uses multiple reference genomes.

Screenshot from 2023-02-01 15-46-20
_Dummy study list to illustrate label_

Changes:
- Label that shows used reference genome (when multiple reference genomes are available)

- Label is disabled by default using property `skin_home_page_show_reference_genome` (backend property can be added in the future).

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -250,6 +250,15 @@ export class QueryStore {
         return referenceGenomes.length === 1;
     }
 
+    @computed
+    get multipleReferenceGenomesPresentInAllStudies() {
+        const allStudies = this.treeData.map_studyId_cancerStudy;
+        const referenceGenomes = _.uniq(
+            Array.from(allStudies.values()).map(s => s.referenceGenome)
+        );
+        return referenceGenomes.length > 1;
+    }
+
     ////////////////////////////////////////////////////////////////////////////////
     // QUERY PARAMETERS
     ////////////////////////////////////////////////////////////////////////////////
--- a/src/shared/components/query/studyList/StudyList.tsx
+++ b/src/shared/components/query/studyList/StudyList.tsx
@@ -61,6 +61,10 @@ export default class StudyList extends QueryStoreComponent<
         if (this.props.showSelectedStudiesOnly) this._view = this.view;
     }
 
+    @computed get shouldShowRefGenome() {
+        return this.store.multipleReferenceGenomesPresentInAllStudies;
+    }
+
     render() {
         let studyList = this.renderCancerType(this.rootCancerType);
 
@@ -238,6 +242,8 @@ export default class StudyList extends QueryStoreComponent<
                                 {!this.store.isDeletedVirtualStudy(
                                     study.studyId
                                 ) && this.renderSamples(study)}
+                                {this.shouldShowRefGenome &&
+                                    this.renderReferenceGenomeLabel(study)}
                                 {this.renderStudyLinks(study)}
                             </div>
                         );
@@ -246,17 +252,6 @@ export default class StudyList extends QueryStoreComponent<
             </li>
         );
     };
-
-    // renderStudyName = (study:CancerStudy, afterName?:any) =>
-    // {
-    // 	return (
-    // 		<CancerTreeCheckbox view={this.view} node={study}>
-    // 			<span className={styles.StudyName}>
-    // 				{study.name} {afterName || null}
-    // 			</span>
-    // 		</CancerTreeCheckbox>
-    // 	);
-    // }
 
     renderSamples = (study: CancerStudy) => {
         return (
@@ -445,6 +440,12 @@ export default class StudyList extends QueryStoreComponent<
             );
         }
     };
+
+    private renderReferenceGenomeLabel(study: CancerStudy) {
+        return study.referenceGenome ? (
+            <span> | {study.referenceGenome} </span>
+        ) : null;
+    }
 }
 
 export interface ICancerTreeCheckboxProps {
--- a/src/shared/components/query/studyList/StudyList.tsx
+++ b/src/shared/components/query/studyList/StudyList.tsx
@@ -62,7 +62,10 @@ export default class StudyList extends QueryStoreComponent<
     }
 
     @computed get shouldShowRefGenome() {
-        return this.store.multipleReferenceGenomesPresentInAllStudies;
+        return (
+            getServerConfig().skin_home_page_show_reference_genome &&
+            this.store.multipleReferenceGenomesPresentInAllStudies
+        );
     }
 
     render() {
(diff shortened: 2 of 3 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -247,6 +247,16 @@ export class QueryStore {
         return referenceGenomes.length === 1;
     }
 
+    @computed
+    get onlyOneReferenceGenomeInAllStudies() {
```
alisman, line 251: can we reverse the sign of this to multipleReferenceGenomesPresentInAllStudies?

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -31,6 +32,13 @@ export interface IStudyListProps {
     showSelectedStudiesOnly?: boolean;
 }
 
+const referenceGenomeBadgeColors: { [referenceGenome: string]: string } = {
```
alisman, line 35: Lets put this in /shared/constants.ts file to encourage reuse elsewhere

src/shared/components/query/studyList/StudyList.tsx
```diff
@@ -445,6 +448,21 @@ export default class StudyList extends QueryStoreComponent<
...
     };
+
+    private renderReferenceGenomeBadge(study: CancerStudy) {
+        return study.referenceGenome ? (
+            <span
+                className={`badge badge-pill badge-secondary ${styles.badgeReferenceGenome}`}
+                style={{
+                    backgroundColor:
+                        referenceGenomeBadgeColors[study.referenceGenome] ||
+                        'grey',
```
alisman, line 459: we should try to avoid using arbitrary color names like this.   Look up variables.scss in our project and you will find a list of colors.  Perhaps $mediumGray would be good since it's used for the sample pills on patient page.   There are a couple ways to access these.  One would be to use the variable in baedReferenceGenome class.  Then here, you only set backroundColor if one resolves from rferenceGenomeBadgeColors.    There is also a way of exporting variables from the scss file.  We've never done this, but it's 
```
:export {
  primaryColor: $primary-color;
  backgroundColor: $background-color;
  padding: $padding;
}
```
And then you can import them supposedly.

alisman: the former strategy is probably simplest?  except that you then have conditionally apply the backgroungColor in the style={}.  Just a little more code.

BasLee: Exporting out of a scss file didn't work for me, I added grey to the study other theme colors:
https://github.com/cBioPortal/cbioportal-frontend/pull/4407/commits/765125bb1ba294e20e059cfe213bce3d7712325b#diff-02e4860ff55191a861f21c782c3685de5c6db94180bf7daa276ed1dc14ab1765R267

src/shared/components/query/studyList/StudyList.tsx
BasLee: fixed in [765125b](https://github.com/cBioPortal/cbioportal-frontend/pull/4407/commits/765125bb1ba294e20e059cfe213bce3d7712325b)

src/shared/components/query/QueryStore.ts
BasLee: fixed in [765125b](https://github.com/cBioPortal/cbioportal-frontend/pull/4407/commits/765125bb1ba294e20e059cfe213bce3d7712325b)

src/pages/studyView/StudyViewConfig.ts
```diff
@@ -263,6 +264,7 @@ const studyViewFrontEnd = {
             secondary: '#DC3912',
             tertiary: '#f88508',
             quaternary: '#109618',
+            mediumGrey: '#ddd',
```
alisman, line 267: this configuration has todo with studyview.   color doesn't belong here.  see my comment about where it belongs.

BasLee: I am trying to follow the code convention/flow that was already here

src/shared/constants.ts
```diff
@@ -120,3 +121,13 @@ export const DataTypeConstants = {
     BINARY: 'BINARY',
     CATEGORICAL: 'CATEGORICAL',
 };
+
+export const ReferenceGenomeBadgeColors: {
+    [referenceGenome: string]: string;
+} = {
+    hg19: STUDY_VIEW_CONFIG.colors.theme.primary,
```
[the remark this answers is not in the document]
BasLee: I tried to use your scss export, but that didn't seem to work. Is there a working example in the current code base that I could use as a reference?

src/pages/studyView/StudyViewConfig.ts
BasLee: If scss importing doesn't work, what would be a good typescript location for these colors to be moved to?

inodb (commented):
@BasLee I think the pills are a little too eye grabbing for highlighting a technical detail. The publication icon and study view icon should prolly be more prominent compared to the reference genome info. There are also a lot of other properties that might be more interesting features of a study too highlight (e.g. does it have mrna, timeline data, etc). Wondering if it's really important to show this when arriving on the homepage. One option is to e.g. show the reference genome in the front (so the text doesn't move) after you are accidentally combining some that shouldn't be combined?

alisman:
@pvannierop @Sjoerd-van-Hagen adding you guys to this discussion.  see comment above

pvannierop:
@inodb I think these badges are more than technical details. It is an indicator to the user whether certain studies can be selected together for integrated analysis. IOW it drives user interaction.

BasLee:
@inodb @pvannierop @alisman the status of this new feature is a bit uncertain at the moment, maybe we can shelve it for the moment?

pvannierop:
@BasLee Lets try to not use an icon but show inline text like so:

```
123434 samples | hg18 <rest of icons>
```

src/pages/studyView/StudyViewConfig.ts
BasLee: Fixed in https://github.com/cBioPortal/cbioportal-frontend/compare/765125bb1ba294e20e059cfe213bce3d7712325b..65bd1087c4c31369fa5874cd6fa05699e829a225

BasLee:
> @BasLee Lets try to not use an icon but show inline text like so: [..]

Fixed in [bc06f45](https://github.com/cBioPortal/cbioportal-frontend/compare/3a9b5b0a6eef27b47f25019cb8d96dd40039ea3b..bc06f45e43015fcbe6ffba25bd5b23e03d437622)

BasLee:
> @BasLee I think the pills are a little too eye grabbing for highlighting a technical detail. [..]

@inodb Fixed in [bc06f45](https://github.com/cBioPortal/cbioportal-frontend/compare/3a9b5b0a6eef27b47f25019cb8d96dd40039ea3b..bc06f45e43015fcbe6ffba25bd5b23e03d437622): the badge is replaced with a less intrusive text 'label'

pvannierop:
[Show reference genome badge in study list](https://trello.com/c/GZfLMhqe/2-show-reference-genome-badge-in-study-list)

PR #4409, opened 2022-11-04
Only allow expression profiles from tcga pancan in plots tab 

alisman:
Filter out expression profiles when rule for sharing profiles is not met.  This is for multi-study queries in which expression data is not normalized and shouldn't be compared (unless part of same group)

image


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4328,11 +4339,23 @@ export class ResultsViewPageStore
...
                 });
+
+                // expression profiles are not allowed
+                // under some circumstances
+                if (allowExpressionProfiles(profiles)) {
+                    return profiles;
+                } else {
+                    return profiles.filter(
+                        p =>
+                            p.molecularAlterationType !=
```
onursumer, line 4355: Is `!=` on purpose?
```suggestion
                            p.molecularAlterationType !==
```

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4328,11 +4349,28 @@ export class ResultsViewPageStore
...
+
+                // expression profiles are not allowed
+                // under some circumstances
+                if (allowExpressionProfiles(this.studies.result)) {
+                    return profiles;
+                } else {
+                    return profiles.filter(
+                        p =>
+                            ![
+                                AlterationTypeConstants.MRNA_EXPRESSION,
```
alisman, line 4366: @inodb i think it's best to exclude rather than include because if we introduce new profiles they will inadvertantly be excluded.  Of course, this is temporary.  If we're serious about this we actually need to model this in data.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -4328,11 +4349,28 @@ export class ResultsViewPageStore
...
+                                AlterationTypeConstants.MICRO_RNA_EXPRESSION,
+                                AlterationTypeConstants.PROTEIN_LEVEL,
+                                AlterationTypeConstants.RNA_EXPRESSION,
+                                AlterationTypeConstants.MICRO_RNA_EXPRESSION,
```
onursumer, line 4370: the last `AlterationTypeConstants.MICRO_RNA_EXPRESSION` is duplicate

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -421,6 +422,26 @@ export type GeneticEntity = {
...
+        )
+    ) {
+        // if there is only one study
+        // or ALL studies are part of pan_can_atlas and thus have normalized data across expression profiles
+        if (
+            studies.length === 1 ||
+            _.every(studies, s => /pan_can_atlas/i.test(s.studyId))
+        ) {
+            return true;
+        }
```
onursumer, line 439: this is missing an `else` condition, we may end up implicitly returning `undefined` for some cases. How about we just return the evaluated condition?

```suggestion
        return (
            studies.length === 1 ||
            _.every(studies, s => /pan_can_atlas/i.test(s.studyId))
        );
```

src/pages/resultsView/ResultsViewPageStore.ts
inodb: sounds good

PR #4413, opened 2022-11-09
Support '::' structural variant notation in OQL 

pvannierop:
# Background

At present, OQL notation support querying fusion events for a specific gene in this way:

```
KIF5B: FUSION
```
Which translates to _find any fusion event that involves KIF5B_. The result does not distinguish event where KIF5B is up- or downstream.

# Description
This PR extends OQL so that specific orientations of gene fusions can be queried like so:

Find all events where KIF5B is in Gene1 field:
```
KIF5B:: 
KIF5B: FUSION::
```
Find all events where RET is in Gene2 field:
```
::RET
RET: ::FUSION
```

Find all KIF5B-RET events (in resp. Gene1/Gene2 fields):	
```
KIF5B::RET
KIF5B: FUSION::RET
RET: KIF5B::FUSION
```

Find all oncogenic somatic upstream fusions with KIF5B:
```
KIF5B:: : DRIVER_SOMATIC
KIF5B: FUSION::_DRIVER_SOMATIC
```

Find all events where KIF5B is in Gene1 field and Gene2 is NULL/undefined (in database):
```
KIF5B::-
KIF5B: FUSION::-
```

# Comments
- Unit tests are included.

### DIFF
--- a/src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
+++ b/src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
@@ -23,6 +23,7 @@ import {
 } from 'shared/components/GeneSelectionBox/GeneSelectionBoxUtils';
 import autobind from 'autobind-decorator';
 import GeneSymbolValidatorMessage from 'shared/components/GeneSelectionBox/GeneSymbolValidatorMessage';
+import { getGenesFromSingleGeneQuery } from 'shared/lib/oql/oqlfilter';
 
 export interface IGeneSymbolValidatorProps {
     deferOqlError?: boolean;
@@ -154,8 +155,16 @@ export default class GeneSymbolValidator extends React.Component<
     @computed get oql() {
         return getOQL(this.props.geneQuery);
     }
+
     @computed private get geneIds(): string[] {
-        return this.oql.error ? [] : this.oql.query.map(line => line.gene);
+        return this.oql.error
+            ? []
+            : _(this.oql.query)
+                  .flatMap(singleGeneQuery =>
+                      getGenesFromSingleGeneQuery(singleGeneQuery)
+                  )
+                  .uniq()
+                  .value();
     }
 
     @computed
--- a/src/shared/lib/oql/oqlfilter.ts
+++ b/src/shared/lib/oql/oqlfilter.ts
@@ -7,17 +7,22 @@ import oql_parser, {
     CNACommand,
     EXPCommand,
     FUSIONCommand,
+    FUSIONCommandDownstream,
+    FUSIONCommandOrientationBase,
+    FUSIONCommandUpstream,
+    MergedGeneQuery,
     MutationModifier,
     MUTCommand,
     PROTCommand,
+    SingleGeneQuery,
 } from './oql-parser';
 import { annotateAlterationTypes } from './annotateAlterationTypes';
-import { SingleGeneQuery, MergedGeneQuery } from './oql-parser';
 import { ExtendedAlteration } from '../../../pages/resultsView/ResultsViewPageStore';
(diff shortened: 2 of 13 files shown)


inodb:
@pvannierop 

We discussed at steering call to additionally allow:

```
KIF5B: FUSION::
KIF5B: ::FUSION
```
So you can combine it with e.g.:
```
KIF5B: FUSION:: AMP MUT
```

TODO:

- Need to figure out somatic/germline driver notation. Does e.g. `FUSION_SOMATIC_DRIVER` work?

src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
```diff
@@ -154,8 +155,16 @@ export default class GeneSymbolValidator extends React.Component<
...
         return getOQL(this.props.geneQuery);
     }
+
     @computed private get geneIds(): string[] {
-        return this.oql.error ? [] : this.oql.query.map(line => line.gene);
+        return this.oql.error
+            ? []
+            : _(this.oql.query)
+                  .flatMap(singleGeneQuery =>
+                      getGenesFromSingleGeneQuery(singleGeneQuery)
```
alisman, line 164: getGenesFromComplexGeneQuery?  just seems too paradoxical (genes from single gene)

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -528,34 +592,137 @@ function isDatumWantedByAnyTypeWithModifiersCommand<T>(
...
+    gene: string
+): number {
+    var d_fusion = accessors.structuralVariant(datum); // null || true
+    if (d_fusion === null) {
+        // If no fusion data, it's not addressed
+        return 0;
+    }
+    const structuralVariant = (datum as unknown) as StructuralVariant;
+    if (
+        !matchGeneByHygoSymbolOrSpecialValues(
```
alisman, line 672: typo (Hygo)

pvannierop:
@inodb 
>So you can combine it with e.g.:
>KIF5B: FUSION:: AMP MUT

Yes, that is possible. There is a unit test that makes sure this functionality is present (`RET: ::FUSION MUT`).

src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
pvannierop: Well, the fusions are indeed queries that involve multiple genes. From the point of OQL fusions are still a query that concern 1 gene, but asks events where this gene may be fuses to a specific other genes. I do not object to renaming it though. Lets discuss.

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -528,34 +592,137 @@ function isDatumWantedByAnyTypeWithModifiersCommand<T>(
...
+            gene,
+            structuralVariant.site1HugoSymbol
+        )
+    ) {
+        return 0;
+    }
+    if (
+        !matchGeneByHygoSymbolOrSpecialValues(
```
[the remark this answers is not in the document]
pvannierop: Fixed!

src/shared/lib/oql/oqlfilter.ts
pvannierop: Fixed!

pvannierop:
[SV '::' notation in OQL](https://trello.com/c/ClEZbZDl/10-sv-notation-in-oql)

src/shared/components/GeneSelectionBox/GeneSymbolValidator.tsx
alisman: i don't want delay but this name would so puzzle me.  what about getInvolvedGenesFromSingleGeneQuery?

alisman (approved):

PR #4415, opened 2022-11-10
Add message when no available mutations and tabs for top ten genes with highest frequency for group comparison lollipop plot

gblaih:
Addresses issue https://github.com/cBioPortal/cbioportal/issues/9888
Follow up to https://github.com/cBioPortal/cbioportal-frontend/pull/4392

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
@@ -14,6 +14,7 @@ import ErrorMessage from 'shared/components/ErrorMessage';
 import { LollipopGeneSelector } from './LollipopGeneSelector';
 import GroupComparisonMutationsTabPlot from './GroupComparisonMutationsTabPlot';
 import OverlapExclusionIndicator from './OverlapExclusionIndicator';
+import { MSKTab, MSKTabs } from 'shared/components/MSKTabs/MSKTabs';
 
 interface IGroupComparisonMutationsTabProps {
     store: GroupComparisonStore;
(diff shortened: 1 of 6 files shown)


src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -24,8 +25,31 @@ export default class GroupComparisonMutationsTab extends React.Component<
...
+        this.geneTab = id;
+        if (id) {
+            this.props.store.setSelectedMutationMapperGene(id);
+        } else {
+            this.props.store.clearSelectedMutationMapperGene();
+        }
+    }
+
+    @computed get tabs() {
+        const tabs: JSX.Element[] = [];
```
alisman, line 45: this is nitpick but try to avoid declaring a variable and then filling it.  instead, just use your array methods.  this can instead be
this.props.store.genesWithMaxFrequency.map(g=><MSKTab key={g.hugoGeneSymbol} id={g.hugoGeneSymbol} linkText={g.hugoGeneSymbol} />));

alisman: and you can just return that

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -45,22 +69,65 @@ export default class GroupComparisonMutationsTab extends React.Component<
                     </div>
                 );
             }
+
+            let activeTabId;
```
alisman, line 73: i would think this logic should be in a computed getter activeTab

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -45,22 +69,65 @@ export default class GroupComparisonMutationsTab extends React.Component<
...
                     <OverlapExclusionIndicator
                         store={this.props.store}
                         only="sample"
                     />
                     <LollipopGeneSelector
                         store={this.props.store}
                         genes={this.props.store.availableGenes.result!}
+                        handleGeneChange={this.handleGeneChange}
                     />
+                    <div style={{ display: 'inline-flex', height: '1.5em' }}>
```
alisman, line 114: what's the height for?

PR #4419, opened 2022-11-14
Format group comparison lollipop plot y-axis label

gblaih:
Fixes an issue in https://github.com/cBioPortal/cbioportal/issues/9901

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
@@ -22,13 +22,13 @@ interface IGroupComparisonMutationsTabPlotProps {
     filters?: any;
 }
 
-function plotYAxisLabelFormatter(symbol?: string, groupName?: string) {
-    // lowercase = 1 and uppercase = 1.3 (based on 'w' and 'W'), if groupName >= 13, stop and + "..."
+function plotYAxisLabelFormatter(symbol: string, groupName: string) {
+    // lowercase = 1 and uppercase = 1.3 (based on 'w' and 'W'), if groupName >= 22 (13 + 9 leniency), stop and + "..."
     let length = 0;
     let label = '';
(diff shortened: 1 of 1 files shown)


src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -23,7 +23,21 @@ interface IGroupComparisonMutationsTabPlotProps {
 }
 
 function plotYAxisLabelFormatter(symbol?: string, groupName?: string) {
-    return `${symbol} ${groupName} Mutations`;
+    // lowercase = 1 and uppercase = 1.3 (based on 'w' and 'W'), if groupName >= 13, stop and + "..."
```
alisman, line 26: weird that the params of this function are optional    lets see what happens if we make them required

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -23,7 +23,21 @@ interface IGroupComparisonMutationsTabPlotProps {
...
+    let length = 0;
+    let label = '';
+    for (let c of groupName!) {
```
alisman, line 29: why not take symbol into account too?

gblaih: the symbol is always % or # I believe with a space after so I just need to look at the groupName. I found that 13 'w' and 10 'W' with the symbol + space before it is the max.

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
alisman: @gblaih

gblaih: Making them required doesn't seem to change anything so I will change it to that.

PR #4422, opened 2022-11-16
Paginated download of Clinical Table in Study View

pvannierop:
Fixes: cbioportal/cbioportal#9702

# Problem 
With studies with large sample counts or a lot of clinical attributes for patients/samples the table in clinical tab on Study View would not show any data on Chrome browsers.

# Cause
When the download for sample or patient clinical data exceeds ~500Mb Chrome silently drops the response. It shows 
HTTP code 200 in DevTools, but the frontend code receives null. This appears to be a Chrome limitation that is not reported much (but see [here](https://support.google.com/chrome/a/thread/28233217/chrome-80-seems-to-enforce-a-new-limit-on-post-response-size-of-around-300mb-is-this-by-design?hl=en)). Because the Clinical Tab downloads __all__ clinical data before displaying, the 500Mb limit is exceeded for large studies and no data is displayed.

# Solution
This PR implements batch-wise download for data for the Clinical Table in Study View in order to bypass the 500 Mb limit for large studies. 

Optimizations that are included:

- Instead of lists with sample identifiers, the endpoint receives the current Study View filter and resolves sample identifiers from that. This bypasses the upload of large amounts of (compressed) data in the request body.
-  The endpoint returns sample and patient level clinical attributes in the same request. For this, a ClinicalDataCollection model is introduced.

# Notes
- This PR depends on backend cbioportal/cbioportal#9896. Do not merge before that.
- The new `/clinical-data-table/fetch` endpoint supports pagination, filtering with a search term, and sorting. As such it prepares for server side data and lazy loading in the frontend. This functionality requires a new Table React component and should be addressed by future updates.


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8256,31 +8256,37 @@ export class StudyViewPageStore
         default: {},
     });
 
-    readonly getDataForClinicalDataTab = remoteData({
-        await: () => [this.clinicalAttributes, this.selectedSamples],
+    readonly getDataForClinicalDataTab: MobxPromiseUnionTypeWithDefault<
```
alisman, line 8259: seems odd that you have to type this?  like, i would think we would be able to use api types as before?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3022,6 +2971,87 @@ export function updateSavedUserPreferenceChartIds(
     return chartSettings;
 }
 
+export async function getAllClinicalDataByStudyViewFilter(
+    studyViewFilter: StudyViewFilter
+): Promise<{ [sampleId: string]: { [attributeId: string]: string } }> {
+    const localClinicalDataCollection: ClinicalDataCollection = {
```
alisman, line 2977: i'm assuming that what's returned is ulimately just some collections of the ClinicalData type? in which case ts should be able to just infer this from the client's fetch method.

src/pages/studyView/StudyViewUtils.spec.tsx
```diff
@@ -2653,6 +2656,81 @@ describe('StudyViewUtils', () => {
...
+                            studyId: 'study1',
+                            uniqueSampleKey: undefined,
+                            uniquePatientKey: 'p1key',
+                            clinicalAttributeId: 'attr2',
+                            value: 'value2',
+                        },
+                    ],
+                } as unknown) as ClinicalDataCollection,
+                {
+                    s1key: {
```
alisman, line 2687: this seems odd to me.  we are takint a patientClinicalData which has no association with a sample and associating with it with a sample?

packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
```diff
@@ -102,6 +102,24 @@ export type ClinicalAttributeCountFilter = {
 
         'sampleListId': string
 
+};
+export type ClinicalData = {
```
alisman, line 106: this is odd because we already have a ClinicalData type from the public client.  maybe this is true for other types that are used across clients.  not sure.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3022,6 +2971,87 @@ export function updateSavedUserPreferenceChartIds(
...
+    // Add patient level clinical data to sample clinical data.
+    patientKeyedSampleKeyedData = _.mapValues(
+        patientKeyedSampleKeyedData,
+        (sampleKeyedData, patientId) =>
+            _.mapValues(sampleKeyedData, (attrs, sampleId) =>
+                attrs.concat(patientKeyedPatientData[patientId] || [])
+            )
+    );
+    // Remove patient id levels (only keep sample id keys).
+    const sampleKeyedData = _.assign(
```
alisman, line 3036: this seems odd to me.  you are assigning an array (_.values) to an object, which seems to produce an object with numerical indexes as keys { 0:thing, 1:thing, 2:thing  }.  That seems analagous to an array and should probablu just be an array?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3022,6 +2971,87 @@ export function updateSavedUserPreferenceChartIds(
...
+    const patientKeyedPatientData = _.groupBy(
+        clinicalDataCollection.patientClinicalData,
+        d => d.uniquePatientKey
+    );
+    // Add patient level clinical data to sample clinical data.
+    patientKeyedSampleKeyedData = _.mapValues(
+        patientKeyedSampleKeyedData,
+        (sampleKeyedData, patientId) =>
+            _.mapValues(sampleKeyedData, (attrs, sampleId) =>
+                attrs.concat(patientKeyedPatientData[patientId] || [])
```
alisman, line 3032: it seems that this would duplicate the patient level items for each sample?

src/pages/studyView/StudyViewPageStore.ts
pvannierop: I have removed this type definition.

src/pages/studyView/StudyViewUtils.spec.tsx
pvannierop: This is not odd. In the clinical table in Study View each row represents a sample. For each sample row the patient data is appended. So, in the logic to create the row data for the clinical table there is association of patient data.

packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
pvannierop: This code is autogenerated. I think that the public and internal API docs are strictly separated by Swagger.

src/pages/studyView/StudyViewUtils.tsx
pvannierop: No, `sampleKeyedData` are arrays of clinical data keyed by unique sample id like so:

```
{
  s1key => [clinical_atrr1, clinical_attr2, ...]
}
```

In `sampleCollapsedAttributes` this is collapsed into the format accepted by the Clinical Table like so:

```
{
  s1key => {
      clinical_atrr1: clinical_value1,
      clinical_attr2: clinicel_value2
  }
}
```

So, the keying in `sampleKeyedData` is needed to reference all clinical data of a sample. An array is not suitable in this case.

src/pages/studyView/StudyViewUtils.tsx
pvannierop: No, this structure merely keys the sample-level clinical data by patient id like so:

```
{
  s1key: [{p1key: 'p1key'},]
}
```

becomes:

```
{
  p1key: {
     s1key: [{p1key: 'p1key'},]
  }
}
```
In essence, it collects the sample data per patient. This is done to later add the same patient data to each sample datum.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8256,31 +8256,37 @@ export class StudyViewPageStore
...
-            return _.reduce(
-                this.selectedSamples.result,
-                (acc, next) => {
-                    let sampleData: { [attributeId: string]: string } = {
-                        studyId: next.studyId,
-                        patientId: next.patientId,
-                        sampleId: next.sampleId,
-                        ...(sampleClinicalDataMap[next.uniqueSampleKey] || {}),
-                    };
+                [sampleId: string]: { [attributeId: string]: string };
```
alisman, line 8273: again, this type should be inferred by typescript from internals or definition of function ```getAllClinicalDataByStudyViewFilter```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -8772,9 +8778,12 @@ export class StudyViewPageStore
...
     public async getDownloadDataPromise(): Promise<string> {
-        let sampleClinicalDataMap = await getClinicalDataBySamples(
-            this.selectedSamples.result
-        );
+        if (this.selectedSamples.result.length === 0) {
+            return Promise.resolve('');
+        }
+        let sampleClinicalDataMap: {
+            [sampleId: string]: { [attributeId: string]: string };
+        } = await getAllClinicalDataByStudyViewFilter(this.filters);
```
alisman, line 8786: ditto above.  should be inferred

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3022,6 +2971,87 @@ export function updateSavedUserPreferenceChartIds(
...
+    const sampleKeyedData = _.assign(
+        {},
+        ..._.values(patientKeyedSampleKeyedData)
+    );
+    // Put all clinical attributes in one object.
+    const sampleCollapsedAttributes = _.mapValues(
+        sampleKeyedData,
+        clinicalData => {
+            const data = _.map(clinicalData, (datum: ClinicalData) => {
+                const obj = {} as { [attrId: string]: string };
```
alisman, line 3045: const data: Record<string, string> = {} is stronger than using "as"

src/pages/studyView/StudyViewUtils.tsx
alisman: i see, so values produces an array of objects which have sampleIds as keys and we are just copying them onto one object

src/pages/studyView/StudyViewUtils.spec.tsx
```diff
@@ -95,6 +97,7 @@ import { remoteData, toPromise } from 'cbioportal-frontend-commons';
 import { autorun, observable, runInAction } from 'mobx';
 
 import { AlterationTypeConstants, DataTypeConstants } from 'shared/constants';
+import { ClinicalDataCollection } from 'cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal';
```
alisman, line 100: this should just be 
```import { ClinicalDataCollection } from 'cbioportal-ts-api-client'```
IDE seems to get it wrong.  but we have a test that catches it

src/pages/studyView/StudyViewUtils.tsx
pvannierop: I do not quite understand what part you are referring to. Do you object against `Promise<{ [sampleId: string]: { [attributeId: string]: string } }>` or `Promise<{ [sampleId: string]: { [attributeId: string]: string } }>`? I think using explicit types is always better than implicit? Let's discuss...

src/pages/studyView/StudyViewPageStore.ts
pvannierop: Ok, I removed this type declaration.

src/pages/studyView/StudyViewPageStore.ts
pvannierop: I removed this type declaration.

src/pages/studyView/StudyViewUtils.tsx
pvannierop: Ah I did not know. Updated.

src/pages/studyView/StudyViewUtils.spec.tsx
pvannierop: For me this does not work for types of the Internal client. We should discuss how this should work.

pvannierop:
I will run _prettier_ when code review is finished.

src/pages/studyView/StudyViewUtils.spec.tsx
alisman: @pvannierop i think i see reason.  if you look at ```packages/cbioportal-ts-api-client/src/index.tsx``` there is a comment which explains that we need to manually export types from the internal client (to avoid clashes with the public client.  So that's why it's not working for you with my suggestion above.

pvannierop: I have added the types manually to index.d.ts of the api client and updated all occurences of this problem in the code.

alisman (approved):

PR #4425, opened 2022-11-17
Add driver/VUS categories for group comparison lollipop plot

gblaih:
Fixes an issue in https://github.com/cBioPortal/cbioportal/issues/9901

- Shows and enables driver/VUS filter icons


src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -622,4 +657,330 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                    | null
+                    | false =
+                    this.driverAnnotationSettings.oncoKb &&
+                    getOncoKbMutationAnnotationForOncoprint &&
+                    !(
+                        getOncoKbMutationAnnotationForOncoprint instanceof Error
+                    ) &&
+                    getOncoKbMutationAnnotationForOncoprint(mutation);
+
+                const isHotspotDriver =
```
alisman, line 739: Are we using isHotspotDriver etc in this context or can we just kill all this code

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -622,4 +657,330 @@ export default class GroupComparisonStore extends ComparisonStore {
...
+                    this.driverAnnotationSettings.cosmicCount,
+                    cosmicCountExceeded,
+                    this.driverAnnotationSettings.customBinary,
+                    this.driverAnnotationSettings.driverTiers
+                );
+            });
+        },
+    });
+
+    readonly oncoKbMutationAnnotationForOncoprint = remoteData<
```
alisman, line 771: change name since not for Oncoprint

src/shared/lib/comparison/AnalysisStore.ts
```diff
@@ -0,0 +1,383 @@
...
+import { fetchHotspotsData } from '../CancerHotspotsUtils';
+import { GenomeNexusAPIInternal } from 'genome-nexus-ts-api-client';
+import {
+    countMutations,
+    mutationCountByPositionKey,
+} from 'pages/resultsView/mutationCountHelpers';
+import ComplexKeyCounter from '../complexKeyDataStructures/ComplexKeyCounter';
+import GeneCache from 'shared/cache/GeneCache';
+
+export default abstract class AnalysisStore {
```
alisman, line 45: @gblaih is there any reason to worry that these members were changed from what they were in the old resultsviewpagestore?  i assume they were just copied but just want to be sure.

gblaih: the abstract members are different for the resultsviewpagestore vs comparisonstore but the rest are just copied and should have the same behavior.

src/shared/lib/comparison/AnalysisStore.ts
```diff
@@ -0,0 +1,383 @@
...
+import {
+    CancerStudy,
+    Gene,
+    Mutation,
+    MutationCountByPosition,
+    Sample,
+} from 'cbioportal-ts-api-client';
+import { computed, makeObservable, observable } from 'mobx';
+import _ from 'lodash';
+import { compileMutations } from '../../../pages/resultsView/ResultsViewPageStoreUtils';
```
onursumer, line 11: Ideally we shouldn't be referencing to `pages/resultsView` here. Might be better to move the function to `AnalysisStoreUtils`

src/shared/lib/comparison/AnalysisStore.ts
```diff
@@ -0,0 +1,383 @@
...
+    fetchOncoKbCancerGenes,
+    fetchOncoKbDataForOncoprint,
+    filterAndAnnotateMutations,
+    getGenomeNexusUrl,
+    makeGetOncoKbMutationAnnotationForOncoprint,
+    makeIsHotspotForOncoprint,
+    ONCOKB_DEFAULT,
+} from 'shared/lib/StoreUtils';
+import MobxPromise, { MobxPromiseUnionTypeWithDefault } from 'mobxpromise';
+import { AnnotatedMutation } from '../../../pages/resultsView/ResultsViewPageStore';
```
onursumer, line 25: We should move `AnnotatedMutation` type somewhere else. It is not ideal to have a circular reference like this.

onursumer (changes requested):
Looks good. We just need to sort out a few imports.

alisman (approved):

onursumer (approved):
Looks good, thanks!

PR #4436, opened 2022-12-01
Capture selected scale in URL in group comparison lollipop plot

gblaih:
Fixes issue in https://github.com/cBioPortal/cbioportal/issues/9901

- Makes scale toggle icons clearer and captures state in URL

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
@@ -132,6 +132,7 @@ export default class GroupComparisonMutationsTab extends React.Component<
                     </div>
                     <GroupComparisonMutationsTabPlot
                         store={this.props.store}
+                        urlWrapper={this.props.urlWrapper}
                         mutations={_(this.props.store.mutationsByGroup.result!)
                             .values()
                             .flatten()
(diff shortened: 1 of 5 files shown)


src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -132,6 +132,7 @@ export default class GroupComparisonMutationsTab extends React.Component<
                     </div>
                     <GroupComparisonMutationsTabPlot
                         store={this.props.store}
+                        urlWrapper={this.props.urlWrapper}
```
alisman, line 135: it would be better to give this component a property onScaleToggle and then update the urlWrapper at this level.

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -130,7 +132,11 @@ export default class GroupComparisonMutationsTabPlot extends React.Component<
                             plotLollipopTooltipCountInfo={
                                 plotLollipopTooltipCountInfo
                             }
-                            axisMode={this.axisMode}
+                            axisMode={
```
alisman, line 135: again, just give the component an axisMode property and keep it agnostic to the url, which is really a concert of the page component

gblaih: fixed

PR #4442, opened 2022-12-05
Default axis scale to frequency in group comparison lollipop plot

gblaih:
Fixes issue when there is no selected axis scale. When there is no selected axis scale, it should default to frequency mode.

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
@@ -124,7 +124,10 @@ export default class GroupComparisonMutationsTabPlot extends React.Component<
                             plotLollipopTooltipCountInfo={
                                 plotLollipopTooltipCountInfo
                             }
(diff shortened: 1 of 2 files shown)


src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -124,7 +124,10 @@ export default class GroupComparisonMutationsTabPlot extends React.Component<
                             plotLollipopTooltipCountInfo={
                                 plotLollipopTooltipCountInfo
                             }
-                            axisMode={this.props.store.userSelectedAxisMode}
+                            axisMode={
+                                this.props.store.userSelectedAxisMode ||
```
alisman, line 128: sorry to be annoying but i think we should probably put this default inside of the the userSelectedAxisMode.

And also, I think we can just call it ```axisMode``` since I don't think we strongly need to distinguish it from a default or something else.

alisman (approved):

PR #4444, opened 2022-12-08
Format and fix gene selector and tab contents in group comparison lollipop plot

gblaih:
Fixes issues in https://github.com/cBioPortal/cbioportal/issues/9901

- Format gene selector and tabs so they are on the same line
- Fix gene tabs so they reflect highest mutation frequencies only (rather than mutation+sv+cna)
- Edit gene selector so it only shows genes with mutations and sort by frequency

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
@@ -47,14 +47,8 @@ export default class GroupComparisonMutationsTab extends React.Component<
 
     @computed get tabs() {
         return this.props.store.genesWithHighestMutationFrequency
-            .slice(0, 10)
-            .map(g => (
-                <MSKTab
-                    key={g.hugoGeneSymbol}
-                    id={g.hugoGeneSymbol}
-                    linkText={g.hugoGeneSymbol}
-                />
-            ));
+            .result!.slice(0, 10)
+            .map(g => <MSKTab key={g} id={g} linkText={g} />);
     }
 
     @computed get activeTabId(): string | undefined {
@@ -85,6 +79,7 @@ export default class GroupComparisonMutationsTab extends React.Component<
         await: () => [
             this.props.store.genes,
             this.props.store.genesWithMutations,
+            this.props.store.genesWithHighestMutationFrequency,
         ],
         render: () => {
             // We only want 2 groups for our mirrored lollipop plot. Display message if not 2 groups
@@ -128,20 +123,10 @@ export default class GroupComparisonMutationsTab extends React.Component<
                         store={this.props.store}
                         only="sample"
                     />
-                    {/* <LollipopGeneSelector
-                        store={this.props.store}
-                        genes={this.props.store.genes.result!}
-                        handleGeneChange={this.handleGeneChange}
-                        key={
-                            'comparisonLollipopGene' +
-                            this.props.store.activeMutationMapperGene!
-                                .hugoGeneSymbol
-                        }
-                    /> */}
                     <div style={{ display: 'flex', alignItems: 'center' }}>
                         <LollipopGeneSelector
                             store={this.props.store}
-                            genes={this.props.store.availableGenes.result!}
+                            genes={this.props.store.genes.result!}
                             genesWithMutations={
                                 this.props.store.genesWithMutations.result!
                             }
--- a/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationsTab.tsx
@@ -46,13 +46,15 @@ export default class GroupComparisonMutationsTab extends React.Component<
     }
 
     @computed get tabs() {
-        return this.props.store.genesWithMaxFrequency.map(g => (
-            <MSKTab
-                key={g.hugoGeneSymbol}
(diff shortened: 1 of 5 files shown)


src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -41,13 +41,9 @@ export default class GroupComparisonMutationsTab extends React.Component<
...
-        return this.props.store.genesWithMaxFrequency.map(g => (
-            <MSKTab
-                key={g.hugoGeneSymbol}
-                id={g.hugoGeneSymbol}
-                linkText={g.hugoGeneSymbol}
-            />
-        ));
+        return this.props.store.genesWithHighestMutationFrequency
+            .result!.slice(0, 10)
+            .map(g => <MSKTab key={g} id={g} linkText={g} />);
```
alisman, line 46: lets put prefix on both the key and id props.  e.g.  `comp-tab-${g}`
React keys need to be unique for whole page so best to prefix them in general

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -454,6 +454,16 @@ export default class GroupComparisonStore extends ComparisonStore {
         },
     });
 
+    readonly genesWithMutations = remoteData<Gene[]>({
+        await: () => [this.genesWithHighestMutationFrequency],
+        invoke: async () => {
+            const genes = await fetchGenes(
```
alisman, line 460: seems odd that we are fetching these genes here when we already fetching all genes right above in availableGenes.  could we not just filter that here?

alisman: also, seems strange that genesWithMutations is produced from genesWithHighestMutationFrequency.  That would imply the it wasn't just "highest" but all those with ANY mutation frequency, perhaps sorted by frequency?

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
gblaih: added

src/pages/groupComparison/GroupComparisonStore.ts
gblaih: yea we don't need to fetch genes here since we already have availableGenes, fixed. and yea genesWithHighestMutationFrequency is not the best name, changed it to genesSortedByMutationFrequency.

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -41,18 +41,20 @@ export default class GroupComparisonMutationsTab extends React.Component<
...
-                id={g.hugoGeneSymbol}
-                linkText={g.hugoGeneSymbol}
-            />
-        ));
+        return this.props.store.genesSortedByMutationFrequency
+            .result!.slice(0, 10)
+            .map(g => (
+                <MSKTab
+                    key={`comp-tab-${g}`}
+                    id={`comp-tab-${g}`}
```
alisman, line 54: ok, i misled you here.  key should be prefixed but not the id property. that will simplify the selected gene thing above.  sorry

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -63,7 +65,7 @@ export default class GroupComparisonMutationsTab extends React.Component<
             activeTabId = this.props.store.activeMutationMapperGene!
                 .hugoGeneSymbol;
         }
-        return activeTabId;
+        return `comp-tab-${activeTabId}`;
```
alisman, line 68: and it will simplify this

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
gblaih: fixed

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
gblaih: fixed

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1012,6 +1012,59 @@ export default abstract class ComparisonStore extends AnalysisStore
...
+    readonly mutationsEnrichmentDataRequestGroups = remoteData({
+        await: () => [
+            this.alterationsEnrichmentAnalysisGroups,
+            this.selectedStudyMutationEnrichmentProfileMap,
+        ],
+        invoke: () => {
+            if (
+                _(this.selectedMutationEnrichmentEventTypes)
+                    .values()
+                    .some()
```
alisman, line 1024: i haven't seen some used like this (without any callback).  it means that there are items in values array?  could we just use length then?

gblaih: this was mostly copied from the alterationsEnrichmentDataRequestGroups and thats what it used. but yes I think it just check if there are items in the values array and we can use length

gblaih: this was mostly copied from alterationsEnrichmentDataRequestGroups and I believe it checks that there is at least one selected event type, or the values array contains a true boolean

PR #4449, opened 2022-12-12
Fix axis label rounding issue in group comparison lollipop

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9930

### DIFF
--- a/packages/react-mutation-mapper/src/util/LollipopPlotUtils.ts
+++ b/packages/react-mutation-mapper/src/util/LollipopPlotUtils.ts
@@ -89,12 +89,18 @@ export function calcYMaxInput(
 
     if (input === undefined) {
         input = yAxisSameScale
-            ? getCommonYAxisMaxSliderValue(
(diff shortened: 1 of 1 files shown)


packages/react-mutation-mapper/src/util/LollipopPlotUtils.ts
```diff
@@ -89,12 +89,18 @@ export function calcYMaxInput(
...
-            ? getCommonYAxisMaxSliderValue(
-                  yMaxStep,
-                  countRange,
-                  oppositeCountRange
+            ? parseFloat(
+                  getCommonYAxisMaxSliderValue(
+                      yMaxStep,
+                      countRange,
+                      oppositeCountRange
+                  ).toFixed(Math.log10(yMaxStep) * -1)
```
alisman, line 97: seems like this would pass a negative number to toFixed(), which seems to be against the rule no?

alisman (approved):

PR #4454, opened 2022-12-14
Use profiled patients count in axis labels and tooltips in comparison view lollipop

gblaih:
Edits axis labels and lollipop tooltips so it uses and shows profiled patient counts instead of profiled sample counts

### DIFF
--- a/packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
+++ b/packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
@@ -146,17 +146,23 @@ export default class SVGAxis extends React.Component<SVGAxisProps, {}> {
                         : horizontalPadding);
                 transform = '';
             }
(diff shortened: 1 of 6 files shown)


alisman (approved):

packages/cbioportal-frontend-commons/src/components/SVGAxis.tsx
```diff
@@ -146,6 +146,17 @@ export default class SVGAxis extends React.Component<SVGAxisProps, {}> {
                         : horizontalPadding);
                 transform = '';
             }
+            let wrappedLabel = this.props.label.split(/\n/g).map(l =>
+                l.includes('**') ? (
+                    <tspan x={x} dy="1em" fontWeight={'bold'}>
+                        {l.replace(/\*{2}/g, '')}
```
alisman, line 152: use replacer to actually target position of markdown with the spans

alisman (approved):

PR #4457, opened 2022-12-19
gene panel charts in portal

Luke-Sikina:


src/pages/patientView/copyNumberAlterations/column/CohortColumnFormatter.spec.tsx
```diff
@@ -13,7 +13,7 @@ import CopyNumberCountCache from '../../clinicalInformation/CopyNumberCountCache
 import { CacheData } from '../../clinicalInformation/SampleGeneCache';
 
 describe('CohortColumnFormatter', () => {
-    const copyNumberData: DiscreteCopyNumberData[] = [
+    const copyNumberData: any[] = [
```
alisman, line 16: how come we had to get rid of type?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3188,6 +3225,13 @@ export class StudyViewPageStore
     private isChartFiltered(chartUniqueKey: string): boolean {
         switch (this.chartsType.get(chartUniqueKey)) {
             case ChartTypeEnum.PIE_CHART:
+                if (isGenePanelChart(chartUniqueKey)) {
+                    return (
+                        this.getGenePanelFilters({
+                            uniqueKey: chartUniqueKey,
+                        } as any).length > 0
```
alisman, line 3232: can we do better than "any"?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3564,6 +3608,23 @@ export class StudyViewPageStore
             filters.patientTreatmentTargetFilters = this.patientTreatmentTargetFilters;
         }
 
+        const genePanelFilters: GenePanelFilter[] = Object.keys(
```
alisman, line 3611: you can use ```_.forEach( (val, key)=>{...})``` 
a little cleaner

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3564,6 +3608,23 @@ export class StudyViewPageStore
...
+        )
+            .map(profile => {
+                return this._genePanelFilters[profile].map(panel => {
+                    return {
+                        genePanelId: panel,
+                        molecularProfileSuffix: chartMetaToProfile(profile),
+                    } as GenePanelFilter;
+                });
+            })
+            .reduce((acc, curVal) => acc.concat(curVal), []);
```
alisman, line 3622: you just call .flat() instead of this reduce.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -558,6 +562,38 @@ export class StudyViewPageStore
             )
         );
 
+        // Add special gene panel charts for each molecular profile
```
alisman, line 565: lets discuss necessity of this reaction.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3966,6 +4027,41 @@ export class StudyViewPageStore
...
+    public getGenePanelDataCount(
+        chartMeta: ChartMeta
+    ): MobxPromise<ClinicalDataCountSummary[]> {
+        const suffix = chartMetaToProfile(chartMeta);
+
+        return remoteData<ClinicalDataCountSummary[]>({
+            await: () => [this.genePanelCounts],
+            invoke: async () => {
+                const unfCounts = this.genePanelCounts.result!;
+                const countsForProfile = unfCounts
```
alisman, line 4039: can we improve the name unfCounts?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3966,6 +4027,41 @@ export class StudyViewPageStore
...
+                        return item.molecularProfileId.endsWith(suffix);
+                    })
+                    .map(i => {
+                        return {
+                            attributeId: i.molecularProfileId,
+                            counts: i.counts,
+                        };
+                    });
+
+                let data = countsForProfile[0];
```
alisman, line 4050: these assumes there will always be a matching profile.  is that founded?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3966,6 +4027,41 @@ export class StudyViewPageStore
...
+                let counts: ClinicalDataCount[] = [];
+                let attributeId: string = '';
+                if (data !== undefined) {
+                    counts = data.counts;
+                    attributeId = data.attributeId;
+                    if (!this.chartToUsedColors.has(attributeId))
+                        this.chartToUsedColors.set(attributeId, new Set());
+                }
+                return this.addColorToCategories(counts, attributeId);
```
alisman, line 4059: seems odd to call this empty values (data===undefined)

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3966,6 +4027,41 @@ export class StudyViewPageStore
...
+                    });
+
+                let data = countsForProfile[0];
+                let counts: ClinicalDataCount[] = [];
+                let attributeId: string = '';
+                if (data !== undefined) {
+                    counts = data.counts;
+                    attributeId = data.attributeId;
+                    if (!this.chartToUsedColors.has(attributeId))
+                        this.chartToUsedColors.set(attributeId, new Set());
```
alisman, line 4057: we shouldn't update state from inside of invoke method like this.  generally invoke only returns a value, which is used by refererer.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9792,4 +9925,33 @@ export class StudyViewPageStore
             this.studyIds[0] === 'heme_onc_nsclc_genie_bpc'
         );
     }
+
+    @action
```
alisman, line 9929: seems odd that we would be adding these general methods as part of this PR.  lets discuss

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -603,6 +603,24 @@ export function makeDensityScatterPlotTooltip(
...
+
+export function getGenePanelChartUniqueKey(profileId: string) {
+    return GENE_PANEL_PREFIX + profileId;
+}
+
+export function isGenePanelChart(uniqueKey: string) {
+    return uniqueKey.startsWith(GENE_PANEL_PREFIX);
+}
+
+export function chartUniqueKeyToProfile(meta: ChartMeta | string) {
```
alisman, line 616: can we just have it accept string and then deference that string at point of call when necessary?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -3966,6 +4027,41 @@ export class StudyViewPageStore
         this.changeChartVisibility(uniqueKey, false);
     }
 
+    public getGenePanelDataCount(
+        chartMeta: ChartMeta
+    ): MobxPromise<ClinicalDataCountSummary[]> {
+        const suffix = chartMetaToProfile(chartMeta);
+
+        return remoteData<ClinicalDataCountSummary[]>({
```
alisman, line 4035: this is gonna produce a new remoteData every time the method is called, which is a potential memory leak.  also, it isn't actually fetching anything remote (that's done by genePanelCounts.  We need to cache this promise (see getGenomicChartDataBin)

src/pages/studyView/StudyViewPageStore.ts
alisman, line 4059: maybe don't call this with empty stuff

src/pages/studyView/StudyViewPageStore.ts
alisman: is this used?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9792,4 +9925,33 @@ export class StudyViewPageStore
...
         );
     }
+
+    @action
+    resetFilterAndToggleChartVisibility(chartUniqueKey: string): void {
+        const visible = this._chartVisibility.has(chartUniqueKey);
+        this.resetFilterAndChangeChartVisibility(chartUniqueKey, !visible);
+    }
+
+    private setCustomChart(chartMeta: ChartMetaWithDimensionAndChartType) {
```
alisman, line 9935: name speciofic to gene panel

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -210,9 +220,37 @@ export class MultiSelectionTable extends React.Component<
                 render: (data: MultiSelectionTableRow) => {
                     return (
                         <div className={styles.labelContent}>
-                            <EllipsisTextTooltip
-                                text={data.label}
-                            ></EllipsisTextTooltip>
+                            <EllipsisTextTooltip text={data.label} />
```
alisman, line 223: see if we can figure out a way to keep this class generic (not have knowledge of genepanelness

PR #4460, opened 2022-12-21
Prevent clinical columns from disappearing

Rajat-Sirohi:
Fix cBioPortal/cbioportal#9869

The issue seems to have been introduced by commit 31876041e087a7a3741883fd88852b778a60f769. I'm not sure what fixes that commit was meant to implement, but effectively reverting that commit by removing the reset call to `this.props.storeColumnVisibility(undefined)` within the results view mutation table's `resetColumnVisibility()` function seems to fix the issue linked at the top, and when I play around with the column visibility, I don't see any new issues arising.

### DIFF
--- a/src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
+++ b/src/pages/resultsView/mutation/ResultsViewMutationTable.tsx
@@ -83,41 +83,27 @@ export default class ResultsViewMutationTable extends MutationTable<
         super(props);
         this.props.columnVisibilityProps!.customDropdown = (
             columnVisibilityControlsProps: IColumnVisibilityControlsProps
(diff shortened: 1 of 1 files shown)


onursumer:
@Rajat-Sirohi The commit https://github.com/cBioPortal/cbioportal-frontend/commit/31876041e087a7a3741883fd88852b778a60f769 was introduced as part of cBioPortal/cbioportal-frontend#4090 to address [this issue](https://github.com/cBioPortal/cbioportal-frontend/pull/4090#pullrequestreview-838697683).

Looks like clinical columns still disappear when you toggle OncoKB icon style. We should probably fix that behavior as well as part of this PR.

PR #4464, opened 2022-12-27
Add abstract reference linkouts in OncoKB Card

calvinlu3:
When hovering over the oncokb annotation icon and on the `Biological Effect` tab, the summaries currently do not have the abstract references parsed(only PMID and NCT). 

### Before changes
image

### After changes
image

This fixes https://github.com/oncokb/oncokb/issues/3350

### DIFF
--- a/packages/oncokb-frontend-commons/src/components/RefComponent.tsx
+++ b/packages/oncokb-frontend-commons/src/components/RefComponent.tsx
@@ -11,46 +11,75 @@ export default class RefComponent extends React.Component<{
     componentType: 'tooltip' | 'linkout';
 }> {
     render() {
-        const parts = this.props.content.split(/pmid|nct/i);
+        const parts = this.props.content.split(/pmid|nct|abstract/i);
 
         if (parts.length < 2) {
             return <span>{this.props.content}</span>;
         }
 
+        // Slicing from index 1 to end to rejoin the abstract title and link
+        // when there's the case that they also contain the string 'Abstract'
+        // Example :
+        //     (Abstract: Fakih et al. Abstract# 3003, ASCO 2019. https://meetinglibrary.asco.org/record/12411/Abstract)
+        const abstractParts = parts
+            .slice(1)
+            .join('Abstract')
+            .split(/(?=http)/i);
+        const isAbstract = !(abstractParts.length < 2);
+
+        let abstract = '';
+        let abstractLink = '';
         const ids = parts[1].match(/[0-9]+/g);
-
-        if (!ids) {
-            return <span>{this.props.content}</span>;
-        }
-
         let prefix: string | undefined;
-
-        if (this.props.content.toLowerCase().indexOf('pmid') >= 0) {
-            prefix = 'PMID: ';
-        } else if (this.props.content.toLowerCase().indexOf('nct') >= 0) {
-            prefix = 'NCT';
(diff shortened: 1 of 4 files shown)


zhx828 (approved):
LGTM

packages/oncokb-frontend-commons/src/components/RefComponent.tsx
```diff
@@ -11,46 +11,75 @@ export default class RefComponent extends React.Component<{
...
+            if (prefix) {
+                link = (
+                    <a
+                        target="_blank"
+                        rel="noopener noreferrer"
+                        href={getNCBIlink(`/pubmed/${ids.join(',')}`)}
+                    >
+                        {`${prefix}${ids.join(', ')}`}
+                    </a>
+                );
```
onursumer, line 69: This is partially duplicate code now since we have a similar code segment in the `if (isAbstract)` block.

packages/oncokb-frontend-commons/src/components/RefComponent.tsx
```diff
@@ -11,46 +11,75 @@ export default class RefComponent extends React.Component<{
...
+        // when there's the case that they also contain the string 'Abstract'
+        // Example :
+        //     (Abstract: Fakih et al. Abstract# 3003, ASCO 2019. https://meetinglibrary.asco.org/record/12411/Abstract)
+        const abstractParts = parts
+            .slice(1)
+            .join('Abstract')
+            .split(/(?=http)/i);
+        const isAbstract = !(abstractParts.length < 2);
 
+        let abstract = '';
```
onursumer, line 30: `abstract` is a reserved keyword. I am surprised compiler is okay with this.

calvinlu3: Yeah compiler didn't give any warnings probably because `abstract` is [not reserved anymore](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Lexical_grammar#reserved_words)?

Also, refactoring the code based on the above comment allowed me to remove the `abstract` and `abstractLink` variables.

packages/oncokb-frontend-commons/src/components/RefComponent.tsx
calvinlu3: Made some changes to move the code outside of the `if (isAbstract) ... else ...`.

packages/oncokb-frontend-commons/src/components/RefComponent.tsx
```diff
@@ -11,46 +11,64 @@ export default class RefComponent extends React.Component<{
...
     render() {
-        const parts = this.props.content.split(/pmid|nct/i);
+        const parts = this.props.content.split(/pmid|nct|abstract/i);
 
         if (parts.length < 2) {
             return <span>{this.props.content}</span>;
         }
 
-        const ids = parts[1].match(/[0-9]+/g);
+        // Slicing from index 1 to end to rejoin the abstract title and link
```
alisman, line 20: to be safe i would surround all of this parsing logic in try/catch just to avoid crashing in some difficult to forsee edgecase.   I would also extract this parsing/composition into a nicely named function and provide in the comment an example of input AND output.   this should probably have a unit test too.

alisman (approved):

PR #4467, opened 2023-01-04
#9945 Mutational Signature Dependency

TJMKuijpers:

- In the original frontend code, the mutational signature version would always default to version 2. This causes an error when there is only version 3 data. 
- The following four points are implemented to remove this dependency on version 2:

1.      retrieveMutationalSignatureVersionFromData() in MutationalSignaturesUtils.tsx: this function check which versions are available based on the MolecularProfileID and returns a string (which can be used for )
2.     When both version 3 and version 2 mutational signatures are available, priority is given to version 3.
3.     The dropdown menu which allows us to switch between versions is now depending on your mutational signature data available for the selected patient (and not for the overall study). This allows some patients to have only one type of data. (changes made in MutationalSignaturesContainer.tsx to the function availableVersions())
4.    Version 3 mutational signature data can also be subset to ID, SBS, and DBS files. If generic assays are loaded with _SBS, _DBS, or _ID suffix (instead of V3), it is possible to display those results in the mutational signature tab.

Unit test has been added to test the function (if v# is present -> should return v#). Functionality can also be tested with the lgg_ucsf_2014_test_generic_assay study in local/studies (data loading with and without a specific mutational signature version file present).

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -189,6 +189,7 @@ import {
     MutationalSignaturesVersion,
     MutationalSignatureStableIdKeyWord,
     validateMutationalSignatureRawData,
+    retrieveMutationalSignatureVersionFromData,
 } from 'shared/lib/GenericAssayUtils/MutationalSignaturesUtils';
 import { getServerConfig } from 'config/config';
 import { StructuralVariantFilter } from 'cbioportal-ts-api-client';
@@ -516,6 +517,7 @@ export class PatientViewPageStore {
                             });
                         }
                     );
+
                     const genericAssayRawData = await client.fetchGenericAssayDataInMultipleMolecularProfilesUsingPOST(
                         {
                             genericAssayDataMultipleStudyFilter: {
@@ -718,11 +720,14 @@ export class PatientViewPageStore {
     readonly hasMutationalSignatureData = remoteData({
         await: () => [this.fetchAllMutationalSignatureData],
         invoke: async () => {
+            var profile_version=retrieveMutationalSignatureVersionFromData(this.fetchAllMutationalSignatureData.result.map(profile=>profile.molecularProfileId))
+            this.setMutationalSignaturesVersion(profile_version)
             return Promise.resolve(
                 this.fetchAllMutationalSignatureData.result &&
                     this.fetchAllMutationalSignatureData.result.length > 0
             );
         },
+
     });
 
     // set version 2 of the mutational signature as default
--- a/src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
+++ b/src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
@@ -34,8 +34,12 @@ export default class MutationalSignaturesContainer extends React.Component<
         // split the id by "_", the last part is the version info
         // we know split will always have results
         // use uniq function to get all unique versions
-        return _.chain(this.props.profiles)
-            .map(profile => _.last(profile.molecularProfileId.split('_'))!)
+
+        // not all patients have the newest mutational signatures --> check if data is present before giving the options
+        var possible_options= this.props.profiles
+            .map(profile => _.last(profile.molecularProfileId.split('_'))!);
+        possible_options=possible_options.filter(item => Object.keys(this.props.data).includes(item));
+        return _.chain(possible_options)
             .uniq()
             .value();
     }
--- a/src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
+++ b/src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
@@ -8,6 +8,9 @@ import { GenericAssayTypeConstants } from 'shared/lib/GenericAssayUtils/GenericA
 export enum MutationalSignaturesVersion {
     V2 = 'v2',
     V3 = 'v3',
+    SBS ='SBS',
+    DBS='DBS',
+    ID='ID'
 }
 
 export enum MutationalSignatureStableIdKeyWord {
@@ -162,4 +165,21 @@ export function validateMutationalSignatureRawData(
 
     // we are expecting contribution and pvalue profiles are in pairs
     return _.every(profileIdsGroupByVersion, ids => ids.length === 2);
+}
+
+export function retrieveMutationalSignatureVersionFromData(SignatureProfiles: any[]){
+    var value_to_set:string='v2';
+    var unique_profile_version=_.uniq(SignatureProfiles.map(function(obj){
+        return _.last(obj.split('_'))}))
+    if(unique_profile_version!==undefined) {
+        // give priority to version 3 over version 2
+        if(unique_profile_version.indexOf('v3')!== -1  && unique_profile_version.indexOf('v2') !==-1){
+            value_to_set=Object.values(MutationalSignaturesVersion)[unique_profile_version.indexOf('v3')]
+        }
+        else {
+            var version_to_set = Object.values(MutationalSignaturesVersion).indexOf(unique_profile_version[0] as unknown as MutationalSignaturesVersion);
+            value_to_set = Object.values(MutationalSignaturesVersion)[version_to_set];
+            }
+        }
+    return value_to_set
 }
--- a/src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.spec.ts
+++ b/src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.spec.ts
@@ -1,6 +1,6 @@
 import { assert } from 'chai';
 import { GenericAssayData } from 'cbioportal-ts-api-client';
-import { validateMutationalSignatureRawData } from './MutationalSignaturesUtils';
+import { validateMutationalSignatureRawData,retrieveMutationalSignatureVersionFromData } from './MutationalSignaturesUtils';
 
 describe('MutationalSignaturesUtils', () => {
     describe('validateMutationalSignatureRawData()', () => {
@@ -124,4 +124,18 @@ describe('MutationalSignaturesUtils', () => {
             );
         });
     });
-});
+    describe('setMutationalSignatureVersion', () => {
+        it('Molecular profile Id _v2 sets MutationalSignature to .V2', () => {
+            var profile_test = ['study_test_v2']
+            assert.isTrue(retrieveMutationalSignatureVersionFromData(profile_test) == 'v2');
+        });
+        it('Molecular profile Id _v3 sets MutationalSignature to .V3', () => {
+            var profile_test = ['test_study_v3'];
+            assert.isTrue(retrieveMutationalSignatureVersionFromData(profile_test) == 'v3');
+        });
+        it('Studies with two version of mutational signatures is set to .V3', () => {
+            var profile_test = ['test_study_v2', 'test_study_v3']
+            assert.isTrue(retrieveMutationalSignatureVersionFromData(profile_test) == 'v3')
+        })
+    });
+})


src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
```diff
@@ -34,8 +34,12 @@ export default class MutationalSignaturesContainer extends React.Component<
         // split the id by "_", the last part is the version info
         // we know split will always have results
         // use uniq function to get all unique versions
-        return _.chain(this.props.profiles)
-            .map(profile => _.last(profile.molecularProfileId.split('_'))!)
+
+        // not all patients have the newest mutational signatures --> check if data is present before giving the options
+        var possible_options= this.props.profiles
```
dippindots, line 39: We are using camel case to name our variables e.g. `variableNameOne`, see  https://en.wikipedia.org/wiki/Camel_case for more details.

dippindots (changes requested):
@TJMKuijpers Welcome to cBioPortal community! And thanks for making a pull request to fix our current bug. I can see your approach would work as expected, which is great! Just the coding styles are different than our current. I made some comments below.

For coding styles, you can reference google's ts guide: https://google.github.io/styleguide/tsguide.html
You can also reference our documentation about `prettier` on how to format your code. https://github.com/cBioPortal/cbioportal-frontend#formatting-code-with-prettierjs
And feel free to pin me anytime if you have any questions.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -718,11 +720,14 @@ export class PatientViewPageStore {
     readonly hasMutationalSignatureData = remoteData({
         await: () => [this.fetchAllMutationalSignatureData],
         invoke: async () => {
+            var profile_version=retrieveMutationalSignatureVersionFromData(this.fetchAllMutationalSignatureData.result.map(profile=>profile.molecularProfileId))
+            this.setMutationalSignaturesVersion(profile_version)
```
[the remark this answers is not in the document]
TJMKuijpers: I have removed this from the readonly method and adjusted the selectedMutationalSignatureVersion() method to set the version at initial loading. The behavior of selectedMutationalSignatureVersion() is not changed when the user changes the version with the dropdown menu.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -718,11 +720,14 @@ export class PatientViewPageStore {
     readonly hasMutationalSignatureData = remoteData({
         await: () => [this.fetchAllMutationalSignatureData],
         invoke: async () => {
+            var profile_version=retrieveMutationalSignatureVersionFromData(this.fetchAllMutationalSignatureData.result.map(profile=>profile.molecularProfileId))
```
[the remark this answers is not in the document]
TJMKuijpers: Var has been changed to let (since the values are expected to change)

TJMKuijpers:
Dear @dippindots, thank you for the review and the comments. I have updated the files as suggested and formatted the code with  prettier.

dippindots:
@TJMKuijpers Thanks for the update! I will take a look again today.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -728,7 +729,21 @@ export class PatientViewPageStore {
     // set version 2 of the mutational signature as default
     @observable _selectedMutationalSignatureVersion: string =
         MutationalSignaturesVersion.V2;
+    // initialLoad to set _selectedMutationalSignatureVersion while loading the tab for the first time
+    @observable initialLoad: boolean = true;
```
dippindots, line 733: We can set `_selectedMutationalSignatureVersion` = undefined by default. and remove `initialLoad`, see more comments below.

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -728,7 +729,21 @@ export class PatientViewPageStore {
...
+        if (
+            this.initialLoad &&
+            this.fetchAllMutationalSignatureData.isComplete
+        ) {
+            let versionPresent = retrieveMutationalSignatureVersionFromData(
+                this.fetchAllMutationalSignatureData.result.map(
+                    profile => profile.molecularProfileId
+                )
+            );
+            this.setMutationalSignaturesVersion(versionPresent);
```
dippindots, line 744: We should avoid setting another observable variable in the `@computed` component. It's because a computed component should only capture the changes of the variables it's monitoring.
We can do this:

1. Change `_selectedMutationalSignatureVersion` and `initialLoad` as described in my previous comments
2. In this computed components, return `version present` if `_selectedMutationalSignatureVersion === undefined` or return `_selectedMutationalSignatureVersion`.
The second step's logic is when  `_selectedMutationalSignatureVersion` is undefined, we know its initial load, and we can select the default available version from `version present`, if `_selectedMutationalSignatureVersion` has value, then that means users already set the version manually, then we can just return `_selectedMutationalSignatureVersion`.

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -163,3 +166,35 @@ export function validateMutationalSignatureRawData(
     // we are expecting contribution and pvalue profiles are in pairs
     return _.every(profileIdsGroupByVersion, ids => ids.length === 2);
 }
+
+export function retrieveMutationalSignatureVersionFromData(
+    SignatureProfiles: any[]
```
dippindots, line 171: signatureProfiles

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -163,3 +166,35 @@ export function validateMutationalSignatureRawData(
...
+            uniqueProfileVersion.indexOf('v3') !== -1 &&
+            uniqueProfileVersion.indexOf('v2') !== -1
+        ) {
+            valueToSet = Object.values(MutationalSignaturesVersion)[
+                uniqueProfileVersion.indexOf('v3')
+            ];
+        } else {
+            let versionToSet = Object.values(
+                MutationalSignaturesVersion
+            ).indexOf(
```
dippindots, line 191: Maybe replace `indexOf` with `includes`?

dippindots (commented):
@TJMKuijpers Thanks again for updating the pull request, I just added more comments here. Sorry for being picky here, but I think it would be beneficial for us to take this opportunity and give you some cBioPortal coding preferences feedback.

src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
```diff
@@ -34,8 +34,14 @@ export default class MutationalSignaturesContainer extends React.Component<
...
         // we know split will always have results
         // use uniq function to get all unique versions
-        return _.chain(this.props.profiles)
-            .map(profile => _.last(profile.molecularProfileId.split('_'))!)
+        // not all patients have the newest mutational signatures --> check if data is present before giving the options
+        let possibleOptions = this.props.profiles.map(
+            profile => _.last(profile.molecularProfileId.split('_'))!
+        );
+        possibleOptions = possibleOptions.filter(item =>
+            Object.keys(this.props.data).includes(item)
```
alisman, line 42: could just be ```item in this.props.data```

src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
```diff
@@ -34,8 +34,14 @@ export default class MutationalSignaturesContainer extends React.Component<
         // split the id by "_", the last part is the version info
         // we know split will always have results
         // use uniq function to get all unique versions
-        return _.chain(this.props.profiles)
-            .map(profile => _.last(profile.molecularProfileId.split('_'))!)
+        // not all patients have the newest mutational signatures --> check if data is present before giving the options
+        let possibleOptions = this.props.profiles.map(
+            profile => _.last(profile.molecularProfileId.split('_'))!
+        );
+        possibleOptions = possibleOptions.filter(item =>
```
alisman, line 41: you can just chain filter on the map above since filter is native to js array

src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
```diff
@@ -34,8 +34,14 @@ export default class MutationalSignaturesContainer extends React.Component<
...
+            Object.keys(this.props.data).includes(item)
+        );
+        return _.chain(possibleOptions)
```
alisman, line 44: this can just be _.uniq(possibleOptions) and then you don't need _.chain or _.value.

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -163,3 +166,35 @@ export function validateMutationalSignatureRawData(
     // we are expecting contribution and pvalue profiles are in pairs
     return _.every(profileIdsGroupByVersion, ids => ids.length === 2);
 }
+
+export function retrieveMutationalSignatureVersionFromData(
+    SignatureProfiles: any[]
+) {
+    let valueToSet: string = 'v2';
```
alisman, line 173: use ```const``` unless you know for sure you are going to change the value later (unusual).

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
TJMKuijpers: `_selectedMutationalSignatureVersion` has been set to undefined (and `initialLoad` removed) in PatientViewPageStore

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
TJMKuijpers: `indexOf` changed to `includes`

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -163,3 +166,42 @@ export function validateMutationalSignatureRawData(
...
+        if (
+            uniqueProfileVersion.includes('v3') &&
+            uniqueProfileVersion.includes('v2')
+        ) {
+            const valueToSet: string = 'v3';
+            return valueToSet;
+        } else {
+            // Can we make this easier? (so just select the version present
+            // To also take into account when there are version outside the enumType)
+            const versionToSet = Object.values(
```
alisman, line 189:
```versionToSet = Object.values(MutationalSignaturesVersion).find(uniqueProfileVersion[0])
```
then you can test for undefined.

alisman: in fact i think this whole thing can be reduced to:

```
return Object.values(MutationalSignaturesVersion)
.find(uniqueProfileVersion[0]) ? uniqueProfileVersion[0]
```

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -163,3 +166,42 @@ export function validateMutationalSignatureRawData(
...
+            return _.last(obj.split('_'));
+        })
+    );
+    if (uniqueProfileVersion !== undefined) {
+        // give priority to version 3 over version 2
+        if (
+            uniqueProfileVersion.includes('v3') &&
+            uniqueProfileVersion.includes('v2')
+        ) {
+            const valueToSet: string = 'v3';
```
alisman, line 184: just 
```
return 'v3'
```

src/pages/patientView/mutationalSignatures/MutationalSignaturesContainer.tsx
```diff
@@ -34,10 +34,11 @@ export default class MutationalSignaturesContainer extends React.Component<
         // split the id by "_", the last part is the version info
         // we know split will always have results
         // use uniq function to get all unique versions
-        return _.chain(this.props.profiles)
+        // not all patients have the newest mutational signatures --> check if data is present before giving the options
+        const possibleOptions = this.props.profiles
```
alisman, line 38: the original code (using lodash chaining is easier to read).  you should try to retain that style.

TJMKuijpers: Retained the style and rewrote the lines of code to:
`_.chain(this.props.profiles)
            .map(profile => _.last(profile.molecularProfileId.split('_'))!).filter(item => item in this.props.data).uniq().value();`

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
TJMKuijpers: I've reduced the code and removed the if/else statements. I believe it can be reduced even more if we don't force the version to be a member of the enum MutationalSignaturesVersion. This could be beneficial when a version exists that is not yet part of MutationalSignaturesVersion. Or, when users use _early or _late to distinguish between mutational signature profiles (and not _v2 or _v3)

The code becomes:
`return (uniqueProfileVersion.includes('v3') && uniqueProfileVersion.includes('v2')) ? 'v3' : uniqueProfileVersion[0]!`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
alisman: @TJMKuijpers this needs to be addressed

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
alisman: @TJMKuijpers @pvannierop can you comment on the above?

TJMKuijpers: In the original code, the _selectedMutationalSignatureVersion is set to version 2 by:
`@observable _selectedMutationalSignatureVersion: string =MutationalSignaturesVersion.V2;`
When loading the mutational signature tab, the version is retrieved via:
```
       version={
                pageComponent.patientViewPageStore.selectedMutationalSignatureVersion
}
```
which will call :
```
		@computed get selectedMutationalSignatureVersion() {
        				return this._selectedMutationalSignatureVersion;
    			}
```
By default this will be v2, which causes a problem when version 2 data is not present (page will turn blank). 

To fix the dependency on version 2, we could check if the mutational signature version of a study belongs to enum MutationalSignatureVersion:

```
export enum MutationalSignaturesVersion {
    V2 = 'v2',
    V3 = 'v3',
}
```
However, this could be a problem if the molecularProfileId of our mutational signature data does not end with _v2 and v3 (because of the string split in availableVersions()). For example, we could have 3 mutational signature files with a molecularProfileId _SBS, _DBS, and _ID. To solve this problem, we could add SBS, DBS, and ID to MutationalSignaturesVersion. However, this means that when we have a molecularProfileId, not in MutationalSignaturesVersion, we have to update it to be able to load the mutational signature data.

By having a function that is independent of MutationalSignaturesVersion, we don't have to worry about molecularProfileIds that are not a member (or in this case, are v2 or v3).

`return (uniqueProfileVersion.includes('v3') && uniqueProfileVersion.includes('v2')) ? 'v3' : uniqueProfileVersion[0]!`
In this case, we give priority over version 3 if both versions 2 and 3 are present. Otherwise, we will return the first value of uniqueProfileVersion. This could be either v2, v3, or any other version (if we regard SBS, DBS, and ID COSMIC signatures as a version).

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -724,10 +725,25 @@ export class PatientViewPageStore {
...
+        default: 'v2',
+        onResult: version => {
+            this._selectedMutationalSignatureVersion = version;
+        },
+    });
 
-    // set version 2 of the mutational signature as default
-    @observable _selectedMutationalSignatureVersion: string =
-        MutationalSignaturesVersion.V2;
     @computed get selectedMutationalSignatureVersion() {
```
alisman, line 747: this can read, 
```
return this._selectedMutationalSignatureVersion || this.initialMutationalSignatureVersion.result)
```

TJMKuijpers: Replaced with:
```
@computed get selectedMutationalSignatureVersion() {
        return this._selectedMutationalSignatureVersion || this.initialMutationalSignatureVersion.result;
    }
```

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -724,10 +725,25 @@ export class PatientViewPageStore {
...
+            return Promise.resolve(
+                retrieveMutationalSignatureVersionFromData(
+                    this.fetchAllMutationalSignatureData.result.map(
+                        profile => profile.molecularProfileId
+                    )
+                )
+            );
+        },
+        default: 'v2',
+        onResult: version => {
```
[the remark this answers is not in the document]
TJMKuijpers: The onResult could indeed be removed. However, I believe `default: 'v2` is needed. If we remove this, the test on the PatientViewPageTabs will return an error because  `string | undefined` is not allowed for:
```
version={
              pageComponent.patientViewPageStore.selectedMutationalSignatureVersion
}
```
The test passes when we set `default:v2`

src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
```diff
@@ -724,12 +725,24 @@ export class PatientViewPageStore {
...
+        invoke: () => {
+            return Promise.resolve(
+                retrieveMutationalSignatureVersionFromData(
+                    this.fetchAllMutationalSignatureData.result.map(
+                        profile => profile.molecularProfileId
+                    )
+                )
+            );
+        },
+        default: 'v2',
```
alisman, line 741: we could get rid of this default.   if typescript complains you can use exaimation point at the end to force typescript to ignore the possibility of undefined.  
```
this.initialMutationalSignatureVersion.result!
```

pvannierop:
[Remove dependency on COSMIC V2 data in profile data](https://trello.com/c/0qwdv3xf/3-remove-dependency-on-cosmic-v2-data-in-profile-data)

TJMKuijpers:
The test file patientview.screenshot.spec.js is updated because we now start with the mutational table showing version 3 by default (and not version 2). The click event changing the version does change version 3 to version 2.

PR #4469, opened 2023-01-05
Show counts separately for each group and fix resize issue in comparison lollipop

gblaih:
Adds features in https://github.com/cBioPortal/cbioportal/issues/9901

- Mirror the counts, showing counts separately for each group
- Edit y-axis max sliders and percent toggle format to fix resize issue
- Add ordinals (A, B) to label groups more clearly


src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,14 +51,101 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+    }
+
+    protected mutationCountsByProteinImpactTypeForGroup(
+        groupIndex: number
+    ): {
+        [proteinImpactType: string]: number;
+    } {
+        const map: { [proteinImpactType: string]: number } = {};
+
+        Object.keys(
```
alisman, line 141: use ```_.forIn(this.mutationsGroupedByProteinImpactTypeForGroup(groupIndex),(v,k)=>{...} ```

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,14 +51,101 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+        // apply filters excluding the protein impact type filters
+        // this prevents number of unchecked protein impact types from being counted as zero
+        let sortedFilteredGroupedData = applyDataFilters(
+            this.store.dataStore.sortedFilteredGroupedData[groupIndex].data,
+            filtersWithoutProteinImpactTypeFilter,
+            this.store.dataStore.applyFilter
         );
+
+        // also apply lazy mobx table search filter
+        sortedFilteredGroupedData = sortedFilteredGroupedData.filter(m =>
```
alisman, line 124: is this necessary in this feature, since there is no filtering?

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,14 +51,101 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                        annotatedProteinImpactTypeFilter={
+                            this.annotatedProteinImpactTypeFilter
+                        }
+                    />
+                </div>
+            </div>
+        );
+    }
+
+    protected mutationsGroupedByProteinImpactTypeForGroup(groupIndex: number) {
```
alisman, line 107: if at all possible we should pull this logic apart.   you must already by filtering by drivers somewhere else right?  so, could we not apply this impact type filtering on the result of that?  instead of applying again?

alisman: i.e. the two other types of filtering we are applying inside this method are not implied by the methods name so it would be easy to misunderstand what it's doing.

onursumer: This is mostly duplicate of https://github.com/cBioPortal/cbioportal-frontend/blob/634421d48da5156540b77993d011a194132339d3/src/shared/components/mutationMapper/MutationMapper.tsx#L466

Actually, we need to apply filters here again by excluding the protein impact type filters. Otherwise we lose the counts for unchecked types since they're filtered out: https://github.com/cBioPortal/cbioportal-frontend/blob/634421d48da5156540b77993d011a194132339d3/src/shared/components/mutationMapper/MutationMapper.tsx#L474-L475

So, we need this additional filtering logic here to dynamically calculate the numbers (inside the boxes) to display for each type. I think we should separate that logic into a utility function or something instead of duplicating the existing code in each new implementation.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
onursumer: not necessary unless we decide to enable the table later

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
onursumer: Actually it's not working properly for this implementation. When a type is deselected, its count immediately drops to zero.

image

We should keep the original count just like in the current production

image

packages/react-mutation-mapper/src/component/lollipopMutationPlot/LollipopMutationPlotControls.tsx
```diff
@@ -215,6 +220,15 @@ export default class LollipopMutationPlotControls extends React.Component<
         );
     }
 
+    protected get percentToggle() {
```
alisman, line 223: this will have negligible benefit in this case, but it's good to make a getter that returns jsx @computed to avoid unnecessary recomputation.

gblaih: yea, this is mostly to group the toggle with the sliders on the left when the filter reset panel is triggered as opposed to the legend and download controls

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,14 +50,127 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                            filter={this.proteinImpactTypeFilter}
+                            counts={this.mutationCountsByProteinImpactTypeForGroup(
+                                0
+                            )}
+                            onSelect={this.onProteinImpactTypeSelect}
+                            onClickSettingMenu={this.props.onClickSettingMenu}
+                            annotatedProteinImpactTypeFilter={
+                                this.annotatedProteinImpactTypeFilter
+                            }
+                        />
```
onursumer, line 82: This should probably be a separate function or component which can also be used for the mirrored part to avoid duplication. I think the parameters should be the `group index` and the `short name`.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,14 +50,127 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                                    .sortedFilteredGroupedData[0].group
+                            }
+                        </div>
+                        <ProteinImpactTypeBadgeSelector
+                            filter={this.proteinImpactTypeFilter}
+                            counts={this.mutationCountsByProteinImpactTypeForGroup(
+                                0
+                            )}
+                            onSelect={this.onProteinImpactTypeSelect}
+                        />
```
onursumer, line 123: Similar to the comment above, this can be a function or component which takes group index and name as parameters

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
onursumer: This seems fixed now. Thanks!

onursumer (changes requested):

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
gblaih: Separated both to a new function proteinImpactTypeBadgeSelectorForGroup

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -479,6 +481,14 @@ export default class MutationMapper<
...
+        return sortedFilteredData;
+    }
+
+    @computed
+    protected get mutationsGroupedByProteinImpactType() {
+        let sortedFilteredData = this
+            .sortedFilteredDataWithoutProteinImpactTypeFilter;
+
         // also apply lazy mobx table search filter
         sortedFilteredData = sortedFilteredData.filter(m =>
```
onursumer, line 493: Might be better to define this as a `const`

```suggestion
        // also apply lazy mobx table search filter
        const sortedFilteredData = this
            .sortedFilteredDataWithoutProteinImpactTypeFilter.filter(m =>
```

onursumer (approved):
Looks good! Just added another minor comment.

src/shared/components/mutationMapper/MutationMapper.tsx
gblaih: Changed it to a const.

alisman (approved):

PR #4473, opened 2023-01-09
Eliminate cbioportal count and cosmic count from driver settings

alisman:


src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -509,34 +509,18 @@ export default class ResultsViewOncoprint extends React.Component<
             get annotateDriversHotspotsDisabled() {
                 return !getServerConfig().show_hotspot;
             },
+            get annotateDriversCosmicDisabled() {
+                return !getServerConfig().show_hotspot;
+            },
```
onursumer, line 514: Is this intentional? Feels like we shouldn't be referring to `show_hotspot` here.

src/shared/alterationFiltering/AnnotationFilteringSettings.ts
```diff
@@ -187,8 +177,6 @@ export function buildDriverAnnotationControlsHandlers(
                 )
                     driverAnnotationSettings.hotspots = true;
 
-                driverAnnotationSettings.cbioportalCount = true;
-                driverAnnotationSettings.cosmicCount = true;
```
alisman, line 191: @gblaih this is where the bug was.  if you clicked the parent checkbox "on" it would set cosmicCount to true and that's why the driver number would/could go up.  because it would start evaluating using cosmic service, which is now defunct.   so, solution was just to fully get rid of this cosmic code since we don't want to use that service anymore.  ditto cbioPortal count.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
alisman: Good catch @onursumer.  I had put this in to test something but forgot to get rid.  It wasn't referenced anywhere.

gblaih (approved):

PR #4474, opened 2023-01-10
Set OncoKB prop usingPublicOncoKbInstance default to false + modify README

scarrero4660:
This fixes [oncokb/oncokb#3359](https://github.com/oncokb/oncokb/issues/3359)


alisman:
@scarrero4660 @zhx828 i noticed this PR sets default to false where the issue requests default to be true?

scarrero4660:
@alisman @zhx828 Yes there was a typo in the original issue. We determined the default should be false

alisman (approved):

PR #4475, opened 2023-01-11
Move driver annotation settings button to mutation filter panel in comparison view lollipop

gblaih:
Moves driver annotation settings button to mutation filter panel above protein impact type badges and hides button that appears on hover next to driver badge.

Related issue https://github.com/cBioPortal/cbioportal/issues/9901


src/shared/components/driverAnnotations/SettingsMenuButton.tsx
```diff
@@ -105,28 +106,63 @@ export default class SettingsMenuButton extends React.Component<
...
-                    this.setVisibilityState(visible);
-                }}
-                onPopupAlign={tooltipEl => setArrowLeft(tooltipEl, '22px')}
-            >
-                <button
-                    data-test="GlobalSettingsButton"
-                    style={{
-                        marginRight: 0,
-                        marginLeft: 5,
+        if (this.props.inFilterPanel) {
```
alisman, line 109: what's difference between these two modes?  hard for me to see.

gblaih: `inFilterPanel` refers to if the button is in the filter panel. In the comparison view, the button is in the filter panel as opposed to the results view where the button is on the top of the page by the "Modify Query" button. If it is in the panel, the button's properties are set so it is more suitable for the filter panel, like setting its trigger to click, its tooltip placement is from the top, the icon is smaller, styling, etc.

alisman (approved):

PR #4476, opened 2023-01-11
Fix localdb test which is failing on retry

alisman:


end-to-end-test/local/specs/core/groupComparison.screenshot.spec.js
```diff
@@ -158,10 +159,13 @@ describe('group comparison page screenshot tests', function() {
         });
 
         it('group comparison alteration enrichments two groups', function() {
-            goToUrlAndSetLocalStorage(
-                `${browser.getUrl()}&unselectedGroups=%5B"GB"%2C"OAST"%2C"ODG"%5D`,
-                true
-            );
+            // this test will not work on retry because groups will be toggled back on
+            this.retries(0);
```
alisman, line 163: @inodb i'm pretty sure that the problem was that the screen shot test was failing (as it should have due to log->log2) but then test runner was retrying it, which was adding redundant url params which then broken the application.  the retry result was then getting reported instead of the initial screenshot failure.

end-to-end-test/local/specs/core/groupComparison.screenshot.spec.js
```diff
@@ -158,10 +159,13 @@ describe('group comparison page screenshot tests', function() {
...
+
+            getElementByTestHandle('groupSelectorButtonGB').click();
```
alisman, line 165: i did this to make it a little more realistic than just editing url to shut off the groups.

end-to-end-test/local/specs/core/groupComparison.screenshot.spec.js
inodb: ah i see - makes sense

PR #4477, opened 2023-01-11
Add missing react-bootstrap dependency for cbioportal-frontend-commons

onursumer:
This missing dependency might be causing the problem observed in [this PR](https://github.com/knowledgesystems/signal/pull/152#issuecomment-1324343685)


inodb (approved):
LGTM

PR #4481, opened 2023-01-13
Fix missing oncokb-frontend-commons dependency and styles

onursumer:
Related to https://github.com/knowledgesystems/signal/pull/152#issuecomment-1380941913

Seems like oncokb-frontend-commons relies on some styles inherited from the `cbioportal-frontend` css class. Outside the cBioPortal context these styles do not exist and results in partially broken tooltip.

### DIFF
--- a/packages/oncokb-frontend-commons/package.json
+++ b/packages/oncokb-frontend-commons/package.json
@@ -40,6 +40,7 @@
     "lodash": "^4.17.15",
     "oncokb-styles": "~1.4.2",
     "oncokb-ts-api-client": "^1.3.2",
+    "rc-tooltip": "^5.0.2",
     "react-bootstrap": "^0.31.5",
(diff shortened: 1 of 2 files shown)


zhx828:
How can we prevent similar issue?

onursumer:
> How can we prevent similar issue?

@zhx828 I think we need an isolated test environment or something like that. Not sure what's the best practice to catch styling issues.

alisman (approved):

leexgh:
Seems all cancer type is unknow in tooltip:
Screen Shot 2023-01-13 at 4 57 14 PM

onursumer:
> Not related to this issue, but seems all cancer type is unknow in tooltip: Screen Shot 2023-01-13 at 4 57 14 PM

Good catch! This issue might be specific to Signal since it already exists in the production. We should address this in a separate PR.

PR #4484, opened 2023-01-17
Add e2e tests for comparison view lollipop plot

gblaih:
Add e2e tests for comparison view lollipop plot that test alerts, tooltips, gene selection, annotation tracks, and driver/vus and protein badge selections

### DIFF
--- a/end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
+++ b/end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
@@ -1,0 +1,234 @@
+var assert = require('assert');
+var goToUrlAndSetLocalStorage = require('../../../shared/specUtils')
+    .goToUrlAndSetLocalStorage;
+var waitForNetworkQuiet = require('../../../shared/specUtils')
+    .waitForNetworkQuiet;
+var setInputText = require('../../../shared/specUtils').setInputText;
+var setSettingsMenuOpen = require('../../../shared/specUtils')
+    .setSettingsMenuOpen;
+const { getElementByTestHandle } = require('../../../shared/specUtils');
+
+const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, '');
+
+describe('group comparison mutations tab tests', function() {
+    describe('lollipop alerts and plot display', function() {
+        before(function() {
+            goToUrlAndSetLocalStorage(
+                `${CBIOPORTAL_URL}/comparison?comparisonId=634006c24dd45f2bc4c3d4aa`
+            );
+            $('div[data-test="ComparisonPageOverlapTabDiv"]').waitForExist();
+        });
+
+        it('too many groups alert displayed when more than 2 groups selected', function() {
+            $('a.tabAnchor_mutations').waitForExist();
+            $('a.tabAnchor_mutations').click();
+            browser.pause(1000);
+            assert.equal(
+                $('[data-test="TooManyGroupsAlert"]').isDisplayed(),
+                true
+            );
+        });
+
+        it('not enough groups alert displayed when less than 2 groups selected', function() {
+            $('a=Deselect all').waitForExist();
+            $('a=Deselect all').click();
+            browser.pause(1000);
+            assert.equal(
+                $('[data-test="NotEnoughGroupsAlert"]').isDisplayed(),
+                true
+            );
+            $(
+                '[data-test="groupSelectorButtonColon Adenocarcinoma"]'
+            ).waitForExist();
+            $('[data-test="groupSelectorButtonColon Adenocarcinoma"]').click();
+            browser.pause(1000);
+            assert.equal(
+                $('[data-test="NotEnoughGroupsAlert"]').isDisplayed(),
+                true
+            );
+        });
+
+        it('lollipop plot displayed when 2 groups selected', function() {
+            $(
+                '[data-test="groupSelectorButtonColorectal Adenocarcinoma"]'
+            ).waitForExist();
+            $(
+                '[data-test="groupSelectorButtonColorectal Adenocarcinoma"]'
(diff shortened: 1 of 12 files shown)


end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+            goToUrlAndSetLocalStorage(
+                `${CBIOPORTAL_URL}/comparison?comparisonId=634006c24dd45f2bc4c3d4aa`
+            );
+            $('div[data-test="ComparisonPageOverlapTabDiv"]').waitForExist();
+        });
+
+        it('too many groups alert displayed when more than 2 groups selected', function() {
+            $('a.tabAnchor_mutations').waitForExist();
+            $('a.tabAnchor_mutations').click();
+            browser.pause(1000);
```
alisman, line 25: we should avoid using pauses in order to minimize test duration.   instead, just do ```$(..).waitForDisplayed()```
Note that you **do not actually have to assert anything after this**.  If waitForDisplayed times out, the test will fail.   If it succeeds, the test will pass.

end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+    describe('lollipop alerts and plot display', function() {
+        before(function() {
+            goToUrlAndSetLocalStorage(
+                `${CBIOPORTAL_URL}/comparison?comparisonId=634006c24dd45f2bc4c3d4aa`
+            );
+            $('div[data-test="ComparisonPageOverlapTabDiv"]').waitForExist();
+        });
+
+        it('too many groups alert displayed when more than 2 groups selected', function() {
+            $('a.tabAnchor_mutations').waitForExist();
```
alisman, line 23: since all of the tests involve the mutation tab, just put in the url and then you don't have to do this clicking thing.   we don't really need to test the clicking of the tab button.

end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+            $('a.tabAnchor_mutations').click();
+            browser.pause(1000);
+            assert.equal(
+                $('[data-test="TooManyGroupsAlert"]').isDisplayed(),
+                true
+            );
+        });
+
+        it('not enough groups alert displayed when less than 2 groups selected', function() {
+            $('a=Deselect all').waitForExist();
```
alisman, line 33: you can probably assume that this button exists based on the wait in the before callback

end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+            $('a=Deselect all').click();
+            browser.pause(1000);
+            assert.equal(
+                $('[data-test="NotEnoughGroupsAlert"]').isDisplayed(),
```
alisman, line 37: we recently added a helper called getElementByTestHandle which just takes 'NotEnoughGroupsAlert' as an arg.

end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+                true
+            );
+            $(
```
alisman, line 40: i believe you can assume that this exists?

end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+    describe('lollipop tooltip display', function() {
+        it('displays double tooltip when lollipop is present in both plots at the same position', function() {
+            $('.lollipop-0').waitForExist();
+            $('.lollipop-0').moveTo();
+            $('[data-test="tooltip-1450-Colon Adenocarcinoma"]').waitForExist();
+            $(
+                '[data-test="tooltip-1450-Colorectal Adenocarcinoma"]'
+            ).waitForExist();
+            assert.equal(
```
alisman, line 71: instead of these big assertions, we can just do 
```$('[data-test="tooltip-1450-Colon Adenocarcinoma"]').waitForExist();```
which as explained above has the same effect as an assertion.   i know if feels weird not do an assertion in a test, but it's ok.

File: end-to-end-test/remote/specs/core/groupComparisonLollipop.spec.js [hunk 7]
```diff
@@ -0,0 +1,234 @@
...
+    });
+
+    describe('adding annotation tracks', function() {
+        it('track visibility stays on gene change', function() {
+            $('div.css-dlmlqy-control').click();
+            $('[data-test="CancerHotspots"]').click();
+            waitForNetworkQuiet();
+            $('svg.css-tj5bde-Svg').click();
+            waitForNetworkQuiet();
+            assert.equal(
```
alisman, line 153: just use waitForDisplayed on the element instead of waitForNetworkQuiet

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -136,7 +136,7 @@ export default class GroupComparisonMutationsTabPlot extends React.Component<
                 );
                 return (
                     <>
-                        <h3>
+                        <h3 data-test="LollipopPlotTitle">
```
alisman, line 139: there's probably an easy way to get this h3 without using data-test and if so, that's a better way to go.  we only want to use data-test when it's totally necessary.   here for example, some parent div already has a data-test attribute and you can then look for h3 inside of that.

alisman (approved):

PR #4485, opened 2023-01-19
Results view mutations tab should show filtered and total counts in filter alert

gblaih:
Fix issue https://github.com/cBioPortal/cbioportal/issues/9958

### DIFF
--- a/src/pages/resultsView/querySummary/PatientSampleSummary.tsx
+++ b/src/pages/resultsView/querySummary/PatientSampleSummary.tsx
@@ -1,0 +1,32 @@
+import * as React from 'react';
+import { SampleIdentifier, Patient, Sample } from 'cbioportal-ts-api-client';
+import _ from 'lodash';
+
+interface IPatientSampleSummaryProps {
+    samples: Sample[] | SampleIdentifier[];
+    patients: Patient[] | string[];
+}
+
+export const PatientSampleSummary: React.FC<IPatientSampleSummaryProps> = ({
+    samples,
+    patients,
+}: IPatientSampleSummaryProps) => {
+    if (samples.length !== patients.length) {
+        const patientUnits = patients.length === 1 ? 'patient' : 'patients';
+        const sampleUnits = samples.length === 1 ? 'sample' : 'samples';
+        return (
+            <span>
+                <strong>{samples.length}</strong> {sampleUnits} /{' '}
+                <strong>{patients.length}</strong> {patientUnits}
+            </span>
+        );
+    } else {
+        const units =
+            samples.length === 1 ? 'sample/patient' : 'samples/patients';
+        return (
+            <span>
+                <strong>{samples.length}</strong> {units}
+            </span>
+        );
+    }
+};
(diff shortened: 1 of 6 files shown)


src/shared/components/mutationMapper/MutationMapperDataStore.ts
```diff
@@ -196,6 +196,19 @@ export default class MutationMapperDataStore
...
+    get dataSamples() {
+        return _.uniqBy(_.flatten(this.allData), m => m.sampleId).map(m => ({
+            sampleId: m.sampleId,
+            studyId: m.studyId,
+        }));
+    }
+
+    @computed
+    get dataPatients() {
+        return _.uniq(_.flatten(this.allData).map(m => m.patientId));
```
alisman, line 209: since the word "patient" strongly suggests were' going to get the Patient type back, i would call this dataPatientIds.    The name "dataPatients" and "dataSamples" could probably be improved.  Like, we know it's data.  Lets make it more descriptive (even if there are other preexisting properties here which suffer from the same vagueness).

src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
```diff
@@ -10,13 +10,13 @@ export function getPatientSampleSummary(samples: any[], patients: any[]) {
         const sampleUnits = samples.length === 1 ? 'sample' : 'samples';
```
alisman, line 10: this is a comment about this whole function.   i do realize this is not your code.   but we could take opportunity to improve it.  1) probably this should be a component.     2) we should improve on the typing of the arguments.  anytime we see use of any, we should try to fix it.   in this case, patients looks like it's string[].   for samples, there is an existing type, SampleIdentifiers which i think will work here.

alisman (changes requested):
see comments

src/pages/resultsView/querySummary/QuerySummaryUtils.tsx
gblaih: sample and patients can be Sample[] or Patient[] respectively as well so I added those along with the SampleIdentifier[] and string[] types

src/shared/components/mutationMapper/MutationMapperDataStore.ts
gblaih: renamed dataPatients to dataPatientIds and dataSamples to dataSampleIdentifiers

src/pages/resultsView/querySummary/PatientSampleSummary.tsx
```diff
@@ -0,0 +1,32 @@
+import * as React from 'react';
+import { SampleIdentifier, Patient, Sample } from 'cbioportal-ts-api-client';
+import _ from 'lodash';
+
+interface IPatientSampleSummaryProps {
+    samples: Sample[] | SampleIdentifier[];
```
alisman, line 6: @gblaih i went ahead and merged but probably we should just pass in counts instead of this slightly odd signature.  But it's not important enough to bother.

PR #4488, opened 2023-01-23
Fire genome nexus variant annotation call earlier

onursumer:
Fix cBioPortal/cbioportal#9967

### DIFF
--- a/src/pages/resultsView/mutation/Mutations.tsx
+++ b/src/pages/resultsView/mutation/Mutations.tsx
@@ -89,6 +89,10 @@ export default class Mutations extends React.Component<
     public render() {
         const activeTabId = this.selectedGeneSymbol;
 
+        // this is to trigger genome nexus variant annotation call as soon as the Mutations tab starts rendering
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -89,6 +89,10 @@ export default class Mutations extends React.Component<
     public render() {
         const activeTabId = this.selectedGeneSymbol;
 
+        // this is to trigger genome nexus variant annotation call as soon as the Mutations tab starts rendering
+        const triggerForAnEarlierGenomeNexusCall = this.props.store
+            .indexedVariantAnnotations.result;
```
onursumer, line 94: @alisman it also works without assigning the result to a variable. Which one do you think is better? My IDE complains about either an unused or an unassigned variable.

```suggestion
        this.props.store.indexedVariantAnnotations.result;
```

alisman: @onursumer this is fine.

alisman (approved):

PR #4489, opened 2023-01-23
Fix issue with indexing/naming samples in timeline

alisman:
Fixes https://github.com/cBioPortal/cbioportal/issues/9962

### DIFF
--- a/src/pages/patientView/SampleManager.tsx
+++ b/src/pages/patientView/SampleManager.tsx
@@ -95,10 +95,11 @@ export function getSpecimenCollectionDayMap(
         // TODO: SAMPLE_ACQUISITION is specific to genie_bpc_test study. We
         // should probably have some config to allow people to choose what
         // timeline tracks get labels
-        const specimenEvents = events.filter(
-            (e: ClinicalEvent) =>
-                e.eventType === 'SPECIMEN' || 'SAMPLE_ACQUISITION'
-        );
+        const specimenEvents = events.filter((e: ClinicalEvent) => {
+            return /SPECIMEN|Sample Acquisition|sample_acquisition'/i.test(
+                e.eventType
+            );
+        });
 
         specimenEvents.forEach((event: ClinicalEvent) => {
             const sampleIdAttr = _.find(


src/pages/patientView/SampleManager.tsx
```diff
@@ -95,10 +95,11 @@ export function getSpecimenCollectionDayMap(
         // TODO: SAMPLE_ACQUISITION is specific to genie_bpc_test study. We
         // should probably have some config to allow people to choose what
         // timeline tracks get labels
-        const specimenEvents = events.filter(
-            (e: ClinicalEvent) =>
-                e.eventType === 'SPECIMEN' || 'SAMPLE_ACQUISITION'
-        );
+        const specimenEvents = events.filter((e: ClinicalEvent) => {
+            return /SPECIMEN|Sample Acquisition|sample_acquisition'/i.test(
```
alisman, line 99: @inodb this bug was two fold.   
First, it turns out this code here was wrong and never evaluated properly.  However, as long as events were returned in order, things worked.  You uncovered a case where the events came back in opposite order, probably because sequencing was done in reverse order of acquisition.

(Note that the samples were always represented properly on the timeline, it was just the integer name that was backward (though consistent in the UI). 

On top of this, we did not have the right set of names at line 99, if there is in fact any standard.  If we don't have a standard then we are subject to the order of the events.

inodb: Thanks for fixing and explaining!

The GENIE BPC studies follow a standard. The rest of the public/MSK studies don't afaik. The order for BPC is: Sample Acquisition, Sequencing.

The "SPECIMEN" track is something that occurs in public or MSK internal studies but shouldn't happen in combination with the other two. Maybe it make sense to do some ordering like:

- Sample Acquisition
- Any other track with a `SAMPLE_ID` attribute

inodb (approved):
LGTM - thanks!

src/pages/patientView/SampleManager.tsx
inodb: Then we could e.g. use "Sample Acquisition" as the main sample collection attribute. This issue also came up here: https://github.com/cBioPortal/cbioportal/issues/9966

alisman: I think just about ever track is going to have a SAMPLE_ID field:
image

alisman: @inodb so you're saying use SPECIMEN only on our internal portals, otherwise use Sample Acquisition?

PR #4492, opened 2023-01-25
Improve Pathway Mapper default genes of interest list to include all alteration types

YusufZiyaOzgul:
Fix cBioPortal/cbioportal#9974

Looks like currently we are relying on genes sorted by mutation frequency for Pathways tab. Actually we need all alterations, not just mutations. So, we need to get sort genes by alteration frequency instead to always get the desired list of genes.

### DIFF
--- a/src/pages/groupComparison/GroupComparisonPage.tsx
+++ b/src/pages/groupComparison/GroupComparisonPage.tsx
@@ -265,6 +265,7 @@ export default class GroupComparisonPage extends React.Component<
                             <GroupComparisonPathwayMapper
                                 genomicData={
                                     this.store.alterationEnrichmentRowData
+                                        .result || []
                                 }
                                 activeGroups={this.store.activeGroups.result}
                                 store={this.store}
--- a/src/pages/groupComparison/pathwayMapper/GroupComparisonPathwayMapper.tsx
+++ b/src/pages/groupComparison/pathwayMapper/GroupComparisonPathwayMapper.tsx
@@ -54,7 +54,7 @@ export default class GroupComparisonPathwayMapper extends React.Component<
     @computed get activeGenes() {
         return (
             this._userSelectedGenes ||
-            this.props.store.genesSortedByMutationFrequency.result?.slice(
+            this.props.store.genesSortedByAlterationFrequency.result?.slice(
                 0,
                 10
             ) ||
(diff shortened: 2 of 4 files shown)


src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1208,6 +1208,24 @@ export default abstract class ComparisonStore extends AnalysisStore
...
+            this.alterationsEnrichmentAnalysisGroups,
+        ],
+        invoke: async () => {
+            const alterationRowData: AlterationEnrichmentRow[] = getAlterationRowData(
+                this.alterationsEnrichmentData.result!,
+                this.resultsViewStore
+                    ? this.resultsViewStore.hugoGeneSymbols
+                    : [],
+                this.alterationsEnrichmentAnalysisGroups.result!
+            );
```
onursumer, line 1223: Looks like we are already computing this part in the [@computed get alterationEnrichmentRowData](https://github.com/cBioPortal/cbioportal-frontend/blob/6ffbe5e2a41ce6e765f40391fda37fa91c221232/src/shared/lib/comparison/ComparisonStore.ts#L1129) function. We should probably reuse that function here as well.

We may try converting `alterationEnrichmentRowData` to a `readonly` function and awaiting it here.

onursumer (changes requested):
Thanks! This change seems working, we may just want to reuse an existing function inside the new one.

src/pages/groupComparison/GroupComparisonPage.tsx
```diff
@@ -265,6 +265,7 @@ export default class GroupComparisonPage extends React.Component<
                             <GroupComparisonPathwayMapper
                                 genomicData={
                                     this.store.alterationEnrichmentRowData
+                                        .result!
```
onursumer, line 268: probably safer to return an empty array here
```suggestion
                                        .result || []
```

src/pages/groupComparison/pathwayMapper/GroupComparisonPathwayMapper.tsx
```diff
@@ -63,9 +63,11 @@ export default class GroupComparisonPathwayMapper extends React.Component<
     }
 
     @computed get alterationEnrichmentRowData(): AlterationEnrichmentRow[] {
-        return this.props.store.alterationEnrichmentRowData.filter(gene => {
-            return this.activeGenes.includes(gene.hugoGeneSymbol);
-        });
+        return this.props.store.alterationEnrichmentRowData.result!.filter(
```
onursumer, line 66: let's assume that result might be undefined
```suggestion
        return (this.props.store.alterationEnrichmentRowData.result || []).filter(
```

onursumer (changes requested):

onursumer (approved):
Looks good. Let's do a squash merge to reduce the number of commit messages.

alisman (approved):

PR #4494, opened 2023-01-26
Oql upstream downstream fusions in oncoprint

pvannierop:
This PR continues from #4413.
This PR depends on backend PR cbioportal/cbioportal#9949

The new OQL syntax for SVs/fusions uses '::' notation. With this PR the OQL will show relevant tracks in Oncoprint.

image

image


packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
```diff
@@ -5537,6 +5553,14 @@ export default class CBioPortalAPIInternal {
             'sampleMolecularIdentifiers0MolecularProfileId' ? : string,
             'sampleMolecularIdentifiers0SampleId' ? : string,
             'structuralVariantFilter': StructuralVariantFilter,
+            'structuralVariantQueries0Gene1ContainingOnlyOneIdentifierOrSpecialValue' ? : boolean,
```
alisman, line 5556: seems kinda strange that there's index in this property name?

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -758,6 +764,55 @@ export class ResultsViewPageStore extends AnalysisStore
         }
     }
 
+    /**
+     * Genes of 'single gene' queries
+     * (i.e. non-fusion queries, without an upstream or downstream gene)
+     */
+    @computed get singleHugoGeneSymbols() {
+        if (
+            this.urlWrapper.query.gene_list &&
```
alisman, line 773: You can now use optional chaining to write this condition ```this.urlWrapper.query?.gene_list.length```
it will evaluate falsy if gene_list is undefined

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -758,6 +764,55 @@ export class ResultsViewPageStore extends AnalysisStore
...
+    }
+
+    /**
+     * All genes ($.gene and $.alterations.gene) of fusion queries
+     * (i.e. queries with an upstream or downstream gene or special value)
+     */
+    @computed get structVarHugoGeneSymbols(): string[] {
+        return _(this.structVarQueries)
+            .flatMap(query => [getFirstGene(query), getSecondGene(query)])
+            .compact()
```
alisman, line 800: unless i'm mistaken, compact and uniq will have no impact here.  flatMap will return a n array of arrays (regardless if contents of child arrays are undefined).  The child arrays will always be A) defined and B) unique.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -3340,15 +3399,26 @@ export class ResultsViewPageStore extends AnalysisStore
             // filters can be an empty list
             // when all selected samples are coming from studies that don't have structural variant profile
             // in this case, we should not fetch structural variants data
-            if (_.isEmpty(filters)) {
+            if (_.isEmpty(sampleMolecularIdentifiers)) {
                 return [];
             } else {
+                const entrezGeneIds = genes
```
alisman, line 3405: I think genes is, like 20k long, so perhaps we should have a computed lookup which is symbol to gene and then you can just iterate through structVarHugoGeneSymbols

src/shared/lib/oql/oqlfilter.ts
```diff
@@ -1205,6 +1242,23 @@ const getGeneFromAlterations = (q: SingleGeneQuery) => {
     return (q.alterations as FUSIONCommandOrientationBase[])[0].gene;
 };
 
+export function fusionsInOQLQuery(oql_query: string): SingleGeneQuery[] {
```
alisman, line 1245: curious why you switched to "fusion" here when everything else in PR is structural variant

alisman (changes requested):
see comments

packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
pvannierop: Ah yes, this required running `yarn updateAPI` against the correct backend branch. It is fixed now.

src/pages/resultsView/ResultsViewPageStore.ts
pvannierop: Fix at all three places that his was done.

src/pages/resultsView/ResultsViewPageStore.ts
pvannierop: No, as far as I can tell for the lodash docs flatMap returns an array. The chained .compart() will remove any falsy values.

src/pages/resultsView/ResultsViewPageStore.ts
pvannierop: I think that `genes` is the genes that are derived from the query. There is a structure called `referenceGenes` that appears to represent all genes in the reference genome (~20k). Howerer, I have I agree that this mapping would be more efficient with a looup. I have created a hugoGeneSymbol => Gene lookup and mapped from there.

src/shared/lib/oql/oqlfilter.ts
pvannierop: Ah good point, I have renamed these methods to the SV equivalent.

src/pages/resultsView/ResultsViewPageStore.ts
alisman: still seems not done?

src/pages/resultsView/ResultsViewPageStore.ts
alisman: Flatmap will return an array of whatever the callback/iteratee returns.  Here you are returning an array, so it will be an array of arrays, none of which will be undefined or duplicated.
image

pvannierop:
[SV in Oncoprint](https://trello.com/c/BbcqxJtu/5-sv-in-oncoprint)

packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
pvannierop: This fixed with merge of the backend PR.

src/pages/resultsView/ResultsViewPageStore.ts
pvannierop: Yes it is done.

src/pages/resultsView/ResultsViewPageStore.ts
pvannierop: No, your example here is not correct. To match the code it should be:

```
_.flatMap([[1,2], [3,4]], m => [m[0], m[1]])
```

Moreover, I know I am right because 1) the code works and 2) the type checking agress with the specified `string[]` contract.

alisman (approved):

PR #4499, opened 2023-01-30
Improve performance of sample identifier filtering in study view

alisman:
Fixes https://github.com/cBioPortal/cbioportal/issues/9982

Apply faster intersection algorithm


inodb (approved):
Nice - def a lot faster! I tried to do a 10K cohort selection on internal MSK cohort and it's almost instant whereas before it was several seconds

PR #4505, opened 2023-02-02
Fusion tab2023

alisman:


src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,108 @@
...
+    @observable _fusionGeneTab: string | undefined = undefined;
+    @computed get fusionGeneTab() {
+        return this._fusionGeneTab || this.props.store.hugoGeneSymbols[0];
+    }
+
+    set fusionGeneTab(t: string | undefined) {
+        this._fusionGeneTab = t;
+    }
+
+    protected generateTabs(genes: string[]) {
```
alisman, line 49: instead of method, make this a @computed getter.    to get the genes input,   this.props.store.hugoGeneSymbols!

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,108 @@
...
+        return tabs;
+    }
+
+    @autobind
+    protected handleTabChange(id: string) {
+        // update the hash if routing exits
+        if (this.props.routing) {
+            this.props.routing.updateRoute({ fusionGeneTab: id });
+        }
+        // update the observable if no routing
```
alisman, line 72: there probably isn't really a case where there's no routing, so just get rid of this

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,108 @@
...
+        else {
+            this.fusionGeneTab = id;
+        }
+    }
+
+    readonly fusionUI = MakeMobxView({
+        await: () => [this.props.store.fusionMapperStores],
```
alisman, line 79: look into whether we can just await the fusion data for the present gene tab that we are on

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -5005,6 +5025,51 @@ export class ResultsViewPageStore extends AnalysisStore
         },
     });
 
+    readonly fusionMapperStores = remoteData<{
```
alisman, line 5028: this instantiates all the gene stores up front when it's possible that you really only need a store for the currently selected gene.  so investigate if it's possible to instantiate the stores lazily, when a gene tab is actually shown.

src/pages/resultsView/fusion/Fusions.tsx
```diff
@@ -0,0 +1,108 @@
...
+    }
+
+    set fusionGeneTab(t: string | undefined) {
+        this._fusionGeneTab = t;
+    }
+
+    protected generateTabs(genes: string[]) {
+        const tabs: JSX.Element[] = [];
+        genes.forEach((gene: string) => {
+            const fusionMapperStore = this.props.store.fusionMapperStores
```
alisman, line 52: we don't actually need these stores until the tab for their gene is actually selected

src/pages/resultsView/fusion/Fusions.tsx
alisman: you don't need await fusionMapperStores, you just need to await genes, because that's all you need for the tabs.

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,96 @@
...
+                        this.props.store.fusionMolecularProfile
+                    }
+                />
+            </>
+        );
+    }
+
+    readonly resultFusionUI = MakeMobxView({
+        await: () => [
+            this.props.store.studyIdToStudy,
```
alisman, line 79: this is gene specific data (one store per gene), so this makes sense to await here.

src/pages/resultsView/fusion/ResultViewFusionMapper.tsx
```diff
@@ -0,0 +1,96 @@
...
+    store: ResultViewFusionMapperStore;
+}
+
+@observer
+export default class ResultViewFusionMapper extends React.Component<
+    IFusionMapperProps,
+    {}
+> {
+    private renderComponents() {
+        const studyMap = new Map(
```
alisman, line 42: see if you can avoid this transformation and just normalize the type expected by ResultsViewFUsionTable

src/pages/resultsView/fusion/ResultViewFusionMapperStore.ts
```diff
@@ -0,0 +1,190 @@
...
+                const transcriptIds = new Set<string>();
+                this.fusions.forEach((fusion: StructuralVariant) => {
+                    transcriptIds.add(fusion.site1EnsemblTranscriptId);
+                    transcriptIds.add(fusion.site2EnsemblTranscriptId);
+                });
+                // Fetch ensembl transcript data with transcript ids as request params
+                return await fetchEnsemblTranscriptsByEnsemblFilter({
+                    transcriptIds: Array.from(transcriptIds),
+                }).then(result =>
+                    ResultViewFusionMapperStore.addColor(
```
alisman, line 118: avoid using static functions because we don't do it very much and it isn't common in js.  just declare the functino above the class and call it

PR #4511, opened 2023-02-09
Do not display a default plot if the survival prefix table is initially empty

onursumer:
Fix cBioPortal/cbioportal#9991


dippindots (approved):
Looks good to me!

onursumer:
Failing e2e tests don't seem to be related. Merging.

PR #4513, opened 2023-02-13
Fix localdb bootup (on local linux machines)

alisman:


inodb (approved):
Ofc the issue is letter case! Restricting to storing lower case only seems like a good solution - thanks! 

>How table and database names are stored on disk and used in MySQL is affected by the [lower_case_table_names](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) system variable. [lower_case_table_names](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) can take the values shown in the following table. This variable does not affect case sensitivity of trigger identifiers. On Unix, the default value of [lower_case_table_names](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_lower_case_table_names) is 0. On Windows, the default value is 1. On macOS, the default value is 2

From: https://dev.mysql.com/doc/refman/8.0/en/identifier-case-sensitivity.html

PR #4519, opened 2023-02-16
10015 patient view tooltip months numbers should be rounded

dubincorey:
Fix cBioPortal/cbioportal#10015

- reduce number of decimals to the hundredth place

## Any screenshots or GIFs?
Screenshot 2023-02-16 at 5 13 23 PM


src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -41,6 +41,19 @@ export default class ClinicalInformationPatientTable extends React.Component<
...
         }
+        try {
+            const parsedFloat = parseFloat(data.value);
+
+            if (parsedFloat != parseInt(data.value)) {
+                // if it's a float, not an integer
+                const prevValue = data.value.split('.')[0];
+                const roundedValue = data.value.split('.')[1];
+                const addDecimal = prevValue.concat('.');
+                ret = addDecimal.concat(roundedValue.slice(0, 2));
```
onursumer, line 52: Have you tried reusing one of the functions defined in https://github.com/cBioPortal/cbioportal-frontend/blob/1730ff3504d8b0daaf03794642a0f51928195ac8/packages/cbioportal-utils/src/format/FormatUtils.ts or https://github.com/cBioPortal/cbioportal-frontend/blob/ad87a472069d17aa68e76881a5581f75b329be1a/src/shared/lib/FormatUtils.tsx?

If none of the existing functions work for this case, we should add a new function in one of those files and just call it here to keep this `getDisplayValue` function simple. 

Also, I wouldn't modify the value of `ret` in a separate try/catch block. Why not just update [this line](https://github.com/cBioPortal/cbioportal-frontend/pull/4519/files#diff-e932fb53846ef921820aeaacb9c87b0135649f156ec85c1a76a95c9033405159R38)?

onursumer (changes requested):
Let's try reusing one of the existing utility functions first.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
dubincorey: @onursumer Thanks for the help! I used one of the `FormatUtils`, and assuming rounding to the nearest tenth place is okay, I think it should be good. I removed the try/catch block, but wasn't sure how I could implement my changes into the switch statement. Instead, I just made a small condition that checks that none of the values I'm reading were integers originally. I think that should be good. Screenshot 2023-02-23 at 2 21 01 PM

onursumer: @dubincorey I think you can just use the function `formatNumberValueInSignificantDigits`, and do something like this within the switch statement.
```JavaScript
case 'Overall Survival (Months)':
    ret = formatNumberValueInSignificantDigits(Number(data.value), 3);
```
Probably no need for additional checks, since this will return no decimal points when `data.value` is an integer, and return `NaN` in case `data.value` is not a number.

onursumer: Hmm, actually I was wrong. `formatNumberValueInSignificantDigits` doesn't always work as I expected for bigger numbers. So maybe we should introduce a similar utility function utilizing the `maximumFractionDigits` option of the `toLocaleString` function.

onursumer: Not directly related to this issue but I also found a bug with `formatNumberValueInSignificantDigits` (see cBioPortal/cbioportal#10046). We should be careful when using the `toLocaleString` function.

dubincorey: Also, I don't think I can use
`case 'Overall Survival (Months)':`
as a condition for formatting `data.value`. Because the `ClinicalInformationPatientTable` has multiple attributes with values that are long floats. Wouldn't I need to use multiple cases to match for each attribute that has a long float?

onursumer: @dubincorey Thanks for the clarification. I was under the impression that we needed this change only for the `Overall Survival (Months)` case. If this logic needs to be applied in any case, then we should probably replace the switch case block with a simpler if/else block. Something like:

```
if the value is numerical
    format it with the new "formatNumberValueInFractionDigits" function (which will keep the integer value as-is)
else
    keep the non-numerical value as-is
```

onursumer: But one other question is do we really always want to keep `Overall Survival (Months)` strictly as integers? If that's what we really want, then we still need to keep that logic.

dubincorey: I can ask during standup, but the `Overall Survival (Months)` rounding change was made 6 years ago so we can see what the priority is at this point.

dubincorey: Also, at the moment, `formatPercentValue` is rounding all the floating values that I've found to the nearest tenth, and leaving all integer values alone. Do you see a problem with this?

inodb (approved):
Lgtm - thank you!

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
dubincorey: @inodb We discussed during standup, and Gaofei mentioned certain papers preferred to use floor rounding. He mentioned this was being used elsewhere as well. @onursumer's concern was that some values like `.36` or `.98` would show as `0`. If this is not a big deal, I can make this a floor function.

onursumer: The only potential issue with `formatPercentValue` might be this condition: https://github.com/cBioPortal/cbioportal-frontend/blob/1730ff3504d8b0daaf03794642a0f51928195ac8/packages/cbioportal-utils/src/format/FormatUtils.ts#L7-L9

Probably not a big deal though.

alisman:
@dubincorey there's something amiss with this PR.  I think you may have started with a local branch from your last project.  See the 11 commits that are present here.

dubincorey:
I've made this a mess but made a clean PR here https://github.com/cBioPortal/cbioportal-frontend/pull/4531

PR #4520, opened 2023-02-17
Study view km plot optimization

dippindots:
Fix cBioPortal/cbioportal#9988
Fix cBioPortal/cbioportal#8378

### **Changes proposed in this pull request:**
- Keep old implementation for comparison page/comparison tab survival plot
- Only apply binning when the sample size is larger than the threshold (default `1000`, we can change this into a configurable value later)
- Logic for binning: flooring month to integer (e.g. flooring `100.2` or `100.4` to `100`) and count any events or censored data that get aggregated
- Add information into tooltip to show the count for events and censored

### **Other comments:**
- [x] Change text in the tooltip:
  - `Time of event: 46-47 months` -> `Events during [46,47) months`. Use brackets/parenthesis notation here depending on how you are binning: https://stackoverflow.com/a/4396303
  - `Patients with an event: `
  - `Censored patients: `
  - Newline
  - `% event free at end of months (46 months)` -> `% event free at interval end: `
  - `Patients at risk at interval end: `
  - Screen Shot 2023-02-20 at 3 46 40 PM
- [x] Indicate in info icon that events were floored to the nearest month
- [x] Indicate the unit for the x axis in the info icon

### **Follow-up ticket:**
cBioPortal/cbioportal#10140


inodb:
Thanks @dippindots - this looks great!

Maybe a few tweaks to the text and ordering here:

Screen Shot 2023-02-20 at 3 46 40 PM

- `Time of event: 46-47 months` -> `Events during (46,47] months`. Use brackets/parenthesis notation here depending on how you are binning: https://stackoverflow.com/a/4396303. I assume 46.9 months is in this particular bin too
- `Patients with an event: `
- `Censored patients: `
- Newline
- `% event free at end of months (46 months)` -> `% event free at interval end: `
- `Patients at risk at interval end: `

It might make sense to highlight the entire bin when mousing over (so one horizontal line). Currently, it highlights points and it's a little hard to see that it's showing an interval:

Screen Shot 2023-02-20 at 4 15 57 PM

tmazor:
This is great! I agree with all of Ino's comments above, and I also noticed that the tooltips will extend off the page:
image

Can you make the tooltips appear to the left of the chart when the chart is on the right side of the page, like pie charts do?
image

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -520,6 +531,80 @@ export function downSampling(
...
+            numberOfEvents: eventInfo.eventCount,
+            numberOfCensored: eventInfo.censorCount,
+        };
+        return aggregatedScatter;
+    });
+    return result;
+}
+
+function updateEventInfo(eventInfo: EventInfo, data: ScatterData): EventInfo {
+    let updatedEventInfo: EventInfo = eventInfo;
```
alisman, line 582: keep in mind that you are not actually cloning this object here.  so any change you make below will actually apply to BOTH updatedEventInfo and eventInfo, since one is just a reference to the other.    This may or may not be a problem.  However, the code should not "pretend" that is producing something new.  I.e. if it's just mutating eventInfo input, then it should simply do that.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -520,6 +531,80 @@ export function downSampling(
...
+        ? eventInfo.eventCount + 1
+        : eventInfo.eventCount;
+    updatedEventInfo.censorCount = data.status
+        ? eventInfo.censorCount
+        : eventInfo.censorCount + 1;
+    return updatedEventInfo;
+}
+
+export function getLineDataFromScatterData(data: ScatterData[]): any[] {
+    let chartData: any[] = [];
```
alisman, line 595: seems it would be simple enough to type this.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -520,6 +531,80 @@ export function downSampling(
...
+            atRisk: eventInfo.lastRiskInMonth,
+            numberOfEvents: eventInfo.eventCount,
+            numberOfCensored: eventInfo.censorCount,
+        };
+        return aggregatedScatter;
+    });
+    return result;
+}
+
+function updateEventInfo(eventInfo: EventInfo, data: ScatterData): EventInfo {
```
alisman, line 581: lets give this a more descriptive name.  like, what is the update it is performing?

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -520,6 +531,80 @@ export function downSampling(
     });
 }
 
+export function floorScatterData(
+    scatterDataWithOpacity: ScatterData[]
+): ScatterData[] {
+    const eventInfoByMonth: { [month: number]: EventInfo } = {};
+    let eventInfo: EventInfo;
+    let result: ScatterData[] = [];
+    _.reduce(
```
alisman, line 540: you should always assign to the variable with any map/reduce/filter/etc operation.
```
const eventInfoByMonth = _.reduce(input, (aggr:TYPE HERE)=> ... })
```

alisman:
@dippindots lets bring this up at standup today to figure out what's next.

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: It's just updating the whole EventInfo object based on the new ScatterData, I don't know what could be a better name, so I just added some comments.

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: Thanks, got it, I removed it.

src/pages/resultsView/survival/SurvivalUtil.tsx
dippindots: updated

alisman (approved):

PR #4525, opened 2023-02-23
fix formatNumberValueInSignificantDigits function for bigger numbers

onursumer:
Fix cBioPortal/cbioportal#10046

### DIFF
--- a/packages/cbioportal-utils/src/format/FormatUtils.spec.ts
+++ b/packages/cbioportal-utils/src/format/FormatUtils.spec.ts
@@ -1,0 +1,100 @@
+import { formatNumberValueInSignificantDigits } from './FormatUtils';
+import { assert } from 'chai';
+
+describe('FormatUtils', () => {
+    describe('formatNumberValueInSignificantDigits', () => {
+        it('returns null if null', () => {
+            assert.equal(formatNumberValueInSignificantDigits(null, 6), null);
+        });
+
+        it('does not format if no significant digits', () => {
+            assert.equal(
+                formatNumberValueInSignificantDigits(1.12345),
+                1.12345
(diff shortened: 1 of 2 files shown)


packages/cbioportal-utils/src/format/FormatUtils.ts
```diff
@@ -34,9 +34,11 @@ export function formatNumberValueInSignificantDigits(
         return null;
     } else if (significantDigits) {
         return Number(
-            value.toLocaleString(undefined, {
-                maximumSignificantDigits: significantDigits,
-            })
+            value
+                .toLocaleString('en-US', {
```
leexgh, line 38: What does `en-US` do?

packages/cbioportal-utils/src/format/FormatUtils.spec.ts
```diff
@@ -0,0 +1,100 @@
+import { formatNumberValueInSignificantDigits } from './FormatUtils';
+import { assert } from 'chai';
+
+describe('FormatUtils', () => {
```
leexgh, line 4: Great! Thanks for adding the tests

packages/cbioportal-utils/src/format/FormatUtils.ts
onursumer: Not sure if this is critical but this is basically to make sure that we are using the same locale everywhere. In some rare cases it may format the number in a different way which may cause problems.

See the `locales` param [here](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toLocaleString) for details.

leexgh (approved):

PR #4530, opened 2023-02-28
:sparkles: Add Clinical Event Type Counts Table

haynescd:
Fix cBioPortal/cbioportal#10000

- Update Study View Summary page to allow filtering on Clinical Event Type

## Checks

## Any screenshots or GIFs?
image
image

## Notify reviewers


src/pages/studyView/table/ClinicalEventTypeCountTable.tsx
```diff
@@ -0,0 +1,412 @@
+import { ClinicalEventTypeCount } from 'cbioportal-ts-api-client/dist/generated/CBioPortalAPIInternal';
```
haynescd, line 1: Adding this Table seems to be redundant... I need to refactor MultiSelectionTable.tsx to be more generic

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1992,6 +1993,15 @@ export class StudyViewPageStore
 
         this.setTreatmentFilters(filters);
 
+        if (
+            filters.clinicalEventFilters !== undefined &&
+            !_.isEmpty(filters.clinicalEventFilters)
```
alisman, line 1998: isEmpty will actually cover undefined, so no need to check it first

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9218,6 +9267,78 @@ export class StudyViewPageStore
             return isMixedReferenceGenome(this.queriedPhysicalStudies.result);
         }
     }
+    private _clinicalEventFilters = observable.array<DataFilter>([], {
```
alisman, line 9270: from what i can see, we are never editing this array except by replacing all of its contents entirely.  this means that we could use a simple array assigned to an observable reference instead of observable.array, which is only necessary when you need to respond to mutations of the array itself.    what i mean by this is, 
```
@observable.ref 
private _clinicalEventsFilters = [];

```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9218,6 +9267,78 @@ export class StudyViewPageStore
...
+        this._clinicalEventFilters.replace(filters);
+    }
+
+    @action
+    public resetClinicalEventTypeFilter() {
+        this._clinicalEventFilters.clear();
+    }
+
+    @computed get clinicalEventTypeFiltersRawStrings(): string[][] {
+        const filters = _.map(this.clinicalEventFilters, filterList =>
```
alisman, line 9289: coupla things here.  1. js arrays have map/reduce/filter methods built in so you don't need to use lodash for simple cases this.  ```this.clinicalEventFilters.map((filterList)=>...)``` will work

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9218,6 +9267,78 @@ export class StudyViewPageStore
...
+            });
+        },
+    });
+
+    @action.bound
+    removeClinicalEventTypeFilter(eventTypeFilterToRemove: string): void {
+        let currentClinicalEventTypeFilters =
+            toJS(this.clinicalEventFilters) || [];
+
+        const updatedClinicalEventTypeFilters = _.forEach(
```
alisman, line 9308: we should probably use map instead of forEach.  You don't typically assign from whatever is returned from forEach.  i can explain.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6554,6 +6587,17 @@ export class StudyViewPageStore
             );
         }
 
+        if (!_.isEmpty(this.clinicalEventTypeCounts.result)) {
```
alisman, line 6590: maybe we can hide this stuff inside a method a la  ```initializeClinicalDataCountCharts```

src/pages/studyView/UserSelections.tsx
```diff
@@ -702,6 +718,55 @@ export default class UserSelections extends React.Component<
         );
     }
 
+    private renderClinicalEventFilter(
+        f: DataFilter[],
+        displayName: string
+    ): JSX.Element {
+        const filters = (f as any).map((oFilter: DataFilter) => {
+            const pills = (oFilter.values as any).map(
```
alisman, line 726: lets discuss the 'any' casting, which should be avoided at all cost.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9218,6 +9267,78 @@ export class StudyViewPageStore
...
+
+    @action.bound
+    removeClinicalEventTypeFilter(eventTypeFilterToRemove: string): void {
+        let currentClinicalEventTypeFilters =
+            toJS(this.clinicalEventFilters) || [];
+
+        const updatedClinicalEventTypeFilters = _.forEach(
+            currentClinicalEventTypeFilters,
+            eventTypeFilter => {
+                _.remove(
```
alisman, line 9311: use filter instead of remove

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -9218,6 +9267,78 @@ export class StudyViewPageStore
...
+        this.setClinicalEventTypeFilter(updatedClinicalEventTypeFilters);
+    }
+
+    @action.bound
+    addClinicalEventTypeFilter(
+        chartMeta: ChartMeta,
+        eventTypeFilters: string[][]
+    ): void {
+        let filtersToAdd = _.map(eventTypeFilters, filters => {
+            let values = _.map(filters, f => {
```
alisman, line 9328: let vales: DataFilterValue[] = ...

src/pages/studyView/table/ClinicalEventTypeCountTable.tsx
```diff
@@ -0,0 +1,412 @@
...
+
+    @computed
+    get cellMargin() {
+        return _.reduce(
+            this.props.columns,
+            (acc, column) => {
+                acc[column.columnKey] = 0;
+                return acc;
+            },
+            {} as { [key in ClinicalEventTypeCountColumnKey]: number }
```
alisman, line 209: you can avoid using AS here by putting the type on the acc argument  of the callback.   i.e. 
```
this.props.columns.reduce((acc:TYPE, column)=>...)
```

haynescd:
Updated images of Event Table
Screenshot 2023-03-09 at 1 22 35 AM
Screenshot 2023-03-09 at 1 22 54 AM

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5964,9 +5995,10 @@ export class StudyViewPageStore
         }
 
         if (!pending && this.isInitiallLoad) {
-            pending = pending || this.selectedSamples.result.length === 0;
-        }
-        if (!pending && this.isInitiallLoad) {
+            pending =
+                pending ||
+                this.clinicalEventTypeCounts.isPending ||
```
haynescd, line 6000: @alisman This is the one change I had to add since you reviewed. Needed to add this to prevent "out of sync" bug

PR #4531, opened 2023-02-28
Patient view tooltip months numbers should be rounded

dubincorey:
Fix for https://github.com/cBioPortal/cbioportal/issues/10015

Supersedes #4519

Clean up commit history of old PR since it was a big mess

### DIFF
--- a/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
+++ b/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
@@ -37,7 +37,7 @@ export default class ClinicalInformationPatientTable extends React.Component<
             return data.value;
         } else {
             const parsedFloat = parseFloat(data.value);
-            if (parsedFloat != parseInt(data.value)) {
+            if (parsedFloat !== parseInt(data.value)) {
                 // the values formatted are not percentages in this case however the function cleanly rounds any float to 1 decimal place
                 return formatPercentValue(parsedFloat);
             } else {
--- a/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
+++ b/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
@@ -35,7 +35,9 @@ export default class ClinicalInformationPatientTable extends React.Component<
     }): string {
         let ret: string;
 
-        if (!isNaN(data.value as any)) {
+        if (isNaN(data.value as any)) {
+            ret = data.value;
+        } else {
             const parsedFloat = parseFloat(data.value);
             if (parsedFloat != parseInt(data.value)) {
                 // the values formatted are not percentages in this case however the function cleanly rounds any float to 1 decimal place
@@ -43,8 +45,6 @@ export default class ClinicalInformationPatientTable extends React.Component<
             } else {
                 ret = data.value;
             }
-        } else {
-            ret = data.value;
         }
         return ret!;
     }
--- a/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
+++ b/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
@@ -35,15 +35,18 @@ export default class ClinicalInformationPatientTable extends React.Component<
     }): string {
         let ret: string;
 
-        const parsedFloat = parseFloat(data.value);
-        if (parsedFloat != parseInt(data.value) && !Number.isNaN(parsedFloat)) {
-            // the values formatted are not percentages in this case however the function cleanly rounds any float to 1 decimal place
-            ret = formatPercentValue(parsedFloat);
+        if (!isNaN(data.value as any)) {
+            const parsedFloat = parseFloat(data.value);
+            if (parsedFloat != parseInt(data.value)) {
+                // the values formatted are not percentages in this case however the function cleanly rounds any float to 1 decimal place
+                ret = formatPercentValue(parsedFloat);
+            } else {
+                ret = data.value;
+            }
         } else {
             ret = data.value;
         }
-
-        return ret;
+        return ret!;
     }
 
     @autobind handleClick(name: string) {
--- a/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
+++ b/src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
@@ -7,6 +7,7 @@ import { SHOW_ALL_PAGE_SIZE } from '../../../shared/components/paginationControl
 import { sortByClinicalAttributePriorityThenName } from '../../../shared/lib/SortUtils';
 import { isUrl } from 'cbioportal-frontend-commons';
 import autobind from 'autobind-decorator';
+import { formatPercentValue } from 'cbioportal-utils';
 
 export interface IClinicalInformationPatientTableProps {
     data: ClinicalData[];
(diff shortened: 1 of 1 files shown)


src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -33,14 +34,14 @@ export default class ClinicalInformationPatientTable extends React.Component<
...
     }): string {
         let ret: string;
-        switch (data.attribute) {
-            case 'Overall Survival (Months)':
-                ret = parseInt(data.value, 10).toFixed(0);
-                break;
-            default:
-                ret = data.value;
-                break;
+        const parsedFloat = parseFloat(data.value);
```
alisman, line 37: data.value is often not numberical, so shouldn't be treated this way.  You need to make sure it's numeric.
image

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -33,14 +34,14 @@ export default class ClinicalInformationPatientTable extends React.Component<
...
+
+        if (parsedFloat != parseInt(data.value)) {
+            ret = formatPercentValue(parsedFloat);
```
alisman, line 40: i don't understand why we are using this formatPercentValue here.   also, why we are able to do away with special handling for Overall Survival (Months) case?

alisman (changes requested):
see comments

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
onursumer: Should we handle `Overall Survival (Months)` like any other value with months, like `Months from Diagnosis`? Do we strictly want month values to be integers without any decimal points? If not, maybe it's better to simply treat all floating point numbers the same way, like showing only 1 or 2 decimal points.

If we want more granularity, then we need to distinguish month attributes from other numerical attributes. Current production code only treats `Overall Survival (Months)` in a special way and probably incorrectly displays categorical attributes as `NaN`.

dubincorey: @onursumer This was my thought process as well. The specific case of `Overall Survival (Months)` was made 6 years ago and I don't know how many attributes existed at that point. So @alisman I had thought it would be okay to handle `Overall Survival (Months)` like the other floating point attributes.

And Onur and I discussed using formatPercentValue here (https://github.com/cBioPortal/cbioportal-frontend/pull/4519#discussion_r1117223083)

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
dubincorey: Corrected this with https://github.com/cBioPortal/cbioportal-frontend/pull/4531/commits/fa9cc113a0d14393cce9f54f25074fde23cbf511

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
dubincorey: @alisman We discussed this last Friday during standup, and we seemed okay with things as long as they work and look good. I've corrected the NaN values from appearing and added a comment clarifying the use of `formatPercentValue`, so if things look okay to you this is hopefully good to go.

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -33,14 +34,15 @@ export default class ClinicalInformationPatientTable extends React.Component<
...
         let ret: string;
-        switch (data.attribute) {
-            case 'Overall Survival (Months)':
-                ret = parseInt(data.value, 10).toFixed(0);
-                break;
-            default:
-                ret = data.value;
-                break;
+
+        const parsedFloat = parseFloat(data.value);
```
alisman, line 38: Corey, this works but really we should just be testing for isNumeric first and then proceeding.   you can use lodash 
_.isNumeric function to do this.

dubincorey: I'm seeing that function as being deprecated. I can create a new utility function for this which is basically isNumeric https://github.com/lodash/lodash/issues/1148#issue-69568538
``` Javascript 
function isNumeric(x: any) {
     return !(isNaN(x)) && (typeof x !== "object") &&
       (x != Number.POSITIVE_INFINITY) && (x != Number.NEGATIVE_INFINITY);
}

Where should this function go?

Comment (onursumer) [reply src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx:38]: Maybe we use [isFinite](https://lodash.com/docs/4.17.15#isFinite) instead?

Comment (dubincorey) [reply src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx:38]: After testing, `isFinite` does not consider any string (even a number in a string like `const x = '2'`) as true. Which doesn't solve the problem Aaron mentioned of testing before converting `data.value`, because we would have to convert the string to see if `isFinite` would be true.

File: src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx [hunk 4]
```diff
@@ -35,15 +35,18 @@ export default class ClinicalInformationPatientTable extends React.Component<
     }): string {
         let ret: string;
 
-        const parsedFloat = parseFloat(data.value);
-        if (parsedFloat != parseInt(data.value) && !Number.isNaN(parsedFloat)) {
-            // the values formatted are not percentages in this case however the function cleanly rounds any float to 1 decimal place
-            ret = formatPercentValue(parsedFloat);
+        if (!isNaN(data.value as any)) {
```
dubincorey, line 38: @alisman Let me know if this works for you. Doesn't need an outside library and parses before any possible casting.

alisman: Hi Corey, yes, this is good.    Can you do me one more favor?  Just reverse this if condition, for clarity.  it should be
```
if (isNan(..)) { } else ...
```
We need to run the e2e tests again anyway so might as well improve this.  THANKS!!!!

alisman: and one more little thing.  far as i can tell, there's no reason to have this local variable ```ret```

where we assign to it, we can just do ```return x```

i see this is legacy btw.

dubincorey: @alisman Added updates to adjust for these suggestions. If that's all I think things look good!

src/pages/patientView/clinicalInformation/ClinicalInformationPatientTable.tsx
```diff
@@ -32,16 +33,17 @@ export default class ClinicalInformationPatientTable extends React.Component<
...
-                ret = parseInt(data.value, 10).toFixed(0);
-                break;
-            default:
-                ret = data.value;
-                break;
+        if (isNaN(data.value as any)) {
+            return data.value;
+        } else {
+            const parsedFloat = parseFloat(data.value);
+            if (parsedFloat != parseInt(data.value)) {
```
onursumer, line 40: Should we use strict inequality instead?
```suggestion
            if (parsedFloat !== parseInt(data.value)) {
```

dubincorey: Good point! Adjusted for this.

alisman (approved):

PR #4533, opened 2023-03-02
Allow numeric data type for custom data charts

TJMKuijpers:
Pull Request frontend implementation of custom numerical data

This pull request adds the functionality of uploading custom numerical data to cBioPortal via the add custom graph option. In the original frontend code of cBioPortal, it is only possible to upload custom categorical data. Here, we implemented the option to upload numerical data. (Note this PR cannot be merged before Binning of custom data #10039 is merged).

New functionality
- Modal to upload custom data is extended with a radio button to select categorical/numerical data type.
- Group comparison is implemented for the numerical custom data
- Users can change the binning method based on quartile / median/ bin generator or custom bins

Screenshot test
- Screenshot test added to test if the numerical/categorical radio button is present (core functionality needed in the frontend)

Screenshots
image
image

This also allows to have numerical custom data after we query based on genes (custom data 2 in the image):
image
image

## Notify reviewers


src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -237,11 +247,23 @@ export class StudySummaryTab extends React.Component<
                     props.getData = () =>
                         this.store.getChartDownloadableData(chartMeta);
                 } else {
-                    props.promise = this.store.getClinicalDataBin(chartMeta);
-                    props.filters = this.store.getClinicalDataFiltersByUniqueKey(
-                        chartMeta.uniqueKey
-                    );
-                    props.onDataBinSelection = this.handlers.onDataBinSelection;
+                    if (chartMeta.dataType === 'Custom_Data') {
```
alisman, line 250: can we make "Custom_Data" into a constant somewhere?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1,6 +1,7 @@
-import _ from 'lodash';
+import _, { uniq } from 'lodash';
```
alisman, line 1: juse use _.uniq to avoid confusion about where it comes from

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -377,6 +377,12 @@ export enum BinMethodOption {
     CUSTOM = 'CUSTOM',
 }
 
+type CustomDataType = 'CATEGORICAL' | 'NUMERICAL';
```
alisman, line 380: https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-compile-time
this should allow you to avoid this duplication

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -413,21 +419,38 @@ export class StudyViewPageStore
...
     public isShowNAChecked = (uniqueKey: string): boolean => {
-        const filter = this.getDataBinFilterSet(uniqueKey).get(uniqueKey)!;
-        const showNA = filter ? filter.showNA : undefined;
-
-        // Show NA bars by default
-        if (showNA === undefined) {
-            return true;
+        if (this._customDataBinFilterSet.has(uniqueKey)) {
+            const filter = this._customDataBinFilterSet.get(uniqueKey);
+            const showNA = filter ? filter.showNA : undefined;
```
alisman, line 438: this logic seems kind of painful.  i would just do 
```
// default to true
return filter.showNA === undefined ? true : filter.showNA;
```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1182,6 +1208,59 @@ export class StudyViewPageStore
                             this.molecularProfileMapByType,
                             selectedSamples
                         );
+                    } else if (chartMeta.dataType === 'Custom_Data') {
+                        const clinicalAttribute = chartMeta.clinicalAttribute;
```
alisman, line 1212: lets move all this logic into a tested helper function that accepts chartMeta and sampleKeyToData as args

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1182,6 +1208,59 @@ export class StudyViewPageStore
...
+                                        item => ({
+                                            sampleId: item.sampleId,
+                                            studyId: item.studyId,
+                                            patientId: item.patientId,
+                                            value: item.value,
+                                        })
+                                    );
+                                    // Combine selectedSamples with sampleKeyToData to get the desired data object
+                                    let merged = [];
+                                    for (
```
alisman, line 1231: try to avoid using for loops.  this is a mapping operation so it should be 

```
const merged = selectedSamples.map(sample=>{...});

File: src/pages/studyView/StudyViewPageStore.ts [hunk 7]
```diff
@@ -1182,6 +1208,59 @@ export class StudyViewPageStore
...
+                                            ...selectedSamples[i],
+                                            ...sampleInformation.find(
+                                                itmInner =>
+                                                    itmInner.sampleId ===
+                                                    selectedSamples[i].sampleId
+                                            ),
+                                        });
+                                    }
+
+                                    const inputObject: any[] = merged.filter(
```
alisman, line 1246: why would uniqueSampleKey ever be undefined?

if there is a valid reason, this filter operation can be chained with the map above.  

```
const mergedWhatevers = arr.map(...)
    .filter(...);
```

```

File: src/pages/studyView/StudyViewPageStore.ts [hunk 8]
```diff
@@ -1182,6 +1208,59 @@ export class StudyViewPageStore
...
+                                        item =>
+                                            item.uniqueSampleKey !== undefined
+                                    );
+                                    const outputObject: any[] = inputObject.map(
```
alisman, line 1250: you might as well just assign this map to data instead of creating this outputObject variable which doesn't do anything.  avoid unecessary steps that complicate the code.

alisman, line 1250: also, we should almost never use the `any` type in our logic.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2642,6 +2726,21 @@ export class StudyViewPageStore
...
         default: false,
     });
+    @action.bound
+    updateCustomDataIntervalFilters(
+        chartUniqueKey: string,
+        dataBins: DataBin[]
+    ): void {
+        const values: DataFilterValue[] = getDataIntervalFilterValues(dataBins);
+        this.setCustomChartNumericalFilters(chartUniqueKey, values);
+        if (!this._customDataBinFilterSet.has(chartUniqueKey)) {
```
alisman, line 2736: seems strange that we update two different things here.   lets discuss.   at very least lets put some comments explaining why two different things need to be set.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2718,7 +2815,29 @@ export class StudyViewPageStore
             this.updateClinicalDataIntervalFilters
         );
     }
-
+    @action.bound
+    setCustomChartNumericalFilters(
```
alisman, line 2819: i see below that we have a method
```setCustomChartFilters```
if this is in fact categorical, we should change the name to avoid confusion with the new numerical option

File: src/pages/studyView/tabs/SummaryTab.tsx (hunk 1 above)
Comment (TJMKuijpers) [reply src/pages/studyView/tabs/SummaryTab.tsx]: I think we can make use of the following:
```
if (  this.store.isUserDefinedCustomDataChart(
       chartMeta.uniqueKey
    )
```
In case of `ChartTypeEnum.BAR_CHART` and `isUserDefinedCustomDataChart` are both true, the custom data has a numerical type.

In other cases, we can make use of `ChartMetaDataTypeEnum.CUSTOM_DATA`

File: src/pages/studyView/StudyViewPageStore.ts (hunk 4 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts]: Updated to : 
```
            const filter = this._customDataBinFilterSet.get(uniqueKey);
            return filter === undefined ? true : (filter!.showNA as boolean);
```

File: src/pages/studyView/StudyViewPageStore.ts (hunk 8 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts]: We can define a new type `comparisonCustomData` with the following code to generate the selectedSampleData:
```
const selectedSampleData: comparisonCustomData[] = 
sampleInformation.map(
              sample =>
                ({ ...sample,                                       
                    ...selectedSamples.find(                                                   
                                           itmInner =>                                                       
                                           itmInner.sampleId ===                                                        
                                           sample.sampleId),
                } as comparisonCustomData)
```
and:
```
data = selectedSampleData.map(item => {
                     return {
                            clinicalAttribute: clinicalAttribute,
                            clinicalAttributeId:
                                              clinicalAttribute.clinicalAttributeId,
                            ...item,
                                        } as ClinicalData;
                                    })
```

File: src/pages/studyView/StudyViewPageStore.ts (hunk 9 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts]: This now is implemented in  `getCustomDataNumerical()`. When the user uploads a custom data set, we don't know the data bin values beforehand (unlike a custom categorical data where we know the categories). After obtaining the binned data from the API, we need to set the bins in the filter.

File: src/pages/studyView/StudyViewPageStore.ts (hunk 10 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts:2809]: I renamed `setCustomChartFilters` to `setCustomChartCategoricalFilters.` This filter takes input `values` (string[]) to filter categorical data. The `setCustomChartNumericalFilters` takes `dataBins` (DataBin[]) as input. It makes sense to have a more explicit name for the custom categorical data filter.

File: src/pages/studyView/StudyViewPageStore.ts (hunk 5 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts:1204]: This has been moved to the function `transformSampleDataToSelectedSampleClinicalData` in StudyViewutils.tsx  with a test in StudyViewUtils.spec.tsx.

File: src/pages/studyView/StudyViewPageStore.ts (hunk 7 above)
Comment (TJMKuijpers) [reply src/pages/studyView/StudyViewPageStore.ts]: The `uniqueSampleKey` is undefined in the `selectedSampleData` when the sample is not in `selectedSamples`. When we filter based on the `uniqueSampleKey`, we only send the selected samples in the studyview filter to the group comparison. (I think an inner join would have been better)

File: src/pages/studyView/StudyViewUtils.tsx [hunk 11]
```diff
@@ -3821,3 +3823,39 @@ export const FGA_VS_MUTATION_COUNT_KEY = makeXvsYUniqueKey(
 
 export const FGA_PLOT_DOMAIN = { min: 0, max: 1 };
 export const MUTATION_COUNT_PLOT_DOMAIN = { min: 0 };
+
+export type comparisonCustomData = {
```
alisman, line 3827: capitalize it:
ComparisonCustomData

inodb:
@TJMKuijpers Great work! Could you help fix the conflict? I think it's good to go otherwise

TJMKuijpers:
@inodb, resolved the conflict.

inodb:
@TJMKuijpers there are some updates screenshots. The barcharts now show "NA" in text rather than as a bar (maybe that's the default we want but just want to make sure - if the change is desired, the screenshot needs to be updated):

https://output.circle-artifacts.com/output/job/15424c4c-7e39-40cd-bf65-4c63a78bdfea/artifacts/0/imageCompare.html

PR #4534, opened 2023-03-03
Enable expression and protein data on multi-study when configurable condition is met

alisman:


src/shared/components/query/MolecularProfileSelector.tsx
```diff
@@ -121,6 +134,17 @@ export default class MolecularProfileSelector extends QueryStoreComponent<
         groupLabel: string
     ) {
         let profiles = this.store.getFilteredProfiles(molecularAlterationType);
+
+        const groupedProfiles = _.groupBy(profiles, profile => {
+            const profileType = profile.molecularProfileId.replace(
+                new RegExp(profile.studyId + '_'),
+                ''
+            );
```
onursumer, line 142: should we add a comment here to explain why we are removing the study id?

src/shared/components/query/QueryStore.ts
```diff
@@ -1556,14 +1565,16 @@ export class QueryStore {
     getFilteredProfiles(
         molecularAlterationType: MolecularProfile['molecularAlterationType']
     ) {
-        return this.molecularProfilesInSelectedStudies.result.filter(
+        const ret = this.molecularProfilesInSelectedStudies.result.filter(
```
onursumer, line 1568: Do we really need to declare a new const `ret`? We just return `ret` without doing anything else, right?

onursumer (commented):
Looks good to me. Just added couple minor comments.

SRodenburg:
should address related issue:
https://github.com/cBioPortal/cbioportal/issues/9865

PR #4535, opened 2023-03-03
Fix the localdb tests for ARM architecture macOS

fuzhaoyuan:
Fix cBioPortal/cbioportal#10059

Fix the localdb tests for ARM architecture macOS


scripts/e2e.sh
```diff
@@ -55,6 +55,17 @@ echo "$TEST_HOME"
...
+if [[ "$(uname -s)" == "Darwin" ]] && [[ "$(sysctl -n machdep.cpu.brand_string)" =~ M.* ]]; then
+  # if macOS and M-series chip, use images for ARM architecture
+  export DOCKER_IMAGE_MYSQL=biarms/mysql:5.7
+  export DOCKER_IMAGE_KEYCLOAK=alemairebe/keycloak:11.0.3
+  export DOCKER_IMAGE_SESSION_SERVICE=fuzhaoyuan/session-service:0.5.0
+else
+  # else use images for x86_64 architecture
+  export DOCKER_IMAGE_MYSQL=mysql:5.7
+  export DOCKER_IMAGE_KEYCLOAK=jboss/keycloak:11.0.3
+fi
```
inodb, line 67: this feels kinda similar to: https://github.com/cBioPortal/cbioportal-docker-compose/blob/master/init.sh#L7-L11. Wondering if we can just run `init.sh` in e2e tests as well?

Ok for now tho

inodb (approved):
Thanks for fixing this @fuzhaoyuan !

scripts/e2e.sh
fuzhaoyuan: Yes of course! I checked the `init.sh` file but there're some notes I considered:
1) The first part of `init.sh` file seems running itself at each sub folder and it not seen at `e2e.sh`. I'm not sure if it will interfere with the following execution (I do assume not).
2) If run from `yarn` command via `node`, since the current version of node we use (v15) only run through Rosetta translation, the `$(arch)` command will actually return `i386`. Therefore I used system check and chip check together to decide the platform.
3) The `keycloak` part is not inside `init.sh` so it needs to be specified in `e2e.sh` here anyway. So I just add others together.

fuzhaoyuan: I tried to go with this approach but it is doing something different because we called `init.sh` within each sub folder, and the override yaml file doesn't seem to work in this case.

PR #4537, opened 2023-03-04
Right Aligned numbers in study view

aditygrg2:
Fix cBioPortal/cbioportal#10006

Right aligned text for numbers in MultiSelectionTable 

Before :
Screenshot 2023-03-04 at 15 06 40

After :
Screenshot 2023-03-04 at 15 09 28

### DIFF
--- a/end-to-end-test/local/specs/user-display-name.spec.js
+++ b/end-to-end-test/local/specs/user-display-name.spec.js
@@ -6,17 +6,6 @@ var useExternalFrontend = require('../../shared/specUtils').useExternalFrontend;
 const CBIOPORTAL_URL = process.env.CBIOPORTAL_URL.replace(/\/$/, '');
 
 describe('displays appropriate user name/email', function() {
-    it('shows email in the logged-in button with deprecated value defined', function() {
-        userEmailAddress = 'test@email.com';
-        goToUrlAndSetLocalStorageWithProperty(CBIOPORTAL_URL, true, {
-            user_email_address: userEmailAddress,
-        });
-        assert.equal(
-            $('#dat-dropdown.btn-sm.dropdown-toggle.btn.btn-default').getText(),
-            'Logged in as ' + userEmailAddress + ' '
-        );
-    });
-
     it('shows email in the logged-in button with new value defined', function() {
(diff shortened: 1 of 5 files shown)


src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -324,7 +329,11 @@ export class MultiSelectionTable extends React.Component<
                         data.numberOfAlteredCases,
                         data.matchingGenePanelIds || [],
                         this.toggleModal,
-                        { marginLeft: cellMargin }
+                        {
+                            flexDirection: 'row-reverse',
+                            display: 'flex',
+                            marginRight: cellMargin,
+                        }
```
onursumer, line 336: Now I see this code segment is being copied 3 times already. Might be better to do a little refactoring and extract this code segment to a single variable/constant to prevent code duplication.

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -232,7 +232,12 @@ export class MultiSelectionTable extends React.Component<
                 headerRender: () => {
                     return (
                         <div
-                            style={{ marginLeft: cellMargin }}
+                            style={{
+                                marginLeft: cellMargin,
+                                flexDirection: 'row-reverse',
+                                display: 'flex',
+                                marginRight: cellMargin,
+                            }}
```
onursumer, line 240: Do we need this? I don't see any effect of this style on the column header. It looks same as before.

src/pages/studyView/table/MultiSelectionTable.tsx
alisman: really we should do this with a css class.   @aditygrg2 i think this would be quite easy to do.   i searched project for "row-reverse" and it only comes up in four places.  so that suggests that with a small amount of refactoring, we can accomplish this using a css class, which would be better for code quality and also optimization since this gets used on a lot of table cells.

alisman (changes requested):
Thank you very much for tackling this!  Please see comment about achieving this with a class instead of this inline style.  we should also update the existing places where we do this.

PR #4538, opened 2023-03-05
MSKTabs should not interpret tooltip clicks as clicks on tab buttons

jfkonecn:
# Description
Fix cBioPortal/cbioportal#9946 

- Removed prevent default call from MSKTabs so that clicking an anchor with an href isn't ignored
- Added target blank to WikiPathways link so that it'll open a new tab (I'm assuming this is the desired behavior)

# Why I don't have tests
I did my best to add a test to MSKTabs.spec.tsx, but I couldn't trigger the tool tip to be visible in the DOM. 😭
I wanted to trigger a click event on a tool tip anchor and assert that the url changed.
I'm willing to fix this, but I'd appreciate it if you could point me in the right direction.

# Other issues (willing to fix myself if you're okay with me doing so)
-  The NCI-PID link results in a [404 error](https://www.nlm.nih.gov/research/umls/sourcereleasedocs/current/NCI_PID/index.html
- The template for PR's has a broken link to the CONTRIBUTING.md file

# Checks

@adamabeshouse

### DIFF
--- a/src/shared/components/MSKTabs/MSKTabs.tsx
+++ b/src/shared/components/MSKTabs/MSKTabs.tsx
@@ -409,18 +409,6 @@ export class MSKTabs extends React.Component<IMSKTabsProps> {
 
                 const href = this.getTabHref(tab.props.id);
 
-                const linkContent = tab.props.linkTooltip ? (
-                    <DefaultTooltip
-                        overlay={tab.props.linkTooltip}
-                        mouseEnterDelay={0}
-                        placement="top"
-                    >
-                        <span>{tab.props.linkText}</span>
-                    </DefaultTooltip>
-                ) : (
-                    tab.props.linkText
-                );
-
                 pages[currentPage - 1].push(
                     <li
                         key={tab.props.id}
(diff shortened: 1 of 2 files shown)


src/shared/components/MSKTabs/MSKTabs.tsx
```diff
@@ -141,10 +141,9 @@ export class MSKTabs extends React.Component<IMSKTabsProps> {
 
     private tabClickHandlers = MemoizedHandlerFactory(
         (
-            e: React.MouseEvent<any>,
+            _: React.MouseEvent<any>,
```
onursumer, line 144: Do we really need this change? This is going to shadow the imported lodash variable `import _ from 'lodash';`. It's not recommended to introduce shadowed variables even if they don't break anything (just like in this case).

onursumer (changes requested):

onursumer:
Now clicking on a link in the tooltip also automatically switches to the corresponding pathway viewer. In other words, clicking on a tooltip link is also interpreted as clicking on a pathway tab pill itself. I think the code segment `e.preventDefault();` was introduced to prevent this kind of behavior. Not sure how critical it is though.

src/shared/components/MSKTabs/MSKTabs.tsx
jfkonecn: No, I'll revert it. I was trying to mark it as an unused variable, but I didn't think about lodash.

jfkonecn:
Do you want me to fix the NCI-PID link too?

onursumer:
> I see. Is it correct to say that the tabClickHandlers shouldn't be called if the user clicks on the tooltip?

Correct, a tooltip click is a different action than a tab click. Ideally we should find a way to make all tooltip links work without actually triggering a tab click event.

onursumer:
> Do you want me to fix the NCI-PID link too?

This is technically a separate issue. The link itself is not a valid link anymore. I am not sure what we should replace it with.

src/shared/components/MSKTabs/MSKTabs.tsx
alisman: yeah, typescript probably will complain if we don't specify the argument, even if we don't use it.  and "e" is fairly conventional for DOM events

alisman (changes requested):
see comments

jfkonecn:
Is everyone okay with me not adding tests?
To be clear I have no issue with adding tests I just couldn't figure out how to test a tooltip with enzyme

alisman:
@jfkonecn I believe that this PR is the [right way](https://github.com/cBioPortal/cbioportal-frontend/pull/4541) to hand the problem.   We want to keep the preventDefault but just make sure that click handler is more selective in what it responds to.

jfkonecn:
> @jfkonecn I believe that this [new PR is the right way](https://github.com/cBioPortal/cbioportal-frontend/pull/4541) to hand the problem. We want to keep the preventDefault but just make sure that click handler is more selective in what it responds to.

Personally it feels like the tabClickHandlers shouldn't be called at all since the click event is from the tooltip and not the tab itself. 

Obviously you have to maintain this code so I no issues with you rejecting my pull request. 

Thank you for reviewing this PR so quickly and being so respectful!

alisman:
@jfkonecn i looked at this one again and realized that you are right!  your way is correct and cleaner.  the tooltip should be the parent of the anchor so that the click handler is inside and that fixes our problem.  will merge this one.

alisman (approved):

PR #4539, opened 2023-03-07
Right Aligned numbers in study view

aditygrg2:
Fix https://github.com/cBioPortal/cbioportal/issues/10006

Sorry for accidentally closing the previous PR here https://github.com/cBioPortal/cbioportal-frontend/pull/4537

Right-aligned text for numbers in MultiSelectionTable

- Updated a CSS class 'pullRight' which when used now displays numeric data aligned to right in multiselection table exception being getFreqColumnRender() which takes style as one of its arguments, and there is no class argument, so we cannot use a class without changing the function.


src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -324,7 +324,11 @@ export class MultiSelectionTable extends React.Component<
                         data.numberOfAlteredCases,
                         data.matchingGenePanelIds || [],
                         this.toggleModal,
-                        { marginLeft: cellMargin }
+                        {
```
alisman, line 327: @aditygrg2 is this still necessary now that we have the class?

aditygrg2: Hi @alisman I have committed after updating the function. I was not doing it earlier as I thought the function might change some important things, sorry for not checking it earlier correctly. Now, the new commit solves this.

alisman (approved):

alisman:
@aditygrg2 thanks for this PR!  I just updates screenshots for end to end tests.  i'll merge this pending the test passing

PR #4541, opened 2023-03-08
MSKTabs should not fire on click of element in tooltip

alisman:
Fix cBioPortal/cbioportal#9946


onursumer (approved):
LGTM. Thanks for the fix!

PR #4549, opened 2023-03-16
Add compression in some GSVA tests

oplantalech:
Add compression in some GSVA tests to prevent future errors similar to #4529


pvannierop (changes requested):
I think you need to activate this in the portal.properties file used by all tests.

PR #4551, opened 2023-03-19
Selected genes list populates when there is only one group selected

jfkonecn:
Fix cBioPortal/cbioportal#10088

- There was a guard in the makeEnrichmentDataPromise function that was preventing the gene list from being populated

Why no test?
This is a network request based issue so it'll be difficult to create a non-brittle test

## Checks

@alisman


src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -1078,7 +1078,7 @@ export default abstract class ComparisonStore extends AnalysisStore
             if (
                 (this.alterationsEnrichmentDataRequestGroups.result &&
                     this.alterationsEnrichmentDataRequestGroups.result.length >
```
onursumer, line 1080: Maybe simplify it this way?
```suggestion
                this.alterationsEnrichmentDataRequestGroups.result?.length
```
Technically, we don't even need the check `> 0`.

jfkonecn: Personally I prefer to have the > 0 since it's more explicit, but I'm happy to change it if you want me to.

onursumer: We can keep the `> 0`. I don't have a strong preference.

onursumer: Ah, just realized if we keep the `> 0` then the above simplification doesn't work. TypeScript complains if we do

```JavaScript
this.alterationsEnrichmentDataRequestGroups.result?.length > 0
```

onursumer: Sorry for dragging this out but if we really want to be explicit we can also do this:

```JavaScript
!_.isEmpty(this.alterationsEnrichmentDataRequestGroups.result)
```

PR #4556, opened 2023-03-22
Add mutations table and fisher exact two-sided p value to comparison lollipop

gblaih:
Implements https://github.com/cBioPortal/cbioportal/issues/10012


packages/react-mutation-mapper/src/component/mutationMapper/FilterResetPanel.tsx
```diff
@@ -7,6 +7,7 @@ import styles from './filterResetPanel.module.scss';
 type FilterResetPanelProps = {
     resetFilters: () => void;
     filterInfo?: JSX.Element | string;
+    shiftClickMessage?: JSX.Element | string;
```
onursumer, line 10: I would name this prop `additionalInfo` or something like that. `FilterResetPanel` doesn't need to know what a `shift+click` is.

packages/react-mutation-mapper/src/component/mutationMapper/FilterResetPanel.tsx
```diff
@@ -34,14 +35,18 @@ export class FilterResetPanel extends React.Component<
                 data-test="filter-reset-panel"
             >
                 <span style={{ verticalAlign: 'middle' }}>
-                    {this.props.filterInfo}
+                    <span style={{ fontWeight: 'bold' }}>
```
alisman, line 38: use <strong> tag instead of inline style

packages/react-mutation-mapper/src/store/DefaultMutationMapperFilterApplier.ts
```diff
@@ -112,6 +115,14 @@ export class DefaultMutationMapperFilterApplier implements FilterApplier {
     ) {
         return applyDefaultMutationStatusFilter(filter, mutation);
     }
+
+    @autobind
```
alisman, line 119: if this is mutating state, should be `@action.bind`

packages/react-mutation-mapper/src/util/FilterUtils.ts
```diff
@@ -152,6 +153,20 @@ export function applyDefaultProteinImpactTypeFilter(
     );
 }
 
+export function applyDefaultProteinChangeFilter(
```
alisman, line 156: i see there is an establishe convention here but i think the naming isn't great.   the name ```applyDefaultProteinChangeFilter``` suggests something that actually applies it but this is just a test, so maybe ```shouldApplyDefaultProteinChangeFilter``` ?   If it's not too much work to change the names of these things, lets do it.

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
```diff
@@ -0,0 +1,86 @@
...
+}
+
+export const FisherExactTwoSidedTestLabel: React.FC<IFisherExactTwoSidedTestLabelProps> = observer(
+    ({
+        dataStore,
+        groups,
+        groupToProfiledPatientCounts,
+    }: IFisherExactTwoSidedTestLabelProps) => {
+        const mutationCountForActiveGeneGroupA = countUniqueMutations(
+            _.intersection(
```
alisman, line 27: we've had performance problems with intersect before.  look up the package "fast-array-intersect" which we use elsehwere in our project and i think will be must faster.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
         return null;
     }
 
+    protected formatPaginationStatusText = (text: string) => {
```
alisman, line 79: this should really be a free function (not a method of class) which accepts text and tableData as arguments.  would be much easier to test.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+        let dataStore = this.props.store.dataStore as MutationMapperDataStore;
+        const mutationsLabel =
+            _.flatten(dataStore.tableData).length === 1
+                ? 'Mutation'
+                : 'Mutations';
+
+        return text.includes('Mutations')
```
alisman, line 86: interesting (and yucky!) we should explain this with a comment if it must be.

alisman: yeah, we really need a comment on this whole function to explain what it does

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                    this.props.store.uniqueSampleKeyToTumorType
+                }
+                oncoKbCancerGenes={this.props.store.oncoKbCancerGenes}
+                pubMedCache={this.props.pubMedCache}
+                genomeNexusCache={this.props.genomeNexusCache}
+                genomeNexusMutationAssessorCache={
+                    this.props.genomeNexusMutationAssessorCache
+                }
+                dataStore={
+                    this.props.store.dataStore as MutationMapperDataStore
```
alisman, line 107: how come we need this casting?

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                }
+                initialSortColumn={'p-Value'}
+                initialSortDirection={'asc'}
+            />
+        );
+    }
+
+    @computed get rowDataByProteinChange() {
+        let dataStore = this.props.store.dataStore as MutationMapperDataStore;
+        let mutationsByProteinChange = _.values(
```
alisman, line 159: you should avoid nesting lodash.  use chaining instead. 
also, this seems strange because, unless i'm mistaken, will result in array of arrays?  where mutationsByPorteinChange suggests a map of proteinChange to mutations.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+        );
+        return getProteinChangeToMutationRowData(
+            mutationsByProteinChange,
+            this.mutatedCountsByProteinChangeForGroup,
+            this.props.groupToProfiledPatientCounts,
+            this.props.groups
+        );
+    }
+
+    @autobind
```
alisman, line 170: i doubt this needs autobind.  only if it's used as a callback by DOM

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -46,8 +76,119 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+    protected rowDataByProteinChangeForEnrichedGroups():
+        | {
+              [proteinChange: string]: ComparisonMutationsRow;
+          }
+        | {} {
+        return _.pickBy(this.rowDataByProteinChange, v => {
+            return this.significanceFilter
+                ? this._enrichedGroups.includes(v.enrichedGroup) &&
+                      v.qValue < 0.05
```
alisman, line 179: 0.05 should be a constant defined somewhere and we should explain in comment it's meaning.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
         );
         return map;
     }
+
+    @autobind
+    protected mutationsGroupedByProteinChangeForGroup(groupIndex: number) {
+        let allGroupedData = this.store.dataStore.allData;
```
alisman, line 363: try to avoid using ```let``` in general.  here, you could just do 
```const allGroupedData = groupDataByGroupFilters``` with third arg being this.store.data.store.allData, no?

File: src/pages/groupComparison/GroupComparisonMutationMapper.tsx [hunk 13]
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+            allGroupedData,
+            this.store.dataStore.applyFilter
+        );
+
+        let proteinChanges = _(allGroupedData[groupIndex].data)
+            .map((d: { proteinChange: any }[]) => d[0].proteinChange)
+            .uniq()
+            .value();
+
+        const filters = proteinChanges.map(value => ({
```
alisman, line 377: just combine this with chain above and use comments to explain/label the steps.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+        groupIndex: number
+    ): {
+        [proteinChange: string]: number;
+    } {
+        const map: { [proteinChange: string]: number } = {};
+
+        _.forIn(
+            this.mutationsGroupedByProteinChangeForGroup(groupIndex),
+            (v, k) => {
+                const uniqueMutations = _.uniqBy(v.data, d => d[0].patientId);
```
alisman, line 405: put in a comment here why it's uniq by patient

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                map[v.group] = uniqueMutations.length;
+            }
+        );
+
+        return map;
+    }
+
+    protected get fisherExactTwoSidedTestLabel(): JSX.Element | null {
```
alisman, line 413: should be @computed

alisman: also, why is return type allow null?  doesn't seem like that's possible?

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+                        options={this.options}
+                        value={this.selectedValues}
+                    />
+                </div>
+                <label className="checkbox-inline" style={{ marginRight: 7 }}>
+                    <input
+                        type="checkbox"
+                        checked={this.significanceFilter}
+                        onClick={this.toggleSignificanceFilter}
+                        data-test="SwapAxes"
```
alisman, line 452: give it a test handle that's a little more specific to feature.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+        });
+    }
+
+    @computed get selectedValues() {
+        return this._enrichedGroups.map(id => ({ value: id }));
+    }
+
+    @action.bound
+    onChange(values: { value: string }[]) {
+        this._enrichedGroups = _.map(values, datum => datum.value);
```
alisman, line 475: seems odd that we are doing two things in this handler.  updating enrichedGroups and also whatever onFilterOptionSelect does.  Like, shouldn't the former be included in the latter?

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+            this._enrichedGroups.length === 2 && !this.significanceFilter,
+            this.store.dataStore,
+            DataFilterType.PROTEIN_CHANGE,
+            PROTEIN_CHANGE_FILTER_ID
+        );
+    }
+
+    @action.bound
+    toggleSignificanceFilter() {
+        this.significanceFilter = !this.significanceFilter;
```
alisman, line 487: ditto.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +357,155 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
...
+            this.store.dataStore,
+            DataFilterType.PROTEIN_CHANGE,
+            PROTEIN_CHANGE_FILTER_ID
+        );
+    }
+
+    @action.bound
+    protected resetFilters() {
+        super.resetFilters();
+        this._enrichedGroups = this.props.groups.map(
```
alisman, line 500: again if _enrichedGroups qualifies as a filter then it's name should reflect that

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -128,16 +136,14 @@ export default class GroupComparisonMutationsTabPlot extends React.Component<
...
-                )
-            ) {
-                const mutationMapperStore = this.mutationMapperToolStore.getMutationMapperStore(
-                    this.props.store.activeMutationMapperGene!.hugoGeneSymbol
-                );
+            const mutationMapperStore = this.mutationMapperToolStore.getMutationMapperStore(
+                this.props.store.activeMutationMapperGene!.hugoGeneSymbol
+            );
+            if (mutationMapperStore) {
+                mutationMapperStore.uniqueSampleKeyToTumorType = this.props.store.uniqueSampleKeyToTumorType.result!;
```
alisman, line 146: this seems weird that we are taking this from one store and giving it to another like this.  can the logic not make use of uniqueSampleKeyToTumorType on ```this.props.store```?   Also, should ```uniqueSampleKeyToTumorType``` be awaited above?

src/shared/components/mutationMapper/MutationMapper.tsx
```diff
@@ -508,6 +525,10 @@ export default class MutationMapper<
         return null;
     }
 
+    protected get fisherExactTwoSidedTestLabel(): JSX.Element | null {
```
alisman, line 528: why does this alway return null?

src/shared/lib/FisherExactTestCalculator.ts
```diff
@@ -51,3 +51,49 @@ export function getCumulativePValue(
     }
     return p;
 }
+
+export function getTwoTailedPValue(
```
alisman, line 55: we definately need to test the crap out of this. did you make sure we didn't already have an implementation of this somewhere?

src/shared/components/mutationTable/column/PValueColumnFormatter.tsx
```diff
@@ -0,0 +1,53 @@
...
+        d: Mutation[]
+    ) {
+        const pValue = this.getPValueData(rowDataByProteinChange, d);
+
+        let content = (
+            <span
+                style={{
+                    whiteSpace: 'nowrap',
+                    fontWeight:
+                        rowDataByProteinChange[d[0].proteinChange].qValue < 0.05
```
alisman, line 42: here's that constant again

src/shared/components/mutationTable/column/LogRatioColumnFormatter.tsx
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    public static renderFunction(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        d: Mutation[]
+    ) {
+        let content = (
+            <div>{this.getLogRatioTextValue(rowDataByProteinChange, d)}</div>
```
alisman, line 36: do we have output a div inside a call?

src/shared/components/mutationTable/column/GroupMutatedCountPercentageColumnFormatter.tsx.tsx
```diff
@@ -0,0 +1,64 @@
...
+import { Mutation } from 'cbioportal-ts-api-client';
+import { ComparisonMutationsRow } from 'shared/model/ComparisonMutationsRow';
+
+export default class GroupMutatedCountPercentageColumnFormatter {
+    public static getMutatedCountData(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        groupIndex: number,
+        d: Mutation[]
```
alisman, line 12: can we just called d = mutations?

alisman: true for all these formatters

src/shared/components/mutationTable/column/GroupMutatedCountPercentageColumnFormatter.tsx.tsx
```diff
@@ -0,0 +1,64 @@
...
+    ) {
+        const rowData = rowDataByProteinChange[d[0].proteinChange];
+
+        return !!!groupIndex
```
alisman, line 16: !!!  ?

src/shared/components/mutationTable/column/EnrichedInColumnFormatter.tsx
```diff
@@ -0,0 +1,68 @@
...
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        d: Mutation[],
+        groups: ComparisonGroup[]
+    ) {
+        const rowData = rowDataByProteinChange[d[0].proteinChange];
+
+        const nameToGroup = _.keyBy(groups, g => g.name);
+        const significant = rowData.qValue < 0.05;
```
alisman, line 44: constant

src/shared/components/mutationTable/column/EnrichedInColumnFormatter.tsx
```diff
@@ -0,0 +1,68 @@
...
+        const groupColor =
+            significant && groups
+                ? nameToGroup[
+                      this.getEnrichedInData(rowDataByProteinChange, d).slice(4)
```
alisman, line 48: why slice(4)?

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
alisman, line 79: see if we can override this in the underlying library (rather than this kinda hacky editing of the initial output)

src/pages/groupComparison/GroupComparisonMutationsTabPlot.tsx
```diff
@@ -4,7 +4,7 @@ import { action, computed, observable, makeObservable } from 'mobx';
 import autobind from 'autobind-decorator';
```
alisman, line 4: Change name of this file/component becaues it's not just the plot.   Wrapper.  Content

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -364,6 +364,19 @@ export default class GroupComparisonStore extends ComparisonStore {
         },
     });
 
+    public readonly groupToProfiledPatientCounts = remoteData({
```
alisman, line 367: this is seems strange.  array would siffice instead of a map with integer keys.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,70 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            groupAMutatedPercentage: groupAMutatedPercentage,
+            groupBMutatedPercentage: groupBMutatedPercentage,
+            logRatio: logRatio,
+            pValue: pValue,
+            qValue: 0,
+            enrichedGroup: enrichedGroup,
+        };
+    });
+
+    rowData = _.sortBy(rowData, ['pValue']);
```
alisman, line 1121: don't use lodash string shortcut.  ```_.sortBy(rowData, r=>r.pValue)```

src/shared/components/mutationMapper/MutationMapper.tsx
alisman: make this general "bottomLabel"

alisman: plotFooter

src/shared/components/mutationTable/column/AlterationOverlapColumnFormatter.tsx
```diff
@@ -0,0 +1,131 @@
...
+                totalQueriedCases) *
+            100;
+        const group1Altered =
+            (rowData.groupAMutatedCount / totalQueriedCases) * 100;
+        const group2Altered =
+            (rowData.groupBMutatedCount / totalQueriedCases) * 100;
+
+        const alterationLanguage = 'alterations';
+
+        const overlay = () => {
```
alisman, line 62: might as well make a quick component out of this

src/shared/components/mutationTable/column/AlterationOverlapColumnFormatter.tsx
```diff
@@ -0,0 +1,131 @@
...
+            sampleIdentifiersForGroupA,
+            sampleSet
+        );
+        const sampleIdentifiersForGroupB = getSampleIdentifiers([groups[1]]);
+        const patientIdentifiersforGroupB = getPatientIdentifiers(
+            sampleIdentifiersForGroupB,
+            sampleSet
+        );
+
+        const totalQueriedCases =
```
alisman, line 37: unit test this

src/shared/components/mutationTable/column/GroupMutatedCountPercentageColumnFormatter.tsx.tsx
```diff
@@ -0,0 +1,64 @@
...
+    ) {
+        const rowData = rowDataByProteinChange[d[0].proteinChange];
+
+        const mutatedCount = this.getMutatedCountData(
+            rowDataByProteinChange,
+            groupIndex,
+            d
+        );
+        const percentage = formatPercentValue(
+            !!!groupIndex
```
alisman, line 36: triple

src/shared/lib/ILazyMobXTableApplicationDataStore.ts
```diff
@@ -149,7 +150,7 @@ export class SimpleGetterLazyMobXTableApplicationDataStore<T>
     }
 
     @computed get showingAllData() {
-        return this.tableData.length === this.allData.length;
+        return _.flatten(this.tableData).length === this.allData.length;
```
alisman, line 153: why is all data ungrouped?  it should be the number of rows that table can possibly display (agnostic to what is being displayed a row (a group of things or one thing or an array of one, whatever).

src/shared/lib/StoreUtils.ts
```diff
@@ -1156,14 +1156,19 @@ export function mergeDiscreteCNAData(
 }
 
 export function mergeMutations(
-    mutationData: MobxPromise<Mutation[]>,
+    mutationData: MobxPromise<Mutation[]> | Mutation[],
```
alisman, line 1159: this should accept only Mutation[]

src/shared/lib/StoreUtils.ts
```diff
@@ -1262,6 +1267,10 @@ export function generateStructuralVariantId(s: StructuralVariant): string {
         .join('_');
 }
 
+export function generateMutationIdByGeneAndProteinChange(m: Mutation): string {
+    return [m.gene.hugoGeneSymbol, m.proteinChange].join('_');
```
alisman, line 1271: string interpolation

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -7,32 +7,107 @@ import {
...
+        let itemsLabel =
+            dataStore.tableData.length === 1
+                ? `Protein Change with ${mutationsLabel}`
+                : `Protein Changes with ${mutationsLabel}`;
+
+        if (itemsLabel.length) {
+            // we need to prepend the space here instead of within the actual return value
+            // to avoid unnecessary white-space at the end of the string
+            itemsLabel = ` ${itemsLabel}`;
+        }
```
onursumer, line 108: you could also call `trim()` on the final string to get rid of extra whitespace.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -7,32 +7,107 @@ import {
...
+            _.flatten(dataStore.tableData).length === 1
+                ? 'Mutation'
+                : 'Mutations';
+
+        // text originally shows Mutation(s) based on rows, since the rows represent protein changes now, replace Mutation(s)
+        // with Protein Change(s) with {mutationsLabel}
+        let itemsLabel =
+            dataStore.tableData.length === 1
+                ? `Protein Change with ${mutationsLabel}`
+                : `Protein Changes with ${mutationsLabel}`;
```
onursumer, line 102: Do we need this much granularity? How about just 
```suggestion
        const itemsLabel = `Protein Change(s) with ${mutationsLabel}`;
```

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -7,32 +7,107 @@ import {
...
+                dataStore.itemsPerPage === PAGINATION_SHOW_ALL
+                    ? dataStore.visibleData.length
+                    : firstVisibleItemDisp + dataStore.visibleData.length - 1;
+        }
+
+        // if there are multiple mutations represented, use Mutations else Mutation in label
+        const mutationsLabel =
+            _.flatten(dataStore.tableData).length === 1
+                ? 'Mutation'
+                : 'Mutations';
```
onursumer, line 95: or maybe just `Mutation(s)`

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -7,32 +7,107 @@ import {
...
     constructor(props: IGroupComparisonMutationMapperProps) {
         super(props);
+        makeObservable(this);
+
+        this._selectedGroupsForEnrichedInFilter = this.props.groups.map(
+            group => group.nameWithOrdinal
+        );
+    }
+
+    @computed get paginationStatusText(): string {
```
onursumer, line 71: This is partially duplicate of https://github.com/cBioPortal/cbioportal-frontend/blob/9ce8d4e6c9216e86f7b8d00d653ccecc74f2faef/src/shared/components/lazyMobXTable/LazyMobXTable.tsx#L594

Not sure how much effort it would take to make this a utility function to share/reuse.

src/pages/groupComparison/GroupComparisonMutationMapper.tsx
```diff
@@ -214,4 +382,140 @@ export default class GroupComparisonMutationMapper extends MutationMapper<
         );
         return map;
     }
+
+    @autobind
+    protected mutationsGroupedByProteinChangeForGroup(groupIndex: number) {
```
onursumer, line 387: This is purely data related, so probably shouldn't be implemented inside a component. It feels like this should be a utility function, or a computed map (keyed by group index) within a store.

Seems like we already have similar functions here like `mutationsGroupedByProteinImpactTypeForGroup` and `mutationCountsByProteinImpactTypeForGroup`. Might be better to refactor all these together in a separate PR.

src/pages/groupComparison/GroupComparisonMutationTable.tsx
```diff
@@ -0,0 +1,310 @@
...
+            MutationTableColumnType.MUTATION_STATUS,
+            MutationTableColumnType.PROTEIN_CHANGE,
+            MutationTableColumnType.MUTATION_TYPE,
+            '(A) Group',
+            '(B) Group',
+            'Log2 Ratio',
+            'Enriched in',
+            'Alteration Overlap',
+            'P_VALUE',
+            'Q_VALUE',
```
onursumer, line 54: We should probably define these extra column names as constants/enums somewhere as well.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+        group: string;
+        data: any[];
+    }>,
+    profiledPatientsCounts: number[],
+    groups: ComparisonGroup[]
+): {
+    [proteinChange: string]: ComparisonMutationsRow;
+} {
+    let rowData = tableData.map(proteinChangeRow => {
+        let groupAMutatedCount: number =
```
onursumer, line 1086:
```suggestion
        const groupAMutatedCount: number =
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            getMutatedCountsByProteinChangeForGroup(
+                mutationsGroupedByProteinChangeForGroup,
+                0
+            )[proteinChangeRow[0].proteinChange] || 0;
+        let groupBMutatedCount: number =
```
onursumer, line 1091:
```suggestion
        const groupBMutatedCount: number =
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            getMutatedCountsByProteinChangeForGroup(
+                mutationsGroupedByProteinChangeForGroup,
+                1
+            )[proteinChangeRow[0].proteinChange] || 0;
+        let groupAMutatedPercentage: number =
```
onursumer, line 1096:
```suggestion
        const groupAMutatedPercentage: number =
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            (groupAMutatedCount / profiledPatientsCounts[0]) * 100;
+        let groupBMutatedPercentage: number =
```
onursumer, line 1098:
```suggestion
        const groupBMutatedPercentage: number =
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            (groupBMutatedCount / profiledPatientsCounts[1]) * 100;
+        let logRatio: number = Math.log2(
```
onursumer, line 1100:
```suggestion
        const logRatio: number = Math.log2(
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            groupAMutatedPercentage / groupBMutatedPercentage
+        );
+        let pValue: number = getTwoTailedPValue(
```
onursumer, line 1103:
```suggestion
        const pValue: number = getTwoTailedPValue(
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+            groupAMutatedCount,
+            profiledPatientsCounts[0] - groupAMutatedCount,
+            groupBMutatedCount,
+            profiledPatientsCounts[1] - groupBMutatedCount
+        );
+        let enrichedGroup: string =
```
onursumer, line 1109:
```suggestion
        const enrichedGroup: string =
```

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,92 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+        return {
+            proteinChange: proteinChangeRow[0].proteinChange,
+            groupAMutatedCount: groupAMutatedCount,
+            groupBMutatedCount: groupBMutatedCount,
+            groupAMutatedPercentage: groupAMutatedPercentage,
+            groupBMutatedPercentage: groupBMutatedPercentage,
+            logRatio: logRatio,
+            pValue: pValue,
+            qValue: 0,
+            enrichedGroup: enrichedGroup,
```
onursumer, line 1123:
```suggestion
            groupAMutatedCount,
            groupBMutatedCount,
            groupAMutatedPercentage,
            groupBMutatedPercentage,
            logRatio,
            pValue,
            qValue: 0,
            enrichedGroup,
```

src/pages/resultsView/mutation/ResultsViewMutationMapper.tsx
```diff
@@ -74,6 +74,10 @@ export default class ResultsViewMutationMapper extends MutationMapper<
         let filterInfo:
             | JSX.Element
             | string = `Showing ${dataStore.tableData.length} of ${dataStore.allData.length} mutations.`;
+        let shiftClickMessage: string =
```
onursumer, line 77:
```suggestion
        const shiftClickMessage: string =
```

src/shared/components/mutationTable/column/EnrichedInColumnFormatter.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+        const nameToGroup = _.keyBy(groups, g => g.nameWithOrdinal);
+        const significant = rowData.qValue < SIGNIFICANT_QVALUE_THRESHOLD;
+        const groupColor =
+            significant && groups
+                ? nameToGroup[
+                      this.getEnrichedInData(rowDataByProteinChange, mutations)
+                  ].color
+                : undefined;
+        let content = (
```
onursumer, line 54: we don't need a variable here

```suggestion
        return (
```

src/shared/components/mutationTable/column/GroupMutatedCountPercentageColumnFormatter.tsx.tsx
```diff
@@ -0,0 +1,64 @@
...
+    }
+
+    public static renderFunction(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        groupIndex: number,
+        mutations: Mutation[]
+    ) {
+        let content = (
```
onursumer, line 52:
```suggestion
        return (
```

src/shared/components/mutationTable/column/LogRatioColumnFormatter.tsx
```diff
@@ -0,0 +1,43 @@
...
+        );
+    }
+
+    public static renderFunction(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        mutations: Mutation[]
+    ) {
+        let content = (
```
onursumer, line 35:
```suggestion
        return (
```

src/shared/components/mutationTable/column/PValueColumnFormatter.tsx
```diff
@@ -0,0 +1,55 @@
...
+
+    public static renderFunction(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        mutations: Mutation[]
+    ) {
+        const pValue = this.getPValueData(rowDataByProteinChange, mutations);
+
+        let content = (
```
onursumer, line 38:
```suggestion
        return (
```

src/shared/components/mutationTable/column/AlterationOverlapColumnFormatter.tsx
```diff
@@ -0,0 +1,92 @@
...
+    getPatientIdentifiers,
+    getSampleIdentifiers,
+} from 'pages/groupComparison/GroupComparisonUtils';
+import { MiniOncoprint } from 'shared/components/miniOncoprint/MiniOncoprint';
+import { DefaultTooltip } from 'cbioportal-frontend-commons';
+import ComplexKeyMap from 'shared/lib/complexKeyDataStructures/ComplexKeyMap';
+import { ComparisonMutationsRow } from 'shared/model/ComparisonMutationsRow';
+import { AlterationOverlapOverlay } from './alterationOverlap/AlterationOverlapOverlay';
+
+export default class AlterationOverlapColumnFormatter {
```
onursumer, line 14: We probably don't need classes and static functions for this new column formatter and all other column formatters implemented in this PR. This was how we implemented all these column formatters long time ago, but I think a simpler functional approach should just work fine.

src/shared/components/mutationTable/column/QValueColumnFormatter.tsx
```diff
@@ -0,0 +1,54 @@
...
+
+    public static renderFunction(
+        rowDataByProteinChange: {
+            [proteinChange: string]: ComparisonMutationsRow;
+        },
+        mutations: Mutation[]
+    ) {
+        const qValue = this.getQValueData(rowDataByProteinChange, mutations);
+
+        let content = (
```
onursumer, line 38:
```suggestion
        return (
```

src/shared/lib/FisherExactTestCalculator.ts
onursumer: Also, if we have to keep this function can we replace for loops with forEach, map, etc wherever applicable?

onursumer (changes requested):

packages/react-mutation-mapper/src/util/FilterUtils.ts
onursumer: Good point. Name choice is not great. This function (and any other applyFilter function) actually just tests the mutation for a given filter and returns a boolean value. We should probably fix the naming in another PR though.

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
```diff
@@ -0,0 +1,100 @@
...
+        ).length;
+
+        const groupBMutatedCount = _.uniq(
+            _.flatten(
+                intersect([
+                    _.flatten(dataStore.tableData),
+                    _.flatten(dataStore.sortedFilteredGroupedData[1].data),
+                ])
+            ).map(m => m.patientId)
+        ).length;
```
onursumer, line 44: These are almost identical. Should we have a function, something like `countMutated`?

src/pages/groupComparison/GroupComparisonMutationTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+            <span id="mutation-annotation">
+                {AnnotationColumnFormatter.renderFunction(
+                    d,
+                    {
+                        hotspotData: this.props.hotspotData,
+                        myCancerGenomeData: this.props.myCancerGenomeData,
+                        oncoKbData: haveDifferentCancerTypes(
+                            d,
+                            this.resolveTumorType
+                        )
```
onursumer, line 274: We have redundant calls to the function `haveDifferentCancerTypes`. Instead we should just call it once, at the very beginning of the function. Something like

```JavaScript
const hasMultipleCancerTypes = haveDifferentCancerTypes(...)
```

src/pages/groupComparison/GroupComparisonMutationTable.tsx
```diff
@@ -0,0 +1,347 @@
...
+        this._columns[MutationTableColumnType.NUM_MUTATED_GROUP_A].order = 40;
+        this._columns[MutationTableColumnType.NUM_MUTATED_GROUP_B].order = 50;
+        this._columns[MutationTableColumnType.MUTATION_OVERLAP].order = 60;
+        this._columns[MutationTableColumnType.LOG_RATIO].order = 70;
+        this._columns[MutationTableColumnType.P_VALUE].order = 80;
+        this._columns[MutationTableColumnType.Q_VALUE].order = 90;
+        this._columns[MutationTableColumnType.ENRICHED_IN].order = 100;
+        this._columns[MutationTableColumnType.MUTATION_STATUS].order = 110;
+
+        //Adjust column visibility according to portal.properties
```
onursumer, line 337: Ideally we shouldn't be mentioning about `portal.properties` in the frontend codebase. But seems like we already did the same thing in `ResultsViewMutationTable` and `PatientViewMutationTable`. It's also a bit concerning that we keep duplicating certain segments of code in each new Mutation Table implementation. I guess this is also something we need to address later.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -1063,3 +1065,111 @@ const AlterationMenuHeader: React.FunctionComponent<{
...
+} {
+    const map: { [attribute: string]: number } = {};
+
+    _.forIn(getMutationsGroupedByAttributeForGroup(groupIndex), (v, k) => {
+        map[v.group] = v.data.length;
+    });
+    return map;
+}
+
+export function getMutatedCountsByAttributeForGroup(
```
onursumer, line 1155: What's the difference between this one and `getMutationCountsByAttributeForGroup`? Would it be better if we unify them with an additional `uniqueCount` flag or a `uniqueBy` function as a third parameter?

packages/react-mutation-mapper/src/component/column/Annotation.tsx
```diff
@@ -79,6 +80,7 @@ export type GenericAnnotationProps = {
     oncoKbContentPadding?: number;
     pubMedCache?: MobxCache;
     userDisplayName?: string;
+    haveDifferentCancerTypes?: boolean;
```
alisman, line 83: hasMultipleCancerTypes

src/shared/components/mutationTable/MutationTable.tsx
```diff
@@ -209,6 +210,13 @@ export enum MutationTableColumnType {
     DBSNP = 'dbSNP',
     GENE_PANEL = 'Gene panel',
     SIGNAL = 'SIGNAL',
+    NUM_MUTATED_GROUP_A = '(A) Group',
+    NUM_MUTATED_GROUP_B = '(B) Group',
```
onursumer, line 214: We may want these column names match the actual group names. Not sure if it is easy to implement though.

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+export const FisherExactTwoSidedTestLabel: React.FC<IFisherExactTwoSidedTestLabelProps> = observer(
+    ({
+        dataStore,
+        hugoGeneSymbol,
+        groups,
+        profiledPatientCounts,
+    }: IFisherExactTwoSidedTestLabelProps) => {
+        const groupAMutatedCount = _.uniq(
+            _.flatten(
```
alisman, line 29: simplify this code (try resist nesting when possible). and put a comment what operation we're doing.

src/shared/model/ComparisonMutationsRow.ts
```diff
@@ -0,0 +1,11 @@
+export interface ComparisonMutationsRow {
```
onursumer, line 1: I have mixed feelings about the naming. It's true that this model represents a single row in the `GroupComparisonMutationTable` which is a special summary/aggregation of multiple mutations. But conceptually it's still a mutation table. So, how about `GroupComparisonMutation`? We have a very similar concept in `SignalDB` where a single mutation actually represents multiple mutations: https://github.com/cBioPortal/cbioportal-frontend/blob/5be7d0425a0c25a3f6a5a7f895fbe948b58146eb/packages/genome-nexus-ts-api-client/src/generated/GenomeNexusAPI.ts#L750

onursumer (changes requested):
Looks good! Just requested a few minor changes.

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
```diff
@@ -0,0 +1,100 @@
...
+                                {`${
+                                    _.filter(
+                                        dataStore.tableDataGroupedByPatients,
+                                        d => d.length > 1
+                                    ).length
+                                } ${
+                                    _.filter(
+                                        dataStore.tableDataGroupedByPatients,
+                                        d => d.length > 1
+                                    ).length == 1
```
alisman, line 89: ===

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
```diff
@@ -0,0 +1,100 @@
...
+                                groups={groups}
+                            />
+                            <span data-test="patientMultipleMutationsMessage">
+                                {`${
+                                    _.filter(
+                                        dataStore.tableDataGroupedByPatients,
+                                        d => d.length > 1
+                                    ).length
+                                } ${
+                                    _.filter(
```
alisman, line 86: _.some()

src/pages/groupComparison/GroupComparisonUtils.tsx
gblaih: Mutated counts are the number of mutated patients (those with mutations) and mutation counts are number of mutations. I can make it one function with an additional parameter that's a boolean for mutation/mutated counts, and depending on that we do the appropriate action

src/shared/components/mutationTable/MutationTable.tsx
gblaih: Yea, it wouldn't be easy and I think keeping the general column name is good enough

src/pages/groupComparison/FisherExactTwoSidedTestLabel.tsx
alisman: i see.   i was wrong.  resolved.

inodb (approved):
Thanks @gblaih ! This looks good to me

As it also partially addresses: https://github.com/cBioPortal/cbioportal/issues/9943. Could we do a follow up PR that updates FAQ etc?

alisman (approved):

PR #4560, opened 2023-03-27
Fix and reduce flakiness in oncoprint tests

gblaih:
Edit oncoprint tests to reduce flakiness

Tests "oncoprint should return to non-clustered state correctly" and "should sort patients and samples by custom case list order correctly".


end-to-end-test/remote/specs/core/oncoprint.spec.js
```diff
@@ -519,12 +519,12 @@ describe('oncoprint', function() {
 
             clickQueryByGeneButton();
 
-            $('[data-test="dataTypePrioritySelector"]')
+            $('[data-test="molecularProfileSelector"]')
```
alisman, line 522: nice. this makes sense.  we changed that recenty.

alisman (approved):

PR #4564, opened 2023-03-31
Fix localdb keycloak docker image issue

fuzhaoyuan:
Fix cBioPortal/cbioportal#10110

- Fixed keycloak database parameters
- Switch back to official keycloak image, which runs through Rosetta translation on ARM platform but shouldn't hurt performance
- Remove platform detection for keycloak and fix it for mysql images


inodb (approved):
LGTM - thanks!

PR #4570, opened 2023-04-04
Update circleci config to parametize Workflow for e2e tests

haynescd:
Fix cBioPortal/cbioportal#10113

- Add Parameters to enable and disable e2e remote and localdb jobs

## Parameters Added to install_test workflow 
- remote_spec_pattern
  - Sets Spec Pattern for Remote e2e test ( Can specify one tests to run)
  - Default: ./remote/specs/**/*.spec.js
- localdb_spec_pattern
  - Sets Spec Pattern for localdb e2e test ( Can specify one tests to run)
  - Default: ./local/specs/**/*.spec.js
- enable_remote_e2e
  - Enables/Disables Remote E2E tests
  - Default: true
- enable_localdb_e2e
  - Enables/Disables local E2E tests
  - Default: true
-   manual_trigger_branch_env
  - When manually triggering jobs this parameter forces Branch_Env to equal manual_trigger_branch_env
  - Default: master

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,6 +2,25 @@
 #
 # Check https://circleci.com/docs/2.0/language-javascript/ for more details
 #
+version: 2.1
+
+parameters:
+  remote_spec_pattern:
+    type: string
+    default: "./remote/specs/**/*.spec.js"
+  localdb_spec_pattern:
+    type: string
+    default: "./local/specs/**/*.spec.js"
+  enable_remote_e2e:
+    type: boolean
+    default: true
+  enable_localdb_e2e:
+    type: boolean
+    default: true
+  manual_trigger_branch_env:
+    type: string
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -11,62 +24,68 @@ defaults: &defaults
...
+          - store_artifacts:
+              path: /tmp/repo/end-to-end-test/remote/junit/customReport.json
+              destination: /customReport.json
+          - store_artifacts:
+              path: /tmp/repo/end-to-end-test/remote/junit/errors
+              destination: /errors
+    - unless:
+        condition: << pipeline.parameters.enable_remote_e2e >>
+        steps:
+          - run: echo "Job didn't Run "
```
alisman, line 87: @haynescd like, here, can we just return nonzero so that the job will fail?

PR #4571, opened 2023-04-04
Impl. study search on Query Page using 'study tags'

BasLee:
This PR allows users to search studies on the index page by their study tags. 

The study tags are now fetched in the QueryStore, and added to the study node metadata as a flat string.
To search in study tags, the `studyTags` metadata property is added to `searchNodeFields`, so the textQueryUtils includes it when searching.

PR depends on this [backend PR](https://github.com/cBioPortal/cbioportal/pull/10116) that adds studyId to the study tags entity.

## Changes

- Add study tags to study list metadata and to full text search fields
- Allow full text search not only in study nodes but also their metadata
- Update API: add study ID to study tags entity
- Refactorings:
  - Rename `DefaultPhrase` to more meaningful `StringPhrase` (vs `ListPhrase`)
  - Rename `defaultNodeFields` to `searchNodeFields`
  - Remove unused `searchTerms` property from study list metadata

### DIFF
--- a/src/shared/components/query/CancerStudyTreeData.ts
+++ b/src/shared/components/query/CancerStudyTreeData.ts
@@ -14,6 +14,10 @@ export type CancerTypeWithVisibility = CancerType & { alwaysVisible?: boolean };
 
 export type CancerTreeNode = CancerTypeWithVisibility | CancerStudy;
 
+export function isStudy(node: CancerTreeNode): node is CancerStudy {
+    return !!(node as CancerStudy).studyId;
+}
+
 export type NodeMetadata = {
     isCancerType: boolean;
     isPriorityCategory: boolean;
@@ -24,7 +28,13 @@ export type NodeMetadata = {
     descendantCancerTypes: CancerTypeWithVisibility[];
     ancestors: CancerTypeWithVisibility[]; // in order of ascending distance from node, ending in root node
     siblings: CancerStudy[]; // descendants of highest level non-root ancestor
-    searchTerms: string; // all related text in a single string for easy searching
+    studyTags?: string;
+};
+
+export type StudyTags = {
+    studyId: string;
+    cancerStudyId: number;
+    tags: string;
 };
 
 export default class CancerStudyTreeData {
@@ -66,12 +76,14 @@ export default class CancerStudyTreeData {
     constructor({
         cancerTypes = [],
         studies = [],
+        allStudyTags = [],
         priorityStudies = {},
         virtualStudies = [],
         maxTreeDepth = 0,
     }: {
         cancerTypes: CancerTypeWithVisibility[];
         studies: CancerStudy[];
+        allStudyTags: StudyTags[];
         priorityStudies?: CategorizedConfigItems;
         virtualStudies?: VirtualStudy[];
         maxTreeDepth: number;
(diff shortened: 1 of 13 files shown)


src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -151,9 +163,17 @@ export default class CancerStudyTreeData {
                     );
                 }
 
-                let searchTerms = isCancerType
-                    ? node.name
-                    : [node.name, (node as CancerStudy).description].join('\n');
+                const foundStudyTags =
+                    allStudyTags.length &&
+                    isStudy(node) &&
+                    allStudyTags.find(
```
alisman, line 169: the naming of things here is kinda confusing. ```find``` will only ever return a single item, so should the variable not be singular?  but i guess then it has a ```tags``` collection?  so a single study can have multiple tags?

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -151,9 +163,17 @@ export default class CancerStudyTreeData {
...
-                let searchTerms = isCancerType
-                    ? node.name
-                    : [node.name, (node as CancerStudy).description].join('\n');
+                const foundStudyTags =
+                    allStudyTags.length &&
+                    isStudy(node) &&
+                    allStudyTags.find(
+                        t => t.studyId === (node as CancerStudy).studyId
+                    );
+                let studyTags = '';
```
alisman, line 172: ternary would be clearer

alisman: also, would be nice if we could do null or undefined instead of empty string

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -252,3 +264,12 @@ export default class CancerStudyTreeData {
         }
     }
 }
+
+function flattenTags(tags: any): string {
```
alisman, line 268: stringify tags?

src/shared/components/query/CancerStudyTreeData.ts
```diff
@@ -14,6 +14,10 @@ export type CancerTypeWithVisibility = CancerType & { alwaysVisible?: boolean };
 
 export type CancerTreeNode = CancerTypeWithVisibility | CancerStudy;
 
+export function isStudy(node: CancerTreeNode): node is CancerStudy {
```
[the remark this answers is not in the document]
BasLee: you can use it to check (sub)types, see [type guard](https://www.typescriptlang.org/docs/handbook/advanced-types.html#user-defined-type-guards)

src/shared/components/query/CancerStudyTreeData.ts
BasLee: Renamed in [2009473](https://github.com/cBioPortal/cbioportal-frontend/pull/4571/commits/2009473b6d81c094ba4246ee45947fd9d316e4fe)

src/shared/components/query/CancerStudyTreeData.ts
BasLee: Fixed in [2009473](https://github.com/cBioPortal/cbioportal-frontend/pull/4571/commits/2009473b6d81c094ba4246ee45947fd9d316e4fe)

src/shared/components/query/CancerStudyTreeData.ts
BasLee: Renamed in [2009473](https://github.com/cBioPortal/cbioportal-frontend/pull/4571/commits/2009473b6d81c094ba4246ee45947fd9d316e4fe)

alisman (approved):

PR #4576, opened 2023-04-05
changes for #9943 on frontend

Eric-Jiang1:
Fix cBioPortal/cbioportal#9943

- Fisher's exact test is giving the two-sided value
- changed mutual exclusivity tab and comparison tab to now show "2-sided fisher's exact test" instead of 1-sided to users
- added in comments to walk future developers through 

## Checks
- [x] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!  (This PR is not adding logic based on one or more clinical attributes in the backend repo)

## Notify reviewers
Doing this through Slack, I've been actively communicating with the team to resolve this issue.

### DIFF
--- a/src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
+++ b/src/pages/resultsView/enrichments/AlterationEnrichmentsTable.tsx
@@ -189,7 +189,7 @@ export default class AlterationEnrichmentTable extends React.Component<
                         Derived from{' '}
                         {_.values(this.props.data[0].groupsSet).length > 2
                             ? 'Chi-squared test'
-                            : 'one-sided Fisher Exact test'}
+                            : 'two-sided Fisher Exact test'}
                     </span>
                 ) : (
                     undefined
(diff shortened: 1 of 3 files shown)


dippindots (commented):
Great to see you added some comments, that's very helpful. One thing is although backend changes will make `comparison tabs` using the `two-sided Fisher Exact test`, the `MutualExclusivityTable.tsx` is still using the `one-sided Fisher Exact test` which is implemented in the front-end, so we might want to change this too.
It would also be great if we can have some unit tests to verify the newly implemented method.
Please see this [example](https://github.com/cBioPortal/cbioportal-frontend/blob/master/src/shared/lib/calculation/BenjaminiHochbergFDRCalculator.spec.ts) on how to create unit tests for a method.

inodb:
Thanks @Eric-Jiang1 ! We will add this to https://github.com/cBioPortal/cbioportal-frontend/pull/4556

PR #4577, opened 2023-04-07
RFC68: Impl. structural variants table in Study View

pvannierop:
Depends on: cBioPortal/cbioportal#10119
RFC document: https://docs.google.com/document/d/1_nhO5dpjIGVbnBzch34keHgQUrxYEFudvFDsybDkgrA/edit
Can be activated using `?featureFlags=STUDY_VIEW_STRUCT_VAR_TABLE`

# Description
This PR will implement the _Structural Variants_ table in Study View:
image

The _Structural Variants_ table allows to define cohorts based on genomic structural variant events where a _gene1_ and/or _gene2_ position is defined.

Using the 'hiden checkbox' system _gene1::gene2_, _gene1::_ or _::gene2_ structural variant queries can be staged for additional inspection in Results View:
image


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -375,6 +388,17 @@ export type OncokbCancerGene = {
     isCancerGene: boolean;
 };
 
+export type OncokbCancerStructVar = {
```
alisman, line 391: Not available from OncokbAPI?  looping in @zhx828

alisman: Also, if we must keep, lets put this in models folder

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -531,6 +545,68 @@ export function updateGeneQuery(
     return updatedQueries;
 }
 
+// This function acts as a toggle. If present in 'geneQueries', the query
```
alisman, line 548: can we move all these new functions into a separate utils file just for SV?  no time like the present to start breaking things up more.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -531,6 +545,68 @@ export function updateGeneQuery(
...
+    const gene2 = selectedGene2 || '*';
+
+    // Remove any SV alteration with the same genes
+    // (both upstream and downstream fusions are evaluated).
+    const updatedQueries = _.filter(geneQueries, (query: SingleGeneQuery) => {
+        const isStructVar = queryContainsStructVarAlteration(query);
+        const isUpstreamMatch =
+            query.gene === gene1 &&
+            query.alterations &&
+            !!_.find(
```
alisman, line 570: i think ```_.some``` is nicer than ```!!_.find```

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -778,9 +854,14 @@ export function getGenericAssayChartUniqueKey(
 const UNIQUE_KEY_SEPARATOR = ':';
 
 export function getUniqueKeyFromMolecularProfileIds(
-    molecularProfileIds: string[]
+    molecularProfileIds: string[],
+    chartType?: ChartTypeEnum
 ) {
-    return _.sortBy(molecularProfileIds).join(UNIQUE_KEY_SEPARATOR);
+    let ids = _.sortBy(molecularProfileIds);
+    if (chartType === ChartTypeEnum.STRUCTURAL_VARIANTS_TABLE) {
```
alisman, line 861: is this to distinguish the different types of SV charts?   i think this needs a comment.  also, if this ID is chart centric, maybe we change the name of the function to reflect that.  Since it not clear by chartType should be an input based on the current name

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3348,6 +3446,76 @@ export function geneFilterQueryFromOql(
     };
 }
 
+export function structVarFilterQueryFromOql(
```
alisman, line 3449: seems odd that we wouldn't reuse parsing from general oql handling for query functionality

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3878,3 +4047,32 @@ export function transformSampleDataToSelectedSampleClinicalData(
         .filter(item => item.uniqueSampleKey !== undefined);
     return clinicalDataSamples;
 }
+
+export type StructVarGene1Gene2 = {
```
alisman, line 4051: StructVarGenePair?

src/pages/studyView/UserSelections.tsx
```diff
@@ -809,6 +828,60 @@ export default class UserSelections extends React.Component<
         });
     }
 
+    private groupedStructVarQueries(
```
alisman, line 831: comment here explaining what "grouped" means.  seems slighly odd because i would expect this function to do some grouping, but it seems just a map from query to pilltag.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -151,6 +155,8 @@ export interface IChartContainerProps {
     selectedGenes?: any;
     cancerGenes: number[];
     onGeneSelect?: any;
+    selectedStructuralVariants?: any;
```
alisman, line 158: seems like we could improve these types

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+                    )
+                )
+            ),
+        };
+        return defaults[columnKey];
+    };
+
+    @computed
+    get maxNumberTotalCount() {
+        return _.max(this.tableData.map(item => item.totalCount));
```
alisman, line 480: this is a tiny quibble but ```_.maxBy(this.tableData, item => item.totalCount).totalCount```
is more efficient i think

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+    @computed
+    get maxNumberAlteredCasesColumn() {
+        return _.max(this.tableData!.map(item => item.numberOfAlteredCases));
+    }
+
+    @computed
+    get totalCountLocaleString() {
+        return this.maxNumberTotalCount === undefined
+            ? ''
+            : this.maxNumberTotalCount.toLocaleString();
```
alisman, line 492: i think we just want toString().   The use of toLocalString will make the output dependent on locale settings of browser.  I don't think we want that.  I believe we want consistent English-language formatting of numbers.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+
+    @autobind
+    isDisabled(uniqueKey: string) {
+        return _.some(this.preSelectedRowsKeys, key => key === uniqueKey);
+    }
+
+    @action.bound
+    onStructVarHover(rowId: string, isHovered: boolean) {
+        if (isHovered) {
+            this.hoveredStructVarTableRowIds.push(rowId);
```
alisman, line 638: in general i think it's a better practice to use immutable state.  i.e. don't mutate in place but update the whole array.  
```
this.hoveredStructVarTableRowIds = [rowId, ...this.hoveredStructVarTableRowIds]
```
Then you can make the array @observable.ref.   Ditto for the remove.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+            this.props.onChangeSelectedRows(
+                this.props.selectedRowsKeys.concat([uniqueKey])
+            );
+        } else {
+            this.props.onChangeSelectedRows(
+                _.xorBy(this.props.selectedRowsKeys, [record])
+            );
+        }
+    }
+    @observable private _selectionType: SelectionOperatorEnum;
```
alisman, line 660: i could be wrong but i thought you had to assign something here to get observable to work properly.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+        }
+        this.props.onChangeSelectedRows([]);
+    }
+
+    @computed get selectionType() {
+        if (this._selectionType) {
+            return this._selectionType;
+        }
+        switch (
+            (localStorage.getItem(this.props.tableType) || '').toUpperCase()
```
alisman, line 681: we just have a more descriptive key so we can understand what it is.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+        ) {
+            case SelectionOperatorEnum.INTERSECTION:
+                return SelectionOperatorEnum.INTERSECTION;
+            case SelectionOperatorEnum.UNION:
+                return SelectionOperatorEnum.UNION;
+            default:
+                return this.props.tableType === FreqColumnTypeEnum.DATA
```
alisman, line 688: maybe comment to explain why we default in this way?

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
```diff
@@ -0,0 +1,782 @@
...
+                        afterSorting={this.afterSorting}
+                        fixedTopRowsData={this.preSelectedRows}
+                        highlightedRowClassName={this.selectedRowClassName}
+                        showSetOperationsButton={true}
+                        numberOfSelectedRows={
+                            this.props.selectedRowsKeys.length
+                        }
+                    />
+                )}
+                {this.props.genePanelCache ? (
```
alisman, line 771: maybe add modalOpen check to the initial condition so that we don't bother rendering the modal in hidden state.   then make the "show" attribute of Modal always = true.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3348,6 +3446,76 @@ export function geneFilterQueryFromOql(
...
+    includeVUS?: boolean,
+    includeUnknownOncogenicity?: boolean,
+    selectedDriverTiers?: { [tier: string]: boolean },
+    includeUnknownDriverTier?: boolean,
+    includeGermline?: boolean,
+    includeSomatic?: boolean,
+    includeUnknownStatus?: boolean
+): StructVarFilterQuery {
+    if (!gene1Gene2Representation.match('::')) {
+        throw new Error(
```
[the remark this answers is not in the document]
pvannierop: It happens at more places in _StudyViewUtils.ts_ with similar technical error messages. ow should we handle these errors instead?

src/pages/studyView/StudyViewUtils.tsx
pvannierop: Yes, better. I replaced this and any other occurence in the code base.

src/pages/studyView/StudyViewPageStore.ts
zhx828: I don't think we have Structural Variant class but `AnnotateStructuralVariantQuery`

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -531,6 +545,68 @@ export function updateGeneQuery(
...
+    // (both upstream and downstream fusions are evaluated).
+    const updatedQueries = _.filter(geneQueries, (query: SingleGeneQuery) => {
+        const isStructVar = queryContainsStructVarAlteration(query);
+        const isUpstreamMatch =
+            query.gene === gene1 &&
+            query.alterations &&
+            !!_.find(
+                query.alterations,
+                (alt: FUSIONCommandUpstream | FUSIONCommandDownstream) =>
+                    alt.alteration_type === 'downstream_fusion' &&
```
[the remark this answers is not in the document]
pvannierop: I have added 2 new constants in oqlfilter.ts:

```
export const STUCTVARDownstreamFusion = 'downstream_fusion';
export const STUCTVARUpstreamFusion = 'upstream_fusion';
```

src/pages/studyView/StudyViewUtils.tsx
pvannierop: Indeed. The unique table key for a chart was derived from the molecular profile name(s) of the data shown in that table. 

1. This method is chart centric as you mention. I propose to rename this method to:

`getUniqueChartKeyFromMolecularProfileIds`

2. ChartType is an optional input. It is only used in my implemention for the new SV table because when different tables use the same data (same molecular profile ids), this strategy no longer works. However, I think that we better incorporate the CartTTypeEnum in the key like so:

```
export function getUniqueChartKeyFromMolecularProfileIds(
    molecularProfileIds: string[],
    chartType?: ChartTypeEnum
) {
    let ids = _.sortBy(molecularProfileIds);
    let returnValue = ids.join(UNIQUE_KEY_SEPARATOR);
    return chartType ? chartType + UNIQUE_KEY_SEPARATOR + returnValue :  returnValue;
}
```

3. How would you feel about making `chartType` a required param? This way is is also clear to developers in the future that it is chart centric.

src/pages/studyView/StudyViewUtils.tsx
pvannierop: Ok, this is possible but not straight forward. I had to convert the gene1 and gene2 values to a parsed OQL object and then convert it to a OQL string using the unparseOQLQueryLine function in _oqlfilter.ts_.

```
    const gene1 = query.gene1Query.hugoSymbol || '';
    const gene2 = query.gene2Query.hugoSymbol || '';
    // Translate to OQL SingleGeneQuery object.
    const parsed_oql_line: SingleGeneQuery = gene1 ?
        { gene: gene1, alterations: [{ gene: gene2, alteration_type: STUCTVARDownstreamFusionStr, modifiers: [] }]}:
        { gene: gene2, alterations: [{ gene: gene1, alteration_type: STUCTVARUpstreamFusionStr, modifiers: [] }]};
    return unparseOQLQueryLine(parsed_oql_line);
```

Do think this is better?

src/pages/studyView/StudyViewUtils.tsx
pvannierop: Yes, good suggestion. Changed.

src/pages/studyView/UserSelections.tsx
pvannierop: I am adhering to the naming convention in the UserSelections.ts file. I think that this discussion is out of scope for this PR.

src/pages/studyView/charts/ChartContainer.tsx
pvannierop: Do you have suggestions?

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
pvannierop: Updated!

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
pvannierop: This is code duplication with MultiSelectionTable.tsx. This specific proposal is out of scope for this PR I think. However, it highlights the need for code-deduplication.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
pvannierop: This code is again directly copied from _MultiSelectionTable.tsx_. I think what you propose is not needed.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
pvannierop: This code is again directly copied from MultiSelectionTable.tsx.

src/pages/studyView/table/StructuralVariantMultiSelectionTable.tsx
pvannierop: This code is again directly copied from MultiSelectionTable.tsx.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -1916,6 +1944,11 @@ export class StudyViewPageStore
         GeneFilterQuery[][]
     >();
 
+    @observable.ref private _structVarFilterSet = observable.map<
```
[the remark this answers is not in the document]
alisman: actually, i see below at 2059, you are not using as immuntable, so this shouldn't need to be observable.ref

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2362,6 +2411,13 @@ export class StudyViewPageStore
         );
     }
 
+    @computed get filterStructVarsTableByCancerGenes(): boolean {
```
alisman, line 2414: this doesn't need to be computed because we aren't saving any computation by doing so.  i just want to make sure it's understood that making it computed does not change it's observability.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6115,13 +6303,14 @@ export class StudyViewPageStore
         }
 
         if (!_.isEmpty(this.structuralVariantProfiles.result)) {
-            const uniqueKey = getUniqueKeyFromMolecularProfileIds(
+            let uniqueKey = getUniqueKeyFromMolecularProfileIds(
```
alisman, line 6306: it seems odd that we make a uniqueKey, use it at 6312, then, conditionally update it at line 6323.

alisman: if in fact that's the intention, then it's more like we are creating 2 independent keys, and the code should reflect that

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6812,10 +7033,11 @@ export class StudyViewPageStore
             }
         }
         if (!_.isEmpty(this.structuralVariantProfiles.result)) {
-            const uniqueKey = getUniqueKeyFromMolecularProfileIds(
+            let uniqueKey = getUniqueKeyFromMolecularProfileIds(
```
alisman, line 7036: ditto here

src/pages/studyView/charts/ChartContainer.tsx
alisman: I mean, these are quite specific types are they not?   just because 'any' got through before, lets not dig deeper.

src/pages/studyView/StudyViewPageStore.ts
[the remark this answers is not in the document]
BasLee: Fixed in [1cc6c5c](https://github.com/cBioPortal/cbioportal-frontend/pull/4577/commits/1cc6c5c31419dff689737a8c24242de7e719823d)

src/pages/studyView/charts/ChartContainer.tsx
BasLee: Fixed in [1cc6c5c](https://github.com/cBioPortal/cbioportal-frontend/pull/4577/commits/1cc6c5c31419dff689737a8c24242de7e719823d)

PR #4587, opened 2023-04-18
:wrench: Add Configurable Param for max number of samples in study vi…

haynescd:
Fix cBioPortal/cbioportal#10125

Add studyview.max_samples_selected property to make Study View limit configurable. Default is set to 100,000


alisman (approved):

alisman:
@haynescd i think we decided that we would not impose the limit when only ONE study was selected.

alisman (approved):

PR #4590, opened 2023-04-20
Paginated clinical table by re-implementing data store

BasLee:
Paginate the clinical table using original (lazy mobx) table components

Requires [backend PR](https://github.com/cBioPortal/cbioportal/pull/10211/files) changes

## Problem
Current table components use a data store that fetches all data from the backend using a very large page size, and then showing a part of the fetches data on a page (like it has been paginated).
This makes the initial loading time larger than needed, and possibly requires downloading large amounts of data.

## Solution
In this poc the data store interface used by tables has been re-implemented in such a way that only the displayed page is loaded. 
The old store interface stays largely intact, so no changes to existing tables are needed (or should be needed). 

### Frontend
The clinical table component is not passed all the data but now receives the new `PaginatedDataStore`. When the user browses to the next page, the data store fetches a new page from the backend. The data store has a new `PageStore` (`ClinicalTablePageStore`) which contains most of the pagination logic.

### Backend
Add a `total-count` header that returns the total number of items (`sampleClinicalData`) of the paginated response. To add the total, a new cached count query also needed to be added. 

## TODO
  - [x] Use pagination headers instead of spring page interface 
  - [x] Do we want to continue in this direction? Y
  - [X] Fix sorting (by subtype)
  - [X] Clean up pagination stores
  - [X] Reset to first page when sorting, filtering, etc.
  - [x] Optimize 'show more button' query: fetch only the new part and amend with already fetched data
  - [ ] Fix failing lazymobxtable unit test
  - [ ] Fix filtering by sample/patient IDs

### DIFF
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -32,12 +32,12 @@ import * as React from 'react';
 import { buildCBioPortalPageUrl } from '../../shared/api/urls';
 import { BarDatum } from './charts/barChart/BarChart';
 import {
+    BinMethodOption,
     GenericAssayChart,
     GenomicChart,
-    XvsYScatterChart,
     XvsYChartSettings,
+    XvsYScatterChart,
     XvsYViolinChart,
-    BinMethodOption,
 } from './StudyViewPageStore';
 import { StudyViewPageTabKeyEnum } from 'pages/studyView/StudyViewPageTabs';
 import { Layout } from 'react-grid-layout';
@@ -84,6 +84,7 @@ import { getServerConfig } from 'config/config';
 import joinJsx from 'shared/lib/joinJsx';
 import { BoundType, NumberRange } from 'range-ts';
 import { ClinicalEventTypeCount } from 'cbioportal-ts-api-client/dist/generated/CBioPortalAPIInternal';
+import { TablePaginationParams } from 'shared/components/lazyMobXTable/TablePaginationParams';
 
 // Cannot use ClinicalDataTypeEnum here for the strong type. The model in the type is not strongly typed
 export enum ClinicalDataTypeEnum {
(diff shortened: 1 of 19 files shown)


packages/cbioportal-ts-api-client/src/generated/CBioPortalAPIInternal.ts
```diff
@@ -196,6 +190,30 @@ export type ClinicalDataFilter = {
 
         'values': Array < DataFilterValue >
 
+};
+export type ClinicalDataPage = {
```
alisman, line 194: seems odd to me that the API would have a type for ClinicalDataPage.   I would think at the API level that it should handle pagination in the same way as it does for GET requests.  Though one problem I see with the GET pagination is that it doesn't tell you in the response how many total items there are.   Let's discuss.

src/shared/components/lazyMobXTable/LazyMobXTableStore.tsx
```diff
@@ -0,0 +1,615 @@
...
+        if (itemsLabel.length) {
+            // we need to prepend the space here instead of within the actual return value
+            // to avoid unnecessary white-space at the end of the string
+            itemsLabel = ` ${itemsLabel}`;
+        }
+
+        return `Showing ${firstVisibleItemDisp}-${lastVisibleItemDisp} of ${this.dataStore.totalElements}${itemsLabel}`;
+    }
+
+    @computed get tds(): JSX.Element[][] {
```
alisman, line 395: this probalby pre-existed this PR, but we should try to get JSX out of store.  it's just "wrong"

src/shared/components/lazyMobXTable/PaginatedDataStore.tsx
```diff
@@ -0,0 +1,140 @@
...
+    PageStore,
+} from 'shared/components/lazyMobXTable/PageStore';
+
+export type PageParams = {
+    direction: PageDirection;
+    pageNumber: number;
+    pageSize: number;
+};
+
+export class PaginatedDataStore<T>
```
alisman, line 22: make more clear that this this is server side pagination

src/shared/components/lazyMobXTable/PaginatedDataStore.tsx
```diff
@@ -0,0 +1,140 @@
+import { action, computed, makeObservable, observable } from 'mobx';
```
alisman, line 1: put a nice big comment explaining that this is a "facade" to trainslate the page store into the data store expected by LazyMobxTable

src/shared/components/lazyMobXTable/LazyMobXTableStore.tsx
BasLee: yes, this code already existed

src/shared/components/lazyMobXTable/PaginatedDataStore.tsx
BasLee: Added in [44ed633](https://github.com/cBioPortal/cbioportal-frontend/pull/4590/commits/44ed633252d40c7bb047f11fb4bc7838e07e7db7)

src/shared/components/lazyMobXTable/PaginatedDataStore.tsx
BasLee: Added in [44ed633](https://github.com/cBioPortal/cbioportal-frontend/pull/4590/commits/44ed633252d40c7bb047f11fb4bc7838e07e7db7)

src/shared/lib/ILazyMobXTableApplicationDataStore.ts
```diff
@@ -186,6 +190,19 @@ export class SimpleGetterLazyMobXTableApplicationDataStore<T>
             'dataFilter' | 'dataSelector'
         >(this);
     }
+
+    @computed get totalElements() {
```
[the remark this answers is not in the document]
BasLee: Renamed in [d93bf78](https://github.com/cBioPortal/cbioportal-frontend/pull/4590/commits/d93bf7860427cff8831b150422b6ed26ec08fd5c)

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3002,45 +3003,86 @@ export function updateSavedUserPreferenceChartIds(
...
 
+export type ClinicalDataBySampleId = {
+    [sampleId: string]: { [attributeId: string]: string };
+};
+
 export async function getAllClinicalDataByStudyViewFilter(
     studyViewFilter: StudyViewFilter
-): Promise<{ [sampleId: string]: { [attributeId: string]: string } }> {
+): Promise<ClinicalDataBySampleId> {
+    console.log('getAllClinicalDataByStudyViewFilter', studyViewFilter);
```
alisman, line 3013: console

BasLee: Fixed in [7aac574](https://github.com/cBioPortal/cbioportal-frontend/commit/7aac5743da3f29903770c21cbc99cd20905a65bf)

PR #4593, opened 2023-04-21
Add a configurable legend to timeline tracks

alisman:
image

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/Timeline.tsx
+++ b/packages/cbioportal-clinical-timeline/src/Timeline.tsx
@@ -145,7 +145,7 @@ function handleMouseEvents(
 }
 
 const hoverCallback = (
-    e: React.MouseEvent<any>,
+    e: React.MouseEvent<Element, MouseEvent>,
     styleTag: MutableRefObject<null>,
     uniqueId: string
 ) => {
@@ -168,12 +168,29 @@ const hoverCallback = (
                      #${uniqueId} .tl-timeline .tl-track:nth-child(${trackIndex +
                         1}) .tl-track-highlight {
                         opacity: 1 !important;
-                    }
+                     }
+                     
+                     #${uniqueId} .tl-tracklegend:nth-of-type(${trackIndex +
+                        1}) {
+                        display:block !important;
+                     }
+                     
                 `);
                 }
                 break;
             default:
-                jQuery(styleTag.current!).empty();
+                // when mouse is moving INTO the tooltip, we do not want to abide the
+                // mouseleave event. we treat it as if it's part of the track element
+                if (
+                    e.type === 'mouseleave' &&
+                    (e?.relatedTarget as Element).getAttribute('class') ===
+                        'arrow'
+                ) {
+                    break;
+                } else {
+                    // get rid of the above created inline style
+                    jQuery(styleTag.current!).empty();
+                }
                 break;
         }
     }
--- a/packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
@@ -10,6 +10,8 @@ import CustomTrack, { CustomTrackSpecification } from './CustomTrack';
 import { TICK_AXIS_HEIGHT } from './TickAxis';
 import { useObserver } from 'mobx-react-lite';
 import { getBrowserWindow } from 'cbioportal-frontend-commons';
+import ReactDOM from 'react-dom';
+import { TimelineTrackSpecification } from './types';
 
 export interface ITimelineTracks {
     store: TimelineStore;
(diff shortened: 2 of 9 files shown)


alisman:
@inodb Here is a track specific legend.  To start with this must be configured in code on an instance basis.   Since it seems people like to assign colors to timeline events in the data itself, we could create a dynamic legend based on whatever data was present.  It might necessitate people putting redundant legend labels on every event.  

Re the order of attributes in tooltips, they are alphabetical.  We stuck the dates at the bottom, but otherwise, we were sorting them.  This PR removes that sorting which presumably would make the order reflect the original data files.  However, I think @tmazor at one point requested that they be sorted alphabetically.  So not sure what to do.  We could make the order configurable in code, but we're stating to get pretty configuration heavy!

Re the order of the tracks, I would suggest that we make the status tracks into a track group (would be done in data), to avoid them getting intermingled with other not status tracks.

inodb (commented):
Looks great!

Feedback:

- mousing over the tooltip makes the legend disappears

>Here is a track specific legend. To start with this must be configured in code on an instance basis. Since it seems people like to assign colors to timeline events in the data itself, we could create a dynamic legend based on whatever data was present. It might necessitate people putting redundant legend labels on every event.

Yeah maybe we can stick with hardcoding this one legend for now for "Assessment" tracks exclusively. Haven't had requests about adding other legends yet

> Re the order of attributes in tooltips, they are alphabetical. We stuck the dates at the bottom, but otherwise, we were sorting them. This PR removes that sorting which presumably would make the order reflect the original data files. However, I think @tmazor at one point requested that they be sorted alphabetically. So not sure what to do. We could make the order configurable in code, but we're stating to get pretty configuration heavy!

Yeah reflecting data order would be ideal I think. You get what you upload

> Re the order of the tracks, I would suggest that we make the status tracks into a track group (would be done in data), to avoid them getting intermingled with other not status tracks

@ritikakundra is this feasible? Can we change the timeline files on our end somehow? Or can Sage point us to the scripts that do the data mangling so we can update it?

tmazor:
The legend is awesome! My only comment on that is the same as Ino's, that it keeps disappearing when I pull up a tooltip.

In terms of the order of attributes, alphabetical order definitely sounds like something I would have requested. I still kinda prefer that, but I can also see the argument for just reflecting the data as it is provided and I'm ok with that. However, I think the reason we got on this topic now is that I think it would be clearer if the attribute that leads to the coloring were the first one in the tooltip, which is still not the case. So then is that something we can update in the data?

ritikakundra:
@inodb just spoke to @alisman regarding STATUS. He said it might be feasible to configure it in the code itself but will let me know if data needs to be updated.

packages/cbioportal-clinical-timeline/src/Timeline.tsx
```diff
@@ -168,12 +168,29 @@ const hoverCallback = (
...
                 `);
                 }
                 break;
             default:
-                jQuery(styleTag.current!).empty();
+                // when mouse is moving INTO the tooltip, we do not want to abide the
+                // mouseleave event. we treat it as if it's part of the track element
+                if (
+                    e.type === 'mouseleave' &&
+                    (e?.relatedTarget as Element).getAttribute('class') ===
```
onursumer, line 186: why do we need `e?` here but not `e?.type` on the line above?

packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
```diff
@@ -38,16 +40,22 @@ export const TimelineTracks: React.FunctionComponent<ITimelineTracks> = observer
...
+                                    <TimelineTrack
+                                        limit={store.trimmedLimit}
+                                        trackData={track.track}
+                                        getPosition={store.getPosition}
+                                        handleTrackHover={handleTrackHover}
+                                        store={store}
+                                        y={y}
+                                        height={track.height}
+                                        width={width}
+                                    ></TimelineTrack>
```
onursumer, line 53:
```suggestion
                                    />
```

packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
```diff
@@ -38,16 +40,22 @@ export const TimelineTracks: React.FunctionComponent<ITimelineTracks> = observer
...
+                                    <TimelineTrackLegend
+                                        y={y + 20}
+                                        track={track.track}
+                                    ></TimelineTrackLegend>
```
onursumer, line 57:
```suggestion
                                    />
```

packages/cbioportal-clinical-timeline/src/types.ts
```diff
@@ -39,6 +46,11 @@ export enum TimelineTrackType {
 
 export type TimeLineColorGetter = (e: TimelineEvent) => string | void;
 
+export type LegendItem = {
```
onursumer, line 49: Maybe name it `TimelineLegendItem` for consistency?

src/pages/patientView/timeline/timeline_helpers.tsx
```diff
@@ -24,6 +24,7 @@ import { ISampleMetaDeta } from 'pages/patientView/timeline/TimelineWrapper';
 import { ClinicalEvent } from 'cbioportal-ts-api-client';
 import { getColor } from 'cbioportal-frontend-commons';
 import ReactMarkdown from 'react-markdown';
+import { LegendItem } from 'cbioportal-clinical-timeline/src';
```
onursumer, line 27: This was also caught by `ci/circleci: check_incorrect_import_statements`

```suggestion
import { LegendItem } from 'cbioportal-clinical-timeline';
```

onursumer (commented):
Looks good. Added a few minor suggestions.

Also, there is an incorrect import statement caught by the tests.

packages/cbioportal-clinical-timeline/src/Timeline.tsx
alisman: because it will definately have a type but not necessarily a relatedTarget.

PR #4598, opened 2023-04-26
Filter Clinical Attribute Charts when priority is -1

haynescd:
Fix cBioPortal/cbioportal#10133

Update chartMetaSet getter to ignore clinical Attribute charts when priority is < 0


dippindots:
@inodb @haynescd I checked the public database and GENIE database, and we currently don't have any attributes that have a value of priority less than 0. So it should be fine to use `-1` to remove the attribute.

dippindots (approved):
LGTM, it's very straightforward. Just one more question regarding the clinical tab in the study view, are we also gonna hide those attributes from those columns too?

haynescd:
Ok, awesome. I am going to make a localdb update to add attributes with priority of -1 so I can test this feature.

@dippindots  Are you talking about the clinical tab under "Add Chart" Button?

dippindots:
@haynescd For example, in this study: https://www.cbioportal.org/study/clinicalData?id=msk_impact_2017
Just wanted to know if we will also hide `priority=-1` attributes in this table
image

alisman (approved):

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6026,16 +6026,19 @@ export class StudyViewPageStore
...
-                    displayName: attribute.displayName,
-                    uniqueKey: uniqueKey,
-                    dataType: getChartMetaDataType(uniqueKey),
-                    patientAttribute: attribute.patientAttribute,
-                    description: attribute.description,
-                    priority: getPriorityByClinicalAttribute(attribute),
-                    renderWhenDataChange: false,
-                    clinicalAttribute: attribute,
-                };
+                const priortiy = getPriorityByClinicalAttribute(attribute);
```
alisman, line 6029: @haynescd just noticed a little typo in that variable name

haynescd: Hilarious! I cannot spell!!

PR #4602, opened 2023-04-27
:wrench: Add configurability for StudyDownloadUrl

haynescd:
Fix cBioPortal/cbioportal#10135

Update StudyDownloadUrl to be configurable via portal properties.

## StudyDownloadUrl Design
This url needs to point to the location of downloadable studies and study_list.json.
- The downloadable studies need to have the extension .tar.gz

### Example
Url - https://cbioportal-datahub.s3.amazonaws.com/

- https://cbioportal-datahub.s3.amazonaws.com/study_list.json (List available studies within s3 bucket ex: study_id_a, study_id_b)
- https://cbioportal-datahub.s3.amazonaws.com/study_id_a.tar.gz
- https://cbioportal-datahub.s3.amazonaws.com/study_id_b.tar.gz

### DIFF
--- a/src/pages/staticPages/datasetView/DatasetPage.tsx
+++ b/src/pages/staticPages/datasetView/DatasetPage.tsx
@@ -11,7 +11,6 @@ import LoadingIndicator from 'shared/components/loadingIndicator/LoadingIndicato
 import Helmet from 'react-helmet';
 import request from 'superagent';
 import { getStudyDownloadListUrl } from 'shared/api/urls';
-import { AppStore } from 'AppStore';
 
 export class DatasetPageStore {
     readonly data = remoteData({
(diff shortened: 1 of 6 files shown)


src/pages/staticPages/datasetView/DatasetPage.tsx
```diff
@@ -20,12 +19,11 @@ export class DatasetPageStore {
         },
     });
 
-    readonly downloadList = remoteData(() => {
-        if (getServerConfig().app_name === 'public-portal') {
```
inodb, line 24: before we were only doing the HTTP request on public portal, maybe it's good to keep it that way? I.e. maybe we can do something like when `study_download_url` is not defined, we don't even fire the request?

inodb (commented):
This looks good! Made a small comment

src/pages/staticPages/datasetView/DatasetPage.tsx
haynescd: Will do!!

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2747,7 +2747,9 @@ export class StudyViewPageStore
     readonly hasRawDataForDownload = remoteData<boolean>({
```
haynescd, line 2747: Want to look and see why this method is here

haynescd:
The e2e test "study should have the raw data available" fails because the property is defaulted to empty so this is expected

alisman:
@haynescd for the broken e2e test, you can always mock the configuration in test.

alisman (approved):

PR #4616, opened 2023-05-09
:sparkles: Add PatientIdentifiers URL Parameter

haynescd:
Fix cBioPortal/cbioportal#10081

The Study View Page currently has a way to construct filterJSON URL with sample IDs:

[https://cbioportal.org/study/summary?id=msk_impact_2017#filterJson={"sampleIdentifiers":[{"studyId":"msk_impact_2017","sampleId":"P-0000004-T01-IM3"}]}](https://cbioportal.org/study/summary?id=msk_impact_2017#filterJson=%7B%22sampleIdentifiers%22:%5B%7B%22studyId%22:%22msk_impact_2017%22,%22sampleId%22:%22P-0000004-T01-IM3%22%7D%5D%7D)

Update StudyViewPageStore to handle the following URL Parameter

JSON that is now supported within filterJson
`{"patientIdentifiers": [{"studyId": "test_studyId", "patientId": "test_patientid"}]}`

Ex:
[https://cbioportal.org/study/summary?id=msk_impact_2017#filterJson={"patientIdentifiers":[{"studyId":"msk_impact_2017","patientId":"P-0000004"}]}](https://cbioportal.org/study/summary?id=msk_impact_2017#filterJson=%7B%22patientIdentifiers%22:%5B%7B%22studyId%22:%22msk_impact_2017%22,%22patientId%22:%22P-0000004%22%7D%5D%7D)

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -57,6 +57,7 @@ import {
     PatientTreatmentRow,
     ResourceData,
     Sample,
+    SampleFilter,
     SampleIdentifier,
     SampleMolecularIdentifier,
     SampleTreatmentRow,
@@ -387,6 +388,15 @@ export enum BinMethodOption {
 enum CustomDataTypeEnum {
     CATEGORICAL = 'CATEGORICAL',
     NUMERICAL = 'NUMERICAL',
+}
+
+interface PatientIdentifierFilter {
+    patientIdentifiers: PatientIdentifier[];
+}
+
+interface PatientIdentifier {
+    patientId: string;
+    studyId: string;
 }
 
 export class StudyViewPageStore
@@ -2165,14 +2175,26 @@ export class StudyViewPageStore
         // We do not support studyIds in the query filters
         let filters: Partial<StudyViewFilter> = {};
         if (query.filterJson) {
-            try {
-                filters = JSON.parse(
-                    decodeURIComponent(query.filterJson)
-                ) as Partial<StudyViewFilter>;
-                this.updateStoreByFilters(filters);
-            } catch (e) {
-                //  TODO: add some logging here?
+            const parsedFilterJson = this.parseRawFilterJson(query.filterJson);
+            if (query.filterJson.includes('patientIdentifiers')) {
+                const sampleListIds = studyIds.map(s => s.concat('', '_all'));
+                const samples = await this.fetchSamplesWithSampleListIds(
+                    sampleListIds
+                );
+                const {
+                    patientIdentifiers,
+                } = parsedFilterJson as PatientIdentifierFilter;
+                const sampleIdentifiers = this.convertPatientIdentifiersToSampleIdentifiers(
+                    patientIdentifiers,
+                    samples
+                );
+                if (sampleIdentifiers.length > 0) {
+                    filters.sampleIdentifiers = sampleIdentifiers;
+                }
+            } else {
+                filters = parsedFilterJson as Partial<StudyViewFilter>;
             }
+            this.updateStoreByFilters(filters);
         } else if (query.filterAttributeId && query.filterValues) {
             const clinicalAttributes = _.uniqBy(
                 await defaultClient.fetchClinicalAttributesUsingPOST({
@@ -2222,6 +2244,41 @@ export class StudyViewPageStore
                 };
             }
         }
+    }
+
+    parseRawFilterJson(filterJson: string): any {
+        let rawJson;
+        try {
+            rawJson = JSON.parse(decodeURIComponent(filterJson));
+        } catch (e) {
+            console.error('FilterJson invalid Json: error: ', e);
+        }
+        return rawJson;
+    }
+
+    fetchSamplesWithSampleListIds(sampleListIds: string[]) {
+        return defaultClient.fetchSamplesUsingPOST({
+            sampleFilter: {
+                sampleListIds: sampleListIds,
+            } as SampleFilter,
+            projection: 'SUMMARY',
+        });
+    }
+
+    convertPatientIdentifiersToSampleIdentifiers(
+        patientIdentifiers: Array<PatientIdentifier>,
+        samples: Sample[]
+    ): SampleIdentifier[] {
+        return samples
+            .filter(s =>
+                patientIdentifiers.some(
+                    p => p.patientId === s.patientId && p.studyId === s.studyId
+                )
+            )
+            .map(s => ({
+                sampleId: s.sampleId,
+                studyId: s.studyId,
+            }));
     }
 
     @computed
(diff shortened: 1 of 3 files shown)


inodb (commented):
Thanks @haynescd ! This is great!

There is some odd behavior I noticed when you use the filter. It doesn't seem to load all the charts?

https://deploy-preview-4616--cbioportalfrontend.netlify.app/study/summary?id=msk_impact_2017#filterJson={%22patientIdentifiers%22:[{%22studyId%22:%22msk_impact_2017%22,%22patientId%22:%22P-0000004%22}]}

It does load all the charts when I access it directly: https://deploy-preview-4616--cbioportalfrontend.netlify.app/study/summary?id=msk_impact_2017

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2165,14 +2174,20 @@ export class StudyViewPageStore
...
-            try {
-                filters = JSON.parse(
-                    decodeURIComponent(query.filterJson)
-                ) as Partial<StudyViewFilter>;
-                this.updateStoreByFilters(filters);
-            } catch (e) {
-                //  TODO: add some logging here?
+            const parsedFilterJson = this.parseRawFilterJson(query.filterJson);
+            if (query.filterJson.includes('patientIdentifiers')) {
+                const samples = await this.fetchSamplesWithStudyIds(studyIds);
```
alisman, line 2179: hmmm. it's a little painful to pull down all the samples in this way.  there ought to be an endpoint to retrieve samples for a set of patients.  from what i can see right now, you can only do it for one patient at a time.  @inodb do you think that this is enough of an obscure usage that we shouldn't worry about this?  @haynescd at the very least, we should use the https://genie.cbioportal.org/api/samples/fetch service because that's a lot more likely to be in cache

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2224,6 +2239,49 @@ export class StudyViewPageStore
...
+        let rawJson;
+        try {
+            rawJson = JSON.parse(decodeURIComponent(filterJson));
+        } catch (e) {
+            console.error('FilterJson invalid Json: error: ', e);
+        }
+        return rawJson;
+    }
+
+    async fetchSamplesWithStudyIds(studyIds: string[]) {
```
alisman, line 2252: you actually don't have to use async/await here.  This function should just return ```internalClient.fetchFilteredSamplesUsingPost(...)```.    You only need to use async/await when you want to do something with the returned data before returning it

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2224,6 +2239,49 @@ export class StudyViewPageStore
...
+        let studyViewFilter: StudyViewFilter = {} as any;
+        studyViewFilter.studyIds = studyIds;
+        return await internalClient.fetchFilteredSamplesUsingPOST({
+            studyViewFilter: studyViewFilter,
+        });
+    }
+
+    convertPatientIdentifiersToSampleIdentifiers(
+        patientIdentifierFilter: PatientIdentifierFilter,
```
alisman, line 2261: this should accept patientIdentiviers, not the filter since i think all that's important is the identifiers, not the parent of them.  clearer interface.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2224,6 +2239,49 @@ export class StudyViewPageStore
...
+        samples: Sample[]
+    ): SampleIdentifier[] {
+        const sampleIdentifiers: SampleIdentifier[] = [];
+        patientIdentifierFilter.patientIdentifiers.forEach(
```
alisman, line 2265: i think this can be optimized and made clearer by reversing the loop.  
```
const sampleIdentifiers = samples.filter(sample=>i belong to a patient in patientIdetnifiers).map(...)
```
further optimization would be to make a dictionary of patients with compoundKey.

src/pages/studyView/StudyViewPageStore.ts
alisman: that's gonna result in a shitload of api requests though.  because samples endpoint is likely to be in cache, it's better to just fetch all.

src/pages/studyView/StudyViewPageStore.ts
haynescd: done

src/pages/studyView/StudyViewPageStore.ts
haynescd: done

src/pages/studyView/StudyViewPageStore.ts
haynescd: Agreed. That is why I initially used the api to get all the samples in a study

haynescd: So you are suggesting to use /studies/{studyId}/samples because it might be cached?

alisman: yes, i'm pretty sure that the filtering endpoints are not well "covered" by caching for reasons i'm happy to explain.  whereas the samples endpoint IS well covered.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -389,6 +390,15 @@ enum CustomDataTypeEnum {
     NUMERICAL = 'NUMERICAL',
 }
 
+interface PatientIdentifierFilter {
```
alisman, line 393: we should put all these types in the shared/models folder.  The reason is that people end up importing them in places and it ends up bloating the webpack bundles.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2224,6 +2246,41 @@ export class StudyViewPageStore
...
+        });
+    }
+
+    convertPatientIdentifiersToSampleIdentifiers(
+        patientIdentifiers: Array<PatientIdentifier>,
+        samples: Sample[]
+    ): SampleIdentifier[] {
+        return samples
+            .filter(s =>
+                patientIdentifiers.some(
```
alisman, line 2274: as ino pointed out today, there could be a crapload of patients potentially, so what i would do is first, before loop, make a dictionary with the compound key p.patientId+p.studyId.   then you can use lookup (string in object) which is a zillion times faster than array scanning.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -2165,14 +2175,26 @@ export class StudyViewPageStore
...
         let filters: Partial<StudyViewFilter> = {};
         if (query.filterJson) {
-            try {
-                filters = JSON.parse(
-                    decodeURIComponent(query.filterJson)
-                ) as Partial<StudyViewFilter>;
-                this.updateStoreByFilters(filters);
-            } catch (e) {
-                //  TODO: add some logging here?
+            const parsedFilterJson = this.parseRawFilterJson(query.filterJson);
```
alisman, line 2178: this is a comment about updateStoreFromURL and the "problem" precedes your work here, but I think we should try to address it.   the issue is that we are calling this async method without explicitly handling "asycnronicity".  I.e. we don't ever wait for it to finish.   It doesn't cause a problem because we await something which is dependent on this stuff.  But it's not super good and I think it would be pretty easy to get right.  Lets zoom for a few mins tomorrow and we can fix it.

PR #4618, opened 2023-05-10
:zap: Update GenomeNexus Api sync

haynescd:
Update Api for Genome Nexus


packages/react-mutation-mapper/src/util/DataFetcherUtils.ts
```diff
@@ -67,7 +67,7 @@ export function getUrl(urlTemplate: string, templateVariables: any) {
 
 export async function fetchVariantAnnotationsByMutation(
     mutations: Partial<Mutation>[],
-    fields: string[] = ['annotation_summary'],
+    fields: any = 'annotation_summary',
```
onursumer, line 70: I think we should keep this param and the default value as is. Hopefully, the API client should still be able to handle array of strings (otherwise it means that we already introduced a breaking change). Instead, let's update `fields` where we call the client function. Something like this:

```TypeScript
client.fetchVariantAnnotationByGenomicLocationPOST({
    genomicLocations,
    fields as any,
    isoformOverrideSource,
})
```

packages/react-mutation-mapper/src/util/DataFetcherUtils.ts
```diff
@@ -85,7 +85,7 @@ export async function fetchVariantAnnotationsByMutation(
 
 export async function fetchVariantAnnotationsIndexedByGenomicLocation(
     mutations: Partial<Mutation>[],
-    fields: string[] = ['annotation_summary'],
+    fields: any = 'annotation_summary',
```
onursumer, line 88: same as above. let's keep this param (and the default value) as is if possible and do `fields as any` where we call the fetch function.

onursumer (changes requested):

PR #4620, opened 2023-05-17
Fix infinite loop in QueryStore

BasLee:
When a user views a virtual study without any mutations, the result view page will fire an infinite number of backend requests to `/api/session/virtual_study`. 
This is caused by an infinite loop in the `QueryStore` that keeps recalculating `sampleLists` when selecting a virtual study.

(This PR also adds some documentation on the `localdev` property that is needed to debug the frontend using https.)

## Reproduce bug
The bug can be reproduced by:
- importing a study without mutations;
- using cbioportal with https
- selecting the virtual study on the index page by clicking the checkbox next to it (not the 'pie chart' link to the study);
- click 'explore selected studies' on the index page

## Fix
The (duplicated) `onResult` part of `sampleList` and `sampleListInSelectedStudies` trigger the infinite loop. By moving this logic to a reaction the bug is fixed.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -124,6 +124,26 @@ export class QueryStore {
 
         makeObservable(this);
 
+        /**
+         * Reset selectedSampleListId when sampleLists and sampleListInSelectedStudies have finished
+         */
+        reaction(
+            () => [this.sampleLists, this.sampleListInSelectedStudies],
+            () => {
+                if (
+                    !this.sampleLists.isComplete ||
+                    !this.sampleListInSelectedStudies.isComplete
+                ) {
+                    return;
+                }
+                if (
+                    !this.initiallySelected.sampleListId ||
+                    this.studiesHaveChangedSinceInitialization
+                ) {
+                    this._selectedSampleListId = undefined;
+                }
+            }
+        );
         this.initialize(urlWithInitialParams);
     }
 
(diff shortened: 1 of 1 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -471,11 +471,21 @@ export class QueryStore {
     @observable rppaScoreThreshold: string = '2.0';
 
     // sample list id
-    @observable private _selectedSampleListId?: string = undefined; // user selection
+    private _selectedSampleListId?: string = undefined; // user selection
```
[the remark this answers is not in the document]
BasLee: Fixed in [c5ee2b2](https://github.com/cBioPortal/cbioportal-frontend/commit/c5ee2b20980330ceb31f44a6a6191a778174744a)

src/shared/components/query/QueryStore.ts
```diff
@@ -1076,26 +1078,21 @@ export class QueryStore {
     }
 
     readonly sampleLists = remoteData({
+        await: () => [
+            this.sampleListsInStudyCache.get(
+                this.selectableSelectedStudyIds[0]
```
alisman, line 1083: we should include a comment explaining why it's ok to just select first one (i think it's b/c you would only be able to select a sample list if there was only a single study selected, right?)

alisman (approved):

BasLee:
In [39c9832](https://github.com/cBioPortal/cbioportal-frontend/commit/39c98328b832f8e14e1dbdd13c6996c22d755c7e) the logic to reset _selectedSampleListId is moved from the getter to a reaction. 
The resetting only needs to happen when the original remoteData calls are done.

src/shared/components/query/QueryStore.ts
BasLee: this logic was already there, and I don't really know why this is ok :man_shrugging:

PR #4627, opened 2023-05-23
Fix add chart button so that it never overlaps tabs

alisman:
Reset charts button is now treated as a tab and so will never overlap other tabs

image


onursumer (approved):
LGTM!

PR #4631, opened 2023-05-26
Update pathway mapper version to 2.3.0

onursumer:


onursumer:
Failing e2e tests seems unrelated. They also fail in master branch currently.

PR #4633, opened 2023-05-30
Remove redundant makeObservable call

aditygrg2:
Fixes "Error: [MobX] No annotations were passed to makeObservable, but no decorator members have been found either” error in dev mode


alisman (approved):

alisman:
@aditygrg2 thank you for this.  where was this error showing up though?  i don't see it in production.

PR #4635, opened 2023-06-02
Feature/ch/10184 support patient filters sent via postdata

haynescd:
Fix cbioportal/cbioportal#10184

- Refactored StudyViewQuery extraction ( extracted logic to separate class)
- Updated StudyViewPage to check if postData has valid filterJson to pass to StudyViewPageStore for filtering

### DIFF
--- a/end-to-end-test/shared/specUtils.js
+++ b/end-to-end-test/shared/specUtils.js
@@ -188,6 +188,25 @@ function setDropdownOpen(
         }
     );
 }
+function getCorrectUrlAndSetLocalStorage(url) {
+    if (!useExternalFrontend) {
+        console.log('Connecting to: ' + url);
+    } else if (useNetlifyDeployPreview) {
+        browser.execute(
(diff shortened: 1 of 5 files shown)


end-to-end-test/shared/specUtils.js
```diff
@@ -189,6 +189,22 @@ function setDropdownOpen(
...
 
+/**
+ * @param {string} url
+ * @returns {string} modifiedUrl
+ */
+function getUrl(url) {
+    if (!useExternalFrontend) {
+        console.log('Connecting to: ' + url);
+    } else {
+        const urlparam = useLocalDist ? 'localdist' : 'localdev';
```
alisman, line 200: where does useLocalDist come from?

haynescd: That is probably a legacy env var. I extracted this logic from goToUrlAndSetLocalStorage fn

PR #4638, opened 2023-06-02
Upgrade Google Analytics to v4

alisman:
Google Analytics requires upgrade to v4


src/shared/lib/tracking.ts
```diff
@@ -38,6 +38,7 @@ export function initializeTracking() {
 }
 
 export function trackEvent(event: GAEvent) {
+    return;
```
inodb, line 41: does this mean track events no longer work?

inodb:
looks like we no longer need these code changes

PR #4641, opened 2023-06-07
Fix infinite reload of sample lists

alisman:
The onResult handlers were mutating state on which the invoke dependend, causing infinite loop.  This gets rid of the onResult and moves the clearing of the sample list selection state to the study selection handler.  Whenever studies are selected/deselected, the sample list is reset to default.

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -8,6 +8,7 @@ import {
     observable,
     ObservableMap,
     reaction,
+    runInAction,
     toJS,
 } from 'mobx';
 import {
@@ -296,7 +297,10 @@ export class QueryStore {
     }
 
     set selectableSelectedStudyIds(val: string[]) {
-        this._allSelectedStudyIds = observable.map(stringListToSet(val));
+        runInAction(() => {
+            this.selectedSampleListId = undefined;
+            this._allSelectedStudyIds = observable.map(stringListToSet(val));
+        });
     }
 
     @action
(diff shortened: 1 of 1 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -1004,14 +1007,6 @@ export class QueryStore {
             );
         },
         default: [],
-        onResult: () => {
```
alisman, line 1007: @BasLee this is the offending onResult.  It depends upon this._selectedSampleListId, so when it mutates that value it re-fires invoke and is circular.

alisman: I believe that it was necessary in order to clear the selection when user selected a different study.  This can/should be handled on read by checking to see if the _selectedSampleListId matches selectedStudies

BasLee: the weird thing was, that the infinite loop even existed without any code in the `onResult`: it seemed like the presence of onResult itself was causing the loop

BasLee: the weird thing was, that the infinite loop even existed without any code in the onResult: it seemed like the presence of onResult itself was causing the loop

alisman: interesting.  i will test that.

src/shared/components/query/QueryStore.ts
```diff
@@ -2295,16 +2286,7 @@ export class QueryStore {
...
-            formSubmit(
-                buildCBioPortalPageUrl('data_download'),
-                urlParams.query,
-                undefined,
-                'smart'
-            );
-        } else {
-            this.singlePageAppSubmitRoutine(urlParams.query, currentTab);
-        }
+        this.singlePageAppSubmitRoutine(urlParams.query, currentTab);
```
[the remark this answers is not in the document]
alisman: yeah, download mode doesn't exist anymore so I've been trying to clean this stuff out, but i should probably do it in a seperate PR

src/shared/components/query/QueryStore.ts
alisman: i see that you are right.  in my testing, the problem is that mobxpromise's onResult references this.result in order to pass it to the callback:

image
 
when i use result instead of this.result, the problem goes away.   will have to make this change to the library.  

nevertheless, i think getting rid of onResult in general is a good idea as it is almost never actually necessary.

PR #4645, opened 2023-06-20
:sparkles: Add Time Range Selection to KM Plots on Study View

haynescd:
Fix cBioPortal/cbioportal#10152
Update KM Plots on Study View Page to add Time Selection filtering


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -594,6 +598,51 @@ export default class SurvivalChart
...
+        const scatterPoints: Array<ScatterData> = data[1].data;
+        if (scatterPoints.length > 2) {
+            const dataBin = this.generateFilterDataBin(scatterPoints);
+            if (this.props.onUserSelection) {
+                this.props.onUserSelection([dataBin]);
+            }
+        }
+    }
+
+    private generateFilterDataBin(scatterPoints: Array<ScatterData>): DataBin {
```
alisman, line 612: just wanted to point out that it would probably be better to make this a pure helper function rather than a private class method.   Were you to write a unit test for this, it would be much easier to do so.  It relies on that id prop, but that should be a function argument so that it's role as input is made clear.

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -594,6 +598,51 @@ export default class SurvivalChart
         this.sliderValue = Number.parseFloat(text);
     }
 
+    @autobind
+    private onSelection(data: any[]) {
+        const scatterPoints: Array<ScatterData> = data[1].data;
```
alisman, line 603: it it possible data could be empty?

src/pages/studyView/charts/survival/StudyViewSurvivalUtils.ts
```diff
@@ -48,7 +51,35 @@ export function makeSurvivalChartData(
...
         sortedGroupedSurvivals,
         pValue,
+        attributeId,
     };
 }
 
 export function makeScatterPlotData() {}
+
+export function isSurvivalAttributeId(attributeId: string) {
+    return attributeId.includes('_MONTHS') || attributeId.includes('_STATUS');
```
alisman, line 61:
```
/_MONTHS$/i.test(attributeId)
```

PR #4649, opened 2023-06-21
Improve tooltips for Altered and Unaltered group columns on comparison page #9718

h164654156465:
Fix cBioPortal/cbioportal#9718

- Edit the tooltip of the column headers to indicate this:
Altered group: Number (percentage) of samples in Altered group that are profiled for, and altered in, listed gene.
- As @jjgao suggested, I've added the tooltip to each data cell under Altered group and Unaltered group, saying "x samples in Altered group are profiled for [gene name], y (z%) of which are altered in [gene name]"

Screenshot 2023-06-21 at 3 24 09 PM


src/pages/resultsView/enrichments/EnrichmentsUtil.tsx
```diff
@@ -114,6 +114,13 @@ export function formatPercentage(
     );
 }
 
+export function getProfiledCount(
```
alisman, line 117: not a big deal but you probably don't need to export this since it's used internally only.  it's sort of equivalent to private.   otherwise it pollutes the global namespace.  but no need to fix.

PR #4654, opened 2023-06-30
Adapt study view to use clickhouse driven endpoints

alisman:


src/shared/api/cbioportalInternalClientInstance.ts
```diff
@@ -1,5 +1,28 @@
...
+        const url = old.apply(this, [params]);
+        return url;
+    };
+}
 
 const internalClient = new CBioPortalAPIInternal();
 
+export const internalClientColumnStore = new CBioPortalAPIInternal();
+
+proxyColumnStore(internalClientColumnStore, 'fetchMutatedGenes');
```
alisman, line 23: @haynescd right here is the ONLY thing you need to change to add an endpoint.  the string obviously corresponds to a fetch method on the client which corresponds to an endpoint

PR #4656, opened 2023-07-03
New route for handling "/results" 

aditygrg2:
Fix cBioPortal/cbioportal#10252

Fixed issue where navigating to /results route resulted in an error by creating a new route redirecting to the homepage.


alisman:
@aditygrg2 thanks for this.  but i'm curious how you got to the results page without params.   if that's something that happens, we should fix that.

alisman:
Unfortunately I don't think this will work because we need to support a default tab of results view.  For example, [this](https://www.cbioportal.org/results?cancer_study_list=coadread_tcga_pub&cancer_study_id=coadread_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&Z_SCORE_THRESHOLD=2.0&case_set_id=coadread_tcga_pub_nonhypermut&gene_list=KRAS+NRAS+BRAF&gene_set_choice=user-defined-list) has always defaulted to the Oncoprint tab.   So to fix the issue you found, we have to detect the absence of params.  I think the right thing to do then is just to fail and show a proper invalid query message.   I believe we already have a mechanism in routes.tsx for validating. Look into that.

aditygrg2:
@alisman Checking the routes, I observed that the current solution to this kind of problem is just showing a blank page. For example, navigating to [https://www.cbioportal.org/comparison](https://www.cbioportal.org/comparison) will open up a blank page. Should we do that? 
For all other nonmatching routes, we do have a 404 page, for specific invalid query pages we might need to create another page, that may be done as well.

alisman:
@aditygrg2 seems that unmatched routes are correctly handled by the 
```
<Route
                    path="*"
                    component={ScrollToTop(() => (
                        <PageNotFound />
                    ))}
                />
```

for /comparison seems we need to improve the validation so that it catches missing params and shows appropriate error message.   not too worried about this.

aditygrg2:
@alisman Yes, agree. So, the solution I am thinking of is to show a similar page like this - [https://www.cbioportal.org/df](https://www.cbioportal.org/df) with a modified title. Maybe we can refactor PageNotFound.tsx a little to avoid duplicacy.

aditygrg2:
@alisman Showing an "Invalid Query" type specific message might require more changes. Please let me know if we need to do this. I have just checked the absence of params, and if absent the user will be redirected to 404 which is our default. It was optional earlier.

aditygrg2:
maybe we can also make the `<PageNotFound/>` component dynamic with different types of error messages, default being the one which is present now, and special error messages for different pages.

PR #4661, opened 2023-07-06
Add VEP link in reVUE tooltip

leexgh:
- Update genome nexus api with the latest revue data model
- Update revue tooltip content
image


inodb (approved):
Great! Thanks so much!

PR #4662, opened 2023-07-07
Option to disable data download but still allow figure download

h164654156465:
Fix cBioPortal/cbioportal#10180

- disable data download in study view, group comparison view, result view, and patient view.
- support enum SHOW_ALL, SHOW_DATA, HIDE_ALL in frontend based on the config in backend skin.hide_download_controls
- pass showCopyDownload or showDownload as props to components to customize its behavior

Related backend PR https://github.com/cBioPortal/cbioportal/pull/10258

### DIFF
--- a/packages/cbioportal-frontend-commons/src/components/appContext/AppContext.tsx
+++ b/packages/cbioportal-frontend-commons/src/components/appContext/AppContext.tsx
@@ -1,11 +1,17 @@
 import React, { Context } from 'react';
 
+export enum DownloadControlOption {
+    SHOW_ALL = 'show',
+    HIDE_DATA = 'data',
+    HIDE_ALL = 'hide',
+}
+
 export interface IAppContext {
-    showDownloadControls: boolean;
+    showDownloadControls: DownloadControlOption;
 }
 
 export const AppContext: Context<IAppContext> = React.createContext<
     IAppContext
 >({
-    showDownloadControls: true,
+    showDownloadControls: DownloadControlOption.SHOW_ALL,
 });
--- a/src/appShell/App/Container.tsx
+++ b/src/appShell/App/Container.tsx
@@ -31,6 +31,7 @@ import {
 import makeRoutes from 'routes';
 import { AppContext } from 'cbioportal-frontend-commons';
 import { IAppContext } from 'cbioportal-frontend-commons';
+import { DownloadControlOption } from 'cbioportal-frontend-commons';
 import { ErrorAlert } from 'shared/components/errorScreen/ErrorAlert';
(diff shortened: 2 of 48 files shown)


src/appShell/App/Container.tsx
```diff
@@ -68,9 +69,27 @@ export default class Container extends React.Component<IContainerProps, {}> {
     }
 
     get appContext(): IAppContext {
+        const downloadConfig: string = getServerConfig()
```
alisman, line 72: @kalletlak, will this be backward compatible?  if they had it as false in portal.properties, it seems it would now default to true?

kkalletla-cls: Yeah, looks like not backward compatible. I'll discuss and let you know.

src/config/serverConfigDefaults.ts
```diff
@@ -202,7 +202,7 @@ export const ServerConfigDefaults: Partial<IServerConfig> = {
     skin_home_page_unauthorized_studies_global_message:
         'The study is unauthorized. You need to request access.',
     comparison_categorical_na_values: 'NA',
-    skin_hide_download_controls: false,
+    skin_hide_download_controls: 'false',
```
kalletlak, line 205: Change this to `show`

kalletlak (changes requested):

alisman:
The context api is a legitimate way to do this but it departs from the usual way of accessing global settings loaded from server, which is getServerConfig().   I think since that's already established, we might as well keep using it.   Let me know if you see a reason otherwise.

kalletlak:
> The context api is a legitimate way to do this but it departs from the usual way of accessing global settings loaded from server, which is getServerConfig(). I think since that's already established, we might as well keep using it. Let me know if you see a reason otherwise.

You mean to replace context with getServerConfig() for `showDownloadControls`?

kalletlak:
@h164654156465  could you replace all occurrence with what's done at https://github.com/cBioPortal/cbioportal-frontend/pull/4662/files#diff-d04b96b830a44847891cea2270dce1679954330a355fb46488bebc9ae466179aR449

packages/cbioportal-frontend-commons/src/components/appContext/AppContext.tsx
```diff
@@ -1,11 +1,17 @@
 import React, { Context } from 'react';
 
+export enum DownloadControlOption {
+    SHOW_ALL = 'show',
+    HIDE_DATA = 'data',
```
kalletlak, line 5: Looks like this enum is never used

PR #4663, opened 2023-07-10
Persist text color on icon in group comparison buttons

zhx828:
It would be nice to use the same text color on the icon and also the left border of the icon. We do that in the study view. Let me know whether you like this change. 

Ideally we should use the PillTag for the button, but I understand the complicity of this component. 

We should also add a bit more padding for the button, so the left border of the delete icon shows up better. But that seems to have bigger impact, so I didn't include.

### Current
Screen Shot 2023-07-10 at 10 33 09 AM

### Proposed
Screen Shot 2023-07-10 at 10 43 21 AM


alisman (approved):

alisman:
@zhx828 looks good.  needs a few screen shot updates.  let me know if you need help with that.

zhx828:
> @zhx828 looks good. needs a few screen shot updates. let me know if you need help with that.

@alisman I only see one test is relevant to this PR. Let me know if others need to be fixed too.

PR #4673, opened 2023-07-25
Add custom url handling for resource tab iframes (enables state persistence for Minerva viewer)

alisman:
The Minerva viewer is used to display slides in the patient view.  The viewer has settings (stored in URL params) which allow it to maintain state.  We want to persist this state as a user navigates across patients.  This requires us to listen to postMessage broadcasts from inside iframe and store that state.

### DIFF
--- a/src/shared/components/resources/ResourceTab.tsx
+++ b/src/shared/components/resources/ResourceTab.tsx
@@ -5,12 +5,12 @@ import WindowStore from '../window/WindowStore';
 import { action, computed, makeObservable } from 'mobx';
 import styles from './styles.module.scss';
 import classNames from 'classnames';
-import autobind from 'autobind-decorator';
 import { getBrowserWindow } from 'cbioportal-frontend-commons';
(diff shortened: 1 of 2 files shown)


inodb (approved):
Thanks so much! Looks great!

src/shared/components/resources/ResourceTab.tsx
```diff
@@ -93,6 +93,14 @@ export default class ResourceTab extends React.Component<
                 url = buildPDFUrl(this.currentResourceDatum.url);
                 break;
         }
+
+        // apply instance specific transformation on url (e.g. to keep state)
+        CUSTOM_URL_TRANSFORMERS.forEach(config => {
+            if (config.re.test(getServerConfig().app_name || '')) {
+                url = config.func(url);
+            }
+        });
```
inodb, line 102: maybe it can be applied for all apps, but we only apply it for those resources where any one of these fields contains the string 'minerva':

```
description: "Minerva story"
displayName: "Minerva Story"
resourceId: "MINERVA_STORY"
```

PR #4685, opened 2023-08-04
Replace twitter with news feed

alisman:
Twitter feed no longer works.    Replace it with the contents of the news page. 

image

### DIFF
--- a/src/shared/components/rightbar/RightBar.tsx
+++ b/src/shared/components/rightbar/RightBar.tsx
@@ -9,13 +9,15 @@ import { Link } from 'react-router-dom';
 import LoadingIndicator from '../loadingIndicator/LoadingIndicator';
 import { buildCBioPortalPageUrl } from '../../api/urls';
 import { ResultsViewTab } from '../../../pages/resultsView/ResultsViewPageHelpers';
+import { mockOrder } from 'pages/patientView/vafPlot/mockData';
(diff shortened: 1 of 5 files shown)


src/shared/components/rightbar/RightBar.tsx
```diff
@@ -44,6 +42,36 @@ export default class RightBar extends React.Component<
         );
     }
 
+    public newsContent = remoteData(async () => {
+        await sleep(5000);
+
+        return fetch(
+            'https://www.cbioportal.org/proxy/docs.cbioportal.org/news/'
```
alisman, line 49: @inodb this will show on all portals except where there is What's New content specified in configuration.  Basically it will function exactly as Twitter feed did.  Is this what we want?

inodb: good question - i guess maybe people may want to point to some custom markdown?

For e.g. GENIE, we could point to genie news: https://docs.cbioportal.org/news-genie/

inodb: The other thing is that if somebody is running an old portal, not sure what a good default would be. I guess if the news markdown is part of the project we could link to that directly rather than the cbioportal.org news page?

alisman: we agreed to just deploy this.  it will function exactly as twitter embed did.

PR #4688, opened 2023-08-08
Adding jitpack build script and label check test for pull request

dippindots:
Two git actions added:
- Trigger jitpack build when a new release is made
- Check pr has the correct labels based on `release-drafter.yml`


inodb (approved):
Looks great! Thanks so much!

PR #4692, opened 2023-08-10
bug fix: case no meta data from mutational signatures endpoint

TJMKuijpers:
Added an additional check when `fetchAllMutationalSignatureCountMetaData` does return an empty array.
In this case, the stable ID (which should always be defined) is used the create the mutational signature label.


TJMKuijpers:
@inodb I will test it with the pancan study locally before making this ready for review (just to be sure)

src/pages/patientView/mutationalSignatures/MutationalSignatureBarChart.tsx
```diff
@@ -53,25 +51,25 @@ theme.legend.style.data = {
...
+/*export function transformMutationalSignatureData(dataset: IMutationalCounts[]) {
     const transformedDataSet = dataset.map((obj: IMutationalCounts) => {
         const transformedDataSet = dataset.map((obj: IMutationalCounts) => {
             let referenceTransformed = -Math.abs(obj.value);
             return { ...obj, referenceTransformed };
         });
         return transformedDataSet;
     });
-}
+}*/
```
inodb, line 72: what were these used for previously?

TJMKuijpers: `formatLegendTopAxisPoints` is not needed anymore since this behavior is captured in the `@computed formatLegendTopAxisPoints`. The second function `transformMutationalSignatureData` is to show the reference counts but we will not show them in this chart. (I have a separate branch with the reference signature so the function is not lost)

PR #4693, opened 2023-08-14
Disable genomic alteration filtering when no relevant molecular profiles

BasLee:
(Replaced by https://github.com/cBioPortal/cbioportal-frontend/pull/4712)

Fix confusing filtering of the oncoprint when no genomic alteration profiles are selected.

For a more detailed explanation, see issue https://github.com/cBioPortal/cbioportal/issues/10316 .

## Changes

When all profiles with genomic alteration data are excluded:
- all genomic alteration checkboxes in the annotation menu are unchecked and deselected
- the custom tier list is hidden, because it is generated on the fly using (the excluded) genomic alteration data

This PR adds e2e tests to check the menu and oncoprint is displayed with mrna data when excluding all genomic alteration profiles.


BasLee:
After a closer look with @inodb, the filter seems broken. 
The filter clearly mentions to only filter genomic alteration data: 
```
[] Exclude alterations (mutations, structural variants and copy number) of unknown significance
```
A better solution would be, to fix this filter. So, all non-genomic-alteration-data should be unaffected by this filter: even when this filter is toggled on, mRNA and proteomic data should stay visible.

alisman:
@BasLee does this mean we should close this PR?

BasLee:
> does this mean we should close this PR?
@alisman yes, this one can be closed

PR #4695, opened 2023-08-15
Enhancements to patient page

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/10323
[RFC 75](https://docs.google.com/document/d/1K7_kDN-tm81GKdgfDEs3AscJhsFnUYc6U8reYH1YLu8/edit)

1. remove patient from cohort
2. navigate back to study page


src/pages/patientView/PatientViewPage.tsx
```diff
@@ -544,6 +552,72 @@ export class PatientViewPageInner extends React.Component<
...
+        );
+        if (patientIdentifiers.length > 0) {
+            (studyPage as any).studyPageFilter = `filterJson=${JSON.stringify({
+                patientIdentifiers,
+            })}`;
+        }
+    }
+
+    @action.bound
+    private handleDeletePatient(deleteAtIndex: number) {
```
alisman, line 582: in my testing, this doesn't work on a big cohort (e.g. >200) when the items are removed from the hash.  we end up stuck on a single patient.  not totally sure why that would be.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -544,6 +552,72 @@ export class PatientViewPageInner extends React.Component<
...
+        this.pageStore.patientIdsInCohort.splice(deleteAtIndex, 1);
```
alisman, line 583: it's odd that you essentially mutate the state on this first line, then later use window.open to arrive at the new url.  better to use a functional method (lodash) to get a new array (without mutating old one in place).   but it also begs question, why do we need to use window.open at all down below?

kalletlak: Yes, without doing `window.open` url won't be getting updated (with `URLWrapper` we could only update params not hash)

alisman: You could use urlWrapper to update hash.   urlWrapper uses the underlying ExtendedRouterStore (src/shared/lib/ExtendedRouterStore.ts) line 97, you can see we are just passing through the existing hash.  but this could easily be adapted to accept a new hash.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -544,6 +552,72 @@ export class PatientViewPageInner extends React.Component<
         return this.patientViewPageStore;
     }
 
+    @action.bound
+    private handleNavigationControlClick() {
```
alisman, line 556: change the method name to be more specific.  ```handleReturnToStudyView```

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -544,6 +558,68 @@ export class PatientViewPageInner extends React.Component<
...
+                studyId: patientIdParts[0],
+            };
+        });
+        const url = getPatientViewUrl(
+            navCaseIds[currentIndex].studyId,
+            navCaseIds[currentIndex].patientId,
+            navCaseIds
+        );
+
+        // update hash if url lenght is less than max length
```
alisman, line 609: hey, can you add to this:  
Because of url length limits, we can only maintain the list in the url hash for small sets of ids.

src/pages/patientView/PatientViewPage.tsx
```diff
@@ -544,6 +558,68 @@ export class PatientViewPageInner extends React.Component<
...
+            return {
+                patientId: patientIdParts[1],
+                studyId: patientIdParts[0],
+            };
+        });
+        const queriedStudies: Set<string> = new Set(
+            patientIdentifiers.map(p => p.studyId)
+        );
+
+        const studyPage = window.open(
```
alisman, line 574: lets put a comment explaining this little hacky mechanism for "posting" data to the new window.  
"... we need to do this because url method would be too long"

PR #4699, opened 2023-08-16
Add user color configuration for oncoprint

gblaih:
Implements https://github.com/cBioPortal/cbioportal/issues/10291


src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -586,6 +598,71 @@ export class ResultsViewPageStore extends AnalysisStore
...
+        color: RGBAColor
+    ) {
+        this.userAlterationColors[alteration][type] = color;
+    }
+
+    @action.bound
+    public setDefaultUserAlterationColors(
+        rule: IGeneticAlterationRuleSetParams
+    ) {
+        for (let alteration in rule.rule_params.conditional) {
```
alisman, line 614: avoid using for loops.  use functional approach instead.   _.forIn()

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -573,6 +574,17 @@ export class ResultsViewPageStore extends AnalysisStore
 
     @observable queryFormVisible: boolean = false;
 
+    @observable userAlterationColors: {
+        [alteration: string]: {
```
alisman, line 578: alterationType

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -586,6 +598,71 @@ export class ResultsViewPageStore extends AnalysisStore
         | ModifyQueryParams
         | undefined = undefined;
 
+    @action.bound
+    public onAlterationColorChange(
+        alteration: string,
```
alisman, line 603: alerationType

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -573,6 +574,17 @@ export class ResultsViewPageStore extends AnalysisStore
 
     @observable queryFormVisible: boolean = false;
 
+    @observable userAlterationColors: {
```
alisman, line 577: userSelectedAlterationColors

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -586,6 +602,69 @@ export class ResultsViewPageStore extends AnalysisStore
...
+    @computed get defaultClinicalTracksColors() {
+        return this._defaultClinicalTracksColors;
+    }
+
+    @computed get userSelectedClinicalTracksColors() {
+        return this._userSelectedClinicalTracksColors;
+    }
+
+    @action.bound
+    public setDefaultClinicalTrackColors(
```
alisman, line 636: we shouldn't ever need to set the defaults.  when the consumer of the track colors (the oncoprint rendring logic) requests a color, we look to see if there is a custom color for that track, and if not, we use a default.   since that default is not static, we may need to "cache" the colors so that they are only generated once during a load of the page.

src/shared/components/oncoprint/ClinicalTrackColorPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+    @computed get colorList() {
+        let colors: string[] = COLORS.slice(0, 20);
+        colors.push(CLI_YES_COLOR);
+        colors.push(CLI_NO_COLOR);
+        colors.push(CLI_FEMALE_COLOR);
+        colors.push(DARK_GREY);
+        return colors;
+    }
+
+    buildColorChooserWidget = () => (
```
alisman, line 74:
```
@computed get colorChooserElement() {

}
```

src/shared/components/oncoprint/ClinicalTrackColorPicker.tsx
```diff
@@ -0,0 +1,129 @@
...
+                    rootClose={true}
+                >
+                    <DefaultTooltip
+                        overlay={
+                            'Optional: Select color for clinical track value to be used in oncoprint. If no color is selected, the default color will be applied.'
+                        }
+                    >
+                        <span
+                            style={{
+                                float: 'right',
```
alisman, line 114: we avoid using float css property.  user flex box instead.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -972,13 +991,54 @@ export function makeClinicalTracksMobxPromise(
                         dataAndColors.categoryToColor,
                         hexToRGBA
                     );
+                    // if default for attribute doesn't already exist
+                    if (
+                        !oncoprint.props.store.defaultClinicalTracksColors[
+                            attribute.displayName
+                        ]
+                    ) {
+                        oncoprint.props.store.setDefaultClinicalTrackColors(
```
alisman, line 1000: we should never set state from inside of an invoke (this is known as a side effect)

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -131,6 +133,71 @@ export type AdditionalTrackGroupRecord = {
...
+    track: ClinicalTrackSpec,
+    value: string,
+    color: RGBAColor | undefined,
+    store: ResultsViewPageStore
+) {
+    if (track.datatype === 'string') {
+        // set new color of value in track's category_to_color
+        // if undefined color, set value's color in track's category_to_color back to its default
+        if (color) {
+            (track as any).category_to_color[value] = color;
```
alisman, line 163: look into whether we can avoid casting as any.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -131,6 +133,71 @@ export type AdditionalTrackGroupRecord = {
...
+        if (color) {
+            (track as any).countsCategoryFills[valueIndex] = color;
+        } else {
+            (track as any).countsCategoryFills[valueIndex] =
+                store.defaultClinicalTracksColors[track.label][value];
+        }
+    }
+}
+
+export function getClinicalTrackColor(
```
alisman, line 182: this should be a method of the store

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -183,6 +250,12 @@ export default class ResultsViewOncoprint extends React.Component<
         );
     }
 
+    @computed get enableWhiteBackgroundForGlyphs() {
```
alisman, line 253: isEnabled...

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -253,6 +326,8 @@ export default class ResultsViewOncoprint extends React.Component<
 
     @observable renderingComplete = false;
 
+    @observable changedTrackKey: string = '';
```
alisman, line 329: trackKeySelectedForEdit

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -1793,6 +1882,50 @@ export default class ResultsViewOncoprint extends React.Component<
         return WindowStore.size.width - 75;
     }
 
+    @action.bound
+    public handleClinicalTrackColorChange(
+        value: string,
+        color: RGBAColor | undefined
+    ) {
+        setClinicalTrackColor(
```
alisman, line 1890: we should not need to update two different things with essentially the same data.  if we update
store, then the track _should_  respond to the observable mutation.  because of the particular nature of the oncopring library, this may be difficult.

src/shared/components/oncoprint/ResultsViewOncoprint.tsx
```diff
@@ -1793,6 +1882,50 @@ export default class ResultsViewOncoprint extends React.Component<
...
+
+    @observable selectedClinicalTrackKey: string | null = null;
+
+    @action.bound
+    setSelectedClinicalTrackKey(key: string | null) {
+        this.selectedClinicalTrackKey = key;
+    }
+
+    @computed get selectedClinicalTrack() {
+        let index = _.findIndex(
```
alisman, line 1911: you should just use _.find and deal with typescrupt in another fashion.

alisman: whoever is using this needs to be able to to handle an undefined case.

src/pages/resultsView/ResultsViewPageStore.ts
```diff
@@ -573,6 +578,16 @@ export class ResultsViewPageStore extends AnalysisStore
...
     @observable queryFormVisible: boolean = false;
 
+    @observable _userSelectedStudiesToClinicalTracksColors: {
+        [studies: string]: {
+            [label: string]: {
+                [value: string]: RGBAColor;
+            };
+        };
+    } = JSON.parse(
+        localStorage.getItem('clinicalTracksColorConfig') || '{"global":{}}'
```
alisman, line 588: put localStorage get into constructor

PR #4701, opened 2023-08-17
Improve non-profiled gene panel links in oncoprint tooltip

BasLee:
Currently the gene panels in an oncoprint tooltip are marked red (instead of blue) when the gene is not on the panel. 

However, just marking the gene panel with the color red is confusing for some users: what does this color mean? And why is my gene not showing up in the panel, when I click on it?

To improve the meaning of non-profiled gene panels, this PR removes the red color and adds a message that the gene is not on the gene panel.

image
_New styling of gene panel without the current gene, including message_

image
_New styling when multiple patients are selected_

### DIFF
--- a/src/shared/components/oncoprint/TooltipUtils.spec.ts
+++ b/src/shared/components/oncoprint/TooltipUtils.spec.ts
@@ -1296,37 +1296,16 @@ describe('Oncoprint TooltipUtils', () => {
                 assert.equal(tooltipOutput.text().match(/panel2/g)!.length, 1);
                 assert.equal(tooltipOutput.text().match(/panel3/g)!.length, 1);
                 assert.equal(
-                    tooltipOutput.text().match(/panel1 \(1\)/g)!.length,
+                    tooltipOutput.text().match(/panel1\s*\(1\)/g)!.length,
                     1
                 );
                 assert.equal(
-                    tooltipOutput.text().match(/panel2 \(1\)/g)!.length,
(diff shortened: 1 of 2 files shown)


alisman:
@BasLee this looks good. would you mind just putting a before and after screenshot in this PR so the product folks can quickly evaluate it?

alisman:
@inodb, @tmazor  can you approve this from product perspective.  see the screenshot above

BasLee:
Instead of a message between brackets we could also formulate it like: 
`Gene Panels: Profiled in TESTPANEL1. Not profiled in TESTPANEL2.`
Any preference?

inodb:
@BasLee Thanks! Tali's proposal looks great! I wonder what happens in the tooltip for this case:

- a zoomed out oncoprint with e.g. two patients but they have different panels. What would the text say?

BasLee:
> I wonder what happens in the tooltip for this case:
> 
> * a zoomed out oncoprint with e.g. two patients but they have different panels. What would the text say?
@inodb 

When rendering the oncoprint tooltip, the profiled and non-profiled gene panels are grouped separately.
So for example, when you have two samples with different gene panels, first the tooltip will list all the gene panels of those two patients that are profiled (plus their count), and after that all gene panels that are unprofiled (with their count). When the patients have different panels, the counts will differ in the tooltip.

image
_Example of two patients with one gene panel only present in one of the two samples, and a gene panel that is shared by the samples_

inodb:
@BasLee thanks for investigating! That looks fine to me!

I guess we can merge this then?

PR #4706, opened 2023-08-21
Discrete molecular data counts

h164654156465:
Fix https://github.com/cBioPortal/cbioportal/issues/10257 and part of https://github.com/cBioPortal/cbioportal/issues/6759

- Adding CNA discrete profile to gene specific charts dropdown
- Main piece - plot visibility and filtering
- Necessary changes to show user selection
- Download functionality
- Showing it in manage Charts
- Check if comparison is working (excluding group NA for the time being)
- Changes to include it in Virtual study description and Group name
- Saving and fetching form user preferences


packages/cbioportal-frontend-commons/src/lib/AlterationColors.ts
```diff
@@ -22,5 +22,6 @@ export const PROT_COLOR_HIGH = '#ff3df8';
 export const PROT_COLOR_LOW = '#00E1FF';
 export const CNA_COLOR_AMP = '#ff0000';
 export const CNA_COLOR_GAIN = '#ffb6c1';
+export const CNA_COLOR_DIPLOID = 'CBC3E3';
```
alisman, line 25: should probably have the # similar to the rest of these hex value constants?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -972,10 +980,24 @@ export function getVirtualStudyDescription(
...
+                const dataFilterValues = genomicDataFilter.values;
+                const dataType = getDataTypeByDataFilterValues(
+                    dataFilterValues
+                );
+
                 if (name) {
                     filterLines.push(
                         `- ${name}: ${intervalFiltersDisplayValue(
-                            genomicDataFilter.values,
+                            dataType === DataType.NUMBER
```
alisman, line 991: i believe that the interval filters can be edited by clicking on the filter pills. can you check see how this is affected?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -972,10 +980,24 @@ export function getVirtualStudyDescription(
...
+                                ? dataFilterValues
+                                : _.map(dataFilterValues, d =>
+                                      _.assign(d, {
+                                          value:
+                                              getCNAByAlteration(
```
alisman, line 996: seems odd that in a stretch of code dealing with "genomicDataFilters" we are calling a function specific to CNA?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -972,10 +980,24 @@ export function getVirtualStudyDescription(
...
+                );
+
                 if (name) {
                     filterLines.push(
                         `- ${name}: ${intervalFiltersDisplayValue(
-                            genomicDataFilter.values,
+                            dataType === DataType.NUMBER
+                                ? dataFilterValues
+                                : _.map(dataFilterValues, d =>
+                                      _.assign(d, {
```
alisman, line 994: assign mutates in place, updating the target (d).  is that really what we want to do?  should we be producing something new?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2494,6 +2522,10 @@ export async function getHugoSymbolByEntrezGeneId(
     return gene.hugoGeneSymbol;
 }
 
+export function getHugoSymbolByChartTitle(title: string): string {
```
alisman, line 2525: is there a better way than parsing the title?

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3959,3 +3992,21 @@ export function showQueryUpdatedToast(message: string) {
         theme: 'light',
     } as any);
 }
+
+export function getDataTypeByDataFilterValues(
```
alisman, line 3996: @kalletlak  if i read this correctly, and it's difficult to read, we resolving to type string if any value in dataFilterValues is either undefined or empty string?   First of all, that seems very strange.  What if the values are all numbers?  Then it will resolve to type string?   There should be at least a comment explaining why it behaves this way and some indication in the name of the function of this quirkiness.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -3959,3 +3992,21 @@ export function showQueryUpdatedToast(message: string) {
...
+    dataFilterValues: DataFilterValue[]
+): DataType {
+    let valueCount = dataFilterValues.length;
+
+    _.map(dataFilterValues, (dataFilterValue: DataFilterValue) => {
```
alisman, line 4001: this isn't a mapping operation so it shouldn't use map.  only use map with the iterator returns a value.   this is really a each loop.  And in fact, I believe it's really a case for ```.some()``` because i think the function says, if at least one member of the array is undefined or empty string, then the type is Number.  In that case, we should exit the loop as soon as we detect one of these.  That's what some will do.

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -509,13 +519,27 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
                         ref={this.handlers.ref}
                         onUserSelection={this.handlers.onValueSelection}
-                        openComparisonPage={this.openComparisonPage}
+                        openComparisonPage={
+                            this.props.store.isGeneSpecificChart(
+                                this.props.chartMeta.uniqueKey
+                            )
+                                ? () =>
+                                      this.openComparisonPage({
+                                          hugoGeneSymbols: [
```
alisman, line 528: lets discuss why this is necessary.

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -204,6 +213,20 @@ export class StudySummaryTab extends React.Component<
                     props.promise = this.store.getGenericAssayChartDataCount(
                         chartMeta
                     );
+                } else if (
```
alisman, line 216: weird that gene specific chart should be exclusive with, say "generic assay chart".    Perhaps it's just semantics but we should improve the name then.

src/pages/studyView/StudyViewUtils.tsx
h164654156465: this will be removed as well. I will get the hugogenesymbol using other data

src/pages/studyView/StudyViewUtils.tsx
h164654156465: Karthik and I had a discussion and agreed with you to remove this. We'll giving it datatype instead of determining it in DataFilterValue.

src/pages/studyView/tabs/SummaryTab.tsx
alisman: how much effort would it take to refactor/break up this switch statement (which then has the internal if else statement.  it's just way too long and complicated.

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5816,12 +5933,12 @@ export class StudyViewPageStore
                 )
                 .uniq()
                 .value();
-            if (newChart.datatype === 'STRING') {
+            if (newChart.datatype === DataType.STRING) {
```
alisman, line 5936: 👍

alisman:
@kalletlak Found one little issue with mapping CNA to label values.  Probably painful : (

image

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -325,6 +348,20 @@ export class StudySummaryTab extends React.Component<
...
```
alisman, line 351: when a chart type has subtypes (requiring these if/else statements), we should detect subtype here and then pass to helper function, which will use another switch statement or lookup table 
```
{
    [chartSubType]: (...)=>props
}

```

PR #4708, opened 2023-08-23
Swapped x-axis in Scatter Chart 

officialasishkumar:
Fix [#10183](https://github.com/cBioPortal/cbioportal/issues/10183)

### DIFF
--- a/src/pages/resultsView/enrichments/MiniScatterChart.tsx
+++ b/src/pages/resultsView/enrichments/MiniScatterChart.tsx
@@ -172,7 +172,16 @@ export default class MiniScatterChart<
         };
     }
 
+    getReversedData() {
+        return this.props.data.map(dataPoint => ({
+            ...dataPoint,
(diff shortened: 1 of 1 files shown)


officialasishkumar:
Can someone please review it?

src/pages/resultsView/enrichments/MiniScatterChart.tsx
```diff
@@ -172,7 +172,16 @@ export default class MiniScatterChart<
         };
     }
 
+    getReversedData() {
```
alisman, line 175: you want to use a computed getter here to avoid redundant recomputation on every render cycle.  see ```@computed``` attribute elsewhere in codebase for example.

alisman: It seems odd that we always want to consume "reversed" data and that would be a fix for the issue.  shouldn't this be conditional based on group order?

PR #4710, opened 2023-08-24
Update mysql 8 connection string for localdb tests

dippindots:
The PR will update the local e2e test with the latest DB connection string and delete outdated ones.

### DIFF
--- a/end-to-end-test/local/runtime-config/portal.properties
+++ b/end-to-end-test/local/runtime-config/portal.properties
@@ -5,7 +5,7 @@ app.name=cbioportal
 db.user=cbio_user
 db.password=somepassword
 db.driver=com.mysql.jdbc.Driver
-db.connection_string=jdbc:mysql://cbioportal-database:3306/cbioportal?useSSL=false\&allowPublicKeyRetrieval=true
+db.connection_string=jdbc:mysql://cbioportal-database:3306/cbioportal?useSSL=false&allowPublicKeyRetrieval=true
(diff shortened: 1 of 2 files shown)


end-to-end-test/local/runtime-config/portal.properties
```diff
@@ -4,10 +4,8 @@ app.name=cbioportal
 # database
 db.user=cbio_user
 db.password=somepassword
-db.host=cbioportal-database-container
-db.portal_db_name=cbioportal
 db.driver=com.mysql.jdbc.Driver
-db.connection_string=jdbc:mysql://cbioportal-database-container/
+db.connection_string=jdbc:mysql://cbioportal-database:3306/cbioportal?useSSL=false&allowPublicKeyRetrieval=true
```
inodb, line 8: Why does this have a different name from:

https://github.com/cBioPortal/cbioportal-docker-compose/blob/master/docker-compose.yml#L24

dippindots: @inodb In my understanding, using a custom container name and service name is the same, in this case, we are using the service name, I just wanted to keep in sync with other places we have:
https://github.com/cBioPortal/cbioportal-docker-compose/pull/22/files
https://github.com/cBioPortal/cbioportal/pull/10037/files#diff-65ae1168eecd2c82b28b52c8a2c70866716e326bfa781144b33f0d74909103d9R49

PR #4712, opened 2023-08-28
Fix hidden non-genomic data when excluding VUS in oncoprint

BasLee:
Fix 'exclude VUS' filter in oncoprint. 

Replaces this [PR](https://github.com/cBioPortal/cbioportal-frontend/pull/4693), related to this [issue](https://github.com/cBioPortal/cbioportal/issues/10316).

When selecting 'exclude alterations of unknown significance', non-genomic mRNA and proteomic data are currently filtered out. We have driver information for Mutations, CNA and SVs, but not for mRNA and proteomic data, so we figured it would be better not to filter those out. If one wants to exclude those, one could deselect the mRNA/proteomic profile instead. This PR introduces that change.

image
_Exclude VUS filter in annotation menu_

image
_Before selecting the exclude VUS filter_

image
_With this fix, and after selecting the exclude VUS filter, mRNA data is now still shown_

### DIFF
--- a/src/pages/resultsView/ResultsViewPageStore.ts
+++ b/src/pages/resultsView/ResultsViewPageStore.ts
@@ -1792,9 +1792,9 @@ export class ResultsViewPageStore extends AnalysisStore
     readonly nonOqlFilteredAlterations = remoteData<ExtendedAlteration[]>({
         await: () => [
             this.filteredAndAnnotatedMutations,
-            this.filteredAndAnnotatedNonGenomicData,
             this.filteredAndAnnotatedCnaData,
             this.filteredAndAnnotatedStructuralVariants,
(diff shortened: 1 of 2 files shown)


inodb (approved):
This looks good to me, but let's maybe get a second opinion from either @tmazor , @schultzn or @jjgao before merging

inodb:
@BasLee 

>Currently the PR simply merged VUS with non-VUS molecular data. However, there seems so have been put quite some effort in splitting VUS from non-VUS specifically for molecular data. Is the change in the current PR to rigorous, and does it break other things? Or can we remove all of this logic?

Looking at the code I think it makes sense to make the genomic vs non-genomic distinction rather than treating all molecular data the same. One could keep the molecularData variable but prolly just as a combination of NonGenomic+Genomic data

PR #4721, opened 2023-09-11
[RFC77] Filter buttons on the landing page study search menu

TJMKuijpers:
Currently, users can only filter their data sets by reference genome but not by available data types. Users have to be either familiar with the data types in a study or click on a study to check whether the data types of interest are available.

This PR introduces an extension to the study search menu by introducing filters related to the available data types in each study. The data type filter will be an AND filter, which is different from the current reference genome filter (OR filter).

Therefore, the filter icon from the patient view page has been reused (with a label that mentions the AND filter).
Screenshot_DataTypeFile

The following tests have been added for the AND filter behavior to `Phrase.spec.tsx`:
- Should return a match when field has measured samples
- Should return no match when field has no measured samples and match in phraselist
- Should return match when two field have measured samples
- Should return no match when field has measured samples and one field having no samples

### DIFF
--- a/src/shared/components/query/filteredSearch/Phrase.tsx
+++ b/src/shared/components/query/filteredSearch/Phrase.tsx
@@ -1,4 +1,3 @@
-import { CancerTreeNode } from 'shared/components/query/CancerStudyTreeData';
 import {
     FullTextSearchFields,
     FullTextSearchNode,
@@ -8,7 +7,6 @@ import {
     FILTER_SEPARATOR,
     FILTER_VALUE_SEPARATOR,
 } from 'shared/components/query/filteredSearch/SearchClause';
-import { CancerStudy } from 'cbioportal-ts-api-client';
 
 /**
  * Phrase and associated fields
--- a/src/shared/components/query/CancerStudySelector.tsx
+++ b/src/shared/components/query/CancerStudySelector.tsx
@@ -27,7 +27,11 @@ import QuickSelectButtons from './QuickSelectButtons';
 import { StudySelectorStats } from 'shared/components/query/StudySelectorStats';
 import WindowStore from 'shared/components/window/WindowStore';
 import { StudySearch } from 'shared/components/query/StudySearch';
-
+import { DataTypeFilter } from 'shared/components/query/DataTypeFilter';
+import {
+    getSampleCountsPerFilter,
+    getStudyCountPerFilter,
+} from 'shared/components/query/filteredSearch/StudySearchControls';
 const MIN_LIST_HEIGHT = 200;
 
 export interface ICancerStudySelectorProps {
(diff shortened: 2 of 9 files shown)


src/shared/components/query/filteredSearch/field/dataTypeFilterField.tsx
```diff
@@ -0,0 +1,86 @@
...
+            checkedOptions.push(option);
+        }
+    }
+    return (
+        <div>
+            <h5>Data type filter</h5>
+            <div>
+                {options.map((option: string) => {
+                    const id = `input-${option}`;
+                    let isChecked = checkedOptions.includes(option);
```
onursumer, line 51:
```suggestion
                    const isChecked = checkedOptions.includes(option);
```

src/shared/components/query/filteredSearch/field/dataTypeFilterField.tsx
```diff
@@ -0,0 +1,86 @@
...
+                    return (
+                        <TableHeaderCellFilterIcon
+                            cellMargin={2}
+                            dataTest="gene-column-header"
+                            className={styles.displayFlex}
+                            showFilter={true}
+                            isFiltered={isChecked}
+                            onClickCallback={() => {
+                                isChecked = !isChecked;
+                                updatePhrases(option, isChecked);
```
onursumer, line 61: Updating `isChecked` is actually a bit confusing. It's a local variable which is not observable/stateful. Maybe keep it as a constant and call the function like this?

```suggestion
                                updatePhrases(option, !isChecked);
```

src/shared/lib/query/textQueryUtils.ts
```diff
@@ -85,7 +88,20 @@ export function performSearchSingle(
         } else if (clause.isAnd()) {
             let allPhrasesMatch = true;
             for (const phrase of clause.getPhrases()) {
-                allPhrasesMatch = allPhrasesMatch && phrase.match(study);
+                if (phrase.toString().split(':')[0] === 'data-type-study') {
+                    const ph = phrase as ListPhrase;
```
onursumer, line 92:
```suggestion
                    const listPhrase = phrase as ListPhrase;
```

onursumer (changes requested):

src/shared/components/query/filteredSearch/Phrase.tsx
```diff
@@ -143,6 +160,27 @@ export class ListPhrase implements Phrase {
         return anyFieldMatch;
     }
 
+    public measuredSamples(
```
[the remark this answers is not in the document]
TJMKuijpers: Generalized it and created `function countPerDataType`

src/shared/components/query/filteredSearch/field/dataTypeFilterField.tsx
TJMKuijpers: `isChecked` is now a constant.

onursumer: Hmm, seems like it's still defined as `let`: https://github.com/cBioPortal/cbioportal-frontend/pull/4721/files#diff-c48e89227902d88456a84cf5840fa79cd17fe9ebbaddc7f331c977fb73af3779R49

TJMKuijpers:
imageTest1

src/shared/components/query/filteredSearch/field/dataTypeFilterField.tsx
```diff
@@ -0,0 +1,84 @@
...
+import { SearchClause } from 'shared/components/query/filteredSearch/SearchClause';
+
+import { FieldProps } from 'shared/components/query/filteredSearch/field/FilterFormField';
+import { ListPhrase } from 'shared/components/query/filteredSearch/Phrase';
+import { TableHeaderCellFilterIcon } from 'pages/studyView/table/TableHeaderCellFilterIcon';
+import { getDataTypeStudyOverlay } from 'pages/studyView/TableUtils';
+import { createQueryUpdate } from 'shared/components/query/filteredSearch/field/CheckboxFilterField';
+import styles from 'pages/studyView/table/tables.module.scss';
+
+export type dataTypeFilterField = {
```
onursumer, line 12: Let's define the type this way.

```suggestion
export type DataTypeFilterField = {
```

We should also name the file `DataTypeFilterField.tsx` for consistency

onursumer (changes requested):
Almost there! Just one more minor request.

TJMKuijpers:
@inodb  I have added the new files to this PR that creates the dropdown menu: https://deploy-preview-4721--cbioportalfrontend.netlify.app/

TJMKuijpers:
@onursumer after a discussion with Ino, we changed the design and the way the filtering works (not with a text query anymore). Could you have a look at the code again?

PR #4738, opened 2023-09-19
StudyView: in manual submit mode, prevent user from clearing DE-selec…

alisman:
Attempting to de-select filters using the chart controls (e.g. checkboxes) results in the following dialog:
image

### DIFF
--- a/src/pages/studyView/charts/ChartContainer.tsx
+++ b/src/pages/studyView/charts/ChartContainer.tsx
@@ -79,6 +79,7 @@ import {
     StructVarMultiSelectionTableColumnKey,
 } from 'pages/studyView/table/StructuralVariantMultiSelectionTable';
 import { StructVarGenePair } from 'pages/studyView/StructVarUtils';
+import { Modal } from 'react-bootstrap';
 
 export interface AbstractChart {
     toSVGDOMNode: () => Element;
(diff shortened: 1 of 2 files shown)


src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -79,6 +79,8 @@ import {
     StructVarMultiSelectionTableColumnKey,
 } from 'pages/studyView/table/StructuralVariantMultiSelectionTable';
 import { StructVarGenePair } from 'pages/studyView/StructVarUtils';
+import { JsxElement } from 'typescript';
```
onursumer, line 82: accidental import? I didn't even know typescript package has JsxElement :)

src/pages/studyView/charts/ChartContainer.tsx
```diff
@@ -1361,76 +1378,92 @@ export class ChartContainer extends React.Component<IChartContainerProps, {}> {
...
-                        <div className={styles.chartError}>
-                            <i className="fa fa-warning" aria-hidden="true"></i>{' '}
-                            Error when loading data.
-                        </div>
-                    )}
+            <>
+                {this.alertContent && (
+                    <Modal
+                        show={true}
+                        onHide={() => (this.alertContent = null)}
```
onursumer, line 1393: not sure if it's a big deal but this function can also be wrapped with `action`, right?

onursumer (approved):
approving, but we should probably remove the accidental unused import before merging.

gblaih (approved):

PR #4740, opened 2023-09-21
RFC69 : survival chart with landmark events and hazard ratio

TJMKuijpers:
RFC69: Add hazard ratio and landmarks in the survival plot

For clinical trials, it is customary to report hazard ratios and/or landmark values in survival plots. Currently, survival plots rendered by cBioPortal do not show hazard ratio or landmark information.

This PR is the implementation of RFC69. The new survival chart can be activated by setting `&featureFlag=SURVIVAL_PLOT_EXTENDED`.

**Implemented functions**

Number at risk
Most scientific papers show a survival plot together with the number at risk table. We have added this option to the survival plot and show the number of risks below the figure (number at risk at every tick on the x-axis).

_Landmark lines_

The user activates the landmarks option via a checkbox (by default, the landmark input field is disabled). The user can add values (comma or space separated) that are translated into vertical lines (dashed) in the survival plot.

Underneath the number at risk table, we show a new table with the percentage of patients at that specific landmark point.
Figure2_Landmarklines

_Hazard ratios_

The user activates the hazard ratio option via a checkbox. The user selects the Control group to serve as a reference (denominator) in the hazard ratio by dragging the Control group to the leftmost position in the Groups menu section. The hazard ratios for each experimental group ( with a 95% confidence interval and p-value) appear below the respective entries in the plot legend.
Figure1_HazardRatio

**Implemented tests**

_Unit test for the new functions_

- Test for the hazard ratio calculations (based on the survival R package for the ground truth)
- Test for the log-rank test to calculate the lower and upper confidence intervals (based on the survival R package)

_Implemented screenshot tests:_

- Survival chart with hazard ratio table and landmark line disabled
- Survival chart with landmark event at time point 20
- Survival chart with hazard ratio table
- Survival chart with hazard ratio table and landmark event at time point 20


src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -133,7 +133,7 @@ export default abstract class ComparisonStore extends AnalysisStore
     @observable public adjustForLeftTruncation = true;
 
     constructor(
-        protected appStore: AppStore,
+        public appStore: AppStore,
```
TJMKuijpers, line 136: This one should be made protected again and the featureFlag for the extended survival plot should be shared with <Survival> in a better way.

src/pages/groupComparison/GroupComparisonUtils.tsx
```diff
@@ -948,6 +948,23 @@ export function getStatisticalCautionInfo() {
         </div>
     );
 }
+export function getHazardRatioCautionInfo() {
```
alisman, line 951: might as well make this a functional component so it's clear what it is.

alisman: see AlterationFilterMenuSection directly below

src/pages/groupComparison/Survival.tsx
```diff
@@ -317,6 +319,7 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
                             style={{ paddingBottom: 0 }}
                         >
                             {getStatisticalCautionInfo()}
+                            {getHazardRatioCautionInfo()}
```
alisman, line 322: maybe do the same for getStatisticalCautionInfo while we're at it. (Function component)

src/pages/resultsView/survival/hazardRatioGroundTruth.r
```diff
@@ -0,0 +1,50 @@
+# This R script calculated the hazard ratio (which we will use to test our hazard ratio function against)
```
alisman, line 1: should this be committed to repo?

src/pages/resultsView/survival/logRankTest.ts
```diff
@@ -309,6 +309,89 @@ export function logRankTest(...survivalCurves: SurvivalCurveItem[][]) {
...
+    controlGroup: string,
+    ...survivalCurves: SurvivalCurveItem[][]
+) {
+    const allEvents = _.sortBy(_.flatten(survivalCurves), s => s.months);
+    const curve = new SurvivalCurve(allEvents);
+    const sortedCurves = survivalCurves.map(curveData => {
+        return _.sortBy(curveData, d => d.months);
+    });
+
+    const groupsTte = sortedCurves.map(curveData => {
```
alisman, line 325: can you improve this name?  No idea what Tte means. Ir Ev.

src/pages/resultsView/survival/logRankTest.ts
```diff
@@ -309,6 +309,89 @@ export function logRankTest(...survivalCurves: SurvivalCurveItem[][]) {
...
         .filter(x => x.d > 0);
-
     return _logRankTest(survivalItems, groupsTte, groupsEv).pValue;
 }
+
+export function calculatePairWiseHazardRatio(
+    controlGroup: string,
+    ...survivalCurves: SurvivalCurveItem[][]
+) {
+    const allEvents = _.sortBy(_.flatten(survivalCurves), s => s.months);
```
alisman, line 319: this is important business logic here so comments about what we're doing would be helpful to devs trying to understand the basics

src/pages/resultsView/survival/SurvivalChartExtended.tsx
```diff
@@ -0,0 +1,1684 @@
...
+                    </td>
+                ))}
+            </tr>
+        ));
+    }
+
+    @computed get tableRowsHazardRatio() {
+        if (this.hazardRatioGroups !== null) {
+            return (
+                <React.Fragment>
```
alisman, line 1498: i don't think you need Fragment here since you are just returning a single element (table)

src/pages/groupComparison/Survival.tsx
TJMKuijpers: Also changed it for `getStatisticalCautionInfo` and also updated `/groupComparison/ClinicalData.tsx` since it also calls `getStatisticalCautionInfo`

alisman:
@pvannierop I wanted to open a discussion about the practice here of copying an existing component in order to "extend it".   Based on discussion with @TJMKuijpers, I now think I understand that ```<SurvivalExtended>``` is intended to fully replace ```<SurvivalChart>``` and so it's not really an "extension."  It's just adding to adding to the existing component (right?).  The issue then is that by copying the component in order to work on it, we leave ourselves vulnerable to losing concurrent edits to the original component.  If we want to use a feature flag to introduce this work then I think it should be be done via  "mode" of the original component.  Alternatively, we don't use a feature flag in this case.  We just keep the feature branch open.  Since this is a pretty isolated feature, I don't see a big cost to that.

TJMKuijpers:
IMageSV

TJMKuijpers:
@inodb updated the labels for the landmark line groups (had to change `CBIOPORTAL_VICTORY_THEME.chart.fontFamily,`  to `CBIOPORTAL_VICTORY_THEME.legend.style.labels.fontFamily`). Also updated the reference screenshots so the surival chart tests should pass. 
ImageSV1

PR #4742, opened 2023-09-22
Resolve fetching genomic chart datatype for logged in user

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/10401


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6977,13 +6977,24 @@ export class StudyViewPageStore
                 chartUserSettings.hugoGeneSymbol &&
                 chartUserSettings.profileType
             ) {
+                const molecularProfileOptionByTypeMap = _.keyBy(
```
alisman, line 6980: since you are only attempting to find one item here, i think it would be as fast and much more clear just to use _.find() on the options array.

alisman (approved):

PR #4588, opened 2023-04-18
Structural Variant table

fuzhaoyuan:
cBioPortal/cbioportal#10075

Describe changes proposed in this pull request:

Features:
- [x] Add columns: Mutation Status, Cancer Type Detailed, Annotation
- [x] Display number of duplicate samples

Refinements:
- [x] Fulfill features: search Cancer Type Detailed, calculate exons
- [x] Clean up: fusion bar / pie charts, unused columns
- [x] Rename: Sample Id -> ID, Annotation -> SV Description
- [x] Fix: sample link, gene tab switch

PR #4745, opened 2023-09-27
Enable Results View Structural Variants Tab

inodb:
Remove the feature flag to enable structural variants tab introduced here:

https://github.com/cBioPortal/cbioportal-frontend/pull/4588

### DIFF
--- a/src/pages/resultsView/ResultsViewPage.tsx
+++ b/src/pages/resultsView/ResultsViewPage.tsx
@@ -283,16 +283,12 @@ export default class ResultsViewPage extends React.Component<
             {
                 id: ResultsViewTab.STRUCTURALVARIANTS,
                 hide: () => {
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/ResultsViewPage.tsx
```diff
@@ -282,18 +282,6 @@ export default class ResultsViewPage extends React.Component<
             },
             {
                 id: ResultsViewTab.STRUCTURALVARIANTS,
-                hide: () => {
-                    if (this.appStore.featureFlagStore.has('SVTAB')) {
-                        return (
-                            !this.resultsViewPageStore.structuralVariants
-                                .isComplete ||
-                            this.resultsViewPageStore.structuralVariants.result
```
alisman, line 290: @inodb we probably still want to hide it when there are no SV results, right?

alisman (approved):

PR #4747, opened 2023-09-29
Configure left logo in the cBioPortal header via 'skin.left_logo' property

pvannierop:
The left logo in the header can be configured in the portal.properties file via the 'skin.left_logo' property. When property is undefined the default header will be used.

### DIFF
--- a/src/appShell/App/PortalHeader.tsx
+++ b/src/appShell/App/PortalHeader.tsx
@@ -4,7 +4,10 @@ import { Link, NavLink } from 'react-router-dom';
 import { If, Then, Else } from 'react-if';
 import { AppStore } from '../../AppStore';
 import { observer } from 'mobx-react';
-import { getInstituteLogoUrl } from '../../shared/api/urls';
+import {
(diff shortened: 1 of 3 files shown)


src/appShell/App/PortalHeader.tsx
```diff
@@ -139,7 +146,7 @@ export default class PortalHeader extends React.Component<
                                 <div className="identity">
                                     <Dropdown id="dat-dropdown">
                                         <Dropdown.Toggle className="btn-sm">
-                                            Logged in as{' '}
+                                            Logged in as +
```
alisman, line 149: @pvannierop did you mean to put that plus sign there (just before the username) 
result:
image

alisman (changes requested):
see comment

src/appShell/App/PortalHeader.tsx
pvannierop: Eh no, most definitely not. I will revert. Must have been a accidental keyboard press?

PR #4748, opened 2023-09-29
On patient page, default to SBS mutational signatures

alisman:
On the patient page, we want to default to the SBS mutational signature profile in spite of arbitrary data ordering. Usually when people speak of mutational signatures they mean single base substitutions


src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
```diff
@@ -188,10 +188,17 @@ export function retrieveMutationalSignatureVersionFromData(
...
     if (uniqueProfileVersion !== undefined) {
-        return uniqueProfileVersion.includes('v3') &&
+        if (
+            uniqueProfileVersion.includes('v3') &&
             uniqueProfileVersion.includes('v2')
-            ? 'v3'
-            : uniqueProfileVersion[0]!;
+        ) {
+            return 'v3';
+        } else if (uniqueProfileVersion.includes('SBS')) {
```
alisman, line 196: @inodb since we already have a mechanism for selecting default mutation sig profile,  i thought it best to enforce the SBS here rather than monkey with data ordering in a hackish way.  what do you think?

inodb (approved):
LGTM - thanks!

src/shared/lib/GenericAssayUtils/MutationalSignaturesUtils.tsx
inodb: yeah that looks good to me 👍

PR #4750, opened 2023-10-03
match cna chart colors with oncoprint and adjust label colors accordi…

h164654156465:
fix https://github.com/cBioPortal/cbioportal/issues/10367 

- match cna pie chart colors with Oncoprint color in study view page
- make all pie chart label display either white or black color based on their background colors, which in turns make it easier for users to distinguish between label and its background

Screenshot 2023-10-03 at 1 28 34 PM


src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -207,6 +208,13 @@ export default class PieChart extends React.Component<IPieChartProps, {}>
             : d.count.toLocaleString();
     }
 
+    @computed
```
alisman, line 211: it doesn't make sense to make this a computed property because it's returning a callback which will be executed outside the scope of mobx's optimization.  you should make this a bound method of this class instead of a computed getter.  the cost of deriving the color for a handful of piechart slices is negligible.

alisman: the importance of this comment is mostly just to make sure you understand what the computed decorator does.  you can read about it in mobx documentation if necessary.

h164654156465: I looked at line 135 fill() methods using @computed as well, it also returns a method but not a value. Should I change its tag to @autobind as well? I've tried to change either fill() or labelColor() and got "@boundMethod decorator can only be applied to methods not: undefined"

PR #4756, opened 2023-10-17
Add group percentages to oncoprint

alisman:
- [x] Show group percentages in Oncoprint when gaps are turned on
- [x] Show labels in SVG download
- [x] Should also work in Oncoprinter
- [x] Should work in merged tracks


packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
```diff
@@ -1898,6 +2065,34 @@ export default class OncoprintWebGLCellView {
...
+                        });
+
+                        textElt.textContent = gap.labelFormatter();
+                        root.appendChild(textElt);
+                    });
+                }
+            }
+
+            // aaron
+            // aaron
```
onursumer, line 2094: we should remove/replace these comments

src/pages/staticPages/tools/oncoprinter/OncoprinterGeneticUtils.ts
```diff
@@ -608,17 +618,49 @@ export function getGeneticTracks(
...
+                    model.data_groups.update(model);
+                    const groupsByTrackMap = model.data_groups.get();
+
+                    const percentagesForGroups = groupsByTrackMap[gene][0].map(
+                        groupData => {
+                            return getPercentAltered(
+                                groupData as OncoprinterGeneticTrackDatum[]
+                            );
+                        }
+                    );
```
onursumer, line 649: would be simpler if we could do something like

```suggestion
                    const percentagesForGroups = groupsByTrackMap[gene][0].map(getPercentAltered);
```

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -804,6 +802,95 @@ export function makeGeneticTrackWith({
...
+                                                    geneSymbolArray,
+                                                    gene =>
+                                                        sequencedPatientKeysByGene[
+                                                            gene
+                                                        ]
+                                                ),
+                                                groupData.map((d: any) => d.uid)
+                                            );
+                                        })
+                                        .value();
```
onursumer, line 867: this `else` block is almost identical to the `if` block above except the variables `sequencedSampleKeysByGene` vs `sequencedPatientKeysByGene`. probably better to extract the logic to a function and reuse.

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
```diff
@@ -294,9 +308,43 @@ export default class OncoprintWebGLCellView {
...
+                            $(
+                                `<span>${overlappingGap.data.tooltipFormatter()}</span>`
+                            ),
+                            false
+                        );
+                    }
+
+                    if (!overlapping_data && !overlappingGap) {
+                        tooltip.hideIfNotAlreadyGoingTo(150);
+                    }
```
onursumer, line 347: so much work to show a tooltip but I guess we have no other choice :disappointed:

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
```diff
@@ -364,6 +427,27 @@ export default class OncoprintWebGLCellView {
...
+            //ctx.clearColor(1.0, 1.0, 1.0, 1.0);
+            //ctx.clear(ctx.COLOR_BUFFER_BIT | ctx.DEPTH_BUFFER_BIT);
+            // ctx.viewportWidth = canvas.width;
+            // ctx.viewportHeight = canvas.height;
+            // ctx.viewport(0, 0, ctx.viewportWidth, ctx.viewportHeight);
+            //ctx.enable(ctx.DEPTH_TEST);
+            //ctx.enable(ctx.BLEND);
+            //ctx.blendEquation(ctx.FUNC_ADD);
+            //ctx.blendFunc(ctx.SRC_ALPHA, ctx.ONE_MINUS_SRC_ALPHA);
+            //ctx.depthMask(false);
```
onursumer, line 443: should we remove this or add a comment why we commented it out?

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
```diff
@@ -534,11 +618,34 @@ export default class OncoprintWebGLCellView {
...
+        //ctx.clearColor(1.0, 1.0, 1.0, 1.0);
+        //ctx.clear(ctx.COLOR_BUFFER_BIT | ctx.DEPTH_BUFFER_BIT);
+        //ctx. = canvas.width;
+        //ctx.viewportHeight = canvas.height;
+        //ctx.viewport(0, 0, ctx.viewportWidth, ctx.viewportHeight);
+        //ctx.enable(ctx.DEPTH_TEST);
+        //ctx.enable(ctx.BLEND);
+        //ctx.blendEquation(ctx.FUNC_ADD);
+        //ctx.blendFunc(ctx.SRC_ALPHA, ctx.ONE_MINUS_SRC_ALPHA);
+        //ctx.depthMask(false);
```
onursumer, line 643: same as above, except here we have a syntax error `ctx. = canvas.width;` :)

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
```diff
@@ -1898,6 +2065,34 @@ export default class OncoprintWebGLCellView {
                 track_id,
                 false
             );
+
+            const custom = model.getTrackCustomOptions(track_id);
+            if (gapOffsets[0]) {
+                const gaps = _.isEmpty(model.ids_after_a_gap.get())
+                    ? undefined
+                    : custom.find(t => !!t.gapLabelsFn)?.gapLabelsFn(model);
```
onursumer, line 2073: duplicate code segment. should we define a `getGaps` function? we can probably include this one as well `const custom = model.getTrackCustomOptions(track_id);`

onursumer (commented):
This is great. Thanks, Aaron! :1st_place_medal:

Just added a few minor suggestions.

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
alisman: @onursumer  I think It's been said the hardest problems in computer science are 1) knowing when to clear a cache and 2) making a tooltip over an HTML canvas.

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
alisman: why?

packages/oncoprintjs/src/js/oncoprintwebglcellview.ts
alisman: good idea

PR #4758, opened 2023-10-18
Fix incorrect oncokb data and cancer type label used in mutations table

gblaih:
Fixes https://github.com/cBioPortal/cbioportal/issues/10421

Currently, when `CANCER_TYPE_DETAILED` is the same for all mutations of a row in the table, it uses this to fetch the oncoKb data; otherwise, it uses "Cancer of Unknown Primary".

This fixes it so that if there are multiple `CANCER_TYPE_DETAILED` for mutations of a row, it checks if `CANCER_TYPE` is the same and if so, uses that to fetch the oncoKb data. If there are multiple `CANCER_TYPE` as well, then it uses "Cancer of Unknown Primary".

### DIFF
--- a/src/pages/groupComparison/GroupComparisonMutationTable.tsx
+++ b/src/pages/groupComparison/GroupComparisonMutationTable.tsx
@@ -57,15 +57,7 @@ export interface IGroupComparisonMutationTableProps
     };
 }
 
-export function hasMultipleCancerTypes(
-    data: Mutation[],
-    resolveTumorType: (mutation: Mutation) => string
-): boolean {
-    return !_.every(
-        data,
-        m => resolveTumorType(m) === resolveTumorType(data[0])
-    );
-}
+export type TumorTypeAttribute = 'CANCER_TYPE' | 'CANCER_TYPE_DETAILED';
 
 export default class GroupComparisonMutationTable extends MutationTable<
     IGroupComparisonMutationTableProps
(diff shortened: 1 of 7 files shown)


src/pages/groupComparison/GroupComparisonMutationTable.tsx
```diff
@@ -92,6 +84,54 @@ export default class GroupComparisonMutationTable extends MutationTable<
...
+    }
+
+    // determine what function to use to resolve tumor type based on what tumor type attribute is the same for data
+    protected getResolveTumorType(data: Mutation[]) {
+        if (this.getSameTumorTypeAttribute(data) === 'CANCER_TYPE_DETAILED') {
+            return this.resolveTumorType;
+        }
+        if (this.getSameTumorTypeAttribute(data) === 'CANCER_TYPE') {
+            return this.resolveCancerType;
+        }
```
inodb, line 120: this `resolveTumorType` vs `resolveCancerType` naming feels a little weird, tho i guess it was already there. Not super clear what the difference between those is

src/shared/lib/StoreUtils.ts
```diff
@@ -1122,17 +1122,43 @@ export function findMrnaRankMolecularProfileId(
...
+                }
+            );
+        }
+
+        // // first priority is CANCER_TYPE_DETAILED in clinical data
+        // _.each(clinicalDataForSamples.result, (clinicalData: ClinicalData) => {
+        //     if (clinicalData.clinicalAttributeId === 'CANCER_TYPE_DETAILED') {
+        //         map[clinicalData.uniqueSampleKey] = clinicalData.value;
+        //     }
+        // });
```
inodb, line 1161: maybe good to remove commented out code?

inodb (commented):
Looks great! Functionality seems to work from product perspective. Made a few minor comments on the code

src/pages/groupComparison/GroupComparisonMutationTable.tsx
gblaih: yea, I wasn't sure what to name it but `resolveTumorType` typically gets the `Cancer Type Detailed` from a mutation while the new `resolveCancerType` will retrieve the `Cancer Type`. any ideas on name changes? throughout the codebase, the term "Tumor Type" is used but mostly just for "Cancer Type Detailed" and I didn't want to rename all of that.

inodb: I see - yeah as long as it is consistent that's fine by me. If we were to change it throughout the code I'd prolly be explicit and just say resolveCancerType vs resolveCancerTypeDetailed

alisman:
@gblaih did you respond to Ino's comments?

PR #4762, opened 2023-10-25
Refactor to allow pie and table charts to download both full and summ…

alisman:
Product team noticed that we only allow Summary and Full download options on pie charts.  This generalizes the behavior.  Only question is are there some times of table chart which should not offer summary?

### DIFF
--- a/src/pages/studyView/tabs/SummaryTab.tsx
+++ b/src/pages/studyView/tabs/SummaryTab.tsx
@@ -240,7 +240,7 @@ export class StudySummaryTab extends React.Component<
                 }
                 props.onChangeChartType = this.handlers.onChangeChartType;
                 props.getData = (dataType?: DataType) =>
-                    this.store.getPieChartDataDownload(chartMeta, dataType);
+                    this.store.getChartDownloadableData(chartMeta, dataType);
                 props.downloadTypes = [
                     'Summary Data',
                     'Full Data',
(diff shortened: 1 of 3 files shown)


src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -374,9 +374,9 @@ export class StudySummaryTab extends React.Component<
                     props.onResetSelection = this.handlers.onValueSelection;
                 }
                 props.onChangeChartType = this.handlers.onChangeChartType;
-                props.getData = () =>
-                    this.store.getChartDownloadableData(chartMeta);
-                props.downloadTypes = ['Data'];
+                props.getData = chartType =>
+                    this.store.getChartDownloadableData(chartMeta, chartType);
+                props.downloadTypes = ['Summary Data', 'Full Data'];
```
alisman, line 379: we add summary and full data to table charts (same as pie charts).

src/pages/studyView/tabs/SummaryTab.tsx
```diff
@@ -374,9 +374,9 @@ export class StudySummaryTab extends React.Component<
                     props.onResetSelection = this.handlers.onValueSelection;
                 }
                 props.onChangeChartType = this.handlers.onChangeChartType;
-                props.getData = () =>
-                    this.store.getChartDownloadableData(chartMeta);
-                props.downloadTypes = ['Data'];
+                props.getData = chartType =>
+                    this.store.getChartDownloadableData(chartMeta, chartType);
```
onursumer, line 378: Minor, but should we name the param `dataType`?

```suggestion
                props.getData = dataType =>
                    this.store.getChartDownloadableData(chartMeta, dataType);
```

onursumer (commented):
Looks good! One thing is that now both files have the same filename. Should we also update file names like `Somatic_Status_summary.txt` and `Somatic_Status_full.txt`?

PR #4767, opened 2023-11-02
Study view: fix comparison session generation from study view bin cha…

alisman:
Charts in study view sometimes have bins which are single integers.  In this case, the "bins" are represtend as having the same start and end value.  In this case, to fall into this bin, the values must equal the start/end


dippindots (approved):
LGTM, I agree it's ok to add this special case to our codebase.

PR #4768, opened 2023-11-07
Clinical attribute types with priority of zero should not appear by default on study view

alisman:
Fix https://github.com/cBioPortal/cbioportal/issues/10351


dippindots (approved):
LGTM! Thanks!

PR #4771, opened 2023-11-09
Mutational signature plot with cosmic reference signature

TJMKuijpers:
This PR adds the cosmic reference signatures to the mutational signature plot.  This will add a new graph to the existing mutational signature graph to show the percentage of a mutation (SBS, DBS, or ID) in a signature of interest. 

image

It also fixes the following issues:
- Table rows are highlighted on hoover and colored when clicked.
-  signatures category 'Unknown' has a color.
-  Removed word "sample" to every sample ID in the sample menu dropdown menu.
-  Summary stacked bar chart only displays top 10 (no overlfow when there are many significant mutational signatures
-  Minimum and maximum height are set for Y-axis when no data is present (for some patients).
-  `selectedSignature` --> empty string is not a good default value is fixed.
- Updated test from the original mutational signatures.

## Checks

## Any screenshots or GIFs?

## Notify reviewers


TJMKuijpers:
@inodb this PR should extend the current mutational signature plot with the cosmic reference. Maybe we can update the current mutational signature plot after code review on the private instance (for the pancan study)?

inodb:
@TJMKuijpers Looks great!

## Product feedback

- [ ] `Mutational count (value)` -> `Mutation count` - I think this is just nr of mutations (or percentages) right?
- [ ] I think counts should always be integers (here y axis shows `.5`):
    
    image
- [ ] Make tooltip on bars more clear, maybe `43 single nucleotide variants (x%) with a T(C>T)A or A(T>C)T change`. Can do same for indels and DBS
    image
- [ ] Add info icon explaining what SBS, DBS and ID are. Maybe right after the dropdown?
    image
- [ ] Maybe make it more clear the bottom shows a reference signature? E.g.: `Percentage of DBS` -> `COSMIC Reference` and `DBS2 (Smoking)` -> `DBS2 (Smoking) (%)`
    
image


Other thoughts:
- Switching the sample signature display inside of the tab feels a little odd, but I think it's ok for now:
     
image

## Code Review
The cosmic reference files are stored directly in the frontend. That seems ok, as it's only 9.9MB and they stay pretty static:

```
9.9M    src/pages/patientView/mutationalSignatures/cosmic_reference.json
```

We also have some other reference files that are stored in the repo:

```
 17M    src/shared/lib/data/reference_genome_hg19.json
```

Just wondering if we should organize this a bit better, so it's clear both of these use a similar mechanism (e.g. they are stored in some shared `reference` files folder)

TJMKuijpers:
Community meeting remarks
- [ ] Add sample summary (id and number of mutations next to y-axis button)
- [ ] Right justify the numbers in mutational signature table
- [ ] Don't show a reference signature by default but let the user pick one.
- [ ] Add the cosmic annotation for the x-axis in ID plot
- [ ] Remove the parenthesis in the SBS axis label
- [ ] Top legend boxes: increase width and spacing
- [ ] General styling of the plot

inodb:
Thanks @TJMKuijpers !

One more minor thing:

- [ ] PDF download does not work for me
- [ ] Can we call the tab `Mutational Signatures` (so remove "Data")
- [ ] For the y axis, can we name it the same as for COSMIC? E.g. for SBS, write it out: `Single Base Substitutions (%)`, `Indels (%)`, `Double Base Substitutions (%)`

Other thoughts:

- When looking at [multiple samples for one patient](https://private.cbioportal.mskcc.org/patient/mutationalSignatures?studyId=pancan_pcawg_2020&caseId=DO51954#navCaseIds=pancan_pcawg_2020:DO51954), the select by row feels a little odd. A select per cell might be cleaner, where you e.g. select the sample + reference in that cell. We can follow up with this later, but just a thought

schultzn:
Thanks @TJMKuijpers 

I agree with Ino's comments above. Some other thoughts:

- [ ] Can you rename "Version" in the top left corner to "Variant Class" to be consistent with COSMIC nomenclature?
- [ ] Does it make sense to always show the y-axis as 100%? COSMIC seems to dynamically change the y-axis to match the data.
- [ ] In the bar chart tooltips for small insertions and deletions, 1 bp insertions and deletions, the last word needs to be removed (deletion / insertion). 
- [ ] It is not clear at all to me how to switch samples for a patient with multiple samples, e.g. [this one](https://private.cbioportal.mskcc.org/patient/mutationalSignatures?studyId=pancan_pcawg_2020&caseId=DO46329).

tmazor:
I agree with Ino & Niki's comments, and have a few additional notes:
- [ ] I think the axis labels & tooltips using a different font than the rest of the site - if that's true, please standardize
image
- [ ] Looks like there's a missing space in the tooltip after each colon
image
- [ ] I suggest some slight changes to the tooltip text that I think will help with clarity & interpretation:
Instead of "20% SBS of [C->T] around A[C->T]G"
image
how about "20% of SBS mutations (x/y) are C>T at A[C>T]G trinucleotide" where x/y is the fraction used to calculate 20%.
Similarly for DBS:
"33% DBS of CC to AG" --> "33% of DBS mutations (1/3) are CC>AG"
image
And for ID:
"11% 1bp deletion(C) with homopolymer length of 0 deletion" --> "11% of small indels (x/y) are 1bp deletion of C in a homopolymer of length 0"
image

- [ ] Tooltip is cutoff when it runs into the edge of the plot
image
https://private.cbioportal.mskcc.org/patient/mutationalSignatures?studyId=pancan_pcawg_2020&caseId=DO46334

TJMKuijpers:
> * [ ]  PDF download does not work for me

@inodb I'm able to download the plot as PDF in Chrome, Brave and FireFox. Is this a Safari issue?

TJMKuijpers:
how to switch samples for a patient with multiple samples, e.g. [this one](https://private.cbioportal.mskcc.org/patient/mutationalSignatures?studyId=pancan_pcawg_2020&caseId=DO46329).

The sample selector disappears when not all samples are profiled for mutational signatures (which is the case for DO46329). I added a banner to make this more clear.
image

TJMKuijpers:
@schultzn 
```
Does it make sense to always show the y-axis as 100%? COSMIC seems to dynamically change the y-axis to match the data.
```
I think it does make sense if you want to compare the sample profile versus the reference profile. I think it would be more difficult to compare the two if the y-axis match the data (for example when sample mutational profile is between 0-50% and the reference plot between 0-10%)

inodb (approved):
Thanks so much for fixing all the feedback! Great job! Updated some localdb screenshots, will merge shortly

PR #4775, opened 2023-11-10
Update civic API

leexgh:
Fix: https://github.com/cBioPortal/cbioportal/issues/9521
Civic stops supporting v1 API and switches to v2 graphql API. This PR changes civic rest API to graphql API, other things remain the same. 
Now we only need to call one API to get civic gene information and all variants associated with the gene, not need to keep two calls for gene and variant separately. 
image


inodb:
@leexgh Thanks for fixing this! Looks like it's not yet working for copy number data:

https://cbioportal.mskcc.org/patient?studyId=ucec_tcga_pub&caseId=TCGA-BK-A0CC

onursumer:
Looks like we are querying the civic API separately for each gene in the CNA and mutation tables on patient view. Maybe that's why the CNA table is failing?

packages/cbioportal-utils/src/civic/CivicUtils.ts
```diff
@@ -24,42 +24,15 @@ export enum CivicAlterationType {
...
     // Assemble a list of promises, each of which will retrieve a batch of genes
     let promises: Array<Promise<Array<ICivicGeneSummary>>> = [];
 
-    // To prevent the request from growing too large, we send it off in multiple chunks
-    const chunkedIds: number[][] = _.chunk(_.uniq(entrezGeneIds), 400);
-
-    chunkedIds.forEach(entrezGeneIds =>
-        promises.push(civicClient.getCivicGenesBatch(entrezGeneIds.join(',')))
+    hugoSymbols.forEach(hugoSymbol =>
+        promises.push(civicClient.getCivicGenesBatch(hugoSymbol))
```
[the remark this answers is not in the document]
leexgh: Yes that's the problem right now. I don't see a way to send "POST" queries that can accept a list of hugo gene symbols. As far as I can see it can only accept single gene or giving the "start civic id" and "end civic id", the response will then include all genes between this range, but we still need to send extra queries to get the civic id, so it's still not ideal. 
For results view it only sends one gene, but for copy number alterations on patients view it sends too many requests. I already asked civic team to see if they have a better way to handle it.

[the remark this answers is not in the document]
leexgh: This is solved. Civic team just made a new release yesterday, they can support a list of hugo symbols now in their genes API.

onursumer (commented):
Looks good. Thanks!

Just added a few minor suggestions.

packages/cbioportal-utils/src/civic/CivicDataFetcher.ts
```diff
@@ -1,192 +1,271 @@
...
                     url: 'https://civicdb.org/genes/' + record.id + '/summary',
-                    variants: createVariantMap(record.variants),
-                }));
+                    variants,
+                };
+                return geneSummary;
             });
+            for (const geneSummary of geneSummaries) {
+                result.push(await geneSummary);
+            }
```
onursumer, line 170: Can we do this within the `_.map` callback above? Looks like we are already awaiting the result there. If not, a `Promise.all` is probably better than an additional for loop, something like this:

```suggestion
            result.push(...(await Promise.all(geneSummaries)));
```

packages/react-mutation-mapper/src/component/civic/CivicCard.tsx
```diff
@@ -30,15 +30,15 @@ export default class CivicCard extends React.Component<ICivicCardProps, {}> {
...
-                        ', ';
+                for (const [key, value] of Object.entries(
+                    variant.evidenceCounts
+                )) {
+                    if (value !== 0) {
+                        if (entryTypes !== '') entryTypes += ', ';
+                        // Remove "Count" substring out of key string
+                        entryTypes += `${key.slice(0, -5)}: ${value}`;
+                    }
                 }
```
onursumer, line 41: can we do `.filter(value => value !== 0)`, `.map()`, and then `.join(', ')`? it's cleaner that way.

onursumer (commented):
Added a few minor suggestions

alisman:
@leexgh i'd like to suggest that we not use Apollo client for this.  It's a rather heavy dependency to bring into the project just for one fetch and brings a lot of unecessary code/complication.    We should be able to fetch data from a graphql endpoint using fetch.  We just have to find the data format the endpoint expects and generate/serialize it.   I'm happy to help with this and it shouldn't take more than an hour i wouldn't think.  

```
fetch(<graphql_endpoint>, {
  method: 'POST',
  headers: { 'Content-Type': 'application/json' },
  body: JSON.stringify({ query: '{ posts { title } }' }),
})
.then(res => res.json())
.then(res => console.log(res.data));
```

packages/cbioportal-utils/src/model/Civic.ts
```diff
@@ -3,34 +3,23 @@ export interface ICivicGeneSummary {
...
 }
-
-export interface ICivicEvidenceSummary {
-    id: number;
-    type: string;
-    clinicalSignificance: string;
-    level: EvidenceLevel;
-    drugs: string[];
-    disease?: string;
+export interface EvidenceCountsByType {
```
alisman, line 16: lets put Civic in the name since it's particular to Civic, right?

packages/cbioportal-utils/src/civic/CivicUtils.ts
```diff
@@ -88,67 +53,58 @@ export function getCivicVariants(
...
     let civicVariants: ICivicVariantIndex = {};
-    let promises: Array<Promise<void>> = [];
-
     if (mutationSpecs) {
-        let calledVariants: Set<number> = new Set([]);
-        for (let mutation of mutationSpecs) {
-            let geneSymbol = mutation.gene.hugoGeneSymbol;
-            let geneEntry = civicGenes[geneSymbol];
-            let proteinChanges = [mutation.proteinChange];
+        const geneToProteinChangeSet: {
```
alisman, line 57: @leexgh just took another look at this code and i think it could use a few unit tests.  it doesn't seem like the split protein changes are covered by the current spec, but i could be wrong.

packages/cbioportal-utils/src/civic/CivicUtils.ts
```diff
@@ -88,67 +53,58 @@ export function getCivicVariants(
...
+            } else {
+                for (const splitProteinChange of splitProteinChanges) {
+                    acc[geneSymbol].add(splitProteinChange);
+                }
+            }
+            return acc;
+        }, {} as { [geneSymbol: string]: Set<String> });
+
+        for (const geneSymbol in geneToProteinChangeSet) {
+            if (geneSymbol in civicGenes) {
```
alisman, line 74: i think you are really just iterating through civicGenes, so no need for the outer for loop.

packages/cbioportal-utils/src/civic/CivicUtils.ts
```diff
@@ -88,67 +53,58 @@ export function getCivicVariants(
...
+                const proteinChangeSet = geneToProteinChangeSet[geneSymbol];
+                const geneVariants = civicGenes[geneSymbol].variants;
+                for (const variantName in geneVariants) {
+                    if (proteinChangeSet.has(variantName)) {
```
alisman, line 78: again, if proteinChangeSet is a subjset of geneVarits, then you could just iterate through proteinChangeSet and make this code clearer.

packages/cbioportal-utils/src/civic/CivicUtils.ts
leexgh: Good point, all genes in civicGenes should be in mutationSpecs, so we could change it to iterating through civicGenes

packages/cbioportal-utils/src/civic/CivicUtils.ts
leexgh: proteinChangeSet is not a subset of geneVarits, they both contain some variants that are not in the other set. We can do something like find shared variants from the two sets and iterate from that list, but it doesn't make the logic easier, so maybe we can keep this

packages/cbioportal-utils/src/civic/CivicUtils.ts
leexgh: Add unit tests for this

alisman (approved):

PR #4787, opened 2023-11-29
Use appropriate genome-nexus url depending on study

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/10471


src/pages/staticPages/tools/mutationMapper/MutationMapperToolStore.ts
```diff
@@ -100,6 +100,12 @@ export default class MutationMapperToolStore {
     ) {
         makeObservable(this);
         this.mutationData = mutationData;
+        if (
```
alisman, line 103: why not just put this down by the existing logic in get genomeNexusClient?

alisman (approved):

PR #4804, opened 2023-12-18
Fix: data type filter set default value for checkbox

TJMKuijpers:
Fix: Without the default value set for the checkbox of the data type filter menu, the state of the checkbox is remembered but the checkbox is not 'checked'.

Bug can be seen when a user filters on data type, select multiple studies and clicks on 'Query by Gene'. When the user clicks on 'modify' selected studies, the 'checked' checkbox is not blue.


inodb (approved):
LGTM - thank you!

PR #4817, opened 2023-12-27
Add user color configuration for oncoprinter

gblaih:
Implements https://github.com/cBioPortal/cbioportal/issues/10502


src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +397,177 @@ export default class Oncoprinter extends React.Component<
...
+    }
+
+    @computed get selectedClinicalTrack() {
+        return _.find(
+            this.props.store.clinicalTracks,
+            t => t.key === this.trackKeySelectedForEdit
+        );
+    }
+
+    @computed get defaultClinicalAttributeColoringForStringDatatype() {
```
alisman, line 428: seems odd to me that the defaults would be derived from selected, no?  maybe i don't quite understand

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +397,177 @@ export default class Oncoprinter extends React.Component<
...
+                hexToRGBA
+            );
+        }
+        return _.mapValues(RESERVED_CLINICAL_VALUE_COLORS, hexToRGBA);
+    }
+
+    @autobind
+    private getDefaultSelectedClinicalTrackColor(value: string) {
+        if (!this.selectedClinicalTrack) {
+            return [255, 255, 255, 1];
```
alisman, line 443: make a constant out of this.  its BLACK right?

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +397,177 @@ export default class Oncoprinter extends React.Component<
...
+        }
+        if (this.selectedClinicalTrack.datatype === 'counts') {
```
alisman, line 445: use SWITCH statement for this instead of if/elseif

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +397,177 @@ export default class Oncoprinter extends React.Component<
...
+            return getClinicalTrackValues(this.selectedClinicalTrack);
+        }
+        return [];
+    }
+
+    @observable clinicalTrackColorChanged: boolean = false;
+
+    @action.bound
+    setClinicalTrackColorChanged(changed: boolean) {
+        this.clinicalTrackColorChanged = changed;
```
alisman, line 473: seems odd to keep track of changed this way.  i would expect this status to be derivable from the selection itself?

src/pages/staticPages/tools/oncoprinter/OncoprinterStore.ts
```diff
@@ -589,4 +606,34 @@ export default class OncoprinterStore {
...
+        ) {
+            delete this._userSelectedClinicalTracksColors[label][value];
+        } else if (color) {
+            if (!this._userSelectedClinicalTracksColors[label]) {
+                this._userSelectedClinicalTracksColors[label] = {};
+            }
+            this._userSelectedClinicalTracksColors[label][value] = color;
+        }
+        localStorage.setItem(
+            'oncoprinterClinicalTracksColorConfig',
```
alisman, line 631: make this a constant so it can be use to read it as well

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
gblaih: the defaults are obtained from `RESERVED_CLINICAL_VALUE_COLORS` but we need the selected track data to get the values not present in `RESERVED_CLINICAL_VALUE_COLORS` so we can assign them their default colors.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
gblaih: the track changed can be found by checking if the value-to-color mappings are equal before and after change. i thought it would be better to have a property we can check that just tells us if something has changed in the mapping rather than doing an equality check between the two objects every time.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +399,168 @@ export default class Oncoprinter extends React.Component<
...
+                    this.selectedClinicalTrack.data
+                ),
+                hexToRGBA
+            );
+        }
+        return _.mapValues(RESERVED_CLINICAL_VALUE_COLORS, hexToRGBA);
+    }
+
+    @autobind
+    private getDefaultSelectedClinicalTrackColor(value: string) {
```
alisman, line 443: maybe "getDefaultSelectedClinicalTrackColorForValue" just so it's clear that it's dependent on the value.  Also, maybe change name of value param to attributeValue.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +399,168 @@ export default class Oncoprinter extends React.Component<
...
+        if (!this.selectedClinicalTrack) {
```
alisman, line 444: is there a reason why we would not have a selected clinical track?  seems odd given the name of the method.  mabye explain in a comment when this scenario would exist.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +399,168 @@ export default class Oncoprinter extends React.Component<
...
+        }
+        switch (this.selectedClinicalTrack.datatype) {
+            case 'counts':
+                return MUTATION_SPECTRUM_FILLS[
+                    _.indexOf(MUTATION_SPECTRUM_CATEGORIES, value)
+                ];
+            case 'string':
+                // Mixed refers to when an event has multiple values (i.e. Sample Type for a patient event may have both Primary and Recurrence values)
+                if (value === 'Mixed') {
+                    return [220, 57, 18, 1];
```
alisman, line 455: make this a constant so we understand what it is.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +399,168 @@ export default class Oncoprinter extends React.Component<
...
+                } else {
+                    return this
+                        .defaultClinicalAttributeColoringForStringDatatype[
+                        value
```
alisman, line 459: can you be sure that value will be present?

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
```diff
@@ -379,9 +399,168 @@ export default class Oncoprinter extends React.Component<
...
+        if (this.selectedClinicalTrack) {
+            return getClinicalTrackValues(this.selectedClinicalTrack);
+        }
+        return [];
+    }
+
     public render() {
         return (
             <div className="posRelative">
+                {this.selectedClinicalTrack && (
```
alisman, line 477: seems not good that we are duplicating this code (presumably it also exists in results view codebase?

src/pages/staticPages/tools/oncoprinter/OncoprinterClinicalAndHeatmapUtils.ts
```diff
@@ -503,14 +521,28 @@ export function getClinicalTracks(
                     attr.countsCategories!.length ===
                         MUTATION_SPECTRUM_FILLS.length
                 ) {
-                    countsCategoryFills = MUTATION_SPECTRUM_FILLS;
+                    countsCategoryFills = MUTATION_SPECTRUM_FILLS.slice();
+                    _.forEach(attr.countsCategories, (label, i) => {
```
alisman, line 525: i think this code would be clearer if you used map and conditional inside of the map wherein you return either the default (from MUTATION_SPECTRUM_FILLS) or userSelectedClinicalTracksColors entry.    That brings the fundamental operation to the top (building countsCategoryFills)

src/pages/staticPages/tools/oncoprinter/OncoprinterStore.ts
```diff
@@ -70,9 +73,24 @@ export default class OncoprinterStore {
     @observable hideGermlineMutations = false;
     @observable customDriverWarningHidden: boolean;
 
+    @observable _userSelectedClinicalTracksColors: {
+        [label: string]: {
```
alisman, line 77: trackLabel and attributeValue will make this clearer

src/pages/staticPages/tools/oncoprinter/OncoprinterToolUtils.ts
```diff
@@ -21,3 +26,21 @@ export function getDataForSubmission(
...
     });
 }
+
+export function getDefaultClinicalAttributeColoringForStringDatatype(
+    data: ClinicalTrackDatum[]
+) {
+    const colorGetter = makeUniqueColorGetter(
+        _.values(RESERVED_CLINICAL_VALUE_COLORS)
+    );
+    let categoryToColor = _.cloneDeep(RESERVED_CLINICAL_VALUE_COLORS);
```
alisman, line 36: i think you can avoid deep cloning here by instead merging RESERVED_CLINICAL_VALUE_COLORS onto a map made from data.  This will be clearer also.   That way RESERVED_CLINICAL_VALUE_COLORS will always win.

src/pages/staticPages/tools/oncoprinter/Oncoprinter.tsx
gblaih: it should always be present since this value comes from the track data which we used to obtain all the default colors.

alisman (approved):

PR #4822, opened 2024-01-05
Remove StudyID Requirement for Single Study in Custom Selection.

deveshidwivedi:
Fix https://github.com/cBioPortal/cbioportal/issues/10445

Changes:
Introduced a ternary operator to decide whether to include the ```studyId:``` prefix based on the ```isSingleStudy``` property. If it is a single study, only ```s.sampleId``` is used; otherwise, the original format with ```studyId:``` prefix is maintained.

### DIFF
--- a/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
+++ b/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
@@ -81,7 +81,7 @@ export default class CustomCaseSelection extends React.Component<
     get sampleSet(): { [id: string]: Sample } {
         return _.keyBy(
             this.props.selectedSamples,
-            s => `${s.studyId}:${s.sampleId}`
+            s => (this.isSingleStudy ? s.sampleId : `${s.studyId}:${s.sampleId}`)
         );
     }
 


deveshidwivedi:
Hi @alisman, could you please review this?

alisman:
@deveshidwivedi does the selection actually work without the study id prefix?  in other words when you click the select button?

deveshidwivedi:
I'm unable to open it in localhost since some time, please give me some time to check and make updates.

alisman:
@deveshidwivedi give us another day and the issue with localhost will be resolved.  we apologize.  it has to do with a big release we are doing (6.0)

deveshidwivedi:
hi @alisman, I'm still facing issues

alisman:
should now be fixed.  i'm checking this PR btw and will merge if things work

alisman:
@deveshidwivedi @ino based on my testing, this ALREADY works in custom selection.   It doesn't work in the custom data feature, but that isn't addressed by this PR:

image

alisman:
Hi Davesh, 
I'm happy to do a zoom with you and try to resolve the problem. It would actually be good to do so on behalf of other contributors.

deveshidwivedi:
Thank you very much! Since i had a problem with the repo on my machine (which is windows), i tried forking it again and wasn't able to do it. Went through tutorials on the internet and finally got it on my device. I'll try to make updates again and get back here! This time it should work..

PR #4824, opened 2024-01-10
Fix path to config_service for dev and dev builds

alisman:


src/config/config.ts
```diff
@@ -300,7 +300,7 @@ export function initializeLoadConfiguration() {
 
     const configServiceUrl =
         getBrowserWindow().frontendConfig.configurationServiceUrl ||
-        `${APIROOT}index`;
```
alisman, line 303: @inodb @haynescd this looks like maybe a typo right?  it DID need to be changed but not sure why it was changed to "index".

PR #4833, opened 2024-01-22
Updated Surivival Chart  for large cohort label overlap (number at risk)

TJMKuijpers:
This PR updates the number at risk table behavior in the survival chart for large cohorts (>10.000 samples). The current number-at-risk table plots the number at risk for every time point (under the x-axis, see Figure 1). However, when groups contain many samples, the labels overlap and become unreadable.

We have added a function that calculates whether the labels will overlap. If labels overlap, the number at risk table is updated to remove the overlapping labels (Figure 2). If no labels overlap, we default to the default behavior of the number of risk tables. 
Figure 1: old number at risk table for large cohorts when labels overlap
old number at risk table
Figure 2: new number at risk table for large cohorts when labels overlap
New number at risk table


src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -320,6 +329,19 @@ export default class SurvivalChartExtended
...
+    // determine whether visibility is controlled
+    // by a store component or by local state
+    @computed get visibilityState() {
+        return this.visible;
+    }
+    @action setVisibilityState(visible: boolean | undefined) {
+        if (visible !== undefined) {
+            this.visible = !!visible;
+        }
+    }
```
onursumer, line 343: I don't see a reference to any of these variables (`visible`, `visibilityState`, `setVisibilityState`). Do we actually use them somewhere?

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -847,45 +872,141 @@ export default class SurvivalChartExtended
...
+            return scaleLinear()
+                .domain([0, this.sliderValue])
+                .ticks(18);
+        }
+    }
 
-        const numberAtRisk = _.groupBy(
-            this.calculateGroupSize(timePoints.map(item => item)),
+    @action.bound
+    calculateLabelWidth(text: number, fontFamily: string, fontSize: number) {
```
onursumer, line 889: We should not mark this function as an `action`, because it does not modify any observable values. Also, we can probably use the `getTextWidth` function instead of writing a new one.
https://github.com/cBioPortal/cbioportal-frontend/blob/0bcf4809624f0dc7bd5d9169c2bef3172281b57c/packages/cbioportal-frontend-commons/src/lib/TextTruncationUtils.ts#L17

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -847,45 +872,141 @@ export default class SurvivalChartExtended
...
-            })
-        );
+            return !(
+                labelX + labelWidth < existingLabelX ||
+                labelX > existingLabelX + existingLabelWidth
+            );
+        };
 
-        return valueAtAxis;
+        const addLabelsForTimePoint = (
```
onursumer, line 931: It would be nice if we can add a comment somewhere in this function to describe how we decide which labels to hide.

Also, we had a similar issue with the lollipop plot labels. I was wondering if we could generalize that logic and use it here as well. It would probably require some refactoring though. So maybe okay to have two separate implementations for now.

https://github.com/cBioPortal/cbioportal-frontend/blob/3e47dd12cb0c1ba13bcff802414e621afb4f2dd9/packages/react-mutation-mapper/src/component/lollipopPlot/LollipopPlotNoTooltip.tsx#L293

onursumer (changes requested):
Thanks for fixing this!

(Requesting changes for some minor issues)

src/pages/resultsView/survival/SurvivalChart.tsx
TJMKuijpers: I moved `calculateLabelWidth` to the `SurvivalUtils ` so it becomes a function.. I checked and there is a difference in the width if we calculate it for a label or just the plain text. I believe the `getTextWidth `function would still allow for overlap in the survival chart.

src/pages/resultsView/survival/SurvivalChart.tsx
TJMKuijpers: Added the explanation on when labels are shown/hidden. Will have to compare to logic of `caculateTicks`

src/pages/resultsView/survival/SurvivalChart.tsx
TJMKuijpers: Removed this part since it became obsolete.

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -4,6 +4,7 @@ import _ from 'lodash';
 import { ClinicalAttribute } from 'cbioportal-ts-api-client';
 import { Dictionary } from 'lodash';
 import { calculateLogConfidenceIntervals } from './SurvivalHelper';
+import { getTextWidth } from 'cbioportal-frontend-commons';
```
onursumer, line 7: this is unused, let's remove

src/pages/resultsView/survival/SurvivalChart.tsx
```diff
@@ -320,6 +330,19 @@ export default class SurvivalChartExtended
...
+    // determine whether visibility is controlled
+    // by a store component or by local state
+    @computed get visibilityState() {
+        return this.visible;
+    }
+    @action setVisibilityState(visible: boolean | undefined) {
+        if (visible !== undefined) {
+            this.visible = !!visible;
+        }
+    }*/
```
onursumer, line 344: I guess it is okay to remove these instead of commenting out

src/pages/resultsView/survival/SurvivalUtil.tsx
```diff
@@ -738,3 +739,21 @@ export function calculateNumberOfPatients(
...
+    const tempElement = document.createElement('div');
+    tempElement.style.position = 'absolute';
+    tempElement.style.opacity = '0';
+    tempElement.style.fontFamily = fontFamily;
+    tempElement.style.fontSize = fontSize.toString();
+    tempElement.textContent = text.toString();
+    document.body.appendChild(tempElement);
+    const labelWidth = tempElement.offsetWidth;
+    document.body.removeChild(tempElement);
+    console.log(labelWidth);
```
onursumer, line 757: lets' remove the `console.log` as well

onursumer (changes requested):
Almost there! Added a couple more cleanup requests.

onursumer (approved):

PR #4839, opened 2024-02-01
#10191 Feat: Auto ordering of groups on comparision tab done

amanbhardwaj12072003:
Fix cBioPortal/cbioportal [#10191
](https://github.com/cBioPortal/cbioportal/issues/10191)
1. Auto-ordering of groups on the comparison tab.
2. Primary placed before Metastasis on the very first render of the component.

### DIFF
--- a/src/pages/groupComparison/groupSelector/GroupSelector.tsx
+++ b/src/pages/groupComparison/groupSelector/GroupSelector.tsx
@@ -206,6 +206,15 @@ export default class GroupSelector extends React.Component<
                         index={buttons.length}
                     />
                 );
+                // Logic To Put Primary First
+                let buttons_: any[] = [];
+                buttons_[0] = buttons[1];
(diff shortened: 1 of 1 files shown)


src/pages/groupComparison/groupSelector/GroupSelector.tsx
```diff
@@ -206,6 +206,15 @@ export default class GroupSelector extends React.Component<
                         index={buttons.length}
                     />
                 );
+                // Logic To Put Primary First
```
alisman, line 209: i think we need to extract the logic here (163 onward) and put it in a testable function.  also we shouldn't be "editing" the props of a group.  we need to use functional approach and produce a new ordered data (or do ordering upstream)

amanbhardwaj12072003: True that, allow me some time to work on this. Thanks!

PR #4846, opened 2024-02-12
Upgrade webdriver.io to v7 and begin migration of tests from sync to asyn

alisman:
We need to eventually upgrade to Webdriver.io 8, which will require we migrate our test style from sync to async per this documentation:

https://webdriver.io/docs/async-migration


end-to-end-test/shared/image-compare/js/main.js
```diff
@@ -117,19 +117,58 @@ function renderList(data) {
...
 
-    runMode = reportData.testHome || 'remote';
+    runMode =
+        reportData.length && reportData[0]?.specs[0]?.includes('/remote/')
+            ? 'remote'
+            : 'local';
+
+    console.log('reportData', reportData);
+
+    var tests = _(reportData)
```
onursumer, line 154: we don't seem to use this `tests` variable

end-to-end-test/shared/image-compare/js/main.js
```diff
@@ -117,19 +117,58 @@ function renderList(data) {
...
+        .flatMap(r => r.suites)
+        .flatMap(s => s.tests)
+        .value();
 
-    const filteredReportData = reportData.tests.filter(test => {
+    const de = deDupTests(reportData);
```
onursumer, line 159: should this be `tests` instead?

end-to-end-test/shared/specUtils_Async.js
```diff
@@ -0,0 +1,830 @@
+const clipboardy = require('clipboardy');
```
onursumer, line 1: this is not entirely new, right? mostly copy of the existing `specUtils.js`?

onursumer (approved):
LGTM! Just added a few questions/comments about minor things.

PR #4847, opened 2024-02-13
🔨Refactor code: retrieve ndex tooltip information dynamically

SAHU-01:
Fix cBioPortal/cbioportal#10595 (see https://help.github.com/en/articles/closing-issues-using-keywords)

- Using this [endpoint](https://iquery-cbio-dev.ucsd.edu/integratedsearch/v1/source) to populate the **Ndex Cancer Pathways** tooltip rather than hardcoding it as it is now
- Not finding the exact match for hardcoded URLs in endpoint, currently all URLs point to "#" (can be adjusted after discussion)

## Checks

## Any screenshots or GIFs?

Before:
image
 
 After:
image

## Notify reviewers
@inodb 

### DIFF
--- a/src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
+++ b/src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
@@ -35,39 +35,78 @@ function makeRemoteData() {
     return remoteData<DatabaseItem[]>({
         await: () => [],
         invoke: async () => {
-            try {
-                const getResponse = await request
-                    .get(
-                        'https://iquery-cbio-dev.ucsd.edu/integratedsearch/v1/source'
-                    )
-                    .query({
-                        param1: 'value1',
-                        param2: 'value2',
-                        param3: 'value3',
-                    })
-                    .set('Accept', 'application/json')
-                    .timeout(60000)
-                    .redirects(0);
-                const jsonData: ApiResponse = getResponse.body;
-                const extractedData: DatabaseItem[] = jsonData.results[0].databases.map(
-                    result => ({
-                        name: result.name,
-                        numberOfNetworks: parseInt(
-                            result.numberOfNetworks,
-                            10
-                        ).toString(),
-                        url: result.url,
-                    })
-                );
-                return extractedData;
-            } catch (error) {
-                console.error('Error fetching tooltip content:', error);
-                return [];
-            }
+            const getResponse = await request
+                .get(
+                    'https://iquery-cbio-dev.ucsd.edu/integratedsearch/v1/source'
+                )
+                .query({
+                    param1: 'value1',
+                    param2: 'value2',
+                    param3: 'value3',
+                })
+                .set('Accept', 'application/json')
+                .timeout(60000)
+                .redirects(0);
+
+            const jsonData: ApiResponse = getResponse.body;
+            const extractedData: DatabaseItem[] = jsonData.results[0]?.databases.map(
+                result => ({
+                    name: result.name,
+                    numberOfNetworks: parseInt(
+                        result.numberOfNetworks,
+                        10
+                    ).toString(),
+                    url: result.url,
+                })
+            );
+            return extractedData;
         },
         default: [],
     });
 }
+
+const TooltipContent: React.FC = observer(() => {
+    const store = useLocalObservable(() => ({
+        tooltipContent: null as DatabaseItem[] | null,
+        isFetchingTooltip: false,
+        tooltipData: makeRemoteData(),
+    }));
+
+    return (
+        <div style={{ maxWidth: 400 }}>
+            <a href="https://www.ndexbio.org/" target="_blank">
+                NDEx
+            </a>{' '}
+            shows 1,352 pathways:
+            {store.tooltipData.isPending ? (
+                <div>Loading tooltip content...</div>
+            ) : (
+                <div>
+                    {store.tooltipData.result && (
+                        <ul>
+                            {store.tooltipData.result.map(
+                                (item: DatabaseItem, index: number) => (
+                                    <li key={index}>
+                                        {item.numberOfNetworks} from{' '}
+                                        <a
+                                            href={`https://www.${item.url}`}
+                                            target="_blank"
+                                        >
+                                            {item.name}
+                                        </a>
+                                    </li>
+                                )
+                            )}
+                        </ul>
+                    )}
+                    {store.tooltipData.isError && (
+                        <div>Error: Failed to load tooltip content.</div>
+                    )}
+                </div>
+            )}
+        </div>
+    );
+});
 
 const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerProps> = observer(
     function({
@@ -79,19 +118,7 @@ const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerPro
             activeTab:
                 urlWrapper.query.pathways_source ||
                 ResultsViewPathwaysSubTab.PATHWAY_MAPPER,
-            tooltipContent: null as DatabaseItem[] | null,
-            isFetchingTooltip: false,
-            tooltipData: makeRemoteData(),
         }));
-
-        autorun(() => {
-            if (store.tooltipData.isComplete) {
-                runInAction(() => {
-                    store.tooltipContent = store.tooltipData.result;
-                    store.isFetchingTooltip = false;
-                });
-            }
-        });
 
         return (
             <>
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -27,8 +40,51 @@ const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerPro
             activeTab:
                 urlWrapper.query.pathways_source ||
                 ResultsViewPathwaysSubTab.PATHWAY_MAPPER,
+            tooltipContent: null as DatabaseItem[] | null,
+            isFetchingTooltip: false,
         }));
 
+        useEffect(() => {
```
alisman, line 47: it would be more consistent with codebase to use a mobxpromise here to manage the loading behavior.  Look elsewhere in codebase for remoteData( ... ).    This can be assigned to the store above.   e.g.

```
{
       someData:makeRemoteData()
}

```
where makeRemoteData is a function local to this file which returns remoteData(...)
```

**Review by alisman — CHANGES_REQUESTED** [review]
Thanks for making this PR!  Would be good to use our remoteData/mobxpromise solution for handling data fetching state

File: src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx [hunk 2]
```diff
@@ -9,14 +9,66 @@ import { ResultsViewPageStore } from 'pages/resultsView/ResultsViewPageStore';
...
+}
+interface ApiResponse {
+    results: ResultItem[];
+}
+
+function makeRemoteData() {
+    return remoteData<DatabaseItem[]>({
+        await: () => [],
+        invoke: async () => {
+            try {
```
alisman, line 38: @SAHU-01 thanks very much for making this change.  a couple more things. it would be best not to catch the error here.  what will then happen is the the promise will will have isError=true.  You can then use that in the tooltip UI to show an error message to the user, inside the tooltip, which will explain why there are no results.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -27,8 +79,20 @@ const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerPro
             activeTab:
                 urlWrapper.query.pathways_source ||
                 ResultsViewPathwaysSubTab.PATHWAY_MAPPER,
+            tooltipContent: null as DatabaseItem[] | null,
+            isFetchingTooltip: false,
+            tooltipData: makeRemoteData(),
         }));
 
+        autorun(() => {
```
alisman, line 87: you don't need this stuff here, you can just reference tooltipData.result, toooltipData.isComplete, tooltipData.isPending, tooltipData.isError in your UI.  Because these properties are mobx observables and the component is wrapped with observer(), the render will automatically be fired.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -152,6 +216,63 @@ const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerPro
                             height={800}
                             url={`${resultsViewPageStore.remoteNdexUrl.result}`}
                         />
+                    </MSKTab>*/}
+                    <MSKTab
+                        key={ResultsViewPathwaysSubTab.NDEX}
+                        id={ResultsViewPathwaysSubTab.NDEX}
+                        linkText={ResultsViewPathwaysSubTab.NDEX}
+                        linkTooltip={
+                            <div style={{ maxWidth: 400 }}>
```
alisman, line 225: you should wrap this whole tooltip thing inside of a component and put the store and remoteData fetching inside of it.  that way it's lifecycle is independent of the tab, which shouldn't care or know about the tooltip (except to render it).    
```
linkTooltip=<TooltipComponent ...props-if-needed />
```

alisman:
@SAHU-01 we really appreciate you doing this work for us.  I will be out of the office next week.  If you have any questions about my change requests, please ask @onursumer.  Sorry I didn't catch the second round of requests earlier.  FYI @inodb

SAHU-01:
@alisman I've made the changes requested.
Please review the changes and let me know if any further adjustments are needed.
-- Thanks!

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -9,14 +9,105 @@ import { ResultsViewPageStore } from 'pages/resultsView/ResultsViewPageStore';
 import ResultsViewURLWrapper from 'pages/resultsView/ResultsViewURLWrapper';
 import { AppStore } from 'AppStore';
 import { useLocalObservable } from 'mobx-react-lite';
-import { runInAction } from 'mobx';
+import { runInAction, autorun } from 'mobx';
```
onursumer, line 12: looks like we don't use these imports anywhere. let's remove them.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -9,14 +9,105 @@ import { ResultsViewPageStore } from 'pages/resultsView/ResultsViewPageStore';
...
+        invoke: async () => {
+            const getResponse = await request
+                .get(
+                    'https://iquery-cbio-dev.ucsd.edu/integratedsearch/v1/source'
+                )
+                .query({
+                    param1: 'value1',
+                    param2: 'value2',
+                    param3: 'value3',
+                })
```
onursumer, line 46: Do we really need these query parameters? It doesn't make much sense to me.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -9,14 +9,105 @@ import { ResultsViewPageStore } from 'pages/resultsView/ResultsViewPageStore';
...
+            return extractedData;
+        },
+        default: [],
+    });
+}
+
+const TooltipContent: React.FC = observer(() => {
+    const store = useLocalObservable(() => ({
+        tooltipContent: null as DatabaseItem[] | null,
+        isFetchingTooltip: false,
```
onursumer, line 71: `isFetchingTooltip` is unused. Let's remove it as well.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -9,14 +9,105 @@ import { ResultsViewPageStore } from 'pages/resultsView/ResultsViewPageStore';
...
+            ) : (
+                <div>
+                    {store.tooltipData.result && (
+                        <ul>
+                            {store.tooltipData.result.map(
+                                (item: DatabaseItem, index: number) => (
+                                    <li key={index}>
+                                        {item.numberOfNetworks} from{' '}
+                                        <a
+                                            href={`https://www.${item.url}`}
```
onursumer, line 92: I think we can omit `www`
```suggestion
                                            href={`https://${item.url}`}
```

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -80,7 +171,7 @@ const PathWayMapperContainer: React.FunctionComponent<IPathwayMapperContainerPro
                             urlWrapper={resultsViewPageStore.urlWrapper}
                         />
                     </MSKTab>
-                    <MSKTab
+                    {/*<MSKTab
```
onursumer, line 174: I think it's okay to delete the old code. we can always retrieve it from the git history if we really need to.

onursumer (changes requested):
Looks good! Just a few more minor requests.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
SAHU-01: Defining `isFetchingTooltip` within the TooltipContent component's store ensures its accessibility and prevents errors when referenced internally. Removing it would disrupt the component's functionality as it's integral to its state management. So retaining isFetchingTooltip is necessary for the component's proper operation.

When you remove `isFetchingTooltip: false` the TooltipContent component attempts to access a property that doesn't exist in its store.

alisman: Unless we're mistaken, isFetchingTooltip is not actually used inside the component anymore now that we are using the mobxpromise.

SAHU-01: I apologize for the confusion and thank you for bringing that to my attention. You are correct that` isFetchingTooltip ` is not utilized within the component anymore, especially considering the use of mobxpromise. Thank you for your patience and clarification on this matter.

src/pages/resultsView/pathwayMapper/PathWayMapperContainer.tsx
```diff
@@ -10,13 +10,98 @@ import ResultsViewURLWrapper from 'pages/resultsView/ResultsViewURLWrapper';
...
+            ) : (
+                <div>
+                    {store.tooltipData.result && (
+                        <ul>
+                            {store.tooltipData.result.map(
+                                (item: DatabaseItem, index: number) => (
+                                    <li key={index}>
+                                        {item.numberOfNetworks} from{' '}
+                                        <a
+                                            href={`https://${item.url}`}
```
inodb, line 86: @coleslaw481 mentioned:

> You can just use the value of the url field. The value may lack a https:// prefix so that should be checked for and added if missing

It looks like in the code here we assume `item.url` never contains the protocol? Maybe better to be safe and strip out all protocol stuff from `item.url` (i.e. `http{s}://`) if it exists?

onursumer (approved):

PR #4848, opened 2024-02-15
Add configuration for disabling study tag fetching (default to true)

alisman:
There is a problem with Spring security (SAML mode) and the study tags endpoint.  We need to disable it on portals which don't use tags until the problem is resolved.


dippindots (approved):
LGTM!

PR #4852, opened 2024-02-19
Remove entity stable ID from generic assay name

MatthijsPon:
Follow up to https://github.com/cBioPortal/cbioportal-core/pull/7

As the ENTITY_STABLE_IDs for Generic Assays are now more strictly validated, we propose to simply hide the stable ID if the `NAME` property is given, instead of showing `NAME (ENTITY_STABLE_ID)`.

- a
- b

## Checks
- [] Has tests or has a separate issue that describes the types of test that should be created. If no test is included it should explicitly be mentioned in the PR why there is no test.

## Any screenshots or GIFs?

## Notify reviewers


inodb:
Could use a "use case" description to better understand why this is necessary

dippindots (changes requested):
@MatthijsPon This pr looks great, I can see it helps in some cases when name can clearly explain one entity.
However in some case like `Methylation` profile, our `NAME` and `Stable Id` are both meaningful and we should keep both of them. In this case, we can't just hide the `Stable Id`
image

alisman:
closing because this solution is not viable in all cases, per @dippindots note above

PR #4867, opened 2024-03-05
Study view custom sample selection with multi study

7xuanlu:
Fix https://github.com/cBioPortal/cbioportal/issues/10603

Currently while selecting samples/patients using custom selection, we need to enter both study ID and sample ID for multi-study queries. Instead we would want to enter only sample ID or patient ID. Add sample from all studies if it belongs to multiple studies.

- enable the filter button even though there are multiple studies
- filter sample from all studies without specifying `study id`
- filter samples by mixing `sample id` with `study id:sample id`

Screenshot 2024-03-05 at 2 54 13 PM


sheridancbio:
I think the language in the warning message in the GUI is too simple. There could be two studies which both include the same sample ... or there could be two studies which include two different samples which happen to have exactly the same sample id. This is especially true with published studies which represent their samples with very simple sample identifiers ... such as "s1". The s1 sample in published_study_id_1 is very likely to be a different sample from the s1 in published_study_id2. So at the very least I would suggest language which indicates that the sample *identifier* "sample_id_1" is present in multiple studies -  rather than saying that the sample is present in multiple studies. In theory we could probe these cases to get an informed idea of whether the sample called "sample_id_1" in study_id_1 refers to something which has identical genetic event characterization of the sample called "sample_id_1" in study_id2. Perhaps if that is true, we can have some confidence to tell the user that this is "likely the same sample".

jjgao:
> I think the language in the warning message in the GUI is too simple. There could be two studies which both include the same sample ... or there could be two studies which include two different samples which happen to have exactly the same sample id. This is especially true with published studies which represent their samples with very simple sample identifiers ... such as "s1". The s1 sample in published_study_id_1 is very likely to be a different sample from the s1 in published_study_id2. So at the very least I would suggest language which indicates that the sample _identifier_ "sample_id_1" is present in multiple studies - rather than saying that the sample is present in multiple studies. 

Thanks, @sheridancbio. 

@7xuanlu  let's improve the text:

"The SAMPLE_ID's below belong to multiple studies. All of them in the different studies will be selected. To select specific samples, pelase specify STUDY_ID, e.g. STUDY_ID:SAMPLE_ID.

[SAMPLE_ID LIST]"

image

> In theory we could probe these cases to get an informed idea of whether the sample called "sample_id_1" in study_id_1 refers to something which has identical genetic event characterization of the sample called "sample_id_1" in study_id2. Perhaps if that is true, we can have some confidence to tell the user that this is "likely the same sample".

@sheridancbio this would be a research project :) I think we'll leave it out for this PR. This PR's purpose was to simplify the input when people are select non-overlapping studies, e.g. TCGA pancan studies. It is hurdle that we have to specify study id even when all sample IDs are unique.

PR #4869, opened 2024-03-13
:wrench: Update keycloak image to quay.io

haynescd:
For some reason jboss/keycloak:11.0.3 does not exists anymore.... updated to quay.io


inodb (approved):
LGTM - thanks!

PR #4883, opened 2024-03-29
Show all survival attributes and remove duplicates in study clinical data tab

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/10696

Continue to hide survival attributes (besides `OS_STATUS`) in Summary tab.

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -674,7 +674,7 @@ export class StudyViewPageStore
         this.reactionDisposers.push(
             reaction(
                 () => [
-                    this.visibleAttributes,
+                    this.visibleAttributesForSummary,
                     this.columns,
                     _.fromPairs(this.chartsDimension.toJSON()),
                     _.fromPairs(this.chartsType.toJSON()),
@@ -2436,7 +2436,7 @@ export class StudyViewPageStore
     @action.bound
     private updateLayout(): void {
         this.currentGridLayout = calculateLayout(
-            this.visibleAttributes,
+            this.visibleAttributesForSummary,
             this.columns,
             _.fromPairs(this.chartsDimension.toJSON()),
             this.currentGridLayout,
@@ -6837,6 +6837,7 @@ export class StudyViewPageStore
         if (this.isSavingUserPreferencePossible) {
             chartSettingsMap = getChartSettingsMap(
                 this.visibleAttributes,
+                this.visibleAttributesForSummary,
                 this.columns,
                 _.fromPairs(this.chartsDimension.toJSON()),
                 _.fromPairs(this.chartsType.toJSON()),
@@ -7043,8 +7044,19 @@ export class StudyViewPageStore
             },
             [] as ChartMeta[]
         );
+        const visibleAttributesForSummary = _.reduce(
+            this._defaultVisibleChartIds,
+            (acc, chartUniqueKey) => {
+                if (this.chartMetaSetForSummary[chartUniqueKey]) {
+                    acc.push(this.chartMetaSetForSummary[chartUniqueKey]);
+                }
+                return acc;
+            },
+            [] as ChartMeta[]
+        );
         return getChartSettingsMap(
             visibleAttributes,
+            visibleAttributesForSummary,
             this.columns,
             _.fromPairs(this._defaultChartsDimension.toJSON()),
             _.fromPairs(this._defaultChartsType.toJSON()),
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -182,7 +182,7 @@ import {
     getCustomChartDownloadData,
     generateColorMapKey,
     MutationCategorization,
-    getChartMetaSetForClinicalAttributes,
+    getChartMetaSet,
     getVisibleAttributes,
 } from './StudyViewUtils';
 import { SingleGeneQuery } from 'shared/lib/oql/oql-parser';
@@ -6583,7 +6583,7 @@ export class StudyViewPageStore
     @computed get chartMetaSetForSummary(): {
         [id: string]: ChartMeta;
     } {
-        let chartMetaSet = getChartMetaSetForClinicalAttributes(
+        let chartMetaSet = getChartMetaSet(
             this._customCharts,
             this.molecularProfiles.result,
             this._geneSpecificCharts,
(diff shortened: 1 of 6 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6405,6 +6405,65 @@ export class StudyViewPageStore
         );
     }
 
+    @computed get chartMetaSetForClinicalAttributes(): {
```
alisman, line 6408: lets put a comment explaining that this derives the list of columns for the clinical attribute tab from the currently visible charts

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6405,6 +6405,65 @@ export class StudyViewPageStore
...
+        [id: string]: ChartMeta;
+    } {
+        let _chartMetaSet = _.fromPairs(this._customCharts.toJSON());
```
alisman, line 6411: i don't think we need the leading underscore here

src/pages/studyView/StudyViewPageStore.ts
alisman: also can we move all this logic into a helper function to avoid bloating this file any more

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6405,6 +6405,65 @@ export class StudyViewPageStore
...
+        if (_.isEmpty(this.molecularProfiles.result)) {
```
alisman, line 6412: i'm guessing this is some legacy thing but can we put a comment explaining why we delete this?

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6405,6 +6405,65 @@ export class StudyViewPageStore
...
+            delete _chartMetaSet[
+                SpecialChartsUniqueKeyEnum.GENOMIC_PROFILES_SAMPLE_COUNT
+            ];
+        }
+        // Add meta information for clinical attributes (includes survival attributes)
+        // Convert to a Set for easy access and to update attribute meta information (would be useful while adding new features)
+        _.reduce(
```
alisman, line 6419: try to always keep to a functional paradigm where _.someOperation only ever assigns to a new variable.  we then merge those at the end.    i.e.:

```
const clinicalWhatevers = _.reduce(...);
const survivalWhatevers = _.reduce(...);

return { ...clinicalWhatevers, ...surivalWhatevers };

```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6668,6 +6727,41 @@ export class StudyViewPageStore
...
 
+    @computed
+    get visibleAttributesForClinicalDataTab(): ChartMeta[] {
+        // keep track of clinical attributes in case of duplicates
+        let clinicalAttributes: {
+            [id: string]: ChartMeta;
+        } = {};
+        return _.reduce(
+            Array.from(this._chartVisibility.entries() || []),
+            (acc, [chartUniqueKey, visible]) => {
```
alisman, line 6738: put the type in the callback definition instead of down below where you pass the argument.  This will prevent you from using 'as' casting which loosens typing. 
```
_.reduce(acc:ChartMeta[], ...) =>
```

src/pages/studyView/StudyViewPageStore.ts
alisman: ditto for the other computed.

src/pages/studyView/StudyViewPageStore.ts
alisman: another way to think about this is in terms of immutability.  We want to always err on the side of operating on/returning a NEW collection of things rather than mutating something in place.   You are effectively doing that here by mutating ```let _chartMetaSet = _.fromPairs(this._customCharts.toJSON());```.    It so happens that this is effectively a new thing because we're calling toJSON and fromPairs etc.  But it would be more clear if we were doing the following:
```
const customCarts = ...
const clinicalWhatevers = _.reduce()
const suruvivaWhaters = _.reduce()

return { ... }; /// this is the immutable here.  It's a NEW object and we run no risk of mistakenly updating state.  

```

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6569,250 +6571,85 @@ export class StudyViewPageStore
...
-                        clinicalAttribute: attribute,
-                    };
-                }
-                return acc;
-            },
-            _chartMetaSet
-        );
-
-        _.reduce(
+        let chartMetaSet = getChartMetaSetForClinicalAttributes(
```
alisman, line 6632: lets put a comment explaining why the general (chartMetaSet) starts with the clinical attributes set.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2895,9 +2900,19 @@ export function getChartSettingsMap(
         }
         chartSettingsMap[id] = chartSetting;
     });
+    // attributes disabled on the summary tab (used in Clinical Data tab but not Summary tab like most survival attributes)
```
alisman, line 2903: good comment!

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -2895,9 +2900,19 @@ export function getChartSettingsMap(
...
+        visibleAttributes,
+        visibleAttributesForSummary,
+        _.isEqual
+    );
     // add layout for each chart
     gridLayout.forEach(layout => {
-        if (layout.i && chartSettingsMap[layout.i]) {
+        if (
+            layout.i &&
+            chartSettingsMap[layout.i] &&
```
alisman, line 2913: would it be possible to avoid putting these disabled attributes into chartSettingsMap in the first place so we wouldn't have to check this here?

src/pages/studyView/StudyViewPageStore.ts
gblaih: the general chartMetaSet doesn't start with the clinical attributes set; it just contains the meta information for clinical attributes. the function name is misleading, changed to `getChartMetaSet(...)`.

src/pages/studyView/StudyViewUtils.tsx
gblaih: we still want to keep the disabled attributes in the map for user settings purposes for `Clinical Data` tab columns, just without the layout or else it will mess with chart layout on summary tab.

PR #4893, opened 2024-04-16
on plots tab add sorting for barcharts

SURAJ-SHARMA27:
Fixes #[10722](https://github.com/cBioPortal/cbioportal/issues/10722)

I have created a dropdown field named 'Sort' which currently contains two options: 'Sort by Alphabet' and 'Sort by Count'. By default, 'Sort by Alphabet' is selected. I implemented the logic for 'Sort by Count' as follows:

i) I passed the 'sortOption' prop throughout the plot components.
ii) In 'MultipleCategoryBarPlotUtils.ts', I added a case where if the 'sortOption' is 'sortByCount', it creates an array named 'totalSumArray'. This array stores fields defined by the interface:

```
interface TotalSumItem {
    majorCategory: string;
    sum: number;
    minorCategory: { name: string; count: number; percentage: number.. }[];
}

```
Then, I sorted the above array with respect to 'sum', and finally returned the individually sorted array. I implemented the same for sorted labels.

## Checks

## Any screenshots or GIFs?

[Screencast from 2024-04-16 15-48-09.webm](https://github.com/cBioPortal/cbioportal-frontend/assets/90304648/0f3d9e6b-7b38-41f1-ae6d-b902b493df08)

### DIFF
--- a/src/pages/groupComparison/ClinicalData.tsx
+++ b/src/pages/groupComparison/ClinicalData.tsx
@@ -937,6 +937,7 @@ export default class ClinicalData extends React.Component<
                                     ? this.highlightedRow.qValue
                                     : null
                             }
+                            sortOption={'sortByAlphabet'}
                         />
                     );
                 }
--- a/src/pages/groupComparison/ClinicalData.tsx
+++ b/src/pages/groupComparison/ClinicalData.tsx
@@ -937,7 +937,6 @@ export default class ClinicalData extends React.Component<
                                     ? this.highlightedRow.qValue
                                     : null
                             }
-                            sortOption={'sortByAlphabet'}
                         />
                     );
                 }
--- a/src/pages/groupComparison/MultipleCategoryBarPlot.tsx
+++ b/src/pages/groupComparison/MultipleCategoryBarPlot.tsx
@@ -58,7 +58,16 @@ export interface IMultipleCategoryBarPlotProps {
     svgRef?: (svgContainer: SVGElement | null) => void;
     pValue: number | null;
     qValue: number | null;
-    sortOption: string;
+    sortOption?: string;
+}
+export interface TotalSumItem {
+    majorCategory: string;
+    sum: number;
+    minorCategory: {
+        name: string;
+        count: number;
+        percentage: number;
+    }[];
 }
 
 export interface IMultipleCategoryBarPlotData {
@@ -235,7 +244,6 @@ export default class MultipleCategoryBarPlot extends React.Component<
 
     private get legend() {
         if (this.legendData.length > 0) {
-            console.log(this.props, 'This.propss');
             return (
                 <VictoryLegend
                     orientation={
@@ -427,25 +435,17 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     @computed get labels() {
-        console.log(this.data, 'thisisis');
-        interface TotalSumItem {
-            majorCategory: string;
-            sum: number;
-            minorCategory: {
-                name: string;
-                count: number;
-                percentage: number;
-            }[];
-        }
-        this.data.forEach(item => {
+        _.forEach(this.data, item => {
+            // Sorting counts within each item
             item.counts.sort((a, b) =>
                 a.majorCategory.localeCompare(b.majorCategory)
             );
         });
         const totalSumArray: TotalSumItem[] = [];
-        this.data.forEach(item => {
-            item.counts.forEach(countItem => {
-                const existingItem = totalSumArray.find(
+        _.forEach(this.data, item => {
+            _.forEach(item.counts, countItem => {
+                const existingItem = _.find(
+                    totalSumArray,
                     sumItem => sumItem.majorCategory === countItem.majorCategory
                 );
                 if (existingItem) {
@@ -471,24 +471,21 @@ export default class MultipleCategoryBarPlot extends React.Component<
             });
         });
 
-        console.log(totalSumArray, 'totalssufrommulti');
         totalSumArray.sort((a, b) => b.sum - a.sum);
 
-        const labelss = totalSumArray.map(item => item.majorCategory);
-        console.log(labelss, 'totalsrommulti');
+        const sortedLabels = totalSumArray.map(item => item.majorCategory);
 
         if (this.props.sortOption == 'sortByCount') {
-            return labelss;
+            return sortedLabels;
         }
 
         if (this.data.length > 0) {
-            const ans = sortDataByCategory(
+            const CategorizedData = sortDataByCategory(
                 this.data[0].counts.map(c => c.majorCategory),
                 x => x,
                 this.majorCategoryOrder
             );
-            console.log('thisisconsoledataans', ans);
-            return ans;
+            return CategorizedData;
         } else {
             return [];
         }
@@ -799,7 +796,7 @@ export default class MultipleCategoryBarPlot extends React.Component<
             !!this.props.horizontalBars,
             !!this.props.stacked,
             !!this.props.percentage,
-            this.props.sortOption
+            this.props.sortOption || 'sortByAlphabet'
         );
         return barSpecs.map(spec => (
             <VictoryBar
@@ -946,7 +943,6 @@ export default class MultipleCategoryBarPlot extends React.Component<
 
                                 {this.horzAxis}
                                 {this.vertAxis}
-                                {console.log(this.chartEtl, 'tjos')}
                                 {this.chartEtl}
                             </VictoryChart>
                         </g>
@@ -979,7 +975,6 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     render() {
-        console.log(this.data, 'this.datas');
         if (!this.data.length) {
             return <div className={'alert alert-info'}>No data to plot.</div>;
         }
--- a/src/pages/groupComparison/MultipleCategoryBarPlot.tsx
+++ b/src/pages/groupComparison/MultipleCategoryBarPlot.tsx
@@ -435,47 +435,46 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     @computed get labels() {
-        _.forEach(this.data, item => {
-            // Sorting counts within each item
-            item.counts.sort((a, b) =>
-                a.majorCategory.localeCompare(b.majorCategory)
-            );
-        });
-        const totalSumArray: TotalSumItem[] = [];
-        _.forEach(this.data, item => {
-            _.forEach(item.counts, countItem => {
-                const existingItem = _.find(
-                    totalSumArray,
-                    sumItem => sumItem.majorCategory === countItem.majorCategory
-                );
-                if (existingItem) {
-                    existingItem.sum += countItem.count;
-                    existingItem.minorCategory.push({
-                        name: item.minorCategory,
-                        count: countItem.count,
-                        percentage: countItem.percentage,
-                    });
-                } else {
-                    totalSumArray.push({
-                        majorCategory: countItem.majorCategory,
-                        sum: countItem.count,
-                        minorCategory: [
-                            {
-                                name: item.minorCategory,
-                                count: countItem.count,
-                                percentage: countItem.percentage,
-                            },
-                        ],
-                    });
-                }
-            });
-        });
-
-        totalSumArray.sort((a, b) => b.sum - a.sum);
-
-        const sortedLabels = totalSumArray.map(item => item.majorCategory);
-
         if (this.props.sortOption == 'sortByCount') {
+            _.forEach(this.data, item => {
+                item.counts.sort((a, b) =>
+                    a.majorCategory.localeCompare(b.majorCategory)
+                );
+            });
+            const totalSumArray: TotalSumItem[] = [];
+            _.forEach(this.data, item => {
+                _.forEach(item.counts, countItem => {
+                    const existingItem = _.find(
+                        totalSumArray,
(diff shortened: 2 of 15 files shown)


src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -58,6 +58,7 @@ export interface IMultipleCategoryBarPlotProps {
     svgRef?: (svgContainer: SVGElement | null) => void;
     pValue: number | null;
     qValue: number | null;
+    sortOption: string;
```
gblaih, line 61: we can add a question mark like `sortOption?: string;` to make this prop optional. then if this prop is not provided, we can assume it to be 
'sortByAlphabet'. that way we don't have to pass `sortOption={'sortByAlphabet'}` every time

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -234,6 +235,7 @@ export default class MultipleCategoryBarPlot extends React.Component<
 
     private get legend() {
         if (this.legendData.length > 0) {
+            console.log(this.props, 'This.propss');
```
gblaih, line 238: remove all the console logs

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +427,68 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     @computed get labels() {
+        console.log(this.data, 'thisisis');
+        interface TotalSumItem {
```
gblaih, line 431: define and export this interface outside this getter

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +427,68 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+            majorCategory: string;
+            sum: number;
+            minorCategory: {
+                name: string;
+                count: number;
+                percentage: number;
+            }[];
+        }
+        this.data.forEach(item => {
```
gblaih, line 440: for array functions such as `forEach`, lets try using the lodash version like _.forEach(this.data,` ...)` when we can throughout

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +427,68 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+        const labelss = totalSumArray.map(item => item.majorCategory);
+        console.log(labelss, 'totalsrommulti');
+
+        if (this.props.sortOption == 'sortByCount') {
+            return labelss;
+        }
+
         if (this.data.length > 0) {
-            return sortDataByCategory(
+            const ans = sortDataByCategory(
```
gblaih, line 485: lets come up with better names for `labelss` and `ans`. maybe `categories`?

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -107,7 +116,8 @@ export function makeBarSpecs(
     categoryCoord: (categoryIndex: number) => number,
     horizontalBars: boolean,
     stacked: boolean,
-    percentage: boolean
+    percentage: boolean,
+    sortOption: string
```
gblaih, line 120: we can also make this optional so we don't have to pass `sortByAlphabet` every time we use this. and if nothing is passed, assume sortOption is using `sortByAlphabet`

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -121,20 +131,114 @@ export function makeBarSpecs(
 }[] {
     // one bar spec per minor category, in correct order - either specified, or alphabetical
     data = sortDataByCategory(data, d => d.minorCategory, minorCategoryOrder);
+
     // reverse the order of stacked or horizontal bars
     if ((!horizontalBars && stacked) || (horizontalBars && !stacked)) {
         data = _.reverse(data);
     }
+    console.log(data, 'iscalled2');
+    data.forEach(item => {
```
gblaih, line 140: this whole logic looks similar to the one above https://github.com/cBioPortal/cbioportal-frontend/blob/a05006df4870e64bd5df1dec6a485d3da8ca237d/src/pages/groupComparison/MultipleCategoryBarPlot.tsx#L440

would be great if we could make a helper function with all this logic to minimize duplicate code.

src/shared/components/plots/PlotsTab.tsx
```diff
@@ -1488,7 +1495,16 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
             selectedCategories: [],
         });
     }
-
+    handleChangeSortOption(selectedOption: OptionType) {
+        if (selectedOption) {
+            console.log(selectedOption, 'thisisselectedoptions');
+            console.log(this.sortOption, 'thisSOrtts');
+            this.sortOption = selectedOption.value;
+            // this.setState({ sortOption: selectedOption.value });
```
gblaih, line 1503: we can remove these comments now

SURAJ-SHARMA27:
@gblaih I have made nearly all the changes you suggested. Could you please review them and merge the branch if everything looks good?

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +435,57 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+                    });
+                }
+            });
+        });
+
+        totalSumArray.sort((a, b) => b.sum - a.sum);
+
+        const sortedLabels = totalSumArray.map(item => item.majorCategory);
+
+        if (this.props.sortOption == 'sortByCount') {
```
gblaih, line 478: i think we can move all the logic above this to inside this if block as we only need to calculate sortedLabels when `sortOption` is `sortByCount`

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -79,7 +80,13 @@ export function makePlotData(
     );
     return data;
 }
+interface TotalSumItem {
+    majorCategory: string;
+    sum: number;
+    minorCategory: { name: string; count: number; percentage: number }[];
+}
 
+export { TotalSumItem };
```
gblaih, line 89: we can delete this and just import and use the one you made here
https://github.com/cBioPortal/cbioportal-frontend/blob/1c2a8f4bfa8983a34816e99d4bbb9beb59658d66/src/pages/groupComparison/MultipleCategoryBarPlot.tsx#L63

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -117,24 +125,98 @@ export function makeBarSpecs(
...
         const sortedCounts = sortDataByCategory(
             counts,
             d => d.majorCategory,
             majorCategoryOrder
         );
-        return {
+
+        const finals = minorCategoryArrays[minorCategory];
+
+        let categoriezedCounts;
```
gblaih, line 211: categorizedCounts

src/pages/groupComparison/ClinicalData.tsx
```diff
@@ -937,6 +937,7 @@ export default class ClinicalData extends React.Component<
                                     ? this.highlightedRow.qValue
                                     : null
                             }
+                            sortOption={'sortByAlphabet'}
```
gblaih, line 940: we can probably delete this now that we made `sortOption` optional and we've taken care of the case when `sortOption` is not `sortByCount`

gblaih:
@SURAJ-SHARMA27 added a couple minor comments to be addressed but looks good overall. thanks!

SURAJ-SHARMA27:
@gblaih , I have made all the changes suggested. Could you review it once more?

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +435,56 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     @computed get labels() {
+        if (this.props.sortOption == 'sortByCount') {
```
alisman, line 438: @SURAJ-SHARMA27  this is going to actually mutate the sorting of counts in ```this.data```.  i.e. sort it in place.  instead, we should use functional/immutable approach to return a sorted list of labels without mutating underlying data.

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +435,56 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+            _.forEach(this.data, item => {
+                item.counts.sort((a, b) =>
+                    a.majorCategory.localeCompare(b.majorCategory)
+                );
+            });
+            const totalSumArray: TotalSumItem[] = [];
```
alisman, line 444: this code is quite confusing.   it would be clearer if you used _.sortBy(collection, ()=>);   where the return of the callback is a _.sum of the item counts within.

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -424,12 +435,56 @@ export default class MultipleCategoryBarPlot extends React.Component<
     }
 
     @computed get labels() {
+        if (this.props.sortOption == 'sortByCount') {
+            _.forEach(this.data, item => {
```
alisman, line 439: i'm curious why you need to sort alphabetical first.  you should put comments explaining each step.

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -739,7 +794,8 @@ export default class MultipleCategoryBarPlot extends React.Component<
             this.categoryCoord,
             !!this.props.horizontalBars,
             !!this.props.stacked,
-            !!this.props.percentage
+            !!this.props.percentage,
+            this.props.sortOption || 'sortByAlphabet'
```
alisman, line 798: you should probably apply this default inside the makeBarSpec

src/shared/components/plots/PlotsTab.tsx
```diff
@@ -4448,6 +4464,32 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                             </label>
                         </div>
                     )}
+                    {this.isStacked && (this.plotType.result==3) && (
```
alisman, line 4467: can we make the 3 into an Enum so it's human understandable?

alisman (changes requested):
@SURAJ-SHARMA27 i think the code here can be improved.  the biggest issue is the sorting of what should be immutable data.   see comments.  one good rule of thumb is to avoid forEach.  The reason is that forEach forces you do do a sideeffect because it isn't "functional".  It doesn't simply return a new thing (e.g. a sorted array that is a computation of underlying array).  Always use more specific operators like map/reduce/sort etc.  Learing the full set of Lodash functions will help you.

alisman:
@SURAJ-SHARMA27 have you had a chance to look at these comments?

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
SURAJ-SHARMA27: The logic is that I will traverse this.data and will make one totalSumArray array which has a schema like:
```
{
    majorCategory: countItem.majorCategory,
    sum: countItem.count,
    minorCategory: [
        {
            name: item.minorCategory,
            count: countItem.count,
            percentage: countItem.percentage,
        },
    ],
}

```

Now, I will traverse this.data and go to each `majorCategory` to update the `totalSum` in which i will store the count of each object. Then, I will sort the `totalSumArray` with respect to `totalSum` and extract the information as per the requirements like label and `majorCategory`.

Let's take an example. For one configuration, this.data contains:
```
{
    0: { minorCategory: 'Male', counts: Array(42) },
    1: { minorCategory: 'Female', counts: Array(42) }
}
```
I will traverse each object, go to the counts array, and create one object for like Melanoma in `totalSumArray` in which I will store the information and sum of counts for each object. Then, I will sort the `totalSumArray` according to sum and retrieve the labels from there.

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
SURAJ-SHARMA27: I was sorting the elements alphabetically first so that all the elements in each object would come in the same order, allowing me to calculate the count corresponding to each element accurately. For example:

```
data = {
    0: { minorCategory: 'Male', counts: Array(42) },
    1: { minorCategory: 'Female', counts: Array(42) }
}
```


Let's say, in the array containing 42 elements in the Male minorCategory, the first index is X, and in the Female minorCategory, the first index is Y in counts. If I calculate sum_of_total_counts = data[0].counts[0].count + data[1].counts[1].count, the sum_of_total_counts calculated is incorrect because it adds the counts for X and Y. However, if I sort it alphabetically, the counts array for each minorCategory will be in the same order, and I can find the total count easily.

SURAJ-SHARMA27:
> @SURAJ-SHARMA27 have you had a chance to look at these comments?

Actually, my laptop's RAM was causing issues, which is why the repository was not running locally.

inodb:
This is superseded by: https://github.com/cBioPortal/cbioportal-frontend/pull/4960

PR #4895, opened 2024-04-26
Fix api sync

dippindots:
Updates for mutation type charts on study view


alisman:
@7xuanlu is fixing this (it's a server side issue)

PR #4899, opened 2024-05-01
Fix getting molecular profile ids from key

kalletlak:
Fix https://github.com/cBioPortal/cbioportal/issues/10775

### DIFF
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -883,6 +883,8 @@ function startsWithSvChartType(chartType?: string) {
 }
 
 export function getMolecularProfileIdsFromUniqueKey(uniqueKey: string) {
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -883,16 +883,10 @@ function startsWithSvChartType(chartType?: string) {
...
-    if (!startsWithSvChartType(parts[0])) {
-        return parts;
-    } else {
-        return _.map(
-            parts,
-            molecularProfileId =>
-                molecularProfileId.split(CHART_TYPE_SEPARATOR)[1]
-        );
-    }
+    return (startsWithSvChartType(uniqueKey)
```
alisman, line 886: can you put a comment with the format of the key and some explaination of why it sometimes but not always has chartype?

PR #4900, opened 2024-05-02
Customizable Dynamic Kaplan-Meier Plot for Survival Analysis

7xuanlu:
Address issue [Dynamic KM Plot](https://github.com/cBioPortal/cbioportal/issues/10375) and [RFC 36](https://docs.google.com/document/d/1WPY8ET9WlN9YAjDOZmHG_z2_fpEpKKmLArndok1f8bo/edit#heading=h.t10nr9te3qvu)
This frontend PR aims to have a customizable Dynamic Kaplan-Meier plot that allows users to specify start point and endpoint or censored point for survival analysis.

- Survival table and survival plot visibility
- Add start, end, censored clinical events dropdown
- Add start and end clinical attributes dropdown if available
- Save/load custom survival plots from session service
- Delete a survival plot from table
- Supports ANY event in censored event dropdown, which would choose the last event as the censored point for specific patient

### DIFF
--- a/src/pages/groupComparison/Survival.tsx
+++ b/src/pages/groupComparison/Survival.tsx
@@ -39,6 +39,7 @@ import {
 import { observable, action, makeObservable, computed } from 'mobx';
 import survivalPlotStyle from './styles.module.scss';
 import SurvivalPrefixTable, {
+    SurvivalChartType,
     SurvivalPrefixTableStore,
 } from 'pages/resultsView/survival/SurvivalPrefixTable';
 import { PatientSurvival } from 'shared/model/PatientSurvival';
@@ -91,7 +92,7 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
     private selectedEndClinicalEventAttributes: ClinicalEventDataWithKey[] = [];
 
     @observable
-    private _selectedCensoredClinicalEventType: string | undefined = undefined;
+    private _selectedCensoredClinicalEventType: string | undefined = 'any';
 
     @observable
     private selectedCensoredClinicalEventAttributes: ClinicalEventDataWithKey[] = [];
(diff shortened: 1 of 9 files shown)


package-lock.json
alisman, line 1: we shouldn't need package lock.  use yarn to add new things, not npm

src/pages/groupComparison/Survival.tsx
```diff
@@ -379,141 +320,262 @@ export default class Survival extends React.Component<ISurvivalProps, {}> {
...
+                    g => g.value
+                );
+                const patientToAnalysisGroups = this.analysisGroupsComputations
+                    .result!.patientToAnalysisGroups;
+                const pValues = this.pValuesByPrefix.result!;
+                const qValues = this.qValuesByPrefix.result!;
+
+                const survivalPrefixes = _.map(
+                    this.props.store.survivalTitleByPrefix
+                        .result! as Dictionary<string>,
```
alisman, line 433: shouldn't need cast if type properly

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -152,58 +172,278 @@ export default abstract class ComparisonStore extends AnalysisStore
...
-        this.tabHasBeenShownReactionDisposer &&
-            this.tabHasBeenShownReactionDisposer();
+    public destroy(): void {
+        for (const disposer of this.reactionDisposers) {
+            disposer();
+        }
+    }
+
+    @observable
+    public startEventPosition: 'FIRST' | 'LAST' = 'FIRST';
```
alisman, line 247: i think this needs to be someone scoped better.  i.e. it needs to be somehow associated with the survival plot stuff, not just hanging out here in top level of store.

jjgao:
When there is only one _status variable, the table will be hidden and there would be no way to add a dynamic KM plot. **Solution**: show the table when there is one _status variable.

When there is no _status variable (no survival data) but there are still timeline data, there is no way currently to add dynamic KM plot. **Solution**: show the survival tab and table when there is no _status variable AND there are timeline data.

PR #4907, opened 2024-05-08
♻️ Refactor code: RAS/RAF alterations in colorectal c…

SAHU-01:
…ancer, can't remove gene track in first attempt

Fix cBioPortal/cbioportal#10780(see https://help.github.com/en/articles/closing-issues-using-keywords)

- To maintain the gene structure in case of an OQL query deletion, a condition was set, if the deleted gene track has a sublabel after it's deletion a new line must be added to the existing genes. But when no sublabel is present a default  sublabel for '' always gets assigned in application. For the first time when remove track was clicked in RAS/RAF, it added a new line[following oql query logic because '' is also a sublabel] , and on second removal I removed the gene. To fix this issue this condition which always in the presence of a sublabel adds a new line first has to be modified
- To modify the condition, in case of oql queries, specially in the query params of `genelist` '%20' is present, on adding this check alongside if a sublabel is present, the new line addition logic now works only for oql queries, and the other remove tracks not affected.

## Checks

## Any screenshots or GIFs?

## Notify reviewers
@onursumer 

### DIFF
--- a/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
+++ b/src/shared/components/oncoprint/ResultsViewOncoprint.tsx
@@ -1428,7 +1428,7 @@ export default class ResultsViewOncoprint extends React.Component<
                     ' ; ' +
                     remainingGeneAfterDeletion
             );
-        } else if (geneticSublabel) {
+        } else if (geneticSublabel && geneListFromURL.includes('%20')) {
             //OQL Queries logic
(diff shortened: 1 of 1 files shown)


onursumer (changes requested):
Thanks for working on this!

Looks like we still need to fix the issue with OQL. See https://github.com/cBioPortal/cbioportal/issues/10780#issuecomment-2101099438 for details

PR #4911, opened 2024-05-10
Fix LazyMobXTable countHeader when only has 1 page

fuzhaoyuan:
Fix cBioPortal/cbioportal#10791

- Fix the behavior of LazyMobxTable when there is only 1 page, it should not display "(page 1 of 1)"

image


src/shared/components/lazyMobXTable/LazyMobXTable.tsx
```diff
@@ -1162,8 +1162,10 @@ export default class LazyMobXTable<T> extends React.Component<
                     fontWeight: 'bold',
                 }}
             >
-                {this.store.displayData.length} {this.store.itemsLabel} (page{' '}
-                {this.store.page + 1} of {this.store.maxPage + 1})
+                {this.store.displayData.length} {this.store.itemsLabel}
+                {this.store.maxPage + 1 > 1 &&
```
alisman, line 1166: why not just this.store.maxPage > 0?

alisman:
gotcha.  it's no biggie.  lets just leave it.

PR #4912, opened 2024-05-14
Study page clinical data tab: fix download after pagination change

alisman:


inodb (approved):
LGTM - thanks so much!

PR #4924, opened 2024-06-14
Group comparison view: update boxplot info in clinical tab

TJMKuijpers:
In the group comparison view, when we go to the clinical tab there is the option to show different clinical attributes per group. Those clinical attributes are either visualized as a stacked bar chart or boxplot.

The figure with the boxplot shows a boxplot with all samples as scatter points.
image

When we set the visualization type to table, we can see that only the boxplot information is shown:
image

This is quite confusing for the user since the minimum and maximum values are the lower and upper whisker of the boxplot, not of the data points. We should add this information but also add to the current table that these numbers are describing the boxplot.

**This would be the proposed solution** (image created with test study es_0)
image


TJMKuijpers:
@inodb in the third image: I'm not sure what the standard is --> should boxplot (and scatterplot) be one or two words. It seems that both are used when I do a quick google search.

inodb:
@TJMKuijpers Good catch! Thanks for trying to come up with a solution to this

I think the "Table" name here is unintuitive. How about we call it "Summary Statistics"?

In the "Summary Statistics" maybe better to not put the words 'boxplot' nor 'scatterplot'. E.g. imagine sharing a link to this table directly. It's unclear what's meant by scatterplot/boxplot. The boxplot stats are already shown as a tooltip on the boxplot itself. I think it's more clear to only show summary statistics similar to e.g. what the R package does here: https://www.reneshbedre.com/blog/describe-function-in-r.html. I.e. just the scatterplot stats and maybe a few others that the R package reports

Does that make sense?

TJMKuijpers:
@inodb I updated the table to match the descriptive information Pandas or R returns:
image

Table will be updated when the user clicks on "log scale':
image
image

TJMKuijpers:
@inodb  and @dippindots could you have a look at this PR?

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -1014,3 +1014,72 @@ export function toBoxPlotData<D extends IBaseBoxScatterPlotPoint>(
             );
         });
 }
+
+export function toDataDescriptive<D extends IBaseBoxScatterPlotPoint>(
+    data: IBoxScatterPlotData<D>[],
+    logScale?: any
```
dippindots, line 1020: Same here, could we try to create a specific data structure instead of using `any`? It would be much harder to maintain certain code if we have too many uncertain types inside our codebase

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -1014,3 +1014,72 @@ export function toBoxPlotData<D extends IBaseBoxScatterPlotPoint>(
...
+        const count = scatterValues.length;
+        // Calculate minimum and maximum
+        const minimum = Math.min(...scatterValues);
+        const maximum = Math.max(...scatterValues);
+        // Calculate mean
+        const total = scatterValues.reduce(
+            (sum: number, value: number) => sum + value,
+            0
+        );
+        const mean = total / scatterValues.length;
```
dippindots, line 1035:
```suggestion
        const mean = _.mean(scatterValues);
```

dippindots (commented):
@TJMKuijpers Overall it looks good to me codewise, just had a few comments about the type, it seems like we have more than one `any` type, and we should try to create specific types for them.

src/shared/components/plots/BoxScatterPlot.tsx
```diff
@@ -1014,3 +1014,72 @@ export function toBoxPlotData<D extends IBaseBoxScatterPlotPoint>(
...
+        );
+        const count = scatterValues.length;
+        // Calculate minimum and maximum
+        const minimum = Math.min(...scatterValues);
+        const maximum = Math.max(...scatterValues);
+        // Calculate mean
+        const total = scatterValues.reduce(
+            (sum: number, value: number) => sum + value,
+            0
+        );
```
dippindots, line 1034: After we changed the mean calculation to `_.mean`, I think we can safely remove these lines

dippindots, line 1034: After we changed the mean calculation to `_.mean`, I think we can safely remove these lines

dippindots (commented):
Thanks for updating those types, they look great now. I just have one more comment on this, then we also need @inodb to sign off from the product side

dippindots (approved):
LGTM

inodb (commented):
Thanks so much @TJMKuijpers !

This looks great!

image

Few thoughts:

- MAD - should we spell this one out like `Mean Absolute Deviation (MAD)`?
- "Lower/higher whisker" - whisker looks odd without seeing the chart. Do we need those? Maybe better to remove? The whisker values show up in the a tooltip of the chart itself on mouseover, so not sure if we need them for the table?

inodb (commented):
LGTM - thank you!

PR #4930, opened 2024-06-24
Fix mutational sig bug

TJMKuijpers:
Fix https://github.com/cBioPortal/cbioportal/issues/10538. This is a fix for studies when only mutational signatures contribution and p-values are loaded (and no mutational signatures count data):
For a study without mutational signatures count data:
   
For a study with mutational signatures count data:


alisman (approved):

inodb:
@TJMKuijpers thanks so much for this fix! Looks like there's a failing unit test

PR #4932, opened 2024-06-24
WSL setup updates - readme and gitignore for VS

pappde:
This is a no-code update. It updates the .gitignore and the readme.

## Changes
- .gitignore: added ".vs" folder for users using Visual Studio
- readme: updated to improve experience for new users. Added several tips for WSL users.

Details:
- added section at bottom for WSL tips
- removed older WSL Tip that seems to be no longer na issue
- replaced deprecated yvm tip with new commands
- added BRANCH_ENV tip
- added localdev bookmarklet tip
- e2e tests: added missing "cd .."

PR #4937, opened 2024-07-03
Fix various warnings from the browser console

pappde:
## Summary

Fixes some warnings that appear in the browser console when loading GroupComparisonPage.

## Warnings fixed

Primarily warnings originating from prop_types package validation.

- <Popover> Warning: Failed prop type: The prop `id` is required to make `Popover` accessible for users of assistive technologies such as screen readers
  - fixed several instances
- GroupSelectorButton: Warning: Failed prop type: The prop `onHide` is marked as required in `Overlay`, but its value is `undefined`
- appBootstrapper.loadCustomJs: fixed resolve() missing parameter warning by explicitly declaring Promise<void> instead of Promise

## Note

These commits are dependent on PR #4927 and PR #4932 - the first one is a dependency for my environment.


packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
```diff
@@ -86,6 +86,7 @@ export const TimelineTracks: React.FunctionComponent<ITimelineTracks> = observer
                     return (
                         <Portal container={document.body}>
                             <Popover
+                                id="TimelineTracksTooltipPopover"
```
alisman, line 89: is this to fix a warning?  i worry this gets transferred to dom id and we want to keep IDs out of DOM in general

src/pages/groupComparison/groupSelector/GroupSelectorButton.tsx
```diff
@@ -166,8 +171,10 @@ class GroupSelectorButton extends React.Component<
                             placement="top"
                             show={this.hovered}
                             target={this.button}
+                            onHide={this.onHideOverlay}
```
alisman, line 174: curious about this one

alisman:
@pappde thanks for doing this.  see my question about the ids.

packages/cbioportal-clinical-timeline/src/TimelineTracks.tsx
pappde: @alisman Yes, this fixes a warning that was showing in the browser console from prop_types. Specifically, "Warning: Failed prop type: The prop id is required to make Popover accessible for users of assistive technologies such as screen readers"

Honestly, I'm not familiar with screen readers or the  implication if you don't have an id here.

src/pages/groupComparison/groupSelector/GroupSelectorButton.tsx
pappde: @alisman This fixes "Warning: Failed prop type: The prop onHide is marked as required in Overlay, but its value is undefined".

pappde:
@alisman Noted which warning those particular changes fix. Let me know if you prefer reverting any of them. 

Aside: These commits only make a little dent in all the prop_types warnings. I wonder if it would be helpful to update the Getting Started (readme) with some tips on how to filter them out.

src/pages/groupComparison/groupSelector/GroupSelectorButton.tsx
alisman: this seems risky to me since there is is nothing functionally broken so far as i know.  if we're just trying to get rid of a warning, then maybe a noop?

alisman:
@pappde yeah, i think it would better to update README to filter these warnings out.   i checked and adding the id does indeed result in a DOM id, which i would rather avoid.     one can see that onHide should not be a required prop (contrary to my initial comment above) which is in line with my general sense that these prop types are of limited value

pappde:
@alisman Agreed, I'll close this out and update dependent PRs.

PR #4957, opened 2024-08-01
Add Histogram Zoom Preview on Study View

olzhasmukayev:
Add Histogram Zoom Preview:
image


src/pages/studyView/chartHeader/ChartHeader.tsx
```diff
@@ -602,6 +606,31 @@ export class ChartHeader extends React.Component<IChartHeaderProps, {}> {
...
+                items.push(
+                    <li>
+                        <a
+                            className="dropdown-item"
+                            onClick={this.props.togglePreview}
+                        >
+                            <FlexAlignedCheckbox
+                                checked={
+                                    !!(
+                                        this.props.chartControls &&
```
gblaih, line 619: simplify to `!!this.props.chartControls?.showPreview`

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -599,6 +599,47 @@ export class StudyViewPageStore
...
+        if (this._customDataBinFilterSet.has(uniqueKey)) {
+            const filter = this._customDataBinFilterSet.get(uniqueKey);
+            return filter === undefined
+                ? false
+                : (filter!.showPreview as boolean);
+        } else {
+            const filter = this.getDataBinFilterSet(uniqueKey).get(uniqueKey)!;
+            return filter?.showPreview === undefined
+                ? false
+                : (filter!.showPreview as boolean);
```
gblaih, line 639: i think this could be simplified. one way could be to use ternary operator (n ? x : y) to get the filter and something like `return filter?.showPreview || false` for the return value

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -74,7 +84,37 @@ export default class BarChart extends React.Component<IBarChartProps, {}>
 
     constructor(props: IBarChartProps) {
         super(props);
+        this.state = {
```
gblaih, line 87: we mainly use mobx for state management so instead of this.state and setState we use decorators like @observable and @action. can we look into this and see if we can do that?

src/pages/studyView/charts/barChart/BarChart.tsx
```diff
@@ -74,7 +84,37 @@ export default class BarChart extends React.Component<IBarChartProps, {}>
...
+        if (prevProps.showPreviewChecked && !this.props.showPreviewChecked) {
+            this.setState({ zoomDomain: null });
+        }
+    }
+
+    private handleBrush(domain: any) {
+        if (!domain || !domain.x) {
+            console.error('Brush domain is not defined.');
+            return;
+        }
```
gblaih, line 105: lets remove this and just check if `domain` and `domain.x` are defined and if so do the actions below.

olzhasmukayev:
This is copy of the https://github.com/cBioPortal/cbioportal-frontend/pull/5075

PR #4958, opened 2024-08-01
Fix plots tab related data loading on initial study view load

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/10897

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -855,7 +855,7 @@ export default class StudyViewPage extends React.Component<
                                             structuralVariants={
                                                 this.store.structuralVariants
                                             }
-                                            hugoGeneSymbols={
+                                            allHugoGeneSymbols={
                                                 this.store.allHugoGeneSymbols
                                             }
                                             selectedGenericAssayEntitiesGroupByMolecularProfileId={
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -850,15 +850,13 @@ export default class StudyViewPage extends React.Component<
                                                     .driverAnnotationSettings
                                             }
                                             studyIdToStudy={
-                                                this.store.studyIdToStudy.result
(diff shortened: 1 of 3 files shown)


src/shared/components/plots/PlotsTab.tsx
```diff
@@ -3679,7 +3679,9 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                     this.vertGenericAssayOptions.result,
                     selectedEntities,
                     this._vertGenericAssaySearchText,
-                    this.props.hugoGeneSymbols,
+                    Array.isArray(this.props.hugoGeneSymbols)
+                        ? this.props.hugoGeneSymbols
+                        : this.props.hugoGeneSymbols.result,
```
alisman, line 3684: seems strange.  i guess this is due to fact that it's used differently in different places. but there should be a better way to handle this.

alisman (changes requested):
lets see if there's a nicer way to handle some of the conditionality here.

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -842,15 +842,13 @@ export default class StudyViewPage extends React.Component<
...
                                             studyIdToStudy={
-                                                this.store.studyIdToStudy.result
+                                                this.store.studyIdToStudy
                                             }
                                             structuralVariants={
                                                 this.store.structuralVariants
-                                                    .result
                                             }
-                                            hugoGeneSymbols={
+                                            allHugoGeneSymbols={
```
alisman, line 850: Hi Bryan, sorry it took me SO long to respond.  This is quite confusing to have a property hugeGeneSymbols AND allHugeGeneSymbols.  Lets think if there's some better we can handle this.

alisman: The basic problem is that we need to wait for the gene symbols to load but we only want to do that when the tab is actually showing.  We can handle that more elegantly with some kind of loading wrapper.  Maybe we can do a pair programming session and figure it out.

alisman:
Superseded by 5087

PR #4960, opened 2024-08-06
Add Stacked Bar Chart and sorting to Plots Tab

SURAJ-SHARMA27:
Addition of Sort By MinorCategory Options:

- Implemented a new feature allowing users to sort by minorCategory.
- Added a dropdown menu with minorCategory options for sorting.

Sorted Stacks by Selected MinorCategory:

- For the selected option from the dropdown, entities will appear at the beginning of each stack in a sorted order with respect to the count of chosen minorCategory.

## Any screenshots or GIFs?
[screen-capture (9).webm](https://github.com/user-attachments/assets/fc1d29a4-fc8a-4980-9444-25eae8a9e2a0)

## Notify reviewers
@alisman 
@sowmiyaa-kumar
@zeynepkaragoz
@TJMKuijpers
@inodb


inodb:
@SURAJ-SHARMA27 this looks awesome - thank you!

For the sorting, can we also sort by "# samples"? The value of the Y-axis basically

image

This would be similar to e.g. how it works here for [treatment response](https://www.cbioportal.org/study/plots?id=ccle_broad_2019&plots_horz_selection=%7B%22dataType%22%3A%22none%22%7D&plots_vert_selection=%7B%22dataType%22%3A%22TREATMENT_RESPONSE%22%7D&plots_coloring_selection=%7B%7D):
image


This is slightly separate, but maybe for treatment response, the sorting should be using your new "Sort By" component too

schultzn:
Looks great. It would be nice to also be able to sort by the overall number of samples (I see that comment above) and also to change the sort order back to alphabetical.

jjgao:
Good work @SURAJ-SHARMA27 !

- Somehow x-axis labels does not change - always alphabetically, but the bars are changing

image

- When switch Plot Type, should we change the sort value, ie. sort by number of samples if it's "Stacked bar chart" and by % samples if it's "100% stacked bar chart"?

- Should the 'Sort By' menu be included in the 'Vertical Axis' section since it's using the values from vertical axis.

SURAJ-SHARMA27:
> Good work @SURAJ-SHARMA27 !
> 
> * Somehow x-axis labels does not change - always alphabetically, but the bars are changing
> 
> image
> 
> * When switch Plot Type, should we change the sort value, ie. sort by number of samples if it's "Stacked bar chart" and by % samples if it's "100% stacked bar chart"?
> * Should the 'Sort By' menu be included in the 'Vertical Axis' section since it's using the values from vertical axis.
> * It would be nice to add the sort by parameter to the url.

Actually, I implemented that earlier, as you can see in the video I attached at the time of the PR. The labels were changing, but I forgot to include the logic when pushing it. Thank you for pointing that out. I have now made the changes and pushed it.

SURAJ-SHARMA27:
> @SURAJ-SHARMA27 this looks awesome - thank you!
> 
> For the sorting, can we also sort by "# samples"? The value of the Y-axis basically
> 
> image
> This would be similar to e.g. how it works here for [treatment response](https://www.cbioportal.org/study/plots?id=ccle_broad_2019&plots_horz_selection=%7B%22dataType%22%3A%22none%22%7D&plots_vert_selection=%7B%22dataType%22%3A%22TREATMENT_RESPONSE%22%7D&plots_coloring_selection=%7B%7D): image
> 
> This is slightly separate, but maybe for treatment response, the sorting should be using your new "Sort By" component too

Sir, I just want to be sure before starting the work: should the feature  include sorting based on the total number of samples per bar? This means the longest bar should come first, followed by the others accordingly.

SURAJ-SHARMA27:
I think I have implemented most of the suggestions and added two more options for sorting: sort by the number of samples and alphabetically. If you could review it. @inodb 

[screen-capture (14).webm](https://github.com/user-attachments/assets/67779d29-2ab3-477d-9a33-73f075b58c11)

inodb:
Amazing - nice work @SURAJ-SHARMA27 !

> Should the 'Sort By' menu be included in the 'Vertical Axis' section since it's using the values from vertical axis.

@jjgao Good question - I kinda like it separate from x/y axis selection since we might also want to allow sorting by multiple values in the future (which could be like a multi-select dropdown element)

Should we rename `# samples` to `Number of Samples`? I like the consistency with the current y-axis and we use `#` in several places, but it's maybe not as obvious in the sorting what `#` means

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -58,6 +58,11 @@ export interface IMultipleCategoryBarPlotProps {
     svgRef?: (svgContainer: SVGElement | null) => void;
     pValue: number | null;
     qValue: number | null;
+    SortByDropDownOptions?: { value: string; label: string }[];
```
gblaih, line 61: sortByDropDownOptions

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -425,6 +430,43 @@ export default class MultipleCategoryBarPlot extends React.Component<
 
     @computed get labels() {
         if (this.data.length > 0) {
+            if (this.props.sortByOption == 'SortByTotalSum') {
```
gblaih, line 433: ===

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -425,6 +430,43 @@ export default class MultipleCategoryBarPlot extends React.Component<
...
+                    });
+                });
+                const sortedMajorCategories = Object.keys(
+                    majorCategoryCounts
+                ).sort(
+                    (a, b) => majorCategoryCounts[b] - majorCategoryCounts[a]
+                );
+                return sortedMajorCategories;
+            } else if (
+                this.props.sortByOption != '' &&
```
gblaih, line 452: lets use === and !== throughout the code

src/shared/components/plots/PlotsTab.tsx
```diff
@@ -5400,6 +5432,13 @@ export default class PlotsTab extends React.Component<IPlotsTabProps, {}> {
                                     horizontalBars={this.horizontalBars}
                                     percentage={isPercentage}
                                     stacked={isStacked}
+                                    SortByDropDownOptions={
+                                        this?.SortByDropDownOptions
+                                    }
+                                    updateDropDownOptions={
+                                        this?.updateDropDownOptions
+                                    }
+                                    sortByOption={this?.sortByOption}
```
gblaih, line 5441: is `this?.` necessary? i'd think `this.` is fine

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -425,6 +430,43 @@ export default class MultipleCategoryBarPlot extends React.Component<
 
     @computed get labels() {
         if (this.data.length > 0) {
+            if (this.props.sortByOption == 'SortByTotalSum') {
+                const majorCategoryCounts: any = {};
+
+                this.data.forEach(item => {
```
gblaih, line 436: this code looks pretty similar to the the function `sortDataByOption` you made below. it would be nice to extract this into a function that can be reused

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,72 @@ export function sortDataByCategory<D>(
...
+                    majorCategoryCounts[majorCategory] = 0;
+                }
+                majorCategoryCounts[majorCategory] += count;
+            });
+        });
+        const sortedMajorCategories = Object.keys(majorCategoryCounts).sort(
+            (a, b) => majorCategoryCounts[b] - majorCategoryCounts[a]
+        );
+
+        const reorderCounts = (counts: any) => {
```
gblaih, line 121: instead of making this function inside, lets extract this into a new function outside for clarity and reuse

SURAJ-SHARMA27:
I have implemented all the changes as suggested. You can review them. I am also attaching a video for reference.
@inodb  @gblaih 

[screen-capture (15).webm](https://github.com/user-attachments/assets/22665468-8985-4293-829c-df336e717a1c)

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
```diff
@@ -435,6 +451,18 @@ export default class MultipleCategoryBarPlot extends React.Component<
         }
     }
 
+    private setInitialSelectedOption = () => {
```
alisman, line 454: this is a bit of an anti-pattern.  calling from the child component to update the state of parent component on render.  it should be possible to do this in parent component prior to mount?

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
         }
     });
 }
+export function getSortedMajorCategories(
```
alisman, line 101: can we have a comment explaining what is meant by "Major Category" ?

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
...
+    data: IMultipleCategoryBarPlotData[],
+    sortByOption: string | undefined
+): string[] {
+    if (sortByOption === 'SortByTotalSum') {
+        const majorCategoryCounts: any = {};
+
+        data.forEach(item => {
```
alisman, line 108: i think you can do this operation with flatMap, which will avoid nesting forEach (which is confusing)

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
...
+    } else if (sortByOption !== '' && sortByOption !== 'alphabetically') {
+        const sortedEntityData = data.find(
+            item => item.minorCategory === sortByOption
+        );
+        if (sortedEntityData) {
+            sortedEntityData.counts.sort((a, b) => b.count - a.count);
+
+            return sortedEntityData.counts.map(item => item.majorCategory);
+        }
+    }
```
alisman, line 130: just complete the else here.

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
         }
     });
 }
+export function getSortedMajorCategories(
+    data: IMultipleCategoryBarPlotData[],
+    sortByOption: string | undefined
+): string[] {
+    if (sortByOption === 'SortByTotalSum') {
```
alisman, line 105: see if SortByTotalSum is available on an enum somewhere.  seems likely that it is

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
...
+                    majorCategoryCounts[majorCategory] = 0;
+                }
+                majorCategoryCounts[majorCategory] += count;
+            });
+        });
+
+        return Object.keys(majorCategoryCounts).sort(
+            (a, b) => majorCategoryCounts[b] - majorCategoryCounts[a]
+        );
+    } else if (sortByOption !== '' && sortByOption !== 'alphabetically') {
```
alisman, line 121: use constant for 'alphabetically' or see if it's available on an enum

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
```diff
@@ -98,6 +98,71 @@ export function sortDataByCategory<D>(
...
+    return [];
+}
+export function sortDataByOption(
+    data: IMultipleCategoryBarPlotData[],
+    sortByOption: string
+): IMultipleCategoryBarPlotData[] {
+    const sortedMajorCategories = getSortedMajorCategories(data, sortByOption);
+
+    if (sortByOption === 'SortByTotalSum' || sortedMajorCategories.length > 0) {
+        const reorderCounts = (counts: any) => {
```
alisman, line 141: avoid 'any'

inodb:
@SURAJ-SHARMA27 Just discussed this during the community call, some feedback from the group:

- When switching to percentage, the bars should sort by the percentage value (rather than absolute value)
- On the Table: maybe disable the sort here? Or if it is easy to enable sorting there too, enable it.
- Remove "Sort By" on charts where it is not functional (e.g. continuous vs continuous). Currently, the "Sort by" seems to show old values from a previous selection

src/pages/groupComparison/MultipleCategoryBarPlot.tsx
SURAJ-SHARMA27: In the parent component, it doesn't contain data in the form of `IMultipleCategoryBarPlotData`, from which I have to extract all the unique major categories for dropdown options. Therefore, to avoid extra calculations, I passed the prop from the parent and updated it when it became available in the child.

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: image


The `data `is defined in a way that it contains x-axis labels referred to as major categories. This functionality is used twice: once during the sorting of data and again when the getLabels function in MultipleCategoryBarPlot.tsx is called. Therefore, I created a separate function for that to enhance reusability.

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: The above function is used to find the sum of each `majorCategory` for different `minorCategories`. For instance, consider the following example:

data contains three `minorCategory` names: Biopsy, Resection, and Cusa. Each of these `minorCategories` contains a large array named counts, and each element in that array is an object containing `majorCategory`, count, and percentage. For example, for Biopsy, the `majorCategories` might include Germ Cell Tumor and Thyroid Cancer....

I am maintaining a totalSum of these majorCategories and then applying sorting.

If I use flatMap, it would first be difficult to determine which majorCategory belongs to which minorCategory. Additionally, I think it would require more space and have an equivalent time complexity of O(n * m) as it does now with the two nested loops. First, I would need to perform the flatMap operation (a map operation followed by a flattening operation, essentially equivalent to a nested loop) and then use reduce to accumulate the counts. What are your suggestions should i follow another approach?

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: yes completed

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: No, it was not there; I have defined an enumeration and used it

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: I have defined an enumeration and used it

src/shared/components/plots/MultipleCategoryBarPlotUtils.ts
SURAJ-SHARMA27: Yes, I have defined the types across all the changes.

SURAJ-SHARMA27:
> @SURAJ-SHARMA27 Thanks for the fixes! I noticed the "Sort by" option is currently missing [here](https://deploy-preview-4960--cbioportalfrontend.netlify.app/study/plots?id=msk_impact_2017&plots_horz_selection=%7B%22dataType%22%3A%22clinical_attribute%22%2C%22selectedDataSourceOption%22%3A%22CANCER_TYPE%22%7D&plots_vert_selection=%7B%22dataType%22%3A%22clinical_attribute%22%2C%22selectedDataSourceOption%22%3A%22SPECIMEN_TYPE%22%7D&plots_coloring_selection=%7B%7D), do you see that?
> 
> image

yes i fixed it

SURAJ-SHARMA27:
> @SURAJ-SHARMA27 Just discussed this during the community call, people liked it! Some more feedback from the group:
> 
> * When switching to percentage, the bars should sort by the percentage value (rather than absolute value)
> * On the Table: maybe disable the sort here? Or if it is easy to enable sorting there too, enable it.
> * Remove "Sort By" on charts where it is not functional (e.g. continuous vs continuous). Currently, the "Sort by" seems to show old values from a previous selection

Thank you @inodb  sir, just few more doubts
i) This means that when the number of samples is selected from the dropdown, the data should be sorted by the percentage value.
ii) Sorting will be shown only for the stacked bar chart; I have fixed that.

inodb:
> i) This means that when the number of samples is selected from the dropdown, the data should be sorted by the percentage value.
 @SURAJ-SHARMA27 correct

SURAJ-SHARMA27:
> > i) This means that when the number of samples is selected from the dropdown, the data should be sorted by the percentage value.
> 
> @SURAJ-SHARMA27 correct
> 
> > ii) Sorting will be shown only for the stacked bar chart; I have fixed that.
> 
> Thanks so much for fixing!

okay sir, I think I did it correctly. You can review it (pushed already). The bars will be sorted according to the percentages of the selected minor category. If the number of samples is selected from the dropdown, the bars are sorted by the total absolute count for each stack. For alphabetical sorting, the bars will be sorted as usual.

[screen-capture (33).webm](https://github.com/user-attachments/assets/f343c5b9-dd55-46c5-8756-0e48ebb00a78)

alisman:
Hi @SURAJ-SHARMA27,
There is a regression error caused by this PR.  I realize GSOC is long over.  Do you have a few minutes to look at the issue?

If you [visit this link](https://www.cbioportal.org/results/comparison?Z_SCORE_THRESHOLD=2.0&cancer_study_id=coadread_tcga_pub&cancer_study_list=coadread_tcga_pub&case_set_id=coadread_tcga_pub_nonhypermut&comparison_selectedGroups=%5B%22Altered%20group%22%2C%22Unaltered%20group%22%2C%22KRAS%22%2C%22NRAS%22%5D&gene_list=KRAS%20NRAS%20BRAF&gene_set_choice=user-defined-list&genetic_profile_ids_PROFILE_COPY_NUMBER_ALTERATION=coadread_tcga_pub_gistic&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=coadread_tcga_pub_mutations&comparison_subtab=clinical
) with the PR active, you'll see the following problem:

image

If you are busy with school or work, we understand and will find someone else to look into it.

dippindots:
Related ticket: https://github.com/cBioPortal/cbioportal/issues/10722

alisman:
@SURAJ-SHARMA27 just FYI, we have a dev @gblaih who is taking care of the last fixes.

src/shared/components/plots/MultipleCategoryBarPlot.tsx
```diff
@@ -913,8 +944,12 @@ export default class MultipleCategoryBarPlot extends React.Component<
         }
     }
 
-    componentDidUpdate() {
+    componentDidUpdate(prevProps: IMultipleCategoryBarPlotProps) {
         this.updateLegendWidth();
+
+        if (this.props.horizontalBars !== prevProps.horizontalBars) {
```
alisman, line 950: @gblaih should be better to use the horizontalBars value in the component "key" attribute.  This will force it to re-mount when that value changes and so you can avoid dirtiness like the above.  You might lose other state, but it's a cleaner approach.  Give it a try.

alisman: and apologies if this was not your change

inodb (approved):
This looks great!

There's one minor issue where occasionally when you mess around with swapping axes you see stuff like this:

image


But haven't be able to figure out the steps to reproduce it reliably so I think ok to merge it and see if someone else comes up with a good way to reproduce it post merge

alisman (approved):

PR #4961, opened 2024-08-08
Add Patient and Sample Resources to Study View 

Nelliney:
1. Add a new tab, "Files & Links," to the Study View.
2. In the tab, add a searchable table that displays patient and sample resources, the type and description of resources, and the number of resources per patient.
3. Add the total number of resources in the study at the top of the page.
4. Define the behavior for when this tab is displayed.
5. Enhance the user interface by adding borders, icons, and a title to the table.
6. Add threshold so it only works for cohorts with max 500 patients. We don't have a GET or POST request API to get all patient and sample resources in a study atm, so this is to avoid having to make >500 GET Requests 

Screenshot 2024-08-08 at 12 31 13


src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,332 @@
...
+                }
+
+                return false;
+            },
+        };
+    }
+
+    @observable searchTerm: string | undefined = undefined;
+
+    readonly getData = remoteData({
```
onursumer, line 181: `getData` is a confusing name because this is not a function. It's better to name it as `sampleData` or something similar.
```suggestion
    readonly sampleData = remoteData({
```

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,332 @@
...
+                'patientId',
+                'asc',
+                RECORD_LIMIT
+            );
+
+            return Promise.resolve(sampleData);
+        },
+    });
+
+    readonly columns = remoteData({
```
onursumer, line 205: Ideally this should be `@computed` not `remoteData`

```suggestion
    @computed get columns() {
```

Here is a sample implementation https://github.com/cBioPortal/cbioportal-frontend/blob/0b095b1c3eef6d2b4119acb36795720202b231a2/src/pages/patientView/copyNumberAlterations/CopyNumberTableWrapper.tsx#L142

onursumer (changes requested):

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    resourcesPerPatient[data[0].patientId] = 0;
+
+                resourcesPerPatient[data[0].patientId] += data.length;
+            }
+        });
+    };
+
+    itemBuilder(sampleIdResourceData);
+    const resourcesForPatients = await fetchResourceDataOfPatient(patientIds);
+    itemBuilder(resourcesForPatients);
```
onursumer, line 109: Can we add some comments here to describe what we are doing? It's not very easy to follow the logic.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+        });
+    });
+
+    const resourcesPerPatient: { [key: string]: number } = {};
+    const items: { [attributeId: string]: any } = [];
+
+    const itemBuilder = (resourceData: { [key: string]: ResourceData[] }) => {
+        forEach(resourceData, (data: ResourceData[]) => {
+            forEach(data, (resource: ResourceData) => {
+                items.push({
```
onursumer, line 87: It's not ideal to modify the variables outside the function. it would be better if this function returned `items`.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    resourcesPerPatient: 0,
+                    typeOfResource: resource?.resourceDefinition?.displayName,
+                    description: resource?.resourceDefinition?.description,
+                    url: resource?.url,
+                });
+            });
+
+            if (data && data.length > 0) {
+                if (!(data[0].patientId in resourcesPerPatient))
+                    resourcesPerPatient[data[0].patientId] = 0;
```
onursumer, line 100: Similar to `items`, it would be better to return `resourcesPerPatient` instead of defining a variable outside the function.

src/pages/studyView/resources/FilesAndLinks.tsx
onursumer, line 109: We can probably combine `sampleIdResourceData` and `resourcesForPatients` and call the `itemBuilder` only once. Also, it's better to rename `itemBuilder` to `buildItemsAndResources` to improve readability.

So, something like this would be better assuming that `buildItemsAndResources` will return those values instead of modifying variables outside the function:

```JavaScript
const {resourcesPerPatient, items} = buildItemsAndResources(resourcesForPatientsAndSamples);
```

src/shared/api/urls.ts
```diff
@@ -120,15 +120,24 @@ export function getSampleViewUrl(
     studyId: string,
     sampleId: string,
     navIds?: { patientId: string; studyId: string }[]
+) {
+    return getSampleViewUrlWithPathname(studyId, sampleId, 'patient', navIds);
+}
+export function getSampleViewUrlWithPathname(
```
onursumer, line 126:
```suggestion
}

export function getSampleViewUrlWithPathname(
```

onursumer (changes requested):

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+
+async function fetchFilesLinksData(
+    filters: StudyViewFilter,
+    sampleIdResourceData: { [sampleId: string]: ResourceData[] },
+    searchTerm: string | undefined,
+    sortAttributeId: string | undefined,
+    sortDirection: 'asc' | 'desc' | undefined,
+    recordLimit: number
+) {
+    let sampleClinicalDataResponse = await getAllClinicalDataByStudyViewFilter(
```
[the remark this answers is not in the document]
Nelliney: I have changed it, however, I used ClinicalDataTab.tsx as an example for my tab, that is - it has the same declaration for the line:

let sampleClinicalDataResponse = await getAllClinicalDataByStudyViewFilter(

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: The reason of this implementation is performance and avoiding merging multiple lists. Otherwise, we need to merge multiple arrays before returning the results, that is - the merge will cost O(n).

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: (the same as above). To avoid performance issues and extra merging, it was implemented in that way. 
Do you want me to modify it?

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Okay, I will change it accordingly, however, it will increase time/space complexity.

src/shared/api/urls.ts
Nelliney: Done

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -5784,6 +5784,43 @@ export class StudyViewPageStore
...
+                d => d.resourceType === 'SAMPLE'
+            );
+            if (!sampleResourceDefinitions.length) {
+                return Promise.resolve({});
+            }
+
+            const samples = this.samples.result!;
+            const ret: { [sampleId: string]: ResourceData[] } = {};
+            const promises = [];
+            for (const sample of samples) {
```
alisman, line 5802: @Nelliney there will routinely be many thousands of samples in this list, so i think this code will fire a request for each of them (in fact multipe requests for each resource type).  so we need to find a way to get a summary from backend.  lets discuss at standup 
(fyi @inodb)

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+        searchTerm,
+        sortAttributeId,
+        sortDirection,
+        recordLimit,
+        0
+    );
+
+    const patientIds = new Map<string, string>();
+
+    forEach(sampleClinicalDataResponse.data, (data, uniqueSampleId) => {
```
alisman, line 75: lets use _.forEach so we know where forEach comes from.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+        forEach(data, item => {
+            patientIds.set(item.patientId, item.studyId);
+        });
+    });
+
+    const resourcesPerPatient: { [key: string]: number } = {};
+    const items: { [attributeId: string]: any } = [];
```
alisman, line 82: this type doesn't see right.  it's type as a map, but we assign an array to it.

src/pages/studyView/StudyViewPageStore.ts
alisman: I spoke @inodb and i now understand there's some limit put on number of items.  @Nelliney does this only make the requests if the user has opened the resource tab?

src/pages/studyView/resources/FilesAndLinks.tsx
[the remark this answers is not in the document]
alisman: yeah, that's an oversight.  in general, we should always use const unless we actually want to be able to update the variable after we declare it.

src/pages/studyView/resources/FilesAndLinks.tsx
alisman: @Nelliney at the relatively small scale of our data, the cost of iteration is very low, almost negligible.  so we generally err on the side of readability unless we can demonstrate there is a performance issue.  for example something like this would be more easy to understand and functional:
```
const things = _(resourceData).values().flatMap(resource=>{ .... }).value();
```
now, in one line, I can understand what we are doing.    i don't have to wrap my head around "itemBuilder"

if you need to iterate through the list AGAIN to accomplish what's at line 98, that's fine.

src/pages/studyView/resources/FilesAndLinks.tsx
alisman: also, lets see if we can improve on "any".   What does the table accept?

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -377,6 +377,14 @@ export default class StudyViewPage extends React.Component<
...
 
+    @computed get shouldShowResourcesTab() {
+        if (this.store.resourceDefinitions.isComplete) {
+            return this.store.resourceDefinitions.result.length > 0;
+        } else {
+            return false;
+        }
+    }
+
     @computed get shouldShowResources() {
```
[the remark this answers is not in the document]
Nelliney: It was used for showing the tab originally, also, there is some old code in the Resource Tab that might need to be used in the future. However, I will remove it according to your comment, if it is needed in the future, it can be added easily.

src/pages/studyView/StudyViewPageStore.ts
Nelliney: Basically, I agree with you @alisman about improving the performance, it is not the best option to implement this logic on the UI. Moreover, there is another the same problem that might affect the performance due to multiple calls. I spoke with @inodb about this matter. Also, I would like to inform you that the function was taken from PatientViewPageStore (1021 line) where it calls many thousands of samples in this list. 

Yes, it will be used when the tab opens.

src/pages/studyView/resources/FilesAndLinks.tsx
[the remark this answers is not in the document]
Nelliney: I changed it.

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: string and numbers (due to sorting of resourcesPerPatient), do you want me to cast number to string? And then we can change any to string.

I saw some parsing in getDefaultColumnConfig -> sortBy, so it might work.

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: the native forEach method was used in the case since it is sufficient and preferred due to its simplicity and performance, that is - it is considered slightly faster as it is a native method. On another hand, _.forEach:  may be slightly slower than forEach due to its additional functionality and overhead. However, the difference is usually negligible in most applications.

So, if the project's guidelines are to use _.forEach, I will make the change according.

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

src/pages/studyView/resources/FilesAndLinks.tsx
alisman: you aren't using the "native" forEach method.  you have imported forEach from lodash (see ln 7).   there is no native forEach method which iterates over a map.  you'd have to use a for in loop.   but we don't recommend that either.

Nelliney: Good catch! Thank you for pointing that out. I've changed it.

src/pages/studyView/resources/FilesAndLinks.tsx
Nelliney: Done

alisman:
@Nelliney somehow some other commits made it into this PR.  I think maybe you merged master into it.  Can you try rebasing instead?  Let me know if you need help with this as it can be tricky.

Nelliney:
@alisman, thank you for letting me know. I did a rebase yesterday, and now I have synced my fork, and I believe that it should be solved. Please let me know if you have any comments or requests.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+                    sampleId: resource.sampleId,
+                    resourcesPerPatient: 0,
+                    typeOfResource: resource?.resourceDefinition?.displayName,
+                    description: resource?.resourceDefinition?.description,
+                    url: resource?.url,
+                }))
+            )
+            .value();
+
+        _(resourceData).forEach(data => {
```
alisman, line 91: not that important but for the future, this is a reduce operation.   so, it would be slightly cleaner: 
```
const resourcesPerPatient = _.reduce(resourceData, ()=>..., {})
```
the idea is always to use direct assignment instead of  declaration and foreach.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -0,0 +1,329 @@
...
+        return Promise.all(resourcesPerPatient).then(resourcesPerPatient =>
+            _(resourcesPerPatient)
+                .flatMap()
+                .groupBy('patientId')
+                .value()
+        );
+    };
+
+    const resourcesForPatients = await getResourceDataOfPatients();
+    const buildItemsAndResources = (resourceData: {
```
alisman, line 71: this function might as well be declared outside of fetchFilesLinksData.  i don't think there's a reason to declare it inside here.   in general is a bad practice because it makes it harder to test and in an iterative setting can be very redundant.

PR #4962, opened 2024-08-12
Update Genome Nexus API client to include Alpha Missense

leexgh:
Update Genome Nexus API client to include Alpha Missense in response model


onursumer (approved):
LGTM

PR #4969, opened 2024-08-20
Bugfix/hide empty tab

Nelliney:
Fixing the bug: in the patient view, when selecting the 'Files and Links' tab and navigating through patients, an empty tab was shown for those without data. This fix ensures that the 'Files & Links' tab will not be displayed for patients who don't have any data.

Before the fix:

Screenshot 2024-08-19 at 18 49 03

After the fix:

Screenshot 2024-08-19 at 18 46 44


inodb (commented):
@Nelliney @alisman  is one option to do something like this:

```
if (resourceURL) {
  if (resourceDataExists) {
    // show data in tab
  } else {
    // show tab and show text "No {tabName} for this patient". 
  }
} else {
 if (resourceDataExists) {
     // show tab
 } else {
    // hide tab
 }
} 
```
I.e. show a message when in the URL the tab is explicitly requested, but hide the tab in other scenarios. It currently feels a bit odd that it jumps back to the "Summary View" when the patient does not have the requested tab

PR #4990, opened 2024-09-03
Add truncation of timeline label

jCHENEBY:
Fix cBioPortal/cbioportal#10448. Add truncation to timeline labels. Also add padding between end of label and arrow.

- Change type of balise for TrackHeader label that display Timeline let bar and modify css accordingly
- Add some padding to end of label
- Remove unused import

With added padding
timeline-leftbar-padding

Demo of tooltip. Notes, I could not find other example so I "forced" ellipse by limiting timeline left bar size.
timeline-leftbar-tootips

### DIFF
--- a/packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
+++ b/packages/cbioportal-clinical-timeline/src/TrackHeader.tsx
@@ -4,12 +4,9 @@ import { TICK_AXIS_HEIGHT } from './TickAxis';
 import { CustomTrackSpecification } from './CustomTrack';
 import { TimelineStore } from './TimelineStore';
 import { useObserver } from 'mobx-react-lite';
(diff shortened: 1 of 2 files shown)


alisman:
@jCHENEBY not sure i'm doing something wrong but in testing it seems like it doesn't ever truncate.  
https://genie.cbioportal.org/patient?studyId=crc_public_genie_bpc&caseId=GENIE-DFCI-001292

jCHENEBY:
I had some time to look at it and the issue was that the left bar increased width to fit the label instead of truncated. I have set a fix max width to solve the issue.

I also noticed that when the font-size it set to 12px and lower, the tooltip was not displayed for the first truncated label only. I did a quick and dirty fix and set the font-size for the left-bar at 12.5px. Please let me know if you have a more elegant solution.

PR #4998, opened 2024-09-13
Add custom filter tests with new samples and assertResponse fields

gblaih:


src/shared/api/validation.ts
```diff
@@ -142,28 +143,39 @@ export function validate(
...
+                    );
+                    result.url = url;
+                    result.hash = hash;
+                    result.data = params;
+                    result.chDuration = chDuration;
+                    result.legacyDuration = legacyDuration;
+                    return result;
+                });
+            } else {
+                return {
```
alisman, line 170: where is compareCounts called when there's an assertResponse?

src/shared/api/validation.ts
```diff
@@ -258,39 +259,75 @@ export function validate(
...
-            return { body, elapsedTime: xhr.getResponseHeader('elapsed-time') };
-        });
-    }
-
-    return chXHR
-        .then(({ body, elapsedTime }: any) => {
-            let legacyUrl = url.replace(/column-store\//, '');
-            if (treatmentLegacyUrl[label]) {
-                legacyUrl = treatmentLegacyUrl[label](legacyUrl);
+        if (assertResponse && assertResponse[0] && assertResponse[0].attributeId) {
```
alisman, line 262: not sure i understand the necessity of this code up here.  the clickhouse fetch should be entirely independent of the assertResponse, no?

src/shared/api/validation.ts
```diff
@@ -258,39 +259,75 @@ export function validate(
...
                 contentType: 'application/json',
+            }).then((body, state, xhr) => {
+                return { body, elapsedTime: xhr.getResponseHeader('elapsed-time') };
             });
-            return legacyXHR.then(legacyResult => {
-                const result: any = compareCounts(body, legacyResult, label);
+        }
+    }
+
+    return chXHR
```
alisman, line 294: this is where assertResponse starts to become relevant in my understanding.

src/shared/api/validation.ts
```diff
@@ -258,39 +259,75 @@ export function validate(
...
+                    result.data = params;
+                    result.chDuration = parseFloat(elapsedTime);
+                    result.legacyDuration = parseFloat(
+                        legacyXHR.getResponseHeader('elapsed-time') || ''
+                    );
+                    return result;
+                });
+            }
+            else {
+                const result: any = compareCounts(
```
alisman, line 320: the code to build the result is redundant.  there should be a way of avoiding it.  e.g., something like around 301, if there's an assertResponse, then legacyXHR is just Promise.resolve(assertReseponse).then(...)

alisman: you will have to figure out a way of normalizing assertResponse and the XHR.  because obviously assertResponse doesn't have getResponseHeader.  That should be easy though.

src/shared/api/validation.ts
```diff
@@ -324,13 +361,21 @@ export function reportValidationResult(result: any, prefix = '') {
...
 
     result.status &&
-        console.log(
-            `${prefix} ${result.label} (${
-                result.hash
-            }) passed :) ch: ${result.chDuration.toFixed(
-                0
-            )} legacy: ${result.legacyDuration.toFixed(0)}`
-        );
+        (!result.legacyDuration
```
alisman, line 364: try to avoid these duplicating logic (console.log(...)) like this for only a "little" thing like whether it has legacyDuration.  better to introduce a variable that gets always gets printed and can sometimes be blank (e.g. when there's no legacyDuration).

alisman (changes requested):
see comments

src/shared/api/validation.ts
gblaih: i used it to get the one attribute that is used in assertResponse. so the clinical data counts clickhouse fetch only fetches data for that attribute. then its an easy compare between that response and assertResponse.

alisman:
I will merge this in a new PR since there are conflicts

PR #5013, opened 2024-09-30
Update Genome Nexus api clients to include new Mutation Assessor

leexgh:
API client changes for new Mutation assessor data (v4).
We need to deploy genome nexus first, rerun this pr, and then https://github.com/cBioPortal/cbioportal-frontend/pull/5015


leexgh:
Close this one and include this commit in the other pr

PR #5019, opened 2024-10-11
Reduce api calls for generic assay meta in study view page

dippindots:
We were unnecessarily firing generic assay meta requests for each ga entity when these could be batched (as in achieved in this PR).

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -291,7 +291,7 @@ import {
 } from 'pages/resultsView/enrichments/EnrichmentsUtil';
 import {
     fetchGenericAssayMetaByMolecularProfileIdsGroupByMolecularProfileId,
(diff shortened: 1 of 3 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -6024,35 +6024,17 @@ export class StudyViewPageStore
...
         await: () => [this.genericAssayProfiles],
         invoke: async () => {
-            return await fetchGenericAssayMetaByMolecularProfileIdsGroupByMolecularProfileId(
+            return await fetchGenericAssayMetaGroupedByMolecularProfileIdSuffix(
                 this.genericAssayProfiles.result
             );
         },
     });
 
-    readonly genericAssayStableIdToMeta = remoteData<{
```
dippindots, line 6038: No reference, deleting this function

PR #5033, opened 2024-10-25
Fix Mutations Lollipop Tool Tip Info to await promise before retrieving patient counts

haynescd:
Mutation tab failed when attempting to grab filteredPatients promise before data was available.

### DIFF
--- a/src/pages/resultsView/mutation/Mutations.tsx
+++ b/src/pages/resultsView/mutation/Mutations.tsx
@@ -334,14 +334,20 @@ export default class Mutations extends React.Component<
         count: number,
         mutations: Mutation[],
         axisMode: AxisScale
-    ): JSX.Element {
-        return (
-            <LollipopTooltipCountInfo
-                count={count}
-                mutations={mutations}
-                axisMode={axisMode}
-                patientCount={this.props.store.filteredPatients.result!.length}
-            />
-        );
+    ) {
+        if (this.props.store.filteredPatients.isComplete) {
+            return (
+                <LollipopTooltipCountInfo
+                    count={count}
+                    mutations={mutations}
+                    axisMode={axisMode}
+                    patientCount={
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -336,12 +336,19 @@ export default class Mutations extends React.Component<
         axisMode: AxisScale
     ): JSX.Element {
         return (
-            <LollipopTooltipCountInfo
-                count={count}
-                mutations={mutations}
-                axisMode={axisMode}
-                patientCount={this.props.store.filteredPatients.result!.length}
-            />
+            <>
```
alisman, line 339: no biggie but I don't think you need the <> </> and also that semicolon down there is probably getting rendered.  should be aced

alisman: i would just do 
```
if (this.props.store.filteredPatients.isComplete) {
   return <Lollipop...>
} else {
  return null;
}
```

src/pages/resultsView/mutation/Mutations.tsx
```diff
@@ -336,12 +336,19 @@ export default class Mutations extends React.Component<
...
+                    <LollipopTooltipCountInfo
+                        count={count}
+                        mutations={mutations}
+                        axisMode={axisMode}
+                        patientCount={
+                            this.props.store.filteredPatients.result.length
+                        }
+                    />
+                )}
+                ;
```
alisman, line 350: this little guy

src/pages/resultsView/mutation/Mutations.tsx
haynescd: fixed

inodb (approved):
LGTM, thanks!

PR #5034, opened 2024-10-27
On Study View handle larger number of resource files

hweej:
Fixes https://github.com/cBioPortal/cbioportal/issues/11004 (associated backend PR is merged, see https://github.com/cBioPortal/cbioportal/pull/11051). This PR modifies the FilesAndLinks table on the Study View to use the new resource-data-all endpoint, which reduces the number of API requests and enables showing thousands of resource files without performance degradation

### DIFF
--- a/src/pages/studyView/resources/FilesAndLinks.tsx
+++ b/src/pages/studyView/resources/FilesAndLinks.tsx
@@ -128,49 +128,26 @@ async function fetchFilesLinksData(
     const selectedStudyIds = [
         ...new Set(selectedSamples.map(item => item.studyId)),
     ];
-    const selectedSampleIds = selectedSamples.map(item => item.sampleId);
-    const selectedPatientIds = [
-        ...new Set(selectedSamples.map(item => item.patientId)),
-    ];
 
+    // sampleIds (+patientIds) for the selectedSamples
+    const selectedIds = new Map([
+        ...new Map(selectedSamples.map(item => [item.sampleId, item.studyId])),
+        ...new Map(selectedSamples.map(item => [item.patientId, item.studyId])),
+    ]);
+
+    // Fetch resources for entire study
     const resourcesForEntireStudy = await getResourceDataOfEntireStudy(
         selectedStudyIds
     );
 
-    const resourcesForPatientsAndSamples = resourcesForEntireStudy
-        // Filter the resources to consist of only studyView selected samples
-        // Also keep patient level resources (e.g. Those don't have a sampleId)
-        .filter(
-            (resource: ResourceData) =>
-                selectedSampleIds.includes(resource.sampleId) ||
-                selectedPatientIds.includes(resource.patientId)
+    // Filter the resources to consist of only studyView selected samples
+    // Also keep patient level resources (e.g. Those don't have a sampleId)
+    const resourcesForPatientsAndSamples = _(resourcesForEntireStudy)
+        .filter(resource =>
+            selectedIds.has(resource.sampleId || resource.patientId)
         )
-        .reduce(
-            (
-                idMap: { [key: string]: ResourceData[] },
-                resource: ResourceData
-            ) => {
-                const { resourceType } = resource.resourceDefinition;
-                const { patientId, sampleId } = resource;
-
-                if (resourceType == 'PATIENT') {
-                    if (idMap.hasOwnProperty(patientId)) {
-                        idMap[patientId].push(resource);
-                    } else {
-                        idMap[patientId] = [resource];
-                    }
-                } else if (resourceType == 'SAMPLE') {
-                    if (idMap.hasOwnProperty(sampleId)) {
-                        idMap[sampleId].push(resource);
-                    } else {
-                        idMap[sampleId] = [resource];
-                    }
-                }
-
-                return idMap;
-            },
-            {}
-        );
+        .groupBy('patientId')
+        .value();
 
     // we create objects with the necessary properties for each resource
     // calculate the total number of resources per patient.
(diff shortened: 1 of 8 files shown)


src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +127,47 @@ async function fetchFilesLinksData(
         0
     );
 
-    const resourcesForPatients = await getResourceDataOfPatients(
+    const selectedSamplesList = selectedSamples.map(item => item.sampleId);
```
alisman, line 130: selectedSampleIds

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +127,47 @@ async function fetchFilesLinksData(
...
+
+    const resourcesForEntireStudy = await getResourceDataOfEntireStudy(
         studyClinicalDataResponse.data
     );
-    const resourcesForPatientsAndSamples: { [key: string]: ResourceData[] } = {
-        ...sampleIdResourceData,
-        ...resourcesForPatients,
-    };
+
+    const resourcesForPatientsAndSamples = resourcesForEntireStudy
```
alisman, line 139: put a comment explaining why we are filtering

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +127,47 @@ async function fetchFilesLinksData(
...
+            (resource: ResourceData) =>
+                selectedSamplesList.includes(resource.sampleId) ||
+                selectedPatientsList.includes(resource.patientId)
+        )
+        .reduce(
+            (
+                idMap: { [key: string]: ResourceData[] },
+                resource: ResourceData
+            ) => {
+                const { resourceType } = resource.resourceDefinition;
```
alisman, line 150: this whole operation would be more clearly handled by lodash's keyBy function.  we like to use these kind of named helpers because they make the operation more explicit.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +127,47 @@ async function fetchFilesLinksData(
         0
     );
 
-    const resourcesForPatients = await getResourceDataOfPatients(
+    const selectedSamplesList = selectedSamples.map(item => item.sampleId);
+    const selectedPatientsList = [
+        ...new Set(selectedSamples.map(item => item.patientId)),
+    ];
+
+    const resourcesForEntireStudy = await getResourceDataOfEntireStudy(
```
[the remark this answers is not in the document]
hweej: Done, removed the request for clinicaldata

src/pages/studyView/resources/FilesAndLinks.tsx
hweej: Hm, I couldn't make lodash keyBy function to work the way I wanted it to for this use case. In python this would be a case for collections.defaultDict. But I couldn't make this work where I could create new arrays and append to them.

LMK if you want me to work on this part some more.

alisman: @hweej sorry, i mean _.groupBy

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +125,52 @@ async function fetchFilesLinksData(
...
     );
-    const resourcesForPatientsAndSamples: { [key: string]: ResourceData[] } = {
-        ...sampleIdResourceData,
-        ...resourcesForPatients,
-    };
+
+    const resourcesForPatientsAndSamples = resourcesForEntireStudy
+        // Filter the resources to consist of only studyView selected samples
+        // Also keep patient level resources (e.g. Those don't have a sampleId)
+        .filter(
```
alisman, line 143: one more thing here:  this is operation is R*samples+R*patients.  could easily get to billion.    easily optimized by putting selectedSamples in a map so you can use IN operator.

hweej: I have replaced this with Maps instead of arrays to remove the linear lookup. I don't think you can use the IN operator with Maps so we're opting for the .has('key') expression instead.

src/pages/studyView/resources/FilesAndLinks.tsx
hweej: I have replaced these lines with a single lodash groupBy('patientId'). Thanks @alisman !

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +125,29 @@ async function fetchFilesLinksData(
...
-        ...resourcesForPatients,
-    };
+
+    // Filter the resources to consist of only studyView selected samples
+    // Also keep patient level resources (e.g. Those don't have a sampleId)
+    const resourcesForPatientsAndSamples = _(resourcesForEntireStudy)
+        .filter(resource =>
+            selectedIds.has(resource.sampleId || resource.patientId)
+        )
+        .groupBy('patientId')
```
alisman, line 149: @hweej one more little tweak.  i'm pretty sure that using the shortcut 'patientId' is not typesafe., so as a rule we use 
```
(r)=>r.patientId)
```
That way  if there is no propertyId, ts will catch it.

src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -112,13 +125,29 @@ async function fetchFilesLinksData(
...
 
-    const resourcesForPatients = await getResourceDataOfPatients(
-        studyClinicalDataResponse.data
+    const selectedStudyIds = [
+        ...new Set(selectedSamples.map(item => item.studyId)),
+    ];
+
+    // sampleIds (+patientIds) for the selectedSamples
+    const selectedIds = new Map([
+        ...new Map(selectedSamples.map(item => [item.sampleId, item.studyId])),
```
alisman, line 134: this is strange.  why instantiate maps inside of the map?    i think just 
```
   new Map([...selectedSamples.map(xxx)])
```

PR #5038, opened 2024-11-04
Modify fetch resources from single study to multi-study

hweej:
Fix https://github.com/cBioPortal/cbioportal/issues/11158. Add multiy-study result for files & resource tab on Study view. Fetch resource data for each study on the frontend and combine the results. Note: in the future we might want to improve the endpoint to handle multiple studies directly

### DIFF
--- a/src/pages/studyView/resources/FilesAndLinks.tsx
+++ b/src/pages/studyView/resources/FilesAndLinks.tsx
@@ -35,14 +35,12 @@ const RECORD_LIMIT = 500;
 
 function getResourceDataOfEntireStudy(studyIds: string[]) {
     // Fetch resource data for each studyId, then return combined results
-    const allResources = _(studyIds)
-        .map(studyId =>
-            internalClient.getAllStudyResourceDataInStudyPatientSampleUsingGET({
-                studyId: studyId,
-                projection: 'DETAILED',
-            })
-        )
-        .value();
+    const allResources = studyIds.map(studyId =>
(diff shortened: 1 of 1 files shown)


src/pages/studyView/resources/FilesAndLinks.tsx
```diff
@@ -34,16 +34,21 @@ class FilesLinksTableComponent extends LazyMobXTable<{
...
-    // Only handle the first studyId for now. Can be expanded to make a call per
-    // studyId.
-    const studyId = studyIds[0];
-    const allResources = internalClient.getAllStudyResourceDataInStudyPatientSampleUsingGET(
-        {
-            studyId: studyId,
-            projection: 'DETAILED',
-        }
+    // Fetch resource data for each studyId, then return combined results
+    const allResources = _(studyIds)
```
alisman, line 38: the js array natively has map on it so you don't need to wrap in lodash here.  this might seem confusing given what i said last week but the real message is that we should strive to use the semantically accurate and descriptive method.  if it's just map, then no reason to use lodash.   the other reason to use lodash is it's support for chaining, but we're not doing that here.

alisman (approved):

src/pages/studyView/resources/FilesAndLinks.tsx
hweej: Okay no problem I can change to the native map. I was just going off of what the previous author in another function in this same file.

inodb (approved):
Looks like it's working from product perspective, thanks @hweej !

alisman (approved):

PR #5044, opened 2024-11-13
Enable request body compression on column store api requests

dippindots:
Fix cBioPortal/cbioportal#11116


dippindots:
@alisman I'm merging this pr to create a new frontend deployment, I will revert this if error occurs

PR #5061, opened 2024-12-11
Change file and links name to resource displayName

hweej:
Adds a condition to the link text where if all resources are of the same type, then rename the Files & Links tab to the resourceDefinition displayName

### DIFF
--- a/src/pages/studyView/StudyViewPage.tsx
+++ b/src/pages/studyView/StudyViewPage.tsx
@@ -721,7 +721,13 @@ export default class StudyViewPage extends React.Component<
                                         id={
                                             StudyViewPageTabKeyEnum.FILES_AND_LINKS
                                         }
-                                        linkText={RESOURCES_TAB_NAME}
+                                        linkText={
+                                            this.store.resourceDefinitions
+                                                .result?.length == 1
+                                                ? this.store.resourceDefinitions
+                                                      .result[0].displayName
+                                                : RESOURCES_TAB_NAME
+                                        }
                                         hide={!this.shouldShowResources}
                                     >
                                         <div>


inodb (approved):
LGTM! Thanks so much! Tested SPECTRUM and PCAWG studies

src/pages/studyView/StudyViewPage.tsx
```diff
@@ -721,7 +721,13 @@ export default class StudyViewPage extends React.Component<
...
                                             StudyViewPageTabKeyEnum.FILES_AND_LINKS
                                         }
-                                        linkText={RESOURCES_TAB_NAME}
+                                        linkText={
+                                            this.store.resourceDefinitions
+                                                .result?.length == 1
+                                                ? this.store.resourceDefinitions
+                                                      .result[0].displayName
+                                                : RESOURCES_TAB_NAME
+                                        }
```
inodb, line 730: how does this work when there is a resource definition for a study resource (e.g. rather than a sample or patient resource)? I guess, technically, it might show up as two tabs with the same name? Maybe ok to not handle this case but just curious

alisman: @inodb @hweej the fact that the content of this is this.openResource suggests it can contain many things no?  So to just choose the first resourceDefinition doesn't seem right.

alisman: ah i see.  it's if the length is one.

alisman: @hweej i think this will have behavior where tab will have text RESOURCE_TAB_NAME while the resource definitions are loading and then the name will switch.  We should probably just the tab until it finishes loading.  see hide propery on other tabs in this collection

inodb: @alisman it looks like the `hide` property already waits for the definition: https://github.com/hweej/cbioportal-frontend/blob/ae9281eff50a9ff7241393de154970cf8c32b3c5/src/pages/studyView/StudyViewPage.tsx#L380-L386. I think that's what you mean right?

inodb: Let's maybe merge this and come back to the `hide` functionality if it is indeed an issue

PR #5066, opened 2024-12-14
Fix/10733: Updated default value of Timeline events availability chart

ShahKirtan2003:
Fix cBioPortal/cbioportal#10733 

- As described in the issue, modified the default value of the Timeline events availability chart and not show that table by default, and let users add whenever they want from the charts menu.

image
image

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -7834,10 +7834,6 @@ export class StudyViewPageStore
     @action
     initializeClinicalEventTypeCountChart(): void {
         if (this.shouldDisplayClinicalEventTypeCounts.result) {
-            this.changeChartVisibility(
-                SpecialChartsUniqueKeyEnum.CLINICAL_EVENT_TYPE_COUNTS,
-                false
-            );
             this.chartsType.set(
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -7816,7 +7816,7 @@ export class StudyViewPageStore
         if (this.shouldDisplayClinicalEventTypeCounts.result) {
             this.changeChartVisibility(
                 SpecialChartsUniqueKeyEnum.CLINICAL_EVENT_TYPE_COUNTS,
-                true
+                false
             );
```
inodb, line 7820: @ShahKirtan2003 wondering if this whole block can be removed? I guess the default is to hide the chart?

kirtan-kdu: Yes @inodb, removing the changeChartVisibility block is the same as setting `false` value. Addressed your suggestion in latest commit.

ShahKirtan2003: Yes @inodb, removing the changeChartVisibility block is the same as setting false value. Addressed your suggestion in latest commit.

PR #5070, opened 2024-12-19
#11285 Fixed issue of server config properties not getting rendered correctly

arishta:

### Problem: 

The server config API returns all values as strings, including boolean values (like 'true' or 'false'). Our current code only converts those strings to actual boolean values if they exist in our `ServerConfigDefaults` file. This caused a few boolean values like `skin_show_donate_button` to be rendered as string: `true`. 

### Solution:

Added a second pass after applying defaults that looks at all config values from the API and converts any 'true'/'false' strings to proper boolean values.

Screenshot 2024-12-19 at 10 30 50 AM

## Checks

## Any screenshots or GIFs?

## Notify reviewers

### DIFF
--- a/src/config/config.ts
+++ b/src/config/config.ts
@@ -96,6 +96,20 @@ function applyDefaultConfigurationValues(serverConfig: any) {
             }
         }
     });
+
+    // This pass handles boolean values that aren't in ServerConfigDefaults
+    // Without this, these boolean properties will be rendered as string
+    Object.keys(serverConfig).forEach(key => {
+        if (typeof serverConfig[key] === 'string') {
+            const lowerVal = serverConfig[key].toLowerCase();
+            if (lowerVal === 'true' || lowerVal === 'false') {
+                serverConfig[key] = lowerVal === 'true';
+            }
+            console.warn(
+                `Warning: Boolean value for key "${key}" encountered, but it's missing from ServerConfigDefaults.`
+            );
+        }
+    });
 }
 
 export class ServerConfigHelpers {


arishta-dev:
This PR fixes the issue:
https://github.com/cBioPortal/cbioportal/issues/11285

src/config/config.ts
```diff
@@ -94,6 +94,17 @@ function applyDefaultConfigurationValues(serverConfig: any) {
             }
         }
     });
+
+    // This pass handles boolean values that aren't in ServerConfigDefaults
+    // Without this, these boolean properties will be rendered as string
+    Object.keys(serverConfig).forEach(key => {
```
alisman, line 100: @arishta-dev might be nice to console a warning message when we encounter such a case because we want to add missing items to the defaults.

src/config/config.ts
```diff
@@ -94,6 +94,17 @@ function applyDefaultConfigurationValues(serverConfig: any) {
...
+        if (typeof serverConfig[key] === 'string') {
+            const lowerVal = serverConfig[key].toLowerCase();
```
alisman, line 102: no big deal but i like to get people to use regexp in such cases.  e.g.
```
/true|false/i.test(serverConfig[key])
```
now you don't have to mess around with lowercasing and you don't have to do null/undefined check.

inodb:
@arishta-dev Thanks so much for all your work! As mentioned on slack this works but was thinking a cleaner solution might be to fix this in the backend code directly (ensuring it generates JSON boolean properties correctly): https://github.com/cBioPortal/cbioportal/blob/be86bae2829f86cf9aeb091d60f582ce2a5f1110/src/main/resources/webapp/config_service.jsp#L145-L150

arishta-dev:
Hi @inodb, from what I understand, the property values are being fetched from various sources (application.properties, env variables, etc) and then being returned as `Map<String, String> ` in the `config_service` API.
To achieve what you suggested, we can change the boolean data types to Object and make the API return `Map<String, Object>`. 
Please let me know how this approach looks, if it looks good, I will put up a different PR in `cbioportal` repo.

```
    private Map<String, Object> getFrontendProperties(HttpServletRequest request, Authentication authentication) {
    String baseUrl = requestUtils.getBaseUrl(request);
    Map<String, Object> properties = new HashMap<>();
    
    Map<String, String> originalProperties = frontendPropertiesService.getFrontendProperties();

    for (Map.Entry<String, String> entry : originalProperties.entrySet()) {
        String value = entry.getValue();
        if (value.equalsIgnoreCase("true") || value.equalsIgnoreCase("false")) {
            properties.put(entry.getKey(), Boolean.valueOf(value));
        } else {
            properties.put(entry.getKey(), value);
        }
    }

```

src/config/config.ts
arishta-dev: Done

src/config/config.ts
arishta-dev: Done

arishta-dev:
Hi @alisman I have addressed your review comments, but @inodb is suggesting an alternate approach to make the API return the correct response instead of processing it in the frontend, please TAL at my comment, if it looks good, will go with this approach. Thanks! 
https://github.com/cBioPortal/cbioportal-frontend/pull/5070#issuecomment-2558042386

inodb:
Thanks so much @arishta-dev ! That makes sense to me, but I'll double check with @alisman regarding the approach on Monday

inodb:
@arishta apologies about the delay. Let's try 1 if possible - thank you!

arishta:
Hi @inodb @alisman 
I have put up another PR for addressing the backend change, please review that. 
I am marking this one as closed. 
Link to the backend PR: https://github.com/cBioPortal/cbioportal/pull/11323

PR #5076, opened 2025-01-02
update screenshost after gene table update

inodb:
Pretty much [all changes ](https://output.circle-artifacts.com/output/job/95bd19f2-0b4e-4cc5-beab-bb20a9ec0996/artifacts/0/imageCompare.html) are expected due to changes to the gene table (just igv one is flaky so didn't include it)

## Notes
- We went from 19,212 to 19,219 genes. This causes all changes in the q-values
- Re-importing of variants changes sort order on some pages (e.g. mutations table)
- Filed a ticket for the tumor type color changes (https://github.com/cBioPortal/cbioportal/issues/11321). It looks fine, but because the coloring is arbitrary it is hard to verify


zainasir (approved):
LGTM!

PR #5079, opened 2025-01-03
Fix log scale breaking box plot in plots tab

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11237


inodb (approved):
Nice work ! Extra kudos for the test! Thank you!

PR #5084, opened 2025-01-14
Update function logScalePossible for generic assay negative data

TJMKuijpers:
This PR fixes the bug describe in this [issue](https://github.com/cBioPortal/cbioportal/issues/11334).

Now, when we select generic assay data with negative values, the logScalePossible does return `false` whereas before it would always return `true`. 

Before:
image
Now:
image

### DIFF
--- a/src/shared/components/plots/PlotsTabUtils.spec.ts
+++ b/src/shared/components/plots/PlotsTabUtils.spec.ts
@@ -767,7 +767,18 @@ describe('PlotsTabUtils', () => {
     });
 
     describe('logScalePossible', () => {
-        it('should return true when data is type number', () => {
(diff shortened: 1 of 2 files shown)


TJMKuijpers:
@inodb if you want to recreate the issue on the public portal, this would be the steps:
` Go to study gbm_cptac_2021 -> query ERBB2 -> go to plot, switch y axis to “mRNA expression FPKM UQ”, note that the “log scale” is not checked -> switch to “Z score of mRNA expression” -> switch back to “mRNA expression FPKM UQ”, note that the “log scale” is now checked -> switch to “Protein level - Phospoprotein abundance ratio”, now the chart fails.`

I think the interesting part here is that this only occurs when you have select mRNA first.

PR #5090, opened 2025-01-23
Fix comparison mutations tab not loading when no mutations selected

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/10499. Changed the logic such that genomic alterations tab selections,  do not affect Mutations Tab

With selection like this on genomics alteration tab:
image

Mutations Tab still shows (previously it would throw an error):
image


src/pages/groupComparison/GroupComparisonMutationMapperWrapper.tsx
```diff
@@ -151,18 +151,21 @@ export default class GroupComparisonMutationMapperWrapper extends React.Componen
...
             if (mutationMapperStore) {
                 return (
                     <div
                         data-test="ComparisonPageMutationsTabPlot"
                         style={{ marginTop: 20 }}
                     >
                         <h3>
-                            {this.props.store.activeMutationMapperGene
-                                ?.hugoGeneSymbol +
+                            {this.props.store.activeMutationMapperGene!
```
alisman, line 167: maybe better to combine the condition at 160 with the condition at 155.  then you wouldn't have to do this assertion.

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -55,8 +56,8 @@ export default class GroupComparisonMutationsTab extends React.Component<
         let activeTabId;
         if (this.props.store.userSelectedMutationMapperGene) {
             activeTabId = this.props.store.userSelectedMutationMapperGene;
-        } else {
-            activeTabId = this.props.store.activeMutationMapperGene!
+        } else if (this.props.store.activeMutationMapperGene) {
```
alisman, line 59: is it possible for actigeMutationMapperGene and userSelectedMutationMapperGene to differ?

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -368,6 +368,9 @@ export default class GroupComparisonStore extends ComparisonStore {
     readonly mutations = remoteData({
         await: () => [this.samples, this.mutationEnrichmentProfiles],
         invoke: async () => {
+            if (!this.activeMutationMapperGene) {
```
alisman, line 371: this is kind of awkward.  ideally, we would never reference this if there wasn't an activeMutationMapperGene.  that's the way our code should be constructed

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -482,15 +486,24 @@ export default class GroupComparisonStore extends ComparisonStore {
...
-            this.genes.result!.find(
-                g => g.hugoGeneSymbol === this.userSelectedMutationMapperGene
-            ) ||
-            this.genes.result!.find(
-                g =>
-                    g.hugoGeneSymbol ===
-                    this.genesSortedByMutationFrequency.result![0]
-            );
+        let gene;
+        if (this.genes.isComplete) {
```
alisman, line 490: we should be checking in the view to make sure that gense is complete before we render any component which would be dependent on it.   this is an anti-pattern that we have a lot in the app unfortunately.

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
gblaih: yes there is a scenario where the userSelectedMutationMapperGene is undefined. i just changed this though as they should really be the same

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -430,14 +430,15 @@ export default class GroupComparisonStore extends ComparisonStore {
...
-        invoke: () => {
-            return Promise.resolve(
-                getCoverageInformation(
-                    this.genePanelDataForMutationProfiles.result!,
-                    this.sampleKeyToSample.result!,
-                    this.patients.result!,
-                    [this.activeMutationMapperGene!]
-                )
+        invoke: async () => {
+            if (!this.activeMutationMapperGene) {
```
alisman, line 434: it's like, really what we need is another property on remoteData which will not fure this unless some condition is met.

src/pages/groupComparison/GroupComparisonMutationMapperWrapper.tsx
```diff
@@ -149,20 +149,25 @@ export default class GroupComparisonMutationMapperWrapper extends React.Componen
...
+            this.props.store.genes,
+            this.props.store.genesSortedByMutationFrequency,
         ],
         render: () => {
-            const mutationMapperStore = this.mutationMapperToolStore.getMutationMapperStore(
-                this.props.store.activeMutationMapperGene!.hugoGeneSymbol
-            );
-            if (mutationMapperStore) {
+            let mutationMapperStore;
+            if (
```
alisman, line 157: i think what it should be is 
```
render:()=> {
   if (this.props.store.activeMutationMapperGene.hugoGeneSymbol){
      const mutationMapperStore = ...
      return ()
   }

}
```

src/pages/groupComparison/GroupComparisonMutationsTab.tsx
```diff
@@ -101,6 +100,15 @@ export default class GroupComparisonMutationsTab extends React.Component<
...
                 );
+            } else if (!this.props.store.activeMutationMapperGene) {
+                return (
+                    <div
+                        className="alert alert-info"
+                        data-test="NoMutationsAvailableAlert"
+                    >
+                        {NO_MUTATIONS_AVAILABLE_MSG}
+                    </div>
+                );
```
alisman, line 111: lets just finish with else {} instead of closing the else and then returning something

src/pages/groupComparison/GroupComparisonStore.ts
```diff
@@ -339,8 +339,8 @@ export default class GroupComparisonStore extends ComparisonStore {
                                 isSampleProfiled(
                                     sample.uniqueSampleKey,
                                     p.molecularProfileId,
-                                    this.activeMutationMapperGene!
-                                        .hugoGeneSymbol,
+                                    this.activeMutationMapperGene
+                                        ?.hugoGeneSymbol,
```
alisman, line 343: i think you want to assert that yes there is a hugoGeneSymbol!   

What the ! means is that you the developer certify that there will be a hugoGeneSymbol.  maybe add a condition above for the negative case.

src/pages/groupComparison/GroupComparisonMutationMapperWrapper.tsx
gblaih: mutationMapperStore can be undefined even if the hugoGeneSymbol is defined, so this would break. you still have to check if mutationMapperStore is defined or not

inodb (approved):
Looks great! Thank you!

src/pages/groupComparison/GroupComparisonStore.ts
alisman: lets fix this

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -713,32 +718,23 @@ export default abstract class ComparisonStore extends AnalysisStore
...
-                            molecularProfiles => molecularProfiles[0]
-                        )
-                    );
-                } else {
-                    return Promise.resolve(this._mutationEnrichmentProfileMap);
-                }
-            } else {
-                return Promise.resolve({});
-            }
+        invoke: async () => {
```
alisman, line 721: put a comment above explaining that this pertains only to the alterations tab.  or move this to be a property of the tab component.

src/shared/lib/comparison/ComparisonStore.ts
```diff
@@ -235,12 +236,16 @@ export default abstract class ComparisonStore extends AnalysisStore
...
         } else {
             // default
-            return mutationEventTypeSelectInit(
-                this.alterationEnrichmentProfiles.result?.mutationProfiles || []
-            );
+            return this.defaultMutationEnrichmentEventTypes;
         }
     }
 
+    @computed get defaultMutationEnrichmentEventTypes() {
```
alisman, line 243: lets not call this this default since it is only meant distinguish from the table level.

alisman: mutationEnrichmentEventTypes

PR #5103, opened 2025-02-11
Add last updated time and embolden exact matches in cosmic column in mutations table

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/10973

It's been decided that we want to keep the current implementation of showing all mutations at a particular residue rather than the exact matching one. To make it more clear, exact matches are embolden in the COSMIC tooltip overlay and update time is added

Screenshot 2025-02-11 at 5 05 25 PM


src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
```diff
@@ -132,7 +184,8 @@ export default class CosmicColumnFormatter {
                 <span className={styles['cosmic-table']}>
                     <b>{value}</b> occurrences of <b>{cosmic[0].keyword}</b>{' '}
                     mutations in COSMIC
-                    <CosmicMutationTable data={cosmic} />
+                    <CosmicMutationTable data={cosmic} columns={columns} />
+                    <b>Last Updated: Before 2017</b>
```
inodb, line 188: @gblaih not sure if there's some hacky way to see if the data is different from what we have in the db for public/msk portals. It's possible local installers have an updated version of COSMIC

src/shared/components/mutationTable/column/CosmicColumnFormatter.tsx
```diff
@@ -113,10 +118,66 @@ export default class CosmicColumnFormatter {
...
+
         let value: number = -1;
         let display: string = '';
         let overlay: (() => JSX.Element) | null = null;
         let content: JSX.Element;
+        let lastUpdatedText: JSX.Element;
+
+        if (
+            getServerConfig().app_name === 'public-portal' ||
+            getServerConfig().app_name === 'mskcc-portal'
```
inodb, line 177: @gblaih do you see a more uniform way of doing this in the code somewhere? I e.g. saw this:

https://github.com/cBioPortal/cbioportal-frontend/blob/dd4d311eb36138b649e2f44f4e0199d19a587f84/src/shared/components/query/QueryContainer.tsx#L125

Maybe that should have a `isMSKPortal` function too?

inodb: https://github.com/cBioPortal/cbioportal-frontend/blob/ee88b3ef5145206539a332470db99537c5cac666/src/AppStore.ts#L65

inodb (commented):
@gblaih could you also change the text at the top of the tooltip? Like:

image

onursumer (changes requested):
Looks good! Just requested a few minor changes.

onursumer (approved):

PR #5110, opened 2025-02-20
Add sample and patient counts toggle for cancer summary chart tooltip

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11311

Adds option to toggle between sample and patient counts in tooltip of cancer summary chart in results view. Previously, it only showed sample counts with no option to switch.

Screenshot 2025-02-20 at 5 15 42 PM


inodb (changes requested):
Thanks @gblaih ! The counts look good, but would be nice to update the "cases" text to "samples" when selecting "Sample Counts" (patient and case are often used as synonyms):

image

inodb (changes requested):
Looks great! Thanks @gblaih ! Can you make the default "Patient Counts", similar to how the oncoprint works

src/shared/lib/alterationCountHelpers.ts
```diff
@@ -74,12 +88,22 @@ export function getAlterationCountsForCancerTypesForAllGenes(
...
         [uniqueSampleKey: string]: ExtendedAlteration[];
     };
-    return countAlterationOccurences(
-        samplesByCancerType,
-        merged,
-        molecularProfileIdsByAlterationType,
-        coverageInformation
-    );
+
+    const alterationCounts = countByProperty
```
onursumer, line 92: `countByProperty` is not a boolean so we should do something like this, right?

```suggestion
    const alterationCounts = countByProperty === 'sampleCounts'
```

src/shared/lib/alterationCountHelpers.ts
```diff
@@ -33,13 +34,25 @@ export function getAlterationCountsForCancerTypesByGene(
...
-                samplesByCancerType,
-                alterationsBySampleId,
-                molecularProfileIdsByAlterationType,
-                coverageInformation,
-                gene
-            );
+
+            let alterationCounts =
+                countByProperty === 'sampleCounts'
+                    ? countAlterationOccurences(
```
onursumer, line 40: should we rename this to `countSampleAlterationOccurences`?

src/shared/lib/alterationCountHelpers.ts
```diff
@@ -272,3 +295,217 @@ export function countAlterationOccurences(
         }
     );
 }
+
+export function countPatientAlterationOccurences(
```
onursumer, line 299: this is a pretty long function with nested function calls. would be nice to simplify/flatten this procedure but not sure if there is an easy way to do that.

onursumer (changes requested):

src/shared/lib/alterationCountHelpers.ts
gblaih: this function is based on the also lengthy ```countAlterationOccurences``` function but with some additional logic to account for the fact we are working with patients. i tried to combine these two into one but that proved to be very sloppy and difficult to read. i don't think it would be easy to simplify this right now but maybe can look into making both these functions cleaner in the future

PR #5118, opened 2025-03-03
RFC93: Custom variant annotation filter in study view page

FlorisVleugels:
## Summary

This PR adds the frontend for custom namespace charts and filtering in the study view, see below screenshots for what it looks like for some example namespaces.

### Key Features:  
- A new tab **Variant Annotations** appears on the Charts menu if namespace attributes are detected for studies in the study view. If no namespace attributes are present, the tab remains hidden.
- Add charts based on namespace attributes.  
- Filter samples using **intersect** and **union** filtering based on custom namespaces.  
- Create a **comparison page session** from selected namespace groups.  
- Display name of the namespace attributes is the postfix + between parentheses the prefix if there are multiple namespaces with the same postfix.
- Variant annotations charts and menu tab are populated with namespace information from all three sources (mutations, structural variants, cna)

For more details on the proposal: [RFC93](https://docs.google.com/document/d/1D4zEdLeyVHm5P99Y7YEfIbv0sZCPsE5NkaNTn4YVi8Q/edit?tab=t.0#heading=h.l41h5shjky00).  

❗This PR should be merged at the same time as the corresponding backend PR (see link below).  

## Code Changes
- Added support for namespace endpoints to the cbioportal API package
- Added Variant Annotations charts (including filtering, comparison and download functionality)
- Expanded chart menu to include namespace attributes & frequency when relevant

## 🔗 Related Issues / PRs
[Backend PR](https://github.com/cBioPortal/cbioportal/pull/11432)

## Screenshots
New chart menu tab: 
image

Study view page with Variant Annotation charts: 
image

Filtering the study view by custom namespaces: 
image

Creating a comparison session based of custom namespace annotations: 
image


src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -817,6 +822,28 @@ export function getUniqueKey(attribute: ClinicalAttribute): string {
...
+export function getUniqueNamespaceKey(attribute: NamespaceAttribute): string {
+    return attribute.outerKey.concat('_', attribute.innerKey);
+}
+
+function getNamespaceChartDisplayName(
+    attribute: NamespaceAttribute,
+    namespaceAttributes: NamespaceAttribute[]
+): string {
+    var count = 0;
+    namespaceAttributes.forEach(item => {
```
alisman, line 834: namespaceAttributes.filter(item=>attribute.innerKey===item.innerKey).length

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -817,6 +822,28 @@ export function getUniqueKey(attribute: ClinicalAttribute): string {
...
+        if (item.innerKey === attribute.innerKey) {
+            count++;
+        }
+    });
+
+    if (count > 1) {
```
alisman, line 840: this logic needs some comment to explain

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -817,6 +822,28 @@ export function getUniqueKey(attribute: ClinicalAttribute): string {
...
+        return attribute.innerKey + ' (' + attribute.outerKey + ')';
```
alisman, line 841: not a big deal but you should use js string interpolation to make string building a little more clear.

src/pages/studyView/StudyViewUtils.tsx
```diff
@@ -4241,6 +4270,56 @@ export async function invokeMutationDataCount(
...
+    const data = result.find(
+        d =>
+            d.outerKey === chartMeta.namespaceAttribute!.outerKey &&
+            d.innerKey === chartMeta.namespaceAttribute!.innerKey
+    );
+
+    let counts: MultiSelectionTableRow[] = [];
+    if (data !== undefined) {
+        counts = data.counts.map(c => {
+            if (c.totalCount == undefined) {
```
alisman, line 4303: why do they sometimes not have totalCount?

src/pages/studyView/UserSelections.tsx
```diff
@@ -347,6 +348,61 @@ export default class UserSelections extends React.Component<
             components
         );
 
+        // Namespace chart filters
+        let namespaceFilterComponents: JSX.Element[] = [];
+        if (this.props.filter.namespaceDataFilters) {
+            _.forEach(
```
alisman, line 354: avoid using forEach and accumulating in a predefined array.  instead use map/reduce/filter etc semantic methods and assign result to variable.

alisman:
@FlorisVleugels this PR looks good.  just a few minor code suggestions.   
@inodb @dippindots this PR follows the established conventions of study view charts.  seems uncontroversial from a code perspective.

alisman:
@FlorisVleugels i just noticed that when we load the study page with this PR, we are eagerly firing the new namespace attrributes request before the charts menu has even been opened.  we should only fire it when the user opens the menu.  you need to move the reference to the promise inside the menu .  let me know if you need assistance with this.

src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -417,6 +454,39 @@ export class MultiSelectionTable extends React.Component<
...
+                        {data.totalCount.toLocaleString()}
+                    </span>
+                ),
+                sortBy: (data: MultiSelectionTableRow) => data.totalCount,
+                defaultSortDirection: 'desc' as 'desc',
+                filter: (
+                    data: MultiSelectionTableRow,
+                    filterString: string
+                ) => {
+                    return _.toString(data.totalCount).includes(filterString);
```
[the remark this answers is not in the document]
FlorisVleugels: Yea I agree it is weird, but this lets users search for rows based on the count in the little search box on the chart (see also mutated genes chart works the same)

src/pages/studyView/UserSelections.tsx
FlorisVleugels: refactored this component to use reduce/ map and made it a bit less nested. Followed the same approach as the mutation chart filters

src/pages/studyView/UserSelections.tsx
```diff
@@ -347,6 +348,61 @@ export default class UserSelections extends React.Component<
...
+                this.props.filter.namespaceDataFilters,
+                namespaceDataFilter => {
+                    const uniqueKey = namespaceDataFilter.outerKey.concat(
+                        '_',
+                        namespaceDataFilter.innerKey
+                    );
+                    const chartMeta = this.props.attributesMetaSet[uniqueKey];
+                    if (chartMeta) {
+                        namespaceFilterComponents.push(
```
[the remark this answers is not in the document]
FlorisVleugels: see previous comment

src/pages/studyView/StudyViewUtils.tsx
FlorisVleugels: Thanks, that shouldnt be nessecary as they should always have a totalcount in the response, removed it!

src/pages/studyView/StudyViewUtils.tsx
FlorisVleugels: Thanks for the suggestion, updated it in latest commit

FlorisVleugels:
> @FlorisVleugels i just noticed that when we load the study page with this PR, we are eagerly firing the new namespace attrributes request before the charts menu has even been opened. we should only fire it when the user opens the menu. you need to move the reference to the promise inside the menu . let me know if you need assistance with this.

I didn't move the reference to the namespace-attributes fetch with my latest commit because the current implementation follows the same cycle as the clinical-attributes. 

Current:   
- When the page is loaded the attributes are fetched and some initializations are done. 
- Then only when the user clicks on the chart menutab the counts for the attributes are fetched, 
- and later when they decide to toggle a specific chart the data counts are fetched.

Do you agree with this approach? We should see the same behavior for clinical attributes when we first render a view

PR #5120, opened 2025-03-06
Allow clickhousifying of the public API client

onursumer:
Related to cBioPortal/cbioportal#11377


alisman:
@onursumer thanks for doing this @onursumer.   is this ready to go?

onursumer:
@alisman this works fine as far as I locally tested, but we should probably wait until after we merge cBioPortal/cbioportal/pull/11393, otherwise I think clickhouse mode will try to request data from not-yet-existing endpoints.

src/shared/api/cbioportalClientInstance.ts
```diff
@@ -2,4 +2,22 @@ import { CBioPortalAPI } from 'cbioportal-ts-api-client';
...
+
+const clientColumnStore = new CBioPortalAPI();
+
+export function getClient() {
+    return isClickhouseMode() ? clientColumnStore : client;
+}
+
+overrideApiRequestForColumnStore(clientColumnStore);
+
+proxyColumnStore(clientColumnStore, 'fetchSamples');
```
alisman, line 17: @fuzhaoyuan when you fire up the branch, you add an entry here corresponding to the client method (which corresponds to your endpoint) and then it will call the clickhouse-ified version

PR #5128, opened 2025-03-12
Replace Jitpack with Github Packages

zainasir:
- Replace jitpack with github packages

## Checks

## Any screenshots or GIFs?

## Notify reviewers


inodb (approved):
Looks good! Thanks for fixing!

Are there any places to remove jitpack from the docs?

.github/workflows/github-package-build.yml
```diff
@@ -0,0 +1,54 @@
+name: Build & Publish to GitHub Packages
+
+on:
+  release:
+    types:
+      - published
```
dippindots, line 6: Just wanted to make sure when do we want to build a new package, it seems this will build the package when we have a draft release? Do we only want it to build when we do pre-release?

dippindots (commented):
I left one comment on this, other than that all looks good to me

.github/workflows/github-package-build.yml
zainasir: Github [docs](https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=published#release:~:text=A%20release%2C%20pre%2Drelease%2C%20or%20draft%20of%20a%20release%20was%20published.) say that this event is only triggered when a release, pre-release, or a draft is "published". So this won't be triggered when we create a draft-release. This also will not be triggered when we change a pre-release to a latest-release because that is considered an update, not a publish.

dippindots (approved):
LGTM

PR #5133, opened 2025-03-14
Improve layout of cancer summary page

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11441

Improves layout of cancer summary page in the results view by grouping controls and moving them to the side. Similarities can be found with the plots tab


alisman:
Horizontal axis becomes “Category”

Vertical becomes “Occurence”

Only show the alert if there items which are filtered out (hidden)

PR #5135, opened 2025-03-15
Change NA to Not Profiled in gene-specific CNA chart

lllinging:
Fix cBioPortal/cbioportal#10809

- Changed "NA" to "Not Profiled" for the gene-specific CNA charts in study view.

## Checks

## Any screenshots or GIFs?
image

### DIFF
--- a/src/pages/studyView/StudyViewUtils.tsx
+++ b/src/pages/studyView/StudyViewUtils.tsx
@@ -1901,6 +1901,11 @@ export function getExponent(value: number): number {
 }
 
 export function getCNAByAlteration(alteration: string | number) {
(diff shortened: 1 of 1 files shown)


alisman:
@tmazor I think the NA comes directly from the data.  How do we know whether it actually means not profiled or just status unknown?

tmazor:
I think same as we define profiled elsewhere, wouldn't it be? If I filter to those 55 samples with NA, the Genomic Profile Sample Counts table shows none of them are profiled for CNA. So the list of profiled samples that we use for eg genomic profile table combined with gene panel when available should be enough to say if sample was profiled for that gene and no alteration called vs sample was not profiled for that gene.

PR #5147, opened 2025-03-25
feat : Use of path URL for comparison subtabs

rishisulakhe:
Fix [cBioPortal/cbioportal#9762 ](https://github.com/cBioPortal/cbioportal/issues/9762)

- Use the path of URL instead of query parameter for comparison subtabs

## Checks
- [ ✓] Has tests or has a separate issue that describes the types of test that should be created. If no test is included it should explicitly be mentioned in the PR why there is no test.
- [✓ ] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [✓ ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!

## Any screenshots or GIFs?
Screenshot 2025-03-25 222117

### DIFF
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -330,6 +330,12 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
+        const pathParts = this.pathName.split('/');
+
+        // Handle paths like "/results/comparison/survival"
+        if (pathParts[1] === 'results' && pathParts[2] === 'comparison') {
+            return ResultsViewTab.COMPARISON;
+        }
         const tabInPath = this.pathName.split('/').pop();
         if (tabInPath && tabInPath in oldTabToNewTabRoute) {
             // map legacy tab ids
@@ -390,6 +396,15 @@ export default class ResultsViewURLWrapper
     }
 
     @computed public get comparisonSubTabId() {
+        const pathParts = this.pathName.split('/');
+        if (
+            pathParts.length >= 4 &&
+            pathParts[1] === 'results' &&
+            pathParts[2] === 'comparison' &&
+            pathParts[3]
+        ) {
+            return pathParts[3];
+        }
         return this.query.comparison_subtab || GroupComparisonTab.OVERLAP;
     }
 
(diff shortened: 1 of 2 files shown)


src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -390,6 +396,15 @@ export default class ResultsViewURLWrapper
...
     @computed public get comparisonSubTabId() {
+        const pathParts = this.pathName.split('/');
+        if (
+            pathParts.length >= 4 &&
+            pathParts[1] === 'results' &&
+            pathParts[2] === 'comparison' &&
+            pathParts[3]
+        ) {
+            return pathParts[3];
+        }
```
alisman, line 407: @rishisulakhe i think this could cause problems in portals that sit inside folders e.g. www.example.com/cbioportal/

instead you could use a regexp to get whatever is after the /comparison/[subtab]?

rishisulakhe: @alisman Yes, this will cause issue.

rishisulakhe: @alisman Done, now we are using regex expression for extracting comparison subtab from path URL
Thankyou

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -330,6 +330,9 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
+        if (this.pathName.match(/\/results\/comparison(\/|$)/)) {
```
alisman, line 333: why is this necessary?

alisman:
@rishisulakhe we have to make a backend change for these new paths to work.  it's limitation of our backend that we don't resolve to the SPA page from all paths, but rather have to ennumerate every valid path in backend.  I will look to see whether we can change this.

rishisulakhe:
@alisman I understood this backend problem.
Can i do this for pathway subtabs also?

alisman:
@rishisulakhe yes, please do.  did you open a backend PR?

rishisulakhe:
@alisman
 Actually I am less aware of Java backend of cBioportal. Can you help me to figure out what backend change i have to do for these new paths to work. You can suggest any idea or reference related to this.
 Thankyou

PR #5161, opened 2025-04-16
Stabilize flaky unprofiled samples oncoprint and breadcrumbs tests

gblaih:
Excluding unprofiled samples oncoprint test is flaky due to mouse hovering effects. This attempts to stabilize that. Breadcrumbs test is still flaky, potentially due to reset charts button not being clicked. Added brief pause to let add charts button contents load.

### DIFF
--- a/end-to-end-test/remote/specs/core/screenshot.spec.js
+++ b/end-to-end-test/remote/specs/core/screenshot.spec.js
@@ -28,6 +28,8 @@ async function waitForAndCheckPlotsTab() {
 function runResultsTestSuite(prefix, options = {}) {
     it(`${prefix} render the oncoprint`, async function() {
         await waitForOncoprint();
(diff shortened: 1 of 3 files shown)


dippindots (approved):
LGTM

end-to-end-test/remote/specs/core/screenshot.spec.js
```diff
@@ -28,6 +28,8 @@ async function waitForAndCheckPlotsTab() {
 function runResultsTestSuite(prefix, options = {}) {
     it(`${prefix} render the oncoprint`, async function() {
         await waitForOncoprint();
+        await (await getElement('body')).moveTo({ xOffset: 0, yOffset: 0 }); // move mouse out of the way
```
alisman, line 31: @gblaih seems like this is redundant with checkElementWithMouseDisabled, no?  so i mean, i just question whether this would really stabilize it give that we are already trying the same thing and it doesn't seem to work?

alisman (approved):

PR #5162, opened 2025-04-17
feat : Use of path URL for comparison and pathways subtab

rishisulakhe:
Fix [cBioportal/cBioportal#9762](https://github.com/cBioPortal/cbioportal/issues/9762)

- Use of path URL instead of query parameter for comparison subtabs
- Use of path URL instead of query parameter for pathways subtabs
## Checks
- [✔] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)

## Any screenshots or GIFs?
For comparison subtabs
Screenshot 2025-04-17 195656
For pathways subtabs
Screenshot 2025-04-17 195745

### DIFF
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -330,6 +330,12 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
+        const pathParts = this.pathName.split('/');
+
+        // Handle paths like "/results/comparison/survival"
+        if (pathParts[1] === 'results' && pathParts[2] === 'comparison') {
+            return ResultsViewTab.COMPARISON;
+        }
         const tabInPath = this.pathName.split('/').pop();
         if (tabInPath && tabInPath in oldTabToNewTabRoute) {
             // map legacy tab ids
@@ -390,6 +396,15 @@ export default class ResultsViewURLWrapper
     }
 
     @computed public get comparisonSubTabId() {
+        const pathParts = this.pathName.split('/');
+        if (
+            pathParts.length >= 4 &&
+            pathParts[1] === 'results' &&
+            pathParts[2] === 'comparison' &&
+            pathParts[3]
+        ) {
+            return pathParts[3];
+        }
         return this.query.comparison_subtab || GroupComparisonTab.OVERLAP;
     }
 
@@ -400,7 +415,7 @@ export default class ResultsViewURLWrapper
 
     @autobind
     public setComparisonSubTabId(tabId: GroupComparisonTab) {
-        this.updateURL({ comparison_subtab: tabId });
+        this.updateURL({}, `results/comparison/${tabId}`);
     }
 
     @computed public get selectedEnrichmentEventTypes() {
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -406,9 +406,11 @@ export default class ResultsViewURLWrapper
         this.updateURL({}, `results/${tabId}`, false, replace);
     }
 
-    public get getSubTab() {
+    public get subTab() {
         const tabSegment = this.getCurrentTabSegment();
-        if (!tabSegment) return '';
+        if (!tabSegment) {
+            return '';
+        }
 
         // Extract subtab name after "results/tabSegment/" in the URL path
         const subtabMatch = this.pathName.match(
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -336,15 +336,11 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
-        // Ensure top-level tab is correctly identified for routes with static subtabs.
-        // This condition checks if the URL matches exactly "/results/comparison"
-        if (this.pathName.match(/\/results\/comparison(\/|$)/)) {
-            return ResultsViewTab.COMPARISON;
-        }
-
-        // Similarly, this checks for the "/results/pathways" route.
-        if (this.pathName.match(/\/results\/pathways(\/|$)/)) {
-            return ResultsViewTab.PATHWAYS;
+        const tabSegment = this.getCurrentTabSegment();
+        if (tabSegment) {
+            return ResultsViewTab[
+                tabSegment.toUpperCase() as keyof typeof ResultsViewTab
+            ];
         }
         const tabInPath = this.pathName.split('/').pop();
         if (tabInPath && tabInPath in oldTabToNewTabRoute) {
@@ -405,42 +401,37 @@ export default class ResultsViewURLWrapper
         );
     }
 
-    @computed public get comparisonSubTabId() {
-        // Extract subtab name after "/comparison/" in the URL path.
-        const subtabMatch = this.pathName.match(/\/comparison\/([^\/?#]+)/);
-        return subtabMatch ? subtabMatch[1] : GroupComparisonTab.OVERLAP;
-    }
-
     @autobind
     public setTabId(tabId: ResultsViewTab, replace?: boolean) {
         this.updateURL({}, `results/${tabId}`, false, replace);
     }
 
-    @autobind
-    public setComparisonSubTabId(tabId: GroupComparisonTab) {
-        this.updateURL({}, `results/comparison/${tabId}`);
-    }
+    public get getSubTab() {
+        const tabSegment = this.getCurrentTabSegment();
+        if (!tabSegment) return '';
 
-    @computed public get pathwaysSubTabId() {
-        // Extract subtab name after "/pathways/" in the URL path
-        const subtabMatch = this.pathName.match(/\/pathways\/([^\/?#]+)/);
-        if (!subtabMatch) {
-            return ResultsViewPathwaysSubTab.PATHWAY_MAPPER;
-        }
-
-        const urlValue = subtabMatch[1];
-        // Matching enum by comparing sanitized versions
-        const match = Object.values(ResultsViewPathwaysSubTab).find(
-            enumValue => sanitizeForUrl(enumValue) === urlValue
+        // Extract subtab name after "results/tabSegment/" in the URL path
+        const subtabMatch = this.pathName.match(
+            new RegExp(`\/results\/${tabSegment}\/([^\/?#]+)`)
         );
-
-        return match || ResultsViewPathwaysSubTab.PATHWAY_MAPPER;
+        return subtabMatch?.[1] || '';
     }
 
     @autobind
-    public setPathwaysSubTabId(tabId: ResultsViewPathwaysSubTab) {
-        const sanitizedValue = sanitizeForUrl(tabId);
-        this.updateURL({}, `results/pathways/${sanitizedValue}`);
+    public setSubTab(
+        subtab: ResultsViewComparisonSubTab | ResultsViewPathwaysSubTab
+    ) {
+        const tabSegment = this.getCurrentTabSegment();
+        if (tabSegment) {
+            const sanitized = sanitizeForUrl(subtab);
+            this.updateURL({}, `results/${tabSegment}/${sanitized}`);
+        }
+    }
+
+    private getCurrentTabSegment(): string | undefined {
+        // Extract tab name after "/results/" in the URL path
+        const match = this.pathName.match(/\/results\/([^\/]+)/);
+        return match?.[1];
     }
 
     @computed public get selectedEnrichmentEventTypes() {
--- a/src/pages/resultsView/ResultsViewURLWrapper.ts
+++ b/src/pages/resultsView/ResultsViewURLWrapper.ts
@@ -330,10 +330,7 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
-        const pathParts = this.pathName.split('/');
-
-        // Handle paths like "/results/comparison/survival"
-        if (pathParts[1] === 'results' && pathParts[2] === 'comparison') {
+        if (this.pathName.match(/\/results\/comparison(\/|$)/)) {
             return ResultsViewTab.COMPARISON;
         }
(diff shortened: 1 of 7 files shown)


src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -330,6 +331,12 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
+        if (this.pathName.match(/\/results\/comparison(\/|$)/)) {
```
onursumer, line 334: let's add a comment here to explain what we are trying to match

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -390,7 +397,8 @@ export default class ResultsViewURLWrapper
     }
 
     @computed public get comparisonSubTabId() {
-        return this.query.comparison_subtab || GroupComparisonTab.OVERLAP;
+        const subtabMatch = this.pathName.match(/\/comparison\/([^\/?#]+)/);
```
onursumer, line 400: similar to above, let's add a comment here to explain what we are trying to match

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -400,7 +408,19 @@ export default class ResultsViewURLWrapper
 
     @autobind
     public setComparisonSubTabId(tabId: GroupComparisonTab) {
-        this.updateURL({ comparison_subtab: tabId });
+        this.updateURL({}, `results/comparison/${tabId}`);
+    }
+
+    @computed public get pathwaysSubTabId() {
+        const subtabMatch = this.pathName.match(/\/pathways\/([^\/?#]+)/);
```
onursumer, line 415: let's add a comment here as well to explain what we are trying to match

src/routes.tsx
```diff
@@ -378,6 +375,60 @@ export const makeRoutes = () => {
...
                 />
+                <Route
+                    path="/results/comparison/:subtab"
+                    component={LocationValidationWrapper(
+                        ResultsViewPage,
+                        tabParamValidator(ResultsViewComparisonSubTab),
+                        ResultsViewQueryParamsAdjuster
+                    )}
+                />
+                <Route
```
onursumer, line 386: This one is for backward compatibility where we can still use the query parameter, right? Maybe add a comment to indicate that

src/routes.tsx
```diff
@@ -378,6 +375,60 @@ export const makeRoutes = () => {
...
+                    path="/results/pathways/:subtab"
+                    component={LocationValidationWrapper(
+                        ResultsViewPage,
+                        tabParamValidator(ResultsViewPathwaysSubTab),
+                        ResultsViewQueryParamsAdjuster
+                    )}
+                />
+                <Route
```
onursumer, line 413: Looks like this one is also for backward compatibility. Let's add a comment to indicate that.

onursumer (commented):
Looks good! Let's add a few comments if possible (see my review comments)

src/routes.tsx
rishisulakhe: Yes, this is for backward compatibility where we can still use the old query parameter format while supporting the new URL path format.

src/routes.tsx
rishisulakhe: Done

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done

onursumer:
> Hi @onursumer I have added the comments.
> 
> I can see that for pathway subtab URLs will looks like
> 
> results/pathways/PathwayMapper (fine) results/pathways/NDEx%20Cancer%20Pathways (ugly with encoded spaces)
> 
> We can change these to URL-friendly slugs like below.
> 
> results/pathways/pathway-mapper results/pathways/ndex
> 
> Pls suggest what can we do now?
> 
> Thankyou

Thanks @rishisulakhe! I think URL friendly paths would be better if it is easy to implement

rishisulakhe:
@onursumer 
Now URL looks good for pathways subtabs.
Screenshot 2025-04-22 202826
Screenshot 2025-04-22 202804
Thankyou

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -330,6 +333,16 @@ export default class ResultsViewURLWrapper
     pathContext = '/results';
 
     @computed public get tabId() {
+        // Ensure top-level tab is correctly identified for routes with static subtabs.
+        // This condition checks if the URL matches exactly "/results/comparison"
+        if (this.pathName.match(/\/results\/comparison(\/|$)/)) {
+            return ResultsViewTab.COMPARISON;
```
alisman, line 339: lets generalize so that we have a map 
{
   "comparison":ResultsViewTab.COMPARISON,
   "pathways": ResultsVIewTab.PATHWAYS
}

alisman:
@rishisulakhe @onursumer i think the slug thing is too messy/verbose.   Lets think of something simpler.

rishisulakhe:
@alisman @onursumer 
Yes, the slug thing makes it messy. 
One simpler way is that we can make changes to pathways tab enum values but it effects many other things  OR
Instead of maintaining mapping objects, we could just convert enum values to lowercase and replace spaces with hyphens for URLs. This would give us clean URLs like "ndex-cancer-pathways" while still using the original enum values internally.
Would these approaches work better? I'm happy to discuss other options if you have something specific in mind.

alisman:
@rishisulakhe yes i think the run time sanitization approach is good.   i.e. compute the clean slugs rather than maintain mappers

rishisulakhe:
@alisman 
This implementation uses a simple sanitization function that converts enum values to lowercase and replaces spaces with hyphens.
Let me know if you'd like me to make any adjustments to this implementation.
Thankyou

src/pages/resultsView/ResultsViewURLWrapper.ts
alisman: @rishisulakhe last one i'ld you to address

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -400,7 +418,29 @@ export default class ResultsViewURLWrapper
 
     @autobind
     public setComparisonSubTabId(tabId: GroupComparisonTab) {
-        this.updateURL({ comparison_subtab: tabId });
+        this.updateURL({}, `results/comparison/${tabId}`);
```
alisman, line 421: @rishisulakhe lest just take this opportunity to clean this stuff up.  we should have one getSubTab method (doens't need to be computed) which is able to use current tab state to extract the subtab.    i.e. subtab = thing after the current tab in the pathname.  then we can get rid of this non-generic stuff (e.g. pathwaysSubTabId()) and everything will be clean and extensible.

alisman: i fully realize it was this way prior to your PR but always a good time to improve the codebase.

rishisulakhe: @alisman Yes this is a good idea. Will make commit soon.

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done. Please let me know if you'd like any adjustments to this approach.

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -389,18 +401,37 @@ export default class ResultsViewURLWrapper
...
-    @computed public get comparisonSubTabId() {
-        return this.query.comparison_subtab || GroupComparisonTab.OVERLAP;
-    }
-
     @autobind
     public setTabId(tabId: ResultsViewTab, replace?: boolean) {
         this.updateURL({}, `results/${tabId}`, false, replace);
     }
 
+    public get getSubTab() {
```
onursumer, line 409: the function name should not repeat `get`

```suggestion
    public get subTab() {
```

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -389,18 +401,37 @@ export default class ResultsViewURLWrapper
...
+        const tabSegment = this.getCurrentTabSegment();
+        if (!tabSegment) return '';
```
onursumer, line 411: minor thing but I think we should avoid one-liner if statements 

```suggestion
        if (!tabSegment) {
            return '';
        }
```

onursumer (changes requested):

src/pages/resultsView/ResultsViewURLWrapper.ts
alisman: @rishisulakhe looks good! thanks.  so lets make Onur's small changes and then i'll test once more and we'll be good to go!!! thanks for all your work

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done.

src/pages/resultsView/ResultsViewURLWrapper.ts
rishisulakhe: Done

rishisulakhe:
Thankyou @alisman @onursumer 
Happy to work on this issue!

gblaih:
@rishisulakhe i found some issues with some of the main tabs while testing this. it seems like the `Cancer Types Summary`, `Mutual Exclusivity` and `CN Segments` tabs are broken. can you take a look and check if any other tabs were affected? as an example if you go [here](https://www.cbioportal.org/results/oncoprint?cancer_study_list=ov_tcga_pub&cancer_study_id=ov_tcga_pub&genetic_profile_ids_PROFILE_MUTATION_EXTENDED=ov_tcga_pub_mutations&Z_SCORE_THRESHOLD=2.0&case_set_id=ov_tcga_pub_3way_complete&gene_list=BRCA1%20BRCA2&gene_set_choice=user-defined-list&comparison_subtab=survival) and select those tabs, they do not work

rishisulakhe:
Hi @gblaih 
Now all the tabs work fine.
image
Thankyou

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -389,18 +407,39 @@ export default class ResultsViewURLWrapper
...
-    @computed public get comparisonSubTabId() {
-        return this.query.comparison_subtab || GroupComparisonTab.OVERLAP;
-    }
-
     @autobind
     public setTabId(tabId: ResultsViewTab, replace?: boolean) {
         this.updateURL({}, `results/${tabId}`, false, replace);
     }
 
+    public get subTab() {
```
onursumer, line 415: @alisman @rishisulakhe Shoud we extract the logic in this function, and also `getCurrentTabSegment` (and maybe even `tabId`) into utility functions and write some unit tests? I feel like these functions became complex enough to add at least some basic unit testing

src/pages/resultsView/ResultsViewURLWrapper.ts
```diff
@@ -389,18 +407,39 @@ export default class ResultsViewURLWrapper
...
+            const sanitized = sanitizeForUrl(subtab);
+            this.updateURL({}, `results/${tabSegment}/${sanitized}`);
+        }
+    }
+
+    private getCurrentTabSegment(): string | undefined {
+        // Extract tab name after "/results/" in the URL path
+        const match = this.pathName.match(/\/results\/([^\/]+)/);
+        return match?.[1];
     }
```
onursumer, line 443: We should probably add a pure utility function outside of the class, something like below, so that we can add unit tests.

```JavaScript
function getCurrentTabSegment(pathName: string): string | undefined {
    // Extract tab name after "/results/" in the URL path
    const match = pathName.match(/\/results\/([^\/]+)/);
    return match?.[1];
}
```

Then we can modify this one as a `get` function

```suggestion
    public get currentTabSegment(): string | undefined {
        return getCurrentTabSegment(this.pathName);
    }
```

rishisulakhe: Hi @onursumer 
I looked into adding a utility function as suggested, and discovered we already have this functionality in ResultsViewPageHelpers via the getTabId() function, which is already unit tested.

I have updated the implementation to reuse this existing function instead of creating a duplicate


```
private get currentTabSegment(): string | undefined {
  return getTabId(this.pathName);
}
```

Let me know any adjustments for this approach.

alisman: Amazing!

src/pages/resultsView/ResultsViewURLWrapper.ts
alisman: @rishisulakhe sorry but what about this one here

rishisulakhe: @alisman 
Now we have to make only `subTab `as a utility method and implement unit testing for the same. Right?
Will make a commit soon.

rishisulakhe:
@alisman 
I have implemented unit tests for new utility `getSubTabId` function.
Let me know any adjustments in this.
Thankyou

PR #5174, opened 2025-04-28
Fix unauthorized study error page

FlorisVleugels:
### Issue
When visiting the study view for an unauthorized study there are three requests that get sent that should not be sent when there are no valid studies. This results in the unauthorized error page not showing and instead we get the error page below:
image

 I also see the frontend spams requests indefinitely when on this error page, which seems related to this [issue](https://github.com/cBioPortal/cbioportal/issues/11160).

### Fix
Only make the calls when there are no unauthorized studies in the study view. This results again in the unauthorized error page when a user does attempt to open a link to an unauthorized study (and the frontend no longer spams requests on the error page):
image

### Related issue
[#11479](https://github.com/cBioPortal/cbioportal/issues/11478)

### DIFF
--- a/src/pages/studyView/StudyViewPageStore.ts
+++ b/src/pages/studyView/StudyViewPageStore.ts
@@ -10115,7 +10115,7 @@ export class StudyViewPageStore
 
     // Poll ClinicalEventTypeCounts API  with no filter to determine if table should be added to StudyView Page
     public readonly shouldDisplayClinicalEventTypeCounts = remoteData({
-        await: () => [this.queriedPhysicalStudyIds],
+        await: () => [this.queriedPhysicalStudyIds, this.unknownQueriedIds],
         invoke: async () => {
             if (this.unknownQueriedIds.result.length === 0) {
                 const filters: Partial<StudyViewFilter> = {};
@@ -10129,6 +10129,8 @@ export class StudyViewPageStore
                         )
                     ).length > 0
                 );
+            } else {
+                return Promise.resolve(false);
             }
         },
     });
@@ -10477,7 +10479,7 @@ export class StudyViewPageStore
     });
 
     public readonly shouldDisplayPatientTreatments = remoteData({
-        await: () => [this.queriedPhysicalStudyIds],
+        await: () => [this.queriedPhysicalStudyIds, this.unknownQueriedIds],
         invoke: () => {
             if (this.unknownQueriedIds.result.length === 0) {
                 return this.internalClient.getContainsTreatmentDataUsingPOST({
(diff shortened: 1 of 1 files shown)


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10477,18 +10479,28 @@ export class StudyViewPageStore
     public readonly shouldDisplayPatientTreatments = remoteData({
         await: () => [this.queriedPhysicalStudyIds],
         invoke: () => {
-            return this.internalClient.getContainsTreatmentDataUsingPOST({
-                studyIds: toJS(this.queriedPhysicalStudyIds.result),
-            });
+            if (this.unknownQueriedIds.result.length === 0) {
```
alisman, line 10482: this is what shouldDisplayClinicalEventTypeCounts should look like

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10499,7 +10511,7 @@ export class StudyViewPageStore
     >({
         await: () => [this.shouldDisplayPatientTreatments],
         invoke: async () => {
-            if (this.shouldDisplayPatientTreatments.result) {
+            if (this.shouldDisplayPatientTreatments.result != false) {
```
alisman, line 10514: it will be boolean once you complete the if statement so you can undo this

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10117,17 +10117,19 @@ export class StudyViewPageStore
...
-            return Promise.resolve(
-                (
-                    await this.internalClient.getClinicalEventTypeCountsUsingPOST(
-                        {
-                            studyViewFilter: filters as StudyViewFilter,
-                        }
-                    )
-                ).length > 0
-            );
+            if (this.unknownQueriedIds.result.length === 0) {
```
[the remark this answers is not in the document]
FlorisVleugels: Yea makes way more sense, changed it in latest commit. Also added awaits for unknownQueriedIds. Let me know if its ok and ill squash it if needed

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10475,20 +10479,30 @@ export class StudyViewPageStore
...
     });
 
     public readonly shouldDisplaySampleTreatments = remoteData({
-        await: () => [this.queriedPhysicalStudyIds],
+        await: () => [this.queriedPhysicalStudyIds, this.unknownQueriedIds],
         invoke: () => {
-            return this.internalClient.getContainsSampleTreatmentDataUsingPOST({
-                studyIds: toJS(this.queriedPhysicalStudyIds.result),
-            });
+            if (this.unknownQueriedIds.result.length === 0) {
```
alisman, line 10497: @FlorisVleugels I realized, it would be better to find where these remotes are called and avoid calling them if their are unkonwnQueriedIds.   I.e the check should really be at the page level.    This is true for both shouldDisplayPatientTreatments and shouldDIsplaySampleTreatments.     I.e. it's not really their concert to check for the unkonwnQueriedIds.

alisman: @FlorisVleugels can you take a look at the above?

alisman:
We will try to fix with a different solution

PR #5175, opened 2025-04-28
Update url navigation to fix failing e2e localdb tests on backend and restore localdb errors screenshots

gblaih:
Fix failing localdb tests on the backend by updating url navigation to first go to a blank page (issues with hash params). Also restore localdb errors screenshots on circleci due to incorrect paths

### DIFF
--- a/end-to-end-test/shared/specUtils_Async.js
+++ b/end-to-end-test/shared/specUtils_Async.js
@@ -258,6 +258,8 @@ async function goToUrlAndSetLocalStorage(url, authenticated = false) {
     const currentUrl = await browser.getUrl();
     const needToLogin =
         authenticated && (!currentUrl || !currentUrl.includes('http'));
+    // navigate to blank page first to prevent issues with url hash params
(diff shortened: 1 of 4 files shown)


end-to-end-test/shared/specUtils_Async.js
```diff
@@ -258,6 +258,8 @@ async function goToUrlAndSetLocalStorage(url, authenticated = false) {
     const currentUrl = await browser.getUrl();
     const needToLogin =
         authenticated && (!currentUrl || !currentUrl.includes('http'));
+    // navigate to blank page first to prevent issues with url hash params
+    await browser.url('about:blank');
```
alisman, line 262: can you see if we have an existing navigation method which does this under the hood.  i have some memory

gblaih: we had this blank page navigation in this same method name in the non async version of this spec file. but it didn't get retained when we moved this to async for some reason, so i'm just restoring this

PR #5182, opened 2025-05-05
Add aria-label attribute to fix the link-name-related and the select-name-related 508 issues

johnyesit:
This PR adds "aria-label" attribute to the elements to improve 508 compliance.


src/pages/staticPages/datasetView/DatasetList.tsx
```diff
@@ -59,7 +59,13 @@ class CancerStudyCell extends React.Component<ICancerStudyCellProps, {}> {
 class ReferenceCell extends React.Component<IReferenceCellProps, {}> {
     render() {
         return (
-            <a target="_blank" href={getNCBIlink(`/pubmed/${this.props.pmid}`)}>
+            <a
+                target="_blank"
+                href={getNCBIlink(`/pubmed/${this.props.pmid}`)}
+                aria-label={
+                    this.props.citation ? undefined : 'No Reference Link'
```
alisman, line 66: this doesn't seem right.  if there IS a citation then we put undefined?  I mean, if there's no citation then the link will have no text

johnyesit: > Let's put the condition higher up and return </> if there's no citation

Does this look good?
Screenshot 2025-05-05 at 3 40 58 PM

alisman (approved):

PR #5188, opened 2025-05-12
Fix bugs in OncoPrint's tabular download due to OQL

oplantalech:
The logic of building the tabular download for OncoPrint assumes each track is a different gene. This causes issues when creating the tabular file for a query with multiple tracks of the same gene, only keeping the alterations of the last track of the gene.

For example, for the query with CDKN2B:MUT and CDKN2B:HOMDEL, when we have HOMDELs as the second track, only HOMDELs are shown in the downloaded tabular file (and no MUTs): https://www.cbioportal.org/results/oncoprint?cancer_study_list=ccle_broad_2019&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cstructural_variants%2Ccna&case_set_id=ccle_broad_2019_cnaseq&gene_list=CDKN2B%253AMUT%250ACDKN2B%253AHOMDEL&geneset_list=%20&tab_index=tab_visualize&Action=Submit&show_samples=true 

When the same query is with MUT displayed in the second track, and HOMDELs in the first track, then we only see the MUT (and no HOMDELs) in the tabular file: https://www.cbioportal.org/results/oncoprint?cancer_study_list=ccle_broad_2019&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cstructural_variants%2Ccna&case_set_id=ccle_broad_2019_cnaseq&gene_list=CDKN2B%253AHOMDEL%250ACDKN2B%253AMUT&geneset_list=%20&tab_index=tab_visualize&Action=Submit&show_samples=true

This PR fixes this issue, combining the alterations that occur in multiple tracks when they belong to the same gene.

### DIFF
--- a/src/shared/components/oncoprint/tabularDownload.ts
+++ b/src/shared/components/oncoprint/tabularDownload.ts
@@ -106,7 +106,7 @@ export function getTabularDownloadData(
                     : geneticTrackDatum.patient!;
 
             // Only initialize each alteration for sample/patient if it doesn't exist
-            // That can happen when we have multiple OncoPrint tracks of the same gene (using OQL)
+            // It may already exist if there are multiple OncoPrint tracks of the same gene
             if (oncoprintData.CNA[currentGeneName][id] == undefined) {
                 oncoprintData.CNA[currentGeneName][id] = '';
             }
@@ -154,7 +154,8 @@ export function getTabularDownloadData(
                 if (
                     oncoprintData.STRUCTURAL_VARIANT[currentGeneName][id] ==
                         '' ||
-                    oncoprintData.CNA[currentGeneName][id] == current_value
+                    oncoprintData.STRUCTURAL_VARIANT[currentGeneName][id] ==
+                        current_value
                 ) {
                     oncoprintData.STRUCTURAL_VARIANT[currentGeneName][
                         id
@@ -162,10 +163,9 @@ export function getTabularDownloadData(
                 } else {
                     oncoprintData.STRUCTURAL_VARIANT[currentGeneName][
                         id
-                    ] = oncoprintData.CNA[currentGeneName][id].concat(
-                        ', ',
-                        current_value
-                    );
+                    ] = oncoprintData.STRUCTURAL_VARIANT[currentGeneName][
+                        id
+                    ].concat(', ', current_value);
                 }
             }
             if (geneticTrackDatum.disp_mrna !== undefined) {
@@ -174,13 +174,16 @@ export function getTabularDownloadData(
                     : geneticTrackDatum.disp_mrna;
                 if (
                     oncoprintData.MRNA[currentGeneName][id] == '' ||
-                    oncoprintData.CNA[currentGeneName][id] == current_value
+                    oncoprintData.MRNA[currentGeneName][id] == current_value
                 ) {
                     oncoprintData.MRNA[currentGeneName][id] = current_value;
                 } else {
-                    oncoprintData.MRNA[currentGeneName][id] = oncoprintData.CNA[
-                        currentGeneName
-                    ][id].concat(', ', current_value);
+                    oncoprintData.MRNA[currentGeneName][
+                        id
+                    ] = oncoprintData.MRNA[currentGeneName][id].concat(
+                        ', ',
+                        current_value
+                    );
                 }
             }
             if (geneticTrackDatum.disp_prot !== undefined) {
@@ -189,13 +192,13 @@ export function getTabularDownloadData(
                     : geneticTrackDatum.disp_prot;
                 if (
                     oncoprintData.PROTEIN[currentGeneName][id] == '' ||
-                    oncoprintData.CNA[currentGeneName][id] == current_value
+                    oncoprintData.PROTEIN[currentGeneName][id] == current_value
                 ) {
                     oncoprintData.PROTEIN[currentGeneName][id] = current_value;
                 } else {
                     oncoprintData.PROTEIN[currentGeneName][
                         id
-                    ] = oncoprintData.CNA[currentGeneName][id].concat(
+                    ] = oncoprintData.PROTEIN[currentGeneName][id].concat(
                         ', ',
                         current_value
                     );
@@ -207,7 +210,8 @@ export function getTabularDownloadData(
                     : geneticTrackDatum.disp_mut;
                 if (
                     oncoprintData.MUTATIONS[currentGeneName][id] == '' ||
-                    oncoprintData.CNA[currentGeneName][id] == current_value
+                    oncoprintData.MUTATIONS[currentGeneName][id] ==
+                        current_value
                 ) {
                     oncoprintData.MUTATIONS[currentGeneName][
                         id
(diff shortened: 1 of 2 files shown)


src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -104,49 +104,122 @@ export function getTabularDownloadData(
...
                     ? geneticTrackDatum.sample!
                     : geneticTrackDatum.patient!;
-            oncoprintData.CNA[currentGeneName][id] = '';
-            oncoprintData.MUTATIONS[currentGeneName][id] = '';
-            oncoprintData.MRNA[currentGeneName][id] = '';
-            oncoprintData.PROTEIN[currentGeneName][id] = '';
-            oncoprintData.STRUCTURAL_VARIANT[currentGeneName][id] = '';
+
+            // Only initialize each alteration for sample/patient if it doesn't exist
+            // That can happen when we have multiple OncoPrint tracks of the same gene (using OQL)
```
alisman, line 109: the comment here is somewhat confusing.   i think you mean,   "it may already exist if there are multiple oncoprint tracks of the same gene (e.g. different subtypes)

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -104,49 +104,122 @@ export function getTabularDownloadData(
...
-                    id
-                ] = structuralVariantMap[
+                var current_value = structuralVariantMap[
                     geneticTrackDatum.disp_structuralVariant + ''
                 ]
                     ? structuralVariantMap[
                           geneticTrackDatum.disp_structuralVariant + ''
                       ]
                     : geneticTrackDatum.disp_structuralVariant;
+                if (
```
alisman, line 154: not sure i understand this if statement.  could use. comment

src/shared/components/oncoprint/tabularDownload.ts
oplantalech: I've adjusted the comment

src/shared/components/oncoprint/tabularDownload.ts
oplantalech: I've reviewed it. There were some mistakes in the code, I've just fixed them.

We might be able to simplify this since we're doing the same for the different alteration types (cna, mutations, mRNA, protein, and structural variants) using a for loop, but we would probably need a dictionary since the parts they change between alterations are not exactly the same (e.g. disp_cna and CNA in CNA, disp_structuralVariant and STRUCTURAL_VARIANT for structural variants, etc.). Do you think it's a good idea?

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -174,13 +174,16 @@ export function getTabularDownloadData(
                     : geneticTrackDatum.disp_mrna;
                 if (
                     oncoprintData.MRNA[currentGeneName][id] == '' ||
-                    oncoprintData.CNA[currentGeneName][id] == current_value
+                    oncoprintData.MRNA[currentGeneName][id] == current_value
```
alisman, line 177: try to always use === instead of == (no coercion)
it's stricter.  probably doesn't make a difference but just a good practice

src/shared/components/oncoprint/tabularDownload.ts
alisman: I agree that these look exactly the same and this should be abstracted.     not sure i see where the dictionary is necessary

oplantalech: I've abstracted the code. Now you can see what I meant with the dictionary

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -78,75 +79,66 @@ export function getTabularDownloadData(
...
-                    geneticTrackDatum.disp_mut
-                ]
-                    ? mutationMap[geneticTrackDatum.disp_mut]
-                    : geneticTrackDatum.disp_mut;
+
+            // Check if the sample/patient is initialized, and combine both results in case they are different
+            for (const alteration in alterationsMap) {
+                var disp = alterationsMap[alteration] as keyof typeof string;
+                if (geneticTrackDatum[disp] !== undefined) {
+                    var current_value = cnaMap[geneticTrackDatum[disp]]
```
alisman: @oplantalech should this really be cnaMap for all alteration type?

oplantalech: Thanks, fixed

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -78,75 +79,75 @@ export function getTabularDownloadData(
...
+                            id
+                        ] = current_value;
+                    } else {
+                        oncoprintData[alteration][currentGeneName][
+                            id
+                        ] = oncoprintData[alteration][currentGeneName][
+                            id
+                        ].concat(', ', current_value);
+                    }
+                }
```
dippindots, line 145: Do you think we can improve this part like this makes it easier to read?
```
    const disp = alterationsMap[alteration];
    const dispValue = geneticTrackDatum[disp];

    const geneticAlterationMap: any = {
        CNA: cnaMap,
        MUTATIONS: mutationMap,
        MRNA: mrnaMap,
        PROTEIN: proteinMap,
        STRUCTURAL_VARIANT: structuralVariantMap,
    };
    
    if (dispValue !== undefined) {
        const map = geneticAlterationMap[alteration];
        const currentValue = map[dispValue] ?? dispValue;

        const geneTrack = oncoprintData[alteration][currentGeneName];
        const existingValue = geneTrack[id];

        if (!existingValue || existingValue === currentValue) {
            geneTrack[id] = currentValue;
        } else {
            geneTrack[id] = `${existingValue}, ${currentValue}`;
        }
    }
```

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -78,75 +79,75 @@ export function getTabularDownloadData(
...
-        }
-        if (oncoprintData.PROTEIN[currentGeneName] === undefined) {
-            oncoprintData.PROTEIN[currentGeneName] = {};
-        }
-        if (oncoprintData.STRUCTURAL_VARIANT[currentGeneName] === undefined) {
-            oncoprintData.STRUCTURAL_VARIANT[currentGeneName] = {};
+        for (const alteration in alterationsMap) {
+            if (oncoprintData[alteration][currentGeneName] === undefined) {
+                oncoprintData[alteration][currentGeneName] = {};
+            }
```
dippindots, line 98: we can try to use nullish coalescing operator at here

```
oncoprintData[alteration][currentGeneName] ??= {};
```

src/shared/components/oncoprint/tabularDownload.ts
```diff
@@ -78,75 +79,75 @@ export function getTabularDownloadData(
...
-                    : geneticTrackDatum.disp_prot;
+
+            // Only initialize each alteration for sample/patient if it doesn't exist
+            // It may already exist if there are multiple OncoPrint tracks of the same gene
+            for (const alteration in alterationsMap) {
+                if (
+                    oncoprintData[alteration][currentGeneName][id] === undefined
+                ) {
+                    oncoprintData[alteration][currentGeneName][id] = '';
+                }
```
dippindots, line 115: same for this one:
```
oncoprintData[alteration][currentGeneName][id] ??= '';
```

oplantalech:
@dippindots Looks like all the tests pass now. I think I need to rebase, though

PR #5193, opened 2025-05-20
Fix filter checked state when navigating back to home page

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11552

### DIFF
--- a/src/shared/components/query/QueryStore.ts
+++ b/src/shared/components/query/QueryStore.ts
@@ -299,6 +299,49 @@ export class QueryStore {
 
     @observable.ref dataTypeFilters: string[] = [];
 
+    @observable studyFilterOptionsFormatted = [
+        {
+            id: 'sequencedSampleCount',
+            name: 'Mutations',
+            checked: false,
+        },
+        {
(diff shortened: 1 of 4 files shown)


src/shared/components/query/QueryStore.ts
```diff
@@ -299,6 +299,49 @@ export class QueryStore {
 
     @observable.ref dataTypeFilters: string[] = [];
 
+    @observable studyFilterOptionsFormatted = [
```
alisman: Is there an obvious reason why this is called "Formatted" ?  doesn't seem formatted to me.   Also, lets put the actual list of items in some helper file somewhere so it doesn't bloat the store.  if you just assign it to this observable here, it will still work fine.

gblaih: i just kept the original name for these options when moving them to the store, but my guess is that it is called "Formatted" as the options are formatted in a way as objects with id, name, and checked fields. i can change the name to `studyFilterOptions` if that fits better. i will also move the actual list from the store.

PR #5209, opened 2025-06-06
Add Variant Allele Frequency in Plots tab - GSoC'25

rishisulakhe:
- Added variant allele frequency in plots tab

## Any screenshots or GIFs?
Screenshot 2025-06-07 003357
Screenshot 2025-06-07 003517

## Notify reviewers
@onursumer


src/shared/components/plots/PlotsTabUtils.tsx
```diff
@@ -1146,101 +1146,137 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
-        _.uniq(
-            sampleMuts.map(
-                m => mutationTypeToDisplayName[getOncoprintMutationType(m)]
-            )
-        )
-    );
+    let data: {
+        uniqueSampleKey: string;
+        value: string | number | string[] | number[];
+    }[] = [];
```
onursumer: Can we keep `data` still a `const`? I don't think we need to modify it later, right?

```suggestion
    const data: {
        uniqueSampleKey: string;
        value: string | number | string[] | number[];
    }[] = samples.map(s => {;
```

src/shared/components/plots/PlotsTabUtils.tsx
```diff
@@ -1146,101 +1146,137 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
+                            if (
+                                Number.isInteger(m.tumorAltCount) &&
+                                Number.isInteger(m.tumorRefCount)
+                            ) {
+                                const total = m.tumorAltCount + m.tumorRefCount;
+                                return total > 0
+                                    ? m.tumorAltCount / total
+                                    : null;
+                            }
+                            return null;
```
onursumer: Let's reuse an existing utility function for this if possible.

https://github.com/cBioPortal/cbioportal-frontend/blob/2894692dc0ae1e02b7985aa2fe49fb44f9c7631d/src/shared/lib/MutationUtils.tsx#L382-L399

src/shared/components/plots/PlotsTab.tsx
```diff
@@ -220,6 +220,7 @@ export enum MutationCountBy {
     MutationType = 'MutationType',
     MutatedVsWildType = 'MutatedVsWildType',
     DriverVsVUS = 'DriverVsVUS',
+    VAF = 'VAF',
```
onursumer: Should we just spell it out for clarity?

```suggestion
    VariantAlleleFrequency = 'VariantAlleleFrequency',
```

src/shared/components/plots/PlotsTabUtils.tsx
```diff
@@ -1146,101 +1146,137 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
-            } else {
-                // otherwise, its profiled, so not mutated
-                plotValue = MUT_PROFILE_COUNT_NOT_MUTATED;
-            }
-        } else {
-            // we have mutations
-            switch (mutationCountBy) {
-                case MutationCountBy.MutationType:
-                    const sampleMutTypes = _.uniq(
+        let value: any;
```
onursumer: Is it possible to be more type specific instead of `any`?

onursumer (changes requested):
It looks good in general, thanks! Just requested changes for a few minor things.

Also, looks like `prettier` is failing. You might have forgotten to run the commit hooks.

src/shared/components/plots/PlotsTabUtils.tsx
rishisulakhe: Right data is not modifying further. Will keep it `const`

src/shared/components/plots/PlotsTabUtils.tsx
rishisulakhe: Sure. Will reuse this function

src/shared/components/plots/PlotsTab.tsx
rishisulakhe: yes.

src/shared/components/plots/PlotsTabUtils.tsx
rishisulakhe: yes value can be number, string or array of numbers or strings.

rishisulakhe: @onursumer 
 Committed all requested changes.
 Also added Unit test for Variant Allele Frequency.
 Thankyou.

rishisulakhe: Working on prettier issue...

src/shared/components/plots/PlotsTabUtils.tsx
```diff
@@ -1146,101 +1147,128 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
+            case MutationCountBy.VariantAlleleFrequency:
+                if (!sampleMutations) {
+                    value = isProfiled ? 0 : NaN;
+                } else {
+                    const vafs: number[] = [];
+                    for (const m of sampleMutations) {
+                        const report = getVariantAlleleFrequency(m);
+                        if (report !== null) {
+                            vafs.push(report.vaf);
+                        }
```
onursumer: Let's use the `map` and `compact` functions for better readability, something like this:

```suggestion
                    const vafs: number[] = _(sampleMutations)
                        .map(m => getVariantAlleleFrequency(m)?.vaf)
                        .compact()
                        .value();
```

rishisulakhe: Sure, will use these functions.

src/shared/components/plots/PlotsTabUtils.tsx
```diff
@@ -1146,101 +1147,132 @@ export function makeAxisDataPromise_Molecular_MakeMutationData(
...
-        case MutationCountBy.MutatedVsWildType:
-        default:
-            categoryOrder = mutVsWildCategoryOrder;
-            break;
+
+    if (mutationCountBy === MutationCountBy.VariantAlleleFrequency) {
+        const hasVafData = mutations.some(
+            m =>
+                typeof m.tumorAltCount === 'number' &&
+                typeof m.tumorRefCount === 'number'
```
onursumer: Let's use `_.isFinite` function instead because `typeof NaN` is also a `'number'`

```suggestion
                _.isFinite(m.tumorAltCount) &&
                _.isFinite(m.tumorRefCount)
```

onursumer:
Just noticed that certain label name changes are not desirable. 

Examples:

`Mutations` -> `Mutations Type`

before:
TP53-Mutations

after:
mutations_type

`TP53: Mutations (MSK-IMPACT)` -> `TP53: Mutations Type`

before:
Screenshot 2025-09-22 at 5 46 06 PM

after:
Screenshot 2025-09-22 at 5 45 34 PM

rishisulakhe:
Hi @inodb
Thanks for the feedback! I've updated the implementation to filter out samples without mutations in that specific gene and samples with mutations but no vaf data. Also updated the top status bar.

Screenshot 2025-10-03 203123

inodb (approved):

PR #5213, opened 2025-06-13
Use new count of alterationsInProfiled genes as numerator in study vi…

alisman:


src/pages/studyView/table/MultiSelectionTable.tsx
```diff
@@ -361,9 +369,23 @@ export class MultiSelectionTable extends React.Component<
                         styles.pullRight
                     );
                 },
-                sortBy: (data: MultiSelectionTableRow) =>
-                    (data.numberOfAlteredCases / data.numberOfProfiledCases) *
-                    100,
+                sortBy: (data: MultiSelectionTableRow) => {
+                    if (
+                        // @ts-ignore
+                        data.numberOfAlteredCasesOnPanel === 0 ||
```
alisman: @fuzhaoyuan i think this needs a gate for backward compat.  will put one in.

PR #5215, opened 2025-06-16
When only RNA profile present in study, show it by default

DavidZ79:
Fix front end ticket [#11569](https://github.com/cBioPortal/cbioportal/issues/11569)

In Query View: (~~MolecularProfileSelector.tsx~~ QueryStore.ts)
Added logic in componentDidMount to auto select an mRNA profile if no mutation profiles exist.
query_page_gif

In Study View: (StudyViewPageStore.ts)
Auto adds the default mRNA profile during study page to query page transition if no molecular profiles are selected.
study_page_gif


inodb (approved):
Thanks for implementing! Looks great from functionality perspective - maybe someone else can look at implementation @onursumer @alisman ?

src/shared/components/query/MolecularProfileSelector.tsx
```diff
@@ -21,6 +21,27 @@ export default class MolecularProfileSelector extends QueryStoreComponent<
...
+
+        if (!hasMutationProfile && hasMrnaProfile) {
+            const mrnaProfiles = this.store.getFilteredProfiles(
+                'MRNA_EXPRESSION' as MolecularProfile['molecularAlterationType']
+            );
+            if (mrnaProfiles.length > 0) {
+                this.store.selectMolecularProfile(mrnaProfiles[0], true);
+            }
+        }
+    }
```
inodb: @DavidZ79 if there's an easy way to do this in the render method instead that would be better,  but no worries if not

alisman: @DavidZ79 this is an antipattern to put state change in the mount method.  i would assume there is somewhere else that we are handling the default profiles selection.  Can you look into that?

DavidZ79: Thanks @inodb and @alisman for the feedback, I moved the default mRNA profile selection logic out of componentDidMount and into QueryStore. Let me know if this is good or you'd like me to tweak it.

src/shared/components/query/QueryStore.ts
```diff
@@ -1766,6 +1772,29 @@ export class QueryStore {
         );
     }
 
+    @computed get shouldSelectDefaultMrnaProfile() {
```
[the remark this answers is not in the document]
DavidZ79: I addressed this by creating a new property `filteredMrnaProfiles`

src/shared/components/query/QueryStore.ts
```diff
@@ -1766,6 +1772,33 @@ export class QueryStore {
...
+
+    @computed get shouldSelectDefaultMrnaProfile() {
+        // select default mRNA profile if no profiles are currently selected
+
+        if (Object.keys(this.selectedProfileIdSet).length > 0) {
+            return false;
+        }
+
+        const hasMutationProfile =
+            this.getFilteredProfiles('MUTATION_EXTENDED').length > 0;
```
alisman: there are other profile types besides MUTATION_EXTENDED and MRNA so I suspect this logic is too narrow.  maybe what we really mean to do is if there is only ONE profile always use it?  or never have zero selected profiles?

DavidZ79: sorry for the delay, I've edited the logic so it checks for when there is only one profile, then use it. Please let me know if this change doesn't look good to you

src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10206,6 +10206,15 @@ export class StudyViewPageStore
                     getSuffixOfMolecularProfile(this.defaultProtProfile)
                 );
             }
+            // Add RNA profile if no other profiles are selected
```
alisman: @inodb why only do this when there are no other profiles selected?  it's arbitrary.  why not just add it either way, just like those above?

src/shared/components/query/QueryStore.ts
```diff
@@ -1766,6 +1772,49 @@ export class QueryStore {
         );
     }
 
+    @computed get filteredMrnaProfiles() {
+        return this.getFilteredProfiles('MRNA_EXPRESSION');
+    }
+
+    @computed get shouldSelectDefaultMrnaProfile() {
```
alisman: sorry to circle back again but this just seems so arbitrary.  can we just automatically turn on whatever profile if there is only one profile?

alisman (changes requested):
See comments.

src/pages/studyView/StudyViewPageStore.ts
DavidZ79: Hi, I removed the previous `molecularProfileFilters.length === 0` condition, and now it just adds the mRNA profile if exists, like the above profile checks in the code.

src/shared/components/query/QueryStore.ts
DavidZ79: I changed it a little bit to it auto selects any profile type when only one exists (not just mRNA), please double check when you have a moment. The functionality remains the same as shown in the gifs above.

alisman:
@inodb @onursumer.   this solution seems kind of arbitrary.  why don'e we just always show ALL profiles when we submit from study view?  that would be a more general solution.

inodb:
@alisman Yea good question. I guess we tend to only show mutations by default when you query because of the complexity of interpreting rnaseq results. That's why the issue was more put up as "if no mutations profile exist show the profile you do have (e.g. rnaseq only studies)". But I like your suggestion, could go either way. @tmazor @jjgao what do you think?

PR #5224, opened 2025-07-02
Add Similarity Map (Patient Embeddings) to Study View

inodb:
We've received various requests to show embeddings (e.g. UMAP/PCA) on the Study View. This implements a prototype using deckgl which can support 1M points:

See [preview](https://deploy-preview-5224.preview.cbioportal.org/study/embeddings?id=696aaf31b2bb32147b01ea70&featureFlags=EMBEDDINGS)
image

### Features
- Recangle Selection
- Reflects filtering from study view

### TODO
- [x] ~Get a realistic embedding (currently bogus for 24K cases)~
- [x] ~Fix driver coloring, currently everything VUS (copy plots tab style)~
- [x] Add tests
- [x] Add a way to pull the coordinates from an S3 bucket
- [x] Add feature flag `featureFlags=EMBEDDINGS`
- [x] Re-enable "hidden dots", come up with good name
- [x] Set up a public virutal study with all the samples used in the embedding
- [x] Change name to Similarity Maps
- [x] Add lasso selection


package.json
```diff
@@ -241,7 +245,6 @@
     "pathway-mapper": "^2.3.0",
     "pdfobject": "^2.0.201604172",
     "pegjs": "^0.10.0",
-    "plotly.js": "^1.42.5",
```
inodb: i don't think this is used anywhere, so removed it

PR #5225, opened 2025-07-06
Align heatmap RNASeq scores with oncoprint tracks #11592

DavidZ79:
Fix frontend ticket [#11592](https://github.com/cBioPortal/cbioportal/issues/11592)

Note: left original logic as fallback in DataUtils.ts

{28C004B7-0F24-4A4F-B026-C69B849CBDB2}


inodb (approved):
Thanks! This def looks better!

Some comment in the description as well how the aggregation logic would be good. Also ideally both tracks should use the same functions for the aggregation logic, I didn't confirm in code if that's how it works

Note that selecting the patient level aggregation logic for RNASeq is a bit more complex than for mutation data (union), but this seems reasonable (if an expression query with threshold exists -> pick thresholded samples absolute max value)

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1175,13 +1175,19 @@ export function makeHeatmapTracksMobxPromise(
...
 
-            return cacheQueries.map(query => {
+            const tracks = cacheQueries.map(query => {
                 const molecularProfileId = query.molecularProfileId;
                 const gene = query.hugoGeneSymbol;
                 const data = oncoprint.props.store.geneMolecularDataCache.result!.get(
                     query
                 )!.data!;
 
+                // get Z-score threshold from the store (default to 2 if not present)
```
alisman: can you see if there is a constand somewhere for default 2?

src/shared/components/oncoprint/DataUtils.ts
```diff
@@ -416,43 +417,99 @@ export function fillHeatmapTrackDatum<
...
-                    representingDatum = selectRepresentingDataPoint(
-                        bestValue!,
-                        dataWithValue,
-                        false
-                    );
-                    break;
-                default:
-                    bestValue = _.maxBy(dataWithValue, (d: HeatmapCaseDatum) =>
-                        Math.abs(d.value)
+
```
alisman, line 423: would be good to have a unit test for this.  let me know if you need help setting that up

src/shared/components/oncoprint/OncoprintUtils.ts
DavidZ79: There was one in another file, but I think it would be best to move that one to shared/constants.ts and use it here as well.

src/shared/components/oncoprint/DataUtils.ts
DavidZ79: Added, all tests are passing. When you got a moment, please have a look, since this is my first time working with these tests, thanks :)

alisman:
@inodb need to discuss this again.  my feeling is that we should now show a heatmap value for a patient where they have to contradictory samples.   there should be some new glyph that indicates "problems" and then in the tooltip, we show the multiple valus.  if you agree with this maybe we can claude it?

onursumer:
@copilot resolve the merge conflicts in this pull request

dippindots:
long time no response, decide to close it

PR #5235, opened 2025-07-17
In local development mode (frontend dev server) clickhouse api endpoints broken

alisman:


rishisulakhe:
Hi @alisman 
Onur and I also discovered this bug recently in our local development environment.
I have just tested these changes locally, found that it still not work, still showing CORS error.
image

Upon debugging, I found that the host variable returns undefined in development mode at this line
https://github.com/cBioPortal/cbioportal-frontend/blob/0f18baf3bc0d064a1ff34069b4586ab4d1569631/src/shared/api/proxyColumnStore.ts#L64
Meanwhile i am using like this as fallback host resolution.
 ```
const host = getLoadConfig().baseUrl;     
 const resolvedHost = host || 'www.cbioportal.org';
```

alisman:
@rishisulakhe this works on on the preview for this very branch.  i merged it.  try rebasing your branch.   it works locally for me as well as long as i am using ```yarn run startSSL```.

If it still doesn't for you, please hit me on the cbio public slack channel

PR #5237, opened 2025-07-17
Add Variant Allele Frequency Heatmap in Oncoprint - GSoC'25

rishisulakhe:
Registered "Variant Allele Frequency" as a new type of molecular profile heatmap that the OncoPrint can display. This allowed it to appear as a selectable option in the "Heatmap" of "Track" menu alongside existing heatmaps.

https://github.com/user-attachments/assets/2bbea084-169e-4b71-8b9f-cad64ec8f438

Preview link: https://deploy-preview-5237.preview.cbioportal.org/

### DIFF
--- a/src/shared/components/oncoprint/DeltaUtils.ts
+++ b/src/shared/components/oncoprint/DeltaUtils.ts
@@ -1636,7 +1636,7 @@ export function transitionHeatmapTrack(
                     .isWhiteBackgroundForGlyphsEnabled
         ) {
             if (
-                typeof trackIdForRuleSetSharing.mutation !== 'undefined' &&
+                trackIdForRuleSetSharing.mutation !== undefined &&
                 trackIdForRuleSetSharing.mutation !== trackId
             ) {
                 oncoprint.shareRuleSet(
--- a/src/shared/components/oncoprint/OncoprintUtils.ts
+++ b/src/shared/components/oncoprint/OncoprintUtils.ts
@@ -627,11 +627,7 @@ function createHeatmapTracksData(
 
             mutations.forEach(m => {
                 const vafReport = getVariantAlleleFrequency(m);
-                if (
-                    vafReport &&
-                    typeof vafReport.vaf === 'number' &&
-                    !isNaN(vafReport.vaf)
-                ) {
+                if (vafReport && _.isFinite(vafReport.vaf)) {
                     const existingVaf = sampleMutationMap.get(
                         m.uniqueSampleKey
                     );


src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,27 @@ export function getCategoricalTrackRuleSetParams(
...
 }
 
+function getGeneProfileQueries(
+    molecularProfileIdToAdditionalTracks: any,
+    geneCache: any
+) {
+    const geneProfiles = _.filter(
+        _.values(molecularProfileIdToAdditionalTracks),
+        d => d.molecularAlterationType !== AlterationTypeConstants.GENERIC_ASSAY
+    );
```
onursumer: chaining usually makes the code more readable

```suggestion
    const geneProfiles = _(molecularProfileIdToAdditionalTracks).values().filter(
        d => d.molecularAlterationType !== AlterationTypeConstants.GENERIC_ASSAY
    ).value();
```

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,27 @@ export function getCategoricalTrackRuleSetParams(
...
+    return _.flatten(
+        geneProfiles.map(entry =>
+            _.keys(entry.entities).map(g => ({
+                molecularProfileId: entry.molecularProfileId,
+                entrezGeneId: geneCache.get({ hugoGeneSymbol: g })?.data
+                    ?.entrezGeneId,
+                hugoGeneSymbol: g.toUpperCase(),
+            }))
+        )
+    ).filter(query => query.entrezGeneId);
```
onursumer: similar to the above, we can chain here as well

```suggestion
    return _(geneProfiles).map(entry =>
            _.keys(entry.entities).map(g => ({
                molecularProfileId: entry.molecularProfileId,
                entrezGeneId: geneCache.get({ hugoGeneSymbol: g })?.data
                    ?.entrezGeneId,
                hugoGeneSymbol: g.toUpperCase(),
            }))
        )
    ).filter(query => query.entrezGeneId)
    .flatten().value();
```

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1130,113 +1153,223 @@ export function makeHeatmapTracksMobxPromise(
...
+
+            const mutationQueries = cacheQueries.filter(query => {
+                const profileType =
+                    molecularProfileIdToMolecularProfile[
+                        query.molecularProfileId
+                    ]?.molecularAlterationType;
+                return (
+                    profileType === AlterationTypeConstants.MUTATION_EXTENDED
+                );
+            });
```
onursumer: it's not ideal to have complex logic in the await function. we may want to move this logic to `invoke` or to another store remote data and await for it for simplicity

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1130,113 +1153,223 @@ export function makeHeatmapTracksMobxPromise(
...
+                oncoprint.props.store.molecularProfileIdToMolecularProfile,
+                oncoprint.props.store.geneMolecularDataCache,
+                ...mutationQueries.map(query =>
+                    oncoprint.props.store.annotatedMutationCache.get({
+                        entrezGeneId: query.entrezGeneId!,
+                    })
+                ),
+            ];
+        },
+        invoke: async () => {
```
onursumer, line 1438: This invoke function now got too long. We should probably try dividing it into sub functions.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1130,113 +1153,223 @@ export function makeHeatmapTracksMobxPromise(
...
+                }[] = [];
+
+                if (profileType === AlterationTypeConstants.MUTATION_EXTENDED) {
+                    const mutationPromise = oncoprint.props.store.annotatedMutationCache.get(
+                        {
+                            entrezGeneId: query.entrezGeneId,
+                        }
+                    );
+
+                    const mutations = mutationPromise.result;
```
onursumer: we should probably await for this result

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1130,113 +1153,223 @@ export function makeHeatmapTracksMobxPromise(
...
+                                        typeof tumorAltCount === 'number' &&
+                                        typeof tumorRefCount === 'number'
+                                    ) {
+                                        const totalCount =
+                                            tumorAltCount + tumorRefCount;
+                                        if (totalCount > 0) {
+                                            vafValue =
+                                                tumorAltCount / totalCount;
+                                        }
+                                    }
```
onursumer: I am surprised that we need this logic. I would assume that it's taken care of by the `getVariantAlleleFrequency` function.

onursumer (changes requested):
The functionality looks good, thanks!

The code can be improved/simplified further. See the comments.

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Thanks for this suggestion!

inodb:
Hey @rishisulakhe - this is cool! I tried it, but ran into this issue, everything shows up blank:

image

rishisulakhe:
Hi @inodb 
Thankyou for review. That issue is now fixed.
You can review it again and let me know if there are any other changes or improvements you'd recommend?

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: I think we do not need to `await` this. Data is already awaiting in main `await` function.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,115 @@ export function getCategoricalTrackRuleSetParams(
     };
 }
 
+function createHeatmapTracksData(
```
onursumer, line 557: Let's just add some comments to this function and the other two new functions to briefly describe what we're doing here.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,115 @@ export function getCategoricalTrackRuleSetParams(
...
+                })
+                .filter(
+                    (
+                        d
+                    ): d is {
+                        uniqueSampleKey: string;
+                        uniquePatientKey: string;
+                        value: number;
+                    } => d !== null
+                );
```
onursumer: we are filtering null values here, right? we can use lodash `compact` function for that.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,115 @@ export function getCategoricalTrackRuleSetParams(
...
+        }
+    } else {
+        const molecularDataResult = oncoprint.props.store.geneMolecularDataCache.result!.get(
+            query
+        );
+        if (molecularDataResult && molecularDataResult.data) {
+            return molecularDataResult.data.map((d: any) => ({
```
onursumer: let's use a more specific data type instead of `any` if possible

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,115 @@ export function getCategoricalTrackRuleSetParams(
...
+                uniquePatientKey: d.uniquePatientKey,
+            }));
+        }
+    }
+    return [];
+}
+
+function getGeneProfileQueries(
+    molecularProfileIdToAdditionalTracks: any,
+    geneCache: any
```
onursumer: can we be more specific here instead of using the `any` type?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +533,115 @@ export function getCategoricalTrackRuleSetParams(
...
+        )
+        .flatten()
+        .filter(query => query.entrezGeneId)
+        .value();
+}
+
+function getMutationHeatMapQueries(
+    molecularProfileIdToAdditionalTracks: any,
+    molecularProfileIdToMolecularProfile: any,
+    geneCache: any
```
onursumer: similarly, is it possible to use more specific types instead of `any`?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -1130,113 +1241,150 @@ export function makeHeatmapTracksMobxPromise(
...
+                oncoprint.molecularProfileIdToAdditionalTracks;
+
             const neededGenes = _.flatten(
-                geneProfiles.map(v => _.keys(v.entities))
+                _.filter(
+                    _.values(molecularProfileIdToAdditionalTracks),
+                    d =>
+                        d.molecularAlterationType !==
+                        AlterationTypeConstants.GENERIC_ASSAY
+                ).map(v => _.keys(v.entities))
```
onursumer: use chaining wherever possible to increase readability. So something like this would be better:

```suggestion
          _.values(molecularProfileIdToAdditionalTracks)
              .filter(d =>
                    d.molecularAlterationType !== AlterationTypeConstants.GENERIC_ASSAY
              )
              .map(v => _.keys(v.entities))
```

src/shared/components/oncoprint/controls/OncoprintControls.tsx
```diff
@@ -474,11 +475,21 @@ export default class OncoprintControls extends React.Component<
...
-                    value: profile.molecularProfileId,
-                    type: profile.molecularAlterationType,
-                })
+                profile => {
+                    let label = profile.name;
+                    if (
+                        profile.molecularAlterationType ===
+                        AlterationTypeConstants.MUTATION_EXTENDED
+                    ) {
+                        label = `Variant Allele Frequency`;
```
onursumer: Hardcoding `Variant Allele Frequency` might be problematic in the future if we want to add a new mutation related track other than VAF. Not sure what's the best way to address this issue though.

onursumer (changes requested):
Almost there. Requested a few more minor changes.

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Yes we are filtering null values. Using lodash `compact` would be much cleaner.

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Yes we have type `GeneCache`

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Yes we already have data type  `NumericGeneMolecularData`

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Thanks for the suggestion.

rishisulakhe:
Hi @inodb @onursumer 
Some updates.
> ## Bug 1
> For [This URL](https://deploy-preview-5237.preview.cbioportal.org/results?

Done✅. Now we are not showing VAF in heatmap option when no VAF data is available for any sample.

> ## Bug 2
> Ensure the VAF heatmap reflects OQL, i.e. the EGFR MUT=L858R filter should be applied to the VAF Heatmap track too:

Still working on it.

> ## Bug 3
> Profiled but EGFR mutation not detected should be white (the line should only show up for the "not profiled" cases, same as for regular mutations track):

Done✅. Now we are showing light grey color for samples with profiled but not mutated. Here is [demo](https://deploy-preview-5237.preview.cbioportal.org/results/oncoprint?cancer_study_list=pancan_pcawg_2020&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Ccna&case_set_id=pancan_pcawg_2020_cnaseq&gene_list=EGFR%2520TP53&geneset_list=%20&tab_index=tab_visualize&Action=Submit&heatmap_track_groups=pancan_pcawg_2020_mutations%2CEGFR%2CTP53)

Also we have one more bug from product team, still need to be address. Like we add clinical track(cancer type) and click on show gaps with %, then add VAF or any heatmap results showing incorrect rendering of genetic track.

tmazor:
This is awesome! A few things I noticed:

1) For other tracks eg RNA, the label in OncoPrint matches the label in the Heatmap dropdown. But for VAF, the OncoPrint is labeled very generically with just 'Mutations':

image

image

Let's call the OncoPrint section 'Variant Allele Frequency' instead of 'Mutations'



2) I think this bug is specific to studies with gene panels. Run a multi-gene query. Add VAF heatmap. Delete a gene from the query. If the study does not have gene panels, the deleted gene is still there w/ VAFs -- this matches how RNA/protein heatmaps behave, so this is reasonable. But if I delete the gene in a study that has gene panels, then the VAF track remains but it changes to all 'not sequenced':
https://deploy-preview-5237.preview.cbioportal.org/results/oncoprint?cancer_study_list=chol_msk_2018&Z_SCORE_THRESHOLD=2.0&RPPA_SCORE_THRESHOLD=2.0&profileFilter=mutations%2Cstructural_variants%2Ccna&case_set_id=chol_msk_2018_cnaseq&gene_list=IDH1&geneset_list=%20&tab_index=tab_visualize&Action=Submit&heatmap_track_groups=chol_msk_2018_mutations%2CIDH1%2CTP53

image



3) Maybe a toggle to respect or ignore OQL? Because I can come up with use cases for both sides of this, so maybe we just let the user decide?

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +538,208 @@ export function getCategoricalTrackRuleSetParams(
...
+        const samples = oncoprint.props.store.filteredSamples.result!;
+        const coverageInfo = oncoprint.props.store.coverageInformation.result!;
+        const queriedGenes = oncoprint.props.store.genes.result || [];
+        const isGeneInQuery = queriedGenes.some(
+            g =>
+                g.hugoGeneSymbol.toUpperCase() ===
+                query.hugoGeneSymbol.toUpperCase()
+        );
+
+        let mutations: any[] | undefined;
```
onursumer: Better to use a type here (Mutation?) instead of `any` if possible

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +538,208 @@ export function getCategoricalTrackRuleSetParams(
...
+                    if (
+                        existingVaf === undefined ||
+                        vafReport.vaf > existingVaf!
+                    ) {
+                        sampleMutationMap.set(m.uniqueSampleKey, vafReport.vaf);
+                    }
+                } else {
+                    if (!sampleMutationMap.has(m.uniqueSampleKey)) {
+                        sampleMutationMap.set(m.uniqueSampleKey, null);
+                    }
```
onursumer, line 646: I think setting an explicit null value to indicate profiled but not mutated is a little bit confusing. we should probably use a different value.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +538,208 @@ export function getCategoricalTrackRuleSetParams(
...
+            mutations = mutationPromise.result;
+            if (mutations) {
+                mutations = mutations.filter(
+                    m => m.molecularProfileId === query.molecularProfileId
+                );
+            }
+        }
+
+        if (mutations && mutations.length > 0) {
+            const sampleMutationMap = new Map<string, number | null>();
```
onursumer, line 626: a value of `null` is a bit confusing. we should probably use some other constant value here.

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +538,208 @@ export function getCategoricalTrackRuleSetParams(
...
+                    );
+                })
+                .map(sample => {
+                    const vafValue = sampleMutationMap.get(
+                        sample.uniqueSampleKey
+                    );
+                    return {
+                        uniqueSampleKey: sample.uniqueSampleKey,
+                        uniquePatientKey: sample.uniquePatientKey,
+                        value: vafValue !== undefined ? vafValue : null, // null for profiled but not mutated
```
onursumer: we should probably use an explicit constant value like `PROFILED_BUT_NOT_MUTATED` instead of `null`

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +538,208 @@ export function getCategoricalTrackRuleSetParams(
...
+                        sample.uniqueSampleKey,
+                        query.molecularProfileId,
+                        query.hugoGeneSymbol,
+                        coverageInfo
+                    );
+                })
+                .map(sample => ({
+                    uniqueSampleKey: sample.uniqueSampleKey,
+                    uniquePatientKey: sample.uniquePatientKey,
+                    value: null,
```
onursumer: same here. might be better to use an explicit constant instead of `null`

inodb:
@rishisulakhe Thank you for the fixes! Would be great if the VAF heatmap could respsect the "use white background" option:

image

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Added `PROFILED_BUT_NOT_MUTATED = null`

src/shared/components/oncoprint/DeltaUtils.ts
```diff
@@ -1587,6 +1627,33 @@ export function transitionHeatmapTrack(
...
+        // Rule set when white background flag changes for mutation (VAF) heatmap tracks
+        if (
+            nextSpec.molecularAlterationType ===
+                AlterationTypeConstants.MUTATION_EXTENDED &&
+            nextProps.isWhiteBackgroundForGlyphsEnabled !==
+                (prevProps as Partial<IOncoprintProps>)
+                    .isWhiteBackgroundForGlyphsEnabled
+        ) {
+            if (
+                typeof trackIdForRuleSetSharing.mutation !== 'undefined' &&
```
onursumer: do we really need `typeof` here? can't we just check for undefined?


```suggestion
                trackIdForRuleSetSharing.mutation !== undefined &&
```

src/shared/components/oncoprint/OncoprintUtils.ts
```diff
@@ -531,6 +549,228 @@ export function getCategoricalTrackRuleSetParams(
...
+
+        if (mutations && mutations.length > 0) {
+            const sampleMutationMap = new Map<string, number | null>();
+
+            mutations.forEach(m => {
+                const vafReport = getVariantAlleleFrequency(m);
+                if (
+                    vafReport &&
+                    typeof vafReport.vaf === 'number' &&
+                    !isNaN(vafReport.vaf)
```
onursumer: Probably better to replace this with a cleaner lodash call


```suggestion
                    _.isFinite(vafReport?.vaf)
```

onursumer (changes requested):
Looks good! Just a few more minor suggestions.

src/shared/components/oncoprint/OncoprintUtils.ts
rishisulakhe: Yes, lodash is much cleaner

PR #5240, opened 2025-07-24
Lazy load detailed sample treatment counts for comparison

onursumer:
Related to cBioPortal/cbioportal/issues/11618

TODO: We need to update the API client to make this work.


src/pages/studyView/StudyViewPageStore.ts
```diff
@@ -10474,6 +10476,28 @@ export class StudyViewPageStore
         },
     });
 
+    public readonly detailedSampleTreatments = remoteData<
```
alisman, line 10843: can you just add a comment explaining why we need this (to harvest sample ids for comparison session)

PR #5247, opened 2025-07-31
Fix study view pie chart numbers being cut off

gblaih:
Fix https://github.com/cBioPortal/cbioportal/issues/11636

If the numbers on a pie chart are too large, they will get cut off. To fix this, use K and M for thousand and million respectively in the number label so that it can fit inside the pie chart.


src/pages/studyView/charts/pieChart/PieChart.tsx
```diff
@@ -27,6 +27,21 @@ import { DEFAULT_NA_COLOR } from 'shared/lib/Colors';
 import ifNotDefined from '../../../../shared/lib/ifNotDefined';
 import { getTextColor } from 'pages/groupComparison/OverlapUtils';
 
+export function formatPieChartNumber(n: number) {
```
alisman, line 31: @gblaih there is a library we have called numeral.js which should do this.  lets see if we can leverage that.  i think you'll still need these conditions but at least not the math

gblaih: i've updated the pr to now use the numeral library

dippindots (approved):
LGTM

PR #5249, opened 2025-08-07
Revert prevent user from entering empty value

gblaih:
Revert https://github.com/cBioPortal/cbioportal-frontend/pull/5242 as https://github.com/cBioPortal/cbioportal/pull/11637 will fix issue


alisman:
@brettvanderwerff unless i misunderstand, i think this is still valuable.

PR #5273, opened 2025-10-07
feat: Add option to hide clinical data tab in study view

rosh2525:
This adds a portal property skin.hide_clinical_data_tab_study_view, which can hide the Clinical Data tab in the Study View page

Fixes cBioPortal/cbioportal#11653

- Added a new portal property, skin.hide_clinical_data_tab_study_view, to the IServerConfig interface.
- This property is used in StudyViewPage.tsx to conditionally hide the "Clinical Data" tab.
- Updated configuration loading to recognize window.portalProperties for local testing.

## Checks
- [ x] Has tests or has a separate issue that describes the types of test that should be created. If no test is included it should explicitly be mentioned in the PR why there is no test.

This is a minor visual change controlled by a configuration flag and does not require a dedicated test.

- [ x] The commit log is comprehensible. It follows [7 rules of great commit messages](http://chris.beams.io/posts/git-commit/). For most PRs a single commit should suffice, in some cases multiple topical commits can be useful. During review it is ok to see tiny commits (e.g. Fix reviewer comments), but right before the code gets merged to master or rc branch, any such commits should be squashed since they are useless to the other developers. Definitely avoid [merge commits, use rebase instead.](http://nathanleclaire.com/blog/2014/09/14/dont-be-scared-of-git-rebase/)
- [x ] Is this PR adding logic based on one or more **clinical** attributes? If yes, please make sure validation for this attribute is also present in the data validation / data loading layers (in backend repo) and documented in [File-Formats Clinical data section](https://github.com/cBioPortal/cbioportal/blob/master/docs/File-Formats.md#clinical-data)!

## Any screenshots or GIFs?

Screenshot 2025-10-05 at 3 32 32 PM
Screenshot 2025-10-05 at 3 34 17 PM

## Notify reviewers

### DIFF
--- a/src/config/IAppConfig.ts
+++ b/src/config/IAppConfig.ts
@@ -199,6 +199,5 @@ export interface IServerConfig {
     feature_study_export: boolean;
     uptime_robot_status_page_url: string | null;
     uptime_robot_api_key: string | null;
-    skin_hide_clinical_data_tab_study_view?: boolean;
     skin_hide_clinical_data_tab_study_view: boolean;
 }


src/config/IAppConfig.ts
```diff
@@ -190,4 +190,7 @@ export interface IServerConfig {
     clickhouse_mode: boolean;
     download_custom_buttons_json: string;
     feature_study_export: boolean;
+    skin?: {
```
alisman: we have a convention above that dot notation gets converted to _.  lets keep that

my-index.ejs
```diff
@@ -7,6 +7,7 @@
     <script>
 
         window.devContext = true;
+        window.portalProperties = { skin: { hide_clinical_data_tab_study_view: false } };
```
alisman: is this for testing purposes?

src/config/config.ts
```diff
@@ -351,6 +351,11 @@ export function initializeServerConfiguration(rawConfiguration: any) {
         }
     }
 
+    let portalProperties: any = {};
```
alisman: what's this for?

alisman (changes requested):
see comments

src/config/IAppConfig.ts
alisman: if you test it this, the backend should automatically convert the dot notation.  so i would think this wouldn't work.   let me know if you need hlep testing this

rosh2525:
Thanks for the review @alisman , Sorry for the late reply since I was busy with my semester exams and when I came back to my hometown I had frequent power cuts due to cyclone. I'll be active here forward to this

- After a better look, I reverted the changes to **src/config/config.ts** since the default loader handles the mapping automatically.
- I reverted **my-index.ejs** to remove the local testing
- I updated **IAppConfig.ts** to use skin_hide_clinical_data_tab_study_view (underscore notation)
- Fixed Rendering Logic: In **StudyViewPage.tsx**, I switched to using conditional rendering (const hideClinicalTab =getServerConfig().skin_hide_clinical_data_tab_study_view ?? false;) which is more reliable 

Kindly let me know if there are any more problems with this commit. Thanks !

PR #5284, opened 2025-10-23
add oncokb patient report tab to patient view page

bprize15:
- Added a tab showcasing the OncoKB Patient Report in the Patient View page.
- The tab will always show up so long as the feature flag ONCOKB_PATIENT_REPORT is set

Preview [link](https://deploy-preview-5284.preview.cbioportal.org/patient/oncokbKBPatientReport?studyId=msk_chord_2024&caseId=P-0000012)

TODO:
- [ ] Get Product Review
- [x] Update description to reflect when this tab shows up (is it only in this PR or everywhere?)


src/pages/patientView/oncokb/PatientReportTab.tsx
```diff
@@ -0,0 +1,277 @@
...
+                            clinicalData.PRIMARY_SITE;
+                        sample.sampleMetadata.metastaticSite =
+                            clinicalData.METASTATIC_SITE;
+                        sample.sampleMetadata.specimenType =
+                            clinicalData.SPECIMEN_TYPE;
+                        samples.push(sample);
+                    }
+                }
+
+                const form = document.createElement('form');
```
alisman, line 56: Hey @bprize15 how come you are creating a form this way rather than through React/JSX?

alisman: i see. this to create a post request to the iframe?

src/pages/patientView/oncokb/PatientReportTab.tsx
```diff
@@ -0,0 +1,277 @@
...
+import { ClinicalData } from 'cbioportal-ts-api-client';
+import WindowStore from 'shared/components/window/WindowStore';
+import IFrameLoader from 'shared/components/iframeLoader/IFrameLoader';
+import { observer } from 'mobx-react';
+
+interface IPatientReportTabProps {
+    patientViewPageStore: PatientViewPageStore;
+}
+
+class PatientMetadata {
```
alisman, line 13: i think these can be types instead of classes.  also, lets put these in a ts namespace so that we don't expose them outside of this file since they are specific to the interface of whatever's getting loaded in the iframe.

src/pages/patientView/oncokb/PatientReportTab.tsx
```diff
@@ -0,0 +1,277 @@
...
+            patientViewPageStore.cnaOncoKbData.isComplete &&
+            patientViewPageStore.structuralVariantData.isComplete &&
+            patientViewPageStore.structuralVariantOncoKbData.isComplete;
+
+        useEffect(() => {
+            if (isDataComplete) {
+                const clinicalDataGroupedBySample =
+                    patientViewPageStore.clinicalDataGroupedBySample.result;
+
+                const patientMetadata = new PatientMetadata();
```
alisman, line 36: lets put this code in a utitility file somewhere and add some comments about what it does.

alisman: i mean the whole contents of the this useEffect hook.  be careful though to dereference data from the mobx promises before you pass them into this function

src/pages/patientView/oncokb/PatientReportTab.tsx
```diff
@@ -0,0 +1,277 @@
...
+    } = {};
+    mutations: IndicatorQueryResp[] = [];
+    copyNumberAlterations: IndicatorQueryResp[] = [];
+    structuralVariants: IndicatorQueryResp[] = [];
+}
+
+const IFRAME_NAME = 'patient-report-iframe';
+
+const PatientReportTab = observer(
+    ({ patientViewPageStore }: IPatientReportTabProps) => {
```
alisman, line 23: you should type this using React.FunctionComponent generic

src/pages/patientView/oncokb/PatientReportTab.tsx
bprize15: Exactly, it's the only way to use an iframe with a post request for a single page app.

src/pages/patientView/oncokb/PatientReportTab.tsx
bprize15: Wondering if we want to do this since React does not recommend. It's a rather trivial change so I'm find with it if it helps maintain the style of the codebase, but wondering your thoughts on this: https://github.com/facebook/create-react-app/pull/8177

src/pages/patientView/oncokb/PatientReportTab.tsx
bprize15: I made them classes so I could use default values, since the OncoKB API expects some of these fields to be non-null. I can make them types and remove the default values, but wondering if it might be simpler this way since a user does not need to remember to instantiate some fields?

Also, I'm a little confused on the namespace idea. Since I'm not exporting the types, they are not exposed outside this file. 

I could do something like this:

```typescript
namespace PatientReport {
    export class PatientMetadata {}
    export class AnnotatedSample {}
}
```

But this would not be any safer than the previous approach. The reason this does not leak is the same reason mine currently does not leak: the namespace is not exported.

alisman: Ah, ok.  I assumed you were exporting them but i see that you are right.

src/pages/patientView/oncokb/PatientReportUtils.ts
```diff
@@ -0,0 +1,296 @@
...
+}
+
+function constructIndicatorQueryResp(
+    alt: Mutation | DiscreteCopyNumberData | StructuralVariant
+): IndicatorQueryResp {
+    const isStructuralVariant = 'site1HugoSymbol' in alt;
+
+    let hugoSymbol = '';
+    let entrezGeneId = 0;
+    if ('gene' in alt) {
```
alisman, line 221: i believe that typescript will not check this.  i.e. you could put ```'banana' in alt```.  better of just doing 
```alt.gene```

File: src/pages/patientView/oncokb/PatientReportUtils.ts [hunk 6]
```diff
@@ -0,0 +1,296 @@
+import {
```
alisman, line 1: @bprize15 this file could use some summary comments and testing.  Claude would do very well helping to build add a few tests.

bprize15: Added doc comment and test

alisman (approved):

alisman:
@bprize15 @inodb what's status of the project?  should we merge this?

PR #5439, opened 2026-03-12
fix(plots): handle mixed generic assay datatypes per genericAssayType (Fixes #11763)

### LINKED ISSUE #11763: [BUG] Generic assay profiles are plotted with a uniform datatype regardless of the actual datatype
**Describe the bug**
When you have multiple generic assay profiles attached to the same `generic_assay_type`, and with different `datatype` (e.g. one `LIMIT-VALUE` and another `CATEGORICAL`), in the plots they will all be visualized in one datatype, instead of looking at the `datatype` of the generic assay profile.

**To Reproduce**
1. Create a study with a duplicated generic assay profile, e.g. with the `datatype` `CATEGORICAL`. Assign one to `generic_assay_type` "Type 1" and the other to "Type 2".
2. Create another generic assay profile with the datatype `LIMIT-VALUE` and assign it to the `generic_assay_type` "Type 1".

When you query the plots in the plots tab you should see a difference on how the duplicated generic assay profile is treated (see screenshots).

**Expected behavior**
The generic assay profiles should be plotted based on their `datatype`.

**Screenshots**
As an example, I've duplicated `treatment_ic50` from `study_es_0` and assigned both profiles to `CATEGORICAL` datatype. Then, I have added one from the two ic50 profiles to `TREATMENT_RESPONSE_CATEGORICAL` generic_assay_type.

The ic50 under `TREATMENT_RESPONSE_CATEGORICAL` handles the plot as `CATEGORICAL`:

Image

The ic50 under `TREATMENT_RESPONSE` handles the plot as `LIMIT-VALUE`, even though in the meta file is set as `CATEGORICAL`:

Image

VarshiniGunti:
Fix cBioPortal/cbioportal#11763

- Fixed the root cause in Plots Generic Assay datatype option construction.
- Previously, options were deduplicated only by `genericAssayType`, which incorrectly collapsed multiple profiles of the same type but different datatypes into a single option.
- Updated deduplication logic to use `genericAssayType + datatype`, so each datatype variant is preserved.
- This ensures plotting behavior is aligned with the selected profile’s actual datatype (e.g., `CATEGORICAL` vs `LIMIT-VALUE`) instead of applying a uniform datatype across that generic assay type.

## Why this is impactful
- Resolves incorrect visualization behavior for mixed-datatype Generic Assay profiles.
- Prevents silent data-mode mismatch in Plots for studies where one `genericAssayType` contains multiple datatypes.
- Improves correctness and trust in Plots interpretation for Generic Assay data.

## Files changed
- `src/shared/components/plots/PlotsTab.tsx`

## Checks
  - Ran: `GREP=PlotsTab yarn run testMain`
  - Passed:
    - `src/shared/components/plots/PlotsTabUtils.spec.ts`
    - `src/shared/components/plots/LastPlotsTabSelectionForDatatype.spec.ts`
    - `src/shared/components/plots/PlotsTab.spec.tsx`
  - No.

## Any screenshots or GIFs?
Screenshot 2026-03-13 104627
plot

### Implementation Summary

Updated Generic Assay option deduplication in `PlotsTab.tsx`:
```ts
- old:
.uniqBy(profile => profile.genericAssayType)
- new:
.uniqBy(profile => `${profile.genericAssayType}|${profile.datatype || ''}`)

This prevents profiles with the same genericAssayType but different datatype from collapsing into a single option.


VarshiniGunti:
> @VarshiniGunti I think your changes are disappeared during the rebase, could you take a look?

Thanks for flagging this @dippindots, you were right, the #11763 change got dropped during rebase.
I restored it and pushed commit `80740d569`.
Confirmed `PlotsTab.tsx` now deduplicates by `genericAssayType + datatype`:
`uniqBy(profile => \`${profile.genericAssayType}|${profile.datatype || ''}\`)`
Could you please re-review?

dippindots (commented):
Thanks @VarshiniGunti! The `uniqBy` change makes sense, but I have a couple of questions:
1. Since the `value` in the mapped options is still just `profile.genericAssayType`, wouldn't two options end up with the same value when they share a type but differ in datatype? What would happen when the user tries to select the second one?
2. Similarly, the labels would also be identical — how would users tell them apart in the dropdown?

Could you try this out with a study that has mixed datatypes for the same generic assay type and share what the dropdown looks like?

VarshiniGunti:
> @VarshiniGunti could you look at my questions here? [#5439 (review)](https://github.com/cBioPortal/cbioportal-frontend/pull/5439#pullrequestreview-4007948209)

Use this reply:

@dippindots Thanks for the detailed questions, these are valid concerns. You’re right that duplicate `value`/label handling needs to be addressed properly for mixed datatypes; I’m working on an update to make selection and labeling unambiguous, and I’ll push the fix shortly.

PR #5444, opened 2026-03-17
Add AlphaFold guidance to structure viewer UI

Rana-Ezzeddine:
Summary
This change adds AlphaFold-aware guidance to the existing structure viewer UI.

What changed
- add a small AlphaFold DB URL helper
- show an AlphaFold DB link in the protein chain panel when a UniProt entry is available
- clarify that the panel lists experimental PDB chains only
- update the disabled 3D structure tooltip to point users to AlphaFold DB when no experimental PDB chains are available

Why
The current structure UI is strongly PDB-oriented. This improves the user experience for cases where no experimental structure is available, while also laying relevant groundwork for future AlphaFold integration.


dippindots:
We have recently fixed the 3d viewer and will refactor with AlphaFold properly in a GSoC project, we will close this PR for now. Thanks for trying to add this improvement!

PR #5451, opened 2026-03-20
Fix single-study detection (Issue #10445)

### LINKED ISSUE #10445: Make custom data work without study_id if one study
image
The current interface requires adding `studyId:` as a prefix. When one is looking at a single study users shouldn't have to supply the study id

tangibletang:
Fixes cBioPortal/cbioportal#10445

This PR clarifies the custom case selection UX for single-study views. In practice, users are not always required to provide a study_id: prefix when only one study is in scope, but the existing placeholder/help examples still display study_id: and make it look mandatory. This mismatch is confusing and suggests a stricter input requirement than intended. The update aligns the examples/help text with actual single-study behavior while preserving the existing multi-study format where study_id: remains appropriate.

### What changed
- Single-study detection now uses unique non-empty study IDs, so duplicate entries for the same study no longer force multi-study behavior.
- Example/help content now adapts to context:
  - **Single study:** no `study_id:` prefix implied.
  - **Multiple studies:** existing `study_id:` format remains unchanged.
- Generated custom case lines stay compatible with existing parser behavior in both modes.

### Testing
Manually validated in Study View for both scenarios (single-study and multi-study).  
No automated test was added in this PR (UI/formatting behavior adjustment); a focused follow-up test can be added for placeholder/help toggling by study-count context.

### Screenshots
1. **Single study selected**: example no longer shows `study_id:` prefix.
Screenshot 2026-03-19 at 8 34 21 PM

2. **Multiple studies selected**: example still shows `study_id:` prefix.
Screenshot 2026-03-19 at 8 34 32 PM

### DIFF
--- a/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
+++ b/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
@@ -88,7 +88,7 @@ export default class CustomCaseSelection extends React.Component<
 
     @computed
     get isSingleStudy() {
-        return this.props.queriedStudies.length === 1;
+        return _.uniq(this.props.queriedStudies.filter(Boolean)).length === 1;
     }
 
     @computed
@@ -139,14 +139,19 @@ export default class CustomCaseSelection extends React.Component<
             });
         }
         let cases = selectedCases.map(sample => {
-            return `${sample.studyId}:${
+            const caseId =
                 this.caseIdsMode === ClinicalDataTypeEnum.SAMPLE
                     ? sample.sampleId
-                    : sample.patientId
-            }${
-                this.props.disableGrouping
-                    ? ''
-                    : ` ${DEFAULT_GROUP_NAME_WITHOUT_USER_INPUT}`
+                    : sample.patientId;
+            // For a single study, users can omit `study_id:`. We represent
+            // "omitted study id" with a leading `:` so the parser can still
+            // extract the case id and the optional group value.
+            const caseIdWithOptionalStudyPrefix = this.isSingleStudy
+                ? `:${caseId}`
+                : `${sample.studyId}:${caseId}`;
+
+            return `${caseIdWithOptionalStudyPrefix}${
+                this.props.disableGrouping ? '' : ` ${DEFAULT_GROUP_NAME_WITHOUT_USER_INPUT}`
             }`;
         });
         if (this.caseIdsMode === ClinicalDataTypeEnum.PATIENT) {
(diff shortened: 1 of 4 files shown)


src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
```diff
@@ -88,7 +88,7 @@ export default class CustomCaseSelection extends React.Component<
 
     @computed
     get isSingleStudy() {
-        return this.props.queriedStudies.length === 1;
+        return _.uniq(this.props.queriedStudies.filter(Boolean)).length === 1;
```
dippindots: why not just `_.uniq(this.props.queriedStudies).length === 1;`?

is `.filter(Boolean)` trying to filter some edge cases?

dippindots:
@tangibletang have you tried this locally, not sure if it's preview instance's issue, but when we try custom selection for a single study, we see this error:
`TypeError: Cannot read properties of undefined (reading 'sampleId')`

src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
tangibletang: Thanks for catching that!

The .filter(Boolean) was defensive coding on my part. I realize now queriedStudies only ever contains real study IDs from the API so it's never needed. Removed.


Regarding the the TypeError: I think it was running on my local host earlier but I double checked and found the root cause. When auto-populating in single-study mode, the parser was producing studyId = "" (empty string) which getData() wasn't handling. Fixed that and added a regression test to cover it. 

Let me know if this is fine otherwise!

src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
```diff
@@ -139,14 +139,19 @@ export default class CustomCaseSelection extends React.Component<
...
-                    : ` ${DEFAULT_GROUP_NAME_WITHOUT_USER_INPUT}`
+                    : sample.patientId;
+            // For a single study, users can omit `study_id:`. We represent
+            // "omitted study id" with a leading `:` so the parser can still
+            // extract the case id and the optional group value.
+            const caseIdWithOptionalStudyPrefix = this.isSingleStudy
+                ? `:${caseId}`
+                : `${sample.studyId}:${caseId}`;
+
+            return `${caseIdWithOptionalStudyPrefix}${
```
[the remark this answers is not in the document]
alisman: @tangibletang can you address this?  We tested and it is still a problem.

alisman:
@tangibletang thank you for this PR.  would you mind adding a logical end-to-end test for it?  logical means that it asserts against the DOM (instead of taking screenshot).   Let me know if you need help.

tangibletang:
> [ ](/tangibletang)

Hi @alisman ,I added CustomCaseSelection.spec.tsx with tests that check the placeholder text shows the correct format in single-study vs multi-study mode. To run: yarn test --testPathPattern="CustomCaseSelection.spec"


Please let me know if this is what you were looking for!

alisman:
@tangibletang we just tested this again and we found one last thing to fix:

When you populate (click 'currently selected button) the custom selection in the study view, the IDs show up with leading colon:

image

Can we address this?

dippindots (approved):
LGTM, @onursumer @alisman all tests are passing. Also tested this on multiple studies and single study scenario, I will merge it

dippindots:
@tangibletang Thanks for making this enhancement!

PR #5455, opened 2026-03-20
Fix single-study detection for custom case example placeholder (Issue #115592)

tangibletang:
## Summary

Fixes a patient-view inconsistency where expression heatmap z-scores and `High/Low` summary labels could disagree when a patient had multiple samples. (https://github.com/cBioPortal/cbioportal/issues/11592)

In patient mode, both display outputs now follow the same representative-selection path:
- heatmap value and color
- expression summary subtype (`mRNA` / `protein` high/low)

This keeps the existing representative aggregation model (no averaging/median change), but removes cross-signal contradictions by using one consistent source of truth.

## What Changed

- Refactored patient-level representative heatmap datum selection into a reusable helper in `DataUtils`.
- Reused that same representative selection path when deriving patient-view expression summary subtype (`disp_mrna` / `disp_prot`).
- Preserved existing representative and sort-order semantics (`ASC` / `DESC` / default), while removing label-vs-z-score divergence in patient mode.

## Rationale

Previously, patient heatmap values used representative sample selection, while `High/Low` labels were derived from separate subtype-count aggregation.  
With opposing sample signals, this could produce mismatches (e.g., negative z-score shown with a `High` label).

We chose to align both outputs to the existing representative path (instead of switching to a new averaging/median strategy) because:
- it fixes the inconsistency without redefining established patient-level semantics,
- it preserves current sorting and representative behavior users already rely on,
- it minimizes scope/risk for regression, while making displayed z-score and `High/Low` interpretation consistent.

## Scope

- Patient view logic only
- No intended behavior change for sample view
- No change to overall representative strategy (still representative, not average)

## Validation

- Confirmed patient-view z-score sign and `High/Low` label now align under representative selection.
- Ran OncoPrint-focused test subset and targeted DataUtils test locally.

## Media

- Attached GIF demonstrating patient-view z-score/label alignment before the fix. In patient view, the displayed heatmap z-score and the expression summary subtype label (mRNA Low/High) could be out of sync.
cbiobeforegif-ezgif com-video-to-gif-converter

- Attached GIF demonstrating patient-view z-score/label alignment after the fix. Both label and mRNA z-score are derived from the same representative signal, so the z-score sign and Low/High label are aligned.

cbiogifafter-ezgif com-video-to-gif-converter

### DIFF
--- a/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
+++ b/src/pages/studyView/addChartButton/customCaseSelection/CustomCaseSelection.tsx
@@ -88,7 +88,7 @@ export default class CustomCaseSelection extends React.Component<
 
     @computed
     get isSingleStudy() {
-        return this.props.queriedStudies.length === 1;
+        return _.uniq(this.props.queriedStudies.filter(Boolean)).length === 1;
     }
 
     @computed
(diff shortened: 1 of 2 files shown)


alisman:
@tangibletang this somehow broke the custom sample/patient list functionality on study view. 

click this button and you'll see error: 

image

alisman:
@tangibletang i think you included commits from another PR here because the title of the PR doesn't seem to align with the content

PR #5508, opened 2026-04-12
Migrate from yarn/webpack to pnpm/rspack

inodb:
## Summary

Migrate the frontend build toolchain:

- **yarn → pnpm** — content-addressable store hardlinks across worktrees (no per-worktree `node_modules` duplication), and strict layout surfaces phantom dependencies.
- **webpack → rspack** — drop-in replacement with the same config API. Low-risk path to a faster build; a larger tooling evaluation (Vite/esbuild/Turbopack) is tracked separately in cBioPortal/cbioportal#12109.
- **+ turborepo** — content-hash cache for the 10 workspace package builds under `buildModules`, which otherwise rebuild every run.

We keep the two-chunk vendor split (`common.bundle.js` + `main.app.js`) and `styles.css` filenames so the cBioPortal backend's hardcoded `<script>`/`<link>` tags keep working. The separate webpack DLL build is gone — rspack's splitChunks emits both files from one pass.

Along the way, declared a handful of phantom dependencies that yarn's flat hoist was silently satisfying (`rimraf`, `clipboard`, `cytoscape-navigator`, `cytoscape-panzoom`, `jspdf-yworks`, `qs`, `react-resizable`, `react-table`, `juice`, `gl-matrix`), pinned a few packages via `pnpm.overrides` to match master's resolved versions, and regenerated the e2e screenshot references against the new bundler output.

### TypeScript is now gated on the main build

`build_frontend` runs with `TRANSPILE_ONLY=true` so `ts-loader` skips type-checking in the rspack pass. `fork-ts-checker-webpack-plugin` now runs in all modes (previously dev/test only), so in production it runs synchronously and fails the build on type errors. The separate `check_typescript` CircleCI job and the `checkTS` npm script are removed — both redundant now (and `checkTS` was broken anyway: `;`-chained `tsc` invocations swallowed all but the last exit code).

### Getting started

```bash
corepack enable
pnpm install
pnpm run buildModules
```

### Deploy preview

https://deploy-preview-5508.preview.cbioportal.org — verified homepage, study summary, patient view, cnSegments, mutation table.

### Related

- cBioPortal/cbioportal#12108 — follow-up: upgrade prettier 1.19.1 → 3.x
- cBioPortal/cbioportal#12109 — follow-up: build tooling evaluation (esbuild, Vitest, Playwright, remote turbo cache)

🤖 Generated with [Claude Code](https://claude.com/claude-code)


packages/cbioportal-clinical-timeline/package.json
```diff
@@ -18,28 +17,28 @@
...
     "mobx-react": "^6.0.0",
-    "mobx-react-lite": "^2.0.0",
+    "mobx-react-lite": "^3.0.0",
     "react": "^15.0.0 || ^16.0.0",
     "react-dom": "^15.0.0 || ^16.0.0"
   },
   "dependencies": {
     "autobind-decorator": "^2.1.0",
-    "cbioportal-frontend-commons": "^0.6.2",
+    "cbioportal-frontend-commons": "workspace:*",
```
onursumer, line 41: this may break lerna's publishing action and create issues with npm publishing. is there no way to keep the versioning as is? this question applies to other internal packages as well.

onursumer: hmm, this might be okay. looks like lerna started supporting this starting v8. I guess we can figure it out after merging this pr.

PR #5525, opened 2026-04-14
Heatmap: empty cell for contradictory patient samples (alt to #5225)

alisman:
## Summary

Alternative fix for [#11592](https://github.com/cBioPortal/cbioportal/issues/11592) / replacement for #5225.

When multiple samples for a single patient have **contradictory** mRNA values — i.e. a mix of positive and negative values — collapsing them into one representative number (either by highest absolute value or by average) misrepresents the patient-level state and hides the conflict from the user.

This PR changes the behavior to:

- Render the patient's heatmap cell as **empty** (`profile_data = null`) when the sample values for that patient include both a positive and a negative value.
- Surface **all** underlying sample values in the cell's tooltip so the user can see the conflict directly (no averaging, no arbitrary pick).

Non-contradictory cases (all positive, all negative, or all zero/same-sign) continue to use the existing aggregation logic unchanged.

## Changes

- `DataUtils.ts`: new `hasContradictoryValues` helper; patient-mode branch in `fillHeatmapTrackDatum` short-circuits to an empty cell with a new `contradictoryValues: number[]` field when signs are mixed.
- `Oncoprint.tsx`: adds optional `contradictoryValues` to `IBaseHeatmapTrackDatum`.
- `TooltipUtils.ts`: `makeHeatmapTrackTooltip` now renders the list of contradictory values instead of the (absent) aggregated value.
- Tests updated: the existing `[-10, 3, 4]` patient-aggregation case now asserts the new empty-cell + `contradictoryValues` behavior; new cases cover mixed-sign, zero/same-sign, and a threshold tie-break case reworked to avoid contradictory input.

## Test plan

- [x] `jest src/shared/components/oncoprint/DataUtils.spec.ts` passes (111 tests).
- [x] `jest src/shared/components/oncoprint/TooltipUtils.spec.ts` passes.
- [ ] Manual: load a study with multiple samples per patient where mRNA z-scores have mixed signs; verify the heatmap cell renders empty and the tooltip lists every sample value.
- [ ] Manual: verify non-contradictory patients still render as before.

### DIFF
--- a/src/shared/components/oncoprint/DataUtils.spec.ts
+++ b/src/shared/components/oncoprint/DataUtils.spec.ts
@@ -1663,6 +1663,8 @@ describe('DataUtils', () => {
                 }
             );
 
+            // contradictory values (mix of positive and negative) are reported
+            // via `contradictoryValues` with an empty cell (profile_data: null)
             data = [{ value: -10 }, { value: 3 }, { value: 4 }];
             assert.deepEqual(
                 fillHeatmapTrackDatum<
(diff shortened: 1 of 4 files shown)


alisman:
@copilot it seems as though for each patient you are considering ALL the sample values instead of just those belonging to the patient when determining the conflicting values collection.  can you confirm this and fix?

dippindots:
@inodb any suggestions on this solution @alisman proposed. We are not sure about the label and text we are using here: `contradictory values` and `Not a number`
image

PR #5563, opened 2026-05-07
PoC: extract batched-fetch utility, migrate VariantCountCache

alisman:
## Proof of concept — not for merge as-is

This PR is a **design-review proposal**, not a complete migration. It introduces a small replacement utility for `LazyMobXCache` and migrates exactly one cache (`VariantCountCache`) as a worked example. Goal: surface the proposed API and migration pattern before committing to the long tail (~19 more subclasses, ~50+ call sites).

## Why

Profiling against production cBioPortal confirmed that the per-row column-formatter batching in `LazyMobXCache` is doing real work:

- Patient view (10 mutation rows visible) → **1** `variant-counts/fetch` request with **10** entries in body
- Results view (30 genes, MSK-IMPACT) → **1** `mutations/fetch` with all 30 genes batched, median batch sizes of 25–30 across other endpoints

So we can't simply delete the caches. But the rest of `LazyMobXCache`'s surface (`peek`, `getPromise`, `addData`, `cache` reactive read, error retry tracking, `meta`-augmented data, `MobxPromise` integration, ~20 thin subclass wrappers) is incidental — most call sites only use `.get()`.

This PR proposes: **keep the batching, drop the rest**.

## What's here

**New** — `src/shared/lib/batchedReactiveFetch.ts` (~50 lines):
```ts
export function createBatchedReactiveFetch<Q, V>(args: {
    fetch: (queries: Q[]) => Promise<V[]>;
    queryToKey: (q: Q) => string;
    dataToKey: (d: V) => string;
}): ReactiveFetchCache<Q, V>
```
A factory that returns `{ get(query) → ReactiveFetchEntry<V> | null }`. All `.get(...)` calls within a single microtask tick collapse into one fetch invocation. Internal state is a single `observable.box`, so mobx observers re-render when results land. 7 unit tests covering: N→1 collapse, reactive updates, in-flight dedup, error termination, cross-tick separation, missing-result handling.

**Migrated** — `VariantCountCache`: was a `LazyMobXCache` subclass, now a 30-line factory function. Same `.get({entrezGeneId, keyword})` shape at the call site. Three downstream files (`CohortColumnFormatter`, `MutationTable`, `PatientViewPageStore`) updated to consume the new type alias.

**Untouched** — `LazyMobXCache.ts`, the 18 other subclasses, `SampleGeneCache`, `SimpleCache`. They keep working as before. This PR adds the utility next to the old infra, doesn't replace it.

## Diff size
6 files, +285 / -34 lines. Most of that is the new utility and its test file.

## What I'd want from review
1. **Does the utility's API feel right?** Anything missing that would force the next migration to leak abstractions? The `meta`-augmented-data pattern (used by e.g. `MutationDataCache`) is a deliberate omission — I don't think it should bleed into the base utility, but worth confirming.
2. **`status: 'pending'` vs `null` on cache miss.** Old code returned `null` for both \"not yet requested\" and \"in flight\"; new code does the same. I dropped a `pending` variant from the type union since it was never observable — let me know if any caller wants to distinguish.
3. **Microtask-tick batching window** vs. the current `accumulatingDebounce` (delay 0). Behaviorally equivalent in practice (both fire on the next macrotask boundary), but if the timing nuance matters anywhere, flag it.

## Test plan
- [x] `pnpm typecheck` clean
- [x] 7 new unit tests in `batchedReactiveFetch.spec.ts` pass
- [x] Existing `LazyMobXCache.spec.ts` still passes (old infra unchanged)
- [x] `CohortColumnFormatter` specs (both variants) pass
- [ ] Live manual smoke test: open a patient view, confirm Cohort column still renders frequencies correctly (could not verify in CI; needs reviewer to click through)
- [ ] CI green

## If approved, follow-up PRs
Each migrating one or two caches at a time, in this rough order of leverage:
1. Other per-row hot paths: `MrnaExprRankCache`, `DiscreteCNACache`, `MutationCountCache`, `CopyNumberCountCache`
2. `getPromise`-using sites (oncoprint/plots): need a small `getPromise`-style helper or inlined `Promise.all`
3. Remaining incidental caches: most can be inlined directly

Once all subclasses are migrated, `LazyMobXCache.ts`, `SampleGeneCache.ts`, `SimpleCache.ts`, and the 19 wrapper subclass files can be deleted in a final cleanup PR. Estimated total: 1–2 weeks of focused work spread across ~10 reviewable PRs.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---
<a href="https://app.blacksmith.sh/cBioPortal/codesmith/cbioportal-frontend/pr/5563"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

### DIFF
--- a/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
+++ b/src/pages/patientView/clinicalInformation/PatientViewPageStore.ts
@@ -52,7 +52,7 @@ import {
     USE_DEFAULT_PUBLIC_INSTANCE_FOR_ONCOKB,
 } from 'react-mutation-mapper';
 import { ClinicalInformationData } from 'shared/model/ClinicalInformation';
-import VariantCountCache from 'shared/cache/VariantCountCache';
+import { createVariantCountCache } from 'shared/cache/VariantCountCache';
 import CopyNumberCountCache from './CopyNumberCountCache';
 import CancerTypeCache from 'shared/cache/CancerTypeCache';
 import MutationCountCache from 'shared/cache/MutationCountCache';
@@ -2435,7 +2435,7 @@ export class PatientViewPageStore {
     }
 
     @cached @computed get variantCountCache() {
-        return new VariantCountCache(this.mutationMolecularProfileId.result);
+        return createVariantCountCache(this.mutationMolecularProfileId.result);
     }
 
     @cached @computed get discreteCNACache() {
--- a/src/shared/cache/VariantCountCache.ts
+++ b/src/shared/cache/VariantCountCache.ts
@@ -1,38 +1,38 @@
-import LazyMobXCache from 'shared/lib/LazyMobXCache';
 import client from 'shared/api/cbioportalInternalClientInstance';
 import { VariantCount, VariantCountIdentifier } from 'cbioportal-ts-api-client';
+import {
+    createBatchedReactiveFetch,
+    ReactiveFetchCache,
+} from 'shared/lib/batchedReactiveFetch';
+
+export type VariantCountCache = ReactiveFetchCache<
+    VariantCountIdentifier,
+    VariantCount
+>;
 
 function getKey<T extends { entrezGeneId: number; keyword?: string }>(
     obj: T
 ): string {
-    if (obj.keyword) {
-        return obj.entrezGeneId + '~' + obj.keyword;
-    } else {
-        return obj.entrezGeneId + '';
-    }
+    return obj.keyword
+        ? `${obj.entrezGeneId}~${obj.keyword}`
+        : `${obj.entrezGeneId}`;
 }
 
-function fetch(
-    queries: VariantCountIdentifier[],
+export function createVariantCountCache(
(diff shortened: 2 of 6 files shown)


alisman:
Superseding with a layered primitives version per discussion — see follow-up PR. The fused `createBatchedReactiveFetch` here works but bakes batching, memoization, and MobX reactivity into one helper. Splitting into three composable primitives (`batchify` + `memoize` + `reactive`) is cleaner — each primitive is independently testable, replaceable, and matches a recognizable industry pattern (DataLoader).

PR #5628, opened 2026-06-23
Stabilize oncoprint generic assay categorical tracks localdb e2e test

gblaih:
### Issue

Oncoprint screenshot tests for generic assay categorical tracks were failing non-deterministically. Two related problems caused this:

 1. Non-deterministic color assignment – oncoprintjs's CategoricalRuleSet assigns colors lazily, in the order columns are first rendered. Because each track restarts from palette index 0, sibling tracks within the same molecular profile would end up sharing colors, and the assignment order could vary between runs, causing screenshot mismatches.
 2. Dimension inconsistency – An unrelated layout issue caused the oncoprint canvas dimensions to be inconsistent across renders, further destabilizing the screenshots.

### Fix

 - Stable color pre-assignment: getCategoricalTrackRuleSetParams now accepts all tracks for a given molecular profile. It pre-builds a deterministic category_to_color map by sorting all unique category values alphabetically across sibling tracks and assigning them colors from the oncoprintjs palette in order, skipping any already claimed by RESERVED_CLINICAL_VALUE_COLORS. This map is passed to the CategoricalRuleSet so colors are fixed before any lazy render-order assignment can occur.
 - Track context threading: transitionCategoricalTrack in DeltaUtils.ts now filters nextProps.categoricalTracks to the relevant profile and passes them to getCategoricalTrackRuleSetParams.
 - Dimension fix: A one-line correction in OncoprintUtils.ts resolves the canvas dimension inconsistency.
 - Test stabilization: The oncoprint.ts helper and affected screenshot spec files were updated to wait for stable render state, and the reference screenshot was refreshed.


alisman:
@gblaih this one scares me cuz changes to source (rather than just test).  lets discuss tomorrow

alisman:
@dippindots @gblaih i'm gonna close this because i think it should be fixed at api level.  it's most likely due to some non-deterministic ordering introduced by clickhouse and should be fixerd server side with order by clause

PR #5669, opened 2026-08-06
Fix Download tab crash for studies with no mutation/CNA profiles

### LINKED ISSUE #12284: user reported results/download page will not load
User reported issue: https://groups.google.com/g/cbioportal/c/yF-z4q5U1IE/m/cPaMlr1DBwAJ

If I click the link in the user provided error message, I also get an error: https://www.cbioportal.org/results/download?cancer_study_list=ovary_geomx_gray_foundation_2024&tab_index=tab_visualize&case_set_id=ovary_geomx_gray_foundation_2024_all&Action=Submit&gene_list=GRB2%250ANEDD9%250ALRRK1%250ALYN%250ADOK1%250ASTAT5B%250ATJP2%250APTPN1

I suspect that the issue might be that this study only has mRNA profile (no mutations/CNA). There is no profileFilter in the URL, which from my limited testing makes me think the query was initiated from study view. If I try to run a query from study view for this study, I get a similar looking URL: 
https://www.cbioportal.org/results?cancer_study_list=ovary_geomx_gray_foundation_2024&tab_index=tab_visualize&case_set_id=ovary_geomx_gray_foundation_2024_all&Action=Submit&gene_list=GRB2
The OncoPrint loads, but if I click over to the Download tab (as in the original link), I get an error (although a slightly different error than the user reported)

dippindots:
Fix https://github.com/cBioPortal/cbioportal/issues/12284
### Problem

The Results View Download tab crashes for studies with no mutation/CNA (or other alteration-type) profile — e.g. mRNA-only studies like `ovary_geomx_gray_foundation_2024`. Reported [here](https://groups.google.com/g/cbioportal/c/yF-z4q5U1IE/m/cPaMlr1DBwAJ).

### Cause

In such studies, genes/tracks queried without explicit OQL syntax (e.g. bare gene names) have no default alteration type, so `CaseAlterationTable.tsx`'s `geneAlterationTypesMap` and `trackAlterationTypesMap` have no entry for them. Two spots in `CaseAlterationTable.tsx` assumed an entry always exists and crashed on the undefined value (`.forEach` and `.length`).

### Fix

Default both lookups to an empty array when missing, so genes/tracks with no alteration type simply contribute no extra columns instead of crashing.

### DIFF
--- a/src/pages/resultsView/download/CaseAlterationTable.tsx
+++ b/src/pages/resultsView/download/CaseAlterationTable.tsx
@@ -444,63 +444,72 @@ export default class CaseAlterationTable extends React.Component<
 
         _.forEach(uniqGenes, gene => {
             //add column for each gene alteration combination
-            this.props.geneAlterationTypesMap[gene].forEach(alteration => {
-                const oql_line = parsedOQLAlterationToSourceOQL(alteration);
-                const alterationType = alteration.alteration_type.toUpperCase();
-                const alterationName = `${gene}: ${oql_line}`;
-                if (
-                    _.isEmpty(
-                        columns.find(column => column.name === alterationName)
-                    )
-                ) {
-                    columns.push({
-                        name: `${gene}: ${oql_line}`,
-                        headerDownload: (name: string) =>
-                            `${gene}: ${oql_line}`,
-                        render: (data: ICaseAlteration) => {
-                            const pseudoOqlSummary = generatePseudoOqlSummary(
-                                data.oqlDataByGene,
-                                gene,
-                                alterationType
-                            );
+            // genes queried in studies with no mutation/CNA/other alteration
+            // profiles have no entry here, since there is no alteration type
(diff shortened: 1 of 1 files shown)


src/pages/resultsView/download/CaseAlterationTable.tsx
```diff
@@ -444,63 +444,72 @@ export default class CaseAlterationTable extends React.Component<
...
-                        render: (data: ICaseAlteration) => {
-                            const pseudoOqlSummary = generatePseudoOqlSummary(
-                                data.oqlDataByGene,
-                                gene,
-                                alterationType
-                            );
+            // genes queried in studies with no mutation/CNA/other alteration
+            // profiles have no entry here, since there is no alteration type
+            // to build a pseudo-OQL column for
+            (this.props.geneAlterationTypesMap[gene] || []).forEach(
```
dippindots, line 453: This is the only change in this PR, all other lines are format related

PR #5697, opened 2026-09-06
Contain Genome Nexus annotation render failures with an error boundary

Prasad8830:
## Summary

Adds a reusable `<AnnotationErrorBoundary>` and applies it to the Genome Nexus annotation components, so a render error in one annotation degrades that cell to a warning icon instead of replacing the entire portal with an error screen.

Addresses cBioPortal/cbioportal#12279. First of three PRs; see **Follow-ups** below.

## Why

The frontend currently has exactly one React error boundary in `appShell/App/Container.tsx`, which wraps the whole app shell. Its fallback replaces the entire page with `<ErrorScreen title={'Oh no! We encountered an error.'} />`.

That means a `TypeError` thrown while rendering a single annotation cell unmounts the portal. This is what happened the last time Genome Nexus changed its data model: the service was reachable and returning HTTP 200, but a field that cBioPortal expected was gone, and the component reading it through during render.

Network failure is already handled, `MobxPromise` has an `error` status, and the annotation columns already render `errorIcon(...)` when a fetch fails. What was unprotected is a successful response whose shape changed. This PR extends the existing degraded-cell affordance from *fetch* failures to *render* failures, reusing `errorIcon` rather than
introducing new UI.

## What changed

**New** `AnnotationErrorBoundary` in `cbioportal-frontend-commons`:

- catches render errors from one annotation and renders a caller-supplied fallback
- reports through an injected logger, registered once in `appBootstrapper.tsx` and forwarded to the existing `sendToLoggly` channel with `tag: 'ANNOTATION_RENDER_ERROR'`, so component failures are searchable separately from fetch failures
- de-duplicates by component name and error message, so an annotation broken on every row of a table produces one log entry rather than one per row

The logger is injected rather than imported because `sendToLoggly` is defined in `src/shared/lib/tracking.ts`, and packages cannot import from the application.

**Adopted** at three places:

- `GenericAnnotation` in `react-mutation-mapper` — each of the four sub-annotations is wrapped individually, so one failing annotation does not take the other three with it
- the Genome Nexus columns in `DefaultMutationTable` — HGVSc, gnomAD, ClinVar, dbSNP, SIGNAL
- the corresponding column formatters in `src/shared/components/mutationTable/column/`

Fallback messages read "could not be displayed", distinct from the existing "Error fetching ..." copy, so the two failure modes stay distinguishable in the UI.

## Notes for review

- **The fallback is a required prop**, so the boundary renders no UI of its own. `errorIcon` currently exists in two copies (`oncokb-frontend-commons` and `react-mutation-mapper`) and `cbioportal-frontend-commons` can import neither, so a built-in fallback would have meant a third copy. Each call site passes its own instead. Consolidating those two copies felt like unrelated cleanup for this PR, happy to do it separately if you'd prefer.
- **CIVIC is wrapped as well**, although it is not a Genome Nexus service. It is a sibling within the same annotation cell, so leaving it unwrapped would cause it to take down the Genome Nexus annotations next to it.
- **Boundaries are per-cell**, at the point where `errorIcon` already renders. A boundary that never trips costs one class-component instance and no extra DOM.

## Test plan

- [x] `pnpm --filter cbioportal-frontend-commons test` — new `AnnotationErrorBoundary.spec.tsx`, 5 tests: children render normally; a throwing child renders the fallback instead of propagating; a sibling annotation in the same cell still renders; the failure reaches the registered logger; the same failure across two boundaries is reported once
- [x] `pnpm run testPackages` — 210 passed, no new failures
- [x] `pnpm run testMain` — 2186 passed, no new failures
- [x] `npx tsc --noEmit` — clean
- [x] `prettier -c` on all changed files — clean

## Follow-ups

- Defensive field access in the Genome Nexus parsing layer. Error boundaries only catch render, lifecycle, and constructor errors, so async and event-handler paths need to be guarded separately, following the pattern already merged in #5651.
- The same treatment for the OncoKB components.

---
<a href="https://app.blacksmith.sh/cBioPortal/codesmith/cbioportal-frontend/pr/5697"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791291205&installation_model_id=19627&pr_number=5697&repository=cBioPortal%2Fcbioportal-frontend&return_to=https%3A%2F%2Fgithub.com%2FcBioPortal%2Fcbioportal-frontend%2Fpull%2F5697&signature=565239e80cd8c1f0f860f90bf90517c64c5d144fc013f71bf3a1fa5d949690d6"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


Prasad8830:
Hi @onursumer and @inodb, @dippindots approved the scope for this on Slack and mentioned he'd be away this week, so I'm reaching out to one of you for review.

This is the first of three PRs for cBioPortal/cbioportal#12279. 
***The background:*** The last Genome Nexus data model change removed a field that cBioPortal was reading, and because the only error boundary in the frontend wraps the entire app shell, the component that read it took down the whole portal rather than just its own cell. This PR contains those failures per annotation; the remaining two cover defensive field access in the parsing layer (which boundaries can't reach, since they don't catch async or event-handler errors) and the same treatment for OncoKB.

Two decisions in here would benefit from your opinion, since the next two PRs build on them:

1. **Boundary granularity.** I went per-cell, wrapping each sub-annotation in `GenericAnnotation`, since that's where the existing `errorIcon` affordance already lives. Per-column-formatter would be cheaper but coarser. You'll have a better sense of the render cost in the large mutation tables than I do.

2. **The fallback is a required prop rather than built into the boundary.** `errorIcon` currently exists in two copies: `oncokb-frontend-commons` and `react-mutation-mapper`, and `cbioportal-frontend-commons` can't import either, so building the fallback in would have meant a third copy. Consolidating those two felt out of scope for a resilience PR, but I'm happy to do it separately if you'd rather have one shared implementation.

Also worth a sanity check: degradation is visible (the existing red warning triangle, with "could not be displayed" wording to distinguish it from the existing "Error fetching" copy). If OncoKB or Genome Nexus changes its schema, that means a triangle on every row of the column. I think that's right, a blank cell reads as "no annotation exists", but it's a UX call and I'd rather you made it.

Happy to split this differently or rework the API if any of the above doesn't fit how you'd want it done.

src/appBootstrapper.tsx
```diff
@@ -171,6 +172,16 @@ const stores = {
 
 browserWindow.globalStores = stores;
 
+// annotation components live in packages that cannot import the tracking
```
alisman: we don't have loggly anymore so i think we can get rid of this.

Prasad8830: Thanks, removed in the latest commit.

One follow-up question: with Loggly gone, nothing registers a logger anymore, so the `setAnnotationErrorLogger` hook and the once-per-failure dedup inside `AnnotationErrorBoundary` are now unused. (I noticed `enableDataDogTracking` exists in `appBootstrapper.tsx` but isn't called anywhere, so I assumed there's no replacement service yet.)

Would you prefer I remove that logging code from the boundary as well, leaving it as a plain containment component, or keep the hook so a future logging service can plug in? Either way, React still prints each caught error to the console.

packages/cbioportal-frontend-commons/src/components/AnnotationErrorBoundary.tsx
```diff
@@ -0,0 +1,68 @@
...
+    constructor(props: AnnotationErrorBoundaryProps) {
+        super(props);
+        this.state = { hasError: false };
+    }
+
+    public static getDerivedStateFromError() {
+        return { hasError: true };
+    }
+
+    public componentDidCatch(error: Error) {
```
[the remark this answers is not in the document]
Prasad8830: The React 15 range in these peer dependencies is already inaccurate on master. `cbioportal-frontend-commons` uses fragments (`<>`, React 16.2+), and `react-mutation-mapper` and `oncokb-frontend-commons` use hooks (React 16.8+), so none of them runs on React 15 today.

Under React 15 the boundary would simply not catch anything, which is the current behavior, so this doesn't regress anything. Correcting the peer ranges changes the published package contract, so I'd prefer to do that in a separate PR if you'd like it done.

Prasad8830:
On the suppressed note in the Copilot review about `hasError` never resetting: that's accurate. `LazyMobXTable` keys rows by index, so after sorting or paging, a boundary that has already tripped can keep showing the warning icon on a different row until it remounts.

It only happens after a render failure, which is already the degraded path, so I've left it as is for now. If you'd like it handled in this PR, I can add an optional `resetKey` prop that clears the error when the row's data changes, with a regression test for recovery.

PR #5706, opened 2026-09-13
Guard OncoKB indicator access against missing fields

Prasad8830:
## Summary

Guards the OncoKB utilities and components against fields disappearing from OncoKB responses, so a data model change degrades the affected field instead of throwing.

Addresses cBioPortal/cbioportal#12279. This is the OncoKB counterpart of #5698 and is independent of both #5697 and #5698: none of the three PRs changes the same files, and they can merge in any order. Wrapping the OncoKB card in the error boundary from #5697 will follow as a separate PR once that one merges.

## Why

The motivating incident was a Genome Nexus response that was still HTTP 200 but no longer contained a field cBioPortal read. OncoKB clients are generated from OncoKB's own Swagger in the same way, so the same failure mode applies here.

The most exposed OncoKB code does not run during render, so no error boundary can contain it:

- `generateIdToIndicatorMap` builds the indicator map in `StoreUtils.ts`. It reads `indicator.query.id` for every indicator, so a single indicator without a `query` fails the whole map build, and every OncoKB annotation on the page is lost with it.
- `defaultOncoKbIndicatorFilter` runs in the mutation mapper stores.
- `isGermlineIndicator` and `isSomaticIndicator` are called by nearly every OncoKB consumer, in and out of render.

The OncoKB card is also rendered outside `GenericAnnotation` (for example, in `OtherBiomarkerAnnotation.tsx`), so the boundary from #5697 does not cover it on every path.

## What changed

**Outside render**

- `util/OncoKbUtils.ts`: `generateIdToIndicatorMap` skips an indicator that arrives without a query rather than failing the whole map. `defaultOncoKbIndicatorFilter` now guards `oncogenic` the same way the germline branch six lines above already guards `pathogenic`.
- `model/OncoKB.ts`: `isGermlineIndicator` and `isSomaticIndicator` treat a response without its query as somatic. That matches how they already treat a response without the germline flag.

**Render paths**

- `OncoKbCardBody.tsx`: the `props.indicator?.mutationEffect.knownEffect` chain, where the optional chain stopped one level short, plus the `mutationEffect`, `treatments`, `diagnosticImplications`, `prognosticImplications` and `query` reads that each tab passes down.
- `OncoKbTrackTooltip.tsx`: `query.tumorType` and `treatments.forEach` in the lollipop plot track, which is not rendered inside any boundary.
- `oncokbCard/EvidenceReferenceContent.tsx`: fixes a bug that exists on master independently of OncoKB changes. `citations` is typed as optional, but omitting it throws `Cannot read properties of undefined (reading 'pmids')`, even when a description is passed. JSX evaluates the props of both `If` branches before one is chosen, so `props.citations!.pmids` is read regardless of the condition. The citation arrays are now resolved up front. Without this, guarding the card would only have moved the crash one level down. `ImplicationContent` sends its per-row citations through this component, so those are covered too.
- `oncokbCard/BiologicalContent.tsx`: props made optional to match `EvidenceReferenceContent`.

## Notes for review

`OncoKbTreatmentTable` still reads per-row fields such as `treatment.drugs` and `treatment.level` without guards. I kept this PR to the top-level fields of the indicator plus the shared citations component, since the row internals render inside the card, which #5697 already contains in the mutation table. Happy to extend it to the rows here if you would prefer.

## Test plan

- [x] 16 new tests across 4 spec files (3 new: `OncoKbCardBody.spec.tsx`,
      `EvidenceReferenceContent.spec.tsx` and `OncoKbTrackTooltip.spec.tsx`). 11 feed responses
      with fields removed, and 5 confirm that fully populated responses behave as before
- [x] **Guards reverted and the suite re-run: all 11 missing-field tests fail**, confirming they
      catch the crashes rather than passing either way
- [x] The card is rendered for all five tab types (biological, sensitive, resistance, diagnostic,
      prognostic) with the nested fields removed
- [x] `pnpm run testPackages`, all passing
- [x] `pnpm run testMain`, 2203 passed, no failures
- [x] `npx tsc --noEmit`, clean
- [x] `prettier -c` on all changed files, clean

---
<a href="https://app.blacksmith.sh/cBioPortal/codesmith/cbioportal-frontend/pr/5706"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-dark-v2.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-with-codesmith-light-v2.svg">View with [code]smith</picture></a> <a href="https://backend.blacksmith.sh/track/enable-autofix?expires=1791885639&installation_model_id=19627&pr_number=5706&repository=cBioPortal%2Fcbioportal-frontend&return_to=https%3A%2F%2Fgithub.com%2FcBioPortal%2Fcbioportal-frontend%2Fpull%2F5706&signature=3fdb34bf592d1ec27000a98fa7203caf0c9653815b268ea455f3b7c9c0242cad"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/autofix-with-codesmith-light.svg">Autofix with [code]smith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith-bot</code> with what you need. Autofix is disabled.</sup>


alisman:
@Prasad8830 we want to find a more generic way of handling this.  Can you try to do it with React error boundaries.  Also, we are going to talk to the maintainers of these services to see why their contract changed.  

We just saw that you are already using error boundary for the annotation column.   Can we take similar approach here?

Prasad8830:
> @Prasad8830 we want to find a more generic way of handling this. Can you try to do it with React error boundaries. Also, we are going to talk to the maintainers of these services to see why their contract changed.
> 
> We just saw that you are already using error boundary for the annotation column. Can we take similar approach here?

Thanks, that makes sense, and I agree a boundary is the more generic fix. Before I rework this, here's what I'd do, so you can tell me if it matches what you had in mind.

Looking at it again, most OncoKB rendering already sits under the boundary from #5697. The OncoKB card opens from the OncoKB cell inside `GenericAnnotation`, which #5697 wraps, and React passes errors from the tooltip up to that boundary. Only two places render OncoKB outside it:

- `OtherBiomarkerAnnotation.tsx`, which renders `<OncoKB>` directly
- the lollipop plot's `OncoKbTrackTooltip`

So the reworked PR would wrap those two in `AnnotationErrorBoundary` and drop the per-field guards.

The parts a boundary can't catch are the indicator map builds (`generateIdToIndicatorMap`, and the `_.keyBy` in `DefaultMutationMapperDataFetcher`). Both run inside `remoteData`, so a throw there already ends up as the existing "Error fetching OncoKB data" icon rather than taking the page down. I'd leave those as they are.

Since this depends on the boundary from #5697, I'll wait for that one to merge and then rebuild this PR on master. I'll also split one unrelated bug I found along the way into its own small PR: `EvidenceReferenceContent` throws when its optional `citations` prop is omitted, even if a description is passed.

One question: should #5698 (the same defensive-guard approach, for the Genome Nexus parsing layer) go the same way, or would you still like the guards there?

